PR #6, opened 2023-09-07
Netsurf xp

francisbouvier:
- Closes https://github.com/Browsercore/browsercore/issues/40

### DIFF
--- a/test.html
+++ b/test.html
@@ -1,0 +1,8 @@
+<div id='content'>
+  <a id='link' href='foo'>OK</a>
+  <p id='para-empty'>
+    <span id='para-empty-child'></span>
+  </p>
+  <p id='para'> And</p>
+  <!--comment-->
+</div>


francisbouvier:
Netsurf as a dependancy on `libiconv` (optional but necessary if we want the parser to handle other encoding than unicode).

Currently on MacOS I'm using the version provided by brew and I'm linking statically to it (path `/opt/homebrew/opt/libiconv/lib/libiconv.a`).
TODO: Handle on Linux

Of course this is not the best way, we should build it locally.

francisbouvier:
[NetSurf experimentation](https://trello.com/c/yw3HJHAe/92-netsurf-experimentation)

francisbouvier:
My main concern with this PR is the build process of netsurf, especially the dependancy on `libiconv`.

I've hardcoded the libiconv path for MacOS (supposing an install through `brew` package manager) which is of course not ideal.
On Linux I let you @krichprollsch add the corresponding.

And if you have better idea, please feel free to change that. Otherwise we can just open a new backlog task to improve that later.

francisbouvier:
An other thing that could be improved in the future is the `wrapper.c` and `wrapper.h` source files.

It was easy to do it in C because`libdom` add already a working example of a function taking an HTML file, parsing, and returning a `dom_document`  that I've just adapted to take a char parameter with the file content.

Ideally we should instead port this function to zig in `netsurf.zig` source code.

francisbouvier:
At the end I've choose to use the internal API (`vendor/netsurf/<lib>/include`) instead of the public API (`vendor/netsurf/include`).

The main reason is to avoid opaque types and therefore alignment issues (the public API just define opaque struct). Plus we avoid all the `vtable` logic which seems to me at least unnecessary. But maybe I'm missing something here, we will have to check.

francisbouvier:
Especially we will need to check the memory layout to be sure that we do not create new pointers each time we call a getter/setter/method function on an existing object.

And check this reference counting stuff (especially if we choose to free objects that have been garbage collected in the JS world, see https://trello.com/c/2ioFr3Cq)

francisbouvier:
Finally I did not removed the lexbor source for now. In fact I choose to have a zig wrapper API for netsurf as close as possible than the existing lexbor wrapper.

We could imagine in the future to have a comptime interface and choose different parser/DOM implementation.

krichprollsch:
Browsercore/jsruntime-lib#67 adds libiconv install from sources

francisbouvier:
I'm deleting this PR in favor of https://github.com/francisbouvier/browsercore/pull/10

PR #10, opened 2023-09-25
Node netsurf

francisbouvier:


src/netsurf.zig
```diff
@@ -0,0 +1,373 @@
...
+pub const Title = c.dom_html_title_element;
+pub const Track = struct { base: c.dom_html_element };
+pub const UList = c.dom_html_u_list_element;
+pub const Video = struct { base: c.dom_html_element };
+
+// Document
+pub const Document = c.dom_document;
+
+pub inline fn documentGetElementById(doc: *Document, id: []const u8) ?*Element {
+    var elem: ?*Element = null;
```
krichprollsch: shouldn't be undefined?
I have no real idea, just asking 🤔

src/generate.zig
```diff
@@ -1,13 +1,33 @@
 const std = @import("std");
 const builtin = @import("builtin");
 
+// Utils
+// -----
+
+fn itoa(comptime i: u8) ![]u8 {
```
krichprollsch: Shouldn't we use a `![]const u8` as return type?

krichprollsch: And what about using a `i8` instead?

src/generate.zig
```diff
@@ -1,13 +1,33 @@
...
+    var len: usize = undefined;
+    if (i < 10) {
+        len = 1;
+    } else if (i < 100) {
+        len = 2;
+    } else {
+        return error.GenerateTooMuchMembers;
```
krichprollsch, line 14: In think if we accept an `i8`, we can handle the last case with a len=3.

src/html/document.zig
```diff
@@ -22,18 +21,19 @@ pub const HTMLDocument = struct {
         return parser.documentHTMLBody(self);
     }
 
-    pub fn _getElementById(self: *parser.DocumentHTML, id: []u8) ?*parser.HTMLElement {
-        const body_html = parser.documentHTMLBody(self);
-        const body_dom = @as(*parser.Element, @ptrCast(body_html));
-        const doc_dom = @as(*parser.Document, @ptrCast(self));
-        const elem_dom = Document.getElementById(doc_dom, body_dom, id);
-        return @as(*parser.HTMLElement, @ptrCast(elem_dom));
+    pub fn _getElementById(self: *parser.DocumentHTML, id: []u8) ?HTMLElem.Union {
```
krichprollsch, line 24: having a public function with a `_` prefix sounds weird :thinking:

src/generate.zig
francisbouvier: `[]u8` is more generic than `[]const u8`.
But in that case the function is intended to be called comptime (and will not work runtime) so anyway we will deal only with const strings, so you're right.

francisbouvier: In that function we know that we will deal only with positive value so I think it's more readable to keep an unsigned integer.

src/html/document.zig
francisbouvier: That's  just how `jsruntime` API works:
- `get_<name>` for getters
- `set_<name>` for setters
- `_<name>` for methods

And `pub` each time to be available.

Of course it's strange to have `_<name>` as `pub` and also those names are not Zig idiomatic.

At the beginning I was using `get<name>` and `set<Name>` but there were problems with some JS APIs. For example `window.setTimeout` is a method, not a setter.

Of course we can change that.

If you want you can open an Issue on `jsruntime` if you have some ideas.

src/generate.zig
krichprollsch: oh indeed,, my bad :facepalm:

src/netsurf.zig
francisbouvier: It works with `undefined` and I think you're right, it's more readable like that (and possibly more efficient as we do not "override" the value in case we found something).

The only thing is that we assume that the function called (here a netsurf C function) is going to initialize `elem` to `NULL` if it does not find anything. We have no guarantee of that.
Of course netsurf do that here (https://source.netsurf-browser.org/libdom.git/tree/src/core/document.c#n853).

But if it's not the case it will lead to a runtime error when we will try to access the value of `elem` (even to check if it's null).

Generally speaking I'm not using `undefined` if I'm not setting the value just afterward in the code.
But for this wrapper I think we can change to `undefined`.

PR #11, opened 2023-09-26
netsurf: move to public only API

francisbouvier:
And add some Node APIs:
- getters: firstChild, lastChild, nextSibling, previoussibling, parentNode, parentElement, nodeName, nodeType, ownerDocument, isConnected
- getters/setters: nodeValue, textContent
- methods: appendChild

And some test comptime optimizations on Document

### DIFF
--- a/src/netsurf.zig
+++ b/src/netsurf.zig
@@ -4,12 +4,50 @@ const c = @cImport({
     @cInclude("wrapper.h");
 });
 
-// Internal
+// Vtable
+// ------
 
+// netsurf libdom is using a vtable mechanism to handle the DOM tree heritage.
+// The vtable allow to select, from a parent, the right implementation of a
+// function for the child.
+
+// For example let's consider the following implementations of Node:
+// - Node <- CharacterData <- Text
+// - Node <- Element <- HTMLElement <- HTMLDivElement
+// If we take the `textContent` getter function on Node, the W3C standard says
+// that the result depends on the interface the Node implements, so
+// Node.textContent will be different depending if Node implements a Text or an
+// HTMLDivElement.
+// To handle that libdom provides a function on the child interface that
+// "override" the default parent function.
+// In this case there is a function dom_characterdata_get_text_content who
+// "override" parent function dom_node_get_text_content.
+// Like in an object-oriented language with heritage.
+// A vtable is attached to each "object" to retrieve the corresponding function.
+
+// NOTE: we can't use the high-level functions of libdom public API to get the
+// vtable as the public API defines only empty structs for each DOM interface,
+// which are translated by Zig as *const anyopaque with unknown alignement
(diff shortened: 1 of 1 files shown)


francisbouvier:
It's a big refacto to use exclusively the public API and the `vtable` mechanism. Why? Because the `vtable` mechanism allows us to go automatically from a parent to a child.

Let's take the `Node.textContent` getter as example.
[The DOM standard](https://dom.spec.whatwg.org/#dom-node-textcontent) says that the implementation of this getter depends on the concrete interface calling it. If it's an `Element` interface it's the concatenation of all children text content, if it's a `CharacterData` interface it's his data.
The `vtable` mechanism allow to go from a node and calls the right implementation:
- `_dom_node_get_text_content` for an `Element`
- `_dom_characterdata_get_text_content` for a `CharacterData`

Without the `vtable` mechanism we should need to do the switch case ourselves and call the corresponding internal function.

francisbouvier:
Unfortunately since zig-0.11 we can not call directly the high-level netsurf public API (in the previous example that will be `dom_node_get_text_content` without the `_` prefix).
It's a compile error alignment, since the netsurf public API defines types as void pointers who are translated in zig as `*const anyopaque` with alignment of 1.

To resolve this we create our own high-level function to retrieve the corresponding `vtable`. For example:

```zig

// generic comptime helper, with @alignCast operations
inline fn getVtable(comptime VtableT: type, comptime NodeT: type, node: anytype) VtableT {
    const node_algined: *align(@alignOf([*c]c.dom_node)) NodeT = @alignCast(node);
    const base = @as([*c]c.dom_node, @ptrCast(node_algined));
    const vtable_aligned: *align(@alignOf([*c]VtableT)) const anyopaque = @alignCast(base.*.vtable.?);
    return @as([*c]const VtableT, @ptrCast(vtable_aligned)).*;
}

// runtime vtable function
fn elementHTMLVtable(elem_html: *ElementHTML) c.dom_html_element_vtable {
    return getVtable(c.dom_html_element_vtable, ElementHTML, elem_html);
}
```

src/netsurf.zig
```diff
@@ -1,62 +1,17 @@
...
-    @cInclude("html/html_tablecol_element.h");
-    @cInclude("html/html_tablerow_element.h");
-    @cInclude("html/html_tablesection_element.h");
-    @cInclude("html/html_text_area_element.h");
-    @cInclude("html/html_title_element.h");
-    @cInclude("html/html_ulist_element.h");
-});
+// Internal
+
+inline fn getVtable(comptime VtableT: type, comptime NodeT: type, node: anytype) VtableT {
```
krichprollsch, line 40: I think you could set a comment explaining:
* why we need the vtable
* why this code is super obscure :sweat_smile:

PR #12, opened 2023-09-27
Node followup

francisbouvier:
- Closes https://github.com/Browsercore/browsercore/issues/35

### DIFF
--- a/src/dom/node.zig
+++ b/src/dom/node.zig
@@ -178,6 +178,12 @@ pub const Node = struct {
 
     pub fn _lookupPrefix(self: *parser.Node, namespace: ?[]const u8) ?[]const u8 {
         // TODO: other is not an optional parameter, but can be null.
+        if (namespace == null) {
+            return null;
+        }
+        if (std.mem.eql(u8, namespace.?, "")) {
+            return null;
+        }
         return parser.nodeLookupPrefix(self, namespace);
     }
 
--- a/src/netsurf.zig
+++ b/src/netsurf.zig
@@ -383,15 +383,9 @@ pub fn nodeIsSameNode(node: *Node, other: *Node) bool {
     return res;
 }
 
-pub fn nodeLookupPrefix(node: *Node, namespace: ?[]const u8) ?[]const u8 {
-    if (namespace == null) {
-        return null;
-    }
-    if (std.mem.eql(u8, namespace.?, "")) {
-        return null;
-    }
+pub fn nodeLookupPrefix(node: *Node, namespace: []const u8) ?[]const u8 {
     var s: ?*String = undefined;
-    _ = nodeVtable(node).dom_node_lookup_prefix.?(node, stringFromData(namespace.?), &s);
+    _ = nodeVtable(node).dom_node_lookup_prefix.?(node, stringFromData(namespace), &s);
     if (s == null) {
         return null;
     }


src/dom/node.zig
```diff
@@ -129,6 +129,75 @@ pub const Node = struct {
         const res = parser.nodeAppendChild(self, child);
         return Node.toInterface(res);
     }
+
+    pub fn _cloneNode(self: *parser.Node, deep: ?bool) Union {
+        const is_deep = if (deep) |deep_set| deep_set else false;
```
krichprollsch: I think we can write:

```suggestion
        const is_deep = deep orelse false;
```

src/dom/node.zig
```diff
@@ -129,6 +129,75 @@ pub const Node = struct {
...
+        const clone = parser.nodeCloneNode(self, is_deep);
+        return Node.toInterface(clone);
+    }
+
+    pub fn _compareDocumentPosition(self: *parser.Node, other: *parser.Node) void {
+        _ = other;
+        _ = self;
+        @panic("Not implemented node.compareDocumentPosition()");
```
krichprollsch: `panic`?  isn't that a problem? I mean, the running process will crash? :thinking:

src/dom/node.zig
```diff
@@ -129,6 +129,75 @@ pub const Node = struct {
...
+    pub fn _hasChildNodes(self: *parser.Node) bool {
+        return parser.nodeHasChildNodes(self);
+    }
+
+    pub fn _insertBefore(self: *parser.Node, new_node: *parser.Node, ref_node: *parser.Node) *parser.Node {
+        return parser.nodeInsertBefore(self, new_node, ref_node);
+    }
+
+    pub fn _isDefaultNamespace(self: *parser.Node, namespace: []const u8) bool {
+        // TODO: namespace is not an optional parameter, but can be null.
```
krichprollsch, line 164: I'm not sure I understand how we can get a `null` w/o optional with zig here :thinking:

src/netsurf.zig
```diff
@@ -340,6 +340,89 @@ pub fn nodeAppendChild(node: *Node, child: *Node) *Node {
     return res.?;
 }
 
+pub fn nodeCloneNode(node: *Node, is_deep: bool) *Node {
+    var res: ?*Node = undefined;
+    _ = nodeVtable(node).dom_node_clone_node.?(node, is_deep, &res);
```
krichprollsch, line 345: we are 100% confident the `nodeVtable(node).dom_node_clone_node` is not null?

src/netsurf.zig
```diff
@@ -340,6 +340,89 @@ pub fn nodeAppendChild(node: *Node, child: *Node) *Node {
...
+    return res;
+}
+
+pub fn nodeLookupPrefix(node: *Node, namespace: ?[]const u8) ?[]const u8 {
+    if (namespace == null) {
+        return null;
+    }
+    if (std.mem.eql(u8, namespace.?, "")) {
+        return null;
+    }
```
krichprollsch: shouldn't this logic be coded in `dom/node.zig` instead? this is not parser dependant isn't it?

src/dom/node.zig
francisbouvier: Yes it's temporary, just a way to know easily that it's not implemented.

The proper way should be returning an `error.NotImplemented` to jsruntime and then escalate that to in JS by throwing the error. But jsuntime does not support that for now (https://github.com/Browsercore/jsruntime-lib/issues/50).

But you right a panic will crash the test suite, I'm going to use something else like log error.

src/dom/node.zig
francisbouvier: We can't with builtins, we have to do something else like NullableTypes, see https://github.com/Browsercore/project/issues/42

src/netsurf.zig
francisbouvier: In C it's a pointer to a function, and the function exists in the source code, so yes we are 100% sure.

src/netsurf.zig
francisbouvier: I did ask my self the same question. In fact this shouldn't be there at all, if the implementation is good on libdom (and I have to reason to believe it's not until tested otherwise).

It's just some kind of optimization for an easy case, to avoid executing the rest of the code. I'm not sure it's really useful.

francisbouvier: But more generally this is a debate: should we put implementations stuff on the parser side or on the API side?

You are right on the fact that if it's not platform dependant it shouldn't be there.

But on the other side if we change the parser one day (and I think we will do it at some point), it's easier if we have no implementations at all in the API side.

In that case (ie. when we are dealing with possible values of input) I think you are right, we can settle to do it on the API side.

PR #27, opened 2023-09-28
CharacterData properties and methods

francisbouvier:
Attributes not implemented:
- methods:
  - after()
  - before()
  - remove()
  - replaceWith()


francisbouvier:
https://github.com/Browsercore/browsercore/issues/23

francisbouvier:
The not implemented methods depends on https://github.com/Browsercore/jsruntime-lib/issues/114
They are also not implemented in libdom characterdata API.

PR #51, opened 2023-10-03
WPT tests

### LINKED ISSUE #28: Use an existing DOM API test suite to ensure correctness
There is also xml tests cases on the libdom lib, maybe we can also leverage that ([libdom.it - Document Object Model Library](https://source.netsurf-browser.org/libdom.git/tree/test))

I don’t know if they have some links with the web platform tests.

---

We can test the browser against the web plateform tests.
https://web-platform-tests.org/index.html

The tests are standardized, it consists to a load an html file and it gives result.
Example: https://wpt.live/dom/attributes-are-nodes.html

Tests are grouped by RFC: for dom: https://wpt.live/dom/

‌

We can easily compare with major browsers results: https://wpt.fyi/results/dom?label=experimental&label=master&aligned

Servo uses them to track its improvements: https://wpt.servo.org/

The last thing: it’s a good metricsto show the browser capabilities in detail to users.

krichprollsch:
```
make wpt [<tests to run>]
```

```
make wpt Node-textContent.html Node-contains.html
```

```
make wpt tests/wpt/dom/nodes/
```

Closes #28

### DIFF
--- a/src/netsurf.zig
+++ b/src/netsurf.zig
@@ -669,16 +669,6 @@ pub inline fn documentCreateElement(doc: *Document, tag_name: []const u8) *Eleme
     return elem.?;
 }
 
-pub inline fn documentBody(doc: *Document) ?*Body {
-    const doc_html = @as(*DocumentHTML, @ptrCast(doc));
-    var body: ?*ElementHTML = undefined;
-    _ = documentHTMLVtable(doc_html).get_body.?(doc_html, &body);
-    if (body == null) {
-        return null;
-    }
-    return @as(*Body, @ptrCast(body.?));
-}
-
 // DocumentHTML
 pub const DocumentHTML = c.dom_html_document;
 


src/run_wpt.zig
```diff
@@ -0,0 +1,165 @@
...
+
+// browse the path to find the tests list.
+fn findWPTTests(allocator: std.mem.Allocator, path: []const u8) ![]const []const u8 {
+    var dir = try std.fs.cwd().openIterableDir(path, .{ .no_follow = true });
+    defer dir.close();
+
+    var walker = try dir.walk(allocator);
+    defer walker.deinit();
+
+    var tc = std.ArrayList([]const u8).init(allocator);
```
francisbouvier: I didn't run the code but here I would rather create the ArrayList in the calling function and take it as an arg in `findWPTTests` func.

src/run_wpt.zig
```diff
@@ -0,0 +1,165 @@
...
+    // read testharness.js content
+    libs[0] = try readFile(alloc, "tests/wpt/resources/testharness.js");
+    defer alloc.free(libs[0]);
+
+    // read testharnessreport.js content
+    libs[1] = try readFile(alloc, "tests/wpt/resources/testharnessreport.js");
+    defer alloc.free(libs[1]);
+
+    // browse the dir to get the tests dynamically.
+    const testcases = try findWPTTests(alloc, wpt_dir);
```
francisbouvier: I think it's better to use directly here `std.testing.allocator` has we don't want to have testcases generations allocations in benchmark.

src/run_wpt.zig
```diff
@@ -0,0 +1,165 @@
...
+    return error.EmptyTest;
+}
+
+// browse the path to find the tests list.
+fn findWPTTests(allocator: std.mem.Allocator, path: []const u8) ![]const []const u8 {
+    var dir = try std.fs.cwd().openIterableDir(path, .{ .no_follow = true });
+    defer dir.close();
+
+    var walker = try dir.walk(allocator);
+    defer walker.deinit();
```
francisbouvier: Maybe the pb is here. By de-allocating the walker before living the function

src/netsurf.zig
```diff
@@ -582,11 +582,17 @@ fn documentHTMLVtable(doc_html: *DocumentHTML) c.dom_html_document_vtable {
     return getVtable(c.dom_html_document_vtable, DocumentHTML, doc_html);
 }
 
-pub fn documentHTMLParse(filename: []const u8) *DocumentHTML {
-    var f: []u8 = @constCast(filename);
-    const doc = c.wr_create_doc_dom_from_file(f.ptr);
+pub fn documentHTMLParse(allocator: std.mem.Allocator, filename: []const u8) !*DocumentHTML {
```
francisbouvier: I think we can have 2 functions here:
- `documentHTMLParseFromString` which takes the buffer, ie. `[]const u8`
- `documentHTMLParseFromFile` which takes the filename, create the buffer and call `documentHTMLParseFromString`

krichprollsch:
current status for `dom/nodes/Node-*` (https://wpt.live/dom/nodes/)  tests
```
run test: error: Running WPT test suite
ERR	tests/wpt/dom/nodes/Node-isConnected-shadow-dom.html
	> ReferenceError: test is not defined
ERR	tests/wpt/dom/nodes/Node-baseURI.html
	> ReferenceError: test is not defined
ERR	tests/wpt/dom/nodes/Node-mutation-adoptNode.html
	> ReferenceError: test is not defined
ERR	tests/wpt/dom/nodes/ChildNode-after.html
	> SyntaxError: Invalid or unexpected token
OK	tests/wpt/dom/nodes/Node-insertBefore.html
	> undefined
ERR	tests/wpt/dom/nodes/append-on-Document.html
	> TypeError: Cannot read properties of undefined (reading 'createDocument')
ERR	tests/wpt/dom/nodes/Node-compareDocumentPosition.html
	> ReferenceError: testNodes is not defined
OK	tests/wpt/dom/nodes/Node-replaceChild.html
	> undefined
ERR	tests/wpt/dom/nodes/Document-createElement.html
	> SyntaxError: Unexpected end of input
ERR	tests/wpt/dom/nodes/Node-nodeName.html
	> ReferenceError: test is not defined
ERR	tests/wpt/dom/nodes/Node-removeChild.html
	> ReferenceError: creators is not defined
ERR	tests/wpt/dom/nodes/Node-appendChild.html
	> ReferenceError: test is not defined
ERR	tests/wpt/dom/nodes/Node-nodeValue.html
	> ReferenceError: test is not defined
ERR	tests/wpt/dom/nodes/Node-isConnected.html
	> ReferenceError: test is not defined
ERR	tests/wpt/dom/nodes/Node-isSameNode.html
	> ReferenceError: test is not defined
ERR	tests/wpt/dom/nodes/Node-parentElement.html
	> ReferenceError: test is not defined
ERR	tests/wpt/dom/nodes/Node-contains.html
	> ReferenceError: testNodes is not defined
ERR	tests/wpt/dom/nodes/Node-properties.html
	> ReferenceError: paras is not defined
ERR	tests/wpt/dom/nodes/Node-cloneNode-on-inactive-document-crash.html
	> ReferenceError: i is not defined
ERR	tests/wpt/dom/nodes/Node-cloneNode-external-stylesheet-no-bc.sub.html
	> SyntaxError: Unexpected end of input
ERR	tests/wpt/dom/nodes/Node-cloneNode-svg.html
	> SyntaxError: Invalid or unexpected token
ERR	tests/wpt/dom/nodes/Node-childNodes.html
	> SyntaxError: Unexpected end of input
ERR	tests/wpt/dom/nodes/Node-cloneNode-XMLDocument.html
	> ReferenceError: test is not defined
ERR	tests/wpt/dom/nodes/Node-parentNode.html
	> ReferenceError: test is not defined
ERR	tests/wpt/dom/nodes/Node-cloneNode-document-with-doctype.html
	> SyntaxError: Invalid or unexpected token
ERR	tests/wpt/dom/nodes/Node-normalize.html
	> SyntaxError: Invalid or unexpected token
ERR	tests/wpt/dom/nodes/Node-parentNode-iframe.html
	> error.EmptyTest
ERR	tests/wpt/dom/nodes/Node-isEqualNode.html
	> ReferenceError: test is not defined
ERR	tests/wpt/dom/nodes/Node-textContent.html
	> SyntaxError: Invalid or unexpected token
ERR	tests/wpt/dom/nodes/Node-lookupNamespaceURI.html
	> TypeError: document.createDocumentFragment is not a function
ERR	tests/wpt/dom/nodes/Node-cloneNode.html
	> SyntaxError: Unexpected end of input
ERR	tests/wpt/dom/nodes/Node-constants.html
	> ReferenceError: setup is not defined
30/32 tests failures
Segmentation fault at address 0x2
/usr/local/zig-0.11.0/lib/compiler_rt/memset.zig:19:14: 0x2df8380 in memset (compiler_rt)
            d[0] = c;
             ^
/usr/local/zig-0.11.0/lib/std/mem/Allocator.zig:313:26: 0xae377a in free__anon_173857 (test)
    @memset(non_const_ptr[0..bytes_len], undefined);
                         ^
/usr/local/zig-0.11.0/lib/std/array_list.zig:75:36: 0xabe818 in deinit (test)
                self.allocator.free(self.allocatedSlice());
                                   ^
/home/pierre/prs/brwsrcr/browsercore/src/run_wpt.zig:55:22: 0xabdb46 in test_0 (test)
    defer list.deinit();
                     ^
/usr/local/zig-0.11.0/lib/test_runner.zig:100:29: 0xae8fe9 in mainServer (test)
                test_fn.func() catch |err| switch (err) {
                            ^
/usr/local/zig-0.11.0/lib/test_runner.zig:34:26: 0xabf13a in main (test)
        return mainServer() catch @panic("internal test runner failure");
                         ^
/usr/local/zig-0.11.0/lib/std/start.zig:564:22: 0xabeb99 in main (test)
            root.main();
                     ^
???:?:?: 0x7f0bd30666c9 in ??? (libc.so.6)
Unwind information for `libc.so.6:0x7f0bd30666c9` was not available, trace may be incomplete
run test: error: while executing test 'test_0', the following command terminated with signal 6 (expected exited with code 0):
/home/pierre/prs/brwsrcr/browsercore/zig-cache/o/aa0066f6a2287c387d7b0dce8cad42f2/test --listen=- 
Build Summary: 2/4 steps succeeded; 1 failed; 1/1 tests passed (disable with --summary none)
wpt transitive failure
+- run test failure
/home/pierre/prs/brwsrcr/browsercore/zig-cache/o/a8e408efd07da54696f4d988c5590446/build /usr/local/zig-0.11.0/zig /home/pierre/prs/brwsrcr/browsercore /home/pierre/prs/brwsrcr/browsercore/zig-cache /home/pierre/.cache/zig wpt -Dengine=v8
```

francisbouvier:
I'm surprised by some of those errors.

For example `appendChild` is handled yet we have:

```
ERR	tests/wpt/dom/nodes/Node-appendChild.html
	> ReferenceError: test is not defined
```

src/netsurf.zig
krichprollsch: I'm creating a separate PR for clarity. 
https://github.com/Browsercore/browsercore/pull/52

krichprollsch:
New version of results.
I still have to digg in some results to understand and fix the failures.

```
Running WPT test suite
FAIL	tests/wpt/dom/nodes/Node-isConnected-shadow-dom.html
Node.isConnected in a closed shadow tree: Fail document.createElement is not a function
Node.isConnected in a open shadow tree: Fail document.createElement is not a function

FAIL	tests/wpt/dom/nodes/Node-baseURI.html
For elements belonging to document, baseURI should be document URL: Fail document.createElement is not a function
For elements unassigned to document, baseURI should be document URL: Fail document.createElement is not a function
For elements belonging to document fragments, baseURI should be document URL: Fail document.createDocumentFragment is not a function
For elements belonging to document fragments in document, baseURI should be document URL: Fail document.createDocumentFragment is not a function
For attributes unassigned to element, baseURI should be document URL: Fail document.createAttribute is not a function
For attributes in elements belonging to document, baseURI should be document URL: Fail document.createElement is not a function
For attributes in elements unassigned to document, baseURI should be document URL: Fail document.createElement is not a function
For attributes in elements belonging to document fragments, baseURI should be document URL: Fail document.createDocumentFragment is not a function
For attributes in elements belonging to document fragments in document, baseURI should be document URL: Fail document.createDocumentFragment is not a function

FAIL	tests/wpt/dom/nodes/Node-mutation-adoptNode.html
simple append of foreign div with text: Fail Cannot read properties of undefined (reading 'createHTMLDocument')

FAIL	tests/wpt/dom/nodes/ChildNode-after.html
TypeError: document.createComment is not a function
    at <anonymous>:155:21
FAIL	tests/wpt/dom/nodes/Node-insertBefore.html
ReferenceError: preInsertionValidateHierarchy is not defined
    at <anonymous>:2:1
FAIL	tests/wpt/dom/nodes/append-on-Document.html
TypeError: Cannot read properties of undefined (reading 'createDocument')
    at test_append_on_Document (<anonymous>:5:40)
    at <anonymous>:44:1
FAIL	tests/wpt/dom/nodes/Node-compareDocumentPosition.html
ReferenceError: testNodes is not defined
    at <anonymous>:4:1
FAIL	tests/wpt/dom/nodes/Node-replaceChild.html
Passing null to replaceChild should throw a TypeError.: Fail document.createElement is not a function
If child's parent is not the context node, a NotFoundError exception should be thrown: Fail document.createElement is not a function
If the context node is not a node that can contain children, a HierarchyRequestError exception should be thrown: Fail getNonParentNodes is not defined
If node is an inclusive ancestor of the context node, a HierarchyRequestError should be thrown.: Fail document.createElement is not a function
If the context node is a document, inserting a document or text node should throw a HierarchyRequestError.: Fail Cannot read properties of undefined (reading 'createHTMLDocument')
If the context node is a document, inserting a DocumentFragment that contains a text node or too many elements should throw a HierarchyRequestError.: Fail Cannot read properties of undefined (reading 'createHTMLDocument')
If the context node is a document (without element children), inserting a DocumentFragment that contains multiple elements should throw a HierarchyRequestError.: Fail Cannot read properties of undefined (reading 'createHTMLDocument')
If the context node is a document, inserting a DocumentFragment with an element if there already is an element child should throw a HierarchyRequestError.: Fail Cannot read properties of undefined (reading 'createHTMLDocument')
If the context node is a document, inserting a DocumentFragment with an element before the doctype should throw a HierarchyRequestError.: Fail Cannot read properties of undefined (reading 'createHTMLDocument')
If the context node is a document, inserting an element if there already is an element child should throw a HierarchyRequestError.: Fail Cannot read properties of undefined (reading 'createHTMLDocument')
If the context node is a document, inserting an element before the doctype should throw a HierarchyRequestError.: Fail Cannot read properties of undefined (reading 'createHTMLDocument')
If the context node is a document, inserting a doctype if there already is a doctype child should throw a HierarchyRequestError.: Fail Cannot read properties of undefined (reading 'createHTMLDocument')
If the context node is a document, inserting a doctype after the document element should throw a HierarchyRequestError.: Fail Cannot read properties of undefined (reading 'createHTMLDocument')
If the context node is a DocumentFragment, inserting a document or a doctype should throw a HierarchyRequestError.: Fail document.createDocumentFragment is not a function
If the context node is an element, inserting a document or a doctype should throw a HierarchyRequestError.: Fail document.createElement is not a function
Replacing a node with its next sibling should work (2 children): Fail document.createElement is not a function
Replacing a node with its next sibling should work (4 children): Fail document.createElement is not a function
Replacing a node with itself should not move the node: Fail document.createElement is not a function
If the context node is a document, inserting a new doctype should work.: Fail Cannot read properties of undefined (reading 'createHTMLDocument')
Replacing the document element with a DocumentFragment containing a single element should work.: Fail Cannot read properties of undefined (reading 'createHTMLDocument')
Replacing the document element with a DocumentFragment containing a single element and comments should work.: Fail Cannot read properties of undefined (reading 'createHTMLDocument')
Replacing the document element with a single element should work.: Fail Cannot read properties of undefined (reading 'createHTMLDocument')
replaceChild should work in the presence of mutation events.: Fail document.addEventListener is not a function
Replacing an element with a DocumentFragment should allow a child of the DocumentFragment to be found by Id.: Fail document.getElementsByTagName is not a function

FAIL	tests/wpt/dom/nodes/Document-createElement.html
TypeError: document.querySelector is not a function
    at <anonymous>:83:26
FAIL	tests/wpt/dom/nodes/Node-nodeName.html
For Element nodes, nodeName should return the same as tagName.: Fail document.createElementNS is not a function
For Text nodes, nodeName should return "#text".: Fail document.createTextNode is not a function
For Comment nodes, nodeName should return "#comment".: Fail document.createComment is not a function
For Document nodes, nodeName should return "#document".: Pass
For DocumentType nodes, nodeName should return the name.: Fail Cannot read properties of undefined (reading 'nodeName')
For DocumentFragment nodes, nodeName should return "#document-fragment".: Fail document.createDocumentFragment is not a function

FAIL	tests/wpt/dom/nodes/Node-appendChild.html
WebIDL tests: Pass
Appending to a leaf node.: Fail document.createTextNode is not a function
Appending a document: Fail frames is not defined
Adopting an orphan: Fail frames is not defined
Adopting a non-orphan: Fail frames is not defined

FAIL	tests/wpt/dom/nodes/Node-nodeValue.html
Text.nodeValue: Fail document.createTextNode is not a function
Comment.nodeValue: Fail document.createComment is not a function
ProcessingInstruction.nodeValue: Fail document.createProcessingInstruction is not a function
Element.nodeValue: Fail document.createElement is not a function
Document.nodeValue: Pass
DocumentFragment.nodeValue: Fail document.createDocumentFragment is not a function
DocumentType.nodeValue: Fail Cannot read properties of undefined (reading 'nodeValue')

FAIL	tests/wpt/dom/nodes/Node-isConnected.html
Test with ordinary child nodes: Fail document.createElement is not a function
Test with iframes: Fail document.createElement is not a function

FAIL	tests/wpt/dom/nodes/Node-isSameNode.html
doctypes should be compared on reference: Fail Cannot read properties of undefined (reading 'createDocumentType')
elements should be compared on reference (namespaced element): Fail document.createElementNS is not a function
elements should be compared on reference (namespaced attribute): Fail document.createElement is not a function
processing instructions should be compared on reference: Fail document.createProcessingInstruction is not a function
text nodes should be compared on reference: Fail document.createTextNode is not a function
comments should be compared on reference: Fail document.createComment is not a function
document fragments should be compared on reference: Fail document.createDocumentFragment is not a function
documents should be compared on reference: Fail Cannot read properties of undefined (reading 'createDocument')
attributes should be compared on reference: Fail document.createAttribute is not a function

FAIL	tests/wpt/dom/nodes/Node-parentElement.html
When the parent is null, parentElement should be null: Pass
When the parent is a document, parentElement should be null (doctype): Fail Cannot read properties of undefined (reading 'parentElement')
When the parent is a document, parentElement should be null (element): Fail Cannot read properties of undefined (reading 'parentElement')
When the parent is a document, parentElement should be null (comment): Fail document.createComment is not a function
parentElement should return null for children of DocumentFragments (element): Fail document.createDocumentFragment is not a function
parentElement should return null for children of DocumentFragments (text): Fail document.createDocumentFragment is not a function
parentElement should work correctly with DocumentFragments (element): Fail document.createDocumentFragment is not a function
parentElement should work correctly with DocumentFragments (text): Fail document.createDocumentFragment is not a function
parentElement should work correctly in disconnected subtrees (element): Fail document.createElement is not a function
parentElement should work correctly in disconnected subtrees (text): Fail document.createElement is not a function
parentElement should work correctly in a document (element): Fail document.createElement is not a function
parentElement should work correctly in a document (text): Fail document.createElement is not a function

FAIL	tests/wpt/dom/nodes/Node-contains.html
ReferenceError: testNodes is not defined
    at <anonymous>:4:1
FAIL	tests/wpt/dom/nodes/Node-properties.html
ReferenceError: paras is not defined
    at <anonymous>:30:26
FAIL	tests/wpt/dom/nodes/Node-cloneNode-on-inactive-document-crash.html
ReferenceError: i is not defined
    at <anonymous>:2:11
FAIL	tests/wpt/dom/nodes/Node-cloneNode-external-stylesheet-no-bc.sub.html
TypeError: Cannot read properties of undefined (reading 'createHTMLDocument')
    at <anonymous>:6:37
FAIL	tests/wpt/dom/nodes/Node-cloneNode-svg.html
TypeError: document.querySelector is not a function
    at <anonymous>:4:22
FAIL	tests/wpt/dom/nodes/Node-childNodes.html
Caching of Node.childNodes: Fail document.createElement is not a function
Node.childNodes on an Element.: Fail document.createElement is not a function
Node.childNodes on a DocumentFragment.: Fail document.createDocumentFragment is not a function
Node.childNodes on a Document.: Fail Illegal constructor
Iterator behavior of Node.childNodes: Fail document.createElement is not a function
Node.childNodes should be a live collection: Fail Cannot read properties of null (reading 'childNodes')

FAIL	tests/wpt/dom/nodes/Node-cloneNode-XMLDocument.html
Created with createDocument: Fail Cannot read properties of undefined (reading 'createDocument')

FAIL	tests/wpt/dom/nodes/Node-parentNode.html

FAIL	tests/wpt/dom/nodes/Node-cloneNode-document-with-doctype.html
Created with the createDocument/createDocumentType: Fail Cannot read properties of undefined (reading 'createDocumentType')
Created with the createHTMLDocument: Fail Cannot read properties of undefined (reading 'createHTMLDocument')
Created with DOMParser: Fail window.DOMParser is not a constructor

FAIL	tests/wpt/dom/nodes/Node-normalize.html
Untitled: Fail document.createDocumentFragment is not a function
Empty text nodes separated by a non-empty text node: Fail document.createElement is not a function
Empty text nodes: Fail document.createElement is not a function
Non-text nodes with empty textContent values.: Fail DOMParser is not defined

FAIL	tests/wpt/dom/nodes/Node-parentNode-iframe.html

FAIL	tests/wpt/dom/nodes/Node-isEqualNode.html
doctypes should be compared on name, public ID, and system ID: Fail Cannot read properties of undefined (reading 'createDocumentType')
elements should be compared on namespace, namespace prefix, local name, and number of attributes: Fail document.createElementNS is not a function
elements should be compared on attribute namespace, local name, and value: Fail document.createElement is not a function
processing instructions should be compared on target and data: Fail document.createProcessingInstruction is not a function
text nodes should be compared on data: Fail document.createTextNode is not a function
comments should be compared on data: Fail document.createComment is not a function
document fragments should not be compared based on properties: Fail document.createDocumentFragment is not a function
documents should not be compared based on properties: Fail Cannot read properties of undefined (reading 'createDocument')
node equality testing should test descendant equality too: Fail document.createElement is not a function

FAIL	tests/wpt/dom/nodes/Node-textContent.html
TypeError: Cannot read properties of undefined (reading 'forEach')
    at <anonymous>:92:11
FAIL	tests/wpt/dom/nodes/Node-lookupNamespaceURI.html
TypeError: document.createDocumentFragment is not a function
    at <anonymous>:15:21
FAIL	tests/wpt/dom/nodes/Node-cloneNode.html
createElement(a): Fail document.createElement is not a function
createElement(abbr): Fail document.createElement is not a function
createElement(acronym): Fail document.createElement is not a function
createElement(address): Fail document.createElement is not a function
createElement(area): Fail document.createElement is not a function
createElement(article): Fail document.createElement is not a function
createElement(aside): Fail document.createElement is not a function
createElement(audio): Fail document.createElement is not a function
createElement(b): Fail document.createElement is not a function
createElement(base): Fail document.createElement is not a function
createElement(bdi): Fail document.createElement is not a function
createElement(bdo): Fail document.createElement is not a function
createElement(bgsound): Fail document.createElement is not a function
createElement(big): Fail document.createElement is not a function
createElement(blockquote): Fail document.createElement is not a function
createElement(body): Fail document.createElement is not a function
createElement(br): Fail document.createElement is not a function
createElement(button): Fail document.createElement is not a function
createElement(canvas): Fail document.createElement is not a function
createElement(caption): Fail document.createElement is not a function
createElement(center): Fail document.createElement is not a function
createElement(cite): Fail document.createElement is not a function
createElement(code): Fail document.createElement is not a function
createElement(col): Fail document.createElement is not a function
createElement(colgroup): Fail document.createElement is not a function
createElement(data): Fail document.createElement is not a function
createElement(datalist): Fail assert_true: HTMLDataListElement is not supported expected true got false
createElement(dialog): Fail document.createElement is not a function
createElement(dd): Fail document.createElement is not a function
createElement(del): Fail document.createElement is not a function
createElement(details): Fail document.createElement is not a function
createElement(dfn): Fail document.createElement is not a function
createElement(dir): Fail assert_true: HTMLDirectoryElement is not supported expected true got false
createElement(div): Fail document.createElement is not a function
createElement(dl): Fail document.createElement is not a function
createElement(dt): Fail document.createElement is not a function
createElement(embed): Fail document.createElement is not a function
createElement(fieldset): Fail document.createElement is not a function
createElement(figcaption): Fail document.createElement is not a function
createElement(figure): Fail document.createElement is not a function
createElement(font): Fail assert_true: HTMLFontElement is not supported expected true got false
createElement(footer): Fail document.createElement is not a function
createElement(form): Fail document.createElement is not a function
createElement(frame): Fail assert_true: HTMLFrameElement is not supported expected true got false
createElement(frameset): Fail document.createElement is not a function
createElement(h1): Fail document.createElement is not a function
createElement(h2): Fail document.createElement is not a function
createElement(h3): Fail document.createElement is not a function
createElement(h4): Fail document.createElement is not a function
createElement(h5): Fail document.createElement is not a function
createElement(h6): Fail document.createElement is not a function
createElement(head): Fail document.createElement is not a function
createElement(header): Fail document.createElement is not a function
createElement(hgroup): Fail document.createElement is not a function
createElement(hr): Fail document.createElement is not a function
createElement(html): Fail document.createElement is not a function
createElement(i): Fail document.createElement is not a function
createElement(iframe): Fail document.createElement is not a function
createElement(img): Fail document.createElement is not a function
createElement(input): Fail document.createElement is not a function
createElement(ins): Fail document.createElement is not a function
createElement(isindex): Fail document.createElement is not a function
createElement(kbd): Fail document.createElement is not a function
createElement(label): Fail document.createElement is not a function
createElement(legend): Fail document.createElement is not a function
createElement(li): Fail document.createElement is not a function
createElement(link): Fail document.createElement is not a function
createElement(main): Fail document.createElement is not a function
createElement(map): Fail document.createElement is not a function
createElement(mark): Fail document.createElement is not a function
createElement(marquee): Fail document.createElement is not a function
createElement(meta): Fail document.createElement is not a function
createElement(meter): Fail document.createElement is not a function
createElement(nav): Fail document.createElement is not a function
createElement(nobr): Fail document.createElement is not a function
createElement(noframes): Fail document.createElement is not a function
createElement(noscript): Fail document.createElement is not a function
createElement(object): Fail document.createElement is not a function
createElement(ol): Fail document.createElement is not a function
createElement(optgroup): Fail document.createElement is not a function
createElement(option): Fail document.createElement is not a function
createElement(output): Fail document.createElement is not a function
createElement(p): Fail document.createElement is not a function
createElement(param): Fail assert_true: HTMLParamElement is not supported expected true got false
createElement(pre): Fail document.createElement is not a function
createElement(progress): Fail document.createElement is not a function
createElement(q): Fail document.createElement is not a function
createElement(rp): Fail document.createElement is not a function
createElement(rt): Fail document.createElement is not a function
createElement(ruby): Fail document.createElement is not a function
createElement(s): Fail document.createElement is not a function
createElement(samp): Fail document.createElement is not a function
createElement(script): Fail document.createElement is not a function
createElement(section): Fail document.createElement is not a function
createElement(select): Fail document.createElement is not a function
createElement(small): Fail document.createElement is not a function
createElement(source): Fail document.createElement is not a function
createElement(spacer): Fail document.createElement is not a function
createElement(span): Fail document.createElement is not a function
createElement(strike): Fail document.createElement is not a function
createElement(style): Fail document.createElement is not a function
createElement(sub): Fail document.createElement is not a function
createElement(summary): Fail document.createElement is not a function
createElement(sup): Fail document.createElement is not a function
createElement(table): Fail document.createElement is not a function
createElement(tbody): Fail document.createElement is not a function
createElement(td): Fail document.createElement is not a function
createElement(template): Fail document.createElement is not a function
createElement(textarea): Fail document.createElement is not a function
createElement(th): Fail document.createElement is not a function
createElement(time): Fail document.createElement is not a function
createElement(title): Fail document.createElement is not a function
createElement(tr): Fail document.createElement is not a function
createElement(tt): Fail document.createElement is not a function
createElement(track): Fail document.createElement is not a function
createElement(u): Fail document.createElement is not a function
createElement(ul): Fail document.createElement is not a function
createElement(var): Fail document.createElement is not a function
createElement(video): Fail document.createElement is not a function
createElement(unknown): Fail document.createElement is not a function
createElement(wbr): Fail document.createElement is not a function
createDocumentFragment: Fail document.createDocumentFragment is not a function
createTextNode: Fail document.createTextNode is not a function
createComment: Fail document.createComment is not a function
createElement with attributes: Fail document.createElement is not a function
createElementNS HTML: Fail document.createElementNS is not a function
createElementNS non-HTML: Fail document.createElementNS is not a function
createProcessingInstruction: Fail document.createProcessingInstruction is not a function
createAttribute: Fail document.createAttribute is not a function
createAttributeNS: Fail document.createAttributeNS is not a function
implementation.createDocumentType: Fail Cannot read properties of undefined (reading 'createDocumentType')
implementation.createDocument: Fail Cannot read properties of undefined (reading 'createDocument')
implementation.createHTMLDocument: Fail Cannot read properties of undefined (reading 'createHTMLDocument')
node with children: Fail document.createElement is not a function
Node with custom prototype: Fail document.createElement is not a function

FAIL	tests/wpt/dom/nodes/Node-constants.html
ReferenceError: testConstants is not defined
    at <anonymous>:11:1
PASS	tests/wpt/success.html
31/32 tests failures
run browsercore-wpt: error: the following command exited with error code 1:
/home/pierre/prs/brwsrcr/browsercore/zig-out/bin/browsercore-wpt 
Build Summary: 9/11 steps succeeded; 1 failed (disable with --summary none)
wpt transitive failure
+- run browsercore-wpt failure
/home/pierre/prs/brwsrcr/browsercore/zig-cache/o/08821c369d3b13b547554a79519d2305/build /usr/local/zig-0.11.0/zig /home/pierre/prs/brwsrcr/browsercore /home/pierre/prs/brwsrcr/browsercore/zig-cache /home/pierre/.cache/zig wpt -Dengine=v8 
```

francisbouvier:
Do we really need to run tests on DOM interfaces that have their HTML counterparts?

For example do we need to implement `Document.createElement` and `HTMLDocument.createElement`, knowing that we will only have `HTMLDocument` for now?
Of course at a later stage the goal is to have a full DOM implementation but I'm not sure we have to prioritize that.
WDYT?

src/netsurf.zig
```diff
@@ -636,12 +659,28 @@ pub inline fn documentGetElementById(doc: *Document, id: []const u8) ?*Element {
...
+}
+
 pub inline fn documentCreateElement(doc: *Document, tag_name: []const u8) *Element {
     var elem: ?*Element = undefined;
     _ = documentVtable(doc).dom_document_create_element.?(doc, stringFromData(tag_name), &elem);
     return elem.?;
 }
 
+pub inline fn documentBody(doc: *Document) ?*Body {
+    const dochtml = @as(*DocumentHTML, @ptrCast(doc));
```
francisbouvier:
```suggestion
    const doc_html = @as(*DocumentHTML, @ptrCast(doc));
```

src/dom/document.zig
```diff
@@ -15,20 +17,21 @@ pub const Document = struct {
...
-        return parser.documentGetElementById(self, id);
-    }
-
     // JS funcs
     // --------
 
-    pub fn get_body(_: *parser.Document) ?*parser.Body {
-        // TODO
-        return null;
+    pub fn get_body(self: *parser.Document) ?*HTMLBodyElement {
```
```suggestion
    pub fn get_body(self: *parser.Document) ?*parser.Body {
```

If the API defines a `Self` (which is the case of all our DOM implementation) we always have to return it, ie. the "real" object and not the "container" one.
Maybe we could add a check on that in jsruntime.

Because it's the pointer of the "real" type who actually points to the underlying data. And it's this pointer who will be stored in the internal field of the JS corresponding object.

src/dom/document.zig
```diff
@@ -15,20 +17,21 @@ pub const Document = struct {
...
+        const b = parser.documentBody(self) orelse null;
+        return @as(*HTMLBodyElement, @ptrCast(b));
```
francisbouvier: The "container" type has absolutely no data, it's size is 0, so here the pointer casting will produce garbage.

src/dom/document.zig
```diff
@@ -15,20 +17,21 @@ pub const Document = struct {
...
-    }
-
     // JS funcs
     // --------
 
-    pub fn get_body(_: *parser.Document) ?*parser.Body {
-        // TODO
-        return null;
+    pub fn get_body(self: *parser.Document) ?*HTMLBodyElement {
+        const b = parser.documentBody(self) orelse null;
```
```suggestion
        return parser.documentBody(self);
```

francisbouvier:
I propose that we adopt a convention for naming variables of DOM interfaces across all the implementation. Otherwise it can be a bit messy, with all the dom/html redundancy (I have myself used different styles...).

- All variables of DOM interfaces should be short but not 1 character (2-5).
- If the variable is without suffix it's the `dom` version, otherwise the suffix will be *snake_case* with the version name
- If the interface is an html tag name (and therefore as no dom counterparts) we can use it directly, without suffix

Examples:
- Node `node`
- Element `elem`
- HTMLElement `elem_html`
- Document `doc`
- HTMLDocument `doc_html`
- HTMLDivElement `div`

src/dom/document.zig
```diff
@@ -3,7 +3,9 @@ const std = @import("std");
 const parser = @import("../netsurf.zig");
 
 const Node = @import("node.zig").Node;
+const NodeUnion = @import("node.zig").Union;
```
francisbouvier: For now I have used the name `Nodes` and you use `NodeUnion`.
`Nodes` feel more *natural* but `NodeUnion` is more explicit about the underlying type so I think it's better.
WDYT?

In any case we have to adopt only 1 name.

src/dom/document.zig
```diff
@@ -15,20 +17,21 @@ pub const Document = struct {
...
-    pub fn get_body(_: *parser.Document) ?*parser.Body {
-        // TODO
-        return null;
+    pub fn get_body(self: *parser.Document) ?*HTMLBodyElement {
+        const b = parser.documentBody(self) orelse null;
+        return @as(*HTMLBodyElement, @ptrCast(b));
+    }
+
+    pub fn _getElementById(self: *parser.Document, id: []const u8) ?NodeUnion {
+        const e = parser.documentGetElementById(self, id) orelse return null;
```
francisbouvier, line 29: I'm not using enough the `orelse return null` pattern. You are right to do so, I think it's shorter while still being explicit.
I will adopt it myself as I think we will have this case very often in the implementation.

src/dom/element.zig
```diff
@@ -3,12 +3,18 @@ const std = @import("std");
 const parser = @import("../netsurf.zig");
 
 const Node = @import("node.zig").Node;
+const Union = @import("node.zig").Union;
```
francisbouvier:
```suggestion
const NodeUnion = @import("node.zig").Union;
```

src/dom/element.zig
```diff
@@ -3,12 +3,18 @@ const std = @import("std");
...
 
 pub const Element = struct {
     pub const Self = parser.Element;
     pub const prototype = *Node;
     pub const mem_guarantied = true;
 
+    pub fn toInterface(e: *parser.Element) Union {
+        const n = @as(*parser.Node, @ptrCast(e));
+        return Node.toInterface(n);
```
francisbouvier: Here we don't need to go through `Node.toInterface` as we already know we have an `Element` as input.
We can directly call `"elements.zig".toInterface` function.

src/netsurf.zig
```diff
@@ -228,6 +228,29 @@ pub const NodeType = enum(u4) {
     notation = c.DOM_NOTATION_NODE, // historical
 };
 
+// NodeList
+pub const NodeList = c.struct_dom_nodelist;
```
francisbouvier:
```suggestion
pub const NodeList = c.dom_nodelist;
```

I think it's better if we stick with the C names of libdom API.

src/netsurf.zig
```diff
@@ -228,6 +228,29 @@ pub const NodeType = enum(u4) {
...
+
+pub fn nodeListLength(nodeList: *NodeList) u32 {
+    var ln: u32 = undefined;
+    _ = c.dom_nodelist_get_length(nodeList, &ln);
+    return ln;
+}
+
+pub fn nodeListItem(nodeList: *NodeList, index: u32) ?*Node {
+    var n: [*c]c.struct_dom_node = undefined;
```
francisbouvier:
```suggestion
    var n: [*c]c.dom_node = undefined;
```

src/netsurf.zig
```diff
@@ -228,6 +228,29 @@ pub const NodeType = enum(u4) {
...
+    _ = c._dom_nodelist_item(nodeList, index, &n);
+
+    if (n == null) {
+        return null;
+    }
+
+    // cast [*c]c.struct_dom_node into *Node
+    const res: *Node = @ptrCast(n);
```
francisbouvier:
```suggestion
    const res: *Node = @ptrCast(n);
    return @as(*Node, @ptrCast(n));
```

src/main_wpt.zig
```diff
@@ -0,0 +1,276 @@
+const std = @import("std");
+
+const parser = @import("netsurf.zig");
+const jsruntime = @import("jsruntime");
+
+const public = @import("jsruntime");
```
francisbouvier: Here `public` is aliasing `jsruntime`. We can remove it and use only the `jsruntime` namespace who makes more sense at the browsercore level.

src/main_wpt.zig
```diff
@@ -0,0 +1,276 @@
...
+
+// FileLoader loads files content from the filesystem.
+const FileLoader = struct {
+    files: std.StringHashMap([]const u8) = undefined,
+    path: []const u8,
+    alloc: std.mem.Allocator,
+
+    const Self = @This();
+
+    fn new(alloc: std.mem.Allocator, path: []const u8) Self {
```
francisbouvier:
```suggestion
    fn init(alloc: std.mem.Allocator, path: []const u8) Self {
```

More zig idiomatic.

src/main_wpt.zig
```diff
@@ -0,0 +1,276 @@
...
+// generate APIs
+const apis = jsruntime.compile(DOM.Interfaces);
+
+// FileLoader loads files content from the filesystem.
+const FileLoader = struct {
+    files: std.StringHashMap([]const u8) = undefined,
+    path: []const u8,
+    alloc: std.mem.Allocator,
+
+    const Self = @This();
```
francisbouvier: I don't think the usage of `Self = @This()` is a good thing here. I know that the _zig-v8_ original autho used it everywhere but IMHO it's simpler and more explicit to use just the struct name `FileLoader`.

`Self = @This()` makes sense when using generics as we do not know the Type provided, but otherwise I prefer to use the actual Type.

src/main_wpt.zig
```diff
@@ -0,0 +1,276 @@
...
+const HTMLElem = @import("html/elements.zig");
+
+const wpt_dir = "tests/wpt";
+
+// generate APIs
+const apis = jsruntime.compile(DOM.Interfaces);
+
+// FileLoader loads files content from the filesystem.
+const FileLoader = struct {
+    files: std.StringHashMap([]const u8) = undefined,
```
francisbouvier: Why provide a default `files` value here, especially if this value is `undefined` and if we provide a initializer to instanciate the Type.

src/main_wpt.zig
```diff
@@ -0,0 +1,276 @@
...
+    path: []const u8,
+    alloc: std.mem.Allocator,
+
+    const Self = @This();
+
+    fn new(alloc: std.mem.Allocator, path: []const u8) Self {
+        return Self{
+            .path = path,
+            .alloc = alloc,
+            .files = std.StringHashMap([]const u8).init(alloc),
```
francisbouvier: It might be clearer to create the type first as declaration `const <Name> = std.StringHashMap([]const u8)` and then it as the type for files: `files: <Name>`.

src/main_wpt.zig
```diff
@@ -0,0 +1,276 @@
...
+    }
+    fn load(self: *Self, name: []const u8) !void {
+        const filename = try std.mem.concat(self.alloc, u8, &.{ self.path, name });
+        defer self.alloc.free(filename);
+        var file = try std.fs.cwd().openFile(filename, .{});
+        defer file.close();
+
+        const file_size = try file.getEndPos();
+        const content = try file.readToEndAlloc(self.alloc, file_size);
+        const namedup = try self.alloc.dupe(u8, name);
```
francisbouvier, line 49: Do we really have to duplicate (ie. store in the heap) the name?

I haven't check the `StringHashMap` implementation but they might do it inside. To be checked.

src/main_wpt.zig
```diff
@@ -0,0 +1,276 @@
...
+// tags containing javascript sources.
+// It loads first the js libs files.
+fn runWPT(arena: *std.heap.ArenaAllocator, f: []const u8, loader: *FileLoader) !jsruntime.JSResult {
+    const alloc = arena.allocator();
+
+    // document
+    const htmldoc = try parser.documentHTMLParseFromFileAlloc(alloc, f);
+    var doc = parser.documentHTMLToDocument(htmldoc);
+
+    // create JS env
```
francisbouvier, line 162: Is there any reason to load the env this way rather than using the `loadEnv` shortcut and providing an `ExecFn`?

src/main_wpt.zig
```diff
@@ -0,0 +1,276 @@
...
+
+    // add document object
+    try js_env.addObject(apis, doc, "document");
+
+    // alias global as self and window
+    try js_env.attachObject(try js_env.getGlobal(), "self", null);
+    try js_env.attachObject(try js_env.getGlobal(), "window", null);
+
+    var res = jsruntime.JSResult{};
+    var cbk_res = jsruntime.JSResult{
```
francisbouvier: If you are never checking the result of `cbk_res` there is no point of providing it to the `run` function, you could just use null. That's why it's an Optional.

src/main_wpt.zig
```diff
@@ -0,0 +1,276 @@
...
+        \\};
+        \\window.removeEventListener = function () {};
+    ;
+    try js_env.run(alloc, init, "init", &res, &cbk_res);
+    if (!res.success) {
+        return res;
+    }
+
+    // TODO load <script src> attributes instead of the static list.
+    try js_env.run(alloc, try loader.get("/resources/testharness.js"), "testharness.js", &res, &cbk_res);
```
francisbouvier: Re-using the same `res` pointer for each script run can lead to strange behavior as some fields could not be updated each time, for example we could have the wrong value for `res.result` or `res.stack`.
It will not be the case for `res.success` so it will not lead to a bug but that's an implementation detail in jsruntime/v8.
Generally speaking this is dangerous.

If you want to avoid some boilerplate you could create a wrapper around the run function to create a JSResult. That's what I've done in https://github.com/Browsercore/jsruntime-lib/commit/592d6a14d02efea2df20c70ccf58289b133e2eea
Maybe after all there it make sense to have this wrapper as an `js_env` method?

src/main_wpt.zig
```diff
@@ -0,0 +1,276 @@
...
+    // initialize VM JS lib.
+    const vm = jsruntime.VM.init();
+    defer vm.deinit();
+
+    // prepare libraries to load on each test case.
+    var loader = FileLoader.new(alloc, "tests/wpt");
+    defer loader.deinit();
+
+    // browse the dir to get the tests dynamically.
+    var list = std.ArrayList([]const u8).init(alloc);
```
francisbouvier, line 88: We could "easily" use a simple array here, as we know at comptime the number of test files. And therefore avoid the use of an allocator. Same in other functions on this file.

But as it is a test command, I'm perfectly fine with the fact to use allocator-based functions if this makes the code easier to develop.
This comment was just an FYI, no action needed.

krichprollsch: I asked the question myself, and I do prefer keeping the runtime listing.
It makes more sense to extract the file list at runtime IMO.

src/main_wpt.zig
krichprollsch: Hum, I guess this is way I had to write https://github.com/Browsercore/jsruntime-lib/pull/112 :/
I need to access directly the JSResult b/c I want to check and display the details in case of error.
I return the js result with `runWPT`.

That said, I don't really understand how I can return and use `res` w/o using the allocator to create it :thinking: `res` should be discarded once `runWPT` ends, isn't it?

src/main_wpt.zig
krichprollsch: I didn't find any way to create an `ExecFn` able to load a given test file :/

src/main_wpt.zig
krichprollsch: AFAIK we have to manage the key memory ourself:
> Builtin hashmap for strings as keys. Key memory is managed by the caller. Keys and values will not automatically be freed.
https://ziglang.org/documentation/0.11.0/std/#A;std:StringHashMap

src/main_wpt.zig
krichprollsch: The idea was to catch field usage w/o initialization.

By reading https://ziglang.org/documentation/0.11.0/#undefined I wanted the behavior:
> Using this value would be a bug. The value will be unused, or overwritten before being used."

src/main_wpt.zig
krichprollsch: Good catch!

src/netsurf.zig
krichprollsch: so `res` is useless I guess?

src/netsurf.zig
krichprollsch: It's not clear to me the difference between the two.
with my editor, I jump into a cimports which contains only `struct_dom_nodelist` declaration...

src/dom/element.zig
krichprollsch: I think I don't understand here, let discuss the point together

src/dom/document.zig
krichprollsch: I think I don't understand the difference between the real object and the container one...

src/dom/document.zig
krichprollsch: We stick with `NodeUnion`.

src/main_wpt.zig
krichprollsch: fixed :heavy_check_mark:

src/html/document.zig
```diff
@@ -20,21 +21,6 @@ pub const HTMLDocument = struct {
     pub fn get_body(self: *parser.DocumentHTML) ?*parser.Body {
```
francisbouvier, line 21: As discussed all the properties are in fact on dom/document, so we should me the body getter too.

src/html/document.zig
```diff
@@ -20,21 +21,6 @@ pub const HTMLDocument = struct {
...
-    }
-
-    pub fn _createElement(self: *parser.DocumentHTML, tag_name: []const u8) HTMLElem.Union {
-        const doc_dom = parser.documentHTMLToDocument(self);
-        const base = parser.documentCreateElement(doc_dom, tag_name);
-        return HTMLElem.toInterface(HTMLElem.Union, base);
-    }
 };
 
 // Tests
```
francisbouvier, line 26: IMHO there is no point to keep tests on html/document if all the properties are on dom/document and tested there.

src/main_wpt.zig
```diff
@@ -0,0 +1,277 @@
...
+const HTMLElem = @import("html/elements.zig");
+
+const wpt_dir = "tests/wpt";
+
+// generate APIs
+const apis = jsruntime.compile(DOM.Interfaces);
+
+// FileLoader loads files content from the filesystem.
+const FileLoader = struct {
+    files: std.StringHashMap([]const u8),
```
francisbouvier: We can create a declaration in the the `FileLoader` struct like `const FilesMap = std.StringHashMap([]const u8)` and then use it as a Type instead of generating here and then after at init.

src/main_wpt.zig
```diff
@@ -0,0 +1,277 @@
...
+    // If the test failed, return detailed logs intead of the simple status.
+    if (!std.mem.eql(u8, res.result, "Pass")) {
+        return try evalJS(js_env, alloc, "report.log", "teststatus");
+    }
+
+    // return the final result.
+    return res;
+}
+
+fn evalJS(env: jsruntime.Env, alloc: std.mem.Allocator, script: []const u8, name: ?[]const u8) !jsruntime.JSResult {
```
francisbouvier, line 255: Do you think we should provide this wrapper, quite common, at the engine level, as a `Env` method?

src/html/document.zig
krichprollsch: But I can't find `body` declared in DOM Document specs: https://dom.spec.whatwg.org/#document
Do you see it?

But I find it in `HTMLDocument` https://html.spec.whatwg.org/#the-document-object

src/main_wpt.zig
krichprollsch: IMO yes, it could be useful.
But we have to explain the caller is responsible to deinit the JSresult in comment.

src/html/document.zig
francisbouvier: You are right

PR #52, opened 2023-10-04
refacto netsurf parse document

krichprollsch:
relates to https://github.com/Browsercore/browsercore/pull/51#discussion_r1344264465

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -57,7 +57,7 @@ pub fn main() !void {
     defer arena.deinit();
 
     // document
-    doc = try parser.documentHTMLParseFromFile(arena.allocator(), "test.html");
+    doc = try parser.documentHTMLParseFromFileAlloc(arena.allocator(), "test.html");
     defer parser.documentHTMLClose(doc);
 
     // remove socket file of internal server
--- a/src/main_shell.zig
+++ b/src/main_shell.zig
@@ -38,7 +38,7 @@ pub fn main() !void {
     defer arena.deinit();
 
     // document
-    doc = try parser.documentHTMLParseFromFile(arena.allocator(), "test.html");
+    doc = try parser.documentHTMLParseFromFileAlloc(arena.allocator(), "test.html");
     defer parser.documentHTMLClose(doc);
 
     // create JS vm
--- a/src/run_tests.zig
+++ b/src/run_tests.zig
@@ -38,7 +38,7 @@ test {
     const apis = jsruntime.compile(DOM.Interfaces);
 
     // document
-    doc = try parser.documentHTMLParseFromFile(std.testing.allocator, "test.html");
+    doc = try parser.documentHTMLParseFromFileAlloc(std.testing.allocator, "test.html");
     defer parser.documentHTMLClose(doc);
 
     // create JS vm


src/netsurf.zig
```diff
@@ -553,11 +553,40 @@ fn documentHTMLVtable(doc_html: *DocumentHTML) c.dom_html_document_vtable {
...
 }
 
-pub fn documentHTMLParse(filename: []const u8) *DocumentHTML {
-    var f: []u8 = @constCast(filename);
-    const doc = c.wr_create_doc_dom_from_file(f.ptr);
+// documentHTMLParseFromFile reads the full document, loads the content in a
+// The allocator is required to create a null terminated string from filename.
+// The buffer is freed by the function.
+// The caller is responsible for closing the document.
+pub fn documentHTMLParseFromFile(allocator: std.mem.Allocator, filename: []const u8) !*DocumentHTML {
```
francisbouvier: We could remove the mandatory allocator by changing the filename type to `[:0]const u8`.

Maybe providing a `documentHTMLParseFromFileAlloc` helper.

src/netsurf.zig
```diff
@@ -553,11 +553,40 @@ fn documentHTMLVtable(doc_html: *DocumentHTML) c.dom_html_document_vtable {
     return getVtable(c.dom_html_document_vtable, DocumentHTML, doc_html);
 }
 
-pub fn documentHTMLParse(filename: []const u8) *DocumentHTML {
-    var f: []u8 = @constCast(filename);
-    const doc = c.wr_create_doc_dom_from_file(f.ptr);
+// documentHTMLParseFromFile reads the full document, loads the content in a
```
francisbouvier: _in a_? `*DocumentHTML`?

src/netsurf.zig
```diff
@@ -553,11 +553,40 @@ fn documentHTMLVtable(doc_html: *DocumentHTML) c.dom_html_document_vtable {
...
+
+    const doc = c.wr_create_doc_dom_from_file(cstr.ptr);
     if (doc == null) {
-        @panic("error parser");
+        return error.ParserError;
+    }
+    return @as(*DocumentHTML, @ptrCast(doc.?));
+}
+
+// documentHTMLParseFromCStrparses the given string.
```
francisbouvier: Space needed between _documentHTMLParseFromCStr_ and _parses_

src/netsurf.zig
```diff
@@ -553,11 +553,40 @@ fn documentHTMLVtable(doc_html: *DocumentHTML) c.dom_html_document_vtable {
...
+// The allocator is required to create a null terminated string.
+// The c string allocated is freed by the function.
+// The caller is responsible for closing the document.
+pub fn documentHTMLParseFromStr(allocator: std.mem.Allocator, str: [:0]const u8) !*DocumentHTML {
```
francisbouvier: Here we should not required an allocator and the call to `dupeZ` has _str_ is already a `[:0]const u8`

src/netsurf.zig
```diff
@@ -553,11 +553,40 @@ fn documentHTMLVtable(doc_html: *DocumentHTML) c.dom_html_document_vtable {
...
+    // create a null terminated c string.
+    const cstr = try allocator.dupeZ(u8, str);
+    defer allocator.free(cstr);
+
+    return documentHTMLParseFromCStr(cstr);
```
francisbouvier: I don't think `documentHTMLParseFromCStr` is required, we can directly call `c.wr_create_doc_dom_from_string(str.ptr)`

krichprollsch:
I'm rebasing

src/netsurf.zig
krichprollsch: Yes but having the zig signature is easier to understand and call IMO

krichprollsch: and we handle the `null` doc error also.

PR #58, opened 2023-10-10
make: add NDEBUG option to netsurf build

krichprollsch:
In order to remove the debug output from the parser.

```
$ make shell 
Building shell...
JS Repl
exit with Ctrl+D or "exit"
> 
Goodbye...
```

How to fix locally?
```
rm -fr vendor/netsurf/{lib,include}/* && make install-netsurf
```


francisbouvier:
FYI there is a `make clean-netsurf` command to remove existing netsurf build.

Issue #18, opened 2023-09-28
Document properties and methods


francisbouvier:
https://dom.spec.whatwg.org/#document

PR #64, opened 2023-10-25
document.getElementsByTagName

krichprollsch:
* [x] impement dom HTMLCollection methods and properties https://dom.spec.whatwg.org/#htmlcollection
* [ ] ~implement array like accessors (relates to https://github.com/Browsercore/jsruntime-lib/issues/126)~ post poned
* [ ] implement symbol iterator #65
* [x] Implement dom document  getElementsByTagName https://dom.spec.whatwg.org/#dom-document-getelementsbytagname

We can't simply use the libdom `dom_document_get_elements_by_tag_name` because it follows an old version of the specifications and returns a `NodeList`.
Since https://github.com/whatwg/dom/commit/190700b7c12ecfd3b5ebdb359ab1d6ea9cbf7749 the spec changed in favor of returning an HTMLCollection.

So I'm trying to re-implement the HTMLCollection in zig.

Relates to #18

### DIFF
--- a/src/netsurf.zig
+++ b/src/netsurf.zig
@@ -621,6 +621,11 @@ pub fn elementGetAttribute(elem: *Element, name: []const u8) ?[]const u8 {
     return stringToData(s.?);
 }
 
+// elementToNode is an helper to convert an element to a node.
+pub inline fn elementToNode(e: *Element) *Node {
+    return @as(*Node, @ptrCast(e));
+}
+
 // ElementHTML
 pub const ElementHTML = c.dom_html_element;
 
@@ -739,11 +744,6 @@ pub inline fn documentGetDocumentElement(doc: *Document) *Element {
     var elem: ?*Element = undefined;
     _ = documentVtable(doc).dom_document_get_document_element.?(doc, &elem);
     return elem.?;
-}
-
-pub inline fn documentGetDocumentNode(doc: *Document) *Node {
-    const res = documentGetDocumentElement(doc);
-    return @as(*Node, @ptrCast(res));
 }
 
 pub inline fn documentCreateElement(doc: *Document, tag_name: []const u8) *Element {
--- a/src/dom/document.zig
+++ b/src/dom/document.zig
@@ -45,9 +45,9 @@ pub const Document = struct {
     // That's why we reimplemented getElementsByTagName by using an
     // HTMLCollection in zig here.
     pub fn _getElementsByTagName(self: *parser.Document, tag_name: []const u8) HTMLCollection {
-        const root = parser.documentGetDocumentNode(self);
+        const root = parser.documentGetDocumentElement(self);
         return HTMLCollection{
-            .root = root,
+            .root = parser.elementToNode(root),
             .match = tag_name,
         };
     }


src/netsurf.zig
```diff
@@ -618,6 +621,40 @@ pub fn elementGetAttribute(elem: *Element, name: []const u8) ?[]const u8 {
...
+pub const HTMLCollection = c.dom_html_collection;
+
+pub fn HTMLCollectionLength(collection: *HTMLCollection) u32 {
+    var ln: u32 = undefined;
+    _ = c.dom_html_collection_get_length(collection, &ln);
+    return ln;
+}
+
+pub fn HTMLCollectionItem(collection: *HTMLCollection, index: u32) ?*Element {
+    var n: [*c]c.dom_node = undefined;
```
francisbouvier: Maybe we should create an alias for `c.dom_node`, like `NodeExt`.
We are already using it elsewhere and it will improve readability to use `?NodeExt` instead of `[*c]c.dom_node`.
If we do so it will be a non-public alias as it's only for an internal use in the wrapper, we do not want to expose it to the APIs.

src/dom/html_collection.zig
```diff
@@ -0,0 +1,198 @@
...
+    cur_node: *parser.Node = undefined,
+
+    // next iterates hover the DOM tree to return the next following node or
+    // null at the end.
+    fn get_next(root: *parser.Node, cur: *parser.Node) ?*parser.Node {
+        // TODO deinit next
+        var next = parser.nodeFirstChild(cur);
+        if (next != null) {
+            return next;
+        }
```
francisbouvier:
```suggestion
        if (parser.nodeFirstChild(cur)) |next| {
            return next;
        }
```

With this form we avoid a `var` that is shadowed afterwards

src/dom/html_collection.zig
```diff
@@ -0,0 +1,198 @@
...
+
+        // TODO deinit next
+        next = parser.nodeNextSibling(cur);
+        if (next != null) {
+            return next;
+        }
```
francisbouvier:
```suggestion
        if (parser.nodeNextSibling(cur)) |next| {
            return next;
        }
```

src/dom/html_collection.zig
```diff
@@ -0,0 +1,198 @@
...
+
+        // TODO deinit parent
+        var parent = parser.nodeParentNode(cur) orelse unreachable;
```
francisbouvier: How can we be sure this will never be reached?

src/dom/html_collection.zig
```diff
@@ -0,0 +1,198 @@
...
+        // TODO deinit lastchild
+        var lastchild = parser.nodeLastChild(parent);
+        var prev = cur;
+        while (prev != root and prev == lastchild) {
+            prev = parent;
+            // TODO deinit parent
+            parent = parser.nodeParentNode(cur) orelse unreachable;
+            // TODO deinit lastchild
+            lastchild = parser.nodeLastChild(parent);
+        }
```
francisbouvier: Even if I have no problem with this implementation I think the algorithm will be more robust with a functional approach.

src/dom/html_collection.zig
```diff
@@ -0,0 +1,198 @@
...
+    // TODO: nodes retrieved must be de-referenced.
+    pub fn get_length(self: *HTMLCollection, allocator: std.mem.Allocator) !u32 {
+        var len: u32 = 0;
+        var node: *parser.Node = self.root;
+        var ntype: parser.NodeType = undefined;
+
+        const imatch = try std.ascii.allocUpperString(allocator, self.match);
+        defer allocator.free(imatch);
+
+        var is_wildcard = std.mem.eql(u8, self.match, "*");
```
francisbouvier:
```suggestion
        const is_wildcard = std.mem.eql(u8, self.match, "*");
```

We do not seem to modify it afterwards

src/dom/html_collection.zig
```diff
@@ -0,0 +1,198 @@
...
+
+        return len;
+    }
+
+    pub fn _item(self: *HTMLCollection, allocator: std.mem.Allocator, index: u32) !?Union {
+        var i: u32 = 0;
+        var node: *parser.Node = self.root;
+        var ntype: parser.NodeType = undefined;
+
+        var is_wildcard = std.mem.eql(u8, self.match, "*");
```
francisbouvier:
```suggestion
        const is_wildcard = std.mem.eql(u8, self.match, "*");
```

We do not seem to modify it afterwards

src/dom/html_collection.zig
```diff
@@ -0,0 +1,198 @@
...
+
+    pub fn _namedItem(self: *HTMLCollection, allocator: std.mem.Allocator, name: []const u8) !?Union {
+        if (name.len == 0) {
+            return null;
+        }
+
+        var node: *parser.Node = self.root;
+        var ntype: parser.NodeType = undefined;
+
+        var is_wildcard = std.mem.eql(u8, self.match, "*");
```
francisbouvier:
```suggestion
        const is_wildcard = std.mem.eql(u8, self.match, "*");
```

src/netsurf.zig
```diff
@@ -731,6 +768,18 @@ pub inline fn documentGetElementsByTagName(doc: *Document, tagname: []const u8)
...
 }
 
+// documentGetDocumentElement returns the root document element.
+pub inline fn documentGetDocumentElement(doc: *Document) *Element {
+    var elem: ?*Element = undefined;
+    _ = documentVtable(doc).dom_document_get_document_element.?(doc, &elem);
+    return elem.?;
+}
+
+pub inline fn documentGetDocumentNode(doc: *Document) *Node {
```
francisbouvier: I'm still not sure if we should do provide this kind of shortcuts in wrapper.
Sure we don't want the APIs to deal with `ptrCast` but we could just provide a more generic `elementToNode` function and let the APIs do the conversion.

src/netsurf.zig
```diff
@@ -618,6 +621,40 @@ pub fn elementGetAttribute(elem: *Element, name: []const u8) ?[]const u8 {
     return stringToData(s.?);
 }
 
+// HTMLCollection
+
+pub const HTMLCollection = c.dom_html_collection;
```
francisbouvier: What is the purpose of having `HTMLCollection` wrapper around netsurf here as we do not use it in the API ?

krichprollsch: None, I forgot to remove it.

src/dom/html_collection.zig
krichprollsch: I rewrote the netsurf implementation in zig: http://source.netsurf-browser.org/libdom.git/tree/src/html/html_collection.c#n160

The depth first iteration si required by the spec:
> Returns the [element](https://dom.spec.whatwg.org/#concept-element) with index index from the [collection](https://dom.spec.whatwg.org/#concept-collection). The [elements](https://dom.spec.whatwg.org/#concept-element) are sorted in [tree order](https://dom.spec.whatwg.org/#concept-tree-order).
https://dom.spec.whatwg.org/#htmlcollection

I add a comment

src/dom/html_collection.zig
krichprollsch: Well the netsurf code is like it's not reachable as far as I can see.
But I guess I can double check and return null if the parent is missing.
WDYT?

krichprollsch: see http://source.netsurf-browser.org/libdom.git/tree/src/html/html_collection.c#n183

src/netsurf.zig
krichprollsch: well I removed this function b/c we didn't need it anymore...

src/netsurf.zig
krichprollsch: I think  I would need more details about that.

PR #68, opened 2023-11-13
ci: add CI workflows for tests

krichprollsch:
Relates to https://github.com/Browsercore/project/issues/45

I had to configure an organization GH Personal Access Token to be able to clone private submodules.
I followed: https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens#creating-a-fine-grained-personal-access-token

The token is stored in the CI secrets variables: https://github.com/Browsercore/browsercore/settings/secrets/actions

:warning: APT tests don't pass at all. So I added a `continue-on-error: true` configuration to force the job to pass instead.
Unfortunately GH desn't give a way to warning a test instead of fail or success. see https://github.com/actions/runner/issues/2347 for details.

I didn't found a better way to do that for now.
https://github.com/Browsercore/browsercore/pull/68/files#diff-f297e07b06710055241730fd4e30f4f97053e47d9a52e61234f4be3e7256bf41R57-R60

### DIFF
--- a/.github/workflows/zig-test.yml
+++ b/.github/workflows/zig-test.yml
@@ -11,6 +11,14 @@ on:
       - "src/**/*.zig"
       - "src/*.zig"
   pull_request:
+
+    # By default GH trigger on types opened, synchronize and reopened.
+    # see https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request
+    # Since we skip the job when the PR is in draft state, we want to force CI
+    # running when the PR is marked ready_for_review w/o other change.
+    # see https://github.com/orgs/community/discussions/25722#discussioncomment-3248917
+    types: [opened, synchronize, reopened, ready_for_review]
+
     branches:
       - main
     paths:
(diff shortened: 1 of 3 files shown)


.github/workflows/zig-test.yml
```diff
@@ -0,0 +1,62 @@
...
+          ln -s /usr/local/lib/lexbor vendor/lexbor
+
+          ln -s /usr/local/lib/libiconv vendor/libiconv
+
+          ln -s /usr/local/lib/netsurf/build vendor/netsurf/build
+          ln -s /usr/local/lib/netsurf/lib vendor/netsurf/lib
+          ln -s /usr/local/lib/netsurf/include vendor/netsurf/include
+
+      - name: zig build debug
+        run: zig build -Dengine=v8
```
francisbouvier, line 71: Same question, is it necessary?

krichprollsch: I'm not sure, see https://github.com/Browsercore/jsruntime-lib/pull/133#discussion_r1394010556

krichprollsch: Screenshot from 2023-11-15 11-47-40

The zig test is faster than the debug build. Maybe we could save a bit of time, but not sure.

PR #69, opened 2023-11-14
ci: create a docker image with browsercore deps

krichprollsch:
Relates to https://github.com/Browsercore/project/issues/45

This PR creates a docker container with pre-build browsercore dependencies.
It aims to improve speed for CI tests.

### DIFF
--- a/.github/workflows/build-deps.yml
+++ b/.github/workflows/build-deps.yml
@@ -3,7 +3,7 @@ name: build-deps
 on:
   push:
     branches:
-      - "master"
+      - "main"
     paths:
       - 'vendor/lexbor-src/**'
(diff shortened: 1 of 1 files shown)


.github/workflows/build-deps.yml
```diff
@@ -0,0 +1,60 @@
+name: build-deps
+
+on:
+  push:
+    branches:
+      - "master"
```
francisbouvier: Is it not `main` instead of `master`?

krichprollsch: Good catch, thanks!

krichprollsch: fixed :heavy_check_mark:

PR #71, opened 2023-11-17
document.getElementsByClassName

krichprollsch:
https://dom.spec.whatwg.org/#dom-document-getelementsbyclassname

Relates to https://github.com/Browsercore/browsercore/issues/18

### DIFF
--- a/src/dom/document.zig
+++ b/src/dom/document.zig
@@ -50,12 +50,12 @@ pub const Document = struct {
     // the spec changed to return an HTMLCollection instead.
     // That's why we reimplemented getElementsByTagName by using an
     // HTMLCollection in zig here.
-    pub fn _getElementsByTagName(self: *parser.Document, tag_name: []const u8) !collection.HTMLCollection {
+    pub fn _getElementsByTagName(self: *parser.Document, tag_name: []const u8) collection.HTMLCollection {
         const root = parser.documentGetDocumentElement(self);
         return collection.HTMLCollectionByTagName(parser.elementToNode(root), tag_name);
     }
 
-    pub fn _getElementsByClassName(self: *parser.Document, classNames: []const u8) !collection.HTMLCollection {
+    pub fn _getElementsByClassName(self: *parser.Document, classNames: []const u8) collection.HTMLCollection {
         const root = parser.documentGetDocumentElement(self);
         return collection.HTMLCollectionByClassName(parser.elementToNode(root), classNames);
     }


src/dom/html_collection.zig
```diff
@@ -5,22 +5,91 @@ const parser = @import("../netsurf.zig");
...
+    }
+};
+
+pub const MatchByTagName = struct {
+    // tag is used to select node against their name.
+    // tag comparison is case insensitive.
+    tag: []const u8,
+    is_wildcard: bool,
+
+    fn init(tag_name: []const u8) !MatchByTagName {
```
francisbouvier: Why do we need an error union here?

src/dom/html_collection.zig
```diff
@@ -5,22 +5,91 @@ const parser = @import("../netsurf.zig");
...
+        .match = Match{
+            .matchByTagName = try MatchByTagName.init(tag_name),
+        },
+    };
+}
+
+pub const MatchByClassName = struct {
+    classNames: []const u8,
+
+    fn init(classNames: []const u8) !MatchByClassName {
```
francisbouvier: Why do we need an error union here?

src/dom/html_collection.zig
```diff
@@ -5,22 +5,91 @@ const parser = @import("../netsurf.zig");
...
+
+    pub fn match(self: MatchByTagName, node: *parser.Node) bool {
+        return self.is_wildcard or std.ascii.eqlIgnoreCase(self.tag, parser.nodeName(node));
+    }
+};
+
+pub fn HTMLCollectionByTagName(root: *parser.Node, tag_name: []const u8) !HTMLCollection {
+    return HTMLCollection{
+        .root = root,
+        .match = Match{
```
francisbouvier: To avoid `match.match` we could rename this `matcher`or something like that?

src/dom/html_collection.zig
```diff
@@ -5,22 +5,91 @@ const parser = @import("../netsurf.zig");
...
+
+    fn init(tag_name: []const u8) !MatchByTagName {
+        return MatchByTagName{
+            .tag = tag_name,
+            .is_wildcard = std.mem.eql(u8, tag_name, "*"),
+        };
+    }
+
+    pub fn match(self: MatchByTagName, node: *parser.Node) bool {
+        return self.is_wildcard or std.ascii.eqlIgnoreCase(self.tag, parser.nodeName(node));
```
francisbouvier, line 39: So using this function we have the same result than `std.ascii.allocUpperString`without the allocator? Great

src/dom/html_collection.zig
```diff
@@ -5,22 +5,91 @@ const parser = @import("../netsurf.zig");
...
+pub const MatchByClassName = struct {
+    classNames: []const u8,
+
+    fn init(classNames: []const u8) !MatchByClassName {
+        return MatchByClassName{
+            .classNames = classNames,
+        };
+    }
+
+    pub fn match(self: MatchByClassName, node: *parser.Node) bool {
```
francisbouvier, line 61: Here we accept a `Node`, so we cast the Element to Node and then again in the function body we cast back the Node to Element
 Maybe we should accept `Element` instead and let the function do the cast to Node if necessary. Unless of course if there is a conceptual reason to manipulate nodes.

src/dom/html_collection.zig
krichprollsch: I forgot to clean this up since I didn't use the allocator anymore...

src/dom/html_collection.zig
krichprollsch: Yes, it's a lot better.

src/dom/html_collection.zig
krichprollsch: Well, one matcher use a node directly, the other the element.
In all cases we will have to cast to something else.
Since we browse nodes, I think passing a node to the matcher makes more sense.
But not sure...

src/dom/html_collection.zig
krichprollsch: I also rename `Match` into `Matcher`

src/dom/document.zig
```diff
@@ -49,12 +50,14 @@ pub const Document = struct {
     // the spec changed to return an HTMLCollection instead.
     // That's why we reimplemented getElementsByTagName by using an
     // HTMLCollection in zig here.
-    pub fn _getElementsByTagName(self: *parser.Document, tag_name: []const u8) HTMLCollection {
+    pub fn _getElementsByTagName(self: *parser.Document, tag_name: []const u8) !collection.HTMLCollection {
```
```suggestion
    pub fn _getElementsByTagName(self: *parser.Document, tag_name: []const u8) collection.HTMLCollection {
```

src/dom/document.zig
```diff
@@ -49,12 +50,14 @@ pub const Document = struct {
...
         const root = parser.documentGetDocumentElement(self);
-        return HTMLCollection{
-            .root = parser.elementToNode(root),
-            .match = tag_name,
-        };
+        return collection.HTMLCollectionByTagName(parser.elementToNode(root), tag_name);
+    }
+
+    pub fn _getElementsByClassName(self: *parser.Document, classNames: []const u8) !collection.HTMLCollection {
```
```suggestion
    pub fn _getElementsByClassName(self: *parser.Document, classNames: []const u8) collection.HTMLCollection {
```

PR #81, opened 2023-11-23
Add DOMException

### LINKED ISSUE #66: Add DOMException

francisbouvier:
Closes #66 

Needs https://github.com/Browsercore/jsruntime-lib/pull/130

### DIFF
--- a/src/dom/exceptions.zig
+++ b/src/dom/exceptions.zig
@@ -44,9 +44,10 @@ pub const DOMException = struct {
     pub const _DATA_CLONE_ERR = 25;
 
     // TODO: deinit
-    pub fn init(alloc: std.mem.Allocator, err: parser.DOMError, callerName: []const u8) anyerror!DOMException {
-        const errName = DOMException.name(err);
-        const str = switch (err) {
+    pub fn init(alloc: std.mem.Allocator, err: anyerror, callerName: []const u8) anyerror!DOMException {
+        const errCast = @as(parser.DOMError, @errSetCast(err));
+        const errName = DOMException.name(errCast);
+        const str = switch (errCast) {
             error.HierarchyRequest => try allocPrint(
                 alloc,
                 "{s}: Failed to execute '{s}' on 'Node': The new child element contains the parent.",
@@ -55,7 +56,7 @@ pub const DOMException = struct {
             error.NoError => unreachable,
             else => "", // TODO: implement other messages
         };
-        return .{ .err = err, .str = str };
+        return .{ .err = errCast, .str = str };
     }
 
     fn name(err: parser.DOMError) []const u8 {


src/netsurf.zig
```diff
@@ -274,6 +274,54 @@ pub const Tag = enum(u8) {
...
+    NotFound,
+    NotSupported,
+    InuseAttribute,
+    InvalidState,
+    Syntax,
+    InvalidModification,
+    Namespace,
+    InvalidAccess,
+    Validation,
+    TypeMismatch,
```
krichprollsch, line 297: You didn't set all existing `DOMException` on purpose?
https://webidl.spec.whatwg.org/#idl-DOMException

src/dom/exceptions.zig
```diff
@@ -0,0 +1,133 @@
...
+    pub const _NOT_FOUND_ERR = 8;
+    pub const _NOT_SUPPORTED_ERR = 9;
+    pub const _INUSE_ATTRIBUTE_ERR = 10;
+    pub const _INVALID_STATE_ERR = 11;
+    pub const _SYNTAX_ERR = 12;
+    pub const _INVALID_MODIFICATION_ERR = 13;
+    pub const _NAMESPACE_ERR = 14;
+    pub const _INVALID_ACCESS_ERR = 15;
+    pub const _VALIDATION_ERR = 16;
+    pub const _TYPE_MISMATCH_ERR = 17;
```
krichprollsch, line 36: I don't really understand why we need these constants here

src/netsurf.zig
```diff
@@ -274,6 +274,54 @@ pub const Tag = enum(u8) {
...
+        c.DOM_NOT_SUPPORTED_ERR => DOMError.NotSupported,
+        c.DOM_INUSE_ATTRIBUTE_ERR => DOMError.InuseAttribute,
+        c.DOM_INVALID_STATE_ERR => DOMError.InvalidState,
+        c.DOM_SYNTAX_ERR => DOMError.Syntax,
+        c.DOM_INVALID_MODIFICATION_ERR => DOMError.InvalidModification,
+        c.DOM_NAMESPACE_ERR => DOMError.Namespace,
+        c.DOM_INVALID_ACCESS_ERR => DOMError.InvalidAccess,
+        c.DOM_VALIDATION_ERR => DOMError.Validation,
+        c.DOM_TYPE_MISMATCH_ERR => DOMError.TypeMismatch,
+        else => DOMError.NoError,
```
krichprollsch: Are we sure we want to return a `NoError` when the error is not part of the switch, a `UnhandledError` or `UnknownError` would be better IMO.
Morevoer, I would add an `unreachable` case for `NoError` b/c we don't want to return an error in this case.

src/netsurf.zig
```diff
@@ -451,9 +499,10 @@ pub fn nodeSetTextContent(node: *Node, value: []const u8) void {
     _ = nodeVtable(node).dom_node_set_text_content.?(node, s);
 }
 
-pub fn nodeAppendChild(node: *Node, child: *Node) *Node {
+pub fn nodeAppendChild(node: *Node, child: *Node) DOMError!*Node {
     var res: ?*Node = undefined;
-    _ = nodeVtable(node).dom_node_append_child.?(node, child, &res);
+    const err = nodeVtable(node).dom_node_append_child.?(node, child, &res);
+    if (err != 0) return DOMExceptionError(err);
```
krichprollsch: What about having helper `if (isErr(err))` instead?
Or `DOMExceptionError` could return an optional, then we could write:

```zig
if (DOMExceptionError(err)) |err| return err;
```

krichprollsch: Event better, since `DOMError` is an error set, we can change `DOMExceptionError` signature to
```zig
fn DOMExceptionError(except: DOMException) DOMError!void {
```

Then we can write:
```zig
const err = nodeVtable(node).dom_node_append_child.?(node, child, &res);
try DOMExceptionError(err)
```

Or directly:
```zig
try DOMExceptionError(nodeVtable(node).dom_node_append_child.?(node, child, &res));
```

src/netsurf.zig
```diff
@@ -274,6 +274,54 @@ pub const Tag = enum(u8) {
...
+    InvalidModification,
+    Namespace,
+    InvalidAccess,
+    Validation,
+    TypeMismatch,
+};
+
+const DOMException = c.dom_exception;
+
+fn DOMExceptionError(except: DOMException) DOMError {
```
krichprollsch: Since this function will be called all along the file, I would try to find a shorter name.
Ideas: `domerr`, `domexc`, ...

src/dom/exceptions.zig
```diff
@@ -0,0 +1,133 @@
...
+            error.NotSupported => "NotSupportedError",
+            error.InuseAttribute => "InuseAttributeError",
+            error.InvalidState => "InvalidStateError",
+            error.Syntax => "SyntaxError",
+            error.InvalidModification => "InvalidModificationError",
+            error.Namespace => "NamespaceError",
+            error.InvalidAccess => "InvalidAccessError",
+            error.Validation => "ValidationError",
+            error.TypeMismatch => "TypeMismatchError",
+            error.NoError => unreachable,
```
krichprollsch, line 89: can't we use comptime to generate this switch?

src/dom/exceptions.zig
```diff
@@ -0,0 +1,133 @@
...
+            error.NotSupported => 9,
+            error.InuseAttribute => 10,
+            error.InvalidState => 11,
+            error.Syntax => 12,
+            error.InvalidModification => 13,
+            error.Namespace => 14,
+            error.InvalidAccess => 15,
+            error.Validation => 16,
+            error.TypeMismatch => 17,
+            error.NoError => unreachable,
```
krichprollsch, line 122: same question about comptime usage...

src/dom/exceptions.zig
```diff
@@ -0,0 +1,133 @@
+const std = @import("std");
+const allocPrint = std.fmt.allocPrint;
+
+const jsruntime = @import("jsruntime");
+const Case = jsruntime.test_utils.Case;
+const checkCases = jsruntime.test_utils.checkCases;
+
+const parser = @import("../netsurf.zig");
+
+pub const DOMException = struct {
```
krichprollsch, line 11:
```suggestion
// https://webidl.spec.whatwg.org/#idl-DOMException
pub const DOMException = struct {
```

src/netsurf.zig
francisbouvier: The others errors are not defined in netsurf https://source.netsurf-browser.org/libdom.git/tree/include/dom/core/exceptions.h#n24

I can add the others errors but we known that they are not going to be handle by netsurf code.

src/dom/exceptions.zig
francisbouvier: Do you mean do the switch at comptime or generate the switch at comptime ?

We can't do the switch at comptime as `err` is a runtime value here.

Generating the switch at comptime might be possible but I'm not sure about that. You can not generate functions at comptime. I think it will not be trivial.
However even if we do so there will be no performance gain as the switch itself will still be at runtime as the actual `err` is only known at runtime.

In my opinion a switch like that with strings literals is more readable and not so verbose.

src/dom/exceptions.zig
francisbouvier: We need them because they are static attributes as per the spec.

src/dom/exceptions.zig
francisbouvier: same answer 😄 

We can not use `@intFromErr` as error number are unique to the whole source code, not only the error set (unlike an enum). + they might change over time.

This switch is a bit verbose but easy to read.

francisbouvier: We could reference the constant values here instead of the actual value to avoid repetition, like `error.IndexSize => _INDEX_SIZE_ERR` but it will be less clear in my opinion.

src/netsurf.zig
francisbouvier: Yes that's a good idea, i will do that.

src/netsurf.zig
francisbouvier: See the implementation of the next comment.

src/netsurf.zig
francisbouvier: You are right

src/netsurf.zig
krichprollsch: I think it's better to setup the whole errors, even if we will not throw them for now.

src/dom/exceptions.zig
krichprollsch: I meant to generate the switch at comptime yes.
My goal is not a performance gain, but to simplify adding new exceptions.
I want to avoid duplicating the error name/const declarations.

But It's a best have more than a requirement. The spec will not introduce new errors often.

src/netsurf.zig
francisbouvier: Done

PR #96, opened 2023-11-28
Check DOMError

francisbouvier:
Closes https://github.com/Browsercore/browsercore/issues/83


francisbouvier:
The `documentHTMLClose` function returns an error `NotSupported`. For now I've choose to just log this error and ignore it, as it's at the end of the process.
But still there is something to check here.

PR #100, opened 2023-12-01
Add some element.*Attribute functions

krichprollsch:
Relates to https://github.com/Browsercore/browsercore/issues/20

### DIFF
--- a/src/netsurf.zig
+++ b/src/netsurf.zig
@@ -351,9 +351,7 @@ pub fn namedNodeMapItem(nnm: *NamedNodeMap, index: u32) !?*Attribute {
     const err = c._dom_namednodemap_item(nnm, index, &n);
     try DOMErr(err);
 
-    if (n == null) {
-        return null;
-    }
+    if (n == null) return null;
 
     // cast [*c]c.dom_node into *Attribute
     return @as(*Attribute, @ptrCast(n));
@@ -364,9 +362,7 @@ pub fn namedNodeMapGetNamedItem(nnm: *NamedNodeMap, qname: []const u8) !?*Attrib
     const err = c._dom_namednodemap_get_named_item(nnm, try stringFromData(qname), &n);
     try DOMErr(err);
 
-    if (n == null) {
-        return null;
-    }
+    if (n == null) return null;
 
     // cast [*c]c.dom_node into *Attribute
     return @as(*Attribute, @ptrCast(n));
@@ -386,9 +382,7 @@ pub fn namedNodeMapGetNamedItemNS(
     );
     try DOMErr(err);
 
-    if (n == null) {
-        return null;
-    }
+    if (n == null) return null;
 
     // cast [*c]c.dom_node into *Attribute
     return @as(*Attribute, @ptrCast(n));
@@ -409,9 +403,7 @@ pub fn namedNodeMapSetNamedItem(nnm: *NamedNodeMap, attr: *Attribute) !?*Attribu
     );
     try DOMErr(err);
 
-    if (n == null) {
-        return null;
-    }
+    if (n == null) return null;
 
     // cast [*c]c.dom_node into *Attribute
     return @as(*Attribute, @ptrCast(n));
@@ -426,9 +418,7 @@ pub fn namedNodeMapSetNamedItemNS(nnm: *NamedNodeMap, attr: *Attribute) !?*Attri
     );
     try DOMErr(err);
 
-    if (n == null) {
-        return null;
-    }
+    if (n == null) return null;
 
     // cast [*c]c.dom_node into *Attribute
     return @as(*Attribute, @ptrCast(n));
@@ -493,9 +483,7 @@ pub fn nodeListItem(nodeList: *NodeList, index: u32) !?*Node {
     const err = c._dom_nodelist_item(nodeList, index, &n);
     try DOMErr(err);
 
-    if (n == null) {
-        return null;
-    }
+    if (n == null) return null;
 
     // cast [*c]c.dom_node into *Node
     return @as(*Node, @ptrCast(n));
@@ -550,9 +538,8 @@ pub fn nodeNextElementSibling(node: *Node) !?*Element {
     var n = node;
     while (true) {
         const res = try nodeNextSibling(n);
-        if (res == null) {
-            return null;
-        }
+        if (res == null) return null;
+
         if (try nodeType(res.?) == .element) {
             return @as(*Element, @ptrCast(res.?));
         }
@@ -572,9 +559,8 @@ pub fn nodePreviousElementSibling(node: *Node) !?*Element {
     var n = node;
     while (true) {
         const res = try nodePreviousSibling(n);
-        if (res == null) {
-            return null;
-        }
+        if (res == null) return null;
+
         if (try nodeType(res.?) == .element) {
             return @as(*Element, @ptrCast(res.?));
         }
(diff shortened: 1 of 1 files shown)


src/dom/attribute.zig
```diff
@@ -0,0 +1,15 @@
...
+const Node = @import("node.zig").Node;
+const DOMException = @import("exceptions.zig").DOMException;
+
+// WEB IDL https://dom.spec.whatwg.org/#attr
+pub const Attr = struct {
+    pub const Self = parser.Attribute;
+    pub const prototype = *Node;
+    pub const mem_guarantied = true;
+
+    pub const Exception = DOMException;
```
francisbouvier: It's not mandatory to repeat the `Exception` here as `Attr` is a child of `EventTarget` who has it already.

src/dom/element.zig
```diff
@@ -2,16 +2,24 @@ const std = @import("std");
...
 
+const DOMException = @import("exceptions.zig").DOMException;
+
 // WEB IDL https://dom.spec.whatwg.org/#element
 pub const Element = struct {
     pub const Self = parser.Element;
     pub const prototype = *Node;
     pub const mem_guarantied = true;
 
+    pub const Exception = DOMException;
```
francisbouvier: Same here, we do not need the `Exception` declaration.

src/dom/namednodemap.zig
```diff
@@ -0,0 +1,80 @@
...
+const checkCases = jsruntime.test_utils.checkCases;
+
+const DOMException = @import("exceptions.zig").DOMException;
+
+// WEB IDL https://dom.spec.whatwg.org/#namednodemap
+pub const NamedNodeMap = struct {
+    pub const Self = parser.NamedNodeMap;
+    pub const mem_guarantied = true;
+
+    pub const Exception = DOMException;
```
francisbouvier, line 16: Here we need it as `NamedNodeMap` is not a child of `Event Target`.

src/netsurf.zig
```diff
@@ -336,6 +336,131 @@ fn DOMErr(except: DOMException) DOMError!void {
...
+}
+
+pub fn namedNodeMapItem(nnm: *NamedNodeMap, index: u32) !?*Attribute {
+    var n: [*c]c.dom_node = undefined;
+    const err = c._dom_namednodemap_item(nnm, index, &n);
+    try DOMErr(err);
+
+    if (n == null) {
+        return null;
+    }
```
francisbouvier: I think we could inline those kind of checks, as this a common pattern and the file is already quite verbose.

src/netsurf.zig
```diff
@@ -336,6 +336,131 @@ fn DOMErr(except: DOMException) DOMError!void {
...
+}
+
+pub fn namedNodeMapGetNamedItem(nnm: *NamedNodeMap, qname: []const u8) !?*Attribute {
+    var n: [*c]c.dom_node = undefined;
+    const err = c._dom_namednodemap_get_named_item(nnm, try stringFromData(qname), &n);
+    try DOMErr(err);
+
+    if (n == null) {
+        return null;
+    }
```
francisbouvier: Inliner

src/netsurf.zig
```diff
@@ -336,6 +336,131 @@ fn DOMErr(except: DOMException) DOMError!void {
...
+        nnm,
+        try stringFromData(namespace),
+        try stringFromData(localname),
+        &n,
+    );
+    try DOMErr(err);
+
+    if (n == null) {
+        return null;
+    }
```
francisbouvier: Inliner

src/netsurf.zig
```diff
@@ -336,6 +336,131 @@ fn DOMErr(except: DOMException) DOMError!void {
...
+
+pub fn namedNodeMapGetLength(nnm: *NamedNodeMap) !u32 {
+    var ln: u32 = undefined;
+    const err = c.dom_namednodemap_get_length(nnm, &ln);
+    try DOMErr(err);
+    return ln;
+}
+
+pub fn namedNodeMapItem(nnm: *NamedNodeMap, index: u32) !?*Attribute {
+    var n: [*c]c.dom_node = undefined;
```
francisbouvier, line 350: We already talked about creating an internal alias `NodeInternal` to avoid using c pointer. Let's do that in a separate PR.

src/netsurf.zig
```diff
@@ -336,6 +336,131 @@ fn DOMErr(except: DOMException) DOMError!void {
...
+    const err = c._dom_namednodemap_item(nnm, index, &n);
+    try DOMErr(err);
+
+    if (n == null) {
+        return null;
+    }
+
+    // cast [*c]c.dom_node into *Attribute
```
francisbouvier, line 356: I think once we have the `NodeInternal` type we can remove those comments who are not so useful.

src/netsurf.zig
```diff
@@ -336,6 +336,131 @@ fn DOMErr(except: DOMException) DOMError!void {
...
+
+pub fn namedNodeMapGetNamedItemNS(
+    nnm: *NamedNodeMap,
+    namespace: []const u8,
+    localname: []const u8,
+) !?*Attribute {
+    var n: [*c]c.dom_node = undefined;
+    const err = c._dom_namednodemap_get_named_item_ns(
+        nnm,
+        try stringFromData(namespace),
```
francisbouvier, line 379: I think we could try to shorter those functions as they are used a lot, like `strFromData` and `strToData`. It's not a lot but still 3 chars less and it's as clear as the full version I think.

Let's do that in a separate PR.

src/netsurf.zig
```diff
@@ -336,6 +336,131 @@ fn DOMErr(except: DOMException) DOMError!void {
...
+    if (n == null) {
+        return null;
+    }
+
+    // cast [*c]c.dom_node into *Attribute
+    return @as(*Attribute, @ptrCast(n));
+}
+
+// Convert a parser pointer to a public dom_node pointer.
+fn toDOMNode(comptime T: type, v: *T) [*c]c.dom_node {
```
francisbouvier, line 392: I think `toNodeInternal` is better to avoid ambiguity with `Node`?

src/netsurf.zig
```diff
@@ -336,6 +336,131 @@ fn DOMErr(except: DOMException) DOMError!void {
...
+    const err = c._dom_namednodemap_set_named_item(
+        nnm,
+        toDOMNode(Attribute, attr),
+        &n,
+    );
+    try DOMErr(err);
+
+    if (n == null) {
+        return null;
+    }
```
francisbouvier: Inliner

src/netsurf.zig
```diff
@@ -336,6 +336,131 @@ fn DOMErr(except: DOMException) DOMError!void {
...
+    const err = c._dom_namednodemap_set_named_item_ns(
+        nnm,
+        toDOMNode(Attribute, attr),
+        &n,
+    );
+    try DOMErr(err);
+
+    if (n == null) {
+        return null;
+    }
```
francisbouvier: Inliner

src/netsurf.zig
krichprollsch: but in the contrary it's a public version of node, not an internal one...

src/netsurf.zig
krichprollsch: https://github.com/Browsercore/browsercore/issues/108

src/netsurf.zig
krichprollsch: after discussion, let use `NodeExternal`

krichprollsch: https://github.com/Browsercore/browsercore/issues/109

src/netsurf.zig
krichprollsch: https://github.com/Browsercore/browsercore/issues/109

src/netsurf.zig
krichprollsch: https://github.com/Browsercore/browsercore/issues/109#issuecomment-1840430816

PR #104, opened 2023-12-04
Use fixed zig version

krichprollsch:
Relates to https://github.com/Browsercore/project/issues/93

Depends to https://github.com/Browsercore/jsruntime-lib/pull/151

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -6,14 +6,12 @@ const jsruntime_path = "vendor/jsruntime-lib/";
 const jsruntime = @import("vendor/jsruntime-lib/build.zig");
 const jsruntime_pkgs = jsruntime.packages(jsruntime_path);
 
-const zig_version = builtin.zig_version;
-
 /// Do not rename this constant. It is scanned by some scripts to determine
 /// which zig version to install.
 const recommended_zig_version = jsruntime.recommended_zig_version;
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -23,7 +23,36 @@ help:
 
 # $(ZIG) commands
 # ------------
-.PHONY: build build-release run run-release shell test bench
+.PHONY: build build-release run run-release shell test bench download-zig
+
+zig_version = $(shell grep 'recommended_zig_version = "' "vendor/jsruntime-lib/build.zig" | cut -d'"' -f2)
```
francisbouvier, line 28: I'm wondering if this make command has more sense in the jsruntime Makefile? Espacially if the recommanded version is set there. It's because the main lib (jsruntime) does use a minimal version that the upper project (browsercore) has to do so.

But of course as `browsercore` is the main project it can have sense to have the command here also.

WDYT?

krichprollsch: Well, I put it here indeed b/c `browsercore` is the main project and b/c this is just an opiniated "easy" way to download the recommended zig version, I don't think it's really useful in `jsruntimelib` by itself...

Also, I'm not sure browsercore will stick w/ the jsruntime version in the future. But not sure at all about that...

PR #112, opened 2023-12-05
netsurf: create alias NodeExternal

### LINKED ISSUE #109: netsurf: create an internal alias `NodeExternal` to avoid using c pointer
> We already talked about creating an internal alias `NodeExternal` to avoid using c pointer. Let's do that in a separate PR.

> I think once we have the NodeExternal type we can remove those comments who are not so useful.

_Originally posted by @francisbouvier in https://github.com/Browsercore/browsercore/pull/100#discussion_r1414114553_

krichprollsch:
fix #109

### DIFF
--- a/src/netsurf.zig
+++ b/src/netsurf.zig
@@ -336,6 +336,52 @@ fn DOMErr(except: DOMException) DOMError!void {
 // EventTarget
 pub const EventTarget = c.dom_event_target;
 
+// NodeType
+
+pub const NodeType = enum(u4) {
+    element = c.DOM_ELEMENT_NODE,
+    attribute = c.DOM_ATTRIBUTE_NODE,
+    text = c.DOM_TEXT_NODE,
+    cdata_section = c.DOM_CDATA_SECTION_NODE,
+    entity_reference = c.DOM_ENTITY_REFERENCE_NODE, // historical
+    entity = c.DOM_ENTITY_NODE, // historical
+    processing_instruction = c.DOM_PROCESSING_INSTRUCTION_NODE,
+    comment = c.DOM_COMMENT_NODE,
+    document = c.DOM_DOCUMENT_NODE,
+    document_type = c.DOM_DOCUMENT_TYPE_NODE,
+    document_fragment = c.DOM_DOCUMENT_FRAGMENT_NODE,
+    notation = c.DOM_NOTATION_NODE, // historical
+};
+
+// NodeList
+pub const NodeList = c.dom_nodelist;
+
+pub fn nodeListLength(nodeList: *NodeList) !u32 {
+    var ln: u32 = undefined;
+    const err = c.dom_nodelist_get_length(nodeList, &ln);
+    try DOMErr(err);
+    return ln;
+}
+
+pub fn nodeListItem(nodeList: *NodeList, index: u32) !?*Node {
+    var n: NodeExternal = undefined;
+    const err = c._dom_nodelist_item(nodeList, index, &n);
+    try DOMErr(err);
+    if (n == null) return null;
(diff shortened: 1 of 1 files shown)


src/netsurf.zig
```diff
@@ -479,16 +466,26 @@ pub fn nodeListLength(nodeList: *NodeList) !u32 {
...
     return @as(*Node, @ptrCast(n));
 }
 
+// NodeExternal is the libdom public representation of a Node.
+// Since we use the internal representation (dom_node_internal), we declare
+// here a private version useful for some netsurf function call.
+const NodeExternal = [*c]c.dom_node;
+
+// Convert a parser pointer to a NodeExternal pointer.
+fn toNodeExternal(comptime T: type, v: *T) NodeExternal {
```
francisbouvier: Here you could use `v: anytype` but I'm not sure what is better. anytype is shorter but your version in more explicit.

src/netsurf.zig
```diff
@@ -479,16 +466,26 @@ pub fn nodeListLength(nodeList: *NodeList) !u32 {
...
     if (n == null) return null;
 
-    // cast [*c]c.dom_node into *Node
     return @as(*Node, @ptrCast(n));
 }
 
+// NodeExternal is the libdom public representation of a Node.
+// Since we use the internal representation (dom_node_internal), we declare
+// here a private version useful for some netsurf function call.
+const NodeExternal = [*c]c.dom_node;
```
francisbouvier: I would have put this before declaring `Node`, so we do not use types that haven't been yet declared.

src/netsurf.zig
```diff
@@ -347,24 +347,22 @@ pub fn namedNodeMapGetLength(nnm: *NamedNodeMap) !u32 {
 }
 
 pub fn namedNodeMapItem(nnm: *NamedNodeMap, index: u32) !?*Attribute {
-    var n: [*c]c.dom_node = undefined;
+    var n: NodeExternal = undefined;
     const err = c._dom_namednodemap_item(nnm, index, &n);
     try DOMErr(err);
```
francisbouvier: Do we need those line returns now?

src/netsurf.zig
```diff
@@ -347,24 +347,22 @@ pub fn namedNodeMapGetLength(nnm: *NamedNodeMap) !u32 {
...
 
     if (n == null) return null;
```
francisbouvier, line 400: Same here, do we need those line returns now?

src/netsurf.zig
krichprollsch: IMO it's not the same: using `v: *T` ensures `v` is a pointer to a value of type `T`.

If I use `v: anytype` I would be allowed to pass anything as `v` no?

src/netsurf.zig
krichprollsch: I'm not sure to understand, `pub const Node = c.dom_node_internal;` is declared after, see https://github.com/Browsercore/browsercore/pull/112/files#diff-eff26b7fb15b686657ba5f430ad7ba96d4a1842a6d51efc0e804f9033ffc66f9L493

src/netsurf.zig
francisbouvier: Well in anycase not passing a ptr will result a compile error, as `@alignCast` requires a ptr. At the end it's the same, your version is more explicit, the anytype version is more concise, I'm OK with boh.

src/netsurf.zig
francisbouvier: Yes sorry. It's more the namednodemap that could be after but let's keep it like that.

krichprollsch: moved

PR #118, opened 2023-12-07
Next Dom document implementation

### LINKED ISSUE #26: CDATASection properties and methods

krichprollsch:
* [x] createText
* [x] createCDATASection
* [x] createComment
* [x] createProcessingInstruction
* [x] importNode
* [x] adoptNode
* [x] createAttribute
* [x] createAttributeNS

Relates to #18 
fix #26

### DIFF
--- a/src/dom/document.zig
+++ b/src/dom/document.zig
@@ -257,22 +257,26 @@ pub fn testExecFn(
     try checkCases(js_env, &new);
 
     var createDocumentFragment = [_]Case{
-        .{ .src = "document.createDocumentFragment()", .ex = "[object DocumentFragment]" },
+        .{ .src = "var v = document.createDocumentFragment()", .ex = "undefined" },
+        .{ .src = "v.nodeName", .ex = "#document-fragment" },
     };
     try checkCases(js_env, &createDocumentFragment);
 
     var createTextNode = [_]Case{
-        .{ .src = "document.createTextNode('foo')", .ex = "[object Text]" },
+        .{ .src = "var v = document.createTextNode('foo')", .ex = "undefined" },
+        .{ .src = "v.nodeName", .ex = "#text" },
     };
     try checkCases(js_env, &createTextNode);
 
     var createCDATASection = [_]Case{
-        .{ .src = "document.createCDATASection('foo')", .ex = "[object CDATASection]" },
+        .{ .src = "var v = document.createCDATASection('foo')", .ex = "undefined" },
+        .{ .src = "v.nodeName", .ex = "#cdata-section" },
     };
     try checkCases(js_env, &createCDATASection);
 
     var createComment = [_]Case{
-        .{ .src = "document.createComment('foo')", .ex = "[object Comment]" },
+        .{ .src = "var v = document.createComment('foo')", .ex = "undefined" },
+        .{ .src = "v.nodeName", .ex = "#comment" },
     };
     try checkCases(js_env, &createComment);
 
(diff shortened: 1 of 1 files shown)


src/dom/character_data.zig
```diff
@@ -10,12 +10,16 @@ const parser = @import("../netsurf.zig");
...
 const Text = @import("text.zig").Text;
+const CDATASection = @import("cdata_section.zig").CDATASection;
+const ProcessingInstruction = @import("processing_instruction.zig").ProcessingInstruction;
 const HTMLElem = @import("../html/elements.zig");
 
 // CharacterData interfaces
 pub const Interfaces = generate.Tuple(.{
     Comment,
     Text,
+    CDATASection,
```
francisbouvier: As `CDATASection` is a child of `Text` I think it's better to create a `pub const Interfaces = generate.Tuple` in text.zig and add it there as `Text.Interfaces`.

src/dom/node.zig
```diff
@@ -54,6 +54,8 @@ pub const Node = struct {
             ),
             .comment => .{ .Comment = @as(*parser.Comment, @ptrCast(node)) },
             .text => .{ .Text = @as(*parser.Text, @ptrCast(node)) },
+            .cdata_section => .{ .CDATASection = @as(*parser.CDATASection, @ptrCast(node)) },
+            .processing_instruction => .{ .ProcessingInstruction = @as(*parser.ProcessingInstruction, @ptrCast(node)) },
```
francisbouvier, line 58: To be consistent with the hierarchy of the inheritance we could create a `CData.toInterface` function and handle there all the cdata children cases (text, cdata_section, processing_instruction, comment).
But maybe it's a unnecessary overhead.
I don't know if we have the case where this `CData.toInterface` will be called directly (if a function returns a CDataUnion`).
WDYT?

src/dom/document.zig
```diff
@@ -226,6 +261,44 @@ pub fn testExecFn(
     };
     try checkCases(js_env, &createDocumentFragment);
 
+    var createTextNode = [_]Case{
+        .{ .src = "document.createTextNode('foo')", .ex = "[object Text]" },
```
francisbouvier: Maybe it will be better to use the `nodeName` getter for the equality (here `#text`) rater than the string representation.

I see that the console of Chrome and Firefox returns different values here: just the value of the text itself (ie. `foo`) for Chrome and the concatenation of the nodeName and the text for Firefox (`#text "foo"`).
But when you call the `toString` function you have `[object Text]` so I guess it's only the console of the dev tools who change the result printed to help web developers with a more explicit return.

So I guess the object string representation `[object Text]` is OK for equality at the JS VM level, but the nodeName could be better (but more verbose). Sincerly I'm not sure. WDYT?

src/dom/node.zig
krichprollsch: >I don't know if we have the case where this CData.toInterface will be called directly (if a function returns a CDataUnion`).

Good question, I don't think so to be honest.

If I create a  `CData.toInterface` I will still have to handle the switch cases in the `Node.toInterface` to be complete...

src/dom/document.zig
krichprollsch: I wasn't really happy with this string representation, I'm ok to use nodeName instead :+1:

francisbouvier: You can also do it in 1 line `document.createTextNode('foo').nodeName` but I'm OK with this version

Issue #20, opened 2023-09-28
HTMLElement properties and methods


francisbouvier:
https://html.spec.whatwg.org/#htmlelement

+ specifics like anchor...

PR #133, opened 2023-12-13
dom: implement NonDocumentTypeChildNode mixin for element

krichprollsch:
* [x] readonly attribute [Element](https://dom.spec.whatwg.org/#element)? [previousElementSibling](https://dom.spec.whatwg.org/#dom-nondocumenttypechildnode-previouselementsibling);
* [x]  readonly attribute [Element](https://dom.spec.whatwg.org/#element)? [nextElementSibling](https://dom.spec.whatwg.org/#dom-nondocumenttypechildnode-nextelementsibling);

relates with #20

### DIFF
--- a/src/dom/element.zig
+++ b/src/dom/element.zig
@@ -181,17 +181,13 @@ pub const Element = struct {
     // https://dom.spec.whatwg.org/#interface-nondocumenttypechildnode
     pub fn get_previousElementSibling(self: *parser.Element) !?Union {
         const res = try parser.nodePreviousElementSibling(parser.elementToNode(self));
-        if (res == null) {
-            return null;
-        }
+        if (res == null) return null;
         return try HTMLElem.toInterface(HTMLElem.Union, res.?);
     }
 
     pub fn get_nextElementSibling(self: *parser.Element) !?Union {
         const res = try parser.nodeNextElementSibling(parser.elementToNode(self));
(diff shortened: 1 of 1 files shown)


src/dom/element.zig
```diff
@@ -177,6 +177,24 @@ pub const Element = struct {
         return try children.get_length();
     }
 
+    // NonDocumentTypeChildNode
+    // https://dom.spec.whatwg.org/#interface-nondocumenttypechildnode
+    pub fn get_previousElementSibling(self: *parser.Element) !?Union {
+        const res = try parser.nodePreviousElementSibling(parser.elementToNode(self));
+        if (res == null) {
+            return null;
+        }
```
francisbouvier: For simple case like that I think we should keep it in 1 line
`if (res == null) return null;`

src/dom/element.zig
```diff
@@ -177,6 +177,24 @@ pub const Element = struct {
...
+        return try HTMLElem.toInterface(HTMLElem.Union, res.?);
+    }
+
+    pub fn get_nextElementSibling(self: *parser.Element) !?Union {
+        const res = try parser.nodeNextElementSibling(parser.elementToNode(self));
+        if (res == null) {
+            return null;
+        }
```
francisbouvier: Same here

PR #134, opened 2023-12-14
basic queryselector

krichprollsch:
relates with https://github.com/lightpanda-io/browsercore/issues/20 and https://github.com/lightpanda-io/browsercore/issues/18

* [x] NodeList
* [x] element.childNodes
* [x] partial element.querySelector
* [x] partial element.querySelectorAll
* [x] partial document.querySelector
* [x] partial document.querySelectorAll

### DIFF
--- a/src/dom/document.zig
+++ b/src/dom/document.zig
@@ -194,7 +194,7 @@ pub const Document = struct {
 
     // TODO netsurf doesn't handle query selectors. We have to implement a
     // solution by ourselves.
-    // We handle only * and single id selector like `#foo`.
+    // For now we handle only * and single id selector like `#foo`.
     pub fn _querySelector(self: *parser.Document, selectors: []const u8) !?ElementUnion {
         if (selectors.len == 0) return null;
 


src/netsurf.zig
```diff
@@ -356,6 +356,10 @@ pub const NodeType = enum(u4) {
 // NodeList
 pub const NodeList = c.dom_nodelist;
 
+pub const NodeListType = enum(c_uint) {
```
francisbouvier: Do we need to use a `c_uint` here?

src/dom/document.zig
```diff
@@ -190,6 +192,56 @@ pub const Document = struct {
         return 1;
     }
 
+    // TODO netsurf doesn't handle query selectors. We have to implement a
+    // solution by ourselves.
+    // We handle only * and single id selector like `#foo`.
```
```suggestion
    // For now we handle only * and single id selector like `#foo`.
```

src/dom/document.zig
```diff
@@ -190,6 +192,56 @@ pub const Document = struct {
...
+            const walker = Walker{};
+            var next: ?*parser.Node = null;
+            while (true) {
+                next = try walker.get_next(root, next) orelse return list;
+                // ignore non-element nodes.
+                if (try parser.nodeType(next.?) != .element) {
+                    continue;
+                }
+                try list.append(alloc, next.?);
+            }
```
francisbouvier, line 232: `return list` here might be more readable

src/dom/nodelist.zig
```diff
@@ -0,0 +1,75 @@
...
+
+    const NodesArrayList = std.ArrayListUnmanaged(*parser.Node);
+
+    nodes: NodesArrayList,
+
+    pub fn init(alloc: std.mem.Allocator) !*NodeList {
+        const list = try alloc.create(NodeList);
+        list.* = NodeList{
+            .nodes = NodesArrayList{},
+        };
```
francisbouvier: Here it's not mandatory to use an allocator for just creating a reference in the heap.
JSRuntime will do it automatically when receiving a value, see https://github.com/lightpanda-io/jsruntime-lib/blob/main/src/engines/v8/generate.zig#L516

What is the best practice?
There is pros and cons.
Returning a value is easier. And requiring an allocator can imply other kind of allocations, while in fact we just use it to create the object itself.
But doing the allocation in the API in more explicit. And as we use the allocator in other methods of the API we already have to deal with a `deinit` function so it can make sens to go all the way on this road.

WDYT?

src/netsurf.zig
krichprollsch: we don't, I forgot to remove it, good catch,

src/dom/nodelist.zig
krichprollsch: I think you are right, the good practice is to return an instance and not a pointer with an `init`.
I will change it.

PR #137, opened 2023-12-15
start for loader and browser APi

krichprollsch:
Relates with https://github.com/lightpanda-io/browsercore/issues/48

```
$ zig build get -Dengine=v8 -- https://www.amazon.com/dp/0134190440/

$ zig build get -Dengine=v8 -- --dump https://www.amazon.com/dp/0134190440/
```

### DIFF
--- a/src/browser/browser.zig
+++ b/src/browser/browser.zig
@@ -14,7 +14,7 @@ const Env = jsruntime.Env;
 const apiweb = @import("../apiweb.zig");
 
 const Window = @import("../html/window.zig").Window;
-const Walker = @import("../dom/html_collection.zig").WalkerDepthFirst;
+const Walker = @import("../dom/walker.zig").WalkerDepthFirst;
 
 const FetchResult = std.http.Client.FetchResult;
 
--- a/src/browser/dump.zig
+++ b/src/browser/dump.zig
@@ -2,7 +2,7 @@ const std = @import("std");
 const File = std.fs.File;
 
 const parser = @import("../netsurf.zig");
-const Walker = @import("../dom/html_collection.zig").WalkerChildren;
+const Walker = @import("../dom/walker.zig").WalkerChildren;
 
 pub fn htmlFile(doc: *parser.Document, out: File) !void {
     try out.writeAll("<!DOCTYPE html>\n");
--- a/src/dom/document.zig
+++ b/src/dom/document.zig
@@ -10,7 +10,7 @@ const Node = @import("node.zig").Node;
 const NodeList = @import("nodelist.zig").NodeList;
 const NodeUnion = @import("node.zig").Union;
 
-const Walker = @import("html_collection.zig").WalkerDepthFirst;
+const Walker = @import("walker.zig").WalkerDepthFirst;
 const collection = @import("html_collection.zig");
 
 const Element = @import("element.zig").Element;
--- a/src/dom/element.zig
+++ b/src/dom/element.zig
@@ -9,7 +9,7 @@ const checkCases = jsruntime.test_utils.checkCases;
 const collection = @import("html_collection.zig");
 
 const Node = @import("node.zig").Node;
-const Walker = @import("html_collection.zig").WalkerDepthFirst;
+const Walker = @import("walker.zig").WalkerDepthFirst;
 const NodeList = @import("nodelist.zig").NodeList;
 const HTMLElem = @import("../html/elements.zig");
 pub const Union = @import("../html/elements.zig").Union;
--- a/src/dom/html_collection.zig
+++ b/src/dom/html_collection.zig
@@ -11,6 +11,11 @@ const utils = @import("utils.z");
 const Element = @import("element.zig").Element;
 const Union = @import("element.zig").Union;
 
+const Walker = @import("walker.zig").Walker;
+const WalkerDepthFirst = @import("walker.zig").WalkerDepthFirst;
+const WalkerChildren = @import("walker.zig").WalkerChildren;
+const WalkerNone = @import("walker.zig").WalkerNone;
+
 const Matcher = union(enum) {
     matchByName: MatchByName,
     matchByTagName: MatchByTagName,
@@ -254,89 +259,6 @@ pub fn HTMLCollectionByAnchors(
         .include_root = include_root,
     };
 }
-
-const Walker = union(enum) {
-    walkerDepthFirst: WalkerDepthFirst,
-    walkerChildren: WalkerChildren,
-    walkerNone: WalkerNone,
-
-    pub fn get_next(self: Walker, root: *parser.Node, cur: ?*parser.Node) !?*parser.Node {
-        switch (self) {
-            inline else => |case| return case.get_next(root, cur),
-        }
-    }
-};
-
-// WalkerDepthFirst iterates over the DOM tree to return the next following
-// node or null at the end.
-//
-// This implementation is a zig version of Netsurf code.
-// http://source.netsurf-browser.org/libdom.git/tree/src/html/html_collection.c#n177
-//
-// The iteration is a depth first as required by the specification.
-// https://dom.spec.whatwg.org/#htmlcollection
-// https://dom.spec.whatwg.org/#concept-tree-order
-pub const WalkerDepthFirst = struct {
-    pub fn get_next(_: WalkerDepthFirst, root: *parser.Node, cur: ?*parser.Node) !?*parser.Node {
-        var n = cur orelse root;
-
-        // TODO deinit next
-        if (try parser.nodeFirstChild(n)) |next| {
-            return next;
-        }
-
-        // TODO deinit next
-        if (try parser.nodeNextSibling(n)) |next| {
-            return next;
-        }
-
-        // TODO deinit parent
-        // Back to the parent of cur.
-        // If cur has no parent, then the iteration is over.
-        var parent = try parser.nodeParentNode(n) orelse return null;
-
-        // TODO deinit lastchild
-        var lastchild = try parser.nodeLastChild(parent);
-        while (n != root and n == lastchild) {
-            n = parent;
-
-            // TODO deinit parent
-            // Back to the prev's parent.
-            // If prev has no parent, then the loop must stop.
-            parent = try parser.nodeParentNode(n) orelse break;
-
-            // TODO deinit lastchild
-            lastchild = try parser.nodeLastChild(parent);
-        }
-
-        if (n == root) {
-            return null;
-        }
-
-        return try parser.nodeNextSibling(n);
-    }
-};
-
-// WalkerChildren iterates over the root's children only.
-pub const WalkerChildren = struct {
-    pub fn get_next(_: WalkerChildren, root: *parser.Node, cur: ?*parser.Node) !?*parser.Node {
-        // On walk start, we return the first root's child.
-        if (cur == null) return try parser.nodeFirstChild(root);
-
-        // If cur is root, then return null.
-        // This is a special case, if the root is included in the walk, we
-        // don't want to go further to find children.
-        if (root == cur.?) return null;
-
-        return try parser.nodeNextSibling(cur.?);
-    }
-};
-
-pub const WalkerNone = struct {
-    pub fn get_next(_: WalkerNone, _: *parser.Node, _: ?*parser.Node) !?*parser.Node {
-        return null;
-    }
-};
 
 pub const HTMLCollectionIterator = struct {
     pub const mem_guarantied = true;
--- a/src/dom/walker.zig
+++ b/src/dom/walker.zig
@@ -1,0 +1,86 @@
+const std = @import("std");
+
+const parser = @import("../netsurf.zig");
+
+pub const Walker = union(enum) {
+    walkerDepthFirst: WalkerDepthFirst,
+    walkerChildren: WalkerChildren,
+    walkerNone: WalkerNone,
+
+    pub fn get_next(self: Walker, root: *parser.Node, cur: ?*parser.Node) !?*parser.Node {
+        switch (self) {
+            inline else => |case| return case.get_next(root, cur),
+        }
+    }
+};
+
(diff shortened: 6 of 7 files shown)


src/html/window.zig
```diff
@@ -0,0 +1,44 @@
...
+
+    pub fn get_document(self: *Window) *parser.Document {
+        return self.document;
+    }
+
+    pub fn get_name(self: *Window) []const u8 {
+        return self.target;
+    }
+
+    // TODO we need to re-implement EventTarget interface.
```
francisbouvier, line 46: Why Window is not a child of EventTarget?

src/html/window.zig
```diff
@@ -0,0 +1,44 @@
...
+
+    pub fn get_self(self: *Window) *Window {
+        return self;
+    }
+
+    pub fn get_parent(self: *Window) *Window {
+        return self;
+    }
+
+    pub fn get_document(self: *Window) *parser.Document {
```
francisbouvier: Calling this method on a window instance with document undefined will lead to a runtime crash.
Is there a way we can avoid that?

src/apiweb.zig
francisbouvier, line 1: I'm not against this name change, but I think if we do that we should have inside first a `DOM.Interfaces` tuple and then an `Interfaces` tuple containing `DOM.Interfaces`.

The rationale is that DOM is not the only Web API we are going to support to it make sense to have those separated.

src/browser/browser.zig
```diff
@@ -0,0 +1,403 @@
...
+const Mime = @import("mime.zig");
+
+const jsruntime = @import("jsruntime");
+const Loop = jsruntime.Loop;
+const Env = jsruntime.Env;
+
+const apiweb = @import("../apiweb.zig");
+
+const Window = @import("../html/window.zig").Window;
+const Walker = @import("../dom/html_collection.zig").WalkerDepthFirst;
```
francisbouvier: Should we move the Walker outside html_collection if we are using it in other contexts?

src/browser/browser.zig
```diff
@@ -0,0 +1,403 @@
...
+
+const log = std.log.scoped(.browser);
+
+// Browser is an instance of the browser.
+// You can create multiple browser instances.
+// It contains only one session but initVM() and deinitVM() must be called only
+// once per main.
+pub const Browser = struct {
+    allocator: std.mem.Allocator,
+    session: *Session = undefined,
```
francisbouvier: I think it's not a good practice to provide `undefined` as a default value.
I will have just let this without default value.

src/browser/browser.zig
```diff
@@ -0,0 +1,403 @@
...
+
+    var vm: jsruntime.VM = undefined;
+    pub fn initVM() void {
+        vm = jsruntime.VM.init();
+    }
+    pub fn deinitVM() void {
+        vm.deinit();
+    }
```
francisbouvier: I'm not sure about having initVM/deinitVM as static methods of Browser.
Maybe it will be better to have it outside Browser and require a vm argument in the Browser.init?

src/browser/browser.zig
```diff
@@ -0,0 +1,403 @@
...
+// You can create successively multiple pages for a session, but you must
+// deinit a page before running another one.
+pub const Session = struct {
+    arena: std.heap.ArenaAllocator,
+    uri: []const u8,
+
+    // TODO handle proxy
+    loader: Loader = undefined,
+    env: Env = undefined,
+    loop: Loop = undefined,
```
francisbouvier: Same here, it's better IMHO to remove the default values.

src/browser/browser.zig
```diff
@@ -0,0 +1,403 @@
...
+
+    var vm: jsruntime.VM = undefined;
+    pub fn initVM() void {
+        vm = jsruntime.VM.init();
+    }
+    pub fn deinitVM() void {
+        vm.deinit();
+    }
+
+    pub fn init(allocator: std.mem.Allocator) !Browser {
```
francisbouvier: I think it's a common practice to call it `alloc`.
I don't mind, just we should try to be consistent inside the project.

src/browser/browser.zig
```diff
@@ -0,0 +1,403 @@
...
+        try self.env.load(&self.jstypes);
+
+        return self;
+    }
+
+    fn deinit(self: *Session) void {
+        self.loader.deinit();
+        self.loop.deinit();
+        self.env.deinit();
+        self.arena.deinit();
```
francisbouvier, line 93: The deinit should also destroy the Session itself no?
As the init create it.

src/browser/browser.zig
```diff
@@ -0,0 +1,403 @@
...
+const FetchResult = std.http.Client.FetchResult;
+
+const log = std.log.scoped(.browser);
+
+// Browser is an instance of the browser.
+// You can create multiple browser instances.
+// It contains only one session but initVM() and deinitVM() must be called only
+// once per main.
+pub const Browser = struct {
+    allocator: std.mem.Allocator,
```
francisbouvier: I think we could remove the allocator field and just pass it as argument on the deinit method (we do not need it anywhere else AFAIK).

src/browser/browser.zig
```diff
@@ -0,0 +1,403 @@
...
+pub const Page = struct {
+    allocator: std.mem.Allocator,
+    loader: *Loader,
+    env: *Env,
+    window: *Window,
+    doc: ?*parser.Document = null,
+
+    // handle url
+    rawuri: ?[]const u8 = null,
+    uri: std.Uri = undefined,
```
francisbouvier, line 117: Same here, no default value.

src/browser/browser.zig
```diff
@@ -0,0 +1,403 @@
...
+            &self.window,
+        );
+    }
+};
+
+// Page navigates to an url.
+// You can navigates multiple urls with the same page, but you have to call
+// end() to stop the previous navigation before starting a new one.
+pub const Page = struct {
+    allocator: std.mem.Allocator,
```
francisbouvier: I believe each page should have it's own arena allocator to release all memory related to the document when page.end is called.

src/browser/browser.zig
```diff
@@ -0,0 +1,403 @@
...
+    loader: *Loader,
+    env: *Env,
+    window: *Window,
```
francisbouvier: As all this elements belongs initially to the Session, maybe it will be more efficient here to just  reference the parent *Session?

src/browser/browser.zig
```diff
@@ -0,0 +1,403 @@
...
+        return Page{
+            .allocator = allocator,
+            .loader = loader,
+            .env = env,
+            .window = window,
+        };
+    }
+
+    pub fn end(self: *Page) void {
+        self.env.stop();
```
francisbouvier: Here we should reset (!= deinit) the arena allocator of the Page to release all memory related to the current document.

src/browser/browser.zig
```diff
@@ -0,0 +1,403 @@
...
+        // https://html.spec.whatwg.org/#reporting-document-loading-status
+
+        // TODO inject the URL to the document including the fragment.
+        // TODO set the referrer to the document.
+
+        self.window.replaceDocument(doc);
+
+        // https://html.spec.whatwg.org/#read-html
+
+        // start JS env
```
francisbouvier, line 226: Maybe we can add a TODO here to do in the future concurrently:
- the parsing of the body
- the loading of the JS env

src/browser/browser.zig
```diff
@@ -0,0 +1,403 @@
...
+        // el, and return.
+    }
+
+    // > type
+    // > Attribute is not set (default), an empty string, or a JavaScript MIME
+    // > type indicates that the script is a "classic script", containing
+    // > JavaScript code.
+    // https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#attribute_is_not_set_default_an_empty_string_or_a_javascript_mime_type
+    fn isJS(stype: ?[]const u8) bool {
+        return stype == null or stype.?.len == 0 or std.mem.eql(u8, stype.?, "application/javascript") or !std.mem.eql(u8, stype.?, "module");
```
francisbouvier: Maybe it will be clearer to have each condition in a separate line?

src/browser/browser.zig
```diff
@@ -0,0 +1,403 @@
...
+        // TODO set document.readyState to complete
+    }
+
+    // evalScript evaluates the src in priority.
+    // if no src is present, we evaluate the text source.
+    // https://html.spec.whatwg.org/multipage/scripting.html#script-processing-model
+    fn evalScript(self: *Page, e: *parser.Element) !void {
+        // https://html.spec.whatwg.org/multipage/webappapis.html#fetch-a-classic-script
+        const opt_src = try parser.elementGetAttribute(e, "src");
+        if (opt_src) |src| {
```
francisbouvier, line 339: Maybe all this logic can be put in a separate method, ie. `fetchRessource`?

src/html/window.zig
krichprollsch: My rationale was: `Window` will be not be a native pointer to a libdom struct.
If `EventTarget` uses native function using `seif`, it will work w/ document or node, but not for window...

WDYT?

src/html/window.zig
krichprollsch: The only way I see is to set document as optional and initialize it as `null`.

src/apiweb.zig
krichprollsch: I'm not sure I understand your comment here.
I renamed the `dom.zig` b/c I first place `window` in a `nav` package.
Currently `apiweb.zig` contains a tuple of interfaces containing `DOM.Interfaces`.
To add another api, we will have to add its `MyAPI.Interfaces`.

see https://github.com/lightpanda-io/browsercore/blob/589830ba7c1b0f1d9a2fa9203c90f8c99c906ac9/src/apiweb.zig

src/browser/browser.zig
krichprollsch: That could be useful, indeed. But should I put it in a `src/dom/walker.zig` file? WDYT?

src/browser/browser.zig
krichprollsch: this vm is awkward to use: the browser need to have a vm started globally, but It doesn't use it at all...
if we pass the vm. then I will not use it....
But the good point is it would be clearer to the caller a VM must be started.
I'm changing this way.

src/browser/browser.zig
krichprollsch: good point, I destroyed it via the browser deinit. but it makes more sens to destroy it here indeed.

src/browser/browser.zig
krichprollsch: I wasn't sure, mainly b/c the js env is at the `session` level...
That's why I preferred to set the arena at the session level instead of the page level.

src/browser/browser.zig
krichprollsch: cf my comment on page's arena. I wasn't sure I wanted to release the memory here.

src/browser/browser.zig
krichprollsch: `undefined` is useful for values I can't or I don't want to setup directly in the struct instantiation.
That's not the case for the session, I agree, but it's the case for others.

src/html/window.zig
francisbouvier: It's a complex question.

In JS world Window needs to have EventTarget has prototype, ie. `window instanceof EventTarget` should be true.

But in the Native world Window does not exists as a libdom struct. So we need to overwrite the EventTarget methods to avoid them to execute.
I think EventTarget has only the methods of event listener handling (add/remove listener and dispatch event). But we are not going to add a new separate event system just for Window. We could just use the window document for that as a workaround.

krichprollsch: I agree, I set the prototype to event target.

src/html/window.zig
krichprollsch: done :heavy_check_mark:

src/browser/browser.zig
krichprollsch: partially fixed, I kept the `undefined` for the `env`.

src/browser/browser.zig
krichprollsch: uri is set only when navigate, it's a kind of a private attribute...

src/browser/browser.zig
krichprollsch: done

src/browser/browser.zig
krichprollsch: changed

src/browser/browser.zig
krichprollsch: done

src/browser/browser.zig
krichprollsch: done

src/html/window.zig
francisbouvier: As we talked today, the best way will be to encapsulate libdom EventTarget inside Window, so the event system will work on it. We have to test it.

src/browser/browser.zig
```diff
@@ -0,0 +1,434 @@
...
+
+        log.debug("starting fetch script {s}", .{src});
+
+        const u = std.Uri.parse(src) catch try std.Uri.parseWithoutScheme(src);
+        const ru = try std.Uri.resolve(self.uri, u, false, alloc);
+
+        var fetchres = try self.session.loader.fetch(alloc, ru);
+        defer fetchres.deinit();
+
+        log.info("fech script {any}: {d}", .{ ru, fetchres.status });
```
francisbouvier, line 401:
```suggestion
        log.info("fetch script {any}: {d}", .{ ru, fetchres.status });
```

src/browser/browser.zig
```diff
@@ -0,0 +1,434 @@
...
+        // el, and return.
+    }
+
+    const FetchError = error{
+        BadStatusCode,
+        NoBody,
+        JsErr,
+    };
+
+    // fetchScript senf a GET request to the src and execute the script
```
francisbouvier, line 388:
```suggestion
    // fetchScript send a GET request to the src and execute the script
```

Issue #19, opened 2023-09-28
HTMLDocument properties and methods


francisbouvier:
https://html.spec.whatwg.org/#the-document-object

PR #144, opened 2024-01-03
dom: DocumentHTML getters

krichprollsch:
relates with #19

### DIFF
--- a/src/dom/document.zig
+++ b/src/dom/document.zig
@@ -108,12 +108,7 @@ pub const Document = struct {
         alloc: std.mem.Allocator,
         tag_name: []const u8,
     ) !collection.HTMLCollection {
-        var elt: ?*parser.Node = null;
-        if (try parser.documentGetDocumentElement(self)) |root| {
-            elt = parser.elementToNode(root);
-        }
-
-        return try collection.HTMLCollectionByTagName(alloc, elt, tag_name, true);
+        return try collection.HTMLCollectionByTagName(alloc, parser.documentToNode(self), tag_name, true);
     }
 
     pub fn _getElementsByClassName(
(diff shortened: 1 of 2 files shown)


src/html/document.zig
```diff
@@ -18,9 +23,175 @@ pub const HTMLDocument = struct {
...
+            next = try walker.get_next(root, next) orelse return null;
+            if (std.ascii.eqlIgnoreCase("head", try parser.nodeName(next.?))) {
+                return @as(*parser.Head, @ptrCast(next.?));
+            }
+        }
+    }
+
+    // TODO: not implemented by libdom
+    pub fn get_cookie(_: *parser.DocumentHTML) ![]const u8 {
+        return error.NotImplemented;
```
francisbouvier, line 65: Which difference are you making between NotSupported and NotImplemented?

Is there a reason to use error vs parser.DOMError?

src/html/document.zig
```diff
@@ -18,9 +23,175 @@ pub const HTMLDocument = struct {
...
+        return v;
+    }
+
+    pub fn _getElementsByName(self: *parser.DocumentHTML, alloc: std.mem.Allocator, name: []const u8) !NodeList {
+        var list = try NodeList.init();
+        errdefer list.deinit(alloc);
+
+        if (name.len == 0) return list;
+
+        const root = try rootNode(self) orelse return list;
```
francisbouvier:
```suggestion
        const root = try self.rootNode() orelse return list;
```

I think it's shows better that `rootNode` is a method of HTMLDocument

src/html/document.zig
```diff
@@ -18,9 +23,175 @@ pub const HTMLDocument = struct {
...
+        while (i < ln) {
+            const n = try c.item(i) orelse break;
+            try list.append(alloc, n);
+            i += 1;
+        }
+
+        return list;
+    }
+
+    inline fn rootNode(self: *parser.DocumentHTML) !?*parser.Node {
```
francisbouvier: Just for my understanding, this is mandatory right ? We can't ptrCast directly the DocumentHTML to Node for that?

src/html/document.zig
krichprollsch: My bad, I wanted to use an `error.NotImplemented` everywhere we have  a kind of `TODO`.

src/html/document.zig
krichprollsch: AFAIK I can't do that b/c `rootNode` is defined into `HTMLDocument` and `self` is a `*parser.DocumentHTML`

src/html/document.zig
krichprollsch: oh good point, indeed, I can just cast the document :facepalm: 
I'm updating

krichprollsch: I'm also applying the same change in `dom/document.zig` when possible.

PR #145, opened 2024-01-08
netsurf: rewrite wrapper.c in pure Zig

krichprollsch:


src/netsurf.zig
```diff
@@ -1280,42 +1281,100 @@ fn documentHTMLVtable(doc_html: *DocumentHTML) c.dom_html_document_vtable {
...
+        .enable_script = false,
+        .ctx = null,
+        .daf = null,
+    };
+
+    err = c.dom_hubbub_parser_create(&params, &parser, &doc);
+    try parserErr(err);
+    defer c.dom_hubbub_parser_destroy(parser);
+
+    var buffer: [1024 * 4]u8 = undefined;
```
francisbouvier: In the original file the buffer size is 1024, is there a reason to move to 1024*4?

krichprollsch: No real reason, I used a sample of code from somewhere. I'm changing back to `1024`.

PR #147, opened 2024-01-09
dom: attr: fix crash w/o parent

krichprollsch:
libdom crashes when setting the value on an attr w/o parent.
The root cause is here: https://github.com/lightpanda-io/libdom/blob/master/src/core/attr.c#L560

I don't really know how to fix it, so I do prefer return a proper error in this case.
I could change netsurf itself, but I think it's better/easier to do this test in zig instead.

WDYT @francisbouvier ?


francisbouvier:
Yes I think you are right, for the moment let just return a proper error, we will see afterwards how we can fix that in libdom.

PR #157, opened 2024-01-17
Window: add EventTargetBase

krichprollsch:
EventTargetBase implements event target vtable for pure zig struct

Relates to #170
blocked by #36


src/html/window.zig
```diff
@@ -1,6 +1,9 @@
 const std = @import("std");
 
 const parser = @import("../netsurf.zig");
+const c = @cImport({
```
francisbouvier: Why do we need this? I do not see any call to `c.<something>` .
And if we need it, it should be imported in netsurf.zig, I think we should try to keep the APIs away from the C code

PR #164, opened 2024-01-19
Add missing html tags

krichprollsch:
em, strong, menu and menuitem


krichprollsch:
+ add a `else` with error

PR #176, opened 2024-02-08
UIEvent

francisbouvier:


src/netsurf.zig
```diff
@@ -443,6 +447,52 @@ pub fn eventPreventDefault(evt: *Event) !void {
...
+pub const UIEvent = c.dom_ui_event;
+
+pub fn uiEventCreate() !*UIEvent {
+    var evt: ?*UIEvent = undefined;
+    const err = c._dom_ui_event_create(&evt);
+    try DOMErr(err);
+    return evt.?;
+}
+
+pub const UIEventInit = struct {
```
krichprollsch, line 460: I'm not sure about declaring this struct here. It's doesn't depend on netsurf, it could be declared directly inside `dom/UIEvent`

PR #181, opened 2024-02-15
xhr: respect 50ms min delay between two progress events

krichprollsch:


francisbouvier:
I understand the rationale but I wonder why 50ms? Do we know what the other browsers are doing? Is there anything on that on the specs?

krichprollsch:
The specs defines the 50ms:

> If not roughly 50ms have passed since these steps were last invoked, then return. 
https://xhr.spec.whatwg.org/#the-send()-method

PR #182, opened 2024-02-19
dom: innerHTML

krichprollsch:


src/browser/dump.zig
```diff
@@ -4,13 +4,14 @@ const File = std.fs.File;
 const parser = @import("../netsurf.zig");
 const Walker = @import("../dom/walker.zig").WalkerChildren;
 
-pub fn htmlFile(doc: *parser.Document, out: File) !void {
+pub fn htmlFile(doc: *parser.Document, out: anytype) !void {
```
francisbouvier: Should it be better to rename it, eg. `htmlWriter` ?

src/browser/dump.zig
```diff
@@ -4,13 +4,14 @@ const File = std.fs.File;
...
     try out.writeAll("<!DOCTYPE html>\n");
     try nodeFile(parser.documentToNode(doc), out);
     try out.writeAll("\n");
 }
 
-fn nodeFile(root: *parser.Node, out: File) !void {
+// out must be a std.io.Writer
+pub fn nodeFile(root: *parser.Node, out: anytype) !void {
```
francisbouvier: And `nodeWriter`?

src/dom/element.zig
```diff
@@ -420,4 +453,15 @@ pub fn testExecFn(
...
     };
     try checkCases(js_env, &attrNode);
+
+    var innerHTML = [_]Case{
+        .{ .src = "document.getElementById('para').innerHTML", .ex = " And" },
+        .{ .src = "document.getElementById('para-empty').innerHTML.trim()", .ex = "<span id=\"para-empty-child\"></span>" },
+
+        .{ .src = "let h = document.getElementById('para-empty')", .ex = "undefined" },
+        .{ .src = "const prev = h.innerHTML", .ex = "undefined" },
+        .{ .src = "h.innerHTML = prev; true", .ex = "true" },
```
francisbouvier, line 468: I think this test is a bit too simple. I would be better to:

- really set something different than the initial value
- set something with HTML tags, otherwise it's like `textContent` and we are not really testing the HTML parsing part

src/dom/node.zig
```diff
@@ -294,27 +294,33 @@ pub const Node = struct {
...
-            }
-        }
-
         // add new children
         for (nodes.?.slice) |node| {
             _ = try parser.nodeAppendChild(self, node);
         }
     }
 
+    pub fn removeChildren(self: *parser.Node) !void {
```
francisbouvier, line 309: I don't think this function should be public

src/dom/node.zig
```diff
@@ -294,27 +294,33 @@ pub const Node = struct {
     // function must accept either node or string.
     // blocked by https://github.com/lightpanda-io/jsruntime-lib/issues/114
     pub fn replaceChildren(self: *parser.Node, nodes: ?Variadic(*parser.Node)) !void {
+        // remove existing children
+        try removeChildren(self);
```
francisbouvier, line 298: I think by moving the logic before the checks on `nodes` we are changing the behavior.
Before is nodes was null or empty we wouldn't have remove the existing nodes. Now we are.
What does the standard says about that?

src/netsurf.zig
```diff
@@ -1859,6 +1875,53 @@ pub fn documentHTMLParse(reader: anytype, enc: ?[:0]const u8) !*DocumentHTML {
     return @as(*DocumentHTML, @ptrCast(doc.?));
 }
 
+pub fn documentParseFragmentFromStr(self: *Document, str: []const u8) !*DocumentFragment {
+    var fbs = std.io.fixedBufferStream(str);
+    return try documentParseFragment(self, fbs.reader(), "UTF-8");
+}
+
+pub fn documentParseFragment(self: *Document, reader: anytype, enc: ?[:0]const u8) !*DocumentFragment {
```
francisbouvier, line 1856: Can we factorize this function with `documentHTMLParse`?

src/browser/dump.zig
krichprollsch: what about `writeHTML(doc: *parser.Document, writer: anytype) !void {` ?

src/browser/dump.zig
krichprollsch: I picked `writeNode` instead.

src/dom/node.zig
krichprollsch: `Node.removeChildren` is used by `element.set_innerHTML`

src/dom/node.zig
krichprollsch: I'm afraid the spec is not crystal clear for this case.
https://dom.spec.whatwg.org/#dom-parentnode-replacechildren

But FF and Chrome remove chlidren with `null` or no args, so I did replicate the behavior.
```
> document.getElementById('content').children.length;
3
> document.getElementById('content').replaceChildren(null);
undefined
> document.getElementById('content').children.length;
0 
```

```
> document.getElementById('content').children.length;
3
> document.getElementById('content').replaceChildren();
undefined
> document.getElementById('content').children.length;
0 
```

src/dom/element.zig
francisbouvier: What about testing if `h.firstChild` is a \<p> with `textContent` _hello world_?

krichprollsch: updated

PR #215, opened 2024-04-16
CDP basic

francisbouvier:


src/msg.zig
```diff
@@ -0,0 +1,171 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+const std = @import("std");
+
+// pub const MaxStdOutSize = 512; // ensure debug msg are not too long
```
krichprollsch: Should we remove?

src/msg.zig
```diff
@@ -0,0 +1,171 @@
...
+
+/// MsgBuffer return messages from a raw text read stream,
```
krichprollsch:
```suggestion
/// MsgBuffer returns messages from a raw text read stream,
```

src/msg.zig
```diff
@@ -0,0 +1,171 @@
...
+/// according to the following format `<msg_size>:<msg>`.
+/// It handles both:
+/// - combined messages in one read
+/// - single message in several read (multipart)
```
krichprollsch:
```suggestion
/// - single message in several reads (multipart)
```

src/msg.zig
```diff
@@ -0,0 +1,171 @@
...
+/// It is safe (and good practice) to reuse the same MsgBuffer
```
krichprollsch:
```suggestion
/// It's safe (and a good practice) to reuse the same MsgBuffer
```

src/msg.zig
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    fn reset(self: *MsgBuffer) void {
+        self.size = 0;
+        self.pos = 0;
+    }
+
+    // read input
+    // - `do_func` is a callback to execute on each message of the input
+    // - `data` is a arbitrary payload that will be passed to the callback along with
```
krichprollsch: I think we should rewrite the comment with something like: `data` is a user data forwarded to `do_func`.

src/msg.zig
```diff
@@ -0,0 +1,171 @@
...
+        var _input = input; // make input writable
+
+        while (true) {
+            var msg: []const u8 = undefined;
+
+            // msg size
+            var msg_size: usize = undefined;
+            if (self.isEmpty()) {
+                // parse msg size metadata
+                const size_pos = std.mem.indexOfScalar(u8, _input, ':').?;
```
krichprollsch: What if `_input` doesn't contain any `:`... It would be better to return an error correctly instead of crashing.

```suggestion
                const size_pos = std.mem.indexOfScalar(u8, _input, ':') orelse return error.ErrNoColon;
```

src/msg.zig
```diff
@@ -0,0 +1,171 @@
...
+                    self.size = msg_size;
+                }
+
+                // get the new position of the cursor
+                const new_pos = self.pos + _input.len;
+
+                // check if the current input can fit in MsgBuffer
+                if (new_pos > self.buf.len) {
+                    // max_size is the max between msg size and current new cursor position
+                    const max_size = @max(self.size, new_pos);
```
krichprollsch: does the max need here? can't we always use `self.size` to realloc?

src/msg.zig
```diff
@@ -0,0 +1,171 @@
...
+                    // resize the MsgBuffer to fit
+                    self.buf = try alloc.realloc(self.buf, max_size);
```
krichprollsch: I think we should set an hard limit to realloc size, a global or a MsgBuf param.
we don't want to alloc depending what an external input data gives to us.

src/msg.zig
```diff
@@ -0,0 +1,171 @@
...
+                @memcpy(self.buf[self.pos..new_pos], _input[0..]);
+
+                // set the new cursor position
+                self.pos = new_pos;
+
+                // if multipart is not finished, go fetch the next input
+                if (!self.isFinished()) return;
+
+                // otherwhise multipart is finished, use its buffer as input
+                _input = self.buf[0..self.pos];
```
krichprollsch, line 122: I'm not sure, but I think something could be wrong if you have a multipart read and a combined one.
Something like:
```
10:--------
--5:-----
```

In this case you lose the `5:----` message isn't it?

src/server.zig
```diff
@@ -0,0 +1,293 @@
...
+        if (!std.mem.eql(u8, res, "")) {
+            std.log.debug("res {s}", .{res});
+            return sendAsync(self, res);
+        }
+    }
+
+    // Inspector
+
+    pub fn sendInspector(self: *Cmd, msg: []const u8) void {
+        if (self.env().getInspector()) |inspector| {
```
krichprollsch, line 311: What about panic if no inspector exists? it's a coding error isn't it?

src/server.zig
```diff
@@ -0,0 +1,293 @@
...
+        // TODO: pointer instead?
+        return self.browser.currentSession().loop;
+    }
+
+    inline fn env(self: Cmd) public.Env {
+        return self.browser.currentSession().env;
+    }
+
+    fn do(self: *Cmd, cmd: []const u8) anyerror!void {
+        const res = try cdp.do(self.alloc(), cmd, self);
```
krichprollsch: Who is the owner of `res`? when the mem is released?

src/server.zig
```diff
@@ -0,0 +1,293 @@
...
+    // Inspector
+
+    pub fn sendInspector(self: *Cmd, msg: []const u8) void {
+        if (self.env().getInspector()) |inspector| {
+            inspector.send(self.env(), msg);
+        }
+    }
+
+    pub fn onInspectorResp(cmd_opaque: *anyopaque, _: u32, msg: []const u8) void {
+        std.log.debug("onResp biz fn called: {s}", .{msg});
```
krichprollsch: `biz`?

src/msg.zig
krichprollsch: worst case I think:
```
10:--------
--10:------
----
```

src/server.zig
```diff
@@ -0,0 +1,293 @@
...
+        const self = @as(*Cmd, @ptrCast(aligned));
+
+        const tpl = "{s},\"sessionId\":\"{s}\"}}";
+        const msg_open = msg[0 .. msg.len - 1]; // remove closing bracket
+        const s = std.fmt.allocPrint(
+            self.alloc(),
+            tpl,
+            .{ msg_open, cdp.ContextSessionID },
+        ) catch unreachable;
+        defer self.alloc().free(s);
```
krichprollsch: since `sendSync` also free the `msg`, don't we double free `s`?

src/server.zig
```diff
@@ -0,0 +1,293 @@
...
+            self.cmd.err = err;
+            return;
+        };
+
+        std.log.debug("send async {d} bytes", .{size});
+        self.deinit(completion);
+    }
+};
+
+pub fn sendLater(ctx: *Cmd, msg: []const u8, ns: u63) !void {
```
krichprollsch: This function is not used and should be remove I think.

src/server.zig
```diff
@@ -0,0 +1,293 @@
...
+        self.msg_buf.read(self.alloc(), input, self, Cmd.do) catch unreachable;
+
+        // continue receving incomming messages asynchronously
+        self.loop().io.recv(*Cmd, self, cbk, completion, self.socket, self.buf);
+    }
+
+    // shortcuts
+    inline fn alloc(self: *Cmd) std.mem.Allocator {
+        // TODO: should we return the allocator from the page instead?
+        return self.browser.currentSession().alloc;
```
krichprollsch: The session initialize the js env with an arena created from the session's allocator.
The page starts and stops the env and free its own arena and the libdom memory.

Since`Cmd` allows navigation through multiple pages, It makes more sense to tie it with the browser's allocator, env and loop.

My real question is: It makes more sense to inject a `*Session` into `Cmd` instead of an `*Browser` itself.
Then you will be able to remove all these `browser.currentSession().XX`

src/server.zig
```diff
@@ -0,0 +1,293 @@
...
+    }
+
+    inline fn loop(self: *Cmd) public.Loop {
+        // TODO: pointer instead?
```
krichprollsch: I would say a pointer is better since the owner is the session.

src/server.zig
krichprollsch: It seems `cdp.do` says: `// The caller is responsible for calling `free` on the returned slice.`
So `sendAsync` must free the memory :thinking:

krichprollsch (commented):
first partial review

src/cdp/cdp.zig
```diff
@@ -0,0 +1,316 @@
...
+    // JSON scanner
+    var scanner = std.json.Scanner.initCompleteInput(alloc, s);
+    defer scanner.deinit();
+
+    std.debug.assert(try scanner.next() == .object_begin);
+
+    // handle 2 possible orders:
+    // - id, method <...>
+    // - method, id <...>
+    var method_key = (try scanner.next()).string;
```
krichprollsch: should you check if the token is really  a string before?

src/cdp/cdp.zig
```diff
@@ -0,0 +1,316 @@
...
+    var method_token: std.json.Token = undefined;
+    var id: ?u16 = null;
+    // check swap order
+    if (std.mem.eql(u8, method_key, "id")) {
+        id = try getId(&scanner, method_key);
+        method_key = (try scanner.next()).string;
```
krichprollsch: same check to do here

src/cdp/cdp.zig
```diff
@@ -0,0 +1,316 @@
...
+        method_token = try scanner.next();
+    } else {
+        method_token = try scanner.next();
+    }
+    try checkKey(method_key, "method");
+
+    // retrieve method
+    const method_name = method_token.string;
+    std.log.debug("cmd: method {s}, id {any}", .{ method_name, id });
```
krichprollsch: but `id` is set only if it's the first token isn't it?

src/cdp/cdp.zig
```diff
@@ -0,0 +1,316 @@
...
+        id = try getId(&scanner, method_key);
+        method_key = (try scanner.next()).string;
+        method_token = try scanner.next();
+    } else {
+        method_token = try scanner.next();
+    }
+    try checkKey(method_key, "method");
+
+    // retrieve method
+    const method_name = method_token.string;
```
krichprollsch, line 99: same about the type check

src/cdp/cdp.zig
```diff
@@ -0,0 +1,316 @@
...
+    std.log.debug("cmd: method {s}, id {any}", .{ method_name, id });
+
+    // retrieve domain from method
+    var iter = std.mem.splitScalar(u8, method_name, '.');
+    const domain = std.meta.stringToEnum(Domains, iter.first()) orelse
+        return error.UnknonwDomain;
+
+    // select corresponding domain
+    return switch (domain) {
+        .Browser => browser(alloc, id, iter.next().?, &scanner, ctx),
```
krichprollsch: it would be great to check if `iter.next()` is non null before using it everywhere...

```zig
const subdomain = iter.next orelse return error.BadMethod
```

src/cdp/cdp.zig
```diff
@@ -0,0 +1,316 @@
...
+fn getSessionId(scanner: *std.json.Scanner, key: []const u8) !?[]const u8 {
+
+    // check key is "sessionId"
+    if (!std.mem.eql(u8, "sessionId", key)) return null;
+
+    // parse "sessionId"
+    return (try scanner.next()).string;
+}
+
+pub fn getMsg(
```
krichprollsch, line 282: The caller must free the the params isn't it?
It doesn't seem to be easy since params are eventually allocated :/

src/browser/browser.zig
```diff
@@ -51,10 +51,9 @@ const log = std.log.scoped(.browser);
 pub const Browser = struct {
     session: *Session,
 
-    pub fn init(alloc: std.mem.Allocator, vm: jsruntime.VM) !Browser {
+    pub fn init(alloc: std.mem.Allocator) !Browser {
         // We want to ensure the caller initialised a VM, but the browser
         // doesn't use it directly...
-        _ = vm;
```
krichprollsch, line 57: Why do you remove the parameter and this line? The comment explain why I'm requesting the parameter even if I don't use it directly.

src/browser/browser.zig
```diff
@@ -132,9 +136,25 @@ pub const Session = struct {
         self.alloc.destroy(self);
     }
 
+    pub fn setInspector(
```
krichprollsch: `initInspector` instead?

src/browser/browser.zig
```diff
@@ -132,9 +136,25 @@ pub const Session = struct {
...
+        self.inspector = try jsruntime.Inspector.init(self.alloc, self.env, ctx, onResp, onEvent);
+        self.env.setInspector(self.inspector.?);
+    }
+
     pub fn createPage(self: *Session) !Page {
         return Page.init(self.alloc, self);
     }
+
+    pub fn callInspector(self: *Session, msg: []const u8) void {
+        if (self.inspector) |inspector| {
```
krichprollsch, line 150: I would panic if the inspector is missing.

src/browser/browser.zig
```diff
@@ -219,7 +239,7 @@ pub const Page = struct {
     }
 
     // spec reference: https://html.spec.whatwg.org/#document-lifecycle
-    pub fn navigate(self: *Page, uri: []const u8) !void {
+    pub fn navigate(self: *Page, uri: []const u8, auxData: ?[]const u8) !void {
```
krichprollsch, line 250: what is `auxData`?

src/cdp/page.zig
```diff
@@ -0,0 +1,399 @@
...
+    // Send Runtime.executionContextsCleared event
+    // TODO: noop event, we have no env context at this point, is it necesarry?
+    try sendEvent(alloc, ctx, "Runtime.executionContextsCleared", void, {}, msg.sessionID);
+
+    // Launch navigate
+    var p = try ctx.browser.currentSession().createPage();
+    ctx.state.executionContextId += 1;
+    const auxData = try std.fmt.allocPrint(
+        alloc,
+        // TODO: we assume this is the default web page, is it right?
```
krichprollsch: Since we have only one page at the same time, I think so yes.

src/cdp/page.zig
```diff
@@ -0,0 +1,399 @@
...
+    const params = msg.params.?;
+
+    // noop executionContextCreated event
+    try Runtime.executionContextCreated(
+        alloc,
+        ctx,
+        0,
+        "",
+        params.worldName,
+        "7102379147004877974.3265385113993241162",
```
krichprollsch, line 232: :thinking:

src/server.zig
francisbouvier: We need to keep a reference to the browser to allow create/delete sessions for multiple CDP calls on 1 connection

src/cdp/cdp.zig
francisbouvier: Yes, the log will display null if id it's not set at this stage.
Anyway I will reduce the logs when everything is working well so this might be removed.

src/cdp/cdp.zig
francisbouvier: The allocator is needed to call `std.json.innerParse` on params (when params_T is provided).
In my understanding this function only allocates temporary for token and free them before returning.
This function uses things like `@memcopy` from the original scanner to avoid allocating the params itself.

src/server.zig
francisbouvier: Yes, I've have removed the free form `sendSync`, the caller owns the string and it should be it's responsibility to free it.

src/server.zig
francisbouvier: Yes the memory is released by the async cbk who calls self.deinit, releasing the message.

src/msg.zig
francisbouvier: No because `realloc` replace the old size, it's not adding it.
But there is a bug, I sould use `@max(self.buf.len + self.size, new_pos)`
I've changed the name of the variable to be more explanatory (new_size) and updated the comment:

                    // we want to realloc at least:
                    // - a size big enough to fit the entire input (ie. new_pos)
                    // - a size big enough (ie. current size + starting size)
                    // to avoid multiple reallocation

src/msg.zig
francisbouvier: Yes you are right, I've put 1mb (hard coded) for now

src/msg.zig
francisbouvier: I think it's covered by this test case https://github.com/lightpanda-io/browser/pull/215/files#diff-c51b3cdde42681eb1426f0bb8396b098c3fbe99a8d5abd348a749534bae1c25cR159
```
        // multipart & combined
        .{ .input = "9:multi", .nb = 0 },
        .{ .input = "part2:ok", .nb = 2 },
```

We have 2 messages at the end _multipart_ and _ok_.

francisbouvier: I've added this test case to cover your example https://github.com/lightpanda-io/browser/pull/215/commits/7750956c7b5d62192958b2905c3c8a5c2e15a357

src/browser/browser.zig
```diff
@@ -99,13 +103,12 @@ pub const Session = struct {
     window: Window,
     // TODO move the shed to the browser?
     storageShed: storage.Shed,
-    page: ?*Page = null,
+    _page: ?Page = null,
```
krichprollsch: Why adding a `_` prefix here?

src/main_get.zig
```diff
@@ -83,18 +83,18 @@ pub fn main() !void {
...
     defer browser.deinit();
 
-    var page = try browser.currentSession().createPage();
-    defer page.deinit();
+    try browser.session.createPage();
 
-    try page.navigate(url, null);
-    defer page.end();
+    try browser.session.page().navigate(url, null);
+    defer browser.session.page().end();
```
krichprollsch: This new API looks weird

src/browser/browser.zig
```diff
@@ -158,8 +158,17 @@ pub const Session = struct {
         }
     }
 
-    pub fn createPage(self: *Session) !Page {
-        return Page.init(self.alloc, self);
+    pub fn createPage(self: *Session) !void {
```
krichprollsch: What about returning the `*Page` pointer? instead of have many `browser.session.page().XXX`

src/server.zig
```diff
@@ -0,0 +1,468 @@
...
+
+        try sendSync(ctx, s);
+    }
+
+    pub fn onInspectorResp(ctx_opaque: *anyopaque, _: u32, msg: []const u8) void {
+        if (std.log.defaultLogEnabled(.debug)) {
+            // msg should be {"id":<id>,...
+            const id_end = std.mem.indexOfScalar(u8, msg, ',') orelse unreachable;
+            const id = msg[6..id_end];
+            std.log.scoped(.cdp).debug("Res (inspector) > id {s}", .{id});
```
krichprollsch, line 340: AFAIK you can avoid this `scoped()` with :
```zig
const log = std.log.scoped(.cdp);
// [...]
log..debug("Res (inspector) > id {s}", .{id});
```

see https://github.com/lightpanda-io/browser/blob/main/src/xhr/xhr.zig#L42

krichprollsch:
You didn't update the deps with zig-jsruntime isn't it?

PR #278, opened 2024-10-21
Cdp full async

francisbouvier:
This PR ensure that all read and write I/O are done asynchronously:
- on the `readCbk` the next read is sent to the kernel _before_ doing the actual command work (so both are concurrent)
- on the write part the `sendSync` command was removed in favor of the `sendAsync`


krichprollsch:
The end of a connection results always with a `error.FileDescriptorNotASocket` error

```console
$ ./zig-out/bin/browsercore
info(server): Listening on: 127.0.0.1:3245...
info(server): accepting new conn...
info(server): close cmd, closing conn...
info(server): accepting new conn...
error(server): Server error: error.FileDescriptorNotASocket
```

PR #288, opened 2024-10-29
cdp: browserContextId is optional in Target.createTarget

krichprollsch:
https://chromedevtools.github.io/devtools-protocol/tot/Target/#method-createTarget


francisbouvier:
I think we should use the `browserContextID` provided (or not) in the request on the `attachedToTarget` event generated in this function.

For now we use the hard-coded `browserContextID`.

krichprollsch:
@francisbouvier PR updated

PR #310, opened 2024-11-27
websocket: first implementation

francisbouvier:


build.zig
```diff
@@ -168,6 +168,11 @@ fn common(
         .root_source_file = b.path("vendor/tls.zig/src/main.zig"),
     });
     step.root_module.addImport("tls", tlsmod);
+
+    const wsmod = b.addModule("ws", .{
```
krichprollsch: What about using the same name here then the name use to expose the module in root_module?
Just to avoid confusion.

 ```suggestion
    const wsmod = b.addModule("websocket", .{
```

File: src/main.zig [hunk 2]
```diff
@@ -170,10 +80,11 @@ const CliMode = union(CliModeTag) {
         host: []const u8 = Host,
         port: u16 = Port,
         timeout: u8 = Timeout,
+        tcp: bool = false, // undocumented TCP mode
 
         // default options
         const Host = "127.0.0.1";
-        const Port = 3245;
+        const Port = 9222;
```
krichprollsch, line 87: You must also change the usage before.
```
\\  --port          Port of the CDP server (default "3245")
```

krichprollsch (commented):
I tested it, it goes fast, but I can send a second connection :/

```
error(server): cancel error: error.NotFound
error(server): Server error: error.NotFound
```

PR #312, opened 2024-11-28
Fetch polyfill

krichprollsch:


krichprollsch:
follow-up: https://github.com/lightpanda-io/browser/issues/319

src/polyfill/fetch.zig
```diff
@@ -0,0 +1,719 @@
...
+//
+// The original code source is available in MIT license.
+//
+// The script comes from the built version from npm.
+// You can get the package with the command:
+//
+// wget $(npm view whatwg-fetch dist.tarball)
+//
+// The source is the content of `package/dist/fetch.umd.js` file.
+pub const source =
```
francisbouvier: What about commit the js file directly and use the directive `@embedFile`?

francisbouvier: It will be easier to read and modify if needed.

tests/html/await-fetch.html
```diff
@@ -0,0 +1,10 @@
+<script>
```
francisbouvier: Where is this test is played?

src/polyfill/fetch.zig
krichprollsch: I didn't know `@embedFile`, ty :pray:
It's much better to keep a separated pure js file.

tests/html/await-fetch.html
krichprollsch: I used the file for manual tests. Removed now.

PR #325, opened 2024-12-08
cdp: runtime, replace `"awaitPromise":true` only if present

francisbouvier:


src/cdp/runtime.zig
```diff
@@ -119,13 +119,16 @@ fn sendInspector(
...
+        if (std.mem.indexOf(u8, msg.json, "\"awaitPromise\":true")) |_| {
+            const buf = try alloc.alloc(u8, msg.json.len + 1);
+            defer alloc.free(buf);
+            _ = std.mem.replace(u8, msg.json, "\"awaitPromise\":true", "\"awaitPromise\":false", buf);
+            ctx.sendInspector(buf);
+            return "";
+        }
     }
+
+    ctx.sendInspector(msg.json);
```
krichprollsch, line 131: Does this change really relate with the PR itself?

PR #332, opened 2024-12-13
add --verbose option

krichprollsch:
use `--verbose` to enable debug output.
In debug mode, verbose is forced to `true`.

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -40,6 +40,14 @@ pub const websocket_blocking = true;
 
 const log = std.log.scoped(.cli);
 
+pub const std_options = .{
+    // Set the log level to info
+    .log_level = .debug,
+
+    // Define logFn to override the std implementation
+    .logFn = logFn,
+};
+
 const usage =
     \\usage: {s} [options] [URL]
     \\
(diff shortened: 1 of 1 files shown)


src/main.zig
```diff
@@ -334,3 +347,18 @@ pub fn main() !void {
         },
     }
 }
+
+var verbose: bool = builtin.mode == .Debug; // In debug mode, force verbose.
```
francisbouvier, line 351: Not a big fan of this global variable but I guess it's not possible to do without, at least if we want to keep using the std lib log function.

src/main.zig
```diff
@@ -334,3 +347,18 @@ pub fn main() !void {
...
+fn logFn(
+    comptime level: std.log.Level,
+    comptime scope: @Type(.EnumLiteral),
+    comptime format: []const u8,
+    args: anytype,
+) void {
+    if (!verbose) {
+        // hide all messages with level greater of equal to debug level.
+        if (@intFromEnum(level) >= @intFromEnum(std.log.Level.debug)) return;
```
francisbouvier, line 360: As `debug` is the last level I think checking an equality should be more clear. But of course like that you can't handle addition lowers levels. So it's also good to me like.

francisbouvier (approved):

src/main.zig
krichprollsch: Well, I think have a global var in the `main.zig` is not perfect, but ok.

src/main.zig
krichprollsch: Good point

krichprollsch: But Since we filter here, I do prefer the `>=` to avoid more verbose messages if a lower level is added by Zig later.

PR #393, opened 2025-01-30
Use https:// instead of git@ for submodules

karlseguin:
Allows easily building the project, following the steps in readme, without a github account or having some special git configuration.

If this change is acceptable, I can submit a similar PR for https://github.com/lightpanda-io/zig-js-runtime/blob/main/.gitmodules


krichprollsch:
Thanks @karlseguin :pray: 
We used ssh for submodules when the repos were private. It doesn't makes sense now, you are right.
The similar PR for zig-jsruntime  is welcome :+1:

PR #395, opened 2025-01-31
Add HTML encoding to text node and HTML attribute values

karlseguin:
Unfortunately, I'm probably still missing cases.


krichprollsch:
@karlseguin Thanks for you PR,
Unfortunately I merged #394 and GH closed yours instead of changing the target to `main` :/
Can you re-send the PR to `main` please? I can't do it via the webUI...

PR #405, opened 2025-02-04
Remove TCP Mode / Server

karlseguin:
Websocket handler now interacts directly with cdp. Simplifies the flow by removing the websocket -> server hop.

I updated the submodule (hopefully correctly), to point to my own branch for lightpanda. I needed the `afterInit` working (alternatively, I also have [this PR on lightpanda-io/websocket.zig](https://github.com/lightpanda-io/websocket.zig/pull/2)


karlseguin:
Just learned about the demo project and the puppeteer tests there. Was looking at writing something like that..useful to have. Will look into these failures tomorrow.

krichprollsch:
> Just learned about the demo project and the puppeteer tests there. Was looking at writing something like that..useful to have. Will look into these failures tomorrow.

 I was writing a comment regarding that. Not sure why the tests are failing.
 
 I started also https://github.com/lightpanda-io/websocket.zig/pull/3 to `zig fmt` + add the check in the CI but the unit tests are failing now... IDK why GH didn't run these tests before.

francisbouvier:
This approach removes entirely the JS loop (https://github.com/lightpanda-io/browser/blob/main/src/server.zig#L472). We need this loop to execute all JS async stuff (callbacks, setTimeout, etc).

That's why the current implementation has 2 infinite loops on 2 different threads:
- one main thread with the Websocket loop (`ws.listen`)
- one secondary thread with one loop mixing JS events and TCP events

I think we can still remove the TCP mode but we need to keep the JS infinite loop somewhere and create a way to communicate with the WS events (eg. stop the JS loop when the ws client closes the browser and start a new one afterwards with the updated new browser and v8 context).

francisbouvier:
Another solution is to have only one infinite event loop with both JS events and WS event (ie. in non-blocking mode). Either:
- use the WS non-blocking event loop (which uses directly epoll/kqueue if I understrand well) on the JS part
- or use the JS event loop (tigerbeetle one) on the WS part

I think that's more work than having 2 threads with 2 infinite loop, that's why I did not go on this path.

PR #408, opened 2025-02-06
Make TCP server websocket-aware

karlseguin:
Adding HTTP & websocket awareness to the TCP server.

HTTP server handles `GET /json/version` and websocket upgrade requests.

Conceptually, websocket handling is the same code as before, but receiving
data will parse the websocket frames and writing data will wrap it in
a websocket frame.

The previous `Ctx` was split into a `Server` and a `Client`. This was
largely done to make it easy to write unit tests, since the `Client` is
a generic, all its dependencies (i.e. the server) can be mocked out. This
also makes it a bit nicer to know if there is or isn't a client (via the
server's client optional).

Added a MemoryPool for the Send object (I thought that was a nice touch!)

Removed MacOS hack on accept/conn completion usage.


krichprollsch:
@karlseguin The change LGTM. thank you!
Can you just fix the conflict created by 39b37867762c575bc0e22674db6ed543436b403a please?

karlseguin:
> @karlseguin The change LGTM. thank you! Can you just fix the conflict created by [39b3786](https://github.com/lightpanda-io/browser/commit/39b37867762c575bc0e22674db6ed543436b403a) please?

done

PR #409, opened 2025-02-07
Add a new unittest build step

karlseguin:
Preserves all existing behavior (i.e. make test and zig build test are not changed in any way).

The new 'unittest' only runs unit tests and is fast to build. It takes ~1.7 to build unittest, vs ~11.09 to build test. This is really the main goal, and hopefully any unit test which are (a) fast and (b) don't impact build times will be run here.

The test runner is based on:
https://gist.github.com/karlseguin/c6bea5b35e4e8d26af6f81c22cb5d76b

It allow filtering, i.e. `make unittest F="parse query dup"`.

'unittest' does memory leak detection when tests use std.testing.allocator.

Fixed a memory leak in url/query which was detected/reported with by the new 'unittest'.

In order to avoid having 3 src/test_xyx.zig files, I merged the existing test_runner.zig and run_tests.zig into a single main_tests.zig. (this change is superfluous, but I thought it was cleaner this way. Happy to revert this).


src/unit_tests.zig
krichprollsch, line 1: I think we must include the license for the file here.
I think you can include a license less restrictive than AGPL here (MIT, apache2...) if you want to @karlseguin 

If it's ok for you @francisbouvier

src/browser/loader.zig
```diff
@@ -79,13 +79,19 @@ pub const Loader = struct {
     }
 };
 
-test "basic url get" {
+test "loader: get" {
     const alloc = std.testing.allocator;
     var loader = Loader.init(alloc);
     defer loader.deinit();
 
-    var result = try loader.get(alloc, "https://en.wikipedia.org/wiki/Main_Page");
```
krichprollsch, line 87: I would like to keep a test using TLS also...
But I agree having something locally is really better.

Maybe we could imagine an optional slower test suite for these kind of test we could run in CI.
But It's not a blocker for now.

krichprollsch (approved):

PR #410, opened 2025-02-08
Improve memory and performance of url.Query

karlseguin:
1 - Use getOrPut to avoid making 2 map lookups where possible.

2 - Use an arena allocator for Values, which makes memory management simpler.

3 - Because of the arena, we no longer need to allocate key or values which don't need
    to be unescaped. The downside is that the input string has to outlive the
    query.Values (but I think this is currently always the case)

4 - Optimize unescape logic & allocations

5 - Improve test coverage

### DIFF
--- a/src/url/query.zig
+++ b/src/url/query.zig
@@ -161,6 +161,8 @@ pub fn parseQuery(alloc: std.mem.Allocator, s: []const u8) !Values {
     var values = Values.init(alloc);
     errdefer values.deinit();
 
+    const arena = values.arena.allocator();
+
     const ln = s.len;
     if (ln == 0) return values;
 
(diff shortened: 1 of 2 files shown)


src/url/query.zig
```diff
@@ -223,61 +188,244 @@ pub fn parseQuery(alloc: std.mem.Allocator, s: []const u8) !Values {
     return values;
 }
 
-test "parse empty query" {
-    var values = try parseQuery(std.testing.allocator, "");
-    defer values.deinit();
+// The return'd string may or may not be allocated. Callers should use arenas
```
krichprollsch, line 193: Shouldn't we change the allocator we pass to `unescape` in `parseQuery` to use the `values`'s arena then?

karlseguin: YES! The test runner doesn't detect memory leaks when using std.testing.allocator right now (https://github.com/lightpanda-io/browser/pull/409)...would have caught that.

PR #422, opened 2025-02-12
Refactor CDP

karlseguin:
This includes / is built on top of https://github.com/lightpanda-io/browser/pull/408

CDP is now a struct, with a cleaner separation between Server, Client and CDP. Server <-> Client <-> CDP -> Browser. The CDP instance is the state.

Optimized how cdp results and events are sent to the client - only a single allocation is now required.

Each CDP message is now processed with an arena, which is re-used for each message. Small ad-hoc message-specific allocations, like the `aux_data` should now be more efficient.

Tried to remove some panics and undefines, or at least reduce the time that a field is undefined. Related, tried to remove the need for stable pointers (i.e. `Browser.init` now returns a `Browser`, rather than needing to be given a `*Browser`).

Improved parsing of CDP messages from the client.

Removed the newly added `browser.currentPage()`, since the cdp already has directly access to the session. (`browser.currentPage()` would make it harder to implement the "TODO allow multiple sessions per browser.")

Failed to get any unit test coverage from CDP, but i think it's important to do as there's more and more logic being added to it. Part of the blocker is global types. I'd consider introducing a global mock Browser/Session:

```zig
const Browser = if (builtin.is_test) MockBrowser else @import("../browser/browser.zig").Browser;
```

But I haven't given up hope on being able to quickly start a lightweight jsruntime.Env in unit tests.

To get a sense of the changes, I think you can look at `dom.discardSearchResults` which shows:
1 - how to get typed params
2 - how to access / modifity the state
3 - how to write a results


francisbouvier:
Hi @karlseguin, first of all thanks for this PR, it's impressive.

It seems that there is a regression (leak) on memory usage. When I launch 1000 runs of the puppeteer demo on my MBA M2, looking Activity Monitor I have:
- 190MB with this PR (release mode)
- 14MB on main (release mode)

karlseguin:
> It seems that there is a regression (leak) on memory usage. When I launch 1000 runs of the puppeteer demo on my MBA M2, looking Activity Monitor I have:

In the page cdp, I was creating a copy of the page, thus copying the arena, which causes issues. I had removed `session.currentPage()`. I've re-added it, as it returns the `&self.page` which helps avoid this issue.

There's still a leak on Mac from the cancel callback not being called and [these](https://github.com/lightpanda-io/zig-js-runtime/blob/main/src/loop.zig#L206) allocations not being freed, but this predates the PR.

francisbouvier:
Great, that fixes the leak, thanks.

I'm reviewing the code itself but it's a pretty big PR so it takes some time :)

src/browser/browser.zig
```diff
@@ -57,30 +59,44 @@ pub const user_agent = "Lightpanda/1.0";
...
+    allocator: Allocator,
+    session_pool: SessionPool,
 
-    const uri = "about:blank";
+    const SessionPool = std.heap.MemoryPool(Session);
 
-    pub fn init(self: *Browser, alloc: std.mem.Allocator, loop: *Loop, vm: jsruntime.VM) !void {
-        // We want to ensure the caller initialised a VM, but the browser
-        // doesn't use it directly...
-        _ = vm;
```
krichprollsch, line 68: I think it's useful to request  a `vm` as parameter here to force the caller to create one before calling.

src/browser/browser.zig
```diff
@@ -57,30 +59,44 @@ pub const user_agent = "Lightpanda/1.0";
...
 pub const Browser = struct {
-    session: Session = undefined,
-    agent: []const u8 = user_agent,
+    loop: *Loop,
+    session: ?*Session,
+    allocator: Allocator,
+    session_pool: SessionPool,
 
-    const uri = "about:blank";
+    const SessionPool = std.heap.MemoryPool(Session);
```
krichprollsch, line 67: Is it OK to use the heap directly here instead of an allocator passed by the caller?

src/browser/browser.zig
karlseguin: Before, there was one global browser, assigned directly to the server. 

Now there's 1 browser per client. Once this is merged, I was planning on making CDP multi-browser (I think when `Target.createBrowserContext`, we should create a new browser - else if a client creates multiple contexts, they won't get the isolation that they should (i.e. cookies/storage)). 

Having this check means storing the VM instance in the Server, in the Client, and eventually in CDP. If this is important, how about a global boolean in zig-js-runtime which is set to true on vm.init and which Env.init checks every time?

src/browser/browser.zig
karlseguin: If you mean that it's in `std.heap`, it's the same as `std.heap.ArenaAllocator`. It's a specialized ArenaAllocator that, knowing the type, can keep a free-list available. There's a bunch of places I want to add it (i.e. the Loop creating all those Completions)

It's initialized in browser's `init` as:

```zig
    pub fn init(allocator: Allocator, loop: *Loop) Browser {
        return .{
             // ...
            .session_pool = SessionPool.init(allocator),
        };
    }
```

So it does use the passed-in allocator.

https://www.openmymind.net/Zig-MemoryPool-Allocator/

src/browser/browser.zig
krichprollsch: AFAIK `vm.Init` https://github.com/lightpanda-io/zig-js-runtime/blob/main/src/engines/v8/v8.zig#L73-L78 calls multiple v8's functions to initiate v8 before any isolate is created (https://github.com/lightpanda-io/zig-v8-fork/blob/fork/src/v8.zig#L78-L91, https://github.com/lightpanda-io/zig-v8-fork/blob/fork/src/v8.zig#L117-L122, https://github.com/lightpanda-io/zig-v8-fork/blob/fork/src/v8.zig#L124-L129)

> Now there's 1 browser per client.

Got it :+1: 

> If this is important, how about a global boolean in zig-js-runtime which is set to true on vm.init and which Env.init checks every time?

I guess it's me, I found weird to have to init a vm and never pass it somewhere :sweat_smile: 
But using a boolean is ok too. The problem w/o this check is the error if you don't init the vm isn't very explicit:

```
$ zig build run


#
# Fatal error in ../../../../v8/src/api/api.cc, line 388
# Check failed: i::GetProcessWideSandbox()->is_initialized().
#
#
#
#FailureMessage Object: 0x7ffe9094fbe0run
└─ run lightpanda failure
error: the following command terminated unexpectedly:
/home/pierre/wrk/browser/.zig-cache/o/b20fc1513f2dcc377631adee7a63ca0f/lightpanda 
Build Summary: 2/4 steps succeeded; 1 failed (disable with --summary none)
run transitive failure
└─ run lightpanda failure
/home/pierre/wrk/browser/.zig-cache/o/7a9935469ab9b5af96ef2ed0fef6c52a/build /usr/local/zig-0.13.0/zig /home/pierre/wrk/browser /home/pierre/wrk/browser/.zig-cache /home/pierre/.cache/zig --seed 0x3890d754 -Z932405236f7f5d84 run
```

krichprollsch:
@karlseguin I'm ready to merge the PR. Can you resolve the conflicts? I'm sorry, I merged others PRs which created them... thanks :pray:

krichprollsch:
ok, we had a small memory regression detected . I guess the test a bit too much restrictive.
I think I will increase the max memory allowed a bit.
I restarted the test, it succeed :)

PR #424, opened 2025-02-13
server: implement /json/list

krichprollsch:
Mimic Chrome's endpoints to enable Selenium CDP connection.

Selenium requires:

On `/json/version`, `Browser` with a Chrome version and `WebkitVersion`.
On `json/list`, at least one item containing `id`, `url` and `type`.

:warning: This PR blocks on WS connection.

### DIFF
--- a/src/server.zig
+++ b/src/server.zig
@@ -578,6 +578,16 @@ fn Client(comptime S: type) type {
                 return self.send(self.server.json_version_response, false);
             }
 
+            const json_list_response =
+                "HTTP/1.1 200 OK\r\n" ++
+                "Content-Length: 47\r\n" ++
+                "Content-Type: application/json; charset=UTF-8\r\n\r\n" ++
+                "[{\"id\":\"1\",\"type\":\"page\",\"url\":\"abount:blank\"}]";
+
+            if (std.mem.eql(u8, url, "/json/list")) {
+                return self.send(json_list_response, false);
+            }
+
             return error.NotFound;
         }
 
@@ -1123,12 +1133,12 @@ pub fn run(
 
 // Utils
 // --------
-
+//
 fn buildJSONVersionResponse(
     allocator: Allocator,
     address: net.Address,
 ) ![]const u8 {
-    const body_format = "{{\"webSocketDebuggerUrl\": \"ws://{}/\"}}";
+    const body_format = "{{\"Browser\": \"Chrome/132.0.6834.110\", \"WebKit-Version\": \"537.36 (@df453a35f099772fdb954e33551388add2ca3cde)\", \"webSocketDebuggerUrl\": \"ws://{}/\"}}";
     const body_len = std.fmt.count(body_format, .{address});
 
     const response_format =


krichprollsch:
Script example can be found here: https://github.com/lightpanda-io/demo/pull/22

karlseguin:
After establishing the connection, the first thing it does is issue this CDP command which we don't support:

```json
{
    "id": 1,
    "method": "Target.getTargets",
    "params": {
        "filter": [
            { "exclude": true, "type": "browser" },
            { "exclude": true, "type": "page" },
            { "exclude": false}
        ]
    }
}
```

src/server.zig
```diff
@@ -1123,12 +1133,12 @@ pub fn run(
...
 // Utils
 // --------
-
+//
 fn buildJSONVersionResponse(
     allocator: Allocator,
     address: net.Address,
 ) ![]const u8 {
-    const body_format = "{{\"webSocketDebuggerUrl\": \"ws://{}/\"}}";
+    const body_format = "{{\"Browser\": \"Chrome/132.0.6834.110\", \"WebKit-Version\": \"537.36 (@df453a35f099772fdb954e33551388add2ca3cde)\", \"webSocketDebuggerUrl\": \"ws://{}/\"}}";
```
karlseguin, line 1141: In the branch for the demo project, I followed the instructions and ran `pip install trio selenium`.  When I tried to run the selenium script against this PR, I got this error:

> Message: session not created: cannot connect to chrome at 127.0.0.1:9222 from session not created: This version of ChromeDriver only supports Chrome version 133 Current browser version is 132.0.6834.110.

I wonder if this version response should be optionally loaded from the command line argument / a config file?

krichprollsch: Good catch.
It looks like selenium is running a `chrome --version` command to select the right webdriver version.
So I guess it expects the same version is returned by the endpoint.

```
chromedriver not found in PATH                          
chrome detected at /usr/bin/google-chrome                                                                       
Running command: /usr/bin/google-chrome --version                                                               
Output: "Google Chrome 132.0.6834.110 "                                                                         
Detected browser: chrome 132.0.6834.110                                                                         
Required driver: chromedriver 132.0.6834.159                                                                    
chromedriver 132.0.6834.159 already in the cache                                                                
Driver path: /home/pierre/.cache/selenium/chromedriver/linux64/132.0.6834.159/chromedriver                                                                                                                                       
Browser path: /usr/bin/google-chrome                                                                            
Started executable: `/home/pierre/.cache/selenium/chromedriver/linux64/132.0.6834.159/chromedriver` in a child process with pid: 993207 using 0 to output -3
```

I would be better to configure selenium with the websocket url to connect to.
But Idk if it's possible.
[Here](https://github.com/SeleniumHQ/selenium/blob/3c87e6bebc2438971fd5c57a0644045ef7d713a1/py/selenium/webdriver/remote/webdriver.py#L1203-L1207) It looks like it can either use an `se:cdp` capability  or retrieve it from the `/json/version` endpoint. But I wasn't able to configure it properly for now.

karlseguin: Figured out that you can force which version it will download:

```
options = webdriver.ChromeOptions()
options.browser_version = '132';  # add this
# ...
```

It downloaded 132 instead of 133 with this, and it didn't complain about the version when I ran it.

PR #426, opened 2025-02-17
In release mode, switch from page_allocator to c_allocator

karlseguin:
I'm not sure if the `page_allocator` was chosen for a specific reason, but I've always heard that the GPA should be used in debug/test, and the c_allocator in release.

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -202,23 +202,13 @@ pub fn main() !void {
 
     // allocator
     // - in Debug mode we use the General Purpose Allocator to detect memory leaks
-    // - in Release mode we use the page allocator
-    var alloc: std.mem.Allocator = undefined;
-    var _gpa: ?std.heap.GeneralPurposeAllocator(.{}) = null;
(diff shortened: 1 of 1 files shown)


krichprollsch:
I suppose we used the page allocator b/c Zig's doc recommends it with arena.
https://ziglang.org/documentation/master/#toc-Choosing-an-Allocator

But indeed using c_allocator is a better option since we link with libc.

before I merge it, @karlseguin can you rebase the branch to run the end-to-end regression tests please?

krichprollsch (approved):

karlseguin:
> I suppose we used the page allocator b/c Zig's doc recommends it with arena. https://ziglang.org/documentation/master/#toc-Choosing-an-Allocator
> 
> But indeed using c_allocator is a better option since we link with libc.
> 
> before I merge it, @karlseguin can you rebase the branch to run the end-to-end regression tests please?

done

krichprollsch:
Thanks! :+1:

PR #427, opened 2025-02-17
[WIP] Start adding structure logging, output to logfmt

karlseguin:
Adds a structured logger with output to logfmt format (which is both human readable and supported by most log ingestors).

Some ideas borrowed from https://github.com/karlseguin/log.zig.

My goal is to give each connection a unique conn_id, and each CDP message a req_id. The logger will get passed into CDP + Browser/Session/Page and even ideally into zig-js-runtime. The goal is to generate logs like:

```
_time=1739795092929 _level=info _ctx=client_connected conn_id=339494
_time=1739795092392 _level=info _ctx=http_request conn_id=339494 path=/json/version status=200
_time=1739795093001 _level=info _ctx=http_request conn_id=339494 path=/
_time=1739795093450 _level=info _ctx=cdp_request conn_id=339494 req_id=1 method=CSS.enable
...
```

Every log has a `_time`, `_level` and `_ctx` field. The "context" is just meant to be a unique label - nothing worse than getting a log message and not knowing where it originated (although it isn't a substitute for metrics, group by on `_ctx` can also be used to detect issues).

If everyone's ok with this, I want to close this PR and re-open it after https://github.com/lightpanda-io/browser/pull/422 has been merged with more of the `std.log` switched over to use this.

### DIFF
--- a/src/log.zig
+++ b/src/log.zig
@@ -1,0 +1,312 @@
+const std = @import("std");
+
+const Allocator = std.mem.Allocator;
+
+var mutex: std.Thread.Mutex = .{};
+
+const LogLevel : Log.Level = blk: {
+    const root = @import("root");
+    break :blk if (@hasDecl(root, "LogLevel")) root.LogLevel else .info;
+};
+
+pub const Log = LogT(std.fs.File);
+
+// Generic so that we can test it against an ArrayList
+fn LogT(comptime Out: type) type {
+    return struct {
+        out: Out,
+        inject: ?[]const u8,
+        allocator: Allocator,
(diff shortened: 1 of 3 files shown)


krichprollsch:
LGTM, the only change I could suggest right now is to format the timestamp in a human friendly format :)

karlseguin:
> LGTM, the only change I could suggest right now is to format the timestamp in a human friendly format :)

Agree. The latest version, in `Debug`, does:

``` 
info   | this is the message | 0 | key=value key=value
error | this is anotehr message | 10 |
```

where the time (0, 10) is the # of milliseconds since the last log.

PR #436, opened 2025-02-19
ci: add unittest execution

krichprollsch:


karlseguin:
I'd like to fix this after [refactor CDP](https://github.com/lightpanda-io/browser/pull/422) is merged. It's a small fix, but it'll just cause a conflict on that PR. 

I believe this issue is long-standing and it's the rapid reconnect of the unit test that's bringing it to light.

Client connects, timeout is registered using (`&server.timeout_completion`), client disconnects. Another client connects _before_ the `timeoutCbk` is called, another timeout is registered with the same `&server.timeout_completion`, causing the issue. The fix is that each client needs its own `timeout_completion` and, on callback, we need to check if the callback completion == current client's completion. 

Interestingly, when I added the tests, I noticed that it was really slow (200ms). I traced it back to this issue, but I didn't think through the full implications of re-submitting the same completion. Linux's behavior of crashing makes more sense to me than Mac being slow.

PR #442, opened 2025-02-21
Add explicit commands to binary

karlseguin:
Changes are fairly superfluous. I wanted to remove the error.NoError, but then thought having more explicit command -> options might be good (especially if we add more command and/or options).

I also dupe the string arguments (i.e. `host`). While not duping them is safe on linux/darwin, it isn't clear to me from the documentation that this is guaranteed.

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -85,7 +85,7 @@ jobs:
       - name: run puppeteer
         run: |
           python3 -m http.server 1234 -d ./public & echo $! > PYTHON.pid
-          ./lightpanda & echo $! > LPD.pid
+          ./lightpanda serve & echo $! > LPD.pid
           RUNS=100 npm run bench-puppeteer-cdp > puppeteer.out || exit 1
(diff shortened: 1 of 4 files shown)


krichprollsch:
@karlseguin thanks for the PR :pray: 
I think we should update some files accordingly:
* README.md https://github.com/lightpanda-io/browser/blob/main/README.md?plain=1#L64 and https://github.com/lightpanda-io/browser/blob/main/README.md?plain=1#L76
* Dockerfile https://github.com/lightpanda-io/browser/blob/main/Dockerfile#L79

Since we have already some users, I would like to keep a backward compatbility (with a warning log displayed maybe?)
ie:
- if nothing is passed, ie `./lightpanda` consider a `serve`
- if no command is sent  but a `--dump` option is passed, ie `./ligthpanda --dump`, consider a `fetch`
- if no command is sent but another option is passed, ie `./lightpanda --host`, consider a `serve`

krichprollsch:
I think we can have this backward compat mode until we tag a next release.

PR #444, opened 2025-02-21
Add an id generator

karlseguin:
Create UUID v4.

Create prefixed ids. To support more of the CDP protocol, we need to remove the hard-coded IDs (session, browser context, frame, loader, ...) and be able to dynamically create them, i.e. creating a new BrowserContextId when Target.createBrowserContext is called.

```zig
var frame_id = Incremental(u16, "FRM"){};
frame_id.next() == "FRM-1"
frame_id.next() == "FRM-2"
```

Generation is allocation-free (the returned string is only valid until the next call to next()). This is not thread safe, each CDP instance will have its own generator (for each id it needs to generate).

The generated IDs are different than what Chrome uses, i.e. BROWSERSESSIONID597D9875C664CAC0. I looked at various drivers and none have any expectations beyond a string. Shorter IDs will be more efficient. Also, the ID can cheeply be converted to and from an integer, allowing for lookups via AutoHashMap(u16) instead of StringHashMap.

### DIFF
--- a/src/id.zig
+++ b/src/id.zig
@@ -1,0 +1,161 @@
+const std = @import("std");
+
+// Generates incrementing prefixed integers, i.e. CTX-1, CTX-2, CTX-3.
+// Wraps to 0 on overflow.
+// Many caveats for using this:
+// - Not thread-safe.
+// - Information leaking
+// - The slice returned by next() is only valid:
+//   - while incrementor is valid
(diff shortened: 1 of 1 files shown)


krichprollsch:
@karlseguin LGTM too, can you rebase the branch pls?

PR #446, opened 2025-02-25
Add Usage Telemetry

karlseguin:
Sends basic usage telemetry. Final URL will need to be changed (probably).

Added Telemetry section to readme (please review). And I the Privacy Policy (linked from the readme)  might need updating to reflect this collection?

Currently, most things need an `Allocator` and `Loop` to run. Now we also need a `Telemetry`, and it isn't hard to imagine a not-so-distant future where various parts also need access to a `Config`. Rather than adding more and more parameters, I've introduced an `App` which is meant to group global data into a single variable, so that future additions are less likely to require touching so much code. Besides this change, the Telemetry code is relatively standalone.

### DIFF
--- a/src/telemetry/lightpanda.zig
+++ b/src/telemetry/lightpanda.zig
@@ -9,7 +9,7 @@ const telemetry = @import("telemetry.zig");
 const RunMode = @import("../app.zig").RunMode;
 
 const log = std.log.scoped(.telemetry);
-const URL = "https://lightpanda.io/browser-stats";
+const URL = "https://telemetry.lightpanda.io";
 
 pub const LightPanda = struct {
     uri: std.Uri,
--- a/src/telemetry/lightpanda.zig
+++ b/src/telemetry/lightpanda.zig
@@ -73,54 +73,55 @@ pub const LightPanda = struct {
         );
     }
 
-    fn handleError(self: *LightPanda, ctx: *Client.Ctx, err: anyerror) anyerror!void {
-        ctx.deinit();
-        self.client_context_pool.destroy(ctx);
+    fn handleError(sending: *Sending, ctx: *Client.Ctx, err: anyerror) anyerror!void {
+        const lightpanda = sending.lightpanda;
 
-        var sending: *Sending = @ptrCast(@alignCast(ctx.userData));
+        ctx.deinit();
+        lightpanda.client_context_pool.destroy(ctx);
+
         sending.deinit();
-        self.sending_pool.destroy(sending);
+        lightpanda.sending_pool.destroy(sending);
         log.info("request failure: {}", .{err});
     }
 
     fn onRequestConnect(ctx: *Client.Ctx, res: anyerror!void) anyerror!void {
-        var sending: *Sending = @ptrCast(@alignCast(ctx.userData));
-        res catch |err| return sending.lightpanda.handleError(ctx, err);
+        const sending: *Sending = @ptrCast(@alignCast(ctx.userData));
+        res catch |err| return handleError(sending, ctx, err);
 
         ctx.req.transfer_encoding = .{ .content_length = sending.body.len };
         return ctx.req.async_send(ctx, onRequestSend) catch |err| {
-            return sending.lightpanda.handleError(ctx, err);
+            return handleError(sending, ctx, err);
         };
     }
 
     fn onRequestSend(ctx: *Client.Ctx, res: anyerror!void) anyerror!void {
-        var sending: *Sending = @ptrCast(@alignCast(ctx.userData));
-        res catch |err| return sending.lightpanda.handleError(ctx, err);
+        const sending: *Sending = @ptrCast(@alignCast(ctx.userData));
+        res catch |err| return handleError(sending, ctx, err);
 
         return ctx.req.async_writeAll(sending.body, ctx, onRequestWrite) catch |err| {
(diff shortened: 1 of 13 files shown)


karlseguin:
Another good reason to replace `std.http.Client` is just how impossible it is to test.

src/telemetry/telemetry.zig
```diff
@@ -0,0 +1,191 @@
...
+}
+
+pub const Event = union(enum) {
+    run: Run,
+    navigate: void,
+    flag: []const u8, // used for testing
+
+    const Run = struct {
+        version: []const u8,
+        mode: RunMode,
```
krichprollsch: I would like to add `arch` (`x86_64`, `aarch64`, `arm64`) and `os` (`linux`,`macos`)

src/telemetry/telemetry.zig
```diff
@@ -0,0 +1,191 @@
...
+    std.fs.cwd().writeFile(.{ .sub_path = ID_FILE, .data = &id }) catch |err| {
+        log.warn("failed to write to id file: {}", .{err});
+        return null;
+    };
+    return id;
+}
+
+pub const Event = union(enum) {
+    run: Run,
+    navigate: void,
```
krichprollsch: navigate should have a `driver` (`cdp`, `none` for fetch navigation)

src/telemetry/telemetry.zig
```diff
@@ -0,0 +1,191 @@
...
+        // a "execution" id is an id that represents this specific run
+        eid: [36]u8,
+        provider: P,
+
+        disabled: bool,
+
+        const Self = @This();
+
+        pub fn init(allocator: Allocator, loop: *Loop) Self {
+            const disabled = std.process.hasEnvVarConstant("LIGHTPANDA_DISABLE_TELEMETRY");
```
krichprollsch, line 33: I would like to add log info here to notify the telemetry is enabled or disabled.

krichprollsch: If we are in test or debug mode we shouldn't log anything.

src/telemetry/lightpanda.zig
```diff
@@ -0,0 +1,200 @@
...
+        try jws.objectField("event");
+        try jws.write(@tagName(self.event.*));
+        try jws.objectField("mode");
+        try jws.write(self.run_mode);
+        try jws.objectField("version");
+        try jws.write(DATA_VERSION);
+        try jws.objectField("os");
+        try jws.write(builtin.os.tag);
+        try jws.objectField("version");
+        try jws.write(build_info.git_commit);
```
krichprollsch: You have `version` twice. It looks like the second is the correct one.

src/telemetry/lightpanda.zig
```diff
@@ -0,0 +1,166 @@
...
+    pub fn send(self: *LightPanda, iid: ?[]const u8, run_mode: RunMode, event: Event) !void {
+        var arena = std.heap.ArenaAllocator.init(self.allocator);
+        errdefer arena.deinit();
+
+        const resp_header_buffer = try arena.allocator().alloc(u8, 4096);
+        const body = try std.json.stringifyAlloc(arena.allocator(), .{
+            .iid = iid,
+            .run_mode = run_mode,
+            .os = builtin.os.tag,
+            .aarch = builtin.cpu.arch,
```
krichprollsch:
```suggestion
            .arch = builtin.cpu.arch,
```

src/telemetry/lightpanda.zig
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    pub fn send(self: *LightPanda, iid: ?[]const u8, run_mode: RunMode, event: Event) !void {
+        var arena = std.heap.ArenaAllocator.init(self.allocator);
+        errdefer arena.deinit();
+
+        const resp_header_buffer = try arena.allocator().alloc(u8, 4096);
+        const body = try std.json.stringifyAlloc(arena.allocator(), .{
+            .iid = iid,
+            .run_mode = run_mode,
```
krichprollsch:
```suggestion
            .mode = run_mode,
```

src/telemetry/telemetry.zig
```diff
@@ -0,0 +1,183 @@
...
+const builtin = @import("builtin");
+
+const Allocator = std.mem.Allocator;
+
+const Loop = @import("jsruntime").Loop;
+const uuidv4 = @import("../id.zig").uuidv4;
+const RunMode = @import("../app.zig").RunMode;
+
+const log = std.log.scoped(.telemetry);
+const ID_FILE = "lightpanda.id";
```
krichprollsch: As discussed, we should create a directory containing this id into the user's home.
Something like `$XDG_DATA_HOME/lightpanda/iid` for linux with fallback on `$HOME/.ligthpanda/iid` maybe?

And in macos maybe `~/Library/Lightpanda/iid`?

I suppose we could extract the dir management logic on its own file.

krichprollsch:
@karlseguin after discussion, here is what we could add for the `navigate` event:
```json
{
  "iid": "ffffffff-ffff-ffff-ffff-ffffffffffff",
  "version": "7b0e725",
  "arch": "x86_64",
  "os": "linux",
  "mode": "serve",
  "driver": "cdp",
  "event": "navigate"

  "tls": true,
  "proxy": false, // is a proxy configured for the request, always false for now.
}
```

PR #448, opened 2025-02-26
Add Set-Cookie parsing

karlseguin:
This code isn't hooked into anything. It's only a Set-Cookie response header parser (which includes a DateTime parser).

Next step:
1. Save them in the browser
2. When sending a request, figure out which cookies should be sent
3. Include them in the request

I added a testing.zig. I needed a function to assert an integer was within a delta:

```zig
test "DateTime: now" {
    const dt = DateTime.now();
    try testing.expectDelta(std.time.microTimestamp(), dt.micros, 1000);
}
```

`std.testing` only has this for floats. 

It also expands `expectEqual` to work against strings. I think is nice on its own, but it's particularly useful when testing against an anytype.

```zig
fn expectCookieAttributes(attributes: anytype, actual: Cookie) !void {
   inline for (@typeInfo(@TypeOf(attributes)).Struct.fields) |field| {
      testing.expectEqual(@field(anytype, field.name), @field(actual, field.name))
   }
}
```

This doesn't work with `std.expectEqual` if any of the values are strings.

### DIFF
--- a/src/storage/cookie.zig
+++ b/src/storage/cookie.zig
@@ -57,7 +57,7 @@ pub const Jar = struct {
         request_time: i64,
         origin_uri: ?Uri,
         target_uri: Uri,
-        navitation: bool,
+        navigation: bool,
     ) !CookieList {
         const target_path = target_uri.path.percent_encoded;
         const target_host = (target_uri.host orelse return error.InvalidURI).percent_encoded;
@@ -93,7 +93,7 @@ pub const Jar = struct {
                 // and cookie.same_site == .lax
                 switch (cookie.same_site) {
                     .strict => continue,
-                    .lax => if (navitation == false) continue,
+                    .lax => if (navigation == false) continue,
                     .none => {},
                 }
             }
@@ -101,14 +101,17 @@ pub const Jar = struct {
             {
                 const domain = cookie.domain;
                 if (domain[0] == '.') {
-                    // when explicitly set, the domain
-                    //  1 - always starts with a .
-                    //  2 - always is a suffix match (or examlpe)
+                    // When a Set-Cookie header has a Domain attribute
+                    // Then we will _always_ prefix it with a dot, extending its
+                    // availability to all subdomains (yes, setting the Domain
+                    // attributes EXPANDS the domains which the cookie will be
+                    // sent to, to always include all subdomains).
                     if (std.mem.eql(u8, target_host, domain[1..]) == false and std.mem.endsWith(u8, target_host, domain) == false) {
                         continue;
                     }
                 } else if (std.mem.eql(u8, target_host, domain) == false) {
-                    // when Domain=XYX isn't specified, it's an exact match only
+                    // When the Domain attribute isn't specific, then the cookie
+                    // is only sent on an exact match.
                     continue;
                 }
             }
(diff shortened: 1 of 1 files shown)


src/storage/cookie.zig
```diff
@@ -0,0 +1,922 @@
...
+        }
+    }
+
+    pub fn forRequest(
+        self: *Jar,
+        allocator: Allocator,
+        request_time: i64,
+        origin_uri: ?Uri,
+        target_uri: Uri,
+        navitation: bool,
```
krichprollsch: shouldn't be `navigation`? :thinking:

src/storage/cookie.zig
```diff
@@ -0,0 +1,922 @@
...
+                    .none => {},
+                }
+            }
+
+            {
+                const domain = cookie.domain;
+                if (domain[0] == '.') {
+                    // when explicitly set, the domain
+                    //  1 - always starts with a .
+                    //  2 - always is a suffix match (or examlpe)
```
krichprollsch:
```suggestion
                    //  2 - always is a suffix match (or example)
```

krichprollsch: I had trouble to understand the comment (and the code block)
Can you specify it's the parser which alwways add a leading dot for the domain set by the cookie?
And w/o leading dot, it's a cookie w/o domain.

src/storage/cookie.zig
```diff
@@ -0,0 +1,922 @@
...
+
+        var arena = ArenaAllocator.init(allocator);
+        errdefer arena.deinit();
+        const aa = arena.allocator();
+        const owned_name = try aa.dupe(u8, cookie_name);
+        const owned_value = try aa.dupe(u8, cookie_value);
+        const owned_path = if (path) |p|
+            try aa.dupe(u8, p)
+        else
+            try defaultPath(aa, uri.path.percent_encoded);
```
krichprollsch, line 338: `defaultPath` don't always return an allocated value. is it a problem?

src/storage/cookie.zig
```diff
@@ -0,0 +1,922 @@
...
+                } else if (std.mem.eql(u8, target_host, domain) == false) {
+                    // when Domain=XYX isn't specified, it's an exact match only
+                    continue;
+                }
+            }
+
+            {
+                const path = cookie.path;
+                if (path[path.len - 1] == '/') {
+                    // If our cookie path is doc/
```
krichprollsch: You mean `/doc/`? The parser ensure the path starts with a `/` isn't it?

src/storage/cookie.zig
karlseguin: i cleared up the comment, but yes, _we_ add the leading dot when reading the cookie. Setting `domain=lightpanda.io` in the cookie response EXPANDS the domains that the cookie will be sent to to include all submdomains. Counter-intuitive, but _not_ include the domain attribute is more restrictive.

src/storage/cookie.zig
karlseguin: Nope, the beauty of arena.

I hate that sometimes you feel like you need to dupe a string literal just to be able to safely call `free` on it when it isn't a string literal. Arena solves this.

src/storage/cookie.zig
karlseguin: Cleaned up the comment

krichprollsch (approved):

Issue #379, opened 2025-01-24
the example used in the readme does not work


frankgreco:
The hello world example does not work. In fact, no website I've tried works. It's usually always panics. 
```
./lightpanda-aarch64-macos --host 127.0.0.1 --port 9222
info(websocket): starting blocking worker to listen on 127.0.0.1:9222
info(server): accepting new conn...
info(server): client connected
info(browser): GET https://wikipedia.com/ 200
info(browser): fetch [https://wikipedia.com/portal/wikipedia.org/assets/js/index-2...](https://wikipedia.com/portal/wikipedia.org/assets/js/index-24c3e2ca18.js): http.Status.ok
info(browser): eval script portal/wikipedia.org/assets/js/index-24c3e2ca18.js: ReferenceError: location is not defined
info(browser): fetch [https://wikipedia.com/portal/wikipedia.org/assets/js/gt-ie9-...](https://wikipedia.com/portal/wikipedia.org/assets/js/gt-ie9-ce3fe8e88d.js): http.Status.ok
error(events): event handler error: error.JSExecCallback
info(events): event handler error try catch: TypeError: Cannot read properties of undefined (reading 'length')
info(server): close cmd, closing conn...
info(server): accepting new conn...
thread 5274880 panic: attempt to use null value
zsh: abort ./lightpanda-aarch64-macos --host 127.0.0.1 --port 9222
```

krichprollsch:
Hello @frankgreco,
Thanks for the issue.
Indeed I can reproduce, something goes wrong on handling the client's disconnection.

But I get the list of the links correctly. Did you get them?

frankgreco:
If I try to dump the page contents, I get a different error.

```javascript
'use strict'

import puppeteer from 'puppeteer-core';

const browser = await puppeteer.connect({
  browserWSEndpoint: "ws://127.0.0.1:9222",
});

const context = await browser.createBrowserContext();
const page = await context.newPage();

await page.goto('https://wikipedia.com/');

console.log(await page.content());

await page.close();
await context.close();
```
```
❯ node index.mjs
file:///Users/frank/projects/superblocks/browser/node_modules/puppeteer-core/lib/esm/puppeteer/common/CallbackRegistry.js:77
            this._reject(callback, new TargetCloseError('Target closed'));
                                   ^

TargetCloseError: Protocol error (Runtime.callFunctionOn): Target closed
[...]

PR #453, opened 2025-03-03
Reset loop event after page stop.

krichprollsch:
Relates with https://github.com/lightpanda-io/zig-js-runtime/pull/281


krichprollsch:
Relates with #379

PR #470, opened 2025-03-14
browser: fix module URL resolution

### LINKED ISSUE #464: module: Invalid URL resolution
downloaded last nightly build on March 12 2025.
lightpanda is executed on a server with Ubuntu 22.04.5 LTS
-rwxr-xr-x   1 root  root 49748912 Mar 12 12:37 lightpanda*

lightpanda dumps core when the following command is executed:
```
./lightpanda fetch --dump https://kiz-shop.com/
```
output:
```
info(browser): GET https://kiz-shop.com/ 200
info(browser): fetch https://kiz-shop.com/assets/index-30s5UcCB.js: http.Status.ok
info(browser): fetch https://kiz-shop.com/vendor-3ELwdC_X.js: http.Status.not_found
error(v8): resolveModuleCallback: load fn: error.BadStatusCode


#
# Fatal error in , line 0
# Check failed: has_pending_exception().
#
#
#
#FailureMessage Object: 0x7fffacca0b70
==== C stack trace ===============================

    ./lightpanda() [0x17d0793]
    ./lightpanda() [0x17cf5db]
    ./lightpanda() [0x1813cd0]
    ./lightpanda() [0x1842ce6]
    ./lightpanda() [0x18251a9]
    ./lightpanda() [0x17cddce]
    ./lightpanda() [0x1524845]
    ./lightpanda() [0x1524fab]
    ./lightpanda() [0x1525982]
    ./lightpanda() [0x1526acb]
    ./lightpanda() [0x152bc21]
    /lib/x86_64-linux-gnu/libc.so.6(+0x29d90) [0x7f552020ad90]
    /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0x80) [0x7f552020ae40]
    ./lightpanda() [0x150f025]
Trace/breakpoint trap (core dumped)
```

krichprollsch:
fix #464

### DIFF
--- a/src/browser/browser.zig
+++ b/src/browser/browser.zig
@@ -184,20 +184,17 @@ pub const Session = struct {
         _ = referrer;
 
         const self: *Session = @ptrCast(@alignCast(ctx));
-
-        if (self.page == null) {
-            return error.NoPage;
-        }
+        const page = &(self.page orelse return error.NoPage);
 
         log.debug("fetch module: specifier: {s}", .{specifier});
         // fetchModule is called within the context of processing a page.
         // Use the page_arena for this, which has a more appropriate lifetime
         // and which has more retained memory between sessions and pages.
         const arena = self.browser.page_arena.allocator();
-        const body = try self.page.?.fetchData(
+        const body = try page.fetchData(
             arena,
             specifier,
-            if (self.page.?.current_script) |s| s.src else null,
+            if (page.current_script) |s| s.src else null,
         );
         return self.env.compileModule(body, specifier);
     }
@@ -605,7 +602,7 @@ pub const Page = struct {
     // It resolves src using the page's uri.
     // If a base path is given, src is resolved according to the base first.
     // the caller owns the returned string
-    fn fetchData(self: *Page, arena: Allocator, src: []const u8, base: ?[]const u8) ![]const u8 {
+    fn fetchData(self: *const Page, arena: Allocator, src: []const u8, base: ?[]const u8) ![]const u8 {
         log.debug("starting fetch {s}", .{src});
 
         var buffer: [1024]u8 = undefined;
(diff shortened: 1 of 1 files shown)


src/browser/browser.zig
```diff
@@ -194,7 +194,11 @@ pub const Session = struct {
         // Use the page_arena for this, which has a more appropriate lifetime
         // and which has more retained memory between sessions and pages.
         const arena = self.browser.page_arena.allocator();
-        const body = try self.page.?.fetchData(arena, specifier);
+        const body = try self.page.?.fetchData(
+            arena,
+            specifier,
+            if (self.page.?.current_script) |s| s.src else null,
```
karlseguin: I don't think think you need to access this here. `fetchPage` is being called on the `page`, so you can access `page.current_script` directly from in there.

src/browser/browser.zig
```diff
@@ -546,6 +554,9 @@ pub const Page = struct {
     // if no src is present, we evaluate the text source.
     // https://html.spec.whatwg.org/multipage/scripting.html#script-processing-model
     fn evalScript(self: *Page, s: Script) !void {
+        self.current_script = &s;
```
karlseguin: Something weird about taking the address of this value and storing it as a field. But it does work, especially since you null it before you exit. 

Would be nice if the module loader took arbitrary state per-load,  but that's a bigger change, and would likely require allocating something on the heap to pass it through.

I do think `evalScript` should take a `*const Script` though. It won't change much - you'll still be storing a stack address in the field - but there's no reason for it to take a copy.

src/browser/browser.zig
krichprollsch: Not sure I'm following you correctly.
We are in `fetchModule` which is a callback function called by zig-js-runtime (and v8 before that).
The way to get the current page is to use the `ctx anyopaque` param which is the `*Session`.

karlseguin: Ah, `fetchData` is called from somewhere else too. I missed that.

karlseguin (approved):

Issue #471, opened 2025-03-15
TypeError: Cannot destructure property 'targetId' of '(intermediate value)' as it is undefined.


xpader:
I just try a simple example like this:

```js
import puppeteer from "puppeteer-core";

const browser = await puppeteer.connect({
	browserWSEndpoint: "ws://127.0.0.1:9222",
});

// The rest of your script remains the same.
const page = await browser.newPage();

// Dump all the links from the page.
await page.goto('https://www.baidu.com/', {waitUntil: 'domcontentloaded'});

// const links = await page.evaluate(() => {
//   return Array.from(document.querySelectorAll('a')).map(row => {
//     return row.getAttribute('href');
//   });
// });

await page.screenshot({path: "./data/tmp/lightpanda-screenshot.png"});

await page.close();
await browser.disconnect();
```

Get error:

```
file:///Volumes/Data/git-repos/lightpanda-browser/node_modules/puppeteer-core/lib/esm/puppeteer/cdp/Browser.js:176
        const { targetId } = await this.#connection.send('Target.createTarget', {
                ^

TypeError: Cannot destructure property 'targetId' of '(intermediate value)' as it is undefined.
    at CdpBrowser._createPageInContext (file:///Volumes/Data/git-repos/lightpanda-browser/node_modules/puppeteer-core/lib/esm/puppeteer/cdp/Browser.js:176:17)
    at async CdpBrowserContext.newPage (file:///Volumes/Data/git-repos/lightpanda-browser/node_modules/puppeteer-core/lib/esm/puppeteer/cdp/BrowserContext.js:120:20)
    at async CdpBrowser.newPage (file:///Volumes/Data/git-repos/lightpanda-browser/node_modules/puppeteer-core/lib/esm/puppeteer/cdp/Browser.js:173:16)
    at async file:///Volumes/Data/git-repos/lightpanda-browser/test.js:8:14

Node.js v18.20.6
```

package.json:
```json
{
  "type": "module",
  "dependencies": {
    "puppeteer-core": "^24.4.0"
  }
}
```

Platform: Linux x86_64, download lightpanda binary from github.

krichprollsch:
Hello @xpader,
Thanks for the issue. We merge #476 which should fix your issue.
The nightly build will integrate it tomorrow if you want to test.
Feel free to comment or create another issue if the problem persists.

xpader:
@krichprollsch Still got error with same code:

```
[...]

PR #476, opened 2025-03-17
Implicitly create BrowserContext on createTarget if one doesn't exist

karlseguin:


karlseguin:
Hopefully this improves compatibility with existing scripts which don't explicitly create a BrowerContext.

Doesn't fully solve https://github.com/lightpanda-io/browser/issues/471, since that script is also trying to take a screenshot (not supported).

krichprollsch (approved):

PR #478, opened 2025-03-19
Share the HTTP client globally

karlseguin:


src/browser/browser.zig
```diff
@@ -75,15 +75,14 @@ pub const Browser = struct {
             .app = app,
             .session = null,
             .allocator = allocator,
-            .http_client = .{ .allocator = allocator },
+            .http_client = @ptrCast(&app.http_client),
```
krichprollsch, line 78: why do we need a `@ptrCast` here?

krichprollsch (approved):

src/browser/browser.zig
krichprollsch: ok `http/client` vs `async/client` :ok_hand:

karlseguin: It's dangerous. But the two structs are identical.

PR #482, opened 2025-03-20
Replace zig-async-io with a custom HTTP client

karlseguin:
https://github.com/lightpanda-io/browser/pull/466

The two main goals are:
1 - Have a single HTTP client that does both sync and async. This makes it easier
    to add things like proxy support, and might help with things like cookies.

2 - Maintaining our own HTTP client isn't trivial, especially given that 
    we're likely to encounter non-conforming servers. But, this removes the need
    to maintain zig-async-io, which is a fork of [a small part] of the stdlib and
    tls.zig. This itself is non-trivial. Also, if we do run into weird server
    behavior, we'd have to maintain that in std.http.Client unless we could get
    a PR accepted.

What's currently missing
1 - ~Redirect handling (critical to add this)~
2 - ~Chunked encoding response (critical to add this)~
3 - Connection pooling (could be added after)
4 - Compression (could be added after)

If you try "fetch" or "serve" a page that doesn't do a direct and that doesn't
use chunked encoding, it should work. Puppeteer demo works.

This is built on the zig-0.14 branch. I initially started working against main 
(Zig 0.13) but then I noticed that tls.zig has _very_ useful support for a 
generic async loop in its master. It's a pretty big game changer and makes the
integration with our event loop trivial (I'm currently pointing to my own fork, 
but I'm hopeful that my PR (https://github.com/ianic/tls.zig/pull/8)
will be accepted, and if not, it's a simple change to accommodate tls.zig as-is).

The async API is simpler, with as single callback. A simplified example:

```zig
fn send (self *XMLHttpRequest, loop: *Loop) !void {
    const request = try self.client.request(self.method, self.uri);
    errdefer request.deinit();

    for (self.headers.list.items) |hdr| {
        try request.addHeader(hdr.name, hdr.value, .{});
    }
    request.body = self.payload;
    try request.sendAsync(loop, self, .{});
}

pub fn onHttpResponse(self: *XMLHttpRequest, progress_: http.Error!http.Progress) !void {
      const progress = try progress_;

      if (progress.first) {
          // the first emitted chunk,
          // get the headers, get the status
      }

      if (progress.data) |data| {
          // we have part of the body,
          // can still get the headers/status here if needed
          self.body.append(data);
      }

      if (progress.done) {
        // we're done!
      }
}
```

Generally, the integration with browser an XMLHttpRequest was rushed. The focus
has been on `src/http/client.zig`.

Finally, I've added the tls.zig dependency as a Zig package (build.zig.zon) 
rather than a submodule. I assume submodules were chosen back when Zig didn't
have a package manager.


src/http/client.zig
```diff
@@ -0,0 +1,2180 @@
...
+                    return .done;
+                };
+
+                if (reader.header_done == false) {
+                    // need more data
+                    return .need_more;
+                }
+
+                // at this point, If `would_be_first == true`, then
+                // `would_be_first` should be thought of as `is_first` because
```
krichprollsch, line 622: is the end if the comment missing here?

src/http/client.zig
```diff
@@ -0,0 +1,2180 @@
...
+                            // still handshaking, nothing to do
+                            return;
+                        }
+                        switch (secure.state) {
+                            .handshake => return self.sendSecureHeader(secure),
+                            .header => {
+                                secure.state = .body;
+                                const handler = self.handler;
+                                const body = handler.request.body orelse {
+                                    // We've sent the haeder, and there's no body
```
krichprollsch:
```suggestion
                                    // We've sent the header, and there's no body
```

src/http/client.zig
```diff
@@ -0,0 +1,2180 @@
...
+
+                // at this point, If `would_be_first == true`, then
+                // `would_be_first` should be thought of as `is_first` because
+
+                if (reader.redirect()) |redirect| {
+                    // We don't redirect until we've drained the body (because,
+                    // if we ever add keepalive, we'll re-use the connection).
+                    // Calling `reader.redirect()` over and over again might not
+                    // be the most efficient (it's a very simple function though),
+                    // but for a redirect resposne, chances are we slurped up
```
krichprollsch:
```suggestion
                    // but for a redirect response, chances are we slurped up
```

src/http/client.zig
```diff
@@ -0,0 +1,2180 @@
...
+                // With chunked-encoding, it's possible that we we've only
+                // partially processed the data. So we need to keep processing
+                // any unprocessed data. It would be nice if we could just glue
+                // this all together, but that would require copying bytes around
+                data = result.unprocessed orelse return .need_more;
+            }
+        }
+
+        fn handleError(self: *Self, comptime msg: []const u8, err: anyerror) void {
+            log.warn(msg ++ ": {any} ({any} {any})", .{ err, self.request.method, self.request.uri });
```
krichprollsch: what about using `log.err` instead?

krichprollsch (approved):
This is awesome :clap: Thank you so much @karlseguin :pray:

PR #495, opened 2025-03-28
Add CDP Node Registry

karlseguin:
This expands on the existing CDP node work used in  DOM.search. It introduces a node registry to track all nodes returned to the client and give lookups to get a node from a Id or a *parser.node.

Eventually, the goal is to have the Registry emit the DOM.setChildNodes event whenever necessary, as well as support many of the missing DOM actions.

Added tests to existing search handlers. Reworked search a little bit to avoid some unnecessary allocations and to hook it into the registry.

The generated Node is currently incomplete. The parentId is missing, the children are missing. Also, we still need to associate the v8 ObjectId to the node.

Finally, I moved all action handlers into a nested "domain" folder.


src/cdp/Node.zig
```diff
@@ -0,0 +1,385 @@
...
+            ._node = n,
+            .id = id,
+            .parent_id = null, // TODO
+            .backend_node_id = id, // ??
+            .node_name = try parser.nodeName(n),
+            .local_name = try parser.nodeLocalName(n),
+            .node_value = try parser.nodeValue(n) orelse "",
+            .node_type = @intFromEnum(try parser.nodeType(n)),
+            .child_node_count = children_count,
+            .children = &.{}, // TODO
```
krichprollsch, line 154: When using `chromedp`, empty children list with `children_count > 0` creates a regression with the previous implementation.

Before we sent:
```
-> {"id":15,"sessionId":"SID-1","method":"DOM.getDocument","params":{}}
<- {"id":15,"result":{"root":{"nodeId":1,"backendNodeId":1,"nodeType":9,"nodeName":"#document","localName":"","nodeValue":"","childNodeCount":2,"children":[{"nodeId":2,"backendNodeId":2,"nodeType":10,"nodeName":"html","localName":"","nodeValue":"","xmlVersion":"","compatibilityMode":"NoQuirksMode","isScrollable":false},{"nodeId":3,"backendNodeId":3,"nodeType":1,"nodeName":"HTML","localName":"html","nodeValue":"","xmlVersion":"","compatibilityMode":"NoQuirksMode","isScrollable":false}],"xmlVersion":"","compatibilityMode":"NoQuirksMode","isScrollable":false}},"sessionId":"SID-1"}
```

src/cdp/Node.zig
```diff
@@ -0,0 +1,385 @@
...
+    NoQuirksMode,
+};
+
+pub fn jsonStringify(self: *const Node, writer: anytype) !void {
+    try writer.beginObject();
+    try writer.objectField("nodeId");
+    try writer.write(self.id);
+
+    try writer.objectField("parentId");
+    try writer.write(self.parent_id);
```
krichprollsch, line 53: I think if `parent_id` is null, you shouldn't send the key at all.

src/cdp/Node.zig
```diff
@@ -0,0 +1,385 @@
...
+    try writer.write(self.compatibility_mode);
+
+    try writer.objectField("isScrollable");
+    try writer.write(self.is_scrollable);
+    try writer.endObject();
+}
+
+// Whenever we send a node to the client, we register it here for future lookup.
+// We maintain a node -> id and id -> node lookup.
+pub const Registry = struct {
```
sjorsdonkers, line 95: How does this relate to `dom/nodelist.zig` ?

krichprollsch: `dom/nodelist.zig` is the web API implementation used by the JS: https://developer.mozilla.org/en-US/docs/Web/API/NodeList

karlseguin: Only indirectly.

NodeList is a web api (https://developer.mozilla.org/en-US/docs/Web/API/NodeList). It's a collection of Nodes, for example from Node.childNodes or document.querySelectorAll

The registry is a lookup for CDP Nodes. It's safe to say that a CDP node is a CDP-specific representation of the web Node. Any CDP-Node that we send to the client is tracked here, for future reference (and, because, we'll probably need to keep track of which nodes we have and haven't already sent, and send extra data for new nodes).

src/cdp/Node.zig
```diff
@@ -0,0 +1,385 @@
...
+// 2 - Dom.getSearchResults
+// 3 - Dom.discardSearchResults
+//
+// For a given browser context, we can have multiple active searches. I.e.
+// performSearch could be called multiple times without getSearchResults or
+// discardSearchResults being called. We keep these active searches in the
+// browser context's node_search_list, which is a SearchList. Since we don't
+// expect many active searches (mostly just 1), a list is fine to scan through.
+pub const Search = struct {
+    name: []const u8,
```
sjorsdonkers, line 194: My understanding is that we use a string as key instead of an int to prevent str_to_int operation on get/remove calls.
However, since that makes us have to potentially allocate more, have an extra indirection as well as needing to to string compare.
I think it would be faster and simpler if we store integers of the search id, instead of the name.

sjorsdonkers: Additionally we may consider that the str <-> int conversion is not our responsibility. Remove and get can just take ints. Allowing the caller to control whether or not a conversion is needed.

karlseguin: I thought about this, and I did hack it a little to avoid an allocation for the first 10 searches..plus it's using an arena.

The issue is that the API expects and passes a string. To avoid an allocation, we'd need to create a custom type for the `performSearch` response and implement a custom `jsonStringify` method to serialize the integer as a string:

https://github.com/lightpanda-io/browser/pull/495/files#diff-b5e0a1f00caf33ad8c0458121ae192b673d64f67b34c20ddb7066ae3e1154cf2R76

Seems like a bit much.

src/cdp/Node.zig
```diff
@@ -0,0 +1,385 @@
...
+local_name: []const u8,
+node_value: []const u8,
+child_node_count: u32,
+children: []const *Node,
+document_url: ?[]const u8,
+base_url: ?[]const u8,
+xml_version: []const u8,
+compatibility_mode: CompatibilityMode,
+is_scrollable: bool,
+_node: *parser.Node,
```
sjorsdonkers, line 41: Should we also store the JSValue?

For resolveNode/wrapObject we would need to go from `node_id -> Node -> JSValue -> v8.Value`
How do we do the Node -> Value step?

karlseguin: Probably...and we might need to add another lookup for ObjectId->Node. But I'd like to address that separately and build this up as needed.

sjorsdonkers: We think we can use: `getV8Object(env.nat_ctx, node_ptr)`

src/cdp/Node.zig
karlseguin: Agree. I've restored the previous child behavior and, adding a custom Node writer because different APIs have different output requirements (i.e. DOM.requestChildNodes can request a specific depth). For now, I only support direct descendants, but all the hooks are in place to add more modes.

chromedp doesnt' work very well for me on main or this branch, so I couldn't test to make sure this got it back at least how it was before.

PR #496, opened 2025-03-28
Enable caller to decide where App lives

sjorsdonkers:
Before App.init would allocate it self on the heap (allocator).
However there appears to be no reason for it not to live on the stack.
This way the caller get's to decide.

### DIFF
--- a/src/app.zig
+++ b/src/app.zig
@@ -28,10 +28,7 @@ pub const App = struct {
         run_mode: RunMode,
     };
 
-    pub fn init(allocator: Allocator, config: Config) !*App {
-        const app = try allocator.create(App);
-        errdefer allocator.destroy(app);
-
+    pub fn init(allocator: Allocator, config: Config) !App {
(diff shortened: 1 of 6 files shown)


src/app.zig
```diff
@@ -49,7 +46,7 @@ pub const App = struct {
                 .tls_verify_host = config.tls_verify_host,
             }),
         };
-        app.telemetry = Telemetry.init(app, config.run_mode);
+        app.telemetry = Telemetry.init(&app, config.run_mode);
```
karlseguin, line 49: This is taking a stack address. It _could_ be ok, depending on what Telemetry does with the app, but in this case, it isn't. 

The alternative to doing what I did is to have `init` take app `*App`, which achieves what you want: allowing the caller to decide where the app lives. Just a slightly less friendly API and, in many cases, just puts the heap allocation burden on the caller.

sjorsdonkers: Thanks!

PR #501, opened 2025-04-02
Add a dumb renderer to get coordinates

karlseguin:
This isn't complete.

Can't test because scripts run into other blockers before reaching this point (i.e. Dom.resolveNode). I do think it's pretty safe to merge this as-is though.

Looking for feedback on the `FlatRenderer` in `browser.zig` and Page.click handling (netsurf code is a bit tedious to us).

Besides the inability to manually test. I'm also looking for feedback about how to test this. These flows are starting to get more complicated and I'm not sure they're reasonable for unit tests.

### DIFF
--- a/src/cdp/domains/input.zig
+++ b/src/cdp/domains/input.zig
@@ -1,0 +1,80 @@
+// Copyright (C) 2023-2024  Lightpanda (Selecy SAS)
+//
+// Francis Bouvier <francis@lightpanda.io>
+// Pierre Tachoire <pierre@lightpanda.io>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+const std = @import("std");
+
+pub fn processMessage(cmd: anytype) !void {
+    const action = std.meta.stringToEnum(enum {
+        dispatchMouseEvent,
+    }, cmd.input.action) orelse return error.UnknownMethod;
+
+    switch (action) {
+        .dispatchMouseEvent => return dispatchMouseEvent(cmd),
+    }
+}
+
+// https://chromedevtools.github.io/devtools-protocol/tot/Input/#method-dispatchMouseEvent
+fn dispatchMouseEvent(cmd: anytype) !void {
+    const params = (try cmd.params(struct {
+        type: []const u8,
+        x: u32,
+        y: u32,
+    })) orelse return error.InvalidParams;
+
+    try cmd.sendResult(null, .{});
+
+    if (std.ascii.eqlIgnoreCase(params.type, "mousePressed") == false) {
+        return;
+    }
+
+    const bc = cmd.browser_context orelse return;
+    const page = bc.session.currentPage() orelse return;
+    const click_result = (try page.click(cmd.arena, params.x, params.y)) orelse return;
+
+    switch (click_result) {
+        .navigate => |uri| try clickNavigate(cmd, uri),
+    }
+    // result already sent
+}
+
+fn clickNavigate(cmd: anytype, uri: std.Uri) !void {
+    const bc = cmd.browser_context.?;
+
+    var url_buf: std.ArrayListUnmanaged(u8) = .{};
+    try uri.writeToStream(.{
+        .scheme = true,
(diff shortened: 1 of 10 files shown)


src/netsurf/netsurf.zig
```diff
@@ -801,6 +801,24 @@ pub fn eventTargetDispatchEvent(et: *EventTarget, event: *Event) !bool {
...
+    const event = try eventCreate();
+    defer eventDestroy(event);
+
+    try eventInit(event, opts.type, .{ .bubbles = opts.bubbles, .cancelable = opts.cancelable });
+
+    var res: bool = undefined;
+    const et: *EventTarget = @ptrCast(element);
+    const err = eventTargetVtable(et).dispatch_event.?(et, event, &res);
+    try DOMErr(err);
+    return res;
```
krichprollsch: you can use `eventTargetDispatchEvent` function here.

src/cdp/domains/page.zig
```diff
@@ -140,20 +152,10 @@ fn navigate(cmd: anytype) !void {
...
-        referrerPolicy: ?[]const u8 = null, // TODO: enum
-    })) orelse return error.InvalidParams;
-
     // change state
     bc.reset();
-    bc.url = params.url;
-
-    // TODO: hard coded ID
-    bc.loader_id = "AF8667A203C5392DBE9AC290044AA4C2";
+    bc.url = url;
```
krichprollsch: I'm afraid this one is not correct is case of redirection...

krichprollsch (approved):
Thanks for this PR, it looks fine to me :+1: 

`Page.click` looks correct to me. The only thing we could change would be to dispatch a `PointerEvent` instead of a simple `Event`. See https://html.spec.whatwg.org/multipage/webappapis.html#fire-a-synthetic-pointer-event

But it implies some changes:
* creating a whole prototype chain of structs to represent the hierarchy:
`Event <- UIEvent <- MouseEvent <- PointerEvent` (https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent) (example w/ xhr progress event: https://github.com/lightpanda-io/browser/blob/main/src/xhr/progress_event.zig
* add the progress event enum: https://github.com/lightpanda-io/browser/blob/main/src/netsurf/netsurf.zig#L520-L523
* set the internal event type on pointer event creation: https://github.com/lightpanda-io/browser/blob/main/src/xhr/progress_event.zig#L50
* add a case to cast the event correctly: https://github.com/lightpanda-io/browser/blob/main/src/events/event.zig#L65

`FlatRenderer` looks good to me.

Regarding tests I don't have strong opinion. We will add end-to-end test for click at least...

krichprollsch:
libdom has a `mouse_event` https://github.com/lightpanda-io/libdom/blob/master/include/dom/events/mouse_event.h

src/cdp/domains/input.zig
```diff
@@ -0,0 +1,80 @@
...
+
+    try cmd.sendResult(null, .{});
+
+    if (std.ascii.eqlIgnoreCase(params.type, "mousePressed") == false) {
+        return;
+    }
+
+    const bc = cmd.browser_context orelse return;
+    const page = bc.session.currentPage() orelse return;
+    const click_result = (try page.click(cmd.arena, params.x, params.y)) orelse return;
```
krichprollsch: according with https://chromedevtools.github.io/devtools-protocol/tot/Input/#method-dispatchMouseEvent we should dispatch the event corresponding to the type given in parameter.
So I think we should:
* call `page.click` if the event type is ` mousePressed` or `mouseReleased`
* add a `TODO` to dispatch only the corresponding event in case of `mouseMoved`, `mouseWheel` (and we don't want to click in this case)

karlseguin: I made it all more generic, page.click -> page.mouseEvent, taking in a struct when an event type enum)

karlseguin:
> libdom has a `mouse_event` https://github.com/lightpanda-io/libdom/blob/master/include/dom/events/mouse_event.h

I switched to the mouse_event, but one of the parameters is a `struct dom_abstract_view *view` which I couldn't find anything about. I pass `null` for now, any idea about this?

src/cdp/domains/page.zig
karlseguin: This was existing code, but I agree. I removed the `url` state from the browsercontext and instead added a `getURL() []const u8` method which gets the URL from the session's page.

I think there's more of this encapsulation / ownership we need to do, and I think that, in most cases, the Browser Session/Page should be the authoritative source.

krichprollsch:
> I switched to the mouse_event, but one of the parameters is a `struct dom_abstract_view *view` which I couldn't find anything about. I pass `null` for now, any idea about this?

it looks like it's the `window` object :thinking: 
https://developer.mozilla.org/fr/docs/Web/API/MouseEvent#exemple

It is inherited from the UIEvent: https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/view

krichprollsch:
@karlseguin can you rebase to fix the conflict please?

karlseguin:
> @karlseguin can you rebase to fix the conflict please?

@krichprollsch ready

PR #504, opened 2025-04-03
browser: update urls after redirection

krichprollsch:
```
$ zig build run -- fetch  http://www.urlfilterdb.com/
debug(cli): Fetch mode: url http://www.urlfilterdb.com/, dump false
debug(browser): start js env
debug(browser): setup global env
debug(polyfill): load polyfill-fetch: undefined
debug(browser): inspector context created
debug: Inspector contextCreated called
debug(browser): starting GET http://www.urlfilterdb.com/
info(http_client): redirecting to: GET https://www.urlfilterdb.com/
info(browser): GET https://www.urlfilterdb.com/ 200
debug(browser): header content-type: text/html; charset=utf-8
debug(browser): parse html with charset utf-8
debug(browser): inspector context created
debug: Inspector contextCreated called
debug(browser): wait: OK
```

### DIFF
--- a/src/browser/browser.zig
+++ b/src/browser/browser.zig
@@ -351,23 +351,7 @@ pub const Page = struct {
             return;
         }
 
-        // own the url
-        self.rawuri = try arena.dupe(u8, uri);
-        self.uri = std.Uri.parse(self.rawuri.?) catch try std.Uri.parseAfterScheme("", self.rawuri.?);
-
-        self.url = try URL.constructor(arena, self.rawuri.?, null);
-        self.location.url = &self.url.?;
-        try self.session.window.replaceLocation(&self.location);
-
-        // prepare origin value.
-        var buf: std.ArrayListUnmanaged(u8) = .{};
-        try self.uri.writeToStream(.{
-            .scheme = true,
-            .authority = true,
-        }, buf.writer(arena));
-        self.origin = buf.items;
-
-        // TODO handle fragment in url.
+        self.uri = std.Uri.parse(uri) catch try std.Uri.parseAfterScheme("", uri);
 
         self.session.app.telemetry.record(.{ .navigate = .{
             .proxy = false,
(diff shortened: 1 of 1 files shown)


src/browser/browser.zig
```diff
@@ -380,7 +364,37 @@ pub const Page = struct {
 
         var response = try request.sendSync(.{});
         const header = response.header;
-        try self.session.cookie_jar.populateFromResponse(self.uri, &header);
+        try self.session.cookie_jar.populateFromResponse(request.uri, &header);
+
+        // update uri after eventual redirection
+        var buf: std.ArrayListUnmanaged(u8) = .{};
+        defer buf.deinit(arena);
```
karlseguin: We can treat the allocator like an `ArenaAllocator` or like an `std.mem.Allocator`.

If we treat it like an `Arena`, we can remove unnecessary frees, which has a super-tiny performance cost.

If we treat it like an `Allocator`, we gain some flexibility with respect to changing the implementation.

It isn't a system-wide decisions, there are places where we'll have an arena and want to treat it like a generic allocator, and there are places where we'll definitely want to take advantage of the arena's behavior. The reason I like to call these ArenaAllocators `arena` is that it's clear to the writer & reader what the behavior is.

TL;DR - The `deinit` is almost certainly a no-op and could be removed. It's up to you.

src/browser/browser.zig
```diff
@@ -380,7 +364,37 @@ pub const Page = struct {
...
+
+        buf.clearRetainingCapacity();
```
karlseguin: Does nothing

src/browser/browser.zig
```diff
@@ -380,7 +364,37 @@ pub const Page = struct {
...
+        try request.uri.writeToStream(.{
+            .scheme = true,
+            .authentication = true,
+            .authority = true,
+            .path = true,
+            .query = true,
+            .fragment = true,
+        }, buf.writer(arena));
+        self.rawuri = try buf.toOwnedSlice(arena);
```
karlseguin: I see that you re-use the buffer later, which I guess is why you copy the contents. But, `toOwnedSlice` frees the underlying memory, so the underlying slice won't get re-used.  I think you can use `buf.items` to avoid the allocation + copy.

src/browser/browser.zig
```diff
@@ -380,7 +364,37 @@ pub const Page = struct {
...
+
+        self.uri = try std.Uri.parse(self.rawuri.?);
+
+        // TODO handle fragment in url.
+        self.url = try URL.constructor(arena, self.rawuri.?, null);
+        self.location.url = &self.url.?;
+        try self.session.window.replaceLocation(&self.location);
+
+        // prepare origin value.
+        buf.clearRetainingCapacity();
```
karlseguin: As-above, if you stick with the `toOwnedSlice(arena)`, then this code does nothing. `toOwnedSlice(arena)` calls `clearAndFree()`.

If, above, you decide to do instead:

```
self.rawuri = buf.items
```

Then here you'll want to reset the buffer:

```
buf = .{};
```

as the `clearRetainingCapacity()` will break `self.rawuri`

src/browser/browser.zig
```diff
@@ -380,7 +364,37 @@ pub const Page = struct {
...
+        try request.uri.writeToStream(.{
+            .scheme = true,
+            .authority = true,
+        }, buf.writer(arena));
+        self.origin = try buf.toOwnedSlice(arena);
```
karlseguin: Also the option to use `buf.items` here.

krichprollsch:
@karlseguin thanks! I fixed the buffer usage.

PR #506, opened 2025-04-06
replace zig-js-runtime

karlseguin:
Big PR description, but 2 important things to highlight:

1 -
The build process has changed a lot. You'll need to run `zig build get-v8`
followed by `zig build build-v8` first. It might be better to test this branch
in a separate / clean folder (it'll leave a large v8 folder in the root of the
project).

2 - 
If you run this and notice that memory usage is high, try running it with the new
`--gc_hints` command line argument. Note that without the `--gc_hints` memory usage
does level off at some point, and it's faster. With `--gc_hints`, the memory 
consumption should be lower than before and it might still be a bit faster.

# Overview
This PR aims to replace `zig-js-runtime` with the main goal of:

1. Removing the global objects (Types and UserContext). (IO was largely removed
by the http client rewrite).
2. Reworking problematic and unsafe proto implementation (i.e. 
https://github.com/lightpanda-io/zig-js-runtime/issues/205)

The intention wasn't to improve performance or memory usage, but, at least on
my tests, both have been improved.

The changes can be broken down into two parts, changes to Browser and reworking
of zig-js-runtime.

## Browser Changes
Most of the changes to browser were made to accommodate the new js runtime.
However, other changes were also made:

* All browser-related code was moved under the browser folder. For example, 
  `src/css/` -> `src/browser/css/`

* `zig-v8-fork` and `tigerbeetle-io` are now Zig packages. Both projects were 
changed slightly to be proper Zig libraries. They both have a `jsruntime` branch 
which this branch references.

* `netsurf.zig` and `mimalloc.zig` are no longer separate modules and can be found in: `src/browser/netsurf.zig` and `src/browser/mimalloc.zig`.

* `zig build unittest` has been removed. `zig build test` is now more idiomatic:
the root source file is src/main.zig and tests are automatically discovered. 
`zig build test` is, as before, slow and something to be looked into.

* All existing web API tests are passing, but the test runner has changed a little
so every single line of test has been superficially touched. Also, 
https://github.com/lightpanda-io/zig-js-runtime/issues/200 is now fixed, so
the XHR test no long has commented out tests cases.

* The rest of the changes exist specifically to accommodate the new jsruntime, 
which will be covered next, but briefly:
  1. `Browser` / `Session` / `Page`, the main integration points with the js runtime,
have changed slightly.
  2. `UserContext` has been renamed to `SessionState` and, while the same in spirit, 
is also quite different.
  3. `Loop` and `Allocator` are no longer "internal" or special types. These are now
accessed through the SessionState.
  4. Callback API changed a little
  5. `Variadic` no longer exists, they map directly to plain Zig slices
  6. `postAttach` has been replaced with `indexed_get` and `named_get` which
capture live changes.
  7. `mem_guarantied` is no longer needed and has been removed where declared.
  8. 2 new build targets: `get-v8` which gets the tools to build v8
as well as the v8 source, and `build-v8` which builds v8. All of this is placed
in a `v8` folder at the root of the project. Also, on mac, we always build a release
build (since the dev build segfaults).

## js runtime
The zig-js-runtime project has been replaced with two files
`src/runtime/loop.zig` and `src/runtime/js.zig`.

The loop is unchanged except that `cancel` on  Mac no longer allocates
(and leaks) memory (it's a complete no-op, versus the current partial no-op).

### Architecture
The js runtime architecture matches the Browser structure:

Browser -> js.Env -> v8.Isolate
Session -> js.Executor -> v8.Context
Page -> js.Scope -> v8.HandleScope

As you can see, the v8.Isolate has been elevated to the Browser and is re-used
across multiple sessions. To make this possible, most (maybe all??) leaks have
been fixed. However, v8.Context are garbage collected by v8 at its own discretion.
Thus, this new approach, while faster, is likely to use more memory. With the new
`--gc_hints` command line argument, we call `isolate.lowMemoryNotification()` when
each executor ends -  encouraging v8 to garbage collect the just-released
context. Note that this is still just a hint. I think it's safe to say that this
approach gives more control to the v8 GC, making memory usage less predictable.

### Code Generation
The v8 bindings are created in a single-pass, directly from Zig's type system.
There is no longer a `reflect` stage nor an internal type representation. The
main advantage is that you don't need to learn two type systems.

### Generics
The global types have been replaced by making most js.* types (i.e. js.Env)
generics. I believe this significantly simplifies our build process, thus the
ability to have the tests' root source file be `src/main.zig`.

The only downside that I've seen is that you will get an _awful_ type description
in an error message involving the js.Env(T) or any of its child types.

### UserContext / SessionState / State
UserContext has been reworked. From the js runtime's point of view, i.e.
`js.Env(T)`, this is `State` generic. From the rest of Browser's point of view,
it is now called `SessionState`.

The `SessionState` belongs to the `Session` and is referenced by the `Executor`.
Rather than calling `setUserContext`, the `session.state` object is
mutated directly.

Most importantly, the js runtime no longer has the concept of special/internal
types.  Thus, `*Loop` and `Allocator` have been added to the SessionState.
For example, the `XMLHttpRequest` constructor was:

```zig
pub fn constructor(alloc: std.mem.Allocator, userctx: UserContext) !XMLHttpRequest {
  ...
}
```

And is now:

```zig
pub fn constructor(session_state: *SessionState) !XMLHttpRequest {
  const arena = session_state.arena;
  ...
}
```

### Compatibility with other engines
Without knowing anything about other JS engines, I feel that some of the changes 
could make integrating a different JS engine harder and some of the changes 
could make it simpler. I think this new approaches exposes a few less types and
presents more of a self-contained, but far from perfect, interface.

Personally, I think it would be great if the abstraction could be built under this layer, so that instead of having a v8.Context, v8.Value and v8.Object, we have a js.Context, js.Value and js.Object, so that binding code can be re-used, but that might be a stretch.


krichprollsch:
FYI I get a non-blocking warning when running `build-v8`

```
$ zig build build-v8
WARNING at the command-line "--args":1:76: Build argument has no effect.
target_os="linux" target_cpu="x64" is_debug=true symbol_level=1 cxx_use_ld="lld"
                                                                           ^----
The variable "cxx_use_ld" was set as a build argument
but never appeared in a declare_args() block in any buildfile.

To view all possible args, run "gn args --list <out_dir>"

The build continued as if that argument was unspecified.

Done. Made 183 targets from 100 files in 99ms
ninja: Entering directory `v8/build/x86_64-linux/debug/ninja'
[882/1972] CXX obj/torque_generated_definitions/feedback-vector-tq.o
```

karlseguin:
I saw this in the original too.

[try gn_args.append("cxx_use_ld=\"lld\"");](https://github.com/lightpanda-io/zig-v8-fork/blob/fork/build.zig#L234)

krichprollsch:
I tried the puppeteer `cdp.js` script with a release build and I get this error randomly on puppeteer:
(but nothing at lightpanda) 
```
$ node ../demo/puppeteer/cdp.js
......................................................................file:///home/pierre/wrk/demo/node_modules/puppeteer-core/lib/esm/puppeteer/common/CallbackRegistry.js:69
            this._reject(callback, new TargetCloseError('Target closed'));
                                   ^

TargetCloseError: Protocol error (Runtime.callFunctionOn): Target closed
    at CallbackRegistry.clear (file:///home/pierre/wrk/demo/node_modules/puppeteer-core/lib/esm/puppeteer/common/CallbackRegistry.js:69:36)
    at CdpCDPSession._onClosed (file:///home/pierre/wrk/demo/node_modules/puppeteer-core/lib/esm/puppeteer/cdp/CDPSession.js:98:25)
    at #onClose (file:///home/pierre/wrk/demo/node_modules/puppeteer-core/lib/esm/puppeteer/cdp/Connection.js:163:21)
    at WebSocket.<anonymous> (file:///home/pierre/wrk/demo/node_modules/puppeteer-core/lib/esm/puppeteer/node/NodeWebSocketTransport.js:43:30)
    at callListener (/home/pierre/wrk/demo/node_modules/ws/lib/event-target.js:290:14)
    at WebSocket.onClose (/home/pierre/wrk/demo/node_modules/ws/lib/event-target.js:220:9)
    at WebSocket.emit (node:events:518:28)
    at WebSocket.emitClose (/home/pierre/wrk/demo/node_modules/ws/lib/websocket.js:265:10)
    at Socket.socketOnClose (/home/pierre/wrk/demo/node_modules/ws/lib/websocket.js:1291:15)
    at Socket.emit (node:events:518:28) {
  cause: ProtocolError
      at <instance_members_initializer> (file:///home/pierre/wrk/demo/node_modules/puppeteer-core/lib/esm/puppeteer/common/CallbackRegistry.js:89:14)
      at new Callback (file:///home/pierre/wrk/demo/node_modules/puppeteer-core/lib/esm/puppeteer/common/CallbackRegistry.js:93:16)
      at CallbackRegistry.create (file:///home/pierre/wrk/demo/node_modules/puppeteer-core/lib/esm/puppeteer/common/CallbackRegistry.js:19:26)
      at Connection._rawSend (file:///home/pierre/wrk/demo/node_modules/puppeteer-core/lib/esm/puppeteer/cdp/Connection.js:86:26)
      at CdpCDPSession.send (file:///home/pierre/wrk/demo/node_modules/puppeteer-core/lib/esm/puppeteer/cdp/CDPSession.js:63:33)
      at #evaluate (file:///home/pierre/wrk/demo/node_modules/puppeteer-core/lib/esm/puppeteer/cdp/ExecutionContext.js:353:50)
      at async ExecutionContext.evaluateHandle (file:///home/pierre/wrk/demo/node_modules/puppeteer-core/lib/esm/puppeteer/cdp/ExecutionContext.js:320:16)
      at async IsolatedWorld.evaluateHandle (file:///home/pierre/wrk/demo/node_modules/puppeteer-core/lib/esm/puppeteer/cdp/IsolatedWorld.js:85:16)
      at async CdpJSHandle.evaluateHandle (file:///home/pierre/wrk/demo/node_modules/puppeteer-core/lib/esm/puppeteer/api/JSHandle.js:147:20)
      at async WaitTask.rerun (file:///home/pierre/wrk/demo/node_modules/puppeteer-core/lib/esm/puppeteer/common/WaitTask.js:100:28)
}

Node.js v22.14.0
```

README.md
```diff
@@ -216,17 +216,19 @@ Note: when Mimalloc is built in dev mode, you can dump memory stats with the
...
+**v8**
 
-Our own Zig/Javascript runtime, which includes the v8 Javascript engine.
-
-This build task is very long and cpu consuming, as you will build v8 from sources.
+First, get the tools necessary for building V8, as well as the V8 source code:
 
 ```
-make install-zig-js-runtime
+zig build get-v8
```
krichprollsch: I would prefer have a `make get-v8` instead

README.md
```diff
@@ -216,17 +216,19 @@ Note: when Mimalloc is built in dev mode, you can dump memory stats with the
...
 ```
 
-For dev env, use `make install-zig-js-runtime-dev`.
+Next, build v8. This build task is very long and cpu consuming, as you will build v8 from sources. 
+
+```
+zig build build-v8
```
krichprollsch: I would prefer have a `make build-v8` and a `make build-v8-dev`

.gitignore
```diff
@@ -4,3 +4,4 @@ zig-out
 /vendor/netsurf/out
 /vendor/libiconv/
 lightpanda.id
+v8/
```
krichprollsch:
```suggestion
/v8/
```

build.zig
```diff
@@ -42,193 +37,190 @@ pub fn build(b: *std.Build) !void {
...
-    const dep_opts = .{ .target = target, .optimize = optimize };
-
-    const jsruntimemod = try jsruntime_pkgs.module(
-        b,
-        options,
-        step.root_module.optimize.?,
-        target,
+    const lib_path = try std.fmt.allocPrint(
+        mod.owner.allocator,
+        "v8/build/{s}-{s}/{s}/ninja/obj/zig/libc_v8.a",
```
krichprollsch, line 179: We have to upgrade the `Dockerfile` also https://github.com/lightpanda-io/browser/blob/main/Dockerfile#L65

src/runtime/js.zig
```diff
@@ -0,0 +1,2273 @@
...
+    // We'll use the PROTOTYPE_TABLE to see if the TaggedAnyType should be
+    // cast to a prototype.
+    const PROTOTYPE_TABLE = comptime blk: {
+        var table: [Types.len]u16 = undefined;
+        const TYPE_LOOKUP = TypeLookup{};
+        for (Types, 0..) |s, i| {
+            var prototype_index = i;
+            const Struct = @field(types, s.name);
+            if (@hasDecl(Struct, "prototype")) {
+                prototype_index = 1;
```
krichprollsch: Is this line useful?  I don't think so :thinking:

src/runtime/js.zig
```diff
@@ -0,0 +1,2273 @@
...
+
+            // Arena whose lifetime is for a single page load, aka a Scope. Where
+            // the call_arena lives for a single function call, the scope_arena
+            // lives for the lifetime of the entire page. The allocator will be
+            // owned by the Scope, but the arena itself is owned by the Executor
+            // so that we can re-use it from scope to scope.
+            scope_arena: ArenaAllocator,
+
+            // When we need to load a resource (i.e. an external script), we call
+            // this function to get the source. This is always a refernece to the
```
```suggestion
            // this function to get the source. This is always a reference to the
```

src/runtime/js.zig
```diff
@@ -0,0 +1,2273 @@
...
+            // Browser Session's fetchModuleSource, but we use a funciton pointer
```
```suggestion
            // Browser Session's fetchModuleSource, but we use a function pointer
```

src/runtime/js.zig
```diff
@@ -0,0 +1,2273 @@
...
+            // executor in the EmbedderData of the v8.Context. So, as long as
+            // we have a v8.Context, we can get the executor.
+            fn mapZigInstanceToJs(context: v8.Context, js_obj_or_template: anytype, value: anytype) !PersistentObject {
+                const executor: *Executor = @ptrFromInt(context.getEmbedderData(1).castTo(v8.BigInt).getUint64());
+                return executor._mapZigInstanceToJs(js_obj_or_template, value);
+            }
+
+            // To turn a Zig instance into a v8 object, we need to do a number of things.
+            // First, if it's a struct, we need to put it on the heap
+            // Second, if we've alrady returned this instance, we should return
```
```suggestion
            // Second, if we've already returned this instance, we should return
```

.github/workflows/zig-test.yml
```diff
@@ -102,11 +102,8 @@ jobs:
 
       - uses: ./.github/actions/install
 
-      - name: zig build unittest
-        run: zig build unittest -freference-trace --summary all
-
       - name: zig build test
-        run: zig build test -Dengine=v8 -- --json > bench.json
+        run: zig build test -- --json > bench.json
```
krichprollsch, line 106: The `--json` option generated a json result used to track the perf in https://perf.lightpanda.io/benchmark/browser.
It's not the case anymore.

```
$ zig build test -- --json 2> /dev/null
        [
  {
    "name": "browser",
    "bench": {
      "duration": 3180096049,
      "alloc_nb": 6,
      "realloc_nb": 278,
      "alloc_size": 24711226
    }
  },
  {
    "name": "libdom",
    "bench": {
      "duration": 3180096049,
      "alloc_nb": 0,
      "realloc_nb": 0,
      "alloc_size": 0
    }
  },
  {
    "name": "v8",
    "bench": {
      "duration": 3180096049,
      "alloc_nb": 0,
      "realloc_nb": 0,
      "alloc_size": 0
    }
  },
  {
    "name": "main",
    "bench": {
      "duration": 3180096049,
      "alloc_nb": 0,
      "realloc_nb": 0,
      "alloc_size": 0
    }
  }
]
```

krichprollsch:
We had a benchmark test to track zig-js-runtime perf (https://perf.lightpanda.io/benchmark/jsruntime)
But idk if we can run it anymore with the new code and if it makes any sense now.
https://github.com/lightpanda-io/zig-js-runtime/blob/main/src/main_bench.zig
https://github.com/lightpanda-io/zig-js-runtime/blob/main/src/bench.zig

But it coud be useful to track the performance someway to detect regressions.

```$ zig build bench -- --json
[
  {
    "name": "With Isolate",
    "bench": {
      "duration": 17579169,
      "alloc_nb": 6,
      "realloc_nb": 484,
      "alloc_size": 154804
    }
  },
  {
    "name": "Without Isolate",
    "bench": {
      "duration": 1996698,
      "alloc_nb": 6,
      "realloc_nb": 558,
      "alloc_size": 155091
    }
  }
]
```

krichprollsch:
This PR is huge, but well explained. The comments in `runtime/` are useful and the code is easy to read, I love it :heart_eyes:  well done and thanks @karlseguin   :clap:

src/runtime/js.zig
karlseguin: it isn't, removed

build.zig
karlseguin: The dockerfile won't build for me, something about compiling netsurf on ARM linux, so it's a bit hard to test, but hopefully I got it right.

.github/workflows/zig-test.yml
karlseguin: I'm struggling with this. For now, copied this exact output and added it to stdout when `--json` flag is passed, but it's just hard-coded values ,not real ones.

karlseguin: This is working now.

build.zig
krichprollsch: Oh it makes sense: the dockerfile uses a clone of last `main` version to build. not your current git local dir. So the change is not compatible w/ the current main until we merge...

README.md
```diff
@@ -216,17 +216,21 @@ Note: when Mimalloc is built in dev mode, you can dump memory stats with the
...
-This build task is very long and cpu consuming, as you will build v8 from sources.
+```
+make get-v8
+```
+
+Next, build v8. This build task is very long and cpu consuming, as you will build v8 from sources. 
 
 ```
-make install-zig-js-runtime
+make build-v8-dev
```
```suggestion
make build-v8
```

krichprollsch (approved):

PR #519, opened 2025-04-09
Add URL struct

karlseguin:
`Page` currently contains 5 URL-ish related fields:

```zig
    rawuri: ?[]const u8 = null,
    uri: std.Uri = undefined,
    origin: ?[]const u8 = null,

    url: ?URL = null,
    location: Location = .{},
```

This PR adds a new URL struct which contains both the raw string and the `std.Uri`, and encapsulates some of the `std.Uri` API. It also tries to clean up some of the ownership. For example, `page` no longer has a `location` field. Instead, the window owns the location. Similarly, `page` no longer has a `url` field, the location now owns its url.

I original made http/client.zig and storage/cookie.zig use the new URL struct, but I backed away from that change in this PR. However, I did change those two classes to use  `*const std.Uri` rather than a `std.Uri`, again to try to clean up ownership.

I think there's more cleanup we can do around URI - the `std.Uri` API isn't great - so starting to encapsulate access through this custom struct is the first step.

### DIFF
--- a/src/browser/browser.zig
+++ b/src/browser/browser.zig
@@ -38,9 +38,7 @@ const apiweb = @import("../apiweb.zig");
 const Window = @import("../html/window.zig").Window;
 const Walker = @import("../dom/walker.zig").WalkerDepthFirst;
 
-const URL = @import("../url/url.zig").URL;
-const Location = @import("../html/location.zig").Location;
-
+const URL = @import("../url.zig").URL;
 const storage = @import("../storage/storage.zig");
 
 const http = @import("../http/client.zig");
@@ -253,7 +251,7 @@ pub const Session = struct {
         self.env.stop();
         // TODO unload document: https://html.spec.whatwg.org/#unloading-documents
 
-        self.window.replaceLocation(null) catch |e| {
+        self.window.replaceLocation(.{ .url = null }) catch |e| {
             log.err("reset window location: {any}", .{e});
         };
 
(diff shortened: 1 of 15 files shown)


src/browser/browser.zig
```diff
@@ -458,20 +440,6 @@ pub const Page = struct {
...
-            return null;
-        }
-
-        const node = parser.elementToNode(element);
-        const tag = try parser.nodeName(node);
-        if (std.ascii.eqlIgnoreCase(tag, "a")) {
-            const href = (try parser.elementGetAttribute(element, "href")) orelse return null;
-            var buf = try allocator.alloc(u8, 1024);
-            return .{ .navigate = try std.Uri.resolve_inplace(self.uri, href, &buf) };
-        }
```
krichprollsch, line 472: This deletion is not relative to the PR isn't it? you are preparing the change in #516?

src/cdp/domains/input.zig
```diff
@@ -64,11 +64,7 @@ fn dispatchMouseEvent(cmd: anytype) !void {
             else => unreachable,
         },
     };
-    const click_result = (try page.mouseEvent(cmd.arena, mouse_event)) orelse return;
-
-    switch (click_result) {
-        .navigate => |uri| try clickNavigate(cmd, uri),
-    }
```
krichprollsch, line 71: same with this deletion

src/browser/browser.zig
karlseguin: Yes, this function returns a `std.Uri`, which gets converted to a string in `input.clickNavigate`, then back into a uri by page.navigate. Since this code isn't currently being used, and it'll get removed anyways, I didn't want to bother making it work with the new URL.

PR #520, opened 2025-04-10
Communicate page navigation state via notifications

karlseguin:
https://github.com/lightpanda-io/browser/pull/516 needs a way for the page/session to communicate with CDP (i.e. the client). This adds a very simple mechanism for the page/session to callback into CDP. 

The `Page.navigate` CDP message has been changed to use the new notification appraoch. Page.navigate just signals the navigation request, and it's only via notifications from the page/session that page lifecycle events are sent back to the client.

I know this isn't the holistic event system that we want. This isn't something that the HTTP client can use to emit events nor is it something that telemetry can register listeners with. My goal is to get clicks working (from mouse events and javascript), then add some type of integration testing. Most of our recent additions don't have great code coverage, and events are hard to unit test. So I think this is the right time to add something to help increase our coverage around these more complicated cases.

### DIFF
--- a/src/browser/browser.zig
+++ b/src/browser/browser.zig
@@ -40,6 +40,7 @@ const Walker = @import("../dom/walker.zig").WalkerDepthFirst;
 
 const URL = @import("../url.zig").URL;
 const storage = @import("../storage/storage.zig");
+const Notification = @import("../notification.zig").Notification;
 
 const http = @import("../http/client.zig");
 const UserContext = @import("../user_context.zig").UserContext;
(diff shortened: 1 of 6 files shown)


src/notification.zig
```diff
@@ -0,0 +1,11 @@
+const URL = @import("url.zig").URL;
+
+pub const Notification = union(enum) {
+    page_navigate: PageEvent,
+    page_navigated: PageEvent,
+
+    pub const PageEvent = struct {
+        ts: u32,
```
krichprollsch: what does `ts` stand for? timestamp?
EDIT: yes it is.

src/browser/browser.zig
```diff
@@ -137,14 +138,32 @@ pub const Session = struct {
 
     jstypes: [Types.len]usize = undefined,
 
+    // recipient of notification, passed as the first parameter to notify
+    ctx: *anyopaque,
```
krichprollsch: shouldn't we specify `ctx` is used for notifications in its name? sounds very generic. Or do you think it could be used for something else?

src/browser/browser.zig
```diff
@@ -835,7 +857,13 @@ const FlatRenderer = struct {
     }
 };
 
-const NoopInspector = struct {
+const NoopContext = struct {
     pub fn onInspectorResponse(_: *anyopaque, _: u32, _: []const u8) void {}
     pub fn onInspectorEvent(_: *anyopaque, _: []const u8) void {}
+    pub fn notify(_: *anyopaque, _: *const Notification) !void {}
```
krichprollsch, line 863: I guess we could eventually use notifications to send also inspector's events...

src/notification.zig
karlseguin: renamed to `timestamp`

src/browser/browser.zig
karlseguin: renamed to `notify_ctx`

Issue #518, opened 2025-04-09
cdp: Implement DOM.describeNode


krichprollsch:
We want to implement [`DOM.describeNode`](https://chromedevtools.github.io/devtools-protocol/tot/DOM/#method-describeNode) which is required for puppeteer click scenario.

## How to reproduce?

Use [`puppeteer/click.js`](https://github.com/lightpanda-io/demo/blob/main/puppeteer/click.js) scenario from the demo repository.

* start the demo webserver with `$ go run ws/main.go`
* optionnally use [cdpproxy](https://github.com/lightpanda-io/cdpproxy) to log all cdp messages
* start Lightanda browser
* run the scenario `node puppeteer/click.js`

## Current result

```
> {"method":"Runtime.callFunctionOn","params":{"functionDeclaration":"() => {\n                    return document;\n                }\n//# sourceURL=pptr:evaluateHandle;%23document%20(file%3A%2F%2F%2Fhome%2Fpierre%2Fwrk%2Fdemo%2Fnode_modules%2Fpuppeteer-core%2Flib%2Fesm%2Fpuppeteer%2Fapi%2FFrame.js%3A263%3A52)\n","executionContextId":2,"arguments":[],"returnByValue":false,"awaitPromise":true,"userGesture":true},"id":17,"sessionId":"SID-1"}
< {"id":17,"result":{"result":{"type":"object","subtype":"node","className":"HTMLDocument","description":"","objectId":"5987356902031041503.2.1"}},"sessionId":"SID-1"}
> {"method":"DOM.describeNode","params":{"objectId":"5987356902031041503.2.1"},"id":18,"sessionId":"SID-1"}
< {"id":18,"code":-31998,"message":"UnknownMethod"}
```

PR #524, opened 2025-04-10
describeNode

sjorsdonkers:
Currently is able return correct data for: https://github.com/lightpanda-io/browser/issues/518.

Depends on: https://github.com/lightpanda-io/zig-js-runtime/pull/306

### DIFF
--- a/src/cdp/domains/dom.zig
+++ b/src/cdp/domains/dom.zig
@@ -30,6 +30,7 @@ pub fn processMessage(cmd: anytype) !void {
         getSearchResults,
         discardSearchResults,
         resolveNode,
+        describeNode,
     }, cmd.input.action) orelse return error.UnknownMethod;
 
     switch (action) {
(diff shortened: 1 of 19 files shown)


sjorsdonkers:
It appears that the reason why both the browser and puppeteer timeout may be unrelated to the describeNode call.
Even before we call describeNode Puppeteer has already lost the #context.
image
I think we should handle this in a separate PR

sjorsdonkers:
When determining what should be a node here are some cases to double check in addition to those that have been set to "node".

Set as Node special cases:
Node (itself)
HTMLCollectionIterator (currently handled)
HTMLCollection (currently handled)

Not set as node:
EventTarget
DOMException
DOMImplementation
MutationObserver
MutationRecords
MutationRecord
NamedNodeMap
NodeListIterator
NodeListEntriesIterator
NodeList
ProcessingInstruction
DOMTokenList
Event
U32Iterator
Bottle
URL
URLSearchParams
XMLHttpRequestEventTarget
XMLHttpRequestUpload
XMLHttpRequest
XMLSerializer

PR #526, opened 2025-04-11
Increase timeout to 1min

Bornlex:


karlseguin:
Is the existing `--timeout 60` (i.e. `./lightpanda serve --timeout 60`)  flag what you want?

This indirectly increases the timeout, by increasing the timeout polling, with the downside that you can never decrease it.

PR #529, opened 2025-04-11
Document cookie

### LINKED ISSUE #525: web api: implement document.cookie
We want to add the implementation of `document.cookie` returning the string of cookies for the page.

```
attribute USVString [cookie](https://html.spec.whatwg.org/multipage/dom.html#dom-document-cookie);
```
https://html.spec.whatwg.org/multipage/dom.html#document

* https://html.spec.whatwg.org/multipage/dom.html#dom-document-cookie
* https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie

krichprollsch:
fix #525

### DIFF
--- a/src/html/document.zig
+++ b/src/html/document.zig
@@ -33,6 +33,9 @@ const Location = @import("location.zig").Location;
 const collection = @import("../dom/html_collection.zig");
 const Walker = @import("../dom/walker.zig").WalkerDepthFirst;
 
+const UserContext = @import("../user_context.zig").UserContext;
+const Cookie = @import("../storage/cookie.zig").Cookie;
+
 // WEB IDL https://html.spec.whatwg.org/#the-document-object
 pub const HTMLDocument = struct {
     pub const Self = parser.DocumentHTML;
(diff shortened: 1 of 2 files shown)


src/html/document.zig
```diff
@@ -81,14 +84,18 @@ pub const HTMLDocument = struct {
...
     }
 
-    // TODO: not implemented by libdom
-    pub fn get_cookie(_: *parser.DocumentHTML) ![]const u8 {
-        return error.NotImplemented;
+    pub fn get_cookie(_: *parser.DocumentHTML, allocator: std.mem.Allocator, userctx: UserContext) ![]const u8 {
+        var buf: std.ArrayListUnmanaged(u8) = .{};
+        defer buf.deinit(allocator);
+        try userctx.cookie_jar.forRequest(&userctx.url.uri, buf.writer(allocator), .{ .navigation = true });
+        return buf.toOwnedSlice(allocator);
```
karlseguin: i know this is consistent with a lot of the existing APIs, but if you _know_ that allocator is an arena, that this an unnecessary/extra allocation + copy. Plus, we never free the dupe, so it's a bit like the worst of both: don't assume it's an arena, but also don't free the allocation.

There are comments like https://github.com/lightpanda-io/browser/blob/main/src/url/url.zig#L75  which imply this should be fixed. But I think relying on the page_arena is fine (though, we should rename the allocator parameter `arena`, IMO)

src/storage/cookie.zig
```diff
@@ -145,7 +145,7 @@ pub const Jar = struct {
             if (first) {
                 first = false;
             } else {
-                try writer.writeAll(", ");
+                try writer.writeAll("; ");
```
karlseguin, line 148: The problem with unit test: the test asserts the wrong behavior. Nice catch :)

src/html/document.zig
krichprollsch: I agree, I changed.

src/html/document.zig
```diff
@@ -81,14 +84,17 @@ pub const HTMLDocument = struct {
...
+        var buf: std.ArrayListUnmanaged(u8) = .{};
+        try userctx.cookie_jar.forRequest(&userctx.url.uri, buf.writer(arena), .{ .navigation = true });
+        return buf.items;
     }
 
-    // TODO: not implemented by libdom
-    pub fn set_cookie(_: *parser.DocumentHTML, _: []const u8) ![]const u8 {
-        return error.NotImplemented;
+    pub fn set_cookie(_: *parser.DocumentHTML, arena: std.mem.Allocator, userctx: UserContext, cookie_str: []const u8) ![]const u8 {
+        const c = try Cookie.parse(arena, &userctx.url.uri, cookie_str);
```
karlseguin: This is a problem, the cookie has to outlive the arena (cookies live for the duration of the session, the arena is for the page). You can parse with `userctx.cookie_jar.allocator`  instead of `arena`. 

Not the first case where I think _most_ web apis need the page_arena, but a few places need something with a longer lifetime. In `jsuruntime`, there is no special parameter for Allocator (or Loop), there's simply a `session_state.arena`. We could add a `session_state.session_arena`  and/or a `session_state.allocator`

krichprollsch: Good catch :+1: 

Did you see other places in that case in web api implementations?

karlseguin: Somewhat related, we see this problem with XHR, where the page_arena is disposed of, clearing out the XHR object, but there are still inflight IO events. It crashes the app. It happens if the page load is slow, and the CDP client disconnects due to a timeout. It's on my radar.

The only other thing which I think currently outlives the page is the storage shed, but I just looked at the code again, and it seems OK as it's using the Session's arena.

PR #530, opened 2025-04-11
webapi: assume we are using an arena in web api implementations

krichprollsch:


karlseguin:
PR makes jsruntime a bit tedious to rebase, it already has a lot of these changes.

krichprollsch:
ok, no pb. let's close this one then.

PR #533, opened 2025-04-15
Recreate fake isolated world when cleared

sjorsdonkers:
Based on https://github.com/lightpanda-io/browser/pull/524 as describeNode is needed to recreate the issue this solves.

When executionContextsCleared is called puppetteer drops all references to contexts it had.
When navigating we already send a executionContextCreated message for the main context, however we did not in the case the client had requested an isolated world.

Puppetteer seems to be using the isolated world to make sure that what it is doing does not interfere with the variables of the actual page. We however still us the main context to do everything, and send a fake context id 0.
This PR continues the hack by making sure we resend the isolated context right after send the executionContextsCleared  event.

NOTE: Likely need to change the target before merging!

### DIFF
--- a/src/cdp/cdp.zig
+++ b/src/cdp/cdp.zig
@@ -303,6 +303,9 @@ pub fn BrowserContext(comptime CDP_T: type) type {
         node_registry: Node.Registry,
         node_search_list: Node.Search.List,
 
+        // Holds the data to be able recreate the fake isolated world, see: @isolated_world
+        fake_isolatedworld: ?@import("domains/runtime.zig").ExecutionContextCreated,
+
         const Self = @This();
 
         fn init(self: *Self, id: []const u8, cdp: *CDP_T) !void {
@@ -323,6 +326,7 @@ pub fn BrowserContext(comptime CDP_T: type) type {
                 .page_life_cycle_events = false, // TODO; Target based value
                 .node_registry = registry,
                 .node_search_list = undefined,
+                .fake_isolatedworld = null,
             };
             self.node_search_list = Node.Search.List.init(allocator, &self.node_registry);
         }
--- a/src/cdp/domains/page.zig
+++ b/src/cdp/domains/page.zig
@@ -98,35 +98,46 @@ fn addScriptToEvaluateOnNewDocument(cmd: anytype) !void {
 }
 
 // TODO: hard coded method
+/// @isolated_world The current understanding is that an isolated world should be a separate isolate and context
+/// that would live in the BrowserContext. We think Puppetee creates this to be able to create variables
+/// that are not interfering with the normal namespace of he webpage.
+/// Similar to the main context we need to pretend to recreate it after a executionContextsCleared event
+/// which happens when navigating to a new page.
+/// Since we do not actually create an isolated context operations on this context are still performed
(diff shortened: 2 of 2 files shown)


src/cdp/cdp.zig
```diff
@@ -303,6 +303,9 @@ pub fn BrowserContext(comptime CDP_T: type) type {
         node_registry: Node.Registry,
         node_search_list: Node.Search.List,
 
+        // Holds the data to be able recreate the fake isolated world, see: @isolated_world
+        fake_isolatedworld: ?@import("domains/runtime.zig").ExecutionContextCreated,
```
karlseguin, line 307: Not against having a `src/cdp/execution_context.zig`  with a `Created` struct inside. Also can keep as-is.

src/cdp/domains/page.zig
```diff
@@ -98,35 +98,46 @@ fn addScriptToEvaluateOnNewDocument(cmd: anytype) !void {
...
-            .uniqueId = "7102379147004877974.3265385113993241162",
-            .auxData = .{
-                .isDefault = false,
-                .type = "isolated",
-                .frameId = params.frameId,
-            },
+    const bc = cmd.browser_context orelse return error.BrowserContextNotLoaded;
+    const session_id = cmd.input.session_id orelse return error.SessionIdRequired;
+
+    const name_copy = try bc.session.arena.allocator().dupe(u8, params.worldName);
```
karlseguin, line 118: Could also give the bc its own arena. My only concern with using the Session's arena is that it's easy to forget the BrowserContext is using it.

src/cdp/domains/page.zig
```diff
@@ -202,9 +213,14 @@ pub fn pageNavigate(bc: anytype, event: *const Notification.PageEvent) !void {
...
 
-    // Send Runtime.executionContextsCleared event
     // TODO: noop event, we have no env context at this point, is it necesarry?
+    // Sending this events will tell make the client drop its contexts and wait for new ones to be created.
     try cdp.sendEvent("Runtime.executionContextsCleared", null, .{ .session_id = session_id });
+
+    // When the execution contexts are cleared the client expect us to send executionContextCreated events with the new ID for each context.
+    // Since we do not actually maintain an isolated context we just send the same message as we did initially.
+    // The contextCreated message is send for the main context by the session by calling the inspector.contextCreated when navigating.
+    if (bc.fake_isolatedworld) |isolatedworld| try cdp.sendEvent("Runtime.executionContextCreated", .{ .context = isolatedworld }, .{ .session_id = session_id });
```
karlseguin, line 223: FYI: you can capture a reference to it:

```zig
if (bc.fake_isolatedworld) |*isolatedworld| { .... }
```

sjorsdonkers:
PR put back into draft as the fake isolated context seems like a rabbit hole

src/cdp/domains/page.zig
sjorsdonkers: Good to know. Generally I don't worry about stack copies. I'd be surprised if this would end up in the final binary.

src/cdp/domains/page.zig
sjorsdonkers: Good catch. One more push to making me think that we should stop with the OO separation of these data structs, especially when we start sending notifications from one to the other, while they are just next to each other.

PR #538, opened 2025-04-17
Add Node.$NODE_TYPE class attributes

karlseguin:
These are used by puppeteer when it's trying to execute code to trigger a click.


src/browser/dom/node.zig
```diff
@@ -79,6 +79,18 @@ pub const Node = struct {
         };
     }
 
+    // class attributes
+
+    pub const _ELEMENT_NODE = @intFromEnum(parser.NodeType.element);
```
sjorsdonkers, line 84: Might be good to add a comment why the historical ones are not added
```
    entity_reference = c.DOM_ENTITY_REFERENCE_NODE, // historical
    entity = c.DOM_ENTITY_NODE, // historical
    notation = c.DOM_NOTATION_NODE, // historical
```

sjorsdonkers (approved):
Thanks! 
LGTM

PR #539, opened 2025-04-17
Add v8 steps to make's install and build commands

sjorsdonkers:
I don't know if it was intentional, I liked the easy and clarity of just having to commands, `make install` and `make build`.


krichprollsch:
Well it was intentional.
It adds a step for compilation which take some time even if v8 is up to date.
Moreover, if you pick a [re-compiled](https://github.com/lightpanda-io/zig-v8-fork/releases/tag/v0.1.17) version of v8, this step will force you to compile it from scratch...

PR #541, opened 2025-04-18
Change TypeLookup values from simple index (usize) to a TypeMeta

karlseguin:
TypeMeta constains the index and the subtype. This allows retrieving the subtype based on the actual value being bound, as opposed to the struct type. (I.e. it returns the correct subtype when a Zig class is a proxy for another using the Self declaration).

Internally store the subtype as an enum. Reduces @sizeOf(TaggedAnyOpaque) from 32 to 16.

Finally, sub_type renamed to subtype for consistency with v8.


src/runtime/subtype.zig
```diff
@@ -0,0 +1,20 @@
+pub const SubType = enum {
```
sjorsdonkers, line 1: Subtype? :P

sjorsdonkers (approved):
LGTM

PR #542, opened 2025-04-18
Set Node subtypes

sjorsdonkers:
Based on: https://github.com/lightpanda-io/browser/pull/541
Adds the "node" subtype to types

### DIFF
--- a/src/runtime/js.zig
+++ b/src/runtime/js.zig
@@ -96,9 +96,7 @@ pub fn Env(comptime S: type, comptime types: anytype) type {
                 @compileError(std.fmt.comptimePrint("Prototype '{s}' for type '{s} must be a pointer", .{ @typeName(Struct.prototype), @typeName(Struct) }));
             }
 
-            @setEvalBranchQuota(40_000); // The number of branches increases as we get more structs with the subtype field
-            const subtype: ?SubType =
-                if (@hasDecl(Struct, "subtype")) std.meta.stringToEnum(SubType, Struct.subtype) else null;
+            const subtype: ?SubType = if (@hasDecl(Struct, "subtype")) Struct.subtype else null;
 
             const R = Receiver(@field(types, s.name));
             fields[i] = .{
--- a/src/runtime/js.zig
+++ b/src/runtime/js.zig
@@ -710,8 +710,6 @@ pub fn Env(comptime S: type, comptime types: anytype) type {
                     if (un.tag_type) |UnionTagType| {
                         inline for (un.fields) |field| {
                             if (value == @field(UnionTagType, field.name)) {
-                                const tname = @typeName(@TypeOf(value));
-                                _ = tname;
                                 return zigValueToJs(templates, isolate, context, @field(value, field.name));
                             }
                         }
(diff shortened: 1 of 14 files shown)


krichprollsch:
@sjorsdonkers can you rebase please since #541 is merged?

src/runtime/js.zig
```diff
@@ -709,6 +710,8 @@ pub fn Env(comptime S: type, comptime types: anytype) type {
                     if (un.tag_type) |UnionTagType| {
                         inline for (un.fields) |field| {
                             if (value == @field(UnionTagType, field.name)) {
+                                const tname = @typeName(@TypeOf(value));
+                                _ = tname;
```
krichprollsch: This code looks useless. Is it for a future usage?

src/runtime/js.zig
```diff
@@ -96,6 +96,7 @@ pub fn Env(comptime S: type, comptime types: anytype) type {
                 @compileError(std.fmt.comptimePrint("Prototype '{s}' for type '{s} must be a pointer", .{ @typeName(Struct.prototype), @typeName(Struct) }));
             }
 
+            @setEvalBranchQuota(40_000); // The number of branches increases as we get more structs with the subtype field
```
karlseguin: I wonder if we use the enum in the class definition, whether this would be faster. We wouldn't need to call `std.meta.stringToEnum`, but could instead just do `Struct.subtype`.

The reason I didn't do this is that I didn't really want to force every file to have to `@import("../../jsruntime/subtype.zig").SubType`.

src/runtime/js.zig
```diff
@@ -908,6 +911,8 @@ pub fn Env(comptime S: type, comptime types: anytype) type {
             // we have a v8.Context, we can get the executor.
             fn mapZigInstanceToJs(context: v8.Context, js_obj_or_template: anytype, value: anytype) !PersistentObject {
                 const executor: *Executor = @ptrFromInt(context.getEmbedderData(1).castTo(v8.BigInt).getUint64());
+                const tname = @typeName(@TypeOf(value));
```
karlseguin: ?

src/runtime/js.zig
```diff
@@ -961,6 +966,8 @@ pub fn Env(comptime S: type, comptime types: anytype) type {
                         const isolate = self.isolate;
 
                         if (isEmpty(ptr.child) == false) {
+                            const tname = @typeName(T);
```
karlseguin: ?

src/runtime/js.zig
karlseguin: looks like you can change

```
if (@hasDecl(Struct, "subtype")) std.meta.stringToEnum(SubType, Struct.subtype) else null;
```

to

```
if (@hasDecl(Struct, "subtype")) Struct.subtype else null;
```

use enum literals everywhere:

```
pub const subtype = .none;
```

and (a) hopefully it works and (b) you can remove the @setEvalBranchQuota(40_000);

src/runtime/js.zig
sjorsdonkers: Fixed: [81400be](https://github.com/lightpanda-io/browser/pull/542/commits/81400beeadbdd84d44f2cd07efc64dba13395492)

src/runtime/js.zig
sjorsdonkers: Fixed: [81400be](https://github.com/lightpanda-io/browser/pull/542/commits/81400beeadbdd84d44f2cd07efc64dba13395492)

src/runtime/js.zig
sjorsdonkers: Fixed: [81400be](https://github.com/lightpanda-io/browser/pull/542/commits/81400beeadbdd84d44f2cd07efc64dba13395492)

src/runtime/js.zig
sjorsdonkers: Cool: [6e6660e](https://github.com/lightpanda-io/browser/pull/542/commits/6e6660ea805196f008f410c4085b0e9f5ecfd9df)

PR #543, opened 2025-04-18
descibeNode for new js runtime

sjorsdonkers:
Replaces: https://github.com/lightpanda-io/browser/pull/524

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ ARG ZIG=0.14.0
 ARG ZIG_MINISIG=RWSGOq2NVecA2UPNdBUZykf1CCb147pkmdtYxgb3Ti+JO/wCYvhbAb/U
 ARG ARCH=x86_64
 ARG V8=11.1.134
-ARG ZIG_V8=v0.1.16
+ARG ZIG_V8=v0.1.18
 
 RUN apt-get update -yq && \
     apt-get install -yq xz-utils \
@@ -49,10 +49,6 @@ WORKDIR /browser
 
 # install deps
 RUN git submodule init && \
-    git submodule update --recursive
-
-RUN cd vendor/zig-js-runtime && \
-    git submodule init && \
     git submodule update --recursive
 
 RUN make install-libiconv && \
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -13,8 +13,8 @@
             .hash = "tigerbeetle_io-0.0.0-ViLgxpyRBAB5BMfIcj3KMXfbJzwARs9uSl8aRy2OXULd",
         },
         .v8 = .{
-            .url = "https://github.com/karlseguin/zig-v8-fork/archive/e5f1c0c9f1ed147617427f22cdaf11df4ab60b79.tar.gz",
-            .hash = "v8-0.0.0-xddH61vYIACI2pT1t-dUbXm18cHAKy-KWT_Qft4sBwam",
+            .url = "https://github.com/lightpanda-io/zig-v8-fork/archive/5790c80fcd12dec64e596f6f66f09de567020e8a.tar.gz",
+            .hash = "v8-0.0.0-xddH66roIAAdXNJpBKN_NO8zBz2H8b9moUzshBCfns2p",
         },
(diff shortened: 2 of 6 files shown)


Dockerfile
```diff
@@ -51,10 +51,6 @@ WORKDIR /browser
 RUN git submodule init && \
     git submodule update --recursive
 
-RUN cd vendor/zig-js-runtime && \
```
sjorsdonkers, line 54: Also removed this

build.zig.zon
```diff
@@ -13,8 +13,8 @@
             .hash = "tigerbeetle_io-0.0.0-ViLgxpyRBAB5BMfIcj3KMXfbJzwARs9uSl8aRy2OXULd",
         },
         .v8 = .{
-            .url = "https://github.com/karlseguin/zig-v8-fork/archive/e5f1c0c9f1ed147617427f22cdaf11df4ab60b79.tar.gz",
-            .hash = "v8-0.0.0-xddH61vYIACI2pT1t-dUbXm18cHAKy-KWT_Qft4sBwam",
```
sjorsdonkers, line 17: e5f1c0c9f1ed147617427f22cdaf11df4ab60b79 is the commit hash.
The hash is in the zig's error if it is missing

karlseguin:
Looks good. 

I don't want to beat a dead horse, but.. `getValueByObjectId` takes an `*const Executor`, so that you can return a `Value` (which needs it) and you need the `Value` to get the `TaggedAnyOpaque` for the ptr+subtype. 

It's a bit less direct, and doesn't depend on an Executor, versus:

```zig
pub fn getNodePtr(self: *const Inspector, allocator: Allocator, , object_id: []const u8) !?*anyopaque {
  const unwrapped = try self.session.unwrapObject(allocator, object_id);
  const toa = getTaggedAnyOpaque(unwrapped.value) orelse return null;
  if (toa.subtype == null or toa.subtype != .node) return error.ObjectIdIsNotANode;
  return toa.ptr;
}
```

src/cdp/domains/dom.zig
```diff
@@ -151,6 +153,37 @@ fn resolveNode(cmd: anytype) !void {
...
+    if (params.backendNodeId != null or params.depth != 1 or params.pierce) {
+        return error.NotYetImplementedParams;
+    }
+
+    const bc = cmd.browser_context orelse return error.BrowserContextNotLoaded;
+
+    if (params.nodeId != null) {
+        const node = bc.node_registry.lookup_by_id.get(params.nodeId.?) orelse return error.NodeNotFound;
+        return cmd.sendResult(.{ .node = bc.nodeWriter(node, .{}) }, .{});
+    } else if (params.objectId != null) {
```
krichprollsch: you don't really need a `else` here isn't it?

I would prefer have:
```zig
if (params.nodeId != null) {
    // ...
}

if (params.objectId != null) {
    // ...
}

return error.MissingParams;
```

src/runtime/js.zig
```diff
@@ -1306,6 +1306,15 @@ pub fn Env(comptime S: type, comptime types: anytype) type {
                     generate_preview,
                 );
             }
+
+            // Gets a value by object ID regardless of which context it is in.
+            // unwrapping the object also tells us the context, for now we assume it is always the default one.
+            // The executor argument is likely to change to somthing to allow us to find the right Executer with the given context
```
```suggestion
            // The executor argument is likely to change to something to allow us to find the right Executor with the given context
```

krichprollsch (approved):

src/runtime/js.zig
sjorsdonkers: Removed [a698ff8](https://github.com/lightpanda-io/browser/pull/543/commits/a698ff8309429b2e9c480c150ea1441817fb4d5c)

src/cdp/domains/dom.zig
sjorsdonkers: Changed : [a698ff8](https://github.com/lightpanda-io/browser/pull/543/commits/a698ff8309429b2e9c480c150ea1441817fb4d5c)
Note that it is correct since every if branch returns.
The if else structure was indicating that only one of the paths should be taking.

PR #545, opened 2025-04-18
feat: update Dockerfile&add github action 

ryoppippi:
Hi!
I tried to build lightpanda's docker image, but because of recent changes in v8 I couldn't build it. Also I found that there is no ci on Dockerfile. In addition, this dockerfile doesn't use multi-stage build efficiency.
So this pr includes
- enhance multi-stage building
- add ci
- Automatic detection of cpu archtecture.

You can see the build result here

https://github.com/ryoppippi/browser/actions/runs/14534577660/job/40780436433

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,38 +1,21 @@
-FROM ubuntu:24.04
+# TARGETARCH is a built-in argument provided by Docker, representing the target architecture (e.g., amd64, arm64)
+# https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope
 
 ARG MINISIG=0.12
 ARG ZIG=0.14.0
 ARG ZIG_MINISIG=RWSGOq2NVecA2UPNdBUZykf1CCb147pkmdtYxgb3Ti+JO/wCYvhbAb/U
+ARG V8=11.1.134
 ARG ZIG_V8=v0.1.17
 
+# Define the base image alias
+FROM ubuntu:24.04 AS base
+
+# Stage 1: Clone source code
+FROM base AS source
+
+# Install dependencies for this stage
 RUN apt-get update -yq && \
-    apt-get install -yq xz-utils \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,78 +1,137 @@
...
 
-# build release
-RUN make build
+WORKDIR /browser
 
-FROM ubuntu:24.04
+# Install build dependencies from source
+RUN make install-libiconv
+RUN make install-netsurf
+RUN make install-mimalloc
```
ryoppippi, line 110: Separated the `make` commands because we can use build cache in each command.
We use multi-stage build so we don't have to worry about the final image size

Dockerfile
```diff
@@ -1,78 +1,137 @@
...
+    esac && \
+    echo "Using V8 build architecture path: ${MAPPED_V8_BUILD_ARCH} based on TARGETARCH: ${TARGETARCH}" && \
+    mkdir -p v8/build/${MAPPED_V8_BUILD_ARCH}-linux/release/ninja/obj/zig/ && \
+    mkdir -p v8/build/${MAPPED_V8_BUILD_ARCH}-linux/debug/ninja/obj/zig/ && \
+    cp /libc_v8.a v8/build/${MAPPED_V8_BUILD_ARCH}-linux/release/ninja/obj/zig/libc_v8.a && \
+    cp /libc_v8.a v8/build/${MAPPED_V8_BUILD_ARCH}-linux/debug/ninja/obj/zig/libc_v8.a
+
+# Build release
+# Ensure git context is available for rev-parse
+RUN zig build --release=safe -Doptimize=ReleaseSafe -Dgit_commit=$(git rev-parse --short HEAD)
```
ryoppippi, line 127: we need to fix Makefile later actually...
I copy-n-pasted it from `build.yml`

ryoppippi:
If I need to change something, let me know! please write it down here!

PR #547, opened 2025-04-19
Re-introduce call_arena

karlseguin:
Because of callbacks, calls into Zig can be nested. Previously, the call_arena was reset after _every_ call. When calls are nested, this doesn't work - the nested call resets the arena, which the caller might still need. A `call_depth` integer was added to the Executor. Each call starts by incrementing the call_depth and, on deinit, decrements the call_depth. The call_arena is only reset when the call_depth == 0. This could result in lower memory use.

Also promoted the call_arena and scope_arena to the Env. Practically speaking, nothing has changed, since they're still reset under the exact same conditions. However, when an executor ends and a new one is started, it can now reuse the retained_capacity of the previous arenas. This should result in fewer allocations.

### DIFF
--- a/src/browser/dom/mutation_observer.zig
+++ b/src/browser/dom/mutation_observer.zig
@@ -284,7 +284,7 @@ const EventHandler = struct {
         const muevt = parser.eventToMutationEvent(evt.?);
 
         // TODO get the allocator by another way?
-        const alloc = data.cbk.executor.call_arena.allocator();
+        const alloc = data.cbk.executor.scope_arena;
 
         if (std.mem.eql(u8, t, "DOMAttrModified")) {
             mrs.first = .{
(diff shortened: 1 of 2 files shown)


krichprollsch (approved):

krichprollsch:
@karlseguin can you rebase the branch please?

karlseguin:
@krichprollsch done

karlseguin:
@krichprollsch This is ready. I reverted part of the change and keept the ArenaAllocator tied to the Executor. This should make having multiple Executors safe (it might be safe sharing it, but it would depend on how/if they interact, so this is just safer). PR still re-introduces the call_arena though, and the call_depth (which can be useful for other things), so still worth merging.

krichprollsch (approved):

PR #554, opened 2025-04-22
Isolated world

sjorsdonkers:


src/runtime/js.zig
```diff
@@ -348,7 +348,7 @@ pub fn Env(comptime S: type, comptime types: anytype) type {
                 },
             };
 
-            errdefer self.stopExecutor(executor);
+            errdefer self.stopExecutor(executor, true); // Note: This likely has issues as context.exit() is errdefered as well
```
sjorsdonkers: Note related to this PR, but this errdefer after another errdefer did not look correct

src/cdp/cdp.zig
```diff
@@ -437,6 +468,20 @@ pub fn BrowserContext(comptime CDP_T: type) type {
...
+/// with the normal namespace and values of the webpage. Similar to the main context we need to pretend to recreate it after
+/// a executionContextsCleared event which happens when navigating to a new page. A client can have a command be executed
+/// in the isolated world by using its Context ID or the worldName.
+/// grantUniveralAccess Indecated whether the isolated world has access to objects like the DOM or other JS Objects.
+/// Generally the client needs to resolve a node into the isolated world to be able to work with it.
+/// An object id is unique across all contexts, different object ids can refer to the same Node in different contexts.
+pub const IsolatedWorld = struct {
+    name: []const u8,
+    grant_universal_access: bool,
+    executor: *@import("../browser/env.zig").Env.Executor,
```
sjorsdonkers: Should `Executor` be called a JS `World` instead. @karlseguin

karlseguin: They're both pretty vague - don't think there's much difference between the two.  I actually like "Session" the best, but that's overloaded, so.

sjorsdonkers: World is not a new word, see: https://chromium.googlesource.com/chromium/src/+/master/third_party/blink/renderer/bindings/core/v8/V8BindingDesign.md#world
It seems an Executor closely models a World

src/runtime/js.zig
```diff
@@ -415,7 +422,7 @@ pub fn Env(comptime S: type, comptime types: anytype) type {
         // object (i.e. the Window), which gets attached not only to the Window
         // constructor/FunctionTemplate as normal, but also through the default
         // FunctionTemplate of the isolate (in startExecutor)
-        fn attachClass(self: *Self, comptime Struct: type, template: v8.FunctionTemplate) void {
+        fn attachClass(self: *const Self, comptime Struct: type, template: v8.FunctionTemplate) void {
```
sjorsdonkers, line 423: Looks like these functions should not be part of the struct

src/cdp/testing.zig
```diff
@@ -107,12 +112,43 @@ const Session = struct {
     }
 };
 
-const Executor = struct {};
+const Env = struct {
```
sjorsdonkers, line 115: These generic CDPT interfaces seem to give us very little for the amount of complexity they introduce

src/cdp/cdp.zig
```diff
@@ -437,6 +471,24 @@ pub fn BrowserContext(comptime CDP_T: type) type {
...
+/// a executionContextsCleared event which happens when navigating to a new page. A client can have a command be executed
+/// in the isolated world by using its Context ID or the worldName.
+/// grantUniveralAccess Indecated whether the isolated world can reference objects like the DOM or other JS Objects.
+/// An isolated world has it's own instance of globals like Window.
+/// Generally the client needs to resolve a node into the isolated world to be able to work with it.
+/// An object id is unique across all contexts, different object ids can refer to the same Node in different contexts.
+pub fn IsolatedWorld(comptime E: type) type {
+    return struct {
+        name: []const u8,
+        grant_universal_access: bool,
```
sjorsdonkers, line 487: We do not necessarily need to store these as they are not used at the moment.
Alternatively we could consider moving them into Executor as I expect that to merge towards being a Generic World type

src/cdp/testing.zig
karlseguin: Might be possible to partially remove now and load the entire real env. Would probably still be a generic over the client though.

src/runtime/js.zig
```diff
@@ -265,10 +265,12 @@ pub fn Env(comptime S: type, comptime types: anytype) type {
             self.isolate.performMicrotasksCheckpoint();
         }
 
-        pub fn startExecutor(self: *Self, comptime Global: type, state: State, module_loader: anytype) !*Executor {
+        pub fn startExecutor(self: *Self, comptime Global: type, state: State, module_loader: anytype, kind: WorldKind) !*Executor {
             if (comptime builtin.mode == .Debug) {
-                std.debug.assert(self.has_executor == false);
-                self.has_executor = true;
+                if (kind == .main) {
```
karlseguin, line 270: In `main`, the debug `has_executor` check has been removed to accommodate this PR. With it removed, you should be able to remove `kind`.

sjorsdonkers: When getting to this statement myself I initially thought about removing it as well, however I found I could also simplify the opening and closing of the context with it. The I also thought it would just be good metadata to have, so I'd prefer keeping it regardless.

krichprollsch:
@sjorsdonkers can you upgrade the `zig-v8` dependency please?

src/cdp/domains/page.zig
```diff
@@ -97,37 +99,28 @@ fn addScriptToEvaluateOnNewDocument(cmd: anytype) !void {
...
+    if (!params.grantUniveralAccess) {
+        std.debug.print("grantUniveralAccess == false is not yet implemented", .{});
+        // When grantUniveralAccess == false and the client attempts to resolve
+        // or otherwise access a DOM or other JS Object from another context that should fail.
+    }
+    const bc = cmd.browser_context orelse return error.BrowserContextNotLoaded;
 
-    // noop executionContextCreated event
-    try cmd.sendEvent("Runtime.executionContextCreated", .{
-        .context = runtime.ExecutionContextCreated{
```
krichprollsch, line 114: I think we can also remove the `runtime.ExecutionContextCreated` struct from `cdp/runtime.zig`

krichprollsch (approved):

src/cdp/domains/page.zig
sjorsdonkers: Fixed: [6f9b9a5](https://github.com/lightpanda-io/browser/pull/554/commits/6f9b9a5def9ceb39f22b57ffd1114d7494f519cc)

PR #562, opened 2025-04-23
Improve MutationObserver

karlseguin:
- Fix get_removedNodes (it was returning addedNodes) 
- get_removedNodes and get addedNodes now return references 
- used enum for dispatching and clean up dispatching in general 
- Remove MutationRecords and simply return an array
  - this allows the returned records to be iterable (as they should be)
  - jsruntime ZigToJs will now map a Zig array to a JS array
-Rely on default initialize of NodeList
-Batch observed records
 - Callback only executed when call_depth == 0
 - Fixes crashes when a MutationObserver callback mutated the nodes being observes.
 - Fixes some WPT issues, but Netsurf's mutationEventRelatedNode does not appear to be to spec, so most tests fail.
 - Allow zig methods to execute arbitrary code when call_depth == 0
   - This is a preview of how I hope to make XHR not crash if the CDP session ends while there's still network activity


src/browser/dom/mutation_observer.zig
```diff
@@ -152,187 +143,200 @@ pub const MutationObserver = struct {
...
-    fn apply(o: *MutationObserver.Observer, target: *parser.Node) bool {
+const Observer = struct {
+    node: *parser.Node,
+    options: MutationObserverInit,
+
+    // record of the mutation, all observed changes in 1 call are batched
+    record: ?MutationRecord = null,
+
+    // reference back to the MutationObserver so that we can access the arena
+    // and batch the mutation record.s
```
krichprollsch:
```suggestion
    // and batch the mutation records.
```

krichprollsch (approved):

PR #566, opened 2025-04-24
pass null namespace/prefix to libdom

karlseguin:
This change, in combination with https://github.com/lightpanda-io/libdom/pull/22 fixes a number of crashes when using these functions.


krichprollsch:
@karlseguin should we upgrade libdom's submodule at the same time?

karlseguin:
@krichprollsch done

PR #568, opened 2025-04-24
Introduce more general notification capabilities

karlseguin:
Replaces the existing, very specialized Notification with something more general.

Currently, the existing page_navigate and page_navigated have been migrated.

Telemetry's page navigation event now also hooks into these events to generate the telemetry record.

There's a comment on the Notification class that talks about scoping, but...

I initially thought I'd have a global singleton, and it would be really easy to register and dispatch events. But then I realized that wouldn't work. If everything's a global, and CDP is registered for "http_bytes_received", then when Telemetry makes an HTTP request, and the client emits a "http_bytes_received" message, CDP will get those messages.

I realized things needed to be scoped. I thought about introducing a "scope_id", but decided to create a notification-per-scope. Currently, there's a `Notification` per `Browser` and one `Notification` for the `App`. I don't expect more. You can form notification topologies, which is exactly how Telemetry now works. Telemetry listens to the `notification_created` event from `app.notification`, and then registers for the `page_navigate` on the newly created `notification` instance.

### DIFF
--- a/src/app.zig
+++ b/src/app.zig
@@ -5,6 +5,7 @@ const js = @import("runtime/js.zig");
 const Loop = @import("runtime/loop.zig").Loop;
 const HttpClient = @import("http/client.zig").Client;
 const Telemetry = @import("telemetry/telemetry.zig").Telemetry;
+const Notification = @import("notification.zig").Notification;
 
 const log = std.log.scoped(.app);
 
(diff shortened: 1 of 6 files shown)


karlseguin:
I'd rather merge this _after_ the isolated world changes are settled.

PR #569, opened 2025-04-24
Make CDP less generic.

karlseguin:
It's still generic over the client - we need to assert messages written to and be able to send specific commands, but it's no longer generic over Browser/ Session/Page/etc..

### DIFF
--- a/src/cdp/cdp.zig
+++ b/src/cdp/cdp.zig
@@ -22,6 +22,8 @@ const json = std.json;
 
 const App = @import("../app.zig").App;
 const asUint = @import("../str/parser.zig").asUint;
+const Browser = @import("../browser/browser.zig").Browser;
+const Session = @import("../browser/browser.zig").Session;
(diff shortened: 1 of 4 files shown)


src/cdp/cdp.zig
```diff
@@ -309,7 +308,7 @@ pub fn BrowserContext(comptime CDP_T: type) type {
         node_registry: Node.Registry,
         node_search_list: Node.Search.List,
 
-        isolated_world: ?IsolatedWorld(CDP_T.Browser.EnvType),
+        isolated_world: ?IsolatedWorld(Browser.EnvType),
```
sjorsdonkers: I'm not sure.
Does Env still need to be generic? Or can it always just be `browser/env.zig`'s Env.
Not just here, I mean in general.

karlseguin: Missed that one. It can always be the real env.

krichprollsch (approved):

PR #571, opened 2025-04-24
Remove log.zig

sjorsdonkers:
log.zig does not appear to be used


karlseguin:
I plan on working on it at some point, but you can remove it for now, I have a better version in a local branch.

PR #575, opened 2025-04-25
Scope a request entry

sjorsdonkers:
Instead of maintaining a scop in Env, Executor and for a Page we enter and exit a scope upon receiving a call or a command.
The problem is that an Isolate has only 1 scope stack, but we have multiple contexts in flight at the same time.
This caused a problem where a page would close the scope, which would also remove the context that the isolatedworld put in there.

By closing all the scopes on exit of a command or call this should not be an issue.
Instead of storing globals in the Env scope, they are now persisted.
Also the context itself needs to persist, so we need to manually reset it when the context is deleted.

### DIFF
--- a/src/cdp/cdp.zig
+++ b/src/cdp/cdp.zig
@@ -168,6 +168,11 @@ pub fn CDPT(comptime TypeProvider: type) type {
         }
 
         fn dispatchCommand(command: anytype, method: []const u8) !void {
+            const v8 = @import("v8");
+            var temp_scope: v8.HandleScope = undefined;
+            v8.HandleScope.init(&temp_scope, command.cdp.browser.env.isolate);
+            defer temp_scope.deinit();
+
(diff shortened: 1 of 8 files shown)


src/cdp/cdp.zig
```diff
@@ -340,7 +345,7 @@ pub fn BrowserContext(comptime CDP_T: type) type {
 
         pub fn deinit(self: *Self) void {
             if (self.isolated_world) |isolated_world| {
-                isolated_world.executor.endScope();
+                isolated_world.executor.context.enter();
```
sjorsdonkers, line 348: TODO cleanup

sjorsdonkers:
PR closed in favor of https://github.com/lightpanda-io/browser/pull/578
May need to reevaluate when we have multiple worlds.
Some of the persistence change have been adopted in 578

PR #578, opened 2025-04-29
Reorganize v8 contexts and scope

karlseguin:
PR does 2 cleanups and 1 hack.

1 - Moves a lot of the `Executor` and `Session` logic/data down to the `Scope` and the `Page`. This provides correct isolation when navigating from page-to-page and minimizes some of the v8 scoping. A Page/Scope now owns the `window` (globla), `session_state` (ooopps, should rename this to `page_state`?) and its own v8.Context.

2 - Moves the inspector out of the `Session` and into the `BrowserContext`. The inspector, it's callback hookup, it's `aux_data` and everything else, is fully managed by the CDP layer. Browser/Session are now inspector-unaware.

3 - To fix the isolated world crash, I force-create the isolated world _before_ the page is created and always shutdown the page before shutting down the isolated world.

(3) is a hack, but I'm still not sure what the correct long term solution is, but I think (1) and (2) move us in a better direction.

### DIFF
--- a/src/browser/browser.zig
+++ b/src/browser/browser.zig
@@ -85,7 +85,7 @@ pub const Browser = struct {
 
     pub fn newSession(self: *Browser, ctx: anytype) !*Session {
         self.closeSession();
-        self.session = undefined;
+        self.session = @as(Session, undefined);
         const session = &self.session.?;
         try Session.init(session, self, ctx);
         return session;
@@ -169,7 +169,7 @@ pub const Session = struct {
         const page_arena = &self.browser.page_arena;
         _ = page_arena.reset(.{ .retain_with_limit = 1 * 1024 * 1024 });
 
-        self.page = undefined;
+        self.page = @as(Page, undefined);
         const page = &self.page.?;
         try Page.init(page, page_arena.allocator(), self);
 
--- a/src/browser/browser.zig
+++ b/src/browser/browser.zig
@@ -173,10 +173,6 @@ pub const Session = struct {
         const page = &self.page.?;
         try Page.init(page, page_arena.allocator(), self);
 
-        self.notify(&.{ .context_created = .{
-            .origin = try page.origin(),
-        } });
-
         // start JS env
         log.debug("start new js scope", .{});
 
@@ -331,9 +327,9 @@ pub const Page = struct {
         log.debug("wait: OK", .{});
     }
 
-    fn origin(self: *const Page) ![]const u8 {
+    pub fn origin(self: *const Page, arena: Allocator) ![]const u8 {
         var arr: std.ArrayListUnmanaged(u8) = .{};
-        try self.url.origin(arr.writer(self.arena));
+        try self.url.origin(arr.writer(arena));
         return arr.items;
     }
 
@@ -367,10 +363,6 @@ pub const Page = struct {
             .url = &self.url,
             .reason = opts.reason,
             .timestamp = timestamp(),
-        } });
-
-        session.notify(&.{ .context_created = .{
-            .origin = try self.origin(),
         } });
 
         var response = try request.sendSync(.{});
@@ -449,7 +441,7 @@ pub const Page = struct {
         // TODO set the referrer to the document.
         try self.window.replaceDocument(html_doc);
         self.window.setStorageShelf(
-            try self.session.storage_shed.getOrPut(try self.origin()),
+            try self.session.storage_shed.getOrPut(try self.origin(self.arena)),
         );
 
         // https://html.spec.whatwg.org/#read-html
--- a/src/browser/browser.zig
+++ b/src/browser/browser.zig
@@ -173,6 +173,10 @@ pub const Session = struct {
         const page = &self.page.?;
         try Page.init(page, page_arena.allocator(), self);
 
+        self.notify(&.{ .context_created = .{
+            .origin = try page.origin(),
+        } });
+
         // start JS env
         log.debug("start new js scope", .{});
 
--- a/src/browser/browser.zig
+++ b/src/browser/browser.zig
@@ -20,6 +20,7 @@ const std = @import("std");
 const builtin = @import("builtin");
 
 const Allocator = std.mem.Allocator;
+const ArenaAllocator = std.heap.ArenaAllocator;
 
 const Dump = @import("dump.zig");
 const Mime = @import("mime.zig").Mime;
@@ -53,13 +54,10 @@ pub const user_agent = "Lightpanda/1.0";
 pub const Browser = struct {
     env: *Env,
     app: *App,
-    session: ?*Session,
+    session: ?Session,
     allocator: Allocator,
     http_client: *http.Client,
-    session_pool: SessionPool,
-    page_arena: std.heap.ArenaAllocator,
-
-    const SessionPool = std.heap.MemoryPool(Session);
+    page_arena: ArenaAllocator,
 
     pub fn init(app: *App) !Browser {
         const allocator = app.allocator;
@@ -75,31 +73,27 @@ pub const Browser = struct {
             .session = null,
             .allocator = allocator,
             .http_client = &app.http_client,
-            .session_pool = SessionPool.init(allocator),
-            .page_arena = std.heap.ArenaAllocator.init(allocator),
+            .page_arena = ArenaAllocator.init(allocator),
         };
     }
 
     pub fn deinit(self: *Browser) void {
         self.closeSession();
         self.env.deinit();
-        self.session_pool.deinit();
         self.page_arena.deinit();
     }
 
     pub fn newSession(self: *Browser, ctx: anytype) !*Session {
         self.closeSession();
-
-        const session = try self.session_pool.create();
+        self.session = undefined;
+        const session = &self.session.?;
         try Session.init(session, self, ctx);
-        self.session = session;
         return session;
     }
 
     pub fn closeSession(self: *Browser) void {
-        if (self.session) |session| {
+        if (self.session) |*session| {
             session.deinit();
-            self.session_pool.destroy(session);
             self.session = null;
         }
     }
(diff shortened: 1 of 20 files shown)


karlseguin:
I initially wanted to have the `contextCreated` events we emit be based on what actually happens. That's how the first commit worked. But it didn't work on chromedp - I missed a message.

It turns out that:
1 - chromedp wants a contextCreated message emitted when the target is created (I assume this relates to the tab)
2 - puppeteer (and possibly chromedp also) wants a contextCreated prior to the `Page.frameNavigated` (puppeteer is pretty sensitive to the timing).
3 - puppeteer also wants a context created when the isolate world is created.

No doubt this is how chrome actually behaves. A context for the tab. A context for the page (and then one per frame) and, if you create an isolated world, a context for that.

Since we aren't currently creating 3 contexts, I decided to go back to fully faking this. However thanks to (1) from the PR details above, this is less of a lie than it was before and thanks to (2) this mess is completely removed from Browser - it's purely a CDP concern.

Now, whether we actually need 3+ distinct contexts created, and whether they need to be created at those exact points in time, is still TDB.

src/cdp/domains/target.zig
```diff
@@ -122,14 +122,21 @@ fn createTarget(cmd: anytype) !void {
...
 
-    // start the js env
-    const aux_data = try std.fmt.allocPrint(
-        cmd.arena,
-        // NOTE: we assume this is the default web page
-        "{{\"isDefault\":true,\"type\":\"default\",\"frameId\":\"{s}\"}}",
-        .{target_id},
-    );
-    _ = try bc.session.createPage(aux_data);
+    try bc.createIsolatedWorld();
```
sjorsdonkers, line 125: How would this work for multiple isolated or worker worlds?
I think if we just make the context persistent we do not have to make them ahead of time.

src/runtime/js.zig
```diff
@@ -259,610 +257,292 @@ pub fn Env(comptime S: type, comptime types: anytype) type {
...
-        const WorldKind = enum {
-            main,
-            isolated,
-            worker,
-        };
-
-        // This is capable of executing JavaScript.
-        pub const Executor = struct {
+        // Loosely maps to a Browser Page.
+        pub const Scope = struct {
```
sjorsdonkers, line 453: Naming:
Since this is significantly heavier than a scope `{}`.
I think it is closer to Deno's ContextScope, which may be a name we can adopt.

src/runtime/js.zig
```diff
@@ -259,610 +257,292 @@ pub fn Env(comptime S: type, comptime types: anytype) type {
...
-                        if (ptr.child == u8) {
-                            // Need to do this to keep the compiler happy
-                            // If this was the case, simpleZigValueToJs would
-                            // have handled it
-                            unreachable;
+                var handle_scope: v8.HandleScope = undefined;
+                v8.HandleScope.init(&handle_scope, isolate);
+                errdefer handle_scope.deinit();
+
+                const context = v8.Context.init(isolate, global_template, null);
```
sjorsdonkers, line 379: For correctness I think we can take the concept of making the context persistent from the other PR: https://github.com/lightpanda-io/browser/blob/008d4241435a627db301c9aca13b02ddbd2cba44/src/runtime/js.zig#L293

src/runtime/js.zig
```diff
@@ -162,7 +162,7 @@ pub fn Env(comptime S: type, comptime types: anytype) type {
         global_scope: v8.HandleScope,
 
         // just kept around because we need to free it on deinit
-        isolate_params: v8.CreateParams,
+        isolate_params: *v8.CreateParams,
```
sjorsdonkers, line 162: I think we can remove `global_scope` if we make the the templates persistent (and start the scope a bit earlier in `startScope(..)`), see [below](https://github.com/lightpanda-io/browser/pull/580/files)

src/runtime/js.zig
```diff
@@ -259,610 +257,292 @@ pub fn Env(comptime S: type, comptime types: anytype) type {
...
-            if (simpleZigValueToJs(isolate, value, false)) |js_value| {
-                return js_value;
-            }
+                // All the FunctionTemplates that we created and setup in Env.init
+                // are now going to get associated with our global instance.
+                const templates = &self.env.templates;
+                inline for (Types, 0..) |s, i| {
+                    const Struct = @field(types, s.name);
+                    const class_name = v8.String.initUtf8(isolate, comptime classNameForStruct(Struct));
+                    global_template.set(class_name.toName(), templates[i], v8.PropertyAttribute.None);
```
sjorsdonkers, line 364: These now all live in global_scope, but it is hard to see that leak.

sjorsdonkers: See fix in this PR: https://github.com/lightpanda-io/browser/pull/580/

src/runtime/js.zig
sjorsdonkers: By making the context persistent we do not need to keep the scope alive.
We can decide whether to keep it open for the main page, or keep it closed and only open it when needed for the isolated World. It think that would solve the issue of multiple isolated / worker worlds.
I'm not sure if that would solve multiple frames in the future if that is a problem.

src/cdp/domains/target.zig
sjorsdonkers: See: https://github.com/lightpanda-io/browser/pull/578#discussion_r2065920808

src/runtime/js.zig
karlseguin: I agree this is an issue. I'd actually like to define class_name in the Env and re-use the value, if that works. Better than having to re-create it on each scope. But, your point still stands.

I'm curious, if handle_scope was to become local and cleared on function end, we'd have to make everything in here a persistent to outlive the handle_scope?

src/cdp/domains/target.zig
karlseguin: What this does is ensure that the we enter and exit the contexts in the correct order. I'm not sure I see how that's related to having a persisted context. Being persisted doesn't change the entering/exiting ordering, does it? If you have multiple isolated workers, when are they created? What's their lifetime? 

In chrome they're created at (1) tab context, (2) isolated world context and (3) page context.

In Browser, we have (1) page  (2) isolated context. This line of code restores the order so that the isolated context is always created before the page context.

src/runtime/js.zig
karlseguin: I dunno, are two generic words better than one?  

Is `executor.context_scope` any more informative than `executor.scope`

src/runtime/js.zig
sjorsdonkers: > I'm curious, if handle_scope was to become local and cleared on function end, we'd have to make everything in here a persistent to outlive the handle_scope?

Surprisingly the answer is no. They just need to live long enough to make the context, but since we want the context to live in the scope we keep, we can't just make a temp HandleScope here.
In the [noscope branch](https://github.com/lightpanda-io/browser/pull/575/files) the HandleScope used here is temporary, but Context is made persistent to outlive it.

src/cdp/domains/target.zig
sjorsdonkers: Right, my thought was if a user is going to make 15 isolated worlds, we can't be making them ahead of time.
If we want to support the worlds would need to be created dynamically, which bring us back to the original problem of worlds being created on the scope of something with a shorter lifetime.

src/runtime/js.zig
sjorsdonkers: Meh🙂

sjorsdonkers:
Note: Requires: https://github.com/lightpanda-io/zig-v8-fork/pull/61

src/cdp/domains/dom.zig
```diff
@@ -163,28 +166,61 @@ fn resolveNode(cmd: anytype) !void {
...
+    const arena = cmd.arena;
+    return cmd.sendResult(.{ .object = .{
+        .type = try remote_object.getType(arena),
+        .subtype = try remote_object.getSubtype(arena),
+        .className = try remote_object.getClassName(arena),
+        .description = try remote_object.getDescription(arena),
+        .objectId = try remote_object.getObjectId(arena),
+    } }, .{});
+
+    // const params = (try cmd.params(struct {
```
sjorsdonkers: Seems rebasing didn't go without issues, not sure if other files are affected as well

src/browser/browser.zig
```diff
@@ -159,130 +136,59 @@ pub const Session = struct {
...
-    pub fn createPage(self: *Session, aux_data: ?[]const u8) !*Page {
+    pub fn createPage(self: *Session) !*Page {
         std.debug.assert(self.page == null);
 
-        _ = self.browser.page_arena.reset(.{ .retain_with_limit = 1 * 1024 * 1024 });
+        const page_arena = &self.browser.page_arena;
+        _ = page_arena.reset(.{ .retain_with_limit = 1 * 1024 * 1024 });
 
-        self.page = Page.init(self);
+        self.page = @as(Page, undefined);
```
sjorsdonkers, line 172: Is this OK?
In `Page.init` we are using the values of the current state as input for the new state!

sjorsdonkers (approved):

src/cdp/cdp.zig
```diff
@@ -353,25 +359,25 @@ pub fn BrowserContext(comptime CDP_T: type) type {
...
+                .name = "",
+                .global = .{},
+                .scope = undefined,
                 .executor = executor,
+                .grant_universal_access = false,
             };
+            var world = &self.isolated_world.?;
+
+            // TODO: can we do something better than passing `undefined` for the state?
+            world.scope = try world.executor.startScope(&world.global, undefined, {});
```
sjorsdonkers, line 380: Just a note, doesn't need to be fixed in this PR.
(When grantUniversalAccess is true)
I'm fairly certain we should put in the page's Window, such that they share the same DOM tree.
I expect we keep running into issues if we combine steps like creating a scope and a context.

src/browser/browser.zig
karlseguin: I'm not sure what you mean by "the values of the current state".  Do you mean things like:

```zig
self.* = .{
  .url = URL.empty,
  .state = .{
    .url = &self.url,
  }
};
```
Isn't `state.url` being given a valid/correct address? The value might not be initialized, but it will be by the time the statement end.

PR #580, opened 2025-04-29
Remove global scope

sjorsdonkers:
- Fixes some locals that were leaking onto the global scope

### DIFF
--- a/src/runtime/js.zig
+++ b/src/runtime/js.zig
@@ -158,9 +158,6 @@ pub fn Env(comptime S: type, comptime types: anytype) type {
         // the global isolate
         isolate: v8.Isolate,
 
-        // this is the global scope that all our classes are defined in
-        global_scope: v8.HandleScope,
-
         // just kept around because we need to free it on deinit
         isolate_params: *v8.CreateParams,
 
@@ -205,9 +202,9 @@ pub fn Env(comptime S: type, comptime types: anytype) type {
             isolate.enter();
             errdefer isolate.exit();
 
-            var global_scope: v8.HandleScope = undefined;
-            v8.HandleScope.init(&global_scope, isolate);
-            errdefer global_scope.deinit();
+            var temp_scope: v8.HandleScope = undefined;
+            v8.HandleScope.init(&temp_scope, isolate);
+            defer temp_scope.deinit();
 
             const env = try allocator.create(Self);
             errdefer allocator.destroy(env);
@@ -218,7 +215,6 @@ pub fn Env(comptime S: type, comptime types: anytype) type {
                 .allocator = allocator,
                 .isolate_params = params,
                 .gc_hints = opts.gc_hints,
-                .global_scope = global_scope,
                 .prototype_lookup = undefined,
             };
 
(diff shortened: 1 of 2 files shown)


src/runtime/js.zig
```diff
@@ -305,20 +301,25 @@ pub fn Env(comptime S: type, comptime types: anytype) type {
...
+                if (self.scope) |scope| {
+                    const isolate = scope.isolate;
                     self.endScope();
+
+                    // V8 doesn't immediately free memory associated with
+                    // a Context, it's managed by the garbage collector. So, when the
+                    // `gc_hints` option is enabled, we'll use the `lowMemoryNotification`
+                    // call on the isolate to encourage v8 to free any contexts which
+                    // have been freed.
+                    if (self.env.gc_hints) {
```
sjorsdonkers, line 313: gc_hints is not accessible from scope otherwise it would be better to move this into `scope.deinit()`

karlseguin: If we start to have narrow and narrow scopes, wouldn't that possibly trigger _a lot_ of GCs?

sjorsdonkers: I see what you mean. I just thought the code I wrote here was ugly.
Note that what we call scope here is what I think we should call a ContextScope. So it would not be very narrow.
But even then you are probably right, as I don't think we would need to call it at every page navigate.

src/runtime/js.zig
```diff
@@ -228,7 +224,8 @@ pub fn Env(comptime S: type, comptime types: anytype) type {
             // we can get its index via: @field(TYPE_LOOKUP, type_name).index
             const templates = &env.templates;
             inline for (Types, 0..) |s, i| {
-                templates[i] = generateClass(@field(types, s.name), isolate);
+                @setEvalBranchQuota(10_000);
+                templates[i] = v8.Persistent(v8.FunctionTemplate).init(isolate, generateClass(@field(types, s.name), isolate)).castToFunctionTemplate();
```
karlseguin, line 228: Are you not freeing these in `deinit` because you figure the isolate is going away anyways?

src/runtime/js.zig
```diff
@@ -305,20 +301,25 @@ pub fn Env(comptime S: type, comptime types: anytype) type {
...
+                    const isolate = scope.isolate;
                     self.endScope();
+
+                    // V8 doesn't immediately free memory associated with
+                    // a Context, it's managed by the garbage collector. So, when the
+                    // `gc_hints` option is enabled, we'll use the `lowMemoryNotification`
+                    // call on the isolate to encourage v8 to free any contexts which
+                    // have been freed.
+                    if (self.env.gc_hints) {
+                        var handle_scope: v8.HandleScope = undefined;
```
karlseguin, line 314: We have to create a HandleScope to free memory?

src/runtime/js.zig
sjorsdonkers: Correct

src/runtime/js.zig
sjorsdonkers: Apparently it crashes otherwise with the HandleScope message.  Anything that creates a local requires a scope to be active. Before this was leaking the object into the global scope.

PR #583, opened 2025-04-29
Share underlying DOM global Window with the isolated World

sjorsdonkers:
- Only the main world page holds a HandleScope and an opened Context
- Isolated Worlds do not hold a HandleScope and the Context is closed by default
- An isolatedWorld context can only be made when there is a page

- v8.context no longer live on the scope allowing the isolated world to be constructed after the page creation.

Note, we should probably:
- rename Executor to World
- rename Scope to Context
- store the HandleScope in the SessionState
- rename the createIsolatedWorld function to something similar to createPage, since the world already exists, adoptPage?, embracePage, assimilatePage? navigatePage, reset/reinitwithPage

[Feel free to merge if approved]

### DIFF
--- a/src/runtime/js.zig
+++ b/src/runtime/js.zig
@@ -345,8 +345,7 @@ pub fn Env(comptime S: type, comptime types: anytype) type {
                 const isolate = env.isolate;
                 const Global = @TypeOf(global.*);
 
-                var context: v8.Context = undefined;
-                {
+                var context: v8.Context = blk: {
                     var handle_scope: v8.HandleScope = undefined;
                     v8.HandleScope.init(&handle_scope, isolate);
                     defer handle_scope.deinit();
@@ -379,7 +378,7 @@ pub fn Env(comptime S: type, comptime types: anytype) type {
                     }
 
                     const context_local = v8.Context.init(isolate, global_template, null);
-                    context = v8.Persistent(v8.Context).init(isolate, context_local).castToContext();
+                    const context = v8.Persistent(v8.Context).init(isolate, context_local).castToContext();
                     context.enter();
                     errdefer if (enter) context.exit();
                     defer if (!enter) context.exit();
@@ -404,8 +403,12 @@ pub fn Env(comptime S: type, comptime types: anytype) type {
                             _ = self_obj.setPrototype(context, proto_obj);
                         }
                     }
-                }
+                    break :blk context;
+                };
 
+                // For a Page we only create one HandleScope, it is stored in the main World (enter==true). A page can have multple contexts, 1 for each World.
+                // The main Context/Scope that enters and holds the HandleScope should therefore always be created first. Following other worlds for this page
+                // like isolated Worlds, will thereby place their objects on the main page's HandleScope. Note: In the furure the number of context will multiply multiple frames support
                 var handle_scope: ?v8.HandleScope = null;
                 if (enter) {
                     handle_scope = @as(v8.HandleScope, undefined);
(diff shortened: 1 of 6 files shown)


src/runtime/js.zig
```diff
@@ -345,62 +345,74 @@ pub fn Env(comptime S: type, comptime types: anytype) type {
...
-
-                // All the FunctionTemplates that we created and setup in Env.init
-                // are now going to get associated with our global instance.
-                const templates = &self.env.templates;
-                inline for (Types, 0..) |s, i| {
-                    const Struct = @field(types, s.name);
-                    const class_name = v8.String.initUtf8(isolate, comptime classNameForStruct(Struct));
-                    global_template.set(class_name.toName(), templates[i], v8.PropertyAttribute.None);
-                }
+                var context: v8.Context = undefined;
```
karlseguin: Another option is to assign the context to the block, turning the `var` -> `const` and removing the `undefined`. 

```zig
const context = blk: {
   // ...
    break :blk context;
};
```

src/runtime/js.zig
```diff
@@ -345,62 +345,74 @@ pub fn Env(comptime S: type, comptime types: anytype) type {
...
-
-                        const proto_index = @field(TYPE_LOOKUP, proto_name).index;
-                        const proto_obj = templates[proto_index].getFunction(context).toObject();
-
-                        const self_obj = templates[i].getFunction(context).toObject();
-                        _ = self_obj.setPrototype(context, proto_obj);
                     }
                 }
 
+                var handle_scope: ?v8.HandleScope = null;
```
karlseguin, line 412: How is the JavaScript to set the custom exception (about 30 lines after this) able to run if there's no scope?

src/runtime/js.zig
sjorsdonkers: Good to know, improved here: [83ef21e](https://github.com/lightpanda-io/browser/pull/583/commits/83ef21e699229e47728582ddf769280bc8ae1e15)

src/runtime/js.zig
sjorsdonkers: Added explanation here: [83ef21e](https://github.com/lightpanda-io/browser/pull/583/commits/83ef21e699229e47728582ddf769280bc8ae1e15)

Basically all context are bound to 1 Page. The number of context we have (without multiple frame support) is 1 x #worlds.
See: https://drive.google.com/file/d/0B1obCOyvTnPKSERSMmpRVjVKQWc/view?resourcekey=0-uq2u_FEcWbEtoI3REw3C_w

So, everytime we removePage we should also remove the contexts in the other world, everytime we createPage we should also Create the context in the other worlds.

Currently we do not do that, and basically rely on the main page creating the new data at the same location.
We should fix that in another PR.

sjorsdonkers: So to actually answer the question, if enter is false, then this is not the main World, so the Main Worlds scope already exists. Note that createIsolatedWorld now takes the main world Page as parameter.

sjorsdonkers: We could make a nested scope to clean up any locals created when making the persistent and string in the lower part of the function. Now it is inconsistent where we properly scope the first half, but leave the latter half on the Page's scope. Probably fine to just leave everything on the Page's scope.. perhaps a bit more memory.
Additionally the cleanup isn't and wasn't great in error cases.

karlseguin: Are you saying that this works because the page's handle_scope exists at this point?

Isn't it more correct to have proper handle_scope for the execution of this little snippet, and whatever other JS-based polyfills we need to do?

karlseguin: Ack, this remind me that none of the polyfils exist in the isolated world. Not specific to this PR, but might need fixing separately.

https://github.com/lightpanda-io/browser/blob/main/src/browser/browser.zig#L279

sjorsdonkers: I'm aware (polyfills), adding it here:
https://github.com/lightpanda-io/browser/pull/584

PR #584, opened 2025-04-30
Intersection observer api

sjorsdonkers:
IntersectionObserver implementation using polyfill.
Additional apis needed to make this work.

This turns out to be a good test to make sure the IsolatedWorld works correctly.
A Zig native implementation is likely preferable in the future.

### DIFF
--- a/src/browser/env.zig
+++ b/src/browser/env.zig
@@ -19,7 +19,14 @@ const Interfaces = generate.Tuple(.{
     @import("url/url.zig").Interfaces,
     @import("xhr/xhr.zig").Interfaces,
     @import("xmlserializer/xmlserializer.zig").Interfaces,
+    CSSStyleDeclaration,
 });
+
+pub const CSSStyleDeclaration = struct {
+    pub fn get_display(_: *CSSStyleDeclaration) []const u8 {
+        return "anythingbutnotnone";
+    }
+};
 
 pub const JsThis = Env.JsThis;
 pub const JsObject = Env.JsObject;
--- a/src/browser/polyfill/intersection_observer.js
+++ b/src/browser/polyfill/intersection_observer.js
@@ -50,15 +50,15 @@
   /**
    * A local reference to the root document.
    */
-  var document = (function(startDoc) {
-    var doc = startDoc;
-    var frame = getFrameElement(doc);
-    while (frame) {
-      doc = frame.ownerDocument;
-      frame = getFrameElement(doc);
-    }
-    return doc;
-  })(window.document);
+  // var document = (function(startDoc) {
+  //   var doc = startDoc;
+  //   var frame = getFrameElement(doc);
+  //   while (frame) {
+  //     doc = frame.ownerDocument;
+  //     frame = getFrameElement(doc);
+  //   }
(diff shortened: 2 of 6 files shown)


src/browser/polyfill/intersection_observer.js
```diff
@@ -0,0 +1,1018 @@
...
+  
+  
+  /**
+   * Returns the root rect after being expanded by the rootMargin value.
+   * @return {ClientRect} The expanded root rect.
+   * @private
+   */
+  IntersectionObserver.prototype._getRootRect = function() {
+    console.log('window.document: '+ window.document);
+    console.log('document: '+ document);
```
sjorsdonkers: `document == null` 🧐

sjorsdonkers: `click.js` Executes the following in the isolated world:
```js
const visibleRatio = await new Promise(resolve => {
    const observer = new IntersectionObserver(entries => {
        resolve(entries[0].intersectionRatio);
        observer.disconnect();
    });
    observer.observe(element);
});
return threshold === 1 ? visibleRatio === 1 : visibleRatio > threshold;
```
Note that `opt_options` is not passed so `this.root` is null meaning `document` should be used as `root`

karlseguin: Is this a lot of work to add natively to Zig?

src/browser/env.zig
```diff
@@ -19,8 +19,15 @@ const Interfaces = generate.Tuple(.{
     @import("url/url.zig").Interfaces,
     @import("xhr/xhr.zig").Interfaces,
     @import("xmlserializer/xmlserializer.zig").Interfaces,
+    CSSStyleDeclaration,
 });
 
+pub const CSSStyleDeclaration = struct {
+    pub fn get_display(_: *CSSStyleDeclaration) []const u8 {
+        return "anythingbutnotnone";
```
karlseguin, line 27: `unset` is probably better (it has a real meaning here)

src/browser/env.zig
```diff
@@ -19,8 +19,15 @@ const Interfaces = generate.Tuple(.{
     @import("url/url.zig").Interfaces,
     @import("xhr/xhr.zig").Interfaces,
     @import("xmlserializer/xmlserializer.zig").Interfaces,
+    CSSStyleDeclaration,
 });
 
+pub const CSSStyleDeclaration = struct {
```
karlseguin, line 25: Maybe moved to the `css` folder?

src/browser/polyfill/intersection_observer.js
sjorsdonkers: I mostly wanted to learn and understand the polyfills, then solve the core issues.
With the current code it runs, but returns false instead of true for the click example.
So I think that is the right time to switch to a native Zig implementation.
I'll make a new PR to at least add loading the polyfill and make issues for the isolated page recreation and delay loading polyfills issues.

src/browser/env.zig
sjorsdonkers: you're too fast lol, I'm closing this PR

PR #585, opened 2025-04-30
Support union parameters

karlseguin:
There's ambiguity in mapping due to the flexible nature of JavaScript. Hopefully most types are unambiguous, like a string or a `*parser.Node`.

We need to "probe" each field to see if it's a possible candidate for the JS value. On a perfect match, we stop probing and set the appropriate union field. There are 2 levels of possible matches: candidate and coerce. A "candidate" match has higher precedence. This is necessary because, in JavaScript, a lot of things can be coerced to a lot of other, seemingly wrong, things.

For example, say we have this union:

a: i32,
b: bool,

Field `a` is a perfect match for the value 123. And field b is a coerce match (because, yes, 123 can be coerced to a boolean). So we map it to `a`.

Field `a` is a candidate match for the value 34.2, because float -> int are both "Numbers" in JavaScript. And field b is a coerce match. So we map it to `a`.

Both field `a` and field `b` are coerce matches for "hello". So we map it to `a` because it's declared first (this relies on how Zig currently works, but I don't think the ordering of type declarations is guaranteed, so that's an issue).

### DIFF
--- a/src/runtime/js.zig
+++ b/src/runtime/js.zig
@@ -2075,7 +2075,7 @@ fn Caller(comptime E: type) type {
         fn jsValueToZig(self: *const Self, comptime named_function: anytype, comptime T: type, js_value: v8.Value) !T {
             switch (@typeInfo(T)) {
                 .optional => |o| {
-                    if (js_value.isNull() or js_value.isUndefined()) {
+                    if (js_value.isNullOrUndefined()) {
                         return null;
                     }
                     return try self.jsValueToZig(named_function, o.child, js_value);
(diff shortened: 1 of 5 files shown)


src/runtime/js.zig
```diff
@@ -2237,6 +2226,230 @@ fn Caller(comptime E: type) type {
...
+
+        // Probing is part of trying to map a JS value to a Zig union. There's
+        // a lot of ambiguity in this process, in part because some JS values
+        // can almost always be coerced. For example, anything can be coerced
+        // into an integer (it just becomes 0), or a float (becomes NaN) or a
+        // string.
+        //
+        // The way we'll do this is that, if there's a direct match, we'll use it
+        // If there's a potential match, we'll keep looking for a direct match
+        // and only use the (first) potential match as a fallback.
```
sjorsdonkers, line 2352: Just the first one as defined in the order of the union variants? No, type based or like an integer is rather a float than a string. I would expect JS to be highly unpredictable instead of something as straightforward as just the first. Dunno :)
I think this is going to be very hard to get exactly right in the first implementation

sjorsdonkers (approved):
I'm not sure if it is needed to do a deep dive review.
I think we should consider moving specifically the jsValueToZig and vice-verse chain of functions to a separate file.

src/runtime/js.zig
karlseguin: Order is used as a tie-breaker only when no exact match is available.

if you have a union:

```
a: i32,
b: bool
c: f32
```

And you pass a `true`, then it'll map to `b`.   If you pass a 1.445, it'll map to `c`. But if you pass "hello" which, in JavaScript, can safely be coerced to an integer, a boolean and a float, it'll pick `a`.

krichprollsch (approved):

PR #587, opened 2025-04-30
cdp: dispatch DOM.setChildNodes event for search results

krichprollsch:


src/cdp/domains/dom.zig
```diff
@@ -76,6 +76,23 @@ fn performSearch(cmd: anytype) !void {
...
+    // dispatch setChildNodesEvents to inform the client of the subpart of node
+    // tree covering the results.
+    for (list.nodes.items) |n| {
+        // retrieve the node's parent
+        if (try parser.nodeParentNode(n)) |p| {
+            // Register the parent and send the node.
+            const parent_node = try bc.node_registry.register(p);
+            const node = bc.node_registry.lookup_by_node.get(n) orelse unreachable;
+            // Should-we return one DOM.setChildNodes event per parentId
+            // containing all its children in the nodes array?
```
karlseguin: Yes.

If it's something we'll need to do often, we could expand the `NodeWriter`. It already takes a placeholder option, so it could be something like:

```
.nodes = bc.nodeWriter(parent_node, .{.mode = .child_list})},
```

krichprollsch:
We have to dispatch the `setChildNodes` event for the whole hierarchy, starting from the root until the nodes.
I tried an implementation but I have a leak error with the unit tests.

karlseguin:
> We have to dispatch the `setChildNodes` event for the whole hierarchy, starting from the root until the nodes. I tried an implementation but I have a leak error with the unit tests.

Really? I was pretty sure it was just the node's sibling, or, maybe more accurately, the node's parent children.

I don't see a leak, both in the CI and locally, it's an `unreachable` which is being hit. 

```zig
const parent_node = bc.node_registry.lookup_by_node.get(p) orelse unreachable;
```

I don't think there's any reason to assume the parent has been seen by the registry before. Always better to use `register`, which does a `getOrPut`

```zig
const node = try bc.node_registry.register(parser.documentToNode(doc));
```

After that, you'll get a crash because the session_id isn't set. I would do 2 things here:

1 - 
Be defensive and add a check after loading `bc.

```zig
var bc = ...
const session_id = bc.session_id orelse return error.SessionIdNotLoaded;
```

2 - Change the cdp/testing.zig `loadBrowserContext` and set a default session_id if there's `html`:

```zig
// existing if statement
if (opts.html) |html| {
   // add
   if (bc.session_id == null) bc.session_id = "SID-X";
   ..
}
```

krichprollsch:
The issue is fixed. I added a bool into `cdp.Node` to know if a node has been already dispatched.

> Really? I was pretty sure it was just the node's sibling, or, maybe more accurately, the node's parent children.

Yes it is as far as I can see. ..

src/cdp/domains/dom.zig
```diff
@@ -76,12 +76,69 @@ fn performSearch(cmd: anytype) !void {
...
+    }
+
+    const plen = parents.items.len;
+    if (plen == 0) return;
+
+    var i: usize = plen;
+    while (i > 0) {
+        i -= 1;
+        const node = parents.items[i];
+        if (node.set_child_nodes_event) continue;
```
karlseguin, line 124:
```suggestion
        // Although our above loop won't add an already-sent node to `parents`
        // this can still be true because two nodes can share the same parent node
        // so we might have just sent the node a previous iteration of this loop
        if (node.set_child_nodes_event) continue;
```

src/cdp/domains/dom.zig
```diff
@@ -76,12 +76,69 @@ fn performSearch(cmd: anytype) !void {
...
+            try parents.append(arena, node);
+            n = p;
+        }
+    }
+
+    const plen = parents.items.len;
+    if (plen == 0) return;
+
+    var i: usize = plen;
+    while (i > 0) {
```
karlseguin, line 118:
```suggestion
    // We're going to iterate in reverse order from how we added them.
    // This ensures that we're emitting the tree of nodes top-down.
    while (i > 0) {
```

PR #592, opened 2025-05-02
Isolated polyfill & create world when needed

sjorsdonkers:
Polyfills are now loaded into the isolated world
The isolated world is created only when needed and removed and created along with the main page if it exists.

### DIFF
--- a/src/cdp/cdp.zig
+++ b/src/cdp/cdp.zig
@@ -377,7 +377,7 @@ pub fn BrowserContext(comptime CDP_T: type) type {
 
             self.isolated_world = .{
                 .name = try self.arena.dupe(u8, world_name),
-                .scope = undefined,
+                .scope = null,
                 .executor = executor,
                 .grant_universal_access = grant_universal_access,
             };
@@ -519,7 +519,8 @@ const IsolatedWorld = struct {
         self.executor.deinit();
         self.scope = null;
     }
-    pub fn removeContext(self: *IsolatedWorld) void {
+    pub fn removeContext(self: *IsolatedWorld) !void {
+        if (self.scope == null) return error.NoIsolatedContextToRemove;
         self.executor.endScope();
         self.scope = null;
     }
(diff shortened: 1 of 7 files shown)


src/cdp/cdp.zig
```diff
@@ -506,12 +511,26 @@ pub fn BrowserContext(comptime CDP_T: type) type {
...
+        self.scope = null;
+    }
+
+    // The isolate world must share at least some of the state with the related page, specifically the DocumentHTML
+    // (assuming grantUniveralAccess will be set to True!).
+    // We just created the world and the page. The page's state lives in the session, but is update on navigation.
+    // This also means this pointer becomes invalid after removePage untill a new page is created.
+    // Currently we have only 1 page/frame and thus also only 1 state in the isolate world.
+    pub fn createContext(self: *IsolatedWorld, page: *Page) !void {
+        self.scope = try self.executor.startScope(&page.window, &page.state, {}, false);
```
karlseguin, line 535: Do you think there's a need to check if `self.scope != null` here and, if so, cleanup?

src/cdp/domains/page.zig
```diff
@@ -112,15 +113,17 @@ fn createIsolatedWorld(cmd: anytype) !void {
     }
     const bc = cmd.browser_context orelse return error.BrowserContextNotLoaded;
 
-    const world = &bc.isolated_world.?;
-    world.name = try bc.arena.dupe(u8, params.worldName);
-    world.grant_universal_access = params.grantUniveralAccess;
+    const world = try bc.createIsolatedWorld(params.worldName, params.grantUniveralAccess);
+    const page = bc.session.currentPage().?;
```
karlseguin: Who knows if there's a client out there that tries to do this before setting up the target/page:

```zig
const page = bc.session.currentPage() orelse return error.PageNotLoaded;
```

sjorsdonkers: Fixed: [9f72c98](https://github.com/lightpanda-io/browser/pull/592/commits/9f72c9896766633598c13ee5bc9a80b84b9efa00)

src/cdp/cdp.zig
sjorsdonkers: If scope != null here it would be a bug most likely ours. Instead of cleanup I changed it to return an error. [9f72c98](https://github.com/lightpanda-io/browser/pull/592/commits/9f72c9896766633598c13ee5bc9a80b84b9efa00)

PR #595, opened 2025-05-03
Improve the debug ergonomics of the Env generic.

karlseguin:
Previously, we were passing our WebAPIs directly as an anonymous tuple. This resulted in Env(T) having an _awful_ name - a name composed of hundreds of classes.

By wrapping the anonymous tuple into a normal struct, the Env now gets a sane name which helps improve stack traces (and profiling, and debugging, ...)


krichprollsch (approved):

krichprollsch:
@karlseguin LGTM, please could you rebase the PR and merge it?
Thanks

Issue #412, opened 2025-02-08
`UnsupportedUriScheme` on DataUri as `script` tag `src`


smashah:
Some websites have a dataurl as a `src` in the script tag.

e.g when running this command:

![Image](https://github.com/user-attachments/assets/bef9069f-5398-47da-809d-31b848773eb6)

The site returns these script tags:

```html
<script src="data:application/x-javascript; charset=utf-8;base64,Oy8qRkJfUEtHX0RFTElNKi8KCnZhciBkYXRhRWxlbWVudD1kb2N1bWVudC5nZXRFbGVtZW50QnlJZCgiZW52anNvbiIpO2lmKGRhdGFFbGVtZW50IT1udWxsKXt2YXIgY29weVZhcmlhYmxlcz1mdW5jdGlvbihhKXtmb3IodmFyIGIgaW4gdmFyaWFibGVzKWFbYl09dmFyaWFibGVzW2JdfSx2YXJpYWJsZXM9SlNPTi5wYXJzZShkYXRhRWxlbWVudC50ZXh0Q29udGVudCk7d2luZG93LnJlcXVpcmVMYXp5P3dpbmRvdy5yZXF1aXJlTGF6eShbIkVudiJdLGNvcHlWYXJpYWJsZXMpOih3aW5kb3cuRW52PXdpbmRvdy5FbnZ8fHt9LGNvcHlWYXJpYWJsZXMod2luZG93LkVudikpfQovLyMgc291cmNlVVJMPWh0dHBzOi8vc3RhdGljLndoYXRzYXBwLm5ldC9yc3JjLnBocC92NC95TC9yL2RiZ3JNc25EMThwLmpzCg==" data-bootloader-hash="nxRKQO6" crossorigin="anonymous" data-p=":0" data-btmanifest="1019961576_main" data-c="1"></script>
[...]

PR #596, opened 2025-05-03
Support Data URI in scripts tags

ksdme:
Related to https://github.com/lightpanda-io/browser/issues/412

Hey guys,

I was trying to play around with Zig and found this project via a tweet from Andreas Kling. I took a shot at the issue linked above (it looks simple and reproducible). I apologize for any stupidity in the solution, like I mentioned, I am new to Zig.

For some context on the TODOs,
I tried fixing up `Mime.parse` to use a `[]const u8` instead so I could use it during data URI parsing. But that required either turning all return values into owned values or moving up the normalized string into the struct. Given that I could find how deinit was being called in `fetchData`, I just left this for later.

I tested the solution with the following,
```
[I] ksdme@bunny ~/m/b/example (ksdme/data-uri-script)> cat srctag.html
<html>
  <head>
    <script src="data:application/x-javascript; charset=utf-8;base64,Y29uc29sZS5sb2coImhlbGxvIik7"></script>
  </head>
</html>
[I] ksdme@bunny ~/m/b/example (ksdme/data-uri-script)>
```

```
[I] ksdme@bunny ~/m/browser (ksdme/data-uri-script)> make run
Building (debug)...
Build OK
Running...
info(server): accepting new conn...
info(server): client connected
info(http_client): redirecting to: GET http://localhost:8000/srctag
info(browser): GET http://localhost:8000/srctag 200
info(console): hello
```


ksdme:

src/browser/browser.zig
```diff
@@ -581,10 +582,18 @@ pub const Page = struct {
     // It resolves src using the page's uri.
     // If a base path is given, src is resolved according to the base first.
     // the caller owns the returned string
+    // TODO: We are leaking the memory associated with the response.
```
karlseguin: We leverage arena allocators a lot, hence the name of the variable is `arena`. It's cleared whenever a page is created in `createPage`. So it isn't leaking, it's just [probably] living for longer than it absolutely has to.

src/browser/browser.zig
```diff
@@ -581,10 +582,18 @@ pub const Page = struct {
...
     fn fetchData(self: *const Page, src: []const u8, base: ?[]const u8) ![]const u8 {
         log.debug("starting fetch {s}", .{src});
 
         const arena = self.arena;
+
+        // Handle data URIs.
+        const data_uri: ?DataURI = DataURI.parse(arena, src) catch null;
```
karlseguin: I would make `parse` return an optional directly, `!?[]const u8`.

To me, the main benefit is that it gives you an easy way to differentiate between it being an invalid data URI and some other error. Like, as-is, if `parse` throws an OutOfMemory error, the catch just tries to move on. Not the end of the world, but also not particularly explicit.

If you stick `error.Invalid`, then this code might be better as:

```zig
const data_uri: ?DataURI = DataURI.parse(arena, src) catch |err| switch (err) blk: {
     error.Invalid => break :blk null,
     else => return err,
};
```

Which is a bit wordy, and why I'd prefer the optional:

```zig
if (try DataURI.parse(arena, src)) |data_uri| {
  return data_uri.data;
}
```

src/browser/datauri.zig
```diff
@@ -0,0 +1,79 @@
...
+    data: []const u8,
+
+    // Parses data:[<media-type>][;base64],<data>
+    pub fn parse(allocator: Allocator, src: []const u8) !DataURI {
+        if (!std.mem.startsWith(u8, src, "data:")) {
+            return error.Invalid;
+        }
+
+        const uri = src[5..];
+        const data_starts = std.mem.indexOf(u8, uri, ",") orelse return error.Invalid;
```
karlseguin: Can use `std.mem.indexOfScalar` directly.  `std.mem.indexOf` _will_ call it, but only after doing 2 length checks on the needle.

src/browser/datauri.zig
```diff
@@ -0,0 +1,79 @@
...
+            data = buffer;
+        }
+
+        return .{
+            .was_base64_encoded = base64_encoded,
+            .data = data,
+        };
+    }
+
+    pub fn deinit(self: *const DataURI, allocator: Allocator) void {
```
karlseguin, line 51: Up to you whether you want to keep it or not (given what I commented on above about the use of an arena). It _isnt_ going to be called, again, because of the arena, but, for completeness, I'm ok with keeping it in. Maybe one day we'll want to call this wit a different type of allocator, so having this is a good reminder that, yes, normally, we should `deinit` it.

karlseguin (commented):
Thanks for the PR. It looks great. A few inline comments, but just a few very minor things.

src/browser/browser.zig
ksdme: Makes sense

src/browser/browser.zig
ksdme: This is how I had it earlier, I think I removed it because it felt like returning null prevents you from calling deinit and clean up and temporary allocations.

Fixed up though.

src/browser/datauri.zig
ksdme: I think it makes sense to keep it. I like it because it sort of documents the allocations in an expected manner.

ksdme:
@karlseguin Fixed up everything.

karlseguin:
Great, again, thanks for the PR! It really helps.

PR #598, opened 2025-05-05
remove unused code, mostly imports

karlseguin:


krichprollsch:
Did you use a tool to cleanup the code?

krichprollsch (approved):

karlseguin:
> Did you use a tool to cleanup the code?

https://github.com/DonIsaac/zlint 

Could be integrated in the CI, but it reported the same 2 errors hundreds of times, and I don't really agree. I guess it can be configured. It has a beautiful log output, I'd love to use that log output in browser!

The first warning was it wants a comment whenever `undefined` is used. I guess after https://github.com/lightpanda-io/browser/pull/597 I should think that's a good idea, but we use `undefined` a lot.

The second was to favor implicit type instead of `@as`:

```zig
const node_base = @as(NodeExternal, @ptrCast(node_aligned));
```

should be

```zig
const node_base: NodeExternal = @ptrCast(node_aligned);
```

PR #599, opened 2025-05-05
Native IntersectionObserver

sjorsdonkers:
Implements:
- IntersectionObserver
- IntersectionObserverEntry
- Element.getClientRects

Changed the client's width +1 to make sure all elements are inside. Should we instead start counting from 0?

Notes:
1. We return [DOMRect](https://developer.mozilla.org/en-US/docs/Web/API/DOMRect) instead of its parent class DOMRectReadOnly.
1. We call the callback immediately, we do not wait for any event that could cause an intersection change, like scrolling, resizing, DOM modification.
1. We always indicate the root and target element are fully intersecting

### DIFF
--- a/src/browser/browser.zig
+++ b/src/browser/browser.zig
@@ -829,7 +829,7 @@ const FlatRenderer = struct {
     }
 
     pub fn width(self: *const FlatRenderer) u32 {
-        return @intCast(self.elements.items.len);
+        return @intCast(self.elements.items.len + 1); // +1 since x starts at 1 (use len after append)
     }
 
     pub fn height(_: *const FlatRenderer) u32 {
--- a/src/browser/browser.zig
+++ b/src/browser/browser.zig
@@ -828,6 +828,15 @@ const FlatRenderer = struct {
         };
     }
 
+    pub fn boundingRect(self: *const FlatRenderer) Element.DOMRect {
+        return .{
+            .x = 0.0,
+            .y = 0.0,
+            .width = @floatFromInt(self.width()),
+            .height = @floatFromInt(self.height()),
+        };
+    }
+
     pub fn width(self: *const FlatRenderer) u32 {
         return @intCast(self.elements.items.len + 1); // +1 since x starts at 1 (use len after append)
     }
--- a/src/browser/dom/element.zig
+++ b/src/browser/dom/element.zig
@@ -348,10 +348,8 @@ pub const Element = struct {
 
     // returns a collection of DOMRect objects that indicate the bounding rectangles for each CSS border box in a client.
     // We do not render so just always return the element's rect.
-    pub fn _getClientRects(self: *parser.Element, state: *SessionState) ![]DOMRect {
-        var heap = try state.arena.create(DOMRect);
-        heap.* = try state.renderer.getRect(self);
-        return heap[0..1];
+    pub fn _getClientRects(self: *parser.Element, state: *SessionState) ![1]DOMRect {
+        return [_]DOMRect{try state.renderer.getRect(self)};
     }
 
     pub fn get_clientWidth(_: *parser.Element, state: *SessionState) u32 {
--- a/src/browser/dom/element.zig
+++ b/src/browser/dom/element.zig
@@ -346,6 +346,14 @@ pub const Element = struct {
         return state.renderer.getRect(self);
     }
 
+    // returns a collection of DOMRect objects that indicate the bounding rectangles for each CSS border box in a client.
+    // We do not render so just always return the element's rect.
+    pub fn _getClientRects(self: *parser.Element, state: *SessionState) ![]DOMRect {
+        var heap = try state.arena.create(DOMRect);
+        heap.* = try state.renderer.getRect(self);
+        return heap[0..1];
+    }
+
     pub fn get_clientWidth(_: *parser.Element, state: *SessionState) u32 {
         return state.renderer.width();
     }
--- a/src/browser/dom/element.zig
+++ b/src/browser/dom/element.zig
@@ -506,7 +506,7 @@ test "Browser.DOM.Element" {
     }, .{});
 
     try runner.testCases(&.{
-        .{ "document.getElementById('para').clientWidth", "0" },
+        .{ "document.getElementById('para').clientWidth", "1" },
         .{ "document.getElementById('para').clientHeight", "1" },
 
         .{ "let r1 = document.getElementById('para').getBoundingClientRect()", "undefined" },
@@ -527,7 +527,7 @@ test "Browser.DOM.Element" {
         .{ "r3.width", "1" },
         .{ "r3.height", "1" },
 
-        .{ "document.getElementById('para').clientWidth", "2" },
+        .{ "document.getElementById('para').clientWidth", "3" },
         .{ "document.getElementById('para').clientHeight", "1" },
     }, .{});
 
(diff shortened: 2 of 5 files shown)


src/browser/browser.zig
```diff
@@ -834,7 +834,7 @@ const FlatRenderer = struct {
     }
 
     pub fn width(self: *const FlatRenderer) u32 {
-        return @intCast(self.elements.items.len);
+        return @intCast(self.elements.items.len + 1); // +1 since x starts at 1 (use len after append)
```
sjorsdonkers, line 841: We can also make x start at 0 instead, this is needed to make sure all elements are withing the width

src/browser/dom/intersection_observer.zig
```diff
@@ -0,0 +1,257 @@
...
+
+    observed_entries: std.ArrayListUnmanaged(IntersectionObserverEntry),
+
+    // new IntersectionObserver(callback)
+    // new IntersectionObserver(callback, options) [not supported yet]
+    pub fn constructor(callback: Env.Callback, options_: ?IntersectionObserverOptions, state: *SessionState) !IntersectionObserver {
+        var options = IntersectionObserverOptions{
+            .root = parser.documentToNode(parser.documentHTMLToDocument(state.document.?)),
+            .rootMargin = "0px 0px 0px 0px",
+            .threshold = &default_threshold,
```
sjorsdonkers: Is there a safe way to make an array in static memory that does not accidentally end up on the stack?

src/browser/dom/intersection_observer.zig
```diff
@@ -0,0 +1,257 @@
...
+        .{ "let entry;", "undefined" },
+        .{ "new IntersectionObserver(entries => { entry = entries[0]; }).observe(document.createElement('div'));", "undefined" },
+        .{ "entry.boundingClientRect.x;", "1" },
+        .{ "entry.intersectionRatio;", "1" },
+        .{ "entry.intersectionRect.x;", "1" },
+        .{ "entry.intersectionRect.y;", "0" },
+        .{ "entry.intersectionRect.width;", "1" },
+        .{ "entry.intersectionRect.height;", "1" },
+        .{ "entry.isIntersecting;", "true" },
+        .{ "entry.rootBounds.x;", "2" }, // This is not the prefered behaviour, the Window rect should wrap all elements so x -> 0
```
sjorsdonkers: Should we make an exception in `renderer.getRect` to detect if the pointer is the Window/document in which case we return a `clientWidth` sized Rect?

src/browser/dom/element.zig
```diff
@@ -348,6 +348,14 @@ pub const Element = struct {
         return state.renderer.getRect(self);
     }
 
+    // returns a collection of DOMRect objects that indicate the bounding rectangles for each CSS border box in a client.
+    // We do not render so just always return the element's rect.
+    pub fn _getClientRects(self: *parser.Element, state: *SessionState) ![]DOMRect {
+        var heap = try state.arena.create(DOMRect);
```
karlseguin: You can return an array, avoiding the heap allocation:

```zig
pub fn _getClientRects(self: *parser.Element, state: *SessionState) ![1]DOMRect {
  return [_]DOMRect{try state.renderer.getRect(self)};
}
```

src/browser/dom/intersection_observer.zig
karlseguin: Not sure if this is what you mean, but I'm like..74% sure you can do:

```zig
.threshold = &.{0.0},
```

And it'll be safe, because you're taking the address of a const value, which'll be in the data section of the binary.

src/browser/dom/intersection_observer.zig
```diff
@@ -0,0 +1,257 @@
...
+            .state = state,
+            .observed_entries = .{},
+        };
+    }
+
+    pub fn _disconnect(self: *IntersectionObserver) !void {
+        self.observed_entries = .{}; // We don't free as it is on an arena
+    }
+
+    pub fn _observe(self: *IntersectionObserver, targetElement: *parser.Element) !void {
```
karlseguin:
```suggestion
    pub fn _observe(self: *IntersectionObserver, target_element: *parser.Element) !void {
```

src/browser/dom/intersection_observer.zig
```diff
@@ -0,0 +1,257 @@
...
+        try self.observed_entries.append(self.state.arena, .{
```
karlseguin, line 84: A minor issue, but I think it's supposed to be a no-op on duplicates.

src/browser/dom/intersection_observer.zig
```diff
@@ -0,0 +1,257 @@
...
+// https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserverEntry
+// https://w3c.github.io/IntersectionObserver/#intersection-observer-entry
+pub const IntersectionObserverEntry = struct {
+    state: *SessionState,
+    target: *parser.Element,
+    options: *IntersectionObserverOptions,
+
+    // Returns the bounds rectangle of the target element as a DOMRectReadOnly. The bounds are computed as described in the documentation for Element.getBoundingClientRect().
+    pub fn get_boundingClientRect(self: *const IntersectionObserverEntry) !Element.DOMRect {
+        return self.state.renderer.getRect(self.target); // Does this ever change?
```
karlseguin: Not currently, but if you were thinking of storing the value, I think it's better as-is, in case the renderer implementation changes.

src/browser/dom/intersection_observer.zig
karlseguin: This is not a bad idea.

src/browser/dom/intersection_observer.zig
sjorsdonkers: Fixed [e32d35b](https://github.com/lightpanda-io/browser/pull/599/commits/e32d35b156b28f7ddad0603ca47dd44bc75981ab)

src/browser/dom/intersection_observer.zig
sjorsdonkers: Changed: [e32d35b](https://github.com/lightpanda-io/browser/pull/599/commits/e32d35b156b28f7ddad0603ca47dd44bc75981ab)

src/browser/dom/element.zig
sjorsdonkers: Changed: [e32d35b](https://github.com/lightpanda-io/browser/pull/599/commits/e32d35b156b28f7ddad0603ca47dd44bc75981ab)

src/browser/dom/intersection_observer.zig
sjorsdonkers: Fixed: [e32d35b](https://github.com/lightpanda-io/browser/pull/599/commits/e32d35b156b28f7ddad0603ca47dd44bc75981ab)

src/browser/dom/intersection_observer.zig
sjorsdonkers: Removed comment: [e32d35b](https://github.com/lightpanda-io/browser/pull/599/commits/e32d35b156b28f7ddad0603ca47dd44bc75981ab)

src/browser/dom/intersection_observer.zig
sjorsdonkers: Fixed: [e32d35b](https://github.com/lightpanda-io/browser/pull/599/commits/e32d35b156b28f7ddad0603ca47dd44bc75981ab)

PR #602, opened 2025-05-06
Subpixel mouse events

sjorsdonkers:
Mouse events can be in the middle of a pixel `{x: 1.5, y: 0.5}`
NetSurf Dom or our rendered do not support that and just need the pixel where the event happened so we convert it to an integer when we get the value.


krichprollsch:
What if we set the box size to 2px, will the click be on `x: 1, y: 1` instead of `0.5`?

sjorsdonkers:
> the box size to 2px, will the click be on `x: 1, y: 1` instead of `0.5`?

x and y are numbers (not integers) according to [the spec](https://chromedevtools.github.io/devtools-protocol/tot/Input/#method-dispatchMouseEvent). So I think it is best to just support them. Initially I thought making them floats internally as well and have the renderer deal with the conversion, but the DOM also expects integers, so I think it is best to just convert them immediately.

krichprollsch (approved):

PR #603, opened 2025-05-06
Mouse event fixes

sjorsdonkers:
Several issues prevents MouserPressed commands from working:
- URL buffer of 1kb resulted in OutOfMemory error for puppeteer click.js
- of by 1 errors in renderer's pixel location -> rendered now starts at x=0 instead of x=1
- Delay navigation as otherwise the dom tree was deleted during the event causing a crash

### DIFF
--- a/src/browser/browser.zig
+++ b/src/browser/browser.zig
@@ -268,7 +268,7 @@ pub const Page = struct {
         // load polyfills
         try polyfill.load(self.arena, self.scope);
 
-        _ = try session.browser.app.loop.timeout(1 * std.time.ns_per_ms, &self.microtask_node);
+        // _ = try session.browser.app.loop.timeout(1 * std.time.ns_per_ms, &self.microtask_node);
     }
 
     fn microtaskCallback(node: *Loop.CallbackNode, repeat_delay: *?u63) void {
--- a/src/browser/browser.zig
+++ b/src/browser/browser.zig
@@ -193,7 +193,7 @@ pub const Session = struct {
         // can't use the page arena, because we're about to reset it
         // and don't want to use the session's arena, because that'll start to
         // look like a leak if we navigate from page to page a lot.
-        var buf: [4096]u8 = undefined;
+        var buf: [2048]u8 = undefined;
         var fba = std.heap.FixedBufferAllocator.init(&buf);
         const url = try self.page.?.url.resolve(fba.allocator(), url_string);
 
@@ -732,7 +732,7 @@ pub const Page = struct {
             _ = repeat_delay;
             const self: *DelayedNavigation = @fieldParentPtr("navigate_node", node);
             self.session.pageNavigate(self.href) catch |err| {
-                log.err("Delayed navigation error {}", .{err});
+                log.err("Delayed navigation error {}", .{err}); // TODO: should we trigger a specific event here?
             };
         }
     };
--- a/src/browser/browser.zig
+++ b/src/browser/browser.zig
@@ -706,11 +706,34 @@ pub const Page = struct {
             .a => {
                 const element: *parser.Element = @ptrCast(node);
                 const href = (try parser.elementGetAttribute(element, "href")) orelse return;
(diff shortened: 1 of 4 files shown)


src/browser/browser.zig
```diff
@@ -705,12 +705,35 @@ pub const Page = struct {
             .a => {
                 const element: *parser.Element = @ptrCast(node);
                 const href = (try parser.elementGetAttribute(element, "href")) orelse return;
-                return self.session.pageNavigate(href);
+
+                // We cannot navigate immediately as navigating will delete the DOM tree, which holds this event's node.
+                // As such we schedule the function to be called as soon as possible.
+                // NOTE Using the page.arena assumes that the scheduling loop does use this object after invoking the callback
+                // If that changes we may want to consider storing DelayedNavigation in the session instead.
+                const arena = self.arena;
```
sjorsdonkers, line 714: Can change to session arena or as optional member in the session if someone does not like using the page arena here

src/browser/browser.zig
```diff
@@ -705,12 +705,35 @@ pub const Page = struct {
...
             else => {},
         }
     }
 
+    const DelayedNavigation = struct {
+        navigate_node: Loop.CallbackNode = .{ .func = DelayedNavigation.delay_navigate },
+        session: *Session,
+        href: []const u8,
+
+        fn delay_navigate(node: *Loop.CallbackNode, repeat_delay: *?u63) void {
```
sjorsdonkers, line 731: In Zig, can we instead just define the function and assign it as function pointer directly at line 726 such that no member functions needs to be defined?

src/browser/browser.zig
```diff
@@ -705,12 +705,37 @@ pub const Page = struct {
...
+            _ = repeat_delay;
+            const self: *DelayedNavigation = @fieldParentPtr("navigate_node", node);
+            self.session.pageNavigate(self.href) catch |err| {
+                log.err("Delayed navigation error {}", .{err});
```
sjorsdonkers: Is this sufficient or should we communicate something to client?

src/browser/browser.zig
krichprollsch: idk if it's possible, but I think I prefer this explicit way.

src/browser/browser.zig
krichprollsch: I think this is OK for now.
In the future calling `pageNavigate` will dispatch events to the CDP client.

krichprollsch: Maybe you can add a `TODO` asking if we should trigger a specific event here.

src/browser/browser.zig
```diff
@@ -192,7 +192,7 @@ pub const Session = struct {
         // can't use the page arena, because we're about to reset it
         // and don't want to use the session's arena, because that'll start to
         // look like a leak if we navigate from page to page a lot.
-        var buf: [1024]u8 = undefined;
+        var buf: [4096]u8 = undefined;
```
krichprollsch: I'm not sure about 4ko here. Did you make tests? `resolve` creates a 1ko buffer, then `fromURI` writes the final url next. maybe 2048 is enough?

krichprollsch (approved):

src/browser/browser.zig
sjorsdonkers: Added TODO: [de7de51](https://github.com/lightpanda-io/browser/pull/603/commits/de7de51e7a4ccb588f206fa0037334f793ccfe9e)

src/browser/browser.zig
sjorsdonkers: Nice catch didn't see the internal 1kb.
For now I added a test to document what the maximum url size is and required memory for the resolve function.

Currently the maximum length is about 500 chars. It would be better if we support at least 2000 chars or just have a dynamic arena as the pre-existing comment suggest.
[de7de51](https://github.com/lightpanda-io/browser/pull/603/commits/de7de51e7a4ccb588f206fa0037334f793ccfe9e)

PR #610, opened 2025-05-07
Optimize intervals, and make sure they're probably cleaned up.

karlseguin:
A loop interval will no longer stop the loop from returning from `run`, and no longer requires mutating event_nb on each iteration.

Re-enable microtask loop, which I accidentally stopped in a previous commit.

### DIFF
--- a/src/runtime/loop.zig
+++ b/src/runtime/loop.zig
@@ -41,6 +41,9 @@ pub const Loop = struct {
     // event are finished.
     events_nb: usize,
 
+    // Used to stop repeating timeouts when loop.run is called.
+    stopping: bool,
+
     // ctx_id is incremented each time the loop is reset.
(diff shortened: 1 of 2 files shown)


src/runtime/loop.zig
```diff
@@ -41,6 +41,9 @@ pub const Loop = struct {
     // event are finished.
     events_nb: usize,
 
+    // Used to stop repeating timeouts when loop.run is called.
+    stopping: bool,
```
karlseguin, line 45: @krichprollsch do operations on `events_nb` really need to be synchronized/atomic?  Isn't this all single threaded? 

If I'm wrong, I should add atomic operations around `stopping` too.

krichprollsch: Well I guess it was in case of future multi-threads use...
I'm not sure we should maintain it since many other things would break in this case...

PR #611, opened 2025-05-07
Add Nix Development Support

mookums:
This adds support for development on systems with the Nix Package Manager installed, through the use of a flake. This is also generally a prerequisite for packaging Lightpanda with Nix.


mookums:

PR #623, opened 2025-05-11
add keepalive to http client

karlseguin:
The idle connection pool is currently sub-optimial, as it only re-uses connections on an exact match. Specifically, it would be nice if a connection could be re-used if the host:port is a match, but the current implementation also requires the blocking mode to match.

So we potentially still need 4x more connections, but in reality, it's probably more like 2x. Most people are only going to be dealing with TLS traffic, so when you connect to lightpanda.io, you'll need 1 connection for the blocking calls and 1 connections for the nonblocking calls.

For plaintext request, it would be simple to share connections between blocking and nonblocking. But for encrypted requests, we need to dig more into tls.zig and figure out how/if a `tls_client` can be converted to <-> from  blocking to nonblocking (they're different types).

### DIFF
--- a/src/http/client.zig
+++ b/src/http/client.zig
@@ -306,7 +306,7 @@ pub const Request = struct {
     fn destroyConnection(self: *Request, connection: *Connection) void {
         const client = self._client;
         connection.deinit(client.allocator);
-        errdefer client.connection_pool.destroy(connection);
+        client.connection_pool.destroy(connection);
     }
 
     const AddHeaderOpts = struct {
@@ -2153,6 +2153,7 @@ const IdleConnections = struct {
 
     fn get(self: *IdleConnections, secure: bool, host: []const u8, port: u16, blocking: bool) ?*Connection {
         self.mutex.lock();
+        defer self.mutex.unlock();
 
         var node = self.idle.first;
         while (node) |n| {
@@ -2161,12 +2162,10 @@ const IdleConnections = struct {
                 self.count -= 1;
                 self.idle.remove(n);
                 self.node_pool.destroy(n);
-                self.mutex.unlock();
                 return connection;
             }
             node = n.next;
         }
-        self.mutex.unlock();
         return null;
     }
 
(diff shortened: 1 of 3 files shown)


src/http/client.zig
```diff
@@ -243,66 +409,102 @@ pub const Request = struct {
...
-                    .tls_client = try tls.nb.Client().init(self.arena, .{
-                        .host = self.host(),
-                        .root_ca = self._client.root_ca,
-                        .insecure_skip_verify = self._tls_verify_host == false,
-                        // .key_log_callback = tls.config.key_log.callback
-                    }),
-                },
+        if (self._secure) {
+            connection.tls = .{
+                .nonblocking = try tls.nb.Client().init(self._client.allocator, .{ .host = connection.host, .root_ca = self._client.root_ca, .insecure_skip_verify = self._tls_verify_host == false, .key_log_callback = tls.config.key_log.callback }),
```
mookums: I think you're missing a comma here to get `zig fmt` to wrap it nicely.

mookums:
Looks fine to me.

1. Just a thought but iIt would be nice if we could have some sort of before and after performance figures considering that keep-alive can have a pretty large impact on request/response times.
2. Not sure if it's applicable here but the last time I had to work with a socket that needed to be TLS secured or bare, I ended up just using a vtable that just wraps the socket and provided like some base functionality (like a `Reader` and `Writer`) and all of the magic was opaque to the caller. It helped make a lot of the logic simpler as you just had to focus on the functionality rather than maintaining two implementations side by side.

karlseguin:
The e2e workflow does a perf test, but you don't see much of a difference. The keepalive on localhost without TLS doesn't really add anything. But I can say that anecdotally, it is a noticeable improvement when I'm trying to work on duckduckgo supprot.

For both the synchronous and asynchornous handlers, I'm using a union to deal with the differences between TLS and plaintext. I think a union is better than an interface when the implementations are known at compile time, no? But, for async, it's very different..not just the APIs, but also the flows..with TLS handshake requiring extra round trips

PR #626, opened 2025-05-12
WIP implement TextEncoder

krichprollsch:
Depends on #627

### DIFF
--- a/src/browser/encoding/text_encoder.zig
+++ b/src/browser/encoding/text_encoder.zig
@@ -1,0 +1,64 @@
+// Copyright (C) 2023-2024  Lightpanda (Selecy SAS)
+//
+// Francis Bouvier <francis@lightpanda.io>
+// Pierre Tachoire <pierre@lightpanda.io>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+const std = @import("std");
+
+const Env = @import("../env.zig").Env;
+
+pub const Interfaces = .{
+    TextEncoder,
+};
+
+// https://encoding.spec.whatwg.org/#interface-textencoder
+pub const TextEncoder = struct {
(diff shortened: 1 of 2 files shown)


src/browser/encoding/text_encoder.zig
```diff
@@ -0,0 +1,60 @@
...
+        return "utf-8";
+    }
+
+    pub fn _encode(_: *const TextEncoder, v: []const u8, state: *SessionState) ![]u8 {
+        // ensure the input is a valid utf-8
+        if (!std.unicode.utf8ValidateSlice(v)) {
+            return error.InvalidUtf8Input;
+        }
+
+        // raw copy b/c we know the input is utf-8
```
karlseguin: I don't think you'll need to allocate, and can use `v` directly:

```zig
pub fn _encode(_: *const TextEncoder, v: []const u8) !Env.TypedArray(u8) {
  ....
  return .{.values = v}
}
```

If you _did_ need to allocate, you could use the `call_arena`, which is shorter-lived, since the data gets copied into v8 anyways.

src/browser/encoding/text_encoder.zig
```diff
@@ -0,0 +1,60 @@
...
+
+const testing = @import("../../testing.zig");
+test "Browser.Encoding.TextEncoder" {
+    var runner = try testing.jsRunner(testing.tracking_allocator, .{});
+    defer runner.deinit();
+
+    try runner.testCases(&.{
+        .{ "var encoder = new TextEncoder();", "undefined" },
+        .{ "encoder.encoding;", "utf-8" },
+        .{ "encoder.encode('€');", "Uint8Array(3) [226, 130, 172]" },
```
karlseguin: Worth testing the error case.

src/browser/encoding/text_encoder.zig
```diff
@@ -0,0 +1,65 @@
+// Copyright (C) 2023-2024  Lightpanda (Selecy SAS)
```
karlseguin, line 1: 2025?

src/browser/encoding/text_encoder.zig
```diff
@@ -0,0 +1,65 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+const std = @import("std");
+
+const Env = @import("../env.zig").Env;
+const SessionState = @import("../env.zig").SessionState;
```
karlseguin: unused

karlseguin (approved):
LGTM

src/browser/encoding/text_encoder.zig
krichprollsch: good question. for now all the headers are to 2024. idk if we should update ti. I will talk that with @francisbouvier

PR #627, opened 2025-05-12
Return Uint8Array

krichprollsch:
For [TextEncoder](https://encoding.spec.whatwg.org/#interface-textencoder) implementation, we need to return a [`Uint8Array`](https://webidl.spec.whatwg.org/#idl-Uint8Array).

But it's not possible for now, `[]u8` are returned to JS as string.


krichprollsch:
closed in favor of #629

PR #630, opened 2025-05-12
Load about blank doc to set window

sjorsdonkers:
Playwright uses the window.document to detect when it is removed from the DOM tree.
Before this change we did not load any DOM tree so these were null pointers.
This makes sure that whenever we navigate to about:blank that a default DOM tree is still loaded and the window / document are set.

Note this behavior is in alignment with the spec: https://html.spec.whatwg.org/#read-html
> 2. If document's [URL](https://dom.spec.whatwg.org/#concept-document-url) is [about:blank](https://html.spec.whatwg.org/#about:blank), then [populate with html/head/body](https://html.spec.whatwg.org/#populate-with-html/head/body) given document.

### DIFF
--- a/src/browser/browser.zig
+++ b/src/browser/browser.zig
@@ -397,21 +397,11 @@ pub const Page = struct {
 
     // A minimal reader for about:blank such that it can be parsed by loadHTMLDoc.
     pub const AboutBlank = struct {
-        const content =
-            \\ <!DOCTYPE html>
-            \\ <html>
-            \\ <head>
-            \\     <title></title>
-            \\ </head>
-            \\ <body>
-            \\ </body>
-            \\ </html>
-        ;
         done: bool = false,
         pub fn next(self: *AboutBlank) !?[]const u8 {
             if (self.done) return null;
             self.done = true;
-            return AboutBlank.content;
+            return ""; // The contents is blank
         }
     };
 
--- a/src/browser/browser.zig
+++ b/src/browser/browser.zig
@@ -335,6 +335,8 @@ pub const Page = struct {
 
         // if the url is about:blank, nothing to do.
         if (std.mem.eql(u8, "about:blank", request_url.raw)) {
+            var about_blank = AboutBlank{};
+            try self.loadHTMLDoc(&about_blank, "utf-8");
             return;
         }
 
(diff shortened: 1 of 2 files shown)


src/browser/browser.zig
```diff
@@ -393,6 +395,26 @@ pub const Page = struct {
         });
     }
 
+    // A minimal reader for about:blank such that it can be parsed by loadHTMLDoc.
+    pub const AboutBlank = struct {
+        const content =
+            \\ <!DOCTYPE html>
+            \\ <html>
+            \\ <head>
+            \\     <title></title>
```
sjorsdonkers: Probably should remove the title :/

src/cdp/domains/target.zig
```diff
@@ -125,6 +125,10 @@ fn createTarget(cmd: anytype) !void {
     bc.target_id = target_id;
 
     var page = try bc.session.createPage();
+    // Navigate to about:blank such that the window / state.document are created and set
+    // Playwright uses these before navigating to a real URL to detect when they are removed.
+    const url = try @import("../../url.zig").URL.parse("about:blank", "");
```
karlseguin, line 130: You could maybe create a const `URL.blank` like the existing `URL.empty` (in  url.zig)

src/browser/browser.zig
```diff
@@ -335,6 +335,8 @@ pub const Page = struct {
 
         // if the url is about:blank, nothing to do.
         if (std.mem.eql(u8, "about:blank", request_url.raw)) {
+            var about_blank = AboutBlank{};
+            try self.loadHTMLDoc(&about_blank, "utf-8");
```
karlseguin, line 339: `loadHTMLDoc` does a lot of stuff you probably don't need (like walking through the the DOM, finding script tabs, setting up click events, ...), but then it probably does stuff you need, let setting the window's `html_doc`.

I wonder if the behavior can be split into two functions, `loadHTMLDoc` and `processHTMLDoc`. You'd only call `loadHTMLDoc`, and the navigate flow would call `loadHTMLDoc` followed by `processHTMLDoc`...might require reorganizing things a little. I'm just thining about the few microseconds that'll be spent on this - not sure it's worth it.

src/browser/browser.zig
```diff
@@ -335,6 +335,8 @@ pub const Page = struct {
 
         // if the url is about:blank, nothing to do.
         if (std.mem.eql(u8, "about:blank", request_url.raw)) {
+            var about_blank = AboutBlank{};
```
karlseguin, line 338: `AboutBlank` is pretty clever, but this would work too I think:

```zig
var fbs = std.io.fixedBufferStream("");
try self.loadHTMLDoc(fbs.reader(), "utf-8");
```

src/cdp/domains/target.zig
```diff
@@ -501,6 +505,11 @@ test "cdp.target: createTarget" {
         // should create a browser context
         const bc = ctx.cdp().browser_context.?;
         try ctx.expectSentEvent("Target.targetCreated", .{ .targetInfo = .{ .url = "about:blank", .title = "about:blank", .attached = false, .type = "page", .canAccessOpener = false, .browserContextId = bc.id, .targetId = bc.target_id.? } }, .{});
+
+        // Even about:blank should set the window and document
+        const page = ctx.cdp().browser_context.?.session.page.?;
+        try std.testing.expect(page.state.document != null);
```
karlseguin, line 511: There's a `const testing = @import("testing.zig");` in this file.

It's `src/cdp/testing.zig` which includes `src/testing.zig`, which extends `std.testing`. There's no `expect` there, but you can add it if you want.I tend to dislike it because it gives less useful errors, but I agree here it's the only real option.

In testing.zig

```zig
pub const expect = std.testing.expect;
```

in cdp/testing.zig

```zig
pub const expect = base.expect;
```

Or not, I think this code here is fine, just wanted to FYI about the existing testing files.

PR #631, opened 2025-05-12
Element closest

sjorsdonkers:
The closest() method of the Element interface traverses the element and its parents (heading toward the document root) until it finds a node that matches the specified CSS selector.
Returns the closest ancestor Element or itself, which matches the selectors. If there are no such element, null.

Docs: https://developer.mozilla.org/en-US/docs/Web/API/Element/closest

### DIFF
--- a/src/browser/dom/element.zig
+++ b/src/browser/dom/element.zig
@@ -135,6 +135,30 @@ pub const Element = struct {
         }
     }
 
+    // The closest() method of the Element interface traverses the element and its parents (heading toward the document root) until it finds a node that matches the specified CSS selector.
+    // Returns the closest ancestor Element or itself, which matches the selectors. If there are no such element, null.
+    pub fn _closest(self: *parser.Element, selector: []const u8, state: *SessionState) !?*parser.Element {
+        const cssParse = @import("../css/css.zig").parse;
+        const CssNodeWrap = @import("../css/libdom.zig").Node;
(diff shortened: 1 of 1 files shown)


src/browser/dom/element.zig
```diff
@@ -135,6 +135,30 @@ pub const Element = struct {
...
+    pub fn _closest(self: *parser.Element, selector: []const u8, state: *SessionState) !?*parser.Element {
+        const cssParse = @import("../css/css.zig").parse;
+        const CssNodeWrap = @import("../css/libdom.zig").Node;
+        const select = try cssParse(state.call_arena, selector, .{});
+
+        var current: CssNodeWrap = .{ .node = parser.elementToNode(self) };
+        while (true) {
+            if (try select.match(current)) {
+                if (!current.isElement()) {
+                    std.debug.print("closest: is not an element: {s}\n", .{try current.tag()});
```
karlseguin: Create a logger and use that. See the `log` global in html/window.zig for example.

src/browser/dom/element.zig
```diff
@@ -135,6 +135,30 @@ pub const Element = struct {
...
+                    return null;
+                }
+                return parser.nodeToElement(current.node);
+            }
+            if (try current.parent()) |parent| {
```
karlseguin: I think this one-liner also works (either is fine though)

```zig
current = (try current.parent()) orelse return null;
```

PR #632, opened 2025-05-12
Page Module Map

mookums:
This adds a `module_map` to the `Page` that prevents circular dependencies from entering a fetch loop. It uses the full path (obtained with the `URL.stitch`) as the key.

### DIFF
--- a/src/browser/browser.zig
+++ b/src/browser/browser.zig
@@ -241,6 +241,9 @@ pub const Page = struct {
 
     scope: *Env.Scope,
 
+    // List of modules currently fetched/loaded.
+    module_map: std.StringHashMapUnmanaged([]const u8),
+
     // current_script is the script currently evaluated by the page.
     // current_script could by fetch module to resolve module's url to fetch.
     current_script: ?*const Script = null,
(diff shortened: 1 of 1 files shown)


src/browser/browser.zig
```diff
@@ -298,10 +302,23 @@ pub const Page = struct {
...
+        const file_src = blk: {
+            if (base) |_base| {
+                break :blk try URL.stitch(self.arena, specifier, _base);
+            } else break :blk specifier;
+        };
+
+        const slot = try self.module_map.getOrPut(self.arena, file_src);
+        if (slot.found_existing) {
+            return slot.value_ptr.*;
+        } else {
```
karlseguin: don't need the `else` if you're returning above.

src/browser/browser.zig
```diff
@@ -298,10 +302,23 @@ pub const Page = struct {
...
+            const module = try self.fetchData(specifier, base);
+            if (module) |_module| slot.value_ptr.* = _module;
```
karlseguin: this is an problem, no?  You now have an entry in the map with an undefined value. If the same module is requested, `slot.found_existing` will be true, and you'll return undefined: `slot.value_ptr.*`

igs-raiderler:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

ig-raiderler:
I'm impressed by the modularity of this codebase. Makes it very maintainable.

PR #636, opened 2025-05-13
Performance now

sjorsdonkers:
Initial implementation of 2 interfaces of performance
https://developer.mozilla.org/en-US/docs/Web/API/Performance/timeOrigin
https://developer.mozilla.org/en-US/docs/Web/API/Performance/now
many still TODO

Performance implementation as part of Window, [WorkerGlobalScope](https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope) not implemented.

These APIs are implemented as part of making Playwright click work.

### DIFF
--- a/src/browser/html/performance.zig
+++ b/src/browser/html/performance.zig
@@ -34,7 +34,7 @@ pub const Performance = struct {
     const ms_resolution = 100;
 
     fn limited_resolution_ms(nanoseconds: u64) f64 {
-        const elapsed_at_resolution = ((nanoseconds / std.time.ns_per_ms) + ms_resolution / 2) / ms_resolution * ms_resolution;
+        const elapsed_at_resolution = ((nanoseconds / std.time.ns_per_us) + ms_resolution / 2) / ms_resolution * ms_resolution;
         const elapsed = @as(f64, @floatFromInt(elapsed_at_resolution));
         return elapsed / @as(f64, std.time.us_per_ms);
     }
--- a/src/browser/html/performance.zig
+++ b/src/browser/html/performance.zig
@@ -1,0 +1,87 @@
+// Copyright (C) 2023-2025  Lightpanda (Selecy SAS)
+//
+// Francis Bouvier <francis@lightpanda.io>
+// Pierre Tachoire <pierre@lightpanda.io>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+const std = @import("std");
+
+const parser = @import("../netsurf.zig");
+const EventTarget = @import("../dom/event_target.zig").EventTarget;
+
+// https://developer.mozilla.org/en-US/docs/Web/API/Performance
+pub const Performance = struct {
+    pub const prototype = *EventTarget;
+
+    // Extend libdom event target for pure zig struct.
+    base: parser.EventTargetTBase = parser.EventTargetTBase{},
+
+    time_origin: std.time.Timer,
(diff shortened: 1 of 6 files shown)


src/browser/html/performance.zig
```diff
@@ -0,0 +1,87 @@
...
+    try testing.expectDelta(@rem(time_origin * std.time.us_per_ms, 100.0), 0.0, 0.1);
+}
+
+test "Performance: now" {
+    var perf = Performance{ .time_origin = try std.time.Timer.start() };
+
+    // Monotonically increasing
+    var now = perf._now();
+    while (now <= 0) { // Loop for now to not be 0
+        try testing.expect(now == 0);
```
karlseguin: You get slightly better error messages when using `expectEqual`, it'll say "expected X, found Y"

src/browser/html/performance.zig
```diff
@@ -0,0 +1,87 @@
...
+
+    // Extend libdom event target for pure zig struct.
+    base: parser.EventTargetTBase = parser.EventTargetTBase{},
+
+    time_origin: std.time.Timer,
+    // if (Window.crossOriginIsolated) -> Resolution in isolated contexts:       5 microseconds
+    // else                            -> Resolution in non-isolated contexts: 100 microseconds
+    const ms_resolution = 100;
+
+    fn limited_resolution_ms(nanoseconds: u64) f64 {
```
karlseguin: `limitedResolutionMs`.

The other ones are `get_timedOrigin` and `_now` because they're exposed to JavaScript, but the internal functions should follow the Zig naming guidelines.

src/browser/html/performance.zig
sjorsdonkers: Ah good, it used to be an inequality. I'll leave the additions to `testing.zig` in for later.
Fixed: [c5fb4f8](https://github.com/lightpanda-io/browser/pull/636/commits/c5fb4f8e2736e8ecbac14d477db3dad2e427436b)

src/browser/html/performance.zig
sjorsdonkers: Aii, I'm switching between languages too much, thanks!
[c5fb4f8](https://github.com/lightpanda-io/browser/pull/636/commits/c5fb4f8e2736e8ecbac14d477db3dad2e427436b)

src/browser/html/performance.zig
```diff
@@ -0,0 +1,87 @@
...
+    try testing.expectDelta(@rem(time_origin * std.time.us_per_ms, 100.0), 0.0, 0.1);
+}
+
+test "Performance: now" {
+    var perf = Performance{ .time_origin = try std.time.Timer.start() };
+
+    // Monotonically increasing
+    var now = perf._now();
+    while (now <= 0) { // Loop for now to not be 0
+        try testing.expectEqual(now, 0);
```
karlseguin, line 74: Parameters are swapped, it's `expected, actual` /shrug.

src/browser/html/performance.zig
```diff
@@ -0,0 +1,87 @@
...
+    const ms_resolution = 100;
+
+    fn limitedResolutionMs(nanoseconds: u64) f64 {
+        const elapsed_at_resolution = ((nanoseconds / std.time.ns_per_us) + ms_resolution / 2) / ms_resolution * ms_resolution;
+        const elapsed = @as(f64, @floatFromInt(elapsed_at_resolution));
+        return elapsed / @as(f64, std.time.us_per_ms);
+    }
+
+    pub fn get_timeOrigin(self: *const Performance) f64 {
+        const is_posix = switch (@import("builtin").os.tag) { // From std.time.zig L125
```
karlseguin, line 43: It's fine, but just FYI, build.zig blocks any builds that aren't on linux or macos. There are a few places, like browser.zig `timestamp` function, which treat anything else as unreachable.

karlseguin:
LGTM

src/browser/html/performance.zig
sjorsdonkers: Yea, this is ugly. I wanted to just use the is_posix constant in time.zig, but that is not pub for some reason.
Then wanted to just check if `@TypeOf(std.time.Instant.timestamp) == i64` but also:
```
        const ts_field = @typeInfo(std.time.Instant).@"struct".fields[0];
        if (!std.mem.eql(u8, ts_field.name, "timestamp")) @compileError(ts_field.name);
        const is_i64 = ts_field.type == i64;
```
Does not seem to work ..

src/browser/html/performance.zig
sjorsdonkers: I'll take this change into the next PR

PR #637, opened 2025-05-13
window.requestAnimationFrame

sjorsdonkers:
Depends on: https://github.com/lightpanda-io/browser/pull/636

Not sure how I feel about immediately executing the callback.
Used in Playwright click


karlseguin:
LGTM, but you might as well implement the corresponding `cancelAnimationFrame`  which would no-op. I checked, it doesn't care if the request_id is invalid or anything.

PR #638, opened 2025-05-13
scrollIntoViewIfNeeded, getContentQuads, innerWidth/Heigh

sjorsdonkers:
- `element.scrollIntoViewIfNeeded`
- `DOM.scrollIntoViewIfNeeded`
- `DOM.getContentQuads`
- `innerHeight`
- `innerWidth`

### DIFF
--- a/src/browser/dom/element.zig
+++ b/src/browser/dom/element.zig
@@ -390,8 +390,7 @@ pub const Element = struct {
         return s.match(CssNodeWrap{ .node = parser.elementToNode(self) });
     }
 
-    pub fn _scrollIntoViewIfNeeded(self: *parser.Element, center_if_needed: ?bool) void {
-        _ = self;
+    pub fn _scrollIntoViewIfNeeded(_: *parser.Element, center_if_needed: ?bool) void {
         _ = center_if_needed;
     }
 };
--- a/src/browser/dom/element.zig
+++ b/src/browser/dom/element.zig
@@ -389,6 +389,11 @@ pub const Element = struct {
         const s = try cssParse(state.call_arena, selectors, .{});
         return s.match(CssNodeWrap{ .node = parser.elementToNode(self) });
     }
+
+    pub fn _scrollIntoViewIfNeeded(self: *parser.Element, center_if_needed: ?bool) void {
+        _ = self;
+        _ = center_if_needed;
+    }
 };
 
 // Tests
@@ -573,6 +578,12 @@ test "Browser.DOM.Element" {
         .{ "el.matches('.ok')", "true" },
         .{ "el.matches('#9000')", "false" },
         .{ "el.matches('.notok')", "false" },
+    }, .{});
+
+    try runner.testCases(&.{
+        .{ "const el3 = document.createElement('div');", "undefined" },
+        .{ "el3.scrollIntoViewIfNeeded();", "undefined" },
+        .{ "el3.scrollIntoViewIfNeeded(false);", "undefined" },
     }, .{});
 
     // before
--- a/src/browser/html/window.zig
+++ b/src/browser/html/window.zig
@@ -120,6 +120,18 @@ pub const Window = struct {
         return &self.history;
     }
 
+    //  The interior height of the window in pixels, including the height of the horizontal scroll bar, if present.
+    pub fn get_innerHeight(_: *Window, state: *SessionState) u32 {
+        // We do not have scrollbars or padding so this is the same as Element.clientHeight
+        return state.renderer.height();
+    }
+
+    // The interior width of the window in pixels. That includes the width of the vertical scroll bar, if one is present.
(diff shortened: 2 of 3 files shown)


src/browser/dom/element.zig
```diff
@@ -389,6 +389,10 @@ pub const Element = struct {
         const s = try cssParse(state.call_arena, selectors, .{});
         return s.match(CssNodeWrap{ .node = parser.elementToNode(self) });
     }
+
+    pub fn _scrollIntoViewIfNeeded(_: *parser.Element, center_if_needed: ?bool) void {
+        _ = center_if_needed;
```
sjorsdonkers, line 394: We could register the element in the renderer..

src/browser/html/window.zig
```diff
@@ -291,4 +303,13 @@ test "Browser.HTML.Window" {
         .{ "let request_id = requestAnimationFrame(timestamp => {});", "undefined" },
         .{ "cancelAnimationFrame(request_id);", "undefined" },
     }, .{});
+
+    try runner.testCases(&.{
+        .{ "innerHeight", "1" },
+        .{ "innerWidth", "1" }, // Width is 1 even if there are no elements
+        .{ "document.createElement('div').getClientRects()", "[object Object]" },
```
karlseguin: I was inspired: https://github.com/lightpanda-io/browser/pull/644

src/cdp/domains/dom.zig
```diff
@@ -233,25 +239,84 @@ fn describeNode(cmd: anytype) !void {
...
+fn scrollIntoViewIfNeeded(cmd: anytype) !void {
+    _ = (try cmd.params(struct {
+        nodeId: ?Node.Id = null,
+        backendNodeId: ?u32 = null,
+        objectId: ?[]const u8 = null,
+        rect: ?DOMRect = null,
+    })) orelse return error.InvalidParams;
+
+    // Only 1 of nodeId, backendNodeId, objectId may be set, but we don't want to error unnecessarily
+    // TBD what do other browsers do in this user error sceneario?
```
karlseguin: Seems to use the `nodeId`.

src/cdp/domains/dom.zig
```diff
@@ -233,25 +239,84 @@ fn describeNode(cmd: anytype) !void {
...
+
+    // Since element.scrollIntoViewIfNeeded is a no-op we do not bother retrieving the node.
+    // This however also means we also do not error in case the node is not found.
```
karlseguin: They _do_ give an error if the node isn't found.

`Could not find node with given id"`

I also saw:

`Node does not have a layout object"`

src/cdp/domains/dom.zig
```diff
@@ -233,25 +239,84 @@ fn describeNode(cmd: anytype) !void {
...
+
+    return cmd.sendResult(null, .{});
+}
+
+fn getNode(arena: Allocator, browser_context: anytype, node_id: ?Node.Id, backend_node_id: ?Node.Id, object_id: ?[]const u8) !*Node {
+    const input_node_id = node_id orelse backend_node_id;
+    if (input_node_id) |input_node_id_| {
+        return browser_context.node_registry.lookup_by_id.get(input_node_id_) orelse return error.NodeNotFound;
```
karlseguin, line 296: Would a `getById` function on registry be useful?

src/cdp/domains/dom.zig
```diff
@@ -233,25 +239,84 @@ fn describeNode(cmd: anytype) !void {
...
     }
-    if (params.objectId != null) {
+    if (object_id) |object_id_| {
         // Retrieve the object from which ever context it is in.
-        const parser_node = try bc.inspector.getNodePtr(cmd.arena, params.objectId.?);
-        const node = try bc.node_registry.register(@ptrCast(parser_node));
-        return cmd.sendResult(.{ .node = bc.nodeWriter(node, .{}) }, .{});
+        const parser_node = try browser_context.inspector.getNodePtr(arena, object_id_);
```
karlseguin, line 300: I don't really know what `getNodePtr` is doing. It's ok that its memory (from the arena) will be gone after this function returns?

src/cdp/domains/dom.zig
sjorsdonkers: To reduce: `.lookup_ _`. I personally am not a fan of convenience functions.

src/cdp/domains/dom.zig
sjorsdonkers: Only the `object_group` may be returned on the allocator, which is not used. (Arena is also used for temporary error strings, but not returned (returned as error enum))

src/cdp/domains/dom.zig
sjorsdonkers: Added check whether node exists and is of the right type. [f74647c](https://github.com/lightpanda-io/browser/pull/638/commits/f74647ccfca5e48d3b4abf660222f71f41cc11c0)

PR #640, opened 2025-05-14
Don't load script tags with the nomodule attribute

karlseguin:
These tags should not be loaded as we support ES modules.


sjorsdonkers (approved):
LGTM

PR #641, opened 2025-05-14
Make Callback.printFunc public

karlseguin:
When calling a Zig function from JS fails due to a parameter type error, log.debug information about the function and parameters.

### DIFF
--- a/src/runtime/js.zig
+++ b/src/runtime/js.zig
@@ -870,7 +870,7 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
             }
 
             // debug/helper to print the source of the JS callback
-            fn printFunc(self: Callback) !void {
+            pub fn printFunc(self: Callback) !void {
                 const scope = self.scope;
                 const value = self.func.castToFunction().toValue();
                 const src = try valueToString(scope.call_arena, value, scope.isolate, scope.context);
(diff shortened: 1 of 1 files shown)


sjorsdonkers (commented):
How do I get this to print?
When I add an argument (to IntersectionObserver.observer()), I get nothing in the server terminal and the same invalid argument response to the client.

karlseguin:
It's a debugging tool, in case you're getting some JS error in a callback, and want to see what that callback is.
```zig
    pub fn constructor(callback: Env.Callback, options_: ?IntersectionObserverOptions, state: *SessionState) !IntersectionObserver {
        // ADDED
        try callback.printFunc();
        ....
```

```
$ make test F="Browser.DOM.IntersectionObserver"
() => {}
entries => {count_a += 1;}
entries => {count_b = entries.length;}
entries => {count_bb = entries.length;}
entries => { count_c = entries.length;}
entries => {}
entries => {}
entries => { entry = entries[0]; }
entries => { new_entry = entries[0]; }
Browser.DOM.IntersectionObserver (13.87ms)
```

sjorsdonkers (approved):
Ah, got it.
We could look into always calling this when we have an invalid arguments error on a callback (in debug)

PR #642, opened 2025-05-14
elementFromPoint + elementsFromPoint

sjorsdonkers:
Depends on: https://github.com/lightpanda-io/browser/pull/638

Implements the last 2 web apis for Playwright click:
- `Document.elementFromPoint`
- `Document.elementsFromPoint`
https://developer.mozilla.org/en-US/docs/Web/API/Document/elementFromPoint

Reimplemented:
- `Node.getRootNode` also in netsurf
- `Element.getBoundingClientRect`
- `Element.getClientRects`

### DIFF
--- a/src/browser/dom/element.zig
+++ b/src/browser/dom/element.zig
@@ -384,7 +384,7 @@ pub const Element = struct {
         if (root != parser.documentToNode(parser.documentHTMLToDocument(state.document.?))) {
             return &.{};
         }
-        const heap_ptr = try state.arena.create(DOMRect);
+        const heap_ptr = try state.call_arena.create(DOMRect);
         heap_ptr.* = try state.renderer.getRect(self);
         return heap_ptr[0..1];
     }
--- a/src/browser/dom/element.zig
+++ b/src/browser/dom/element.zig
@@ -370,7 +370,7 @@ pub const Element = struct {
     pub fn _getBoundingClientRect(self: *parser.Element, state: *SessionState) !DOMRect {
         // Since we are lazy rendering we need to do this check. We could store the renderer in a viewport such that it could cache these, but it would require tracking changes.
         const root = try parser.nodeGetRootNode(parser.elementToNode(self));
-        if (root != parser.documentToNode(parser.documentHTMLToDocument(state.document.?))) {
+        if (root != parser.documentToNode(parser.documentHTMLToDocument(state.window.document.?))) {
             return DOMRect{ .x = 0, .y = 0, .width = 0, .height = 0 };
         }
         return state.renderer.getRect(self);
@@ -381,7 +381,7 @@ pub const Element = struct {
     // Returns an empty array if the element is eventually detached from the main window
     pub fn _getClientRects(self: *parser.Element, state: *SessionState) ![]DOMRect {
         const root = try parser.nodeGetRootNode(parser.elementToNode(self));
-        if (root != parser.documentToNode(parser.documentHTMLToDocument(state.document.?))) {
+        if (root != parser.documentToNode(parser.documentHTMLToDocument(state.window.document.?))) {
             return &.{};
         }
         const heap_ptr = try state.call_arena.create(DOMRect);
--- a/src/browser/dom/element.zig
+++ b/src/browser/dom/element.zig
@@ -365,14 +365,28 @@ pub const Element = struct {
         return Node.replaceChildren(parser.elementToNode(self), nodes);
     }
 
+    // A DOMRect object providing information about the size of an element and its position relative to the viewport.
+    // Returns a 0 DOMRect object if the element is eventually detached from the main window
     pub fn _getBoundingClientRect(self: *parser.Element, state: *SessionState) !DOMRect {
+        // Since we are lazy rendering we need to do this check. We could store the renderer in a viewport such that it could cache these, but it would require tracking changes.
+        const root = try parser.nodeGetRootNode(parser.elementToNode(self));
+        if (root != parser.documentToNode(parser.documentHTMLToDocument(state.document.?))) {
+            return DOMRect{ .x = 0, .y = 0, .width = 0, .height = 0 };
+        }
         return state.renderer.getRect(self);
     }
 
-    // returns a collection of DOMRect objects that indicate the bounding rectangles for each CSS border box in a client.
-    // We do not render so just always return the element's rect.
-    pub fn _getClientRects(self: *parser.Element, state: *SessionState) ![1]DOMRect {
-        return [_]DOMRect{try state.renderer.getRect(self)};
+    // Returns a collection of DOMRect objects that indicate the bounding rectangles for each CSS border box in a client.
+    // We do not render so it only always return the element's bounding rect.
+    // Returns an empty array if the element is eventually detached from the main window
+    pub fn _getClientRects(self: *parser.Element, state: *SessionState) ![]DOMRect {
+        const root = try parser.nodeGetRootNode(parser.elementToNode(self));
+        if (root != parser.documentToNode(parser.documentHTMLToDocument(state.document.?))) {
+            return &.{};
+        }
+        const heap_ptr = try state.arena.create(DOMRect);
+        heap_ptr.* = try state.renderer.getRect(self);
+        return heap_ptr[0..1];
     }
 
     pub fn get_clientWidth(_: *parser.Element, state: *SessionState) u32 {
@@ -568,6 +582,26 @@ test "Browser.DOM.Element" {
 
         .{ "document.getElementById('para').clientWidth", "2" },
         .{ "document.getElementById('para').clientHeight", "1" },
+
+        .{ "let r4 = document.createElement('div').getBoundingClientRect()", null },
+        .{ "r4.x", "0" },
+        .{ "r4.y", "0" },
+        .{ "r4.width", "0" },
+        .{ "r4.height", "0" },
+
+        // Test setup causes WrongDocument or HierarchyRequest error unlike in chrome/firefox
+        // .{ // An element of another document, even if created from the main document, is not rendered.
+        //     \\ let div5 = document.createElement('div');
+        //     \\ const newDoc = document.implementation.createHTMLDocument("New Document");
+        //     \\ newDoc.body.appendChild(div5);
+        //     \\ let r5 = div5.getBoundingClientRect();
+        //     ,
+        //     null,
+        // },
+        // .{ "r5.x", "0" },
+        // .{ "r5.y", "0" },
+        // .{ "r5.width", "0" },
+        // .{ "r5.height", "0" },
     }, .{});
 
     try runner.testCases(&.{
--- a/src/browser/dom/node.zig
+++ b/src/browser/dom/node.zig
@@ -41,6 +41,8 @@ const Walker = @import("walker.zig").WalkerDepthFirst;
 const HTML = @import("../html/html.zig");
 const HTMLElem = @import("../html/elements.zig");
 
+const log = std.log.scoped(.node);
+
 // Node interfaces
 pub const Interfaces = .{
     Attr,
@@ -262,13 +264,15 @@ pub const Node = struct {
         return try parser.nodeContains(self, other);
     }
 
-    pub fn _getRootNode(self: *parser.Node) !?HTMLElem.Union {
-        // TODO return this’s shadow-including root if options["composed"] is true
-        const res = try parser.nodeOwnerDocument(self);
-        if (res == null) {
-            return null;
-        }
-        return try HTMLElem.toInterface(HTMLElem.Union, @as(*parser.Element, @ptrCast(res.?)));
+    // Returns itself or ancestor object inheriting from Node.
+    // - An Element inside a standard web page will return an HTMLDocument object representing the entire page (or <iframe>).
+    // - An Element inside a shadow DOM will return the associated ShadowRoot.
+    // - An Element that is not attached to a document or a shadow tree will return the root of the DOM tree it belongs to
+    pub fn _getRootNode(self: *parser.Node, options: ?struct { composed: bool = false }) !Union {
+        if (options) |options_| if (options_.composed) {
+            log.warn("getRootNode composed is not implemented yet", .{});
+        };
+        return try Node.toInterface(try parser.nodeGetRootNode(self));
     }
 
     pub fn _hasChildNodes(self: *parser.Node) !bool {
--- a/src/browser/html/document.zig
+++ b/src/browser/html/document.zig
@@ -230,7 +230,7 @@ pub const HTMLDocument = struct {
 
     // Returns the topmost Element at the specified coordinates (relative to the viewport).
     // Since LightPanda requires the client to know what they are clicking on we do not return the underlying element at this moment
-    // This can currenty only happen if the first pixel is click without having rendered any element. This will change when css properties are supported.
+    // This can currenty only happen if the first pixel is clicked without having rendered any element. This will change when css properties are supported.
     // This returns an ElementUnion instead of a *Parser.Element in case the element somehow hasn't passed through the js runtime yet.
     pub fn _elementFromPoint(_: *parser.DocumentHTML, x: f32, y: f32, state: *SessionState) !?ElementUnion {
         const ix: i32 = @intFromFloat(@floor(x));
@@ -247,8 +247,9 @@ pub const HTMLDocument = struct {
         const element = state.renderer.getElementAtPosition(ix, iy) orelse return &.{};
         // TODO if pointer-events set to none the underlying element should be returned (parser.documentGetDocumentElement(self.document);?)
 
-        var list = try std.ArrayList(ElementUnion).initCapacity(state.call_arena, 3);
-        try list.append(try Element.toInterface(element));
+        var list: std.ArrayListUnmanaged(ElementUnion) = .empty;
+        try list.ensureTotalCapacity(state.call_arena, 3);
+        list.appendAssumeCapacity(try Element.toInterface(element));
 
         // Since we are using a flat renderer there is no hierarchy of elements. What we do know is that the element is part of the main document.
         // Thus we can add the HtmlHtmlElement and it's child HTMLBodyElement to the returned list.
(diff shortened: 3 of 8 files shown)


src/browser/html/document.zig
```diff
@@ -207,6 +207,23 @@ pub const HTMLDocument = struct {
     pub fn set_bgColor(_: *parser.DocumentHTML, _: []const u8) []const u8 {
         return "";
     }
+
+    pub fn _elementFromPoint(_: *parser.DocumentHTML, x: f32, y: f32, state: *SessionState) !?*parser.Element {
```
sjorsdonkers:
```suggestion
    pub fn _elementFromPoint(_: *parser.DocumentHTML, x: f32, y: f32, state: *SessionState) !?ElementUnion {
```

Should `ElementUnion` be returned?

karlseguin: Yes, probably. If you return it as a `*parser.Element`, it'll be the v8 ObjectTemplate associated with that type, i.e. the `Element` struct in `dom/element.zig`. That's a problem for any "subtype", since you wouldn't be able to call any of the subtype methods. For example, if the element is actually a `HTMLAnchorElement`, you'd want to be able to call `el.href` in JavaScript. It's only by returning the union that the js runtime can pick the correct ObjectTemplate.

src/browser/html/document.zig
```diff
@@ -207,6 +207,28 @@ pub const HTMLDocument = struct {
...
+        return element;
+    }
+
+    pub fn _elementsFromPoint(_: *parser.DocumentHTML, x: f32, y: f32, state: *SessionState) ![]*parser.Element {
+        const ix: i32 = @intFromFloat(@floor(x));
+        const iy: i32 = @intFromFloat(@floor(y));
+        const element = state.renderer.getElementAtPosition(ix, iy) orelse return &.{};
+        // TODO if pointer-events set to none the underlying element should be returned (parser.documentGetDocumentElement(self.document);?)
+
+        // We need to return either 0 or 1 item, so we cannot fix the size to [1]*parser.Element
```
karlseguin: The only way I can think of to solve this is to introduce a new discriminatory type, something like 

```zig
fn MaybeArray(comptime T: type, max: usize) type {
   return struct {
      len: usize,
      values: [max]T,
  };
}
```

This is akin to Zig's `std.BoundedArray`, but we can't use that directly in js.zig, because we'd have no way to detect that type and apply special handling.


Then you could do:

```zig
var arr = MaybeArray(*parser.Element, 1);
const element = state.renderer.getElem orelse return arr;

arr.values.append(element);
return arr;
```

Think it's worth it?

sjorsdonkers: Hmm, a fixed capacity array, right.
Performance wise we are using an arena, so it is similar to a stack copy of 8 bytes.
Considering the interface of the function an unbounded slice is marginally more correct to the spec.
Complexity wise I would prefer not introducing new types/abstractions.
Safety wise, I think the MaybeArray is harder to mess up (once implemented, I think the append is wrong in the above as it is not updating len :) ), Though on second though, the abstraction may make it easier to accidentally return a slice to a stack value.., may be not.

I would keep it as is. Good to consider tho.

src/browser/html/document.zig
sjorsdonkers: Are you sure? Happy to change it of course, but the tests do indicate a `[object HTMLDivElement]` is returned. (checking atm)

karlseguin: It works in this case, because the element being returned has already passed through the js runtime, so it isn't returning a new object (from the point of view of the js runtime), but an existing one.

But if `document.elementFromPoint(0.5, 0.5)` returned an object that hadn't been seen before, it would create a new instance of the wrong type.

src/browser/html/document.zig
```diff
@@ -260,4 +282,14 @@ test "Browser.HTML.Document" {
         .{ "document.cookie = 'favorite_food=tripe; SameSite=None; Secure'", "favorite_food=tripe; SameSite=None; Secure" },
         .{ "document.cookie", "name=Oeschger; favorite_food=tripe" },
     }, .{});
+
+    try runner.testCases(&.{
+        .{ "document.elementFromPoint(0.5, 0.5)", "null" },
```
karlseguin: I think this is a problem. Why does it return `null`? I mean, I know _why_ it returns `null`, but I don't think it should.

This is the HTML document:
```html
 <div id="content">
  <a id="link" href="foo" class="ok">OK</a>
  <p id="para-empty" class="ok empty">
    <span id="para-empty-child"></span>
  </p>
  <p id="para"> And</p>
  <!--comment-->
</div>
```

shouldn't there be something at 0.5, 0.5?

src/browser/html/document.zig
```diff
@@ -260,4 +282,14 @@ test "Browser.HTML.Document" {
...
+        .{ "document.elementsFromPoint(0.5, 0.5)", "" },
+        .{ "document.createElement('div').getClientRects()", null },
```
karlseguin: This might highlight a problem with `getClientRects`. The `div` is never added to the document, so I don't think it should be possible for it to be returned from a `document.elementFromPoint` call.

src/browser/html/document.zig
karlseguin: I know that, right now,`elementFromPoint` can only returned a previously-seen elements, but I'm not sure that's the right behavior, as I pointed out in my comment that I'm not sure if `.{ "document.elementFromPoint(0.5, 0.5)", "null" }` is correct.

src/browser/html/document.zig
sjorsdonkers: Our renderer is lazy so only renders things that have been registered.
The user cannot really know what the location is of any element without having queried for its location.

I think that returning `document.documentElement` is better.
Not sure if we want to (re-)register all elements when we call `getElementAtPosition`. The only use case I can currently see is when the script relies on the renderer being filled in a deterministic way, which I think they should not.

sjorsdonkers: We will however likely need to make our rendered a bit less flat.
Meaning a bit of a tree where an element can hold a list of other documents.

src/browser/html/document.zig
sjorsdonkers: I think getClientRects is fine since it specifies the Rect wrt the whole viewport.
Currently the viewport and the Window document are the same, but won't be in the future.

I think we should store the root document in the renderer, such that we can check whether a queried element (from elementFromPoint) is a descendant of the root. Later we can support multiple documents in the renderer.

sjorsdonkers: Fixed in [4cf7000](https://github.com/lightpanda-io/browser/pull/642/commits/4cf7000b4e7f2e4ca580667dc3a410e1aeeed706)

src/browser/html/document.zig
sjorsdonkers: I think we should document very clearly how our renderer works and what is and what is not a valid use case scenario.
We need to avoid falling into a pit where we end up having to implement a full renderer.

I'll start a discussion to document the current design ideas and goals of the renderer, and how that changes what kind of script we support (like no guessing what is as 0.5 0.5).
The only alternative that I see that would be better is if we have an on-the-fly renderer. That only renders everything on demand for a function such as this, but still does not keep track of changes and such.
For now I think we should stick with the no guessing (and no overlaying approach until it breaks)

src/browser/dom/element.zig
```diff
@@ -365,14 +365,28 @@ pub const Element = struct {
...
     }
 
+    // A DOMRect object providing information about the size of an element and its position relative to the viewport.
+    // Returns a 0 DOMRect object if the element is eventually detached from the main window
     pub fn _getBoundingClientRect(self: *parser.Element, state: *SessionState) !DOMRect {
+        // Since we are lazy rendering we need to do this check. We could store the renderer in a viewport such that it could cache these, but it would require tracking changes.
+        const root = try parser.nodeGetRootNode(parser.elementToNode(self));
+        if (root != parser.documentToNode(parser.documentHTMLToDocument(state.document.?))) {
+            return DOMRect{ .x = 0, .y = 0, .width = 0, .height = 0 };
+        }
```
sjorsdonkers: This logic is here for now as I think it definitely should not be in the FlatRenderer.
We may wrap the FlatRenderer in a ViewPort kind of object which would handle these as well as multiple documents and document renderer relationships.
But I do not think this is the right time to go there.

src/browser/dom/element.zig
```diff
@@ -568,6 +582,26 @@ test "Browser.DOM.Element" {
...
         .{ "document.getElementById('para').clientWidth", "2" },
         .{ "document.getElementById('para').clientHeight", "1" },
+
+        .{ "let r4 = document.createElement('div').getBoundingClientRect()", null },
+        .{ "r4.x", "0" },
+        .{ "r4.y", "0" },
+        .{ "r4.width", "0" },
+        .{ "r4.height", "0" },
+
+        // Test setup causes WrongDocument or HierarchyRequest error unlike in chrome/firefox
```
sjorsdonkers, line 592: I believe that we should not be throwing these errors in this use case, so a pre-exiting bug that is not fixed in the PR

src/browser/html/document.zig
```diff
@@ -212,6 +214,45 @@ pub const HTMLDocument = struct {
...
+        const ix: i32 = @intFromFloat(@floor(x));
+        const iy: i32 = @intFromFloat(@floor(y));
+        const element = state.renderer.getElementAtPosition(ix, iy) orelse return &.{};
+        // TODO if pointer-events set to none the underlying element should be returned (parser.documentGetDocumentElement(self.document);?)
+
+        var list = try std.ArrayList(ElementUnion).initCapacity(state.call_arena, 3);
+        try list.append(try Element.toInterface(element));
+
+        // Since we are using a flat renderer there is no hierarchy of elements. What we do know is that the element is part of the main document.
+        // Thus we can add the HtmlHtmlElement and it's child HTMLBodyElement to the returned list.
```
sjorsdonkers, line 255: Thoughts on this behavior are welcome.
Chrome returns: 
image
We return [0, 9, 10]

src/browser/html/document.zig
sjorsdonkers: ElementUnion has been changed, resolving this discussion other discussion continues here: https://github.com/lightpanda-io/browser/pull/642#discussion_r2095245929

src/browser/html/document.zig
krichprollsch: it makes sense. I'm wondering if it's possible to have something out of the body, but I don't think so.

src/browser/dom/element.zig
```diff
@@ -365,14 +365,28 @@ pub const Element = struct {
         return Node.replaceChildren(parser.elementToNode(self), nodes);
     }
 
+    // A DOMRect object providing information about the size of an element and its position relative to the viewport.
+    // Returns a 0 DOMRect object if the element is eventually detached from the main window
     pub fn _getBoundingClientRect(self: *parser.Element, state: *SessionState) !DOMRect {
+        // Since we are lazy rendering we need to do this check. We could store the renderer in a viewport such that it could cache these, but it would require tracking changes.
+        const root = try parser.nodeGetRootNode(parser.elementToNode(self));
+        if (root != parser.documentToNode(parser.documentHTMLToDocument(state.document.?))) {
```
karlseguin: Can also do:

```zig
@intFromPtr(root) == @intFromPtr(state.document.?)
```

but I'm not sure one is that much better than the other.

src/browser/dom/element.zig
```diff
@@ -365,14 +365,28 @@ pub const Element = struct {
...
-        return [_]DOMRect{try state.renderer.getRect(self)};
+    // Returns a collection of DOMRect objects that indicate the bounding rectangles for each CSS border box in a client.
+    // We do not render so it only always return the element's bounding rect.
+    // Returns an empty array if the element is eventually detached from the main window
+    pub fn _getClientRects(self: *parser.Element, state: *SessionState) ![]DOMRect {
+        const root = try parser.nodeGetRootNode(parser.elementToNode(self));
+        if (root != parser.documentToNode(parser.documentHTMLToDocument(state.document.?))) {
+            return &.{};
+        }
+        const heap_ptr = try state.arena.create(DOMRect);
```
karlseguin: You should be safe to use the `call_arena` her, as the slice will get turned into a VM array

src/browser/dom/node.zig
```diff
@@ -262,13 +264,15 @@ pub const Node = struct {
...
-        return try HTMLElem.toInterface(HTMLElem.Union, @as(*parser.Element, @ptrCast(res.?)));
+    // Returns itself or ancestor object inheriting from Node.
+    // - An Element inside a standard web page will return an HTMLDocument object representing the entire page (or <iframe>).
+    // - An Element inside a shadow DOM will return the associated ShadowRoot.
+    // - An Element that is not attached to a document or a shadow tree will return the root of the DOM tree it belongs to
+    pub fn _getRootNode(self: *parser.Node, options: ?struct { composed: bool = false }) !Union {
+        if (options) |options_| if (options_.composed) {
+            log.warn("getRootNode composed is not implemented yet", .{});
+        };
+        return try Node.toInterface(try parser.nodeGetRootNode(self));
```
karlseguin, line 275: Just trying to understand.

If `@intFromPtr(self) == @intFromPtr(state.document.?)`, you could use the existing `parser.nodeOwnerDocument` ?

But for cases where the document isn't attached, and possibly also the shadow dom, you use the new `nodeGetRootNode` (which is also safe to use if it _is_ attached, just not as efficient).

Is that right?

src/browser/html/document.zig
```diff
@@ -212,6 +214,45 @@ pub const HTMLDocument = struct {
     pub fn set_bgColor(_: *parser.DocumentHTML, _: []const u8) []const u8 {
         return "";
     }
+
+    // Returns the topmost Element at the specified coordinates (relative to the viewport).
+    // Since LightPanda requires the client to know what they are clicking on we do not return the underlying element at this moment
+    // This can currenty only happen if the first pixel is click without having rendered any element. This will change when css properties are supported.
```
karlseguin:
```suggestion
    // This can currenty only happen if the first pixel is clicked without having rendered any element. This will change when css properties are supported.
```

src/browser/html/document.zig
```diff
@@ -212,6 +214,45 @@ pub const HTMLDocument = struct {
...
+    }
+
+    // Returns an array of all elements at the specified coordinates (relative to the viewport). The elements are ordered from the topmost to the bottommost box of the viewport.
+    pub fn _elementsFromPoint(_: *parser.DocumentHTML, x: f32, y: f32, state: *SessionState) ![]ElementUnion {
+        const ix: i32 = @intFromFloat(@floor(x));
+        const iy: i32 = @intFromFloat(@floor(y));
+        const element = state.renderer.getElementAtPosition(ix, iy) orelse return &.{};
+        // TODO if pointer-events set to none the underlying element should be returned (parser.documentGetDocumentElement(self.document);?)
+
+        var list = try std.ArrayList(ElementUnion).initCapacity(state.call_arena, 3);
```
karlseguin: Stick with the unmanaged variant (i.e. the ones that don't take an allocator). There's a chance the managed one get removed.

```zig
var list: std.ArrayListUnmanaged(ElementUnion) = .empty;
try list.ensureTotalCapacity(state.call_arena, 3) ;
```

src/browser/html/document.zig
```diff
@@ -212,6 +214,45 @@ pub const HTMLDocument = struct {
...
+        try list.append(try Element.toInterface(element));
```
karlseguin: If you know you've sized it correctly, and you want to avoid length comparison per append,  you can use `appendAssumeCapacity` (even the unmanaged version of this doesn't take an allocator, since the assumption is that it won't have to allocate).

src/browser/html/document.zig
```diff
@@ -212,6 +214,45 @@ pub const HTMLDocument = struct {
...
+
+        // Since we are using a flat renderer there is no hierarchy of elements. What we do know is that the element is part of the main document.
+        // Thus we can add the HtmlHtmlElement and it's child HTMLBodyElement to the returned list.
+        // TBD Should we instead return every parent that is an element? Note that a child does not physically need to be overlapping the parent.
+        // Should we do a render pass on demand?
+        const doc_elem = try parser.documentGetDocumentElement(parser.documentHTMLToDocument(state.document.?)) orelse {
+            return list.items;
+        };
+        const body = try parser.documentHTMLBody(state.document.?) orelse {
```
karlseguin: This you can deduplicate just a little bit here:

```zig
if (try parser.documentHTMLBody(state.document.?)) |body| {
  try list.append(try Element.toInterface(parser.bodyToElement(body)));
}
try list.append(try Element.toInterface(doc_elem));
return list.items;
```

src/browser/dom/node.zig
sjorsdonkers: 1. if `self == state.document` than we can return `self` [htmldocument ]
1. if `self == state.document.documentElement` than we can return `state.document` [htmldocument ]
In those cases that would also be the same as `parser.nodeOwnerDocument`

Note that `nodeOwnerDocument` is set even for detached nodes. It references the original document until it is appended as child to another document.
So we cannot use `nodeOwnerDocument` to determine whether a node is detached.
So yes, if we know that the node is attached `parser.nodeOwnerDocument` would yield the correct answer.

I actually now use the difference between getRootNode and nodeOwnerDocument to determine whether an element is or is not detached.

src/browser/dom/element.zig
sjorsdonkers: Tnx: [c9ce527](https://github.com/lightpanda-io/browser/pull/642/commits/c9ce5276566144e851d87be4b59f0b0e5097f4bb)

src/browser/html/document.zig
sjorsdonkers: fixed [c9ce527](https://github.com/lightpanda-io/browser/pull/642/commits/c9ce5276566144e851d87be4b59f0b0e5097f4bb)

src/browser/html/document.zig
sjorsdonkers: Fixed [c9ce527](https://github.com/lightpanda-io/browser/pull/642/commits/c9ce5276566144e851d87be4b59f0b0e5097f4bb)

src/browser/html/document.zig
sjorsdonkers: Fixed: [c9ce527](https://github.com/lightpanda-io/browser/pull/642/commits/c9ce5276566144e851d87be4b59f0b0e5097f4bb)

src/browser/html/document.zig
sjorsdonkers: yup: [c9ce527](https://github.com/lightpanda-io/browser/pull/642/commits/c9ce5276566144e851d87be4b59f0b0e5097f4bb)

src/browser/dom/element.zig
sjorsdonkers: Ye, I choose not to as making it a habit means every time that code is copied one would need to make sure the pointers of comparable types. Now if there is a function netsurf.zig that does the conversion it should be safe. But also meh :) I left it as it was.

PR #644, opened 2025-05-14
Make expected runner value optional to skip assertion

karlseguin:


src/testing.zig
```diff
@@ -455,7 +455,7 @@ pub const JsRunner = struct {
     }
 
     const RunOpts = struct {};
-    pub const Case = std.meta.Tuple(&.{ []const u8, []const u8 });
+    pub const Case = std.meta.Tuple(&.{ []const u8, ?[]const u8 });
```
sjorsdonkers, line 458: We probably want this in `src/runtime/testing.zig` as well

sjorsdonkers (approved):
LGTM

src/testing.zig
karlseguin: yes, done.

PR #645, opened 2025-05-14
Normalize HTML Title Whitespace

mookums:
This solves the failing WPT: `html/dom/documents/dom-tree-accessors/document.title-03.html` by ensuring that any whitespace subslices are replaced by a single space.


karlseguin:
You _might_ be able to do this without an allocator.

The web apis, i.e. `set_title`, can take a `title: []u8` rather than a `[]const u8`, so you could normalize it in-place.

That leaves `documentHTMLParse`, and the question becomes, can the `dom_string` you get from `get_title` be mutated and passed back to `set_title`?? Worst case you could use the netsurf arena to create the new string, i.e. `c.dom_string_create`

mookums:
Going to close this. With our general focus to be as slim and performant as possible, feels antithetical to do an allocation just to pass a single WPT.

PR #646, opened 2025-05-15
Move Session, Page and Renderer into their own respective files

karlseguin:
I'm fine with keeping things as-is, but it is getting a little messy in there, and we're only going to add more and more logic. I think this is cleaner, and the sooner we do it, the easier.

The main downside is that some functions, like session.init need to be made `pub`, whereas before, they weren't.


sjorsdonkers (approved):
Yea, I agree we should do this.
Haven't checked if we have any unused imports, but it looks good.

krichprollsch (approved):

PR #647, opened 2025-05-15
add FormData web API

karlseguin:


src/browser/url/form_data.zig
```diff
@@ -0,0 +1,245 @@
...
+
+    pub fn _values(self: *const FormData) ValueIterable {
+        return .{ .inner = .{ .entries = &self.entries } };
+    }
+
+    pub fn _entries(self: *const FormData) EntryIterable {
+        return .{ .inner = .{ .entries = &self.entries } };
+    }
+
+    pub fn _symbol_iterator(self: *const FormData) EntryIterable {
```
sjorsdonkers: fyi I can't find this a spec online..

sjorsdonkers (approved):
LGTM

src/browser/url/form_data.zig
karlseguin: Not 100% sure how to read the spec here, but it does mention: `iterable<USVString, FormDataEntryValue>`.

MDN lists the iterator here: https://developer.mozilla.org/en-US/docs/Web/API/FormData#browser_compatibility

And, a quick test in browser shows that the form data is iterable:

```js
let f = new FormData()
f.set('a', '1');
for (entry of f) console.log(entry)
```

will print `Array [ "a", "1" ]`

PR #651, opened 2025-05-16
Rework HTMLAllCollection

karlseguin:
Capture its unique properties:
1- instances are falsy, and
2- instance can be called as a function

The behavior is used for browser detection (i.e. duckduckgo treats us as a legacy browser because we document.all != false)

### DIFF
--- a/src/browser/dom/html_collection.zig
+++ b/src/browser/dom/html_collection.zig
@@ -149,17 +149,43 @@ pub fn HTMLCollectionByName(
     };
 }
 
-pub fn HTMLCollectionAll(
-    root: ?*parser.Node,
-    include_root: bool,
-) !HTMLCollection {
-    return HTMLCollection{
-        .root = root,
-        .walker = .{ .walkerDepthFirst = .{} },
-        .matcher = .{ .matchTrue = .{} },
-        .include_root = include_root,
+// HTMLAllCollection is a special type: instances of it are falsy. It's the only
+// object in the WebAPI that behaves like this - in fact, it's even a special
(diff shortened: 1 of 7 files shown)


src/browser/dom/html_collection.zig
```diff
@@ -149,17 +149,43 @@ pub fn HTMLCollectionByName(
...
+
+// It's also possible to use an instance like a function:
+//   document.all(3)
+//   document.all('some_id')
+pub const HTMLAllCollection = struct {
+    pub const prototype = *HTMLCollection;
+
+    proto: HTMLCollection,
+
+    pub const mark_as_undetectable = true;
```
krichprollsch, line 166: I saw the PR in zig v8, but I'm not sure about what undetectable means exactly.

Setting undetectable allows the falsy behavior?

krichprollsch (approved):

src/browser/dom/html_collection.zig
karlseguin: This is the V8 terminology. From the documentation

> In many ways, undetectable objects behave as though they are not there. They behave like 'undefined' in conditionals and when printed. However, properties can be accessed and called as on normal objects. 

But, in the JavaScript spec, it's more specific and refers to `IsHTMLDDA` slot (where DDA stands for document dot all, as far as I can tell..which is kind of crazy that the JS spec has a special handling for this specific object).

I guess v8 decided to be more generic in case it was needed for something else and went with "undetectable" (although, the v8 source also specifically mentions document.all as the only known use-case).

PR #652, opened 2025-05-16
Introduce a "transfer_arena"

karlseguin:
Some data has to exist specifically for the navigation of one page to another. For example, if a hyperlink is clicked, the URL begins its life with the original page, but is transferred to the new page. The page_arena cannot be used for such data.

It's possible to use the session_arena, but it's lifetime is much longer and, given enough navigation, could accumulate a lot of memory.

The new transfer_arena exists within the session, but only exists until the next navigation.

While currently only used for the navigation URL, the main goal here is to have a place to put the request body on form submission, which has a lifetime similar to a click url.

While I'm at it, I promoted the existing session arena and the new transfer arena to the browser, allowing better memory re-use between sessions.

### DIFF
--- a/src/browser/session.zig
+++ b/src/browser/session.zig
@@ -18,7 +18,7 @@
 
 const std = @import("std");
 
-const ArenaAllocator = std.heap.ArenaAllocator;
+const Allocator = std.mem.Allocator;
 
 const Env = @import("env.zig").Env;
 const Page = @import("page.zig").Page;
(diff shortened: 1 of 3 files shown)


src/browser/session.zig
```diff
@@ -116,12 +126,12 @@ pub const Session = struct {
         // it isn't null!
         std.debug.assert(self.page != null);
 
-        // can't use the page arena, because we're about to reset it
-        // and don't want to use the session's arena, because that'll start to
-        // look like a leak if we navigate from page to page a lot.
-        var buf: [2048]u8 = undefined;
-        var fba = std.heap.FixedBufferAllocator.init(&buf);
-        const url = try self.page.?.url.resolve(fba.allocator(), url_string);
+        _ = self.browser.transfer_arena.reset(.{ .retain_with_limit = 1 * 1024 * 1024 });
```
krichprollsch: It surprises me the `transfer_arena` is reset at the beginning of the page navigate.
What about instead reset the transfer_arena after the end of `pageNavigate`? The data must live longer?

If we reset it at the beginning of the `pageNavigate` it must be explicit we shouldn't use it outside the function...

karlseguin: I wasn't thinking. I originally had the reset _inside_ `page.navigate`, but I was happier with the `page` knowing nothing about it. Don't know why I didn't think of just defer inside the function.

krichprollsch (approved):

PR #656, opened 2025-05-16
module: report module's evaluation error

### LINKED ISSUE #654: module execution fails silently
When a modules contains a JS error, it isn't correctly reported by the browser and fails silently.

Example with the following HTML:
```html
<!DOCTYPE html>
<html><head></head><body>
    <script type="module">
        console.log("in");
        const a = undefined;
        let b = a.foo;
        console.log("out");
    </script>
</body></html>
```

**expected result**

```
$ zig build run -- fetch  http://127.0.0.1:1234/error.html
info(page): GET http://127.0.0.1:1234/error.html 200
info(console): in
info(page): eval script inline: TypeError: Cannot read properties of undefined (reading 'foo')
```

**current result**
```
$ zig build run -- fetch  http://127.0.0.1:1234/error.html
info(page): GET http://127.0.0.1:1234/error.html 200
info(console): in
```

krichprollsch:
depends on https://github.com/lightpanda-io/zig-v8-fork/pull/71

fix #654

### DIFF
--- a/src/runtime/js.zig
+++ b/src/runtime/js.zig
@@ -1147,7 +1147,7 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
             scope: *const Scope,
 
             // the caller needs to deinit the string returned
-            pub fn exception(self: Exception, allocator: Allocator) !?[]const u8 {
+            pub fn exception(self: Exception, allocator: Allocator) ![]const u8 {
                 const scope = self.scope;
                 return try valueToString(allocator, self.inner, scope.isolate, scope.context);
             }
--- a/src/runtime/js.zig
+++ b/src/runtime/js.zig
@@ -610,13 +610,12 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
 
             // compile and eval a JS module
             // It doesn't wait for callbacks execution
-            pub fn module(self: *Scope, src: []const u8, name: []const u8) !Value {
+            pub fn module(self: *Scope, src: []const u8, name: []const u8) !union(enum) { value: Value, exception: Exception } {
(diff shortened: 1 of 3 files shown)


krichprollsch:
```$ zig build run -- fetch   --dump  http://127.0.0.1:1234/error.html
info(page): GET http://127.0.0.1:1234/error.html 200
info(console): in
info(page): eval module inline: TypeError: Cannot read properties of undefined (reading 'foo')
warning(page): evaljs: error.JsErr
<!DOCTYPE html>
<html><head></head><body>
    <script type="module">
        console.log("in");
        const a = undefined;
        let b = a.foo;
        console.log("out");
    </script>

</body></html>
```

Having the stack trace seems more challenging :thinking: 
It works differently than the `try catch`.

src/runtime/js.zig
```diff
@@ -1133,6 +1142,17 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
 
         pub const RemoteObject = v8.RemoteObject;
 
+        pub const Exception = struct {
+            inner: v8.Value,
+            scope: *const Scope,
+
+            // the caller needs to deinit the string returned
+            pub fn exception(self: Exception, allocator: Allocator) !?[]const u8 {
```
karlseguin: Don't think this can actually return null.

karlseguin:
FWIW, I spent at least an hour yesterday trying to figure out why DDG  scripts were silently failing/aborting. I figured they had some global try/catch, or a window.onerror hook or something else that prevent the error from bubbling out. 

I tried this branch, and I'm seeing an `eval module` now, so that should help.

PR #660, opened 2025-05-19
implementation: remove the setTitle method call

krichprollsch:
Libdom uses the doc's body and title attributes by default. But it fallback to the DOM tree if the attributes are NULL.

I think it's better to have only the DOM tree set on document creation.

### DIFF
--- a/src/browser/netsurf.zig
+++ b/src/browser/netsurf.zig
@@ -1912,7 +1912,6 @@ pub inline fn domImplementationCreateHTMLDocument(title: ?[]const u8) !*Document
     _ = try nodeAppendChild(elementToNode(html), elementToNode(head));
 
     if (title) |t| {
-        try documentHTMLSetTitle(doc_html, t);
         const htitle = try documentCreateElement(doc, "title");
         const txt = try documentCreateTextNode(doc, t);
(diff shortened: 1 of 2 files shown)


src/browser/netsurf.zig
```diff
@@ -1912,7 +1912,6 @@ pub inline fn domImplementationCreateHTMLDocument(title: ?[]const u8) !*Document
     _ = try nodeAppendChild(elementToNode(html), elementToNode(head));
 
     if (title) |t| {
-        try documentHTMLSetTitle(doc_html, t);
```
sjorsdonkers, line 1915: Seems it would be more performant if we do set it as it would not have to search for it [here](https://github.com/lightpanda-io/libdom/blob/b0f4a4314711ed9a91bf3f40f0e0354b8f4706f2/src/html/html_document.c#L1060-L1093).
If we keep it, it would be good to comment why it is there.

krichprollsch: `documentCreateDocument` calls already `documentHTMLSetTitle`. But moving this tree creation to `documentCreateDocument` will be clearer.

sjorsdonkers (approved):

PR #664, opened 2025-05-19
Add TreeWalker

mookums:
This introduces the `TreeWalker` and related `NodeFilter`.


src/browser/dom/document.zig
```diff
@@ -238,6 +241,10 @@ pub const Document = struct {
     pub fn _replaceChildren(self: *parser.Document, nodes: []const Node.NodeOrText) !void {
         return Node.replaceChildren(parser.documentToNode(self), nodes);
     }
+
+    pub fn _createTreeWalker(_: *parser.Document, root: *parser.Node, whatToShow: ?u32, filter: ?Env.Callback) TreeWalker {
```
karlseguin:
```suggestion
    pub fn _createTreeWalker(_: *parser.Document, root: *parser.Node, what_to_show: ?u32, filter: ?Env.Callback) TreeWalker {
```

src/browser/dom/tree_walker.zig
```diff
@@ -0,0 +1,93 @@
...
+const Env = @import("../env.zig").Env;
+
+// https://developer.mozilla.org/en-US/docs/Web/API/TreeWalker
+pub const TreeWalker = struct {
+    root: *parser.Node,
+    currentNode: *parser.Node,
+    whatToShow: u32,
+    filter: ?Env.Callback,
+
+    pub fn init(node: *parser.Node, whatToShow: ?u32, filter: ?Env.Callback) TreeWalker {
```
karlseguin:
```suggestion
    pub fn init(node: *parser.Node, what_to_show: ?u32, filter: ?Env.Callback) TreeWalker {
```

src/browser/dom/tree_walker.zig
```diff
@@ -0,0 +1,93 @@
...
+const std = @import("std");
+const parser = @import("../netsurf.zig");
+
+const Env = @import("../env.zig").Env;
+
+// https://developer.mozilla.org/en-US/docs/Web/API/TreeWalker
+pub const TreeWalker = struct {
+    root: *parser.Node,
+    currentNode: *parser.Node,
+    whatToShow: u32,
```
karlseguin:
```suggestion
    what_to_show: u32,
```

src/browser/dom/tree_walker.zig
```diff
@@ -0,0 +1,93 @@
...
+
+const std = @import("std");
+const parser = @import("../netsurf.zig");
+
+const Env = @import("../env.zig").Env;
+
+// https://developer.mozilla.org/en-US/docs/Web/API/TreeWalker
+pub const TreeWalker = struct {
+    root: *parser.Node,
+    currentNode: *parser.Node,
```
karlseguin:
```suggestion
    current_node: *parser.Node,
```

src/browser/dom/tree_walker.zig
```diff
@@ -0,0 +1,93 @@
...
+    whatToShow: u32,
+    filter: ?Env.Callback,
+
+    pub fn init(node: *parser.Node, whatToShow: ?u32, filter: ?Env.Callback) TreeWalker {
+        return .{
+            .root = node,
+            .currentNode = node,
+            .whatToShow = whatToShow orelse std.math.maxInt(u32),
```
karlseguin: Would improve readability to use `NodeFilter._SHOW_ALL`, or at least a comment saying the orelse is show all.

karlseguin (commented):
This looks ok, but it seems like the implementation is missing a lot, e.g. the `filter` and `whatToShow` fields aren't being used. That's fine, but I just wanted to make sure that was the intention.

src/browser/dom/tree_walker.zig
```diff
@@ -0,0 +1,93 @@
...
+    pub fn get_whatToShow(self: *TreeWalker) u32 {
+        return self.whatToShow;
+    }
+
+    pub fn get_filter(self: *TreeWalker) ?Env.Callback {
+        return self.filter;
+    }
+
+    pub fn _firstChild(self: *TreeWalker) ?*parser.Node {
+        const first_child = parser.nodeFirstChild(self.currentNode) catch return null;
```
sjorsdonkers: Hmm `dom_node_get_first_child` increments the ref_count, walker.zig says we should be deiniting these Nodes.
Though I don't see us doing that in the rest of the code.
Worst case they would be cleaned up when the page is removed.

karlseguin: Because everything is in an mimalloc arena. Wonder if we took out all the refcnt++ and refcnt-- from libdom if we'd shave a few µs.

sjorsdonkers: Yea, may want to remove to comments from walker.zig.
Or actually maybe this PR can replace the whole thing, maybe.

src/browser/dom/tree_walker.zig
```diff
@@ -0,0 +1,187 @@
...
+        return .{
+            .root = node,
+            .current_node = node,
+            .what_to_show = what_to_show orelse NodeFilter._SHOW_ALL,
+            .filter = filter_func,
+            .depth = 0,
+        };
+    }
+
+    fn verify_what_to_show(self: *const TreeWalker, node: *parser.Node) !bool {
```
karlseguin: `verifyWhatToShow`

Function names are `camelCase`, variables are `snake_case` and types are `PascalCase`.

I know with the WebAPI functions this gets a little mixed up, i.e. `get_readyState`.

src/browser/dom/tree_walker.zig
```diff
@@ -0,0 +1,187 @@
...
+    pub fn _lastChild(self: *TreeWalker) !?*parser.Node {
+        const children = try parser.nodeGetChildNodes(self.current_node);
+        const child_count = try parser.nodeListLength(children);
+
+        for (0..child_count) |i| {
+            const index: u32 = @intCast(child_count - 1 - i);
+            const child = (try parser.nodeListItem(children, index)) orelse return null;
+
+            if (!try self.verify_what_to_show(child)) continue;
+            if (!try self.verify_filter(child)) continue;
```
karlseguin: since you always call `verify_filter` and `verify_what_to_show` together, I wonder combining them, or having a helper that calls both is worthwhile.

src/browser/dom/tree_walker.zig
```diff
@@ -0,0 +1,187 @@
...
+        }
+
+        return null;
+    }
+
+    pub fn _lastChild(self: *TreeWalker) !?*parser.Node {
+        const children = try parser.nodeGetChildNodes(self.current_node);
+        const child_count = try parser.nodeListLength(children);
+
+        for (0..child_count) |i| {
```
karlseguin: I was confused why `lastChild` was walking through the list in this order. Now I see the `child_count - 1 - i`.

It's fine as-is, but I think this alternative is the same and might be more explicit:

```zig
var index: u32 = child_count;
while (index > 0) {
   index -= 1
  ...
```

src/browser/dom/tree_walker.zig
```diff
@@ -0,0 +1,187 @@
...
+            self.depth += 1;
+            self.current_node = child;
+            return child;
+        }
+
+        return null;
+    }
+
+    pub fn _nextNode(self: *TreeWalker) !?*parser.Node {
+        return self._firstChild();
```
karlseguin: Is this right? I'd expect it to be relative to `self.current_node` .. that'll make the depth an issue...

src/browser/dom/tree_walker.zig
```diff
@@ -0,0 +1,187 @@
...
+        if (!try self.verify_what_to_show(parent)) return null;
+        if (!try self.verify_filter(parent)) return null;
+
+        self.depth -= 1;
+        self.current_node = parent;
+        return parent;
+    }
+
+    pub fn _previousNode(self: *TreeWalker) !?*parser.Node {
+        return self._parentNode();
```
karlseguin: same as `nextNode`..is it really always the parent? It can't be a previous sibling?

src/browser/dom/tree_walker.zig
```diff
@@ -0,0 +1,187 @@
...
+        const child_count = try parser.nodeListLength(children);
+
+        for (0..child_count) |i| {
+            const index: u32 = @intCast(child_count - 1 - i);
+            const child = (try parser.nodeListItem(children, index)) orelse return null;
+
+            if (!try self.verify_what_to_show(child)) continue;
+            if (!try self.verify_filter(child)) continue;
+
+            self.depth += 1;
```
karlseguin: does every node you visit necessarily increase the depth? I'm thinking if `child` is a sibling of `current_node`, it shouldn't increase the depth.

Maybe I don't understand..but if I'm right, it seems like `depth` is going to get tricky to keep track of. If `depth` is only used in `_parentNode` to protect against going outside the tree, maybe there's a better way (like comparing against the root node?)

src/browser/dom/tree_walker.zig
mookums: Yeah, I collapsed these into a single `verify` function.

src/browser/dom/tree_walker.zig
mookums: Ditched depth in favor of just comparing against the `self.root` when we are moving up the tree.

src/browser/dom/tree_walker.zig
mookums: Fixed this too, moving in document order.

mookums:
Just as a note, some newly passing WPT tests:
`dom/nodes/Document-createTreeWalker.html`
`dom/traversal/TreeWalker-traversal-skip-most.html`
`dom/traversal/TreeWalker-traversal-reject.html`
`dom/traversal/TreeWalker-previousNodeLastChildReject.html`
`dom/traversal/TreeWalker-walking-outside-a-tree.html`
`dom/traversal/TreeWalker-traversal-skip.html`

karlseguin:
LGTM

PR #665, opened 2025-05-20
Tweak debug logging

karlseguin:
1 - Add a log_level build option to control the default log level from
    the build (e.g. -Dlog_level=debug). Defaults to info

2 - Add a new boolean log_unknown_properties build option to enable
    logging unknown properties. Defautls to false.

3 - Remove the log debug for script eval - this can be a huge value
    (i.e. hundreds of KB), which makes the debug log unusable IMO.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -44,6 +44,18 @@ pub fn build(b: *std.Build) !void {
         b.option([]const u8, "git_commit", "Current git commit") orelse "dev",
     );
 
+    opts.addOption(
+        std.log.Level,
+        "log_level",
+        b.option(std.log.Level, "log_level", "The log level") orelse std.log.Level.info,
+    );
+
+    opts.addOption(
+        bool,
+        "log_unknown_properties",
+        b.option(bool, "log_unknown_properties", "Log access to unknown properties") orelse false,
+    );
+
     const target = b.standardTargetOptions(.{});
     const optimize = b.standardOptimizeOption(.{});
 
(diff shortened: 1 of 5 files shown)


build.zig
```diff
@@ -44,6 +44,18 @@ pub fn build(b: *std.Build) !void {
         b.option([]const u8, "git_commit", "Current git commit") orelse "dev",
     );
 
+    opts.addOption(
+        std.log.Level,
+        "log_level",
+        b.option(std.log.Level, "log_level", "The log level") orelse std.log.Level.info,
```
sjorsdonkers, line 50: What about defaulting to .debug if build.mode == debug?

karlseguin: I find the noise to signal ratio is too low right now. The point of the flag is to let everyone set it to whatever they want, without having to touch the source. So the default can be anything.

krichprollsch:
What about adding a `make run-debug` command to active `-Dlog_level=debug` and `-Dlog_unknown_properties=true`?

EDIT: maybe it's `make build-dev` we should update since it's a "Build in debug mode"?

karlseguin:
Well, you were right...depending on the page, `-Dlog_unknown_properties=true` can be a bit overwhelming. It might be better with the incoming CSS stuff, but right now, I think it's only useful as a one-off in really mysterious situations.

But I'll add a `make run-debug` and change `make build-dev` to add the `-Dlog_level=debug`  then have 2 `debug` commands that we can tweak the parameters for as things progress.

PR #666, opened 2025-05-20
Playwright support disclaimer

sjorsdonkers:
Adding a disclaimer on Playwright to make sure user's are aware that scripts may break as correctly implemented features get added to the browser.


README.md
```diff
@@ -36,6 +36,9 @@ Fast web automation for AI agents, LLM training, scraping and testing:
 _Puppeteer requesting 100 pages from a local website on a AWS EC2 m5.large instance.
 See [benchmark details](https://github.com/lightpanda-io/demo)._
 
+> **Playwright support disclaimer:**  
+> Due to the nature of Playwright, a script that works with the current version of the browser may not function correctly with a future version. Playwright uses an intermediate JavaScript layer that selects an execution strategy based on the browser's available features. If LightPanda adds a new [Web API](https://developer.mozilla.org/en-US/docs/Web/API), Playwright may choose to execute different code for the same script. This new code path could attempt to use features that are not yet implemented. Puppeteer has proven to be more resilient to changes in the browser.
```
krichprollsch: Some comments:
* why do you use the blockquote `>` char? It's intentional?
* I would like to remove "Puppeteer has proven to be more resilient to changes in the browser." part. I'm not sure it's really helpful.
* I would add 2 things: 1 saying we make the effort to add compatibility tests, but we can't cover all scenarios + asking user to create issue giving a previous version which was working.

sjorsdonkers: - I made it a footnote, so it is now linked to the bottom of the page.
image
- Removed puppeteer
- Added 2 things

README.md
```diff
@@ -36,6 +36,9 @@ Fast web automation for AI agents, LLM training, scraping and testing:
 _Puppeteer requesting 100 pages from a local website on a AWS EC2 m5.large instance.
 See [benchmark details](https://github.com/lightpanda-io/demo)._
 
+[^1]: **Playwright support disclaimer:**  
+Due to the nature of Playwright, a script that works with the current version of the browser may not function correctly with a future version. Playwright uses an intermediate JavaScript layer that selects an execution strategy based on the browser's available features. If Lightpanda adds a new [Web API](https://developer.mozilla.org/en-US/docs/Web/API), Playwright may choose to execute different code for the same script. This new code path could attempt to use features that are not yet implemented. Lightpanda makes an effort to add compatibility tests, but we can't cover all scenarios. If you encounter an issue, please create[ a GitHub issue](https://github.com/lightpanda-io/browser/issues) and include the last known working version of the script.
```
```suggestion
Due to the nature of Playwright, a script that works with the current version of the browser may not function correctly with a future version. Playwright uses an intermediate JavaScript layer that selects an execution strategy based on the browser's available features. If Lightpanda adds a new [Web API](https://developer.mozilla.org/en-US/docs/Web/API), Playwright may choose to execute different code for the same script. This new code path could attempt to use features that are not yet implemented. Lightpanda makes an effort to add compatibility tests, but we can't cover all scenarios. If you encounter an issue, please create a [GitHub issue](https://github.com/lightpanda-io/browser/issues) and include the last known working version of the script.
```

PR #668, opened 2025-05-20
Load aboutblank doc

sjorsdonkers:
Replaces PR:https://github.com/lightpanda-io/browser/pull/630

Loads the about:blank window / document when creating a target. Scripts evaluated before navigating to a page will now have access to a #document.


sjorsdonkers:
With a similar change this should make it possible for document to not be optional.
Most likely will delete this PR in favor of: https://github.com/lightpanda-io/browser/pull/670

PR #669, opened 2025-05-20
FormData constructor form & submitter parameter

karlseguin:
FormData takes two optional parameters: a form and a submitter.

Building the FormData from these is a first step in supporting form submission.

Basic extension of the HTMLForm element. There was more work done on the Select web api, because the netsurf implementation isn't great. But all of the input elements will need to have their web api extended.

### DIFF
--- a/src/browser/xhr/form_data.zig
+++ b/src/browser/xhr/form_data.zig
@@ -221,8 +221,8 @@ fn collectForm(arena: Allocator, form: *parser.Form, submitter_: ?*parser.Elemen
         const tag = try parser.elementHTMLGetTagType(@as(*parser.ElementHTML, @ptrCast(element)));
         switch (tag) {
             .input => {
-                const input_type = try parser.elementGetAttribute(element, "type") orelse "";
-                if (std.ascii.eqlIgnoreCase(input_type, "image")) {
+                const tpe = try parser.elementGetAttribute(element, "type") orelse "";
+                if (std.ascii.eqlIgnoreCase(tpe, "image")) {
                     if (submitter_name_) |submitter_name| {
                         if (std.mem.eql(u8, submitter_name, name)) {
                             try entries.append(arena, .{
@@ -237,8 +237,14 @@ fn collectForm(arena: Allocator, form: *parser.Form, submitter_: ?*parser.Elemen
                     }
                     continue;
                 }
-                if (std.ascii.eqlIgnoreCase(input_type, "checkbox") or std.ascii.eqlIgnoreCase(input_type, "radio")) {
+
+                if (std.ascii.eqlIgnoreCase(tpe, "checkbox") or std.ascii.eqlIgnoreCase(tpe, "radio")) {
                     if (try parser.inputGetChecked(@ptrCast(element)) == false) {
+                        continue;
+                    }
+                }
+                if (std.ascii.eqlIgnoreCase(tpe, "submit")) {
+                    if (submitter_name_ == null or !std.mem.eql(u8, submitter_name_.?, name)) {
                         continue;
                     }
                 }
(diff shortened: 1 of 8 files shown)


src/browser/xhr/form_data.zig
```diff
@@ -186,9 +199,167 @@ const EntryIterator = struct {
...
+fn collectSelectValues(arena: Allocator, select: *parser.Select, name: []const u8, entries: *std.ArrayListUnmanaged(Entry), state: *SessionState) !void {
+    const HTMLSelectElement = @import("../html/select.zig").HTMLSelectElement;
+
+    // Go through the HTMLSelectElement because it has specific logic for handling
+    // the default selected option, which libdom doesn't properly handle
+    const selected_index = try HTMLSelectElement.get_selectedIndex(select, state);
+    if (selected_index == -1) {
+        return;
+    }
+    std.debug.assert(selected_index >= 0);
```
krichprollsch, line 288: To remove?

src/browser/xhr/form_data.zig
```diff
@@ -186,9 +199,167 @@ const EntryIterator = struct {
...
+    if (is_multiple == false) {
+        const option = try parser.optionCollectionItem(options, @intCast(selected_index));
+        const value = try parser.optionGetValue(option);
+        return entries.append(arena, .{ .key = name, .value = value });
+    }
+
+    const len = try parser.optionCollectionGetLength(options);
+
+    // we can go directly to the first one
+    for (@intCast(selected_index)..len) |i| {
```
krichprollsch, line 305: We are sure selected_index is the first one? 🤔

src/browser/xhr/form_data.zig
```diff
@@ -186,9 +199,167 @@ const EntryIterator = struct {
...
+        const option = try parser.optionCollectionItem(options, @intCast(i));
+        if (try parser.elementGetAttribute(@ptrCast(option), "disabled") != null) {
```
krichprollsch, line 307: You don't check disable attr in the single select case. Is it intentional?

src/browser/xhr/form_data.zig
```diff
@@ -186,9 +199,167 @@ const EntryIterator = struct {
...
+
+    const tag = try parser.elementHTMLGetTagType(submitter);
+    const element: *parser.Element = @ptrCast(submitter);
+    const name = try parser.elementGetAttribute(element, "name");
+
+    switch (tag) {
+        .button => return name,
+        .input => {
+            const tpe = (try parser.elementGetAttribute(element, "type")) orelse "";
+            // only an image type can be a sumbitter
```
krichprollsch, line 329: What about input type=submit?

https://developer.mozilla.org/en/docs/Web/HTML/Reference/Elements/input/submit

src/browser/xhr/form_data.zig
karlseguin: I think it's useful to have both to make sure nothing unexpected ever happens and as documentation. JavaScript code could set it to anything, i.e. `sel.selectedIndex = -3232`, but it should always revert to -1.

src/browser/xhr/form_data.zig
karlseguin: Yes, both by the spec "Returns the index of the **first** selected item", and based on libdom's implementation which loops through the collection and returns the first match:

https://github.com/lightpanda-io/libdom/blob/master/src/html/html_select_element.c#L244

src/browser/xhr/form_data.zig
krichprollsch: I read to fast, my bad :facepalm: :+1:

src/browser/xhr/form_data.zig
karlseguin: Good catch, fixed in https://github.com/lightpanda-io/browser/pull/669/commits/72944a4e5e1cac4ba6ad6b3ffa34a4ed781bc402

src/browser/xhr/form_data.zig
karlseguin: Nope, I missed it. I checked the browser, if the default item (the first one) is disabled, then it sends nothing. Fixed in https://github.com/lightpanda-io/browser/pull/669/commits/72944a4e5e1cac4ba6ad6b3ffa34a4ed781bc402

PR #670, opened 2025-05-20
Not optional document

sjorsdonkers:
Builds on loading document for about:blank: https://github.com/lightpanda-io/browser/pull/668

Since half the code of that PR is removed again, might as well just review this one.
This make the window.document not optional, and sets it to about:blank by default.

### DIFF
--- a/src/browser/html/window.zig
+++ b/src/browser/html/window.zig
@@ -44,7 +44,7 @@ pub const Window = struct {
     // Extend libdom event target for pure zig struct.
     base: parser.EventTargetTBase = parser.EventTargetTBase{},
 
-    document: ?*parser.DocumentHTML = null,
+    document: *parser.DocumentHTML,
     target: []const u8 = "",
     history: History = .{},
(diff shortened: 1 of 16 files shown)


src/browser/html/window.zig
```diff
@@ -60,7 +60,13 @@ pub const Window = struct {
     performance: Performance,
 
     pub fn create(target: ?[]const u8, navigator: ?Navigator) !Window {
+        var fbs = std.io.fixedBufferStream("");
+        const html_doc = try parser.documentHTMLParse(fbs.reader(), "utf-8");
+        const doc = parser.documentHTMLToDocument(html_doc);
+        try parser.documentSetDocumentURI(doc, "about:blank");
+
```
sjorsdonkers, line 67: Should we `setStorageShelf` as well?

karlseguin: I think it's fine to leave null, for now.

karlseguin:
LGTM

PR #671, opened 2025-05-20
Allow webapis to register a destructor to do cleanup on scope (page) end

karlseguin:
Add destructor to XHR to abort any inflight requests.

### DIFF
--- a/src/runtime/js.zig
+++ b/src/runtime/js.zig
@@ -539,8 +539,15 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
             // no init, started with executor.startScope()
 
             fn deinit(self: *Scope) void {
-                for (self.destructor_callbacks.items) |cb| {
-                    cb.destructor(self);
+                {
+                    // reverse order, as this has more chance of respecting any
+                    // dependencies objects might have with each other.
+                    const items = self.destructor_callbacks.items;
+                    var i = items.len;
+                    while (i > 0) {
+                        i -= 1;
+                        items[i].destructor();
+                    }
                 }
 
                 {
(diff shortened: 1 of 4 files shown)


src/runtime/js.zig
```diff
@@ -536,6 +539,10 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
             // no init, started with executor.startScope()
 
             fn deinit(self: *Scope) void {
+                for (self.destructor_callbacks.items) |cb| {
```
sjorsdonkers: I think we should call these in the reverse order of which the where constructed. Since later objects may depend on objects created before it.
Worst case it doesn't matter, best case it prevent problems.

src/runtime/js.zig
```diff
@@ -1672,7 +1683,35 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
...
+            destructorFn: *const fn (ptr: *anyopaque, scope: *Scope) void,
+
+            fn init(ptr: anytype) DestructorCallback {
+                const T = @TypeOf(ptr);
+                const ptr_info = @typeInfo(T);
+
+                const gen = struct {
+                    pub fn destructor(pointer: *anyopaque, scope: *Scope) void {
+                        const self: T = @ptrCast(@alignCast(pointer));
+                        return ptr_info.pointer.child.destructor(self, scope);
```
sjorsdonkers: Q: Why are we passing the scope? In case we own persistent objects?

karlseguin: Dunno, when I build jsScopeEnd I thought they might need access to the arena.

https://github.com/lightpanda-io/browser/pull/671/commits/ce832a8063733b4df916ad1b7602a73051531966  takes it out from both `destructor` and `jsScopeEnd`. We can add the *SessionState in the future, if something does need it.

src/runtime/js.zig
karlseguin: Good idea. Done.

karlseguin:
I re-worked this PR to focus only on the destructor callback hook. Nothing is currently using it. A separate commit will add the destructor to XHR...it requires more changes than I was expecting.

sjorsdonkers (approved):
Nice

PR #677, opened 2025-05-21
Move jsValueToZig from Caller to the Scope

karlseguin:
Caller is a transient object that exists only for calling Zig functions from JS. But jsValueToZig is more generally useful and can be used outside of an explicit JS call. The scope is a better place for these as it's generally referenced already by any code that would need to map values (i.e. a Callback).

### DIFF
--- a/src/runtime/js.zig
+++ b/src/runtime/js.zig
@@ -775,6 +775,403 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
                 }
             }
 
+            fn jsValueToZig(self: *Scope, comptime named_function: NamedFunction, comptime T: type, js_value: v8.Value) !T {
+                switch (@typeInfo(T)) {
+                    .optional => |o| {
+                        if (js_value.isNullOrUndefined()) {
+                            return null;
+                        }
+                        return try self.jsValueToZig(named_function, o.child, js_value);
+                    },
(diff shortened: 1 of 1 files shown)


src/runtime/js.zig
```diff
@@ -775,6 +775,403 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
                 }
             }
 
+            fn jsValueToZig(self: *Scope, comptime named_function: NamedFunction, comptime T: type, js_value: v8.Value) !T {
```
sjorsdonkers, line 778: For consideration, IIRC this and related function can all be free functions as they are not guarding any kind of invariant. As also documented by self previously being const. It could reduce coupling as the functions contracts can be narrowed down to the params they really depend on.

karlseguin: I can move out the simpler ones (ints), but there's a dependency on `TypeLookup`, which is essentially what Env is a generic over.

karlseguin: oh..I already did the ints

sjorsdonkers: Alright then.
Not having generic types is still the most surprising language design choice 🤷

sjorsdonkers (approved):
LGTM

PR #680, opened 2025-05-22
CSSStyleDeclaration implementation

RaphUn:
Added also a css parser and a css value analyzer util

### DIFF
--- a/src/browser/cssom/css_parser.zig
+++ b/src/browser/cssom/css_parser.zig
@@ -1,0 +1,308 @@
+const std = @import("std");
+
+const CSSConstants = struct {
+    const IMPORTANT_KEYWORD = "!important";
+    const IMPORTANT_LENGTH = IMPORTANT_KEYWORD.len;
+    const URL_PREFIX = "url(";
+    const URL_PREFIX_LENGTH = URL_PREFIX.len;
+};
+
+pub const CSSParserState = enum {
+    seekName,
+    inName,
+    seekColon,
+    seekValue,
+    inValue,
+    inQuotedValue,
+    inSingleQuotedValue,
+    inUrl,
+    inImportant,
+};
+
+pub const CSSDeclaration = struct {
+    name: []const u8,
+    value: []const u8,
+    is_important: bool,
+
+    pub fn init(name: []const u8, value: []const u8, is_important: bool) CSSDeclaration {
+        return .{
+            .name = name,
+            .value = value,
+            .is_important = is_important,
+        };
+    }
+};
+
+pub const CSSParser = struct {
+    state: CSSParserState,
+    name_start: usize,
+    name_end: usize,
+    value_start: usize,
+    position: usize,
+    paren_depth: usize,
+    escape_next: bool,
+
+    pub fn init() CSSParser {
+        return .{
+            .state = .seekName,
+            .name_start = 0,
+            .name_end = 0,
+            .value_start = 0,
+            .position = 0,
+            .paren_depth = 0,
+            .escape_next = false,
+        };
+    }
+
+    pub fn parseDeclarations(text: []const u8, allocator: std.mem.Allocator) ![]CSSDeclaration {
+        var parser = init();
+        var declarations = std.ArrayList(CSSDeclaration).init(allocator);
+        errdefer declarations.deinit();
+
+        while (parser.position < text.len) {
+            const c = text[parser.position];
+
+            switch (parser.state) {
+                .seekName => {
+                    if (!std.ascii.isWhitespace(c)) {
+                        parser.name_start = parser.position;
+                        parser.state = .inName;
+                        continue;
+                    }
+                },
+                .inName => {
+                    if (c == ':') {
+                        parser.name_end = parser.position;
+                        parser.state = .seekValue;
+                    } else if (std.ascii.isWhitespace(c)) {
+                        parser.name_end = parser.position;
+                        parser.state = .seekColon;
+                    }
+                },
+                .seekColon => {
+                    if (c == ':') {
+                        parser.state = .seekValue;
+                    } else if (!std.ascii.isWhitespace(c)) {
+                        parser.state = .seekName;
+                        continue;
+                    }
+                },
+                .seekValue => {
+                    if (!std.ascii.isWhitespace(c)) {
+                        parser.value_start = parser.position;
+                        if (c == '"') {
+                            parser.state = .inQuotedValue;
+                        } else if (c == '\'') {
+                            parser.state = .inSingleQuotedValue;
+                        } else if (c == 'u' and parser.position + 3 < text.len and
+                            std.mem.eql(u8, text[parser.position .. parser.position + 4], CSSConstants.URL_PREFIX))
+                        {
+                            parser.state = .inUrl;
+                            parser.paren_depth = 1;
+                            parser.position += 3;
+                        } else {
+                            parser.state = .inValue;
+                            continue;
+                        }
+                    }
+                },
+                .inValue => {
+                    if (parser.escape_next) {
+                        parser.escape_next = false;
+                    } else if (c == '\\') {
+                        parser.escape_next = true;
+                    } else if (c == '(') {
+                        parser.paren_depth += 1;
+                    } else if (c == ')' and parser.paren_depth > 0) {
+                        parser.paren_depth -= 1;
+                    } else if (c == ';' and parser.paren_depth == 0) {
+                        try parser.finishDeclaration(text, &declarations);
+                        parser.state = .seekName;
+                    }
+                },
+                .inQuotedValue => {
+                    if (parser.escape_next) {
+                        parser.escape_next = false;
+                    } else if (c == '\\') {
+                        parser.escape_next = true;
+                    } else if (c == '"') {
+                        parser.state = .inValue;
+                    }
+                },
+                .inSingleQuotedValue => {
+                    if (parser.escape_next) {
+                        parser.escape_next = false;
+                    } else if (c == '\\') {
+                        parser.escape_next = true;
+                    } else if (c == '\'') {
+                        parser.state = .inValue;
+                    }
+                },
+                .inUrl => {
+                    if (parser.escape_next) {
+                        parser.escape_next = false;
+                    } else if (c == '\\') {
+                        parser.escape_next = true;
+                    } else if (c == '(') {
+                        parser.paren_depth += 1;
+                    } else if (c == ')') {
+                        parser.paren_depth -= 1;
+                        if (parser.paren_depth == 0) {
+                            parser.state = .inValue;
+                        }
+                    }
+                },
+                .inImportant => {},
+            }
+
+            parser.position += 1;
+        }
+
+        try parser.finalize(text, &declarations);
+
+        return declarations.toOwnedSlice();
+    }
+
+    fn finishDeclaration(self: *CSSParser, text: []const u8, declarations: *std.ArrayList(CSSDeclaration)) !void {
+        const name = std.mem.trim(u8, text[self.name_start..self.name_end], &std.ascii.whitespace);
+        if (name.len == 0) return;
+
+        const raw_value = text[self.value_start..self.position];
+        const value = std.mem.trim(u8, raw_value, &std.ascii.whitespace);
+
+        var final_value = value;
+        var is_important = false;
+
+        if (std.mem.endsWith(u8, value, CSSConstants.IMPORTANT_KEYWORD)) {
+            is_important = true;
+            final_value = std.mem.trim(u8, value[0 .. value.len - CSSConstants.IMPORTANT_LENGTH], &std.ascii.whitespace);
+        }
+
+        const declaration = CSSDeclaration.init(name, final_value, is_important);
+        try declarations.append(declaration);
+    }
+
+    fn finalize(self: *CSSParser, text: []const u8, declarations: *std.ArrayList(CSSDeclaration)) !void {
+        if (self.state == .inValue) {
+            const name = text[self.name_start..self.name_end];
+            const trimmed_name = std.mem.trim(u8, name, &std.ascii.whitespace);
+
+            if (trimmed_name.len > 0) {
+                const raw_value = text[self.value_start..self.position];
+                const value = std.mem.trim(u8, raw_value, &std.ascii.whitespace);
+
+                var final_value = value;
+                var is_important = false;
+                if (std.mem.endsWith(u8, value, CSSConstants.IMPORTANT_KEYWORD)) {
+                    is_important = true;
+                    final_value = std.mem.trim(u8, value[0 .. value.len - CSSConstants.IMPORTANT_LENGTH], &std.ascii.whitespace);
+                }
+
+                const declaration = CSSDeclaration.init(trimmed_name, final_value, is_important);
+                try declarations.append(declaration);
+            }
+        }
+    }
+
+    pub fn getState(self: *const CSSParser) CSSParserState {
+        return self.state;
+    }
+
+    pub fn getPosition(self: *const CSSParser) usize {
+        return self.position;
+    }
+
+    pub fn reset(self: *CSSParser) void {
+        self.* = init();
+    }
+};
+
+const testing = std.testing;
+
+test "CSSParser - Simple property" {
+    const text = "color: red;";
+    const allocator = testing.allocator;
+
+    const declarations = try CSSParser.parseDeclarations(text, allocator);
+    defer allocator.free(declarations);
+
+    try testing.expect(declarations.len == 1);
+    try testing.expectEqualStrings("color", declarations[0].name);
+    try testing.expectEqualStrings("red", declarations[0].value);
+    try testing.expect(!declarations[0].is_important);
+}
+
+test "CSSParser - Property with !important" {
+    const text = "margin: 10px !important;";
+    const allocator = testing.allocator;
+
+    const declarations = try CSSParser.parseDeclarations(text, allocator);
+    defer allocator.free(declarations);
+
+    try testing.expect(declarations.len == 1);
+    try testing.expectEqualStrings("margin", declarations[0].name);
+    try testing.expectEqualStrings("10px", declarations[0].value);
+    try testing.expect(declarations[0].is_important);
+}
+
+test "CSSParser - Multiple properties" {
+    const text = "color: red; font-size: 12px; margin: 5px !important;";
+    const allocator = testing.allocator;
+
+    const declarations = try CSSParser.parseDeclarations(text, allocator);
+    defer allocator.free(declarations);
+
+    try testing.expect(declarations.len == 3);
+
+    try testing.expectEqualStrings("color", declarations[0].name);
+    try testing.expectEqualStrings("red", declarations[0].value);
+    try testing.expect(!declarations[0].is_important);
+
+    try testing.expectEqualStrings("font-size", declarations[1].name);
+    try testing.expectEqualStrings("12px", declarations[1].value);
+    try testing.expect(!declarations[1].is_important);
+
+    try testing.expectEqualStrings("margin", declarations[2].name);
+    try testing.expectEqualStrings("5px", declarations[2].value);
+    try testing.expect(declarations[2].is_important);
+}
+
+test "CSSParser - Quoted value with semicolon" {
+    const text = "content: \"Hello; world!\";";
+    const allocator = testing.allocator;
+
+    const declarations = try CSSParser.parseDeclarations(text, allocator);
+    defer allocator.free(declarations);
+
+    try testing.expect(declarations.len == 1);
+    try testing.expectEqualStrings("content", declarations[0].name);
+    try testing.expectEqualStrings("\"Hello; world!\"", declarations[0].value);
+    try testing.expect(!declarations[0].is_important);
+}
+
+test "CSSParser - URL value" {
+    const text = "background-image: url(\"test.png\");";
+    const allocator = testing.allocator;
+
+    const declarations = try CSSParser.parseDeclarations(text, allocator);
+    defer allocator.free(declarations);
+
+    try testing.expect(declarations.len == 1);
+    try testing.expectEqualStrings("background-image", declarations[0].name);
+    try testing.expectEqualStrings("url(\"test.png\")", declarations[0].value);
+    try testing.expect(!declarations[0].is_important);
+}
+
+test "CSSParser - Whitespace handling" {
+    const text = "  color  :  purple  ;  margin  :  10px  ;  ";
+    const allocator = testing.allocator;
+
+    const declarations = try CSSParser.parseDeclarations(text, allocator);
+    defer allocator.free(declarations);
+
+    try testing.expect(declarations.len == 2);
+    try testing.expectEqualStrings("color", declarations[0].name);
+    try testing.expectEqualStrings("purple", declarations[0].value);
+    try testing.expectEqualStrings("margin", declarations[1].name);
+    try testing.expectEqualStrings("10px", declarations[1].value);
+}
--- a/src/browser/cssom/css_style_declaration.zig
+++ b/src/browser/cssom/css_style_declaration.zig
@@ -1,0 +1,236 @@
+// Copyright (C) 2023-2024  Lightpanda (Selecy SAS)
+//
+// Francis Bouvier <francis@lightpanda.io>
+// Pierre Tachoire <pierre@lightpanda.io>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+const std = @import("std");
+
+const CSSParser = @import("./css_parser.zig").CSSParser;
+const CSSValueAnalyzer = @import("./css_value_analyzer.zig").CSSValueAnalyzer;
+const SessionState = @import("../env.zig").SessionState;
+
+pub const Interfaces = .{
+    CSSStyleDeclaration,
+    CSSRule,
+};
+
+const CSSRule = struct {};
+
+pub const CSSStyleDeclaration = struct {
+    store: std.StringHashMapUnmanaged(Property),
+    order: std.ArrayListUnmanaged([]const u8),
+
+    const Property = struct {
+        value: []const u8,
+        priority: bool,
+    };
+
+    pub fn get_cssFloat(self: *const CSSStyleDeclaration) []const u8 {
+        return self._getPropertyValue("float");
+    }
+
+    pub fn set_cssFloat(self: *CSSStyleDeclaration, value: ?[]const u8, state: *SessionState) !void {
+        const final_value = value orelse "";
+        return self._setProperty("float", final_value, null, state);
+    }
+
+    pub fn get_cssText(self: *const CSSStyleDeclaration, state: *SessionState) ![]const u8 {
+        var buffer = std.ArrayList(u8).init(state.arena);
+        const writer = buffer.writer();
+        for (self.order.items) |name| {
+            const prop = self.store.get(name).?;
+            const escaped = try CSSValueAnalyzer.escapeCSSValue(state.arena, prop.value);
+            try writer.print("{s}: {s}", .{ name, escaped });
+            if (prop.priority) try writer.writeAll(" !important");
+            try writer.writeAll("; ");
+        }
+        return buffer.toOwnedSlice();
+    }
+
+    // TODO Propagate also upward to parent node
+    pub fn set_cssText(self: *CSSStyleDeclaration, text: []const u8, state: *SessionState) !void {
+        var store_it = self.store.iterator();
+        while (store_it.next()) |entry| {
+            state.arena.free(entry.key_ptr.*);
+            state.arena.free(entry.value_ptr.value);
+        }
+        self.store.clearRetainingCapacity();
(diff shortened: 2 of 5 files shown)


src/browser/cssom/css_parser.zig
```diff
@@ -0,0 +1,308 @@
...
+
+const CSSConstants = struct {
+    const IMPORTANT_KEYWORD = "!important";
+    const IMPORTANT_LENGTH = IMPORTANT_KEYWORD.len;
+    const URL_PREFIX = "url(";
+    const URL_PREFIX_LENGTH = URL_PREFIX.len;
+};
+
+pub const CSSParserState = enum {
+    seekName,
```
karlseguin: I think we would normally use snake_case for these, like a variable.

src/browser/cssom/css_parser.zig
```diff
@@ -0,0 +1,308 @@
...
+            .name_start = 0,
+            .name_end = 0,
+            .value_start = 0,
+            .position = 0,
+            .paren_depth = 0,
+            .escape_next = false,
+        };
+    }
+
+    pub fn parseDeclarations(text: []const u8, allocator: std.mem.Allocator) ![]CSSDeclaration {
```
karlseguin: As a general convention in Zig code, the `allocator` tends to be the first parameter (or the 2nd one if it's a method, where the 1st has to be the receiver).

src/browser/cssom/css_parser.zig
```diff
@@ -0,0 +1,308 @@
...
+        var parser = init();
+        var declarations = std.ArrayList(CSSDeclaration).init(allocator);
```
karlseguin: There's a chance the "managed" structures get removed from Zig. So we tend to favor `std.ArrayListUmamanged(CSSDeclaration)`. But I can see here how that would be a bit annoying, since you pass it around, you'd also have to pass the allocator around.

I can think of 3 options.

1 - leave it as-is, we can deal with it if ever the stdlib removes this. I'm fine with this.

2 - Pass the allocator wherever you pass `&declarations`

3 - Add the allocator and declarations as fields to `CSSParser`, then you don't need to pass either around. You could create a helper.  `self.addDeclaration(...)`:

```zig
fn addDeclaration(self: *CSSParser, decl: CSSDeclaration) !void {
  return self.declarations.append(self.allocator, decl);
```

src/browser/cssom/css_parser.zig
```diff
@@ -0,0 +1,308 @@
...
+                    }
+                },
+                .seekValue => {
+                    if (!std.ascii.isWhitespace(c)) {
+                        parser.value_start = parser.position;
+                        if (c == '"') {
+                            parser.state = .inQuotedValue;
+                        } else if (c == '\'') {
+                            parser.state = .inSingleQuotedValue;
+                        } else if (c == 'u' and parser.position + 3 < text.len and
```
karlseguin: I think this is equivalent:

```zig
else if (c == 'u' and std.mem.startsWith(u8, text[parser.position...], URL_PREFIX)) {
```

src/browser/cssom/css_parser.zig
```diff
@@ -0,0 +1,308 @@
...
+                },
+                .inImportant => {},
+            }
+
+            parser.position += 1;
+        }
+
+        try parser.finalize(text, &declarations);
+
+        return declarations.toOwnedSlice();
```
karlseguin: A bit of an eternal debate here. If this was a library, this would probably be right. But within our code, we know that `allocator` is actually an arena, so all this is doing is creating _another_ copy of the slice that we don't actually need.

You can keep it as-is, which makes this code more universally usable.

You could also return `declarations.items` and "leak" the ArrayList..but it wouldn't really be leaking because we know we're using an arena.  If you opt to go this route, you could rename `allocator` to `arena`, just so that you make it clear you expect callers to provide an arena-backed allocator. If you go this route, you might as well remove the `deinit` at the top of the function.

It would make your testing a bit more complicated, but instead of `const testing = std.testing`, you can do:

```zig
```

And then do:

```zig
test "CSSParser - Property with !important" {
     defer testing.reset(); // this resets the testing.arena_allocator
     const declarations = try CSSParser.parseDeclarations(text, testing.arena_allocator);
     // ...
}

File: src/browser/cssom/css_parser.zig [hunk 6]
```diff
@@ -0,0 +1,308 @@
...
+        if (std.mem.endsWith(u8, value, CSSConstants.IMPORTANT_KEYWORD)) {
+            is_important = true;
+            final_value = std.mem.trim(u8, value[0 .. value.len - CSSConstants.IMPORTANT_LENGTH], &std.ascii.whitespace);
+        }
+
+        const declaration = CSSDeclaration.init(name, final_value, is_important);
+        try declarations.append(declaration);
+    }
+
+    fn finalize(self: *CSSParser, text: []const u8, declarations: *std.ArrayList(CSSDeclaration)) !void {
```
karlseguin: As-is, I think this could be a `*const CSSParser`, but if you create an `addDeclaration` method, then it would need to stay non-const.

karlseguin: Can always alias `std.ArrayList(CSSDeclaration)`  to `const DeclarationList = std.ArrayList(CSSDeclaration)` if you want

or `const DeclarationList = std.ArrayListUnmanaged(CSSDeclaration)` if you go that route.

src/browser/cssom/css_parser.zig
```diff
@@ -0,0 +1,308 @@
...
+        if (self.state == .inValue) {
```
karlseguin: Since this is the only case you could inverse it.

```zig
if (self.state != .inValue) {
  return;
}
```

src/browser/cssom/css_parser.zig
```diff
@@ -0,0 +1,308 @@
...
+            const name = text[self.name_start..self.name_end];
+            const trimmed_name = std.mem.trim(u8, name, &std.ascii.whitespace);
+
+            if (trimmed_name.len > 0) {
```
karlseguin: also could be inversed

src/browser/cssom/css_parser.zig
```diff
@@ -0,0 +1,308 @@
...
+                const raw_value = text[self.value_start..self.position];
```
karlseguin: Not sure if `finishDeclaration` could be leveraged here to deduplicate any of this.

src/browser/cssom/css_parser.zig
```diff
@@ -0,0 +1,308 @@
...
+                    final_value = std.mem.trim(u8, value[0 .. value.len - CSSConstants.IMPORTANT_LENGTH], &std.ascii.whitespace);
+                }
+
+                const declaration = CSSDeclaration.init(trimmed_name, final_value, is_important);
+                try declarations.append(declaration);
+            }
+        }
+    }
+
+    pub fn getState(self: *const CSSParser) CSSParserState {
```
karlseguin: I know it isn't pretty, but in Zig, direct field access is pretty common, so these getters aren't too useful.

Does anything call `getState`, `getPosition` and `reset`?

src/browser/cssom/css_parser.zig
```diff
@@ -0,0 +1,308 @@
...
+const testing = std.testing;
+
+test "CSSParser - Simple property" {
+    const text = "color: red;";
+    const allocator = testing.allocator;
+
+    const declarations = try CSSParser.parseDeclarations(text, allocator);
+    defer allocator.free(declarations);
+
+    try testing.expect(declarations.len == 1);
```
karlseguin: `expectEqual` gives a better error message

src/browser/cssom/css_style_declaration.zig
```diff
@@ -0,0 +1,236 @@
...
+        var buffer = std.ArrayList(u8).init(state.arena);
+        const writer = buffer.writer();
+        for (self.order.items) |name| {
+            const prop = self.store.get(name).?;
+            const escaped = try CSSValueAnalyzer.escapeCSSValue(state.arena, prop.value);
+            try writer.print("{s}: {s}", .{ name, escaped });
+            if (prop.priority) try writer.writeAll(" !important");
+            try writer.writeAll("; ");
+        }
+        return buffer.toOwnedSlice();
```
karlseguin: just return `buffer.items`.

src/browser/cssom/css_style_declaration.zig
```diff
@@ -0,0 +1,236 @@
...
+        return self._getPropertyValue("float");
+    }
+
+    pub fn set_cssFloat(self: *CSSStyleDeclaration, value: ?[]const u8, state: *SessionState) !void {
+        const final_value = value orelse "";
+        return self._setProperty("float", final_value, null, state);
+    }
+
+    pub fn get_cssText(self: *const CSSStyleDeclaration, state: *SessionState) ![]const u8 {
+        var buffer = std.ArrayList(u8).init(state.arena);
```
karlseguin: Use `state.call_arena`. It's shorter-lived than `arena`..it'll live (at least) until the JS runtime is done copying the value to/from v8.

src/browser/cssom/css_style_declaration.zig
```diff
@@ -0,0 +1,236 @@
...
+            try writer.writeAll("; ");
+        }
+        return buffer.toOwnedSlice();
+    }
+
+    // TODO Propagate also upward to parent node
+    pub fn set_cssText(self: *CSSStyleDeclaration, text: []const u8, state: *SessionState) !void {
+        var store_it = self.store.iterator();
+        while (store_it.next()) |entry| {
+            state.arena.free(entry.key_ptr.*);
```
karlseguin: `free` and `destroy` on an arena are almost always no-ops.

```zig
self.store.clearRetainingCapacity();
self.order.clearRetainingCapacity();
```

Is going to be good enough. (Yes, we could grow a lot of memory until the arena is eventually cleared on page unload).

src/browser/cssom/css_style_declaration.zig
```diff
@@ -0,0 +1,236 @@
...
+            state.arena.free(entry.value_ptr.value);
+        }
+        self.store.clearRetainingCapacity();
+        for (self.order.items) |name| state.arena.free(name);
+        self.order.clearRetainingCapacity();
+
+        const declarations = try CSSParser.parseDeclarations(text, state.arena);
```
karlseguin: Testing in firefox, it seems like if I set this to an invalid CSS, it doesn't return an error (but it does blank it out), so maybe you need to catch the error?

```zig
const declarations = CSSParser.parseDeclarations(text, state.arena) catch return;
```

(Although, if there's a bug in the parser, this will make it hard to track....not sure...)

karlseguin: Worth having a test case for setting invalid CSS I think.

src/browser/cssom/css_style_declaration.zig
```diff
@@ -0,0 +1,236 @@
...
+    pub fn _getPropertyValue(self: *const CSSStyleDeclaration, name: []const u8) []const u8 {
+        return if (self.store.get(name)) |prop| prop.value else "";
+    }
+
+    pub fn _item(self: *const CSSStyleDeclaration, index: usize) []const u8 {
+        return if (index < self.order.items.len) self.order.items[index] else "";
+    }
+
+    pub fn _removeProperty(self: *CSSStyleDeclaration, name: []const u8, state: *SessionState) ![]const u8 {
+        if (self.store.get(name)) |prop| {
```
karlseguin: Combine the get + remove in a single operation:

```zig
const prop = self.store.fetchRemove(name) orelse return "";
```

Or, if you switch to a `StringArrayHashMapUnmanaged`, I think `fetchOrderedRemove` would do what you need???

src/browser/cssom/css_style_declaration.zig
```diff
@@ -0,0 +1,236 @@
...
+
+pub const Interfaces = .{
+    CSSStyleDeclaration,
+    CSSRule,
+};
+
+const CSSRule = struct {};
+
+pub const CSSStyleDeclaration = struct {
+    store: std.StringHashMapUnmanaged(Property),
```
karlseguin, line 33: You might be able to combine these two using a `std.StringArrayHashMapUnmanaged(Property)`.

See if the documented ordering guarantees satisfy the requirements:
https://ziglang.org/documentation/master/std/#std.array_hash_map.ArrayHashMapUnmanaged

src/browser/cssom/css_style_declaration.zig
```diff
@@ -0,0 +1,236 @@
...
+            }
+            return value_copy;
+        }
+        return "";
+    }
+
+    pub fn _setProperty(self: *CSSStyleDeclaration, name: []const u8, value: []const u8, priority: ?[]const u8, state: *SessionState) !void {
+        const is_important = priority != null and std.ascii.eqlIgnoreCase(priority.?, "important");
+
+        const value_copy = try state.arena.dupe(u8, value);
```
karlseguin: To be clear, in this case `arena` is correct. You would not want to use `call_arena` here, as the value has to outlive the call to `setProperty`.

src/browser/cssom/css_value_analyzer.zig
```diff
@@ -0,0 +1,327 @@
...
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+const std = @import("std");
+
+pub const CSSValueAnalyzer = struct {
+    pub fn isNumericWithUnit(value: []const u8) bool {
+        if (value.len == 0) return false;
+
+        if (!std.ascii.isDigit(value[0]) and
+            value[0] != '+' and value[0] != '-' and value[0] != '.')
```
karlseguin, line 26: Can a digit really start with a decimal?

src/browser/cssom/css_value_analyzer.zig
```diff
@@ -0,0 +1,327 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+const std = @import("std");
+
+pub const CSSValueAnalyzer = struct {
+    pub fn isNumericWithUnit(value: []const u8) bool {
```
karlseguin, line 22: `32+23+4 == true`

src/browser/cssom/css_value_analyzer.zig
```diff
@@ -0,0 +1,327 @@
...
+                break;
+            }
+        }
+
+        if (!has_digit) return false;
+
+        if (i == value.len) return true;
+
+        const unit = value[i..];
+        return CSSKeywords.isValidUnit(unit);
```
karlseguin, line 64: it's case sensitive, so `12PX` returns false

src/browser/cssom/css_value_analyzer.zig
```diff
@@ -0,0 +1,327 @@
...
+    }
+
+    pub fn isHexColor(value: []const u8) bool {
+        if (!std.mem.startsWith(u8, value, "#")) return false;
+
+        const hex_part = value[1..];
```
karlseguin, line 70: In `isNumericWithUnit` you guard against an empty length, here you don't. Maybe it's safe to assume `value.len > 0` here, dunno.

src/browser/cssom/css_value_analyzer.zig
```diff
@@ -0,0 +1,327 @@
...
+
+        for (hex_part) |c| {
+            if (!std.ascii.isHex(c)) return false;
+        }
+
+        return true;
+    }
+
+    pub fn isMultiValueProperty(value: []const u8) bool {
+        var parts = std.mem.splitAny(u8, value, " ");
```
karlseguin, line 81: I think `splitScalar` will perform better...but can the separator be any whitespace?

src/browser/cssom/css_value_analyzer.zig
```diff
@@ -0,0 +1,327 @@
...
+        while (parts.next()) |part| {
+            if (part.len == 0) continue;
+            multi_value_parts += 1;
+
+            const is_numeric = isNumericWithUnit(part);
+            const is_hex_color = isHexColor(part);
+            const is_known_keyword = CSSKeywords.isKnownKeyword(part);
+            const is_function = CSSKeywords.startsWithFunction(part);
+
+            if (!is_numeric and !is_hex_color and !is_known_keyword and !is_function) {
```
karlseguin, line 94: As an optimization, you could check each of these one at a time, before doing the next check. I think this is the same:

```zig
if (isNumericWithUnit(part)) {
   continue;
}

if (isHexColor(part)) {
   continue;
}

if (CSSKeywords.isKnownKeyword(part)) {
  continue;
}

if (CSSKeywords.startsWithFunction(part)) {
   continue;
}

all_parts_valid = false;
break;
```

src/browser/cssom/css_value_analyzer.zig
```diff
@@ -0,0 +1,327 @@
...
+                all_parts_valid = false;
```
karlseguin, line 95: can return `false` here directly?

src/browser/cssom/css_value_analyzer.zig
```diff
@@ -0,0 +1,327 @@
...
+                break;
+            }
+        }
+
+        return multi_value_parts >= 2 and all_parts_valid;
+    }
+
+    pub fn isAlreadyQuoted(value: []const u8) bool {
+        return value.len >= 2 and ((value[0] == '"' and value[value.len - 1] == '"') or
```
karlseguin, line 104: I think this is more readable.

```zig
if (value.len < 2) {
    return false;
}
const first = value[0];
const last = value[value.len - 1]
return first == last and (first == '"' or first == '\'');
```

src/browser/cssom/css_value_analyzer.zig
```diff
@@ -0,0 +1,327 @@
...
+        }
+
+        return true;
+    }
+
+    pub fn extractImportant(value: []const u8) struct { value: []const u8, is_important: bool } {
+        const trimmed = std.mem.trim(u8, value, &std.ascii.whitespace);
+
+        if (std.mem.endsWith(u8, trimmed, "!important")) {
+            const clean_value = std.mem.trim(u8, trimmed[0 .. trimmed.len - 10], &std.ascii.whitespace);
```
karlseguin: you only need a `trimEnd` here. I know it's super minor, but when I first read this, I thought "why trim this again?". I think the `trimEnd` draws the attention that the "end" has changed and needs to be re-trimmed.

src/browser/cssom/css_value_analyzer.zig
```diff
@@ -0,0 +1,327 @@
...
+        }
+
+        return .{ .value = trimmed, .is_important = false };
+    }
+
+    pub fn needsQuotes(value: []const u8) bool {
+        if (value.len == 0) return true;
+        if (isAlreadyQuoted(value)) return false;
+
+        const has_spaces = std.mem.indexOf(u8, value, " ") != null;
```
karlseguin: `indexOf` will detect that the needle is len == 1 and call `indexOfScalar` for you, but you can also call it directly to avoid that extra check.

But, again, do we care about other whitespace characters?

src/browser/cssom/css_value_analyzer.zig
```diff
@@ -0,0 +1,327 @@
...
+        const has_special_chars = CSSKeywords.containsSpecialChar(value);
```
karlseguin: Seems like if this is `true`, you can immediately return `true` (and you don't need to check `has_spaces`)

src/browser/cssom/css_value_analyzer.zig
```diff
@@ -0,0 +1,327 @@
...
+        const space_requires_quotes = has_spaces and
+            !isMultiValueProperty(value) and
+            !is_url and
+            !is_function;
+
+        return has_special_chars or space_requires_quotes;
+    }
+
+    pub fn escapeCSSValue(allocator: std.mem.Allocator, value: []const u8) ![]const u8 {
+        var out = std.ArrayList(u8).init(allocator);
```
karlseguin: See previous comments re using `Unamanged` variant and assuming that `allocator` is backed by an Arena (which, if you do, I'd strongly recommend you rename `allocator` to `arena` as a form of documenting your expectations)

karlseguin: Actually, assuming an `arena` here has 1 added benefit. You can return an allocated or non-allocated value.

If you _don't_ assume an arena, then you have to assume that the caller is going to eventually call `allocator.free(return_value)`. So you _need_ to return an allocated value.

If you assume an arena, then the caller should never call `allocator.free(return_value)`, so in cases where the value doesn't need to be escaped, you can return `value` directly, avoiding the duplication.

src/browser/cssom/css_value_analyzer.zig
```diff
@@ -0,0 +1,327 @@
...
+        errdefer out.deinit();
+        const writer = out.writer();
+
+        if (isAlreadyQuoted(value)) {
+            try writer.writeAll(value);
+            return out.toOwnedSlice();
+        }
+
+        const needs_quotes = needsQuotes(value);
```
karlseguin: `needsQuotes` calls `isAlreadyQuoted`, so you can tighten this up a little I think

```zig
if (!needsQuotes(value)) {
  // return value directly, assumes the allocator is an arena
  // if you opt to keep it more generic, you could: `return allocator.dupe(u8, value);`
  return value; 
}`

File: src/browser/cssom/css_value_analyzer.zig [hunk 32]
```diff
@@ -0,0 +1,327 @@
...
+
+            try writer.writeByte('"');
+        } else {
+            try writer.writeAll(value);
+        }
+
+        return out.toOwnedSlice();
+    }
+
+    pub fn isKnownKeyword(value: []const u8) bool {
```
karlseguin, line 213: A lot of these don't need to be public..they're only used internally. Either is fine, just pointing it out.

src/browser/cssom/css_value_analyzer.zig
```diff
@@ -0,0 +1,327 @@
...
+    const SpecialChars = [_]u8{
+        '"', '\'', ';', '{', '}', '\\', '<', '>', '/',
+    };
+
+    const Functions = [_][]const u8{
+        "rgb",             "rgba",            "hsl",            "hsla",      "url",    "calc",  "var",  "attr",
+        "linear-gradient", "radial-gradient", "conic-gradient", "translate", "rotate", "scale", "skew", "matrix",
+    };
+
+    pub fn isKnownKeyword(value: []const u8) bool {
```
karlseguin, line 291: Re case-sensitivity...

if you assume the longest special value is 15 characters, i.e. `radial-gradient`, you could:

```zig
// 15 is our longest special value
// TODO: would be nice not to have this automatically derived from our definition somehow?
if (value.len > 15) {
  return false;
}

var buf: [15]u8 = undefined;
const v = std.ascii.lowerString(&buf, value);
```

and then use `v` instead of `value` to get case-insensitive comparison. You'll have to make sure all the constants are lowercase, e.g. `Hz` -> `hz`

At this point, you could look at alternatives to looping, like `std.meta.stringToEnum`(e.g. https://github.com/lightpanda-io/browser/blob/main/src/browser/mime.zig#L174)..but I've benchmarked this before and the difference tends to be pretty small (but more obvious for short values, since the SIMD operation of `std.mem.eql` can't be leveraged, which is what we have here).

Well, I think the case-sensitivity needs to be fixed, we can optimize at a later time if needed.

src/browser/cssom/css_value_analyzer.zig
```diff
@@ -0,0 +1,327 @@
...
+        return CSSKeywords.containsSpecialChar(value);
+    }
+};
+
+const CSSKeywords = struct {
+    const BorderStyles = [_][]const u8{
+        "none", "solid", "dotted", "dashed", "double", "groove", "ridge", "inset", "outset",
+    };
+
+    const ColorNames = [_][]const u8{
```
karlseguin: PascalCase is generally used for Types.

For non-global consts, I think we do `color_names`.

src/browser/html/elements.zig
```diff
@@ -103,15 +104,19 @@ pub const Union = generate.Union(Interfaces);
...
     pub const subtype = .node;
 
-    pub fn get_style(_: *parser.ElementHTML) CSSProperties {
-        return .{};
+    style: CSSStyleDeclaration = .{
+        .store = .{},
+        .order = .{},
+    },
+
+    pub fn get_style(e: *parser.ElementHTML, state: *SessionState) !*CSSStyleDeclaration {
```
karlseguin, line 117: Can you add some basic tests for this? Nothing nearly as exhaustive as what you've already done, but maybe just calling `el.style.setProperty` with a straightforward key-value and then asserting the value with `el.style.getPropertyValue`

src/browser/cssom/css_value_analyzer.zig
```diff
@@ -0,0 +1,327 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+const std = @import("std");
+
+pub const CSSValueAnalyzer = struct {
```
karlseguin, line 21: I know the tests in `css_style_declarations` are good, but this is very unit-testing-friendly code where you can easily test a number of edge / error cases. I could see test cases that cover things like weird numbers, case sensitivity, various whitespaces, invalid quoting, etc, being quite useful to prevent any regression if we update the code in the future.

src/browser/cssom/css_value_analyzer.zig
```diff
@@ -0,0 +1,327 @@
...
+            }
+        }
+
+        return false;
+    }
+
+    pub fn startsWithFunction(value: []const u8) bool {
+        for (Functions) |func| {
+            if (value.len >= func.len + 1 and
+                std.mem.startsWith(u8, value, func) and
```
karlseguin: Could include the `(` in the constants of `Functions` ?

src/browser/cssom/css_parser.zig
```diff
@@ -0,0 +1,308 @@
+const std = @import("std");
```
krichprollsch, line 19: The license header is missing.

src/browser/cssom/css_style_declaration.zig
```diff
@@ -0,0 +1,236 @@
...
+const CSSParser = @import("./css_parser.zig").CSSParser;
+const CSSValueAnalyzer = @import("./css_value_analyzer.zig").CSSValueAnalyzer;
+const SessionState = @import("../env.zig").SessionState;
+
+pub const Interfaces = .{
+    CSSStyleDeclaration,
+    CSSRule,
+};
+
+const CSSRule = struct {};
```
krichprollsch, line 30: shouldn't be `pub const`?

src/browser/cssom/css_parser.zig
```diff
@@ -0,0 +1,308 @@
...
+        var final_value = value;
+        var is_important = false;
+
+        if (std.mem.endsWith(u8, value, CSSConstants.IMPORTANT_KEYWORD)) {
+            is_important = true;
+            final_value = std.mem.trim(u8, value[0 .. value.len - CSSConstants.IMPORTANT_LENGTH], &std.ascii.whitespace);
+        }
+
+        const declaration = CSSDeclaration.init(name, final_value, is_important);
+        try declarations.append(declaration);
```
krichprollsch: I think I would remove ` CSSDeclaration.init` function and use instead

```zig
        try declarations.append(.{
                .name = name,
                .value = value,
                .is_important = is_important,
        });
```

src/browser/cssom/css_parser.zig
```diff
@@ -0,0 +1,308 @@
...
+
+                var final_value = value;
+                var is_important = false;
+                if (std.mem.endsWith(u8, value, CSSConstants.IMPORTANT_KEYWORD)) {
+                    is_important = true;
+                    final_value = std.mem.trim(u8, value[0 .. value.len - CSSConstants.IMPORTANT_LENGTH], &std.ascii.whitespace);
+                }
+
+                const declaration = CSSDeclaration.init(trimmed_name, final_value, is_important);
+                try declarations.append(declaration);
```
krichprollsch: same here

src/browser/cssom/css_style_declaration.zig
krichprollsch: you should use the `state.call_arena` if the buffer must live only build the string returned by the function.

src/browser/cssom/css_style_declaration.zig
krichprollsch: since we are using an arena and we know it, you can direclyt retun `buffer.items` to avoid a useless double allocation.

src/browser/cssom/css_style_declaration.zig
```diff
@@ -0,0 +1,236 @@
...
+        const final_value = value orelse "";
+        return self._setProperty("float", final_value, null, state);
+    }
+
+    pub fn get_cssText(self: *const CSSStyleDeclaration, state: *SessionState) ![]const u8 {
+        var buffer = std.ArrayList(u8).init(state.arena);
+        const writer = buffer.writer();
+        for (self.order.items) |name| {
+            const prop = self.store.get(name).?;
+            const escaped = try CSSValueAnalyzer.escapeCSSValue(state.arena, prop.value);
```
krichprollsch: same about the call_arena

src/browser/cssom/css_style_declaration.zig
```diff
@@ -0,0 +1,236 @@
...
+        }
+        return buffer.toOwnedSlice();
+    }
+
+    // TODO Propagate also upward to parent node
+    pub fn set_cssText(self: *CSSStyleDeclaration, text: []const u8, state: *SessionState) !void {
+        var store_it = self.store.iterator();
+        while (store_it.next()) |entry| {
+            state.arena.free(entry.key_ptr.*);
+            state.arena.free(entry.value_ptr.value);
```
krichprollsch: in arena, `free` are noop. So this is useless...

src/browser/cssom/css_style_declaration.zig
karlseguin: I also wasn't sure, but I checked, and it doesn't have to be (I made `XMLHttpRequest` non public and everything worked fine)

RaphUn:

PR #681, opened 2025-05-22
MouseEvent Interface

mookums:
Adds the interface for the `MouseEvent`. it wraps the `parser.MouseEvent` as this seems like the correct procedure for this.

It includes the subset of functionality needed to continue making progress on Reddit fetching.

### DIFF
--- a/src/browser/events/event.zig
+++ b/src/browser/events/event.zig
@@ -29,6 +29,7 @@ const EventTargetUnion = @import("../dom/event_target.zig").Union;
 
 const CustomEvent = @import("custom_event.zig").CustomEvent;
 const ProgressEvent = @import("../xhr/progress_event.zig").ProgressEvent;
+const MouseEvent = @import("mouse_event.zig").MouseEvent;
 
 const log = std.log.scoped(.events);
 
@@ -37,6 +38,7 @@ pub const Interfaces = .{
     Event,
     CustomEvent,
     ProgressEvent,
+    MouseEvent,
 };
 
 pub const Union = generate.Union(Interfaces);
@@ -60,6 +62,7 @@ pub const Event = struct {
             .event => .{ .Event = evt },
             .custom_event => .{ .CustomEvent = @as(*CustomEvent, @ptrCast(evt)).* },
             .progress_event => .{ .ProgressEvent = @as(*ProgressEvent, @ptrCast(evt)).* },
+            .mouse_event => .{ .MouseEvent = @as(**parser.MouseEvent, @ptrCast(evt)).* },
         };
     }
 
--- a/src/browser/events/mouse_event.zig
+++ b/src/browser/events/mouse_event.zig
@@ -1,0 +1,105 @@
+// Copyright (C) 2023-2024  Lightpanda (Selecy SAS)
+//
+// Francis Bouvier <francis@lightpanda.io>
+// Pierre Tachoire <pierre@lightpanda.io>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+const parser = @import("../netsurf.zig");
+const Event = @import("event.zig").Event;
+const JsObject = @import("../env.zig").JsObject;
+
+// https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent
+pub const MouseEvent = struct {
+    pub const Self = parser.MouseEvent;
+    pub const prototype = *Event;
+
+    const MouseEventInit = struct {
+        screenX: i32 = 0,
+        screenY: i32 = 0,
+        clientX: i32 = 0,
+        clientY: i32 = 0,
+        ctrlKey: bool = false,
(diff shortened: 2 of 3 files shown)


src/browser/events/event.zig
```diff
@@ -60,6 +62,7 @@ pub const Event = struct {
             .event => .{ .Event = evt },
             .custom_event => .{ .CustomEvent = @as(*CustomEvent, @ptrCast(evt)).* },
             .progress_event => .{ .ProgressEvent = @as(*ProgressEvent, @ptrCast(evt)).* },
+            .mouse_event => .{ .MouseEvent = @as(**parser.MouseEvent, @ptrCast(evt)).* },
```
karlseguin: is `**` right? I'm confused.

src/browser/events/mouse_event.zig
```diff
@@ -0,0 +1,105 @@
...
+        return mouse_event;
+    }
+
+    // These is just an alias for clientX.
+    pub fn get_x(self: *Self) !i32 {
+        return self.cx;
+    }
+
+    // These is just an alias for clientY.
+    pub fn get_y(self: *Self) !i32 {
```
karlseguin: Don't think these need to be an error union? In the other APIs, we normally put the explicit receiver type, i.e. `self: *parser.MouseEvent` rather than `self: *Self`. I know it's the same, but it does help you remember that the receiver isn't actually a `MouseEvent`.

src/browser/events/mouse_event.zig
```diff
@@ -0,0 +1,105 @@
...
+        clientX: i32 = 0,
+        clientY: i32 = 0,
+        ctrlKey: bool = false,
+        shiftKey: bool = false,
+        altKey: bool = false,
+        metaKey: bool = false,
+        button: u16 = 0,
+    };
+
+    pub fn constructor(event_type: []const u8, opts_: ?MouseEventInit) !*Self {
```
sjorsdonkers: Can button be an enum?
Should we check the value of event_type? Currently we only have listeners for "click", should we warn if it is something else?

src/browser/events/event.zig
sjorsdonkers: Likely related to the type here and returned by the constructor being `parser.MouseEvent` where I would expect the newly created struct MouseEvent from mouse_event.zig
Does `proto: parser.Event,` need to be added?

src/browser/events/mouse_event.zig
```diff
@@ -0,0 +1,105 @@
...
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+const parser = @import("../netsurf.zig");
+const Event = @import("event.zig").Event;
+const JsObject = @import("../env.zig").JsObject;
+
+// https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent
+pub const MouseEvent = struct {
+    pub const Self = parser.MouseEvent;
+    pub const prototype = *Event;
```
sjorsdonkers: Note, we are skipping [UIEvent](https://developer.mozilla.org/en-US/docs/Web/API/UIEvent) here in the prototype chain.
I think we whould either make UIEvent or put a TODO note here.
I'd probably make an alias `const UIEvent = Event;` with a comment explaining it

src/browser/events/mouse_event.zig
```diff
@@ -0,0 +1,105 @@
...
+        shiftKey: bool = false,
+        altKey: bool = false,
+        metaKey: bool = false,
+        button: u16 = 0,
+    };
+
+    pub fn constructor(event_type: []const u8, opts_: ?MouseEventInit) !*Self {
+        const opts = opts_ orelse MouseEventInit{};
+
+        const mouse_event = try parser.mouseEventCreate();
```
sjorsdonkers: If I'm not mistaken this leaves the `evt->internal_type = 0;`

I expect we need to add `try parser.eventSetInternalType(event, .mouse_event);`

Maybe a good way to test is to test by dispatching the event (for a "click"):
```
element.dispatchEvent(event);
```
edit: or maybe easier to create your own event listener like ProgressEvent

src/browser/events/event.zig
mookums: Adding `proto: parser.Event`didn't seem to change anything for me. It seems to expect a pointer to whatever the constructor returns so that ends up being a pointer to the pointer to the `parser.MouseEvent`

src/browser/events/mouse_event.zig
mookums: Yeah, added to `jsValueToZig` to support integer tagged enums.

src/browser/events/mouse_event.zig
mookums: Added basically the same style of event listener that `ProgressEvent` uses to the test for `MouseEvent`

src/browser/events/mouse_event.zig
mookums: Added an alias and a note about UIEvent.

src/browser/events/event.zig
sjorsdonkers: I think this should be `.mouse_event => .{ .MouseEvent = @as(*parser.MouseEvent, @ptrCast(evt)) },` so without the `.*`
similar to other toInterface functions where the underlying parser type is used (like htmlElements), note that the other Events are our own zig types, similar to Window and such.

I was hoping the new evenlistener test would run this code, but it doesn't. Currently if you replace this line with `else => unreachable,` the tests still work. I think it should be correct tho with the already added `eventSetInternalType(event, .mouse_event);` and the change I mention here.

karlseguin:
LGTM

src/browser/events/event.zig
sjorsdonkers: @mookums We can merge this after changing:
> I think this should be `.mouse_event => .{ .MouseEvent = @as(*parser.MouseEvent, @ptrCast(evt)) },` so without the `.*`

Unless that breaks something with reddit fetching

mookums: Just pushed it. No new issues with reddit fetching.

PR #689, opened 2025-05-23
new Image constructor

sjorsdonkers:
Related to https://github.com/lightpanda-io/project/issues/115#issue-3039251901
`fetch 'https://www.google.com/search?q=browser+headless'`

Constructors + all non deprecated properties implemented by netsurf
Q: Do we want the deprecated ones as well, or do they just take up memory?

### DIFF
--- a/src/browser/html/elements.zig
+++ b/src/browser/html/elements.zig
@@ -565,23 +565,49 @@ pub const HTMLImageElement = struct {
     pub const subtype = .node;
     pub const js_name = "Image";
 
-    pub fn constructor(width: ?[]const u8, height: ?[]const u8, state: *const SessionState) !*parser.Image {
+    pub fn constructor(width: ?u32, height: ?u32, state: *const SessionState) !*parser.Image {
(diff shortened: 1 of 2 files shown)


src/browser/html/elements.zig
```diff
@@ -563,6 +563,26 @@ pub const HTMLImageElement = struct {
...
+        const width = try parser.elementGetAttribute(@ptrCast(self), "width") orelse return 0;
+        return try std.fmt.parseInt(u32, width, 10);
+    }
+    pub fn get_height(self: *parser.Image) !u32 {
+        const width = try parser.elementGetAttribute(@ptrCast(self), "height") orelse return 0;
+        return try std.fmt.parseInt(u32, width, 10);
+    }
+    pub fn get_src(self: *parser.Image) ![]const u8 {
+        return try parser.elementGetAttribute(@ptrCast(self), "src") orelse return "";
+    }
```
sjorsdonkers, line 585: Why do we need to define the getters, but not the setters? 🤔

karlseguin:
> Q: Do we want the deprecated ones as well, or do they just take up memory?

We can add them later when we find a use case that needs them

PR #691, opened 2025-05-25
Replace std.log with a structured logger

karlseguin:
Trying to achieve a few different things:

1 - In a deployed / production environment, using a standard format (logfmt) so that the logs can easily be ingested by tools like vector or logstash, etc..

2 - In a development environment, printing easier to read logs (subjective)

3 - Generally having more standardized logs. Imagine an HTTP fetch fails. With Zig's format-based logging, you have something like:   `log.err("request {s} failed: {}", .{url, err});`. It works, but it isn't particularly parser friendly. Plus, it can be hard to keep consistent with other logs. For example: `log.info("fetching: '{s}'", .{url});` Now the URL is in quotes and it's after the colon.  With the structured logs, the `url` and `err` are arguments: `log.err(.http_client, "request failed", .{.url = url, .err = err}); which, in `logfmt` generates key=value attribute.

The pretty output isn't that special, but my old eyes will appreciate the padding:
Screenshot 2025-05-25 at 7 32 26 PM

### DIFF
--- a/src/log.zig
+++ b/src/log.zig
@@ -81,12 +81,16 @@ pub fn log(comptime scope: @Type(.enum_literal), comptime level: Level, comptime
     if (comptime enabled(scope, level) == false) {
         return;
     }
-    tryLog(scope, level, msg, data) catch |log_err| {
+
+    std.debug.lockStdErr();
+    defer std.debug.unlockStdErr();
+
+    logTo(scope, level, msg, data, std.io.getStdErr().writer()) catch |log_err| {
         std.debug.print("$time={d} $level=fatal $scope={s} $msg=\"log err\" err={s} log_msg=\"{s}\"", .{ timestamp(), @errorName(log_err), @tagName(scope), msg });
     };
 }
 
-fn tryLog(comptime scope: @Type(.enum_literal), comptime level: Level, comptime msg: []const u8, data: anytype) !void {
+fn logTo(comptime scope: @Type(.enum_literal), comptime level: Level, comptime msg: []const u8, data: anytype, out: anytype) !void {
     comptime {
         if (msg.len > 30) {
             @compileError("log msg cannot be more than 30 characters: '" ++ msg ++ "'");
@@ -102,12 +106,7 @@ fn tryLog(comptime scope: @Type(.enum_literal), comptime level: Level, comptime 
         }
     }
 
-    out_lock.lock();
-    defer out_lock.unlock();
-
-    const stderr = std.io.getStdErr().writer();
-    var bw = std.io.bufferedWriter(stderr);
-
+    var bw = std.io.bufferedWriter(out);
     switch (format) {
         .logfmt => try logLogfmt(scope, level, msg, data, bw.writer()),
         .pretty => try logPretty(scope, level, msg, data, bw.writer()),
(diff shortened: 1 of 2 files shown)


src/log.zig
```diff
@@ -0,0 +1,480 @@
...
+        out: Out,
+        format: Format,
+        allocator: Allocator,
+        buffer: std.ArrayListUnmanaged(u8),
+
+        const Self = @This();
+
+        fn init(allocator: Allocator, opts: Opts) !Self {
+            return initTo(allocator, opts, std.io.getStdOut());
+        }
```
sjorsdonkers: Not used

src/log.zig
```diff
@@ -0,0 +1,480 @@
...
+    };
+}
+
+// Generic so that we can test it against an ArrayList
+fn LogT(comptime Out: type) type {
+    return struct {
+        out: Out,
+        format: Format,
+        allocator: Allocator,
+        buffer: std.ArrayListUnmanaged(u8),
```
sjorsdonkers: Unmanaged since we think ArrayList may become deprecated?

src/log.zig
```diff
@@ -0,0 +1,480 @@
...
+        }
+
+        fn deinit(self: *Self) void {
+            self.buffer.deinit(self.allocator);
+        }
+
+        fn log(self: *Self, comptime scope: @Type(.enum_literal), comptime level: Level, comptime msg: []const u8, data: anytype) !void {
+            comptime {
+                if (msg.len > 30) {
+                    @compileError("log msg cannot be more than 30 characters: '" ++ msg ++ "'");
```
sjorsdonkers: What is the reason for this? Especially since when data is added the length is still any size

src/log.zig
```diff
@@ -0,0 +1,480 @@
...
+                break :blk lm ++ "\"" ++ msg ++ "\"";
+            };
+            buffer.appendSliceAssumeCapacity(level_and_msg);
+            inline for (@typeInfo(@TypeOf(data)).@"struct".fields) |f| {
+                const key = " " ++ f.name ++ "=";
+
+                // + 5 covers null/true/false
+                try buffer.ensureUnusedCapacity(allocator, key.len + 5);
+                buffer.appendSliceAssumeCapacity(key);
+                try writeValue(allocator, buffer, true, @field(data, f.name));
```
sjorsdonkers: Having the +5 as a contract to writeValue is a bit fragile.
It may be better to make is a writeKeyValue or writeField function, that internally has a local writeValue function for it to recurse into.

src/log.zig
```diff
@@ -0,0 +1,480 @@
...
+            try buffer.append(allocator, '\n');
+        }
+    };
+}
+
+const Pool = struct {
+    loggers: []*Log,
+    available: usize,
+    mutex: Thread.Mutex,
+    cond: Thread.Condition,
```
sjorsdonkers: It is not clear to me why we have a Log Pool as opposed to a thread local logger. So 1 per thread.
Specifically since we are also locking the `Out` before writing.
I understand it may save a bit of memory if we have many threads, but does introduce some complex performance dependency. It feels is a bit premature.

src/log.zig
```diff
@@ -0,0 +1,480 @@
...
+
+    // + 2 for the quotes
+    // + escape_count because every character that needs escaping is + 1
+    try buffer.ensureUnusedCapacity(allocator, 2 + value.len + escape_count);
+
+    buffer.appendAssumeCapacity('"');
+
+    var rest = value;
+    while (rest.len > 0) {
+        const pos = std.mem.indexOfAny(u8, rest, "\r\n\"") orelse {
```
sjorsdonkers, line 252: Maybe `\r` should just be illegal

src/log.zig
karlseguin: Yes, there's some "insight" here: https://github.com/ziglang/zig/pull/22087

src/log.zig
karlseguin: If you're ingesting the logs into a system, this is essentially the key. It's what you'd `group by` on.  It's a bit like msgid from syslog, but allows for more human-readable values.

src/log.zig
karlseguin: I agree. I removed it and just do a normal append(allocator, ...) with no assumption in all value cases.

src/log.zig
karlseguin: You might be logging a variable:

```zig
log.debug(.js, "callback", .{.src = callback.toString()});
```

src/log.zig
karlseguin: Well, initially the logging had more advanced use-cases that would result in more than 1 logger active per thread.

I'm ok with starting with TLS, but I'm not sure how I'd initialize the instance given it needs an allocator.

karlseguin:
> Does this PR have performance impact?

I didn't know about the performance thing, but I doubt this would be any faster. It adds a bit more formatting, i.e. escaping strings.

This is really about adding structured logging to make it easier to parse the output programmatically. I'd also like to make the log-level a runtime configuration (i.e.  ./lightpanda --log_level debug) rather than a compile-time one. And that's a bit messy to do with std.log...although that'll also add performance overhead.

PR #695, opened 2025-05-26
trusted types

sjorsdonkers:
Limited implementation of TrustedTypes
May not merge PR as not all browser support this yet.
Google seems to work around in case the browser does not support it. (just concats the code string)
When loading google search at this moment `u.trustedTypes` is called, but no follow up calls yet.


krichprollsch:
Closed due to `main` push force. see #1181

PR #705, opened 2025-05-27
Replace SessionState directly with the Page.

karlseguin:
I want to make it so the `call_arena` is never undefined, but I'll do that in a separate PR since this PR is already big enough. With the SessionState, the `call_arena` was also undefined.

The main reason I want this is so that the WebAPIs can call into the page/session/browser. For example, on `location.reload()` we need to trigger a page navigate and, with the SessionState, there was no link to the browser/session/page.


src/browser/polyfill/fetch.zig
```diff
@@ -31,16 +31,16 @@ test "Browser.fetch" {
...
+    //         \\  var ok2 = false;
+    //         \\  const request2 = new Request("http://127.0.0.1:9582/loader");
+    //         \\  (async function () { resp = await fetch(request2); ok2 = resp.ok; }());
+    //         \\  false;
+    //         ,
+    //         "false",
+    //     },
+    //     // all events have been resolved.
+    //     .{ "ok2", "true" },
+    // }, .{});
```
sjorsdonkers: Is this intentional still commented out?

src/browser/storage/cookie.zig
```diff
@@ -232,7 +232,7 @@ fn areCookiesEqual(a: *const Cookie, b: *const Cookie) bool {
 
 fn areSameSite(origin_uri_: ?*const std.Uri, target_host: []const u8) !bool {
     const origin_uri = origin_uri_ orelse return true;
-    const origin_host = (origin_uri.host orelse return error.InvalidURI).percent_encoded;
+    const origin_host = (origin_uri.host orelse return error.InvalidURI333).percent_encoded;
```
sjorsdonkers: 333?

sjorsdonkers (commented):
I'm getting crashes when running the playwright demo scripts. I'm looking into it

src/browser/polyfill/fetch.zig
karlseguin: nope..I was having problems with it, forgot about it.

src/browser/storage/cookie.zig
karlseguin: Heh..was also having problems here (because of the test runner needed to change)..and trying to figure out which InvalidURI was being returned to JS. Fixed now

PR #708, opened 2025-05-27
Re-enable microtask loop

karlseguin:
Must have gotten disabled in a merge?


krichprollsch:
I think we loose the change in 1820e79617a6d1d0e905257ed1e2790c478ab505 in the refacto #646

PR #709, opened 2025-05-27
Take scope out of scope

sjorsdonkers:
Decouples, scope & context creation & entering
- Moves the HandleScope out of the Scope as it is not possible to have 1 HandleScope per Context.
- Exposes the main context's responsibility to enter and exit it's v8_context
- Any additional Locals remaining during context creation have their lifetime limited (previously was page length, now just temp scope)

### DIFF
--- a/src/browser/page.zig
+++ b/src/browser/page.zig
@@ -85,6 +85,11 @@ pub const Page = struct {
     // execute any JavaScript
     scope: *Env.Scope,
 
+    // For a Page we only create one HandleScope and keep it alive for the duration of the page.
+    // If needed JS Locals' lifetimes can be reduced by layering on additional Scopes.
+    // Note: An isolate has only 1 scope stack shared by all main and isolated worlds.
+    handle_scope: Env.HandleScope,
+
     // List of modules currently fetched/loaded.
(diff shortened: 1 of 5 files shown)


karlseguin:
I'm not sure I see the point. Does it let us add a feature that's otherwise hard to add or improve performance? 

Having the Page know about HandleScope - a detail of V8 - sounds like a step backwards with respect to coupling.

sjorsdonkers:
The most important thing to me what these PRs work toward is changing the name Scope as I consider it misleading.
I understand your point that exposing the concept of an actual scope may work differently in other runtimes, but I don't know.
What I do know is that Context and Scope are orthogonal concepts for which we now have some enter boolean to determine whether we should create a scope.
This PR gives the main context the responsibility to enter it (which should be only in 1 place in a program), so that we do not hide away this important operation. Same for the HandleScope.
I don't think we should be preemptively trying to solve making a generic interface for a runtime, before we have at least 3 or so runtimes we would want to support, especially since we do not know if we will ever need to be generic. Since there is no way to tell (without exploring other runtimes) if the current Scope is a good abstraction I think we should not make the abstraction in this case (given the orthogonality of the concepts).
So better to fix it sooner than later, hence the PR.

PR #710, opened 2025-05-27
Rename scope to context

sjorsdonkers:
Based on https://github.com/lightpanda-io/browser/pull/709

Renames The Zig Scope struct to Context as it is closely modeling/wrapping the v8_context.

Renamed the Page's Context to `main_context` to clarify that it is holding the main Worlds context

### DIFF
--- a/src/runtime/js.zig
+++ b/src/runtime/js.zig
@@ -932,7 +932,7 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
                     },
                     .@"enum" => |e| {
                         switch (@typeInfo(e.tag_type)) {
-                            .int => return std.meta.intToEnum(T, try jsIntToZig(e.tag_type, js_value, self.context)),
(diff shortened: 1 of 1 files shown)


karlseguin:
I'm against this PR. I think naming things the same just makes communication more complicated.

> I think we have a leak in the context. 
> Which context? Our context or the v8 context?

And just the ambiguity of a variable name `context` while reviewing a PR.

sjorsdonkers:
I'm very open an alternative name. I considered ExecutionContext.
The current name Scope I consider strictly incorrect and very confusing as Scope and Context are established concepts that should not be interchanged.
Previously this held the handle_scope and a context, but only the scope in the cases this was the main scope since there is only 1 scope per isolate, but there are 2 (possibly many) contexts. Since at maximum 1 of these can hold the scope it was removed from this struct in the previous PR. So at this stage there is no scopiness about this struct.
Additionally I do think it is a Context, because that is the main component it is wrapping and loading. So naming it very differently, introducing yet another new name is adding complication for things that are basically the same.

PR #719, opened 2025-05-28
xmlserializer for doctype

sjorsdonkers:
Puppeteer and Playwright explicitly want to write the DOCTYPE during `page.content()`
`retVal = new XMLSerializer().serializeToString(document.doctype);`
See: https://github.com/lightpanda-io/demo/pull/45

doctype used to be serialized as an empty string as only the top level document should have a doctype. As such it should be handled separately just like Document

### DIFF
--- a/src/browser/xmlserializer/xmlserializer.zig
+++ b/src/browser/xmlserializer/xmlserializer.zig
@@ -62,4 +62,3 @@ test "Browser.XMLSerializer with DOCTYPE" {
         .{ "new XMLSerializer().serializeToString(document.doctype)", "<!DOCTYPE html>" },
     }, .{});
 }
-test


src/browser/xmlserializer/xmlserializer.zig
```diff
@@ -54,3 +54,12 @@ test "Browser.XMLSerializer" {
...
 }
+test "Browser.XMLSerializer with DOCTYPE" {
+    var runner = try testing.jsRunner(testing.tracking_allocator, .{ .html = "<!DOCTYPE html><html><head></head><body></body></html>" });
+    defer runner.deinit();
+
+    try runner.testCases(&.{
+        .{ "new XMLSerializer().serializeToString(document.doctype)", "<!DOCTYPE html>" },
+    }, .{});
+}
+test
```
krichprollsch: Should be removed I suppose

sjorsdonkers: Ha, thanks. I thought it was having the same CI downloading issues as the other PRs :)

PR #721, opened 2025-05-28
Input element properties

sjorsdonkers:
Depends on: https://github.com/lightpanda-io/libdom/pull/28

https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement
Properties that are implemented by Netsurf, not deprecated, and not inherited from HTMLElement.

With these we probable have enough examples to make a generic implementation, for that note that there are some special cases:
- Some netsurf properties are deprecated
- Netsurf does not support all properties
- Some properties are inherited through the prototype chain
- Some properties have default values if the netsurf says null or the unsigned version of -1
- The initial value and the default value do not need to be the same. init -1 default 0 for example.
- The set and get types may be different get may return -1 `i32`, but set must be positive `u32`
- Some specific values instead of defaulting throw an error.
- Sometimes the Set functions is not implemented bynetsurf even if the property is not readonly
- The default initialization and reverting to default on bad value assignment also needs to work for `Element.setAttribute` and `Element.getAttribute` 
- Values may related default value properties used until the value itself is set.
- src properties need to be stiched with the pages url.
- Netsurf may create more DOM events than defined by the spec upon setting a property without using setAttribute. TBD

PR created to make Playwright `page.getByRole` work.

### DIFF
--- a/src/browser/html/elements.zig
+++ b/src/browser/html/elements.zig
@@ -626,6 +626,85 @@ pub const HTMLInputElement = struct {
     pub const Self = parser.Input;
     pub const prototype = *HTMLElement;
     pub const subtype = .node;
+
+    pub fn get_defaultValue(self: *parser.Input) ![]const u8 {
+        return try parser.inputGetDefaultValue(self);
+    }
+    pub fn set_defaultValue(self: *parser.Input, default_value: []const u8) !void {
+        try parser.inputSetDefaultValue(self, default_value);
+    }
+    pub fn get_defaultChecked(self: *parser.Input) !bool {
+        return try parser.inputGetDefaultChecked(self);
+    }
+    pub fn set_defaultChecked(self: *parser.Input, default_checked: bool) !void {
+        try parser.inputSetDefaultChecked(self, default_checked);
+    }
+    pub fn get_from(self: *parser.Input) !?*parser.Form {
+        return try parser.inputGetForm(self);
+    }
+    pub fn get_accept(self: *parser.Input) ![]const u8 {
+        return try parser.inputGetAccept(self);
+    }
+    pub fn set_accept(self: *parser.Input, accept: []const u8) !void {
+        try parser.inputSetAccept(self, accept);
+    }
+    pub fn get_alt(self: *parser.Input) ![]const u8 {
+        return try parser.inputGetAlt(self);
+    }
+    pub fn set_alt(self: *parser.Input, alt: []const u8) !void {
+        try parser.inputSetAlt(self, alt);
+    }
+    pub fn get_checked(self: *parser.Input) !bool {
+        return try parser.inputGetChecked(self);
+    }
+    pub fn set_checked(self: *parser.Input, checked: bool) !void {
+        try parser.inputSetChecked(self, checked);
+    }
+    pub fn get_disabled(self: *parser.Input) !bool {
+        return try parser.inputGetDisabled(self);
+    }
+    pub fn set_disabled(self: *parser.Input, disabled: bool) !void {
+        try parser.inputSetDisabled(self, disabled);
+    }
+    pub fn get_maxLength(self: *parser.Input) !i32 {
+        return try parser.inputGetMaxLength(self);
+    }
+    pub fn set_maxLength(self: *parser.Input, max_length: u32) !void {
+        try parser.inputSetMaxLength(self, max_length);
+    }
+    pub fn get_name(self: *parser.Input) ![]const u8 {
+        return try parser.inputGetName(self);
+    }
+    pub fn set_name(self: *parser.Input, name: []const u8) !void {
+        try parser.inputSetName(self, name);
+    }
+    pub fn get_readOnly(self: *parser.Input) !bool {
+        return try parser.inputGetReadOnly(self);
+    }
+    pub fn set_readOnly(self: *parser.Input, read_only: bool) !void {
+        try parser.inputSetReadOnly(self, read_only);
+    }
+    pub fn get_size(self: *parser.Input) !u32 {
+        return try parser.inputGetSize(self);
+    }
+    pub fn set_size(self: *parser.Input, size: u32) !void {
+        try parser.inputSetSize(self, size);
+    }
+    pub fn get_src(self: *parser.Input) ![]const u8 {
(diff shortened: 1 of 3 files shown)


karlseguin:
No tests (should have), but this case wrongly fails:

```zig
.{ "document.createElement('input').type", "text" }
```

Assuming it would also fail with the HTML `<input>`, then it's realistically important enough to fix I think. 

Any other wrong defaults should at least be documented, but probably also fixed.

src/browser/html/elements.zig
```diff
@@ -1144,3 +1228,144 @@ test "Browser.HTML.Element" {
...
+    try testProperty(&runner, "input", "disabled", "false", "false", &bool_valids, &.{});
+    try testProperty(&runner, "input", "maxLength", "-1", "0", &.{.{ .input = "5", .is_str = false }}, &.{.{ .input = "banana", .is_str = true }});
+    try testing.expectError(error.ExecutionError, runner.testCases(&.{.{ "elem_input.maxLength = -45", null }}, .{}));
+    try testProperty(&runner, "input", "name", "", "", &str_valids, &.{});
+    try testProperty(&runner, "input", "readOnly", "false", "false", &bool_valids, &.{});
+    try testProperty(&runner, "input", "size", "20", "20", &.{.{ .input = "5", .is_str = false }}, &.{.{ .input = "-26", .is_str = false }});
+    try testing.expectError(error.ExecutionError, runner.testCases(&.{.{ "elem_input.size = 0", null }}, .{}));
+    try testing.expectError(error.ExecutionError, runner.testCases(&.{.{ "elem_input.size = 'banana'", null }}, .{}));
+    try testProperty(&runner, "input", "src", "", "", &.{
+        .{ .input = "foo", .is_str = true, .expected = "https://lightpanda.io/foo" }, // TODO stitch should work with spaces -> %20
```
sjorsdonkers: @mookums stitch should probably work with spaces -> `%20` and such

src/browser/html/elements.zig
```diff
@@ -1144,3 +1228,144 @@ test "Browser.HTML.Element" {
...
+    input: []const u8,
+    is_str: bool,
+};
+
+fn testCreateElement(runner: *testing.JsRunner, comptime name: []const u8) !void {
+    try runner.testCases(&.{
+        .{ "let elem_" ++ name ++ " = document.createElement('" ++ name ++ "')", null },
+    }, .{});
+}
+// TODO reduce comptime
```
sjorsdonkers: TODO

src/browser/html/elements.zig
```diff
@@ -1144,3 +1228,144 @@ test "Browser.HTML.Element" {
         .{ "lyric.src", "15" },
     }, .{});
 }
+test "Browser.HTML.Element.propeties" {
```
sjorsdonkers: There should be no need for us to be writing these tests, as others have already written them.
Adopt WPT as required tests?

src/browser/html/elements.zig
```diff
@@ -617,6 +618,89 @@ pub const HTMLInputElement = struct {
...
+        return try parser.inputGetSize(self);
+    }
+    pub fn set_size(self: *parser.Input, size: i32) !void {
+        try parser.inputSetSize(self, size);
+    }
+    pub fn get_src(self: *parser.Input) ![]const u8 {
+        return try parser.inputGetSrc(self);
+    }
+    pub fn set_src(self: *parser.Input, src: []const u8, page: *Page) !void {
+        const new_src = try urlStitch(page.call_arena, src, page.url.raw);
```
sjorsdonkers: Having this layer here as well as in netsurf.zig does not add value IMO

src/browser/netsurf.zig
```diff
@@ -2593,3 +2585,183 @@ pub fn imageSetIsMap(image: *Image, is_map: bool) !void {
...
+}
+
+pub fn inputGetSrc(input: *Input) ![]const u8 {
+    var s_: ?*String = null;
+    const err = c.dom_html_input_element_get_src(input, &s_);
+    try DOMErr(err);
+    const s = s_ orelse return "";
+    return strToData(s);
+}
+// url should already be stitched!
```
sjorsdonkers, line 2738: Should we pass the allocator and base url to netsurf and stitch it here?

sjorsdonkers: I think we should just remove the layer of netsurf functions

src/browser/html/elements.zig
karlseguin: I agree...in theory. But I don't think it'll work in practice given the current state.

First, it's too slow. Second, there's too much failing / noise.   Say you're working on some input=text feature. It isn't clear which WPT tests you should run, but say you find them in 4 files...currently, they'll be 0/4.  You fix some things, but it's still 0/4. You can look at the individual cases, say that's 20/94 BEFORE your fix, and 21/94 AFTER your fix. Did you fix 1 thing? or did you fix 2 things and break 1 thing? Good luck figuring it out.

It could be neat if we could only add individual passing cases, to help us with regression, but that seems like a lot of effort to setup and maintain.

src/browser/netsurf.zig
karlseguin: Having URL resolution at some base layer would be nice, since the behavior seems pretty consistent across multiple APIs. Netsurf knows the document url; it would be nice if it took care of this. It doesn't AFAIK.

src/browser/html/elements.zig
sjorsdonkers: I agree, the WPT are not structured in a way that we could leverage in this way. I'm looking what other browsers are doing for inspiration.
I'm also researching what our options are regarding code and test generation from a schema as most follow a regular pattern. Though, there are also more complex relational constraints event with simple properties.

(It could be possible to handle all schema in the netsurf.zig layer and relational contraints in the HtmlElement layer. Not sure what the benefit of that separation would be. That would also mean that `Element.get/setAttribute()` would need to go through the HtmlElement layer (which it probably should anyways), but that is just a concern for later.)

src/browser/netsurf.zig
sjorsdonkers: I'll leave it consistent with the other url functions for now, so we'll deal with it later

PR #730, opened 2025-05-30
Fix HTMLImageElement

karlseguin:
HTMLImageElement is the correct class name. However, it has a "legacy factory": Image (i.e. new Image()).

### DIFF
--- a/src/browser/html/elements.zig
+++ b/src/browser/html/elements.zig
@@ -57,6 +57,7 @@ pub const Interfaces = .{
     HTMLHtmlElement,
     HTMLIFrameElement,
     HTMLImageElement,
+    HTMLImageElement.Factory,
     HTMLInputElement,
     HTMLLIElement,
     HTMLLabelElement,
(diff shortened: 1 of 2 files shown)


src/browser/html/elements.zig
```diff
@@ -611,6 +603,21 @@ pub const HTMLImageElement = struct {
     pub fn set_isMap(self: *parser.Image, is_map: bool) !void {
         try parser.imageSetIsMap(self, is_map);
     }
+
+    pub const Factory = struct {
```
sjorsdonkers, line 607: How about calling it `Alias`?
```suggestion
    pub const Alias = struct {
```

karlseguin: Not quite an alias though, not per the spec and not how it's implemented. The most obvious practical difference is:

```js
new Image() // valid;
new HTMLImageElement() // invalid...which I find ridiculous, but...
```

But, also, `HTMLImageElement != Image`

sjorsdonkers: I see, it's a strong typedef/alias, I also have seen it be called distinct variant, but I don't like that one.
Factory is somewhat implying it is creating an HTMLImageElement.
Up to you :) 🤷 I'd probably go with StrongAlias or Alias (and have the strong part be implicit). Probably StrongAlias as that would encode the knowledge of your reply.

PR #739, opened 2025-06-02
Fix url constructor

karlseguin:
url, base were being joined in the wrong order. Switch to using URL.stitch if a base is given.

### DIFF
--- a/src/url.zig
+++ b/src/url.zig
@@ -84,7 +84,7 @@ pub const URL = struct {
     }
 
     const StitchOpts = struct {
-        alloc: AllocWhen = .if_needed,
+        alloc: AllocWhen = .always,
 
         const AllocWhen = enum {
             always,
--- a/src/url.zig
+++ b/src/url.zig
@@ -83,12 +83,28 @@ pub const URL = struct {
         return WebApiURL.init(allocator, self.uri);
     }
 
+    const StitchOpts = struct {
+        alloc: AllocWhen = .if_needed,
+
+        const AllocWhen = enum {
+            always,
+            if_needed,
+        };
+    };
     /// Properly stitches two URL fragments together.
     ///
     /// For URLs with a path, it will replace the last entry with the src.
     /// For URLs without a path, it will add src as the path.
-    pub fn stitch(allocator: Allocator, src: []const u8, base: []const u8) ![]const u8 {
+    pub fn stitch(
+        allocator: Allocator,
+        src: []const u8,
+        base: []const u8,
+        opts: StitchOpts,
+    ) ![]const u8 {
         if (base.len == 0) {
+            if (opts.alloc == .always) {
+                return allocator.dupe(u8, src);
+            }
             return src;
         }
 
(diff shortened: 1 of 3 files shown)


src/url.zig
```diff
@@ -83,12 +83,28 @@ pub const URL = struct {
...
     /// For URLs without a path, it will add src as the path.
-    pub fn stitch(allocator: Allocator, src: []const u8, base: []const u8) ![]const u8 {
+    pub fn stitch(
+        allocator: Allocator,
+        src: []const u8,
+        base: []const u8,
+        opts: StitchOpts,
+    ) ![]const u8 {
         if (base.len == 0) {
+            if (opts.alloc == .always) {
```
sjorsdonkers, line 105: fyi, in another PR and empty src is also added. Feel free to adopt in this PR

https://github.com/lightpanda-io/browser/pull/721/files#diff-e84dfe2cc8a38ad62da56117bfcad3d57569be91619624b3383dded4c5aae319

src/url.zig
```diff
@@ -161,7 +177,7 @@ test "URL: Stitching Base & Src URLs (Basic)" {
 
     const base = "https://www.google.com/xyz/abc/123";
     const src = "something.js";
-    const result = try URL.stitch(allocator, src, base);
+    const result = try URL.stitch(allocator, src, base, .{});
```
sjorsdonkers, line 180: Should we remove the `defer allocator.free(result);`
Statements as I would not recommend them (crashes if empty, hence this PR).
Alternatively could  set the default to `.always`, but I find it generally better to not cleanup individual items.

karlseguin: If you remove it here, the test will fail because of a memory leak.

There is a `testing.arena_allocator` that can be used. You can place `defer testing.reset()`  at the top and then, it doesn't matter. If it allocates, great, it's freed. If it doesn't, then no harm.

But that's really if you don't care..or maybe the testing code itself is doing allocations. It's kind of useful here, because part of the behavior of this function is the lifetime of the returned string. If you used the testing arena, then you wouldn't catch a regression - calling code might expect the returned string to be allocated via the allocator, and suddenly it isn't and has a lifetime tied to an input.

sjorsdonkers: OK.
Note that the lifetime is tied to the input if the input is `""`, so generally a user should not use stitch and defer the free the string unless it is set to `.always` or the caller has checked the input is str (in which case why still call the function?).
Not my favorite memory contract.

karlseguin: See my other comment, we need to clean up URLs at some point.

sjorsdonkers: Right got it, to be clear my comment was highlighting that a functions returning a result with 2 or 3 possible lifetimes is not a preferable design pattern as it is only beneficial if all 3 lifetimes are long enough for the users' purpose. In every other case the user needs to:
1. Copy the string even if neither base nor src was empty (unnecessary most of the time).
2. Check beforehand if base or src is empty and not call the function if not needed.

Since the `.if_needed` case is exceptional (src is often user input) and an optimization, I think `.always` is a better default.

karlseguin: I agree `.always` is a better default.

PR #740, opened 2025-06-02
Fix anchor href

karlseguin:


src/browser/html/elements.zig
```diff
@@ -289,10 +291,9 @@ pub const HTMLAnchorElement = struct {
         try parser.anchorSetHref(self, href);
     }
 
-    // TODO return a disposable string
     pub fn get_hostname(self: *parser.Anchor, page: *Page) ![]const u8 {
         var u = try url(self, page);
-        return try page.arena.dupe(u8, u.get_hostname());
+        return u.get_hostname();
```
sjorsdonkers, line 296: I think we could construct the url here with `.if_needed`, no?

sjorsdonkers (approved):

src/browser/html/elements.zig
karlseguin: The whole URL constructing _per_ getter is a bit much.  I guess it isn't called too often, plus it's maybe just as likely to get mutated.

Everything touching URLs needs a rewrite at some point. It's a bit of a mess, starting with std.Uri.

PR #742, opened 2025-06-02
Focus and active element

karlseguin:
Hmm..I'm sure I opened this PR already..

This is in preparation for keyboard input from CDP.


src/browser/html/elements.zig
```diff
@@ -148,6 +148,18 @@ pub const HTMLElement = struct {
         });
         _ = try parser.elementDispatchEvent(@ptrCast(e), @ptrCast(event));
     }
+
+    pub fn _focus(e: *parser.ElementHTML, page: *Page) !void {
```
sjorsdonkers: May as well add the Options parameter and ignore them as we wouldn't use them

src/browser/page.zig
```diff
@@ -95,6 +95,8 @@ pub const Page = struct {
     // indicates intention to navigate to another page on the next loop execution.
     delayed_navigation: bool = false,
 
+    active_element: ?*parser.Element = null,
```
sjorsdonkers: It is not clear to me yet why the field and the methods are in Page as opposed to in Document.
If it is a convenience thing I would definitely move it.
If I'm not mistaken every iFrame has it's own activeElement

In Document the activeElement would never be `null`.

src/browser/html/elements.zig
karlseguin: You're right. I didn't check. Didn't know it had any. Fixed.

src/browser/page.zig
karlseguin: I wasn't sure where to put it. I moved it to, having to use the "NodeWrapper" path.

You say it would never be null, but `get_documentElement` returns an optional...so I kept the signature as optional.

sjorsdonkers: ah right

PR #751, opened 2025-06-03
add noop window.scrollTo

karlseguin:


src/browser/html/window.zig
```diff
@@ -261,11 +261,26 @@ pub const Window = struct {
...
     // each call.
-    pub fn _getComputedStyle(_: *Window, element: *parser.Element, pseudo_element: ?[]const u8) !CSSStyleDeclaration {
+    pub fn _getComputedStyle(_: *const Window, element: *parser.Element, pseudo_element: ?[]const u8) !CSSStyleDeclaration {
         _ = element;
         _ = pseudo_element;
         return .empty;
     }
+
+    const ScrollToOpts = union(enum) {
+        x: i32,
```
sjorsdonkers, line 271: interesting solution

karlseguin: You mean the fact that it supports:

```(x, y)```

or

```(opts)```

So I did `(union, ?y)`.

Don't know how else to do it.

sjorsdonkers: yea, works. Good to know

PR #757, opened 2025-06-04
Fix crash when event target is the window.

karlseguin:
On page load, emitted by the page, the target is the window, but it's improperly cast since the pointer is actually `window.base`. This is going to be a problem in general for any Zig type dispatched as a target, but the Window one is the most obvious and the easiest to fix. If this issue comes up with other types, we'll need to come up with a more robust solution.

### DIFF
--- a/src/browser/dom/event_target.zig
+++ b/src/browser/dom/event_target.zig
@@ -33,9 +33,15 @@ pub const EventTarget = struct {
     pub const Self = parser.EventTarget;
     pub const Exception = DOMException;
 
-    pub fn toInterface(et: *parser.EventTarget) !Union {
-        // NOTE: for now we state that all EventTarget are Nodes
-        // TODO: handle other types (eg. Window)
+    pub fn toInterface(et: *parser.EventTarget, page: *Page) !Union {
+        // Not all targets are *parser.Nodes. page.zig emits a "load" event
+        // where the target is a Window, which cannot be cast directly to a node.
+        // Ideally, we'd remove this duality. Failing that, we'll need to embed
+        // data into the *parser.EventTarget should we need this for other types.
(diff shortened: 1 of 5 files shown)


src/browser/dom/event_target.zig
```diff
@@ -33,9 +33,15 @@ pub const EventTarget = struct {
...
-    pub fn toInterface(et: *parser.EventTarget) !Union {
-        // NOTE: for now we state that all EventTarget are Nodes
-        // TODO: handle other types (eg. Window)
+    pub fn toInterface(et: *parser.EventTarget, page: *Page) !Union {
+        // Not all targets are *parser.Nodes. page.zig emits a "load" event
+        // where the target is a Window, which cannot be cast directly to a node.
+        // Ideally, we'd remove this duality. Failing that, we'll need to embed
+        // data into the *parser.EventTarget should we need this for other types.
+        // For now, for the Window, which is a singleton, we can do this:
+        if (@intFromPtr(et) == @intFromPtr(&page.window.base)) {
```
sjorsdonkers, line 42: Personally I would make every struct a `extern struct` and ban `struct`.
Would also help with slowly taking over libDOM functionality.
IMO zig got the default wrong in this case.

karlseguin: Hmmm...I was going to say "I don't want to make that decision".

Then I thought, "well, we could at least do it on Window for this specific problem...try it out".

But ya..no. There's a bunch of restrictions on what you can put into an `extern struct`. You can't, for example, have slices. So `target: []const u8 = "",` isn't valid.

If this is the way we go, it's a bigger change / discussion.

sjorsdonkers: Oh my😱

PR #758, opened 2025-06-04
Fix activeElement

karlseguin:
https://github.com/lightpanda-io/browser/pull/742 was only correct in isolation.

The problem is that both html/document and dom/document are now attempting to attach state to the same *parser.Node. This is a problem because they're two different types of state. If an HTMLDocument is attached to the node, and you then try to access that as a DOMDocument, it'll crash.

I moved the dom/document state (active_element) into the html/document.

The other option is to introduce something like src/browser/state/document.zig and have both HTMLDocument and DOMDocument share/load that.

### DIFF
--- a/src/browser/dom/document.zig
+++ b/src/browser/dom/document.zig
@@ -42,7 +42,8 @@ pub const Document = struct {
     pub const prototype = *Node;
     pub const subtype = .node;
 
-    active_element: ?*parser.Element = null,
+    // DO NOT ADD STATE HERE
+    // ADD IT TO html/document
 
     pub fn constructor(page: *const Page) !*parser.DocumentHTML {
         const doc = try parser.documentCreateDocument(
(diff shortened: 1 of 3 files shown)


src/browser/dom/document.zig
```diff
@@ -246,8 +247,9 @@ pub const Document = struct {
     }
 
     pub fn get_activeElement(doc: *parser.Document, page: *Page) !?ElementUnion {
-        const self = try page.getOrCreateNodeWrapper(Document, @ptrCast(doc));
-        if (self.active_element) |ae| {
+        const HTMLDocument = @import("../html/document.zig").HTMLDocument;
+        const html_doc = try page.getOrCreateNodeWrapper(HTMLDocument, @ptrCast(doc));
```
sjorsdonkers, line 251:
```suggestion
        const doc_elem = try parser.documentGetDocumentElement(doc);
        const html_doc = try page.getOrCreateNodeWrapper(HTMLDocument, @ptrCast(doc_elem));
```
I'd expect something like this ^

karlseguin: No, that doesn't work. The DocumentElement is something different, different node.

karlseguin:
This issue impacts multiple cases. I'm going to try to "fix" them all.

PR #761, opened 2025-06-04
Improve usability of NodeWrapper

karlseguin:
The NodeWrapper pattern attaches a Zig instance to a libdom Node. That works in isolation, but for 1 given node, we might want to attach different instances.

For example, for an HTMLScriptElement we want to attach an `onError`, but for that same node viewed as an HTMLElement we want to a `CSSStyleDeclaration`. We can only have one. Currently, this code will crash if, for example, we create the embedded data as an HTMLScriptElement, then try to read the embedded data as an HTMLElement.

This PR introduces dedicated state class. So if you want the onError property, you no longer ask the NodeWrapper for an HTMLSCriptElement. Instead, you ask for a storage/HTMLElement.

Nothing fancy here, just memory-inefficient optional fields. If it gets out of hand, we'll think of something more clever.


sjorsdonkers (approved):
Could consider not making the optimization of defining multiple State structs and just use 1 for everything. So combine Document and HtmlElement state structs as well for now.

karlseguin:
> Could consider not making the optimization of defining multiple State structs and just use 1 for everything. So combine Document and HtmlElement state structs as well for now.

Actually, I think we have no choice. With the way I did it, we'll still run into the same problem as soon as we need something on both HTMLElement and Element or Node.

I've added a global `State`. This has the benefit that we can use a MemoryPool(State) which gets re-used between pages.

PR #768, opened 2025-06-05
setExtraHTTPHeaders

sjorsdonkers:
https://chromedevtools.github.io/devtools-protocol/tot/Network/#method-setExtraHTTPHeaders

Confirmed working via puppeteer script. 
```
await page.setExtraHTTPHeaders({ 'food': 'bard', 'foods': 'bars' })
console.log((await page.goto('https://example.org/')));
```
image

Would want to test it from puppeteer by inspecting the headers of the response, but `response` is null for LP (works with chromium)
```
let response = await page.goto('https://example.org/');
console.log(response.request().headers());
```
A good way to test suggestion is welcome, can't Page.navigate afaik

### DIFF
--- a/src/cdp/cdp.zig
+++ b/src/cdp/cdp.zig
@@ -71,6 +71,9 @@ pub fn CDPT(comptime TypeProvider: type) type {
         // Used for processing notifications within a browser context.
         notification_arena: std.heap.ArenaAllocator,
 
+        // Extra headers to add to all requests. TBD under which conditions this should be reset.
+        extra_headers: std.ArrayListUnmanaged(std.http.Header) = .empty,
+
         const Self = @This();
 
         pub fn init(app: *App, client: TypeProvider.Client) !Self {
--- a/src/cdp/domains/network.zig
+++ b/src/cdp/domains/network.zig
@@ -18,6 +18,7 @@
 
 const std = @import("std");
 const Notification = @import("../../notification.zig").Notification;
+const log = @import("../../log.zig");
 
 const Allocator = std.mem.Allocator;
 
@@ -26,12 +27,14 @@ pub fn processMessage(cmd: anytype) !void {
         enable,
         disable,
         setCacheDisabled,
+        setExtraHTTPHeaders,
     }, cmd.input.action) orelse return error.UnknownMethod;
 
     switch (action) {
         .enable => return enable(cmd),
         .disable => return disable(cmd),
         .setCacheDisabled => return cmd.sendResult(null, .{}),
+        .setExtraHTTPHeaders => return setExtraHTTPHeaders(cmd),
     }
 }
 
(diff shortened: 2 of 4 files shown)


src/http/client.zig
```diff
@@ -736,10 +736,11 @@ pub const Request = struct {
         }
         self._notified_start = true;
         notification.dispatch(.http_request_start, &.{
+            .arena = self.arena,
             .id = self.id,
             .url = self.request_uri,
             .method = self.method,
-            .headers = self.headers.items,
+            .headers = &self.headers,
```
sjorsdonkers, line 743: Headers are now send as a modifiable array

src/cdp/domains/network.zig
```diff
@@ -59,6 +83,21 @@ pub fn httpRequestStart(arena: Allocator, bc: anytype, request: *const Notificat
     const target_id = bc.target_id orelse unreachable;
     const page = bc.session.currentPage() orelse unreachable;
 
+    // Modify request with extra CDP headers
+    const original_len = request.headers.items.len;
+    try request.headers.ensureTotalCapacity(arena, original_len + cdp.extra_headers.items.len);
+    outer: for (cdp.extra_headers.items) |extra| {
+        for (request.headers.items[0..original_len]) |*existing_header| {
+            if (std.mem.eql(u8, existing_header.name, extra.name)) {
```
sjorsdonkers: Double loop, but number of headers is small, using hashmap for either is likely slower.

src/cdp/cdp.zig
```diff
@@ -71,6 +71,9 @@ pub fn CDPT(comptime TypeProvider: type) type {
         // Used for processing notifications within a browser context.
         notification_arena: std.heap.ArenaAllocator,
 
+        // Extra headers to add to all requests. TBD under which conditions this should be reset.
+        extra_headers: std.ArrayListUnmanaged(std.http.Header) = .empty,
```
sjorsdonkers, line 75: My current understanding is that this has the lifetime of a browserContext, the lifetime is not clear as it is Network is ambiguous. I'm sure this should outlive a page.navigate. I think every Target.createTarget the lifetime would create a new instance.

src/cdp/domains/network.zig
```diff
@@ -59,6 +83,21 @@ pub fn httpRequestStart(arena: Allocator, bc: anytype, request: *const Notificat
...
+                // If the header already exists, we overwrite it
+                log.debug(.cdp, "request header overwritten", .{ .name = extra.name });
+                existing_header.value = extra.value;
+                continue :outer;
```
karlseguin: Not that it matters ,but pretty sure you can just `break` here.  Not sure if you find `continue: label` more explicit than `break`, since I guess it might not be immediately obviously that you're just breaking the inner loop.

src/cdp/domains/network.zig
```diff
@@ -59,6 +83,21 @@ pub fn httpRequestStart(arena: Allocator, bc: anytype, request: *const Notificat
     const target_id = bc.target_id orelse unreachable;
     const page = bc.session.currentPage() orelse unreachable;
 
+    // Modify request with extra CDP headers
+    const original_len = request.headers.items.len;
+    try request.headers.ensureTotalCapacity(arena, original_len + cdp.extra_headers.items.len);
```
karlseguin:
```suggestion
    try request.headers.ensureTotalCapacity(request.arena, original_len + cdp.extra_headers.items.len);
```

src/cdp/cdp.zig
karlseguin: i tested this with chrome and the lifetime is shorter than the browser context. I believe it's tied to the (CDP) SessionId. In theory, you could hold map sessionid=>headers, but since we only allow 1 active session at a time, I believe the simplest solution is to clear `extra_headers` in `attachToTarget`.

Thus, if we internally navigate from page to page (i.e. button click, form submit, ...) the headers continue to be used because it's the same target/sessionid. But if the driver closes the page and opens a new one - within the same browser context - the headers are removed. This is what I observed in chrome.

src/cdp/domains/network.zig
sjorsdonkers: It is an upsert on each iteration of the loop, a simple break mean we update and append. I prefer the current code over:
```zig
for (extra_headers) |extra| {
    var found = false;
    for (headers) |header| {
        if (extra = header) {
            found = true;
            break;
         }
    }
    if (found) continue;
    headers.appendAssumeCapacity(extra);
}
```
 But then the problem with it is that it appears to be not obvious, so should improve for clarity. So maybe best to create a `put` function indicating hashmap put semantics.

src/cdp/domains/network.zig
sjorsdonkers: Nice! [bacef41](https://github.com/lightpanda-io/browser/pull/768/commits/bacef41a3b37192fea33327daa61cfec6b585ccb)

src/cdp/domains/network.zig
sjorsdonkers: Changed: [bacef41](https://github.com/lightpanda-io/browser/pull/768/commits/bacef41a3b37192fea33327daa61cfec6b585ccb)

src/cdp/cdp.zig
sjorsdonkers: Added: [bacef41](https://github.com/lightpanda-io/browser/pull/768/commits/bacef41a3b37192fea33327daa61cfec6b585ccb)

PR #772, opened 2025-06-06
Import html_element

sjorsdonkers:
Turns HtmlElement from an `opaque` type to an `extern struct`, this also bring in some others most notable `Node` and `HtmlDocument` are now also no longer opaque.

This change allows us to embed these DOM objects directly into zig structs.

### DIFF
--- a/src/browser/netsurf.zig
+++ b/src/browser/netsurf.zig
@@ -26,6 +26,8 @@ const c = @cImport({
     @cInclude("events/event.h");
     @cInclude("events/mouse_event.h");
     @cInclude("utils/validate.h");
+    @cInclude("html/html_element.h");
+    @cInclude("html/html_document.h");
 });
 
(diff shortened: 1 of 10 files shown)


src/browser/netsurf.zig
```diff
@@ -26,6 +26,7 @@ const c = @cImport({
     @cInclude("events/event.h");
     @cInclude("events/mouse_event.h");
     @cInclude("utils/validate.h");
+    @cInclude("html/html_element.h");
```
sjorsdonkers, line 29: This is the only change

src/browser/netsurf.zig
```diff
@@ -26,6 +26,8 @@ const c = @cImport({
...
+    @cInclude("html/html_document.h");
```
sjorsdonkers, line 30: This is the only change

PR #773, opened 2025-06-09
Add basic support for key events

karlseguin:
Support CDP's Input.dispatchKeyEvent and DOM key events. Currently only keydown is supported and expects every key to be a displayable character.

It turns out that manipulating the DOM via key events isn't great because the behavior really depends on the cursor. So, to do this more accurately, we'd have to introduce some concept of a cursor.

Personally, I don't think we'll run into many pages that are purposefully using keyboard events. But driver (puppeteer/playwright) scripts might be another issue.

### DIFF
--- a/src/browser/netsurf.zig
+++ b/src/browser/netsurf.zig
@@ -2834,3 +2834,11 @@ pub fn inputSetValue(input: *Input, value: []const u8) !void {
     const err = c.dom_html_input_element_set_value(input, try strFromData(value));
     try DOMErr(err);
 }
+
+pub fn buttonGetType(button: *Button) ![]const u8 {
+    var s_: ?*String = null;
+    const err = c.dom_html_button_element_get_type(button, &s_);
+    try DOMErr(err);
+    const s = s_ orelse return "button";
+    return strToData(s);
+}
--- a/src/browser/netsurf.zig
+++ b/src/browser/netsurf.zig
@@ -25,6 +25,7 @@ const c = @cImport({
     @cInclude("events/event_target.h");
     @cInclude("events/event.h");
     @cInclude("events/mouse_event.h");
+    @cInclude("events/keyboard_event.h");
     @cInclude("utils/validate.h");
     @cInclude("html/html_element.h");
     @cInclude("html/html_document.h");
@@ -862,6 +863,59 @@ pub fn mouseEventInit(evt: *MouseEvent, typ: []const u8, opts: MouseEventOpts) !
 
 pub fn mouseEventDefaultPrevented(evt: *MouseEvent) !bool {
     return eventDefaultPrevented(@ptrCast(evt));
+}
+
+// KeyboardEvent
+
+pub const KeyboardEvent = c.dom_keyboard_event;
+
+pub fn keyboardEventCreate() !*KeyboardEvent {
+    var evt: ?*KeyboardEvent = undefined;
+    const err = c._dom_keyboard_event_create(&evt);
+    try DOMErr(err);
+    return evt.?;
+}
+
+pub fn keyboardEventDestroy(evt: *KeyboardEvent) void {
+    c._dom_keyboard_event_destroy(evt);
+}
+
+const KeyboardEventOpts = struct {
+    key: []const u8,
+    code: []const u8,
+    bubbles: bool = false,
+    cancelable: bool = false,
+    ctrl: bool = false,
+    alt: bool = false,
+    shift: bool = false,
+    meta: bool = false,
+};
+
+pub fn keyboardEventInit(evt: *KeyboardEvent, typ: []const u8, opts: KeyboardEventOpts) !void {
+    const s = try strFromData(typ);
+    const err = c._dom_keyboard_event_init(
+        evt,
+        s,
+        opts.bubbles,
+        opts.cancelable,
+        null, // dom_abstract_view* ?
+        try strFromData(opts.key),
+        try strFromData(opts.code),
+        0, // location 0 == standard
+        opts.ctrl,
+        opts.shift,
+        opts.alt,
+        opts.meta,
+        false, // repease
+        false, // is_composiom
+    );
+    try DOMErr(err);
+}
+
+pub fn keyboardEventGetKey(evt: *KeyboardEvent) ![]const u8 {
+    var s: ?*String = undefined;
+    _ = c._dom_keyboard_event_get_key(evt, &s);
+    return strToData(s.?);
 }
 
 // NodeType
@@ -2391,6 +2445,11 @@ pub fn textareaGetValue(textarea: *TextArea) ![]const u8 {
     try DOMErr(err);
     const s = s_ orelse return "";
     return strToData(s);
+}
+
+pub fn textareaSetValue(textarea: *TextArea, value: []const u8) !void {
+    const err = c.dom_html_text_area_element_set_value(textarea, try strFromData(value));
+    try DOMErr(err);
 }
 
 // Select
--- a/src/browser/page.zig
+++ b/src/browser/page.zig
@@ -80,6 +80,7 @@ pub const Page = struct {
 
     microtask_node: Loop.CallbackNode,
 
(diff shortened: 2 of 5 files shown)


src/browser/netsurf.zig
```diff
@@ -862,6 +863,59 @@ pub fn mouseEventDefaultPrevented(evt: *MouseEvent) !bool {
...
+        false, // is_composiom
+    );
+    try DOMErr(err);
+}
+
+pub fn keyboardEventGetKey(evt: *KeyboardEvent) ![]const u8 {
+    var s: ?*String = undefined;
+    _ = c._dom_keyboard_event_get_key(evt, &s);
+    return strToData(s.?);
+}
```
sjorsdonkers, line 919: Does this return `"B"` for `shift + b`, dead keys, unidentified, etc. correctly?
See: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key

src/browser/netsurf.zig
```diff
@@ -2391,6 +2445,11 @@ pub fn textareaGetValue(textarea: *TextArea) ![]const u8 {
     return strToData(s);
 }
 
+pub fn textareaSetValue(textarea: *TextArea, value: []const u8) !void {
```
sjorsdonkers, line 2450: Note `value` has a defaultValue. So likely needs the same care as HtmlInputElement.Value.
It may be implemented correctly [here](https://github.com/lightpanda-io/libdom/blob/master/src/html/html_text_area_element.c), but it is not obvious to me.

I think it may be best to also implement defaultValue as they considered the same property.

src/browser/xhr/form_data.zig
```diff
@@ -162,7 +162,7 @@ fn collectForm(form: *parser.Form, submitter_: ?*parser.ElementHTML, page: *Page
                     }
                     submitter_included = true;
                 }
-                const value = (try parser.elementGetAttribute(element, "value")) orelse "";
+                const value = try parser.inputGetValue(@ptrCast(element));
```
sjorsdonkers, line 165: cool

src/cdp/domains/input.zig
```diff
@@ -21,14 +21,60 @@ const Page = @import("../../browser/page.zig").Page;
...
     }
 }
 
+// https://chromedevtools.github.io/devtools-protocol/tot/Input/#method-dispatchKeyEvent
+fn dispatchKeyEvent(cmd: anytype) !void {
+    const params = (try cmd.params(struct {
+        type: Type,
+        key: []const u8,
+        code: []const u8,
+        modifiers: u4,
```
sjorsdonkers:
```suggestion
        type: Type,
        key: []const u8 = "",
        code: []const u8 = "",
        modifiers: u4 = 0,
```
Making sure that the optional parameters have defaults

src/browser/page.zig
```diff
@@ -593,6 +601,76 @@ pub const Page = struct {
...
+    fn _keydownCallback(page: *Page, event: *parser.Event) !void {
+        const kbe: *parser.KeyboardEvent = @ptrCast(event);
+        const target = (try parser.eventTarget(event)) orelse return;
+        const node = parser.eventTargetToNode(target);
+        const tag = (try parser.nodeHTMLGetTagType(node)) orelse return;
+        switch (tag) {
+            .input => {
+                const element: *parser.Element = @ptrCast(node);
+                const input_type = (try parser.elementGetAttribute(element, "type")) orelse "text";
+                if (std.mem.eql(u8, input_type, "text")) {
```
sjorsdonkers:
```suggestion
                const input: *parser.Input= @ptrCast(node);
                const input_type = try parser.inputGetType(input);
                if (std.mem.eql(u8, input_type, "text")) {
```
`elementGetAttribute` is kinda broken in libDom, I think we should avoid using it where possible.

Perhaps you are using it here, because in the test use case the value was set via `elementSetAttribute` which has the same broken behavior.. In which case I think we should at least leave a comment.

src/browser/page.zig
```diff
@@ -593,6 +601,76 @@ pub const Page = struct {
         }
     }
 
+    pub const KeyboardEvent = struct {
```
sjorsdonkers, line 606: Should this be a `Event -> UIEvent -> KeyboardEvent`?

src/browser/page.zig
```diff
@@ -593,6 +601,76 @@ pub const Page = struct {
...
+                    const value = try parser.inputGetValue(@ptrCast(element));
+                    const new_key = try parser.keyboardEventGetKey(kbe);
+                    const new_value = try std.mem.concat(page.arena, u8, &.{ value, new_key });
+                    try parser.inputSetValue(@ptrCast(element), new_value);
+                }
+            },
+            .textarea => {
+                const value = try parser.textareaGetValue(@ptrCast(node));
+                const new_key = try parser.keyboardEventGetKey(kbe);
+                const new_value = try std.mem.concat(page.arena, u8, &.{ value, new_key });
```
sjorsdonkers: This is probably fine I just can't find where it is defined why this should be happening.
Perhaps some comments would be helpful to help the reader.

src/browser/netsurf.zig
karlseguin: Yes, it'll handle capital letters.

I'm trying to get a specific use-case working: puppeteer's `page.type` function. In most cases, this just gets a string, so weird characters aren't really possible.

A lot of direct keypresses are sent with the `rawKeyDown` type in CDP - which we'll just ignore.

in a more recent commit, I ignore "Dead" and I special-case "Enter", since that's commonly used to submit a form.

src/browser/page.zig
karlseguin: This isn't the DOM API, it's the public page API.

I think what you mean would go in `browser/events/keyboard_event.zig`. We currently have a `MouseEvent`, but I guess no one has run into a site that explicitly raises `KeyboardEvent`. It's probably pretty rare and we might get away with never having to do it.

src/browser/page.zig
karlseguin: I tried to find where in the spec's it says that clicking on a hyperlink should load a new page, but I couldn't (seriously). I also couldn't find where it says that when an input or textarea is the event target, its value should be edited (or how that edit takes place).

src/browser/page.zig
karlseguin: I made the change in a separate commit..I changed other cases and I added a `buttonGetType`

src/browser/netsurf.zig
karlseguin: From my testing, both TextArea and Input will return something like `value orelse defaultValue` if the value hasn't been set. So that's right.

However, they both only have `defaultValue` if `defaultValue` is explicitly set. Given this html:

```html
<textarea id=ta>abc</textarea>
```

Then this JS should return `abc`:

```js
document.getElementById("ta").value;
```

But it returns empty.

Only if we explicitly set `defaultValue`: 

```js
document.getElementById("ta").defaultValue = 'abc'
```

Does everything work as expected.

In other words, I guess when we're parsing the page, we should be looking for any input/textarea and setting their defaultValue as needed since libdom doesn't do it for us.

sjorsdonkers: Yes, generally the logic is something like `x orelse defaultX orelse initialX` where initialX is the value as defined if not set upon construction, not to be confused with the `afterInvalidX` and then there are `errorX`s, which when set throw an error instead of using the `afterInvalidX`.

Other surprising behavior is that `Element.setAttribute(element, 'value', 'abc')` will set `defaultValue`, not `value`.

> ```
> <textarea id=ta>abc</textarea>
> document.getElementById("ta").value;
> ```
That this does not work is surprising, but libDOM struggles a lot with these details.
I'm a fair way towards being able to implement `HtmlElement`s without C in Zig such that we can have much easier control. Mainly I just need to work out which DOM events to send (deprecated for external usage, but we still use them internally). Tho I would need to look into why that `abc` is dropped in the first example.

PR #777, opened 2025-06-11
CDP Cookies

sjorsdonkers:
[Network.setCookie](https://chromedevtools.github.io/devtools-protocol/tot/Network/#method-setCookie)
[Network.setCookies](https://chromedevtools.github.io/devtools-protocol/tot/Network/#method-setCookies)
deprecated ~~[Network.getAllCookies](https://chromedevtools.github.io/devtools-protocol/tot/Network/#method-getAllCookies)~~
[Network.deleteCookies](https://chromedevtools.github.io/devtools-protocol/tot/Network/#method-deleteCookies)
[Network.getCookies](https://chromedevtools.github.io/devtools-protocol/tot/Network/#method-getCookies)
[Network.clearBrowserCookies](https://chromedevtools.github.io/devtools-protocol/tot/Network/#method-clearBrowserCookies)
deprecated ~~[Network.canClearBrowserCookies](https://chromedevtools.github.io/devtools-protocol/tot/Network/#method-canClearBrowserCookies)~~
deprecated  ~~[Page.deleteCookie](https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-deleteCookie)~~

[Storage.getCookies](https://chromedevtools.github.io/devtools-protocol/tot/Storage/#method-getCookies)
[Storage.setCookies](https://chromedevtools.github.io/devtools-protocol/tot/Storage/#method-setCookies)
[Storage.clearCookies](https://chromedevtools.github.io/devtools-protocol/tot/Storage/#method-clearCookies)

NOTE:
We currently do not support all cookie parameters, this PR only sets and uses the existing ones.
Existing cookies behavior has been modified as follows:
- Domains and paths are percent encoded in case they were not.
- Domains are converted to lower case
- HttpOnly cookies are no longer passed or accepted from JS (document.cookie)
- expires now supports googles dashes in the date, and is stored as f64

Puppeteer test: https://github.com/lightpanda-io/demo/pull/51


src/browser/storage/cookie.zig
```diff
@@ -365,7 +372,7 @@ pub const Cookie = struct {
                     if (std.mem.endsWith(u8, host, value) == false) {
                         return error.InvalidDomain;
                     }
-                    domain = value;
+                    domain = value; // TODO to lower case: https://www.rfc-editor.org/rfc/rfc6265#section-5.2.3
```
sjorsdonkers: TODO CDP functions should also make the domain whether from domain or from url lowercase

sjorsdonkers: done [d03b73f](https://github.com/lightpanda-io/browser/pull/777/commits/d03b73fbd7f29d4135641d4ebd4892b35c7797c4)

src/cdp/domains/storage.zig
```diff
@@ -0,0 +1,236 @@
...
+    try w.beginObject();
+    {
+        try w.objectField("name");
+        try w.write(cookie.name);
+
+        try w.objectField("value");
+        try w.write(cookie.value);
+
+        try w.objectField("domain");
+        try w.write(cookie.domain); // Should we hide a leading dot?
```
sjorsdonkers, line 206: We probably, should. I'm not sure if it matters.
To support it we probably need to store the domain with a flag whether it came from url or was set explicitly.

src/cdp/domains/storage.zig
```diff
@@ -0,0 +1,236 @@
...
+            return error.OutOfMemory;
+        };
+    }
+
+    fn writeCookies(self: CookieWriter, w: anytype) !void {
+        try w.beginArray();
+        if (self.urls) |urls| {
+            for (self.cookies) |*cookie| {
+                for (urls) |*url| {
+                    if (cookie.appliesTo(url, false, false)) { // TBD same_site, should we compare to the pages url?
```
sjorsdonkers: When getting cookies we can only provide a list of urls, not other configuration (or get all cookies).
We probably should not filter out cookies that are not related to the current page, also because this is not part of the Page api.

src/browser/storage/cookie.zig
```diff
@@ -423,6 +300,100 @@ pub const Cookie = struct {
         };
     }
 
+    pub fn parsePath(arena: Allocator, uri: ?*const std.Uri, explicit_path: ?[]const u8) ![]const u8 {
+        // path attribute value either begins with a '/' or we
+        // ignore it and use the "default-path" algorithm
+        if (explicit_path) |path| {
+            if (path.len > 0 and path[0] == '/') {
+                return try arena.dupe(u8, path);
```
sjorsdonkers, line 321: Should we make sure it ends with a `/`?
Same as when we extract it from the URL
Else, should we take this difference into account when comparing cookie paths

src/cdp/domains/network.zig
```diff
@@ -71,6 +82,112 @@ fn setExtraHTTPHeaders(cmd: anytype) !void {
...
+const Cookie = @import("../../browser/storage/storage.zig").Cookie;
+
+// Only matches the cookie on provided parameters
+fn cookieMatches(cookie: *const Cookie, name: []const u8, domain: ?[]const u8, path: ?[]const u8) bool {
+    if (!std.mem.eql(u8, cookie.name, name)) return false;
+
+    if (domain) |domain_| {
+        const c_no_dot = if (std.mem.startsWith(u8, cookie.domain, ".")) cookie.domain[1..] else cookie.domain;
+        const d_no_dot = if (std.mem.startsWith(u8, domain_, ".")) domain_[1..] else domain_;
+        if (!std.mem.eql(u8, c_no_dot, d_no_dot)) return false;
```
sjorsdonkers, line 94: Related to the other domain `.` prefix comment.
Is this the right way to compare domains for the delete scenario?
What about for detection duplicate cookies?

src/cdp/domains/storage.zig
sjorsdonkers: Note also that Network.getCookies will default to the Pages url if no urls are provided.
`same_site` and `navigation` are hard coded to `true` now

src/cdp/domains/storage.zig
```diff
@@ -0,0 +1,304 @@
...
+
+        // TODO size
+
+        try w.objectField("httpOnly");
+        try w.write(cookie.http_only);
+
+        try w.objectField("secure");
+        try w.write(cookie.secure);
+
+        // TODO session
```
karlseguin: Session is true when `cookie.expires == null`

src/cdp/domains/storage.zig
```diff
@@ -0,0 +1,304 @@
...
+        browserContextId: ?[]const u8 = null,
+    })) orelse return error.InvalidParams;
+
+    if (params.browserContextId) |browser_context_id| {
+        if (std.mem.eql(u8, browser_context_id, bc.id) == false) {
+            return error.UnknownBrowserContextId;
+        }
+    }
+    bc.session.cookie_jar.removeExpired(null);
+    const writer = CookieWriter{ .cookies = bc.session.cookie_jar.cookies.items };
```
karlseguin, line 68: This doesn't seem right. `Storge.getCookies` has the unhelpful description of "Returns all browser cookies."  I don't know what "all browser cookies" means, but this behavior is different than Chrome.

If you run:

```js
await page.goto('https://www.google.com/');
console.log(await page.cookies());    // Network.getCookies
console.log(await browser.cookies()); // Storage.getCookies
```

Then in Chrome, this will return 2 cookies for the first one, and 0 for the second. But with this implementation, it returns 2 cookies for both. I wonder if "brower.cookies", aka `Storage.getCookies` is only those cookies set directly via `Storage.setCookies`? Either need to figure out how this is implemented in Chrome, or to play with more permutations to figure it...I tried to see if puppeteer had an explanation about the difference, but also nothing.

src/cdp/domains/storage.zig
```diff
@@ -0,0 +1,304 @@
...
+    // sourcePort: Temporary ability and it will be removed from CDP
+    partitionKey: ?CookiePartitionKey = null,
+};
+
+pub fn setCdpCookie(cookie_jar: *CookieJar, param: CdpCookie) !void {
+    if (param.priority != .Medium or param.sameParty != null or param.sourceScheme != null or param.partitionKey != null) {
+        return error.NotYetImplementedParams;
+    }
+    if (param.name.len == 0) return error.InvalidParams;
+    if (param.value.len == 0) return error.InvalidParams;
```
karlseguin: Looks like Chrome allows an empty value.

src/cdp/domains/storage.zig
```diff
@@ -0,0 +1,304 @@
...
+
+    var arena = std.heap.ArenaAllocator.init(cookie_jar.allocator);
+    errdefer arena.deinit();
+    const a = arena.allocator();
+
+    // NOTE: The param.url can affect the default domain, (NOT path), secure, source port, and source scheme.
+    const uri = if (param.url) |url| std.Uri.parse(url) catch return error.InvalidParams else null;
+    const uri_ptr = if (uri) |*u| u else null;
+    const domain = try Cookie.parseDomain(a, uri_ptr, param.domain);
```
karlseguin, line 142: Probably doesn't matter, since it's weird user-input, but if the domain value is 'htttps://lightpanda.io' (i.e. not a domain, which is why I think we can ignore it), this handles it differently than Chrome. Chrome seems like it'll extract the domain correctly, whereas we'll treat the entire input as a domain as-is.

sjorsdonkers: I'm not able to reproduce this one, for me Chrome sets `http` as the domain, which I think we can consider equally useless. (regardless of how many `t`s)
image

src/cdp/domains/storage.zig
sjorsdonkers: browser.cookies()
> {"method":"Storage.getCookies","params":{},"id":29}
> 0 results

context.cookies()
> {"method":"Storage.getCookies","params":{"browserContextId":"F84F263F9B18E886D63394D299BF5B59"},"id":30}
> 2 results

The problem is that when we call `browser.createBrowserContext();` we set it as the default and only context.
Chrome however already has a default context, so this will create a 2nd one.

It you replace:
`const context = await browser.createBrowserContext();`
With:
`const context = await browser.defaultBrowserContext();`
(and not close that context!)
`Storage.getCookies` returns 2 results as expected.

This is an issue we need to solve, but not in this PR as the implementation of using the default context is correct. It's just that the default context is incorrect.
Issue created: https://github.com/lightpanda-io/browser/issues/783

src/cdp/domains/storage.zig
sjorsdonkers: Old checks removed: [2e05d29](https://github.com/lightpanda-io/browser/pull/777/commits/2e05d29a1bf14890349be818c0061ed1e5c85aaf)

src/cdp/domains/storage.zig
sjorsdonkers: Added: [2e05d29](https://github.com/lightpanda-io/browser/pull/777/commits/2e05d29a1bf14890349be818c0061ed1e5c85aaf)

PR #780, opened 2025-06-12
Fix loop run (Page.wait)

karlseguin:
Fixes the DDG result regression.

In https://github.com/lightpanda-io/browser/pull/767 I tried to call loop.run from within a loop.run (spoiler, it didn't work), in order to make sure aborted connections were properly cleaned up before starting a new navigation. That resulted in having loop.run no longer wait for timeouts for fear of having to wait on a long timeout. The ended up breaking page.wait (used in the fetch command).

This commit brings back the original behavior where loop.run() waits for all completions. Which is now safe to do since the nested loop.run() call has been removed.

### DIFF
--- a/src/runtime/loop.zig
+++ b/src/runtime/loop.zig
@@ -81,12 +81,13 @@ pub const Loop = struct {
 
         // run tail events. We do run the tail events to ensure all the
         // contexts are correcly free.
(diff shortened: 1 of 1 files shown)


krichprollsch:
It works :tada: :clap:

PR #782, opened 2025-06-16
Use css.querySelectorAll to find form elements

karlseguin:
Libdom's formGetCollection doesn't work (like I would expect) for dynamically added elements.

For example, given:

```
let el = document.createElement('input');
document.getElementsByTagName('form')[0].append(el);
```

(and assume the page has a form), I'd expect `el.form` to be equal to the form the input was added to. Instead, it's null. This is a problem given that `dom_html_form_element_get_elements` uses the element's `form` attribute to "collect" the elements.

This uses our existing querySelector to find the form elements.

### DIFF
--- a/src/browser/xhr/form_data.zig
+++ b/src/browser/xhr/form_data.zig
@@ -115,17 +115,16 @@ const EntryIterable = iterator.Iterable(kv.EntryIterator, "FormDataEntryIterator
 // TODO: handle disabled fieldsets
 fn collectForm(form: *parser.Form, submitter_: ?*parser.ElementHTML, page: *Page) !kv.List {
     const arena = page.arena;
-    const collection = try parser.formGetCollection(form);
-    const len = try parser.htmlCollectionGetLength(collection);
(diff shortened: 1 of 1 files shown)


src/browser/xhr/form_data.zig
```diff
@@ -356,6 +353,8 @@ test "Browser.FormData" {
 
     try runner.testCases(&.{
         .{ "let form1 = document.getElementById('form1')", null },
+        .{ "let input = document.createElement('input');", null },
+        .{ "input.name = 'dyn'; input.value= 'dyn-v'; form1.appendChild(input);", null },
```
sjorsdonkers, line 365: Assuming my understanding of the issues is correct.
`input.form` should be equal to `form1`.
If that is not the case for us then I think we should fix the underlying issues.
If this is a "temporary" hack I would prefer a big flashy comment explaining the situation.

karlseguin: That's true, I'll add a comment. But I don't think this is a hack. I think moving the collection of form elements to our own 'walker' is the right approach so that we can eventually address _other_ issues, like supporting disabled fieldsets.

But dynamically added items don't get the `form` property set, and that should be addressed also.

PR #787, opened 2025-06-16
Propagate set returnvalue

sjorsdonkers:
This does not work yet.
I expected this to be enough, but I do not see yet where the input is actually used instead of the return value.


karlseguin:
I think this is accidentally fixed in https://github.com/lightpanda-io/browser/pull/785

sjorsdonkers:
@karlseguin 
> I think this is accidentally fixed in #785

That fix does indeed do the same as in this PR.
Neither fixes the return value issues tho.
I'll close this PR

PR #789, opened 2025-06-17
browser context proxyServer

sjorsdonkers:
Currently we rely on there existing only 1 browser context. We just need to make sure that we set it back after the session ends

Additional:
Should we rename `--http-proxy` to `--proxy-server` to align with chromium?

### DIFF
--- a/src/cdp/domains/target.zig
+++ b/src/cdp/domains/target.zig
@@ -66,10 +66,29 @@ fn getBrowserContexts(cmd: anytype) !void {
 }
 
 fn createBrowserContext(cmd: anytype) !void {
+    const params = try cmd.params(struct {
+        disposeOnDetach: bool = false,
+        proxyServer: ?[]const u8 = null,
+        proxyBypassList: ?[]const u8 = null,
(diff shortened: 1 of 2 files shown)


src/cdp/domains/target.zig
```diff
@@ -66,11 +66,30 @@ fn getBrowserContexts(cmd: anytype) !void {
...
         error.AlreadyExists => return cmd.sendError(-32000, "Cannot have more than one browser context at a time"),
         else => return err,
     };
 
+    if (params) |p| {
+        if (p.proxyServer) |proxy| {
+            // For now the http client is not in the browser context so we assume there is just 1.
+            bc.http_proxy_before = cmd.cdp.browser.http_client.http_proxy;
+            const proxy_cp = try cmd.cdp.browser.http_client.allocator.dupe(u8, proxy);
+            cmd.cdp.browser.http_client.http_proxy = try std.Uri.parse(proxy_cp);
```
sjorsdonkers, line 89: Is this too hacky? 🧐

PR #790, opened 2025-06-18
Minimal CSSStyleSheet

mookums:
This is a super barebones CSSStyleSheet, mostly for just resolving various Custom Elements complaining about a lack of the CSSStyleSheet interface.

A lot of functionality is not implemented but it seems to allow more components to load on the Reddit example.

### DIFF
--- a/src/browser/cssom/css_rule.zig
+++ b/src/browser/cssom/css_rule.zig
@@ -43,38 +43,3 @@ pub const CSSImportRule = struct {
     style_sheet: CSSStyleSheet,
     supports_text: ?[]const u8,
 };
-
-// pub const CSSGroupingRule = struct {
-//     pub const prototype = *CSSRule;
-//     list: std.ArrayListUnmanaged(CSSRule),
-
-//     pub fn _insertRule(self: *CSSGroupingRule, rule: []const u8, _index: ?usize, page: *Page) !usize {
-//         const index = _index orelse 0;
-//         if (index > self.list.items.len) return error.IndexSizeError;
-
-//         const css_rule: CSSRule = .{ .css_text = rule, .parent_rule = null, .parent_stylesheet = null };
-//         try self.list.insert(page.arena, index, css_rule);
-//         return self.list.items.len;
-//     }
-
-//     pub fn _deleteRule(self: *CSSGroupingRule, index: usize) !void {
-//         if (index > self.list.items.len) return error.IndexSizeError;
-//         _ = self.list.orderedRemove(index);
-//     }
-// };
-
-// pub const CSSStyleRule = struct {
-//     pub const prototype = *CSSGroupingRule;
-//     selector_text: []const u8,
-//     style: CSSStyleDeclaration,
-// };
-
-// pub const CSSFontFaceRule = struct {
-//     pub const prototype = *CSSRule;
-// };
-
-// pub const CSSPageRule = struct {
-//     pub const prototype = *CSSGroupingRule;
-//     selector_text: []const u8,
-//     style: CSSStyleDeclaration,
-// };
--- a/src/browser/cssom/css_rule.zig
+++ b/src/browser/cssom/css_rule.zig
@@ -18,9 +18,6 @@
 
 const std = @import("std");
 
-const Page = @import("../page.zig").Page;
-
-const CSSStyleDeclaration = @import("css_style_declaration.zig").CSSStyleDeclaration;
 const CSSStyleSheet = @import("css_stylesheet.zig").CSSStyleSheet;
 
 pub const Interfaces = .{
--- a/src/browser/cssom/css_rule.zig
+++ b/src/browser/cssom/css_rule.zig
@@ -1,0 +1,80 @@
+// Copyright (C) 2023-2024  Lightpanda (Selecy SAS)
+//
+// Francis Bouvier <francis@lightpanda.io>
+// Pierre Tachoire <pierre@lightpanda.io>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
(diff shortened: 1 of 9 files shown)


src/browser/cssom/css_rule.zig
```diff
@@ -0,0 +1,45 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+const std = @import("std");
+
+const Page = @import("../page.zig").Page;
```
karlseguin:
```suggestion
```

src/browser/cssom/css_rule.zig
```diff
@@ -0,0 +1,45 @@
...
+
+const CSSStyleDeclaration = @import("css_style_declaration.zig").CSSStyleDeclaration;
```
karlseguin:
```suggestion
```

src/browser/cssom/css_rule_list.zig
```diff
@@ -0,0 +1,62 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+const std = @import("std");
+
+const Page = @import("../page.zig").Page;
```
karlseguin:
```suggestion
```

src/browser/cssom/css_rule_list.zig
```diff
@@ -0,0 +1,62 @@
...
+const StyleSheet = @import("stylesheet.zig").StyleSheet;
```
karlseguin, line 21:
```suggestion
```

src/browser/cssom/css_rule_list.zig
```diff
@@ -0,0 +1,62 @@
...
+
+const CSSRule = @import("css_rule.zig").CSSRule;
+const CSSImportRule = @import("css_rule.zig").CSSImportRule;
```
karlseguin, line 23:
```suggestion
```

src/browser/cssom/css_stylesheet.zig
```diff
@@ -0,0 +1,92 @@
...
+const std = @import("std");
+
+const Page = @import("../page.zig").Page;
+const StyleSheet = @import("stylesheet.zig").StyleSheet;
+
+const CSSRuleList = @import("css_rule_list.zig").CSSRuleList;
+const CSSImportRule = @import("css_rule.zig").CSSImportRule;
+
+pub const CSSStyleSheet = struct {
+    pub const prototype = *StyleSheet;
```
karlseguin, line 28: There's 2 types of ways we implement prototype inheritance. For libdom objects, we rely on the fact that the libdom types can be casted from each other. A `dom_html_input_element` can be safely `@ptrCast` to a `dom_html_element`. This is because the (a) C types have guaranteed memory layout and (b) the libdom types were designed for it.

For Zig types, this isn't safe to rely on. If you do:

```js
let css = new CSSStyleSheet()
css.href;
```

The behavior is undefined.

The code in question is:
https://github.com/lightpanda-io/browser/blob/8033e41d4a0cd19542af09d5886788209d5b0d02/src/runtime/js.zig#L2266

Here we see that if the `offset` value is -1, we just cast directly. That `offset` field is set here: https://github.com/lightpanda-io/browser/blob/8033e41d4a0cd19542af09d5886788209d5b0d02/src/runtime/js.zig#L757

You can see it's set if the type being returned (i.e. `CSSStyleSheet`) has a `proto` field. The `offset` is the `@offsetOf` of that `proto` field.

So, the way it is now, it's like doing:

```zig
var decl = CSSStyleSheet{...};
var sheet: *StyleSheet = @alignCast(@ptrCast(&decl));
```

tl;dr - there should be a `proto` field here which is of type`StyleSheet`.  Probably reasonable to give all/most of the `StyleSheet` fields default values.

src/browser/cssom/css_stylesheet.zig
```diff
@@ -0,0 +1,92 @@
...
+
+    css_rules: *CSSRuleList,
```
karlseguin: Why is this a pointer?

src/browser/cssom/css_stylesheet.zig
```diff
@@ -0,0 +1,92 @@
...
+        _ = opts;
+
+        const arena = page.arena;
+        const rules = try arena.create(CSSRuleList);
+        rules.* = .constructor();
+
+        return .{ .css_rules = rules, .owner_rule = null };
+    }
+
+    pub fn get_ownerRule(_: *CSSStyleSheet) ?*CSSImportRule {
```
karlseguin, line 49:
```suggestion
    pub fn get_ownerRule(_: *CSSStyleSheet) ?*CSSImportRule {
```

Not a big deal. Can also be done in `get_cssRules`

src/browser/cssom/stylesheet.zig
```diff
@@ -0,0 +1,55 @@
...
+// https://developer.mozilla.org/en-US/docs/Web/API/StyleSheet#specifications
+pub const StyleSheet = struct {
+    disabled: bool,
+    href: []const u8,
+    owner_node: *parser.Node,
+    parent_stylesheet: ?*StyleSheet,
+    title: []const u8,
+    type: []const u8,
+
+    pub fn get_disabled(self: *StyleSheet) bool {
```
karlseguin:
```suggestion
    pub fn get_disabled(self: *const StyleSheet) bool {
```
(same as for the other ones)

src/browser/cssom/css_stylesheet.zig
mookums: No longer a pointer

src/browser/cssom/css_stylesheet.zig
mookums: Thanks for the explaination, changed this to be a `proto` field.

karlseguin: It's redundant, but you still nee the prototype declaration. I fixed it in:

https://github.com/lightpanda-io/browser/pull/790/commits/54bd55d45d66d44f74f98c55cb8ab88e9d5473d0

Could change js.zig so that it can use either I guess.

PR #792, opened 2025-06-18
Fixes the scoping of page.current_script

karlseguin:
This was previously being set back to null before it was actually needed.

Also, added a more logs / log details.


src/browser/page.zig
```diff
@@ -461,9 +474,9 @@ pub const Page = struct {
         };
 
         var script_source: ?[]const u8 = null;
+        defer self.current_script = null;
```
karlseguin, line 477: moving this outside the `if`  block, so that `self.current_script` is set when `script.eval` is called is the only meaningful change in this PR.

krichprollsch (approved):

PR #795, opened 2025-06-19
Add dummy PerformanceObserver

karlseguin:
Adds a dummy PerformanceObserver. Only the supportedEntryTypes static attribute is supported, and it currently returns an empty array. This hopefully prevents code from trying to use it. For example, before using it, reddit checks if specific types are supported and, if not, doesn't use it.

This introduced complexity in the js runtime. Our current approach to attributes only works with primitive types. Non-primitive types can't be attached to a FunctionTemplate (v8 will crash saying only primitive types can be set). Plus, all non primitive types require a context to create anyways.

We now detect "primitive" attributes and "complex" attributes. Primitive attributes are setup as before. Complex attributes are setup per-context, requiring another loop through our types to detect & setup on each context creation.


src/browser/dom/performance_observer.zig
```diff
@@ -0,0 +1,34 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+const std = @import("std");
+
+// https://developer.mozilla.org/en-US/docs/Web/API/PerformanceObserver
+pub const PerformanceObserver = struct {
+    pub const _supportedEntryTypes = [0][]const u8{};
```
sjorsdonkers, line 23: Would the need to register the Array as a type be resolved if we handle static properties the same as instance properties, like `get_x()`?
It would just return statis data instead of instance data.
I think we can do `static_supportedEntryTypes()` can we not?

karlseguin: Arrays are already supported. This code goes through the existing `zigValueToJs`. The issue is that they're only supported when we have a context. You're effectively doing `new Array();` - you need a handlescope + context to be able to do that.

We have `simpleZigValueToJs` (which could be renamed to "primitive" now that I know that's what V8 calls them) which specifically exists for cases where we don't have a context, but need a JS value.

`static_` creates a function - it needs to be invoked with `()`in JS, else you get a function reference, not the value. We could do `static_get_`, but I think I might still implement it the same way (directly against the JSObject) because I think it'll perform better than invoking a function callback.

krichprollsch (approved):

PR #798, opened 2025-06-20
Fix module loading

karlseguin:
When V8 calls the ResolveModuleCallback that we give it, it passes the specifier which is essentially the string given to `from`:

```
import {x} from './blah.js';
```

We were taking that specifier and giving it to the page. The page knew the currently executing script, an thus could resolve the full URL. Given the full URL, it could either return the JS content from its module cache or fetch the source.

At best though, this isn't efficient. If two files import the same module, yes we cache the src, but we still ask v8 to re-compile it. At worse, it crashes due to resource exhaustion in the case of cyclical dependencies.

ResolveModuleCallback should instead detect that it has already loaded the module and return the previously loaded module. Essentially, we shouldn't be caching the JavaScript source, we should be caching the v8 module.

However, in order to do this, we need more than the specifier, which might only be a relative path (and thus isn't unique). So, in addition to a module cache, we now also maintain an module identifier lookup. Given a module, we can get its full path. Thankfully ResolveModuleCallback gives us the referring module, so we can look up that modules URL, stitch it to the specifier, and get the full url (the unique identifier) within the JS runtime.

Need more real world testing, and a fully working example before I celebrate, but for sites with many import, this appears to improve performance by many orders of magnitude.

### DIFF
--- a/src/browser/page.zig
+++ b/src/browser/page.zig
@@ -87,13 +87,6 @@ pub const Page = struct {
     // execute any JavaScript
     main_context: *Env.JsContext,
 
-    // List of modules currently fetched/loaded.
-    module_map: std.StringHashMapUnmanaged([]const u8),
-
-    // current_script is the script currently evaluated by the page.
-    // current_script could by fetch module to resolve module's url to fetch.
-    current_script: ?*const Script = null,
-
     // indicates intention to navigate to another page on the next loop execution.
     delayed_navigation: bool = false,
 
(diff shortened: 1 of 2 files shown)


karlseguin:
Similar to chrome:
https://github.com/chromium/chromium/blob/main/services/accessibility/features/v8_manager.cc#L108

krichprollsch (approved):

PR #802, opened 2025-06-23
Don't allow timeouts to be registered when shutting down

karlseguin:
Currently, a timeout that sets a timeout can cause loop.run to block forever.

Also, cleanup URL stitch with leading './'. The resulting URL was valid, but since we use the URL as the module cache key, it's important to resolve various representations of the same URL in the same way.

### DIFF
--- a/src/runtime/loop.zig
+++ b/src/runtime/loop.zig
@@ -187,6 +187,12 @@ pub const Loop = struct {
     }
 
     pub fn timeout(self: *Self, nanoseconds: u63, callback_node: ?*CallbackNode) !usize {
+        if (self.stopping) {
+            // Prevents a timeout callback from creating a new timeout, which
+            // would make `loop.run` run forever.
+            return 0;
+        }
+
         const completion = try self.alloc.create(Completion);
         errdefer self.alloc.destroy(completion);
         completion.* = undefined;
--- a/src/url.zig
+++ b/src/url.zig
@@ -110,7 +110,13 @@ pub const URL = struct {
             }
             return src;
         }
-        if (src.len == 0) {
+
+        var normalized_src = if (std.mem.startsWith(u8, src, "/")) src[1..] else src;
+        while (std.mem.startsWith(u8, normalized_src, "./")) {
+            normalized_src = normalized_src[2..];
+        }
+
+        if (normalized_src.len == 0) {
             if (opts.alloc == .always) {
                 return allocator.dupe(u8, base);
             }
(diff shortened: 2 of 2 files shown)


karlseguin:
Initial version of this didn't work well.

The goal is to prevent this type of code:

```js
function logEvents() {
   send(events);
   setTimeout(logEvents, 3000);
```

From making loop.run() run forever.

However, simply preventing timeouts from being registered during "shutdown" can break pages. For example, our fetch.js does this:

```js
onload: function() {
  // ...
  setTimeout(function() {
    resolve(new Response(body, options));
  }, 0);
}
```

Which means that something like:

```js
fetch(request).then((res) => {.....});
```

may not work if `onload` is called during `loop.run` (because we'll ignore the timeout to resolve the promise).


In other words, some timeouts we want to ignore, and some we don't. But I don't think we can tell which is why so, for now, I just added a timeout.

src/runtime/loop.zig
```diff
@@ -102,12 +102,16 @@ pub const Loop = struct {
...
     // Note that I/O events callbacks might register more I/O events
     // on the go when they are executed (ie. nested I/O events).
-    pub fn run(self: *Self) !void {
+    pub fn run(self: *Self, wait_time: usize) !void {
         // stop repeating / interval timeouts from re-registering
         self.stopping = true;
         defer self.stopping = false;
 
-        while (self.pending_network_count != 0 or self.pending_timeout_count != 0) {
+        const max_iterations = wait_time / (std.time.ns_per_ms * 10);
```
sjorsdonkers: Nits: Consider: `wait_nanosec` and giving `10` a name like `run_duration`.

Side note `run_for_ns` is likely quite inaccurate, no idea how much it would accumulate and overshoot `wait_time`, hmm maybe it's fine on unix, and we probably won't care.

src/main.zig
```diff
@@ -126,7 +126,7 @@ fn run(alloc: Allocator) !void {
                 },
             };
 
-            try page.wait();
+            try page.wait(std.time.ns_per_s * 3);
```
krichprollsch, line 129: I guess we will like to set a user-facing option at some point.

krichprollsch (approved):

src/main.zig
karlseguin: If we take a param, it should get used/passed to:
https://github.com/lightpanda-io/browser/blob/cb35b3624aa192a1c65d7726a25e9dbbdece064f/src/browser/html/window.zig#L220

I was thinking something a bit fancier. Accepting an arbitrary JS script, e.g. `--eval "document.getElementbyId('comments')"` which we run on every loop and, when it returns a truthy JS value, we exit.

PR #805, opened 2025-06-23
add PerformanceEntry and PerformanceMark

mookums:
This adds the `PerformanceEntry` and `PerformanceMark` interfaces. This also adds the ability to return enums as long as they have a `toString()` in `js.zig`.

### DIFF
--- a/src/browser/dom/performance.zig
+++ b/src/browser/dom/performance.zig
@@ -20,6 +20,19 @@ const std = @import("std");
 
 const parser = @import("../netsurf.zig");
 const EventTarget = @import("../dom/event_target.zig").EventTarget;
+const Env = @import("../env.zig").Env;
+const Page = @import("../page.zig").Page;
+
+pub const Interfaces = .{
+    Performance,
+    PerformanceEntry,
+    PerformanceMark,
+};
+
+const MarkOptions = struct {
+    detail: ?Env.JsObject = null,
+    start_time: ?f64 = null,
+};
 
 // https://developer.mozilla.org/en-US/docs/Web/API/Performance
 pub const Performance = struct {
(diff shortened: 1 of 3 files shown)


src/browser/dom/performance.zig
```diff
@@ -52,6 +65,91 @@ pub const Performance = struct {
...
+        const options = _options orelse MarkOptions{};
+        const start_time = options.start_time orelse perf._now();
+        if (start_time < 0.0) {
+            return error.TypeError;
+        }
+
+        const duped_name = try page.arena.dupe(u8, name);
+        const proto = PerformanceEntry{ .name = duped_name, .entry_type = .mark, .start_time = start_time };
+
+        return .{ .proto = proto, .detail = options.detail };
```
karlseguin: By default, JsObject is non-persistent. It can become invalid as v8 sees fit. That's fine for functions that take a JsObject and immediately make use of it, but if it's being stored in a Zig instance, we essentially need to increase its reference count so v8 doesn't free it.

```zig
.detail = if (options.detail) |d| try d.persist() else null
```


(For completeness, by default, Env.Function _is_ persistent. The reason JsObject *isn't*, but Function *is* is just a matter of how they're typically used (or at least, how it looked like we were mostly using them at the time.))

src/runtime/js.zig
```diff
@@ -2317,6 +2317,13 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
                     return v8.initNull(isolate).toValue();
                 },
                 .error_union => return zigValueToJs(templates, isolate, v8_context, value catch |err| return err),
+                .@"enum" => {
```
karlseguin: I think it makes more sense to put this in `simpleZigValueToJs`. 

There are cases where we call `simpleZigValueToJs` directly, so having it there means that enums can be used in those cases. You will have to add `.enum` above, in the unreachable case, of `zigValueToJs` to get it to compile.

(`simpleZigValueToJs` exists for types that can be converted without a context, there are times where we like to be able to create a JS value without a context, like when setting up a class attribute...so by moving this logic directly in `simpleZigValueToJs`, we'll be able to use enums in those cases too)

PR #806, opened 2025-06-23
add AbstractRange and Range

mookums:
This adds a pretty basic implementation of the `AbstractRange` and `Range`.

### DIFF
--- a/src/browser/dom/range.zig
+++ b/src/browser/dom/range.zig
@@ -1,0 +1,120 @@
+// Copyright (C) 2023-2024  Lightpanda (Selecy SAS)
+//
+// Francis Bouvier <francis@lightpanda.io>
+// Pierre Tachoire <pierre@lightpanda.io>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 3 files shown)


src/browser/dom/range.zig
```diff
@@ -0,0 +1,122 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+const std = @import("std");
+
+const log = @import("../../log.zig");
```
karlseguin:
```suggestion
```

src/browser/dom/range.zig
```diff
@@ -0,0 +1,122 @@
...
+};
+
+const testing = @import("../../testing.zig");
+test "Browser.Range" {
+    var runner = try testing.jsRunner(testing.tracking_allocator, .{});
+    defer runner.deinit();
+
+    try runner.testCases(&.{
+        // Test Range constructor
+        .{ "let range = new Range()", "undefined" },
```
karlseguin, line 106: This is fine, but just FYI, if you don't care about the output, you can pass `null` as the 2nd parameter:

```zig
.{ "let range = new Range()", null },
```

But maybe in this case you _do_ care to check that its `undefined`

PR #807, opened 2025-06-23
add CSS utility interface

mookums:
This adds in a super minimal `CSS` utility interface. It also modifies the `js.zig` and allows for utility interfaces (Interfaces that contain static elements but can't be constructed) to work.

### DIFF
--- a/src/browser/css/css.zig
+++ b/src/browser/css/css.zig
@@ -23,6 +23,18 @@ const std = @import("std");
 const Selector = @import("selector.zig").Selector;
 const parser = @import("parser.zig");
 
+pub const Interfaces = .{
+    Css,
+};
+
+// https://developer.mozilla.org/en-US/docs/Web/API/CSS
+pub const Css = struct {
+    pub fn _supports(_: *Css, _: []const u8, _: ?[]const u8) bool {
+        // TODO: Actually respond with which CSS features we support.
+        return true;
+    }
+};
+
 // parse parse a selector string and returns the parsed result or an error.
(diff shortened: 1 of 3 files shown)


src/browser/css/css.zig
```diff
@@ -23,6 +23,18 @@ const std = @import("std");
 const Selector = @import("selector.zig").Selector;
 const parser = @import("parser.zig");
 
+pub const Interfaces = .{
```
karlseguin, line 26: This has to be included in the Env.Interface (browser/env.zig) ??

src/browser/css/css.zig
```diff
@@ -23,6 +23,18 @@ const std = @import("std");
...
+    CSS,
+};
+
+// https://developer.mozilla.org/en-US/docs/Web/API/CSS
+pub const CSS = struct {
+    pub fn _supports(_: []const u8, _: ?[]const u8) bool {
```
karlseguin: I know it's doing nothing, but since this is currently the only code that's using the new `checkSelfReceiver` and the subsequent flow, I think a test, as trivial as it might seem, is needed:


```zig
test "Browser.HTML.CSS" {
    var runner = try testing.jsRunner(testing.tracking_allocator, .{});
    defer runner.deinit();

    try runner.testCases(&.{
        .{ "CSS.supports('over 9000')", "true" },
    }, .{});
}
```

mookums:
I don't really see the point anymore in keeping that `checkSelfReceiver` just because we end up needing to do it this way anyways as `window` gets treated as our global. Just stripped it out and I treat the `Css` as the actual interface and the `CSS` as instance with the methods attached.

PR #809, opened 2025-06-24
Add element.dataset API

karlseguin:
Uses the State to store the dataset, but, on first load, loads the data attributes from the DOM.

### DIFF
--- a/src/browser/html/DataSet.zig
+++ b/src/browser/html/DataSet.zig
@@ -16,64 +16,66 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 const std = @import("std");
+const parser = @import("../netsurf.zig");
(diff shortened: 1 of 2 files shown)


krichprollsch:
LGTM, but I'm wondering about `attributes` synchronization:
* should be `data-x` attributes accessible from `attributes` list?
* what about generating the html dump? if we don't sync with attr, we could loop hover the dataset to build the valid final html dump.

It's similar to our `input` problem I suppose. That's something we can work on in a future PR.

karlseguin:
> LGTM, but I'm wondering about `attributes` synchronization:

You're right. I did a follow up commit. DataSet now just wraps the *parser.Element and uses `elementGetAttribute`, `elementSetAttribute`, `elementRemoveAttribute`. 

I tested in browser, and this is more accurate to how it should work - changes to the dataset are reflected in the attributes and vice-versa.

PR #812, opened 2025-06-25
We cannot have empty Zig structs mapping to JS instances

karlseguin:
An empty struct will share the same address as its sibling (1) which will cause an collision in the identity map.

(1) - This depends on Zig's non-guaranteed layout, so the collision might not be with its sibling, but rather some other [seemingly random] field.

### DIFF
--- a/src/browser/crypto/crypto.zig
+++ b/src/browser/crypto/crypto.zig
@@ -21,6 +21,8 @@ const uuidv4 = @import("../../id.zig").uuidv4;
 
 // https://w3c.github.io/webcrypto/#crypto-interface
 pub const Crypto = struct {
+    _not_empty: bool = true,
+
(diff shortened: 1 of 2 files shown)


sjorsdonkers (approved):
Lgtm
Could add an assert checking for zero sized fields

karlseguin:
In `_mapZigInstanceToJs` I tried to add this:

```zig
if (isEmpty(ptr.child) == false) {
    // existing code
} else if (@TypeOf(js_obj_or_template) != v8.Object and @typeInfo(Receiver(ptr.child)) != .@"opaque" and @sizeOf(ptr.child) == 0) {
    // This else if was added
    @compileLog(@typeName(ptr.child));
} else {
    // existing comment
}
```

But I get a few false positives.I guess we could create a whitelist of types, but that gets pretty messy.

PR #815, opened 2025-06-25
TLS for CONNECT proxy 

sjorsdonkers:
Depends on https://github.com/ianic/tls.zig/pull/14

### DIFF
--- a/src/http/client.zig
+++ b/src/http/client.zig
@@ -239,11 +239,6 @@ pub const Client = struct {
         const proxy_type = self.proxy_type orelse return false;
         return proxy_type == .forward;
     }
-
-    fn isProxyTLS(self: *const Client) bool {
-        const proxy = self.http_proxy orelse return false;
-        return std.ascii.eqlIgnoreCase(proxy.scheme, "https");
-    }
 };
 
 const RequestOpts = struct {
@@ -388,9 +383,6 @@ pub const Request = struct {
     // List of request headers
     headers: std.ArrayListUnmanaged(std.http.Header),
 
-    // whether or not we expect this connection to be secure
-    _secure: bool,
-
     // whether or not we should keep the underlying socket open and and usable
     // for other requests
     _keepalive: bool,
@@ -398,6 +390,10 @@ pub const Request = struct {
     // extracted from request_uri
     _request_port: u16,
     _request_host: []const u8,
+    // Whether or not we expect this connection to be secure, connection may still be secure due to proxy
+    _request_secure: bool,
+    // Whether or not we expect the SIMPLE/CONNECT proxy connection to be secure
+    _proxy_secure: bool,
 
     // extracted from connect_uri
     _connect_port: u16,
@@ -483,11 +479,12 @@ pub const Request = struct {
             .method = method,
             .notification = null,
             .arena = state.arena.allocator(),
-            ._secure = decomposed.secure,
             ._connect_host = decomposed.connect_host,
             ._connect_port = decomposed.connect_port,
+            ._proxy_secure = decomposed.proxy_secure,
             ._request_host = decomposed.request_host,
             ._request_port = decomposed.request_port,
+            ._request_secure = decomposed.request_secure,
             ._state = state,
             ._client = client,
             ._aborter = null,
@@ -519,12 +516,13 @@ pub const Request = struct {
     }
 
     const DecomposedURL = struct {
-        secure: bool,
         connect_port: u16,
         connect_host: []const u8,
         connect_uri: *const std.Uri,
+        proxy_secure: bool,
         request_port: u16,
         request_host: []const u8,
+        request_secure: bool,
     };
     fn decomposeURL(client: *const Client, uri: *const Uri) !DecomposedURL {
         if (uri.host == null) {
@@ -539,32 +537,31 @@ pub const Request = struct {
             connect_host = proxy.host.?.percent_encoded;
         }
 
-        const is_connect_proxy = client.isConnectProxy();
-        const is_proxy_tls = client.isProxyTLS();
-
-        var secure: bool = undefined;
-        const scheme = if (is_connect_proxy) uri.scheme else connect_uri.scheme;
-        if (std.ascii.eqlIgnoreCase(scheme, "https")) {
-            secure = true;
-        } else if (std.ascii.eqlIgnoreCase(scheme, "http")) {
-            secure = false;
+        var request_secure: bool = undefined;
+        if (std.ascii.eqlIgnoreCase(uri.scheme, "https")) {
+            request_secure = true;
+        } else if (std.ascii.eqlIgnoreCase(uri.scheme, "http")) {
+            request_secure = false;
         } else {
             return error.UnsupportedUriScheme;
         }
-        const request_port: u16 = uri.port orelse if (secure) 443 else 80;
+        const proxy_secure = client.http_proxy != null and std.ascii.eqlIgnoreCase(client.http_proxy.?.scheme, "https");
+
+        const request_port: u16 = uri.port orelse if (request_secure) 443 else 80;
         const connect_port: u16 = connect_uri.port orelse blk: {
-            if (is_connect_proxy) {
-                if (is_proxy_tls) break :blk 443 else break :blk 80;
+            if (client.isConnectProxy()) {
+                if (proxy_secure) break :blk 443 else break :blk 80;
             } else break :blk request_port;
         };
 
         return .{
-            .secure = secure,
             .connect_port = connect_port,
             .connect_host = connect_host,
             .connect_uri = connect_uri,
+            .proxy_secure = proxy_secure,
             .request_port = request_port,
             .request_host = request_host,
+            .request_secure = request_secure,
         };
     }
 
@@ -682,12 +679,11 @@ pub const Request = struct {
 
             // proxy
             const is_connect_proxy = self._client.isConnectProxy();
-            const is_proxy_tls = self._client.isProxyTLS();
 
             if (is_connect_proxy) {
                 var proxy_conn: SyncHandler.Conn = .{ .plain = self._connection.?.socket };
 
-                if (is_proxy_tls) {
+                if (self._proxy_secure) {
                     // Create an underlying TLS stream with the proxy
                     var proxy_tls_config = tls_config;
                     proxy_tls_config.host = self._connect_host;
@@ -698,8 +694,8 @@ pub const Request = struct {
                 // Connect to the proxy
                 try SyncHandler.connect(self, &proxy_conn);
 
-                if (is_proxy_tls) {
-                    if (self._secure) {
+                if (self._proxy_secure) {
+                    if (self._request_secure) {
                         // If secure endpoint, create the main TLS stream encapsulated into the TLS stream proxy
                         self._connection.?.tls = .{
                             .blocking_tls_in_tls = .{
@@ -715,7 +711,7 @@ pub const Request = struct {
                     }
                 }
             }
-            if (self._secure and !is_proxy_tls) {
+            if (self._request_secure and !self._proxy_secure) {
                 self._connection.?.tls = .{
                     .blocking = try tls.client(std.net.Stream{ .handle = socket }, tls_config),
                 };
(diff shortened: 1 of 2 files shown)


src/http/client.zig
```diff
@@ -322,11 +327,19 @@ const Connection = struct {
 
     const TLSClient = union(enum) {
         blocking: tls.Connection(std.net.Stream),
+        blocking_tls_in_tls: struct {
```
krichprollsch: the name sounds a bit weird, what about something like `blocking_proxy`? or `blocking_tunnel`?

src/http/client.zig
```diff
@@ -239,6 +239,11 @@ pub const Client = struct {
         const proxy_type = self.proxy_type orelse return false;
         return proxy_type == .forward;
     }
+
+    fn isProxyTLS(self: *const Client) bool {
+        const proxy = self.http_proxy orelse return false;
+        return std.mem.eql(u8, proxy.scheme, "https");
```
krichprollsch:
```suggestion
        return std.ascii.eqlIgnoreCase(proxy.scheme, "https");
```

src/http/client.zig
```diff
@@ -655,19 +673,55 @@ pub const Request = struct {
...
+                                .proxy = proxy_conn.tls.*,
+                                .destination = tls_in_tls,
+                            },
+                        };
+                    } else {
+
+                        // otherwise, just use the TLS stream proxy
+                        self._connection.?.tls = .{ .blocking = proxy_conn.tls.* };
+                    }
+                }
```
krichprollsch, line 712: What about refactoring this way from L687 to L726 ?
Not sure if it's correct, but it seems to be simpler to first check `is_proxy_tls`.

```zig
 if (is_connect_proxy) {
   if (!is_proxy_tls) {
     var proxy_conn: SyncHandler.Conn = .{ .plain = self._connection.?.socket };
     try SyncHandler.connect(self, &proxy_conn);
            if (self._secure and !is_proxy_tls) {
                self._connection.?.tls = .{
                    .blocking = try tls.client(std.net.Stream{ .handle = socket }, tls_config),
                };
            }
   } else {
                     // create an underlying TLS stream with the proxy
                    var proxy_tls_config = tls_config;
                    proxy_tls_config.host = self._connect_host;
                    var proxy_conn_tls = try tls.client(std.net.Stream{ .handle = socket }, proxy_tls_config);
                    try SyncHandler.connect(self, &proxy_conn);
                    if (self._secure) {

                        // if secure endpoint, create the main TLS stream
                        // encapsulated into the TLS stream proxy
                        const tls_in_tls = try tls.client(proxy_conn.tls, tls_config);
                        self._connection.?.tls = .{
                            .blocking_tls_in_tls = .{
                                .proxy = proxy_conn.tls.*,
                                .destination = tls_in_tls,
                            },
                        };
                    } else {

                        // otherwise, just use the TLS stream proxy
                        self._connection.?.tls = .{ .blocking = proxy_conn.tls.* };
                    }
   }
 }

File: src/http/client.zig [hunk 4]
```diff
@@ -538,7 +552,11 @@ pub const Request = struct {
             return error.UnsupportedUriScheme;
         }
         const request_port: u16 = uri.port orelse if (secure) 443 else 80;
-        const connect_port: u16 = connect_uri.port orelse (if (is_connect_proxy) 80 else request_port);
+        const connect_port: u16 = connect_uri.port orelse blk: {
+            if (is_connect_proxy) {
```
karlseguin: Could call `client.isProxyTLS()` instead of `is_connect_proxy`, so that you only call `client.isProxyTLS()` if it's needed.

src/http/client.zig
```diff
@@ -655,19 +673,55 @@ pub const Request = struct {
             };
             self._connection = connection;
 
+            const tls_config = tls.config.Client{
```
karlseguin, line 673: Think this can be scoped to the `if (is_connect_proxy) {` case.

src/http/client.zig
```diff
@@ -1880,15 +1941,22 @@ const SyncHandler = struct {
...
     const Conn = union(enum) {
+        tls_in_tls: *tls.Connection(*tls.Connection(std.net.Stream)),
         tls: *tls.Connection(std.net.Stream),
         plain: posix.socket_t,
 
         fn sendRequest(self: *Conn, header: []const u8, body: ?[]const u8) !void {
             switch (self.*) {
-                .tls => |tls_client| {
-                    try tls_client.writeAll(header);
+                .tls => |_| {
```
karlseguin: If you aren't capturing, I think you can merge the two cases

```zig
.tls, .tls_in_tls => {...}
```

And, you can do one better and inline the cases:


```
inline .tls, .tls => |tls_client| {
    try tls_client.writeAll(header);
    if (body) |b| {
        try tls_client.writeAll(b);
    }
},
```

This has the added benefit that  in the two TLS cases,`sendRequest` isn't switch on `self` just to pass it to `writeAll` which was switching again.

src/http/client.zig
```diff
@@ -2081,6 +2156,7 @@ const Reader = struct {
         if (result.done == false) {
             // CONNECT responses should not have a body. If the header is
             // done, then the entire response should be done.
+            log.err(.http_client, "InvalidConnectResponse", .{ .unprocessed = result.unprocessed.? });
```
karlseguin: I don't think `unprocessed` is guaranteed to be non-null here. Leave it as `result.unprocessed` and the logger will unwrap the null. `self.response.status` might be interesting too.

Also, assuming our implementation is right, this is out of our control. I don't think we should log `error` level for things we can't control. `debug` or `info` would be better.

src/http/client.zig
sjorsdonkers: There are 6 cases to support
| Proxy | Destination |
|--------|--------------|
| - | plain |
| - | tls |
| plain| plain |
| plain | tls |
| tls | plain |
| tls | tls |

Fixed suggested change:
```zig
if (is_connect_proxy) {
    if (!is_proxy_tls) {
        var proxy_conn: SyncHandler.Conn = .{ .plain = self._connection.?.socket };
        try SyncHandler.connect(self, &proxy_conn);
        if (self._secure) {
            self._connection.?.tls = .{
                .blocking = try tls.client(std.net.Stream{ .handle = socket }, tls_config),
            };
        } // else neither proxy nor the destination is TLS so the tls_client is null
    } else { // Create an underlying TLS stream with the proxy
        var proxy_tls_config = tls_config;
        proxy_tls_config.host = self._connect_host;
        var proxy_conn_tls = try tls.client(std.net.Stream{ .handle = socket }, proxy_tls_config);
        var proxy_conn: SyncHandler.Conn = .{ .tls = &proxy_conn_tls };

        try SyncHandler.connect(self, &proxy_conn);
        if (self._secure) {
            // If secure endpoint, create the main TLS stream encapsulated into the TLS stream proxy
            self._connection.?.tls = .{
                .blocking_tls_in_tls = .{
                    .proxy = proxy_conn_tls,
                    .destination = undefined,
                },
            };
            const proxy = &self._connection.?.tls.?.blocking_tls_in_tls.proxy;
            self._connection.?.tls.?.blocking_tls_in_tls.destination = try tls.client(proxy, tls_config);
        } else { // Otherwise, just use the TLS stream proxy
            self._connection.?.tls = .{ .blocking = proxy_conn_tls };
        }
    }
} else if (self._secure) {
    self._connection.?.tls = .{
        .blocking = try tls.client(std.net.Stream{ .handle = socket }, tls_config),
    };
}
```

Current:
```zig
if (is_connect_proxy) {
    var proxy_conn: SyncHandler.Conn = .{ .plain = self._connection.?.socket };

    if (is_proxy_tls) {
        // Create an underlying TLS stream with the proxy
        var proxy_tls_config = tls_config;
        proxy_tls_config.host = self._connect_host;
        var proxy_conn_tls = try tls.client(std.net.Stream{ .handle = socket }, proxy_tls_config);
        proxy_conn = .{ .tls = &proxy_conn_tls };
    }

    // Connect to the proxy
    try SyncHandler.connect(self, &proxy_conn);

    if (is_proxy_tls) {
        if (self._secure) {
            // If secure endpoint, create the main TLS stream encapsulated into the TLS stream proxy
            self._connection.?.tls = .{
                .blocking_tls_in_tls = .{
                    .proxy = proxy_conn.tls.*,
                    .destination = undefined,
                },
            };
            const proxy = &self._connection.?.tls.?.blocking_tls_in_tls.proxy;
            self._connection.?.tls.?.blocking_tls_in_tls.destination = try tls.client(proxy, tls_config);
        } else {

            // Otherwise, just use the TLS stream proxy
            self._connection.?.tls = .{ .blocking = proxy_conn.tls.* };
        }
    }
}
if (self._secure and !is_proxy_tls) {
    self._connection.?.tls = .{
        .blocking = try tls.client(std.net.Stream{ .handle = socket }, tls_config),
    };
}
```

Current feels a bit more calm, though the last if is iffy. Can do either.

src/http/client.zig
sjorsdonkers: `blocking_tlsproxy` could work. `blocking_proxy` is a bit odd as `blocking` can be a proxied too. `tls-in-tls` is a term used in the wild sometimes. Let me know

src/http/client.zig
sjorsdonkers: This is the destination tls config, meaning it does not relate to whether there is connect proxy.
It could be an optional constructed if: `self._secure == true`

src/http/client.zig
sjorsdonkers: Can you elaborate, I do not understand.
Currently the `connect_port` is set to `request_port` if it is not in he url/not connect. Otherwise it uses `is_proxy_tls` to use the default ports.

Are you suggesting:
`const connect_port: u16 = connect_uri.port orelse if (is_proxy_tls) 443 else 80;`
Since the value does not matter if the proxy is not CONNECT?

src/http/client.zig
sjorsdonkers: Done [087e269](https://github.com/lightpanda-io/browser/pull/815/commits/087e26959375fe4dcd77936f5fa00e5f2da9e35c)

src/http/client.zig
sjorsdonkers: Done: [2e89683](https://github.com/lightpanda-io/browser/pull/815/commits/2e89683e8db0180390ce86072f54dfc5957e6bc5)

src/http/client.zig
sjorsdonkers: fixed: [2e89683](https://github.com/lightpanda-io/browser/pull/815/commits/2e89683e8db0180390ce86072f54dfc5957e6bc5)

src/http/client.zig
karlseguin:
```zig
if (client.isProxyTLS()) {
    if (is_proxy_tls) break :blk 443 else break :blk 80;
} else break :blk request_port;
```

the `is_connect_proxy` isn't used anywhere else, and it's not needed unless `is_connect_proxy == true`

src/http/client.zig
karlseguin: You're right, I missed the last usage.

src/http/client.zig
```diff
@@ -655,19 +673,51 @@ pub const Request = struct {
...
+                    var proxy_conn_tls = try tls.client(std.net.Stream{ .handle = socket }, proxy_tls_config);
+                    proxy_conn = .{ .tls = &proxy_conn_tls };
+                }
 
-            if (self._secure) {
+                // Connect to the proxy
+                try SyncHandler.connect(self, &proxy_conn);
+
+                if (is_proxy_tls) {
+                    if (self._secure) {
```
sjorsdonkers: This is not the current meaning of `_secure`. Here it is expected to mean the destination is tls, but _secure currently is based on `if (is_connect_proxy) uri.scheme else connect_uri.scheme`

It is used in the connection manager
`if (std.ascii.eqlIgnoreCase(connection.host, host) and connection.port == port and connection.blocking == blocking and ((connection.tls == null) == !secure)) {`

sjorsdonkers: Addressed [4ed39f](https://github.com/lightpanda-io/browser/pull/815/commits/44ed39f366f7204bf61a0180292926d96bfd6a65)
May need to revisit the connection manager later

src/http/client.zig
krichprollsch: ok then :+1:

src/http/client.zig
krichprollsch: I prefer `blocking_tlsproxy`

sjorsdonkers: [a7739cc](https://github.com/lightpanda-io/browser/pull/815/commits/a7739ccdb70d72cb42ff9dd0e899f8b445ddbc5e)

PR #821, opened 2025-06-27
Abort controller

karlseguin:
Missing the [AbortSignal.any()](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/any_static) API.

This API is superficially simple, but required a number of deeper changes.

1 - The poor support for namespace/static functions introduced in https://github.com/lightpanda-io/browser/pull/749 has been improved. Static functions, like `AbortSignal.timeout`, are now correctly supported. 

I still think the `console` behavior is weird, but it's now implemented as a receiver-less method (which is now supported for all methods). Possibly still not correct, but at least it's simpler and and better aligned with how we structure our code.

2 - Like the `Window` before, this introduces an EventTarget which is not a libdom node, an AbortSignal. Unlike `Window`, this isn't a singleton and can't be easily detected. I corrupted the internal event type to serve as a discriminator to the event target. If this doesn't work for future cases, we'll have to come up with a more general solution, like attaching some internal_type_id to the event_target.

3 - Ran into a variation of the empty-struct identity map issue (see: https://github.com/lightpanda-io/browser/pull/812). This time I tried to solve this properly, but it turns out that making the identity map type-aware isn't without its challenges and drawback. Instead, I solved this specific by forcing unique addresses between conflicted objects. As punishment for my failure, I documented the issue: https://github.com/lightpanda-io/project/discussions/165

4 - I believe this is the first case where we've had to raise an arbitrary exception from Zig code (with a user-defined message). My solution exposes the underlying JS context (aka scope), which I don't like, but I don't expect this to be a common requirement, so if we need to change things around, it shouldn't be a problem.

### DIFF
--- a/src/browser/console/console.zig
+++ b/src/browser/console/console.zig
@@ -30,39 +30,39 @@ pub const Console = struct {
     timers: std.StringHashMapUnmanaged(u32) = .{},
     counts: std.StringHashMapUnmanaged(u32) = .{},
 
-    pub fn static_lp(values: []JsObject, page: *Page) !void {
+    pub fn _lp(values: []JsObject, page: *Page) !void {
         if (values.len == 0) {
             return;
         }
         log.fatal(.console, "lightpanda", .{ .args = try serializeValues(values, page) });
     }
 
-    pub fn static_log(values: []JsObject, page: *Page) !void {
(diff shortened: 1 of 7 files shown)


sjorsdonkers (commented):
Works for airbnb

krichprollsch (approved):

PR #822, opened 2025-06-27
Add UndefinedOr(T) union

karlseguin:
Some apis want a value or undefined. For these, we can't use an Optional return type, null maps to JS null. Adds an Env.UndefinedOr(T) generic union for such return types.

### DIFF
--- a/src/runtime/js.zig
+++ b/src/runtime/js.zig
@@ -1860,8 +1860,8 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
 
         pub fn UndefinedOr(comptime T: type) type {
             return union(enum) {
-                value: T,
                 undefined: void,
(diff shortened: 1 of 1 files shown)


src/runtime/js.zig
```diff
@@ -1858,6 +1858,13 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
             }
         };
 
+        pub fn UndefinedOr(comptime T: type) type {
+            return union(enum) {
+                value: T,
+                undefined: void,
```
sjorsdonkers: Very minor, I would make undefined the first variant such that a zero initialized `UndefinedOr` is an undefined

karlseguin: No guarantee about that, but ok, done.

PR #824, opened 2025-06-27
create a DOM tree for non-html files

krichprollsch:
When using a non-html document through CDP, you still want to be able to take actions on a DOM tree.
Chrome builds a pseudo DOM tree for non-html webpages.

We display the content in a `<pre>` tag for well known content type and an empty document for others.


karlseguin:
LGTM

PR #829, opened 2025-06-30
add pump message loop calls

krichprollsch:


src/runtime/js.zig
```diff
@@ -260,6 +263,24 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
             self.isolate.performMicrotasksCheckpoint();
         }
 
+        pub fn pumpMessageLoop(self: *const Self) bool {
+            if (self.platform == null) {
```
karlseguin: Since these are potentially called in a tight loop, I think you can do this at comptime:

```zig
if (comptime builtin.is_test) {
    if (self.platform == null) return false;
}
// assume it's not-null in non-test.
return self.platform.?.inner.pumpMessageLoop(self.isolate, false);

karlseguin:
Besides the small [optional] optimization tweak above, PR is good.

Probably some opportunities to tighten up the test code. There's always a platform, but the fact that you don't have access to it in some parts of the test code is unfortunate. I think it's just that `serveCDP` that's the issue. Maybe we don't need those tests and instead rely on the integration / demo.

src/runtime/js.zig
krichprollsch: I was wondering if I should use a different slower loop for these tasks separated from the microtasks one.
WDYT?

karlseguin: I don't understand what this is doing, so it's hard to say. 

The page's `messageLoopCallback` is not being called very often. _all_ of page.navigate (including processing all synchronous scripts) are executed before it's ever called. This could change if navigate became fully asynchronous though.

It's only being called often when we're "done" navigating and waiting for
1 - New cdp sessions
2 - I/O (i.e. an XHR request is live)

In both cases, I would think it's mostly a no-op.

krichprollsch:
pump message loop could be relative to v8's garbage collector, but I'm not sure about that.
I suppose it's useful to run it time to time.
here we run it 1ms after page's initialize and then every 100ms. I think it's a good enough balance to start.

idk for idleTasks...

PR #833, opened 2025-07-01
Signal option for EventHandler

sjorsdonkers:
Support the `signal` option in `EventTarget.addEventListener`
https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#signal


sjorsdonkers:
Closed in favor of: https://github.com/lightpanda-io/browser/pull/851

PR #835, opened 2025-07-02
Remove HTTP client generic Loop parameter

karlseguin:
I think we initially thought we might need different clients for different parts of the system, each with a unique loop  (e.g. we thought telemetry might need some isolation). But that never happened, so it's just needless now, especially since the async connect uses the non-generic *Loop type directly.


sjorsdonkers:
Merging this for use in https proxy

PR #836, opened 2025-07-02
secure changes

sjorsdonkers:
In a separate branch atm as it wasn't always clear what the expected meaning of _secure was in different places.
Mainly in: `fn get(self: *ConnectionManager`


karlseguin:
lgtm

PR #840, opened 2025-07-03
Add readystate change event to XHR

karlseguin:
Deal with non-node current target crashing. Builds ontop of abort_signal, but with the new event-target specific union, I think this will work in for all future cases.


karlseguin:
> I'm wondering whether it would be good to also implement the state change on abort now: https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/abort

I tried to improve that too now.

PR #842, opened 2025-07-04
Rely on js.zig for float->int translation

karlseguin:
Not only does this ensure compatibility with browsers, it doesn't crash when the value is NaN of Infinity.

### DIFF
--- a/src/browser/html/document.zig
+++ b/src/browser/html/document.zig
@@ -233,19 +233,15 @@ pub const HTMLDocument = struct {
     // Since LightPanda requires the client to know what they are clicking on we do not return the underlying element at this moment
     // This can currenty only happen if the first pixel is clicked without having rendered any element. This will change when css properties are supported.
     // This returns an ElementUnion instead of a *Parser.Element in case the element somehow hasn't passed through the js runtime yet.
-    pub fn _elementFromPoint(_: *parser.DocumentHTML, x: f32, y: f32, page: *Page) !?ElementUnion {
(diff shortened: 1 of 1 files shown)


src/browser/html/document.zig
```diff
@@ -233,19 +233,15 @@ pub const HTMLDocument = struct {
     // Since LightPanda requires the client to know what they are clicking on we do not return the underlying element at this moment
     // This can currenty only happen if the first pixel is clicked without having rendered any element. This will change when css properties are supported.
     // This returns an ElementUnion instead of a *Parser.Element in case the element somehow hasn't passed through the js runtime yet.
-    pub fn _elementFromPoint(_: *parser.DocumentHTML, x: f32, y: f32, page: *Page) !?ElementUnion {
-        const ix: i32 = @intFromFloat(@floor(x));
-        const iy: i32 = @intFromFloat(@floor(y));
-        const element = page.renderer.getElementAtPosition(ix, iy) orelse return null;
+    pub fn _elementFromPoint(_: *parser.DocumentHTML, x: i32, y: i32, page: *Page) !?ElementUnion {
```
sjorsdonkers, line 240: I get why.. tho given that the interface should be floats (conversion to int is an internal renderer choice)
I would leave a comment informing why this decision was made, otherwise it would easily be revert be someone who notices the interface is off or people can just be surprised.

karlseguin: done

PR #843, opened 2025-07-04
An empty anchor should return empty strings for its getters

karlseguin:
document.createElement('a').host  or .href or .. should return an empty string.

However, URL.constructor(document.createElement('a')) should fail.

Because HTMLAnchorElement uses URL.constructor, we have the wrong behavior.

This adds a guard for an empty anchor. This might not cover all of the cases which are valid for an anchor but invalid for a URL.constructor, but it's the most common.

### DIFF
--- a/src/browser/html/elements.zig
+++ b/src/browser/html/elements.zig
@@ -274,7 +274,7 @@ pub const HTMLAnchorElement = struct {
     fn url(self: *parser.Anchor, page: *Page) !URL {
         // Although the URL.constructor union accepts an .{.element = X}, we
         // can't use this here because the behavior is different.
-        //    new URL(document.createElement('a')
+        //    URL.constructor(document.createElement('a')
         // should fail (a.href isn't a valid URL)
         // But
         //     document.createElement('a').host


src/browser/html/elements.zig
```diff
@@ -271,8 +271,18 @@ pub const HTMLAnchorElement = struct {
         return try parser.nodeSetTextContent(parser.anchorToNode(self), v);
     }
 
-    inline fn url(self: *parser.Anchor, page: *Page) !URL {
-        return URL.constructor(.{ .element = @alignCast(@ptrCast(self)) }, null, page); // TODO inject base url
+    fn url(self: *parser.Anchor, page: *Page) !URL {
+        // Although the URL.constructor union accepts an .{.element = X}, we
+        // can't use this here because the behavior is different.
+        //    new URL(document.createElement('a')
```
sjorsdonkers: Image

Image

I suspect the code is right, just the PR description uses `URL.constructor` instead of `new URL(..)`

karlseguin: Fixed, thanks.

PR #848, opened 2025-07-07
Fix non-tls forward-proxy

karlseguin:
Also renamed `isSimpleProxy` -> `isForwardProxy` for consistent with the proxy type name.

### DIFF
--- a/src/http/client.zig
+++ b/src/http/client.zig
@@ -711,12 +711,7 @@ pub const Request = struct {
                     }
                 }
             }
-
-
-            if (
-                (self._request_secure and !self._proxy_secure) and
-                (!self._client.isForwardProxy() or self._proxy_secure)
-            ) {
+            if (self._request_secure and !self._proxy_secure and !self._client.isForwardProxy()) {
                 self._connection.?.tls = .{
                     .blocking = try tls.client(std.net.Stream{ .handle = socket }, tls_config),
                 };


src/http/client.zig
```diff
@@ -711,7 +711,7 @@ pub const Request = struct {
                     }
                 }
             }
-            if (self._request_secure and !self._proxy_secure) {
+            if (self._request_secure and !self._proxy_secure and !self._client.isForwardProxy()) {
```
karlseguin, line 714: This is the only real change.

sjorsdonkers, line 714: I believe we would need this to support a secure forward proxy (`proxy_tls_config` need to be put in scope)

```suggestion
            if (self._client.isForwardProxy()) {
                if (self._proxy_secure) {
                    self._connection.?.tls = .{
                        .blocking = try tls.client(std.net.Stream{ .handle = socket }, proxy_tls_config),
                    };
                }
            } else if (self._request_secure and !self._proxy_secure) { // handles both insecure connect proxy and no proxy case
                self._connection.?.tls = .{
                    .blocking = try tls.client(std.net.Stream{ .handle = socket }, tls_config),
                };
            }
```

Starting to think the refactored version would have been better here anyway.
https://github.com/lightpanda-io/browser/pull/815#discussion_r2179573824

karlseguin: now it's:
```zig
if (
    (self._request_secure and !self._proxy_secure) and
    (!self._client.isForwardProxy() or self._proxy_secure)
) {
```

sjorsdonkers (approved):

src/http/client.zig
sjorsdonkers: Should the tls_config be changed to point to the right endpoint? if it is the tls for the proxy?

sjorsdonkers: ` or self._proxy_secure` can never make the whole statement true since  `and !self._proxy_secure` is also part of it

PR #851, opened 2025-07-07
Add support for the signal option of addEventListener

karlseguin:


src/browser/events/event.zig
```diff
@@ -262,6 +282,50 @@ pub const EventHandler = struct {
...
+    signal_listener: *parser.EventListener,
+
+    fn handle(node: *parser.EventNode, _: *parser.Event) void {
+        const self: *SignalCallback = @fieldParentPtr("node", node);
+        self._handle() catch |err| {
+            log.err(.app, "event signal handler", .{ .err = err });
+        };
+    }
+
+    fn _handle(self: *SignalCallback) !void {
```
sjorsdonkers, line 301: _name is generally used in this code base to mean it is a JavaScript callable function.
In this case I think it means this is the private implementation. So for consistency perhaps calling it handle_impl or so would be better

src/browser/events/event.zig
```diff
@@ -188,25 +189,44 @@ pub const EventHandler = struct {
...
             }
         }
 
         const callback = (try listener.callback(target)) orelse return null;
 
+        if (signal) |s| {
+            std.debug.print("add signal\n", .{});
+            const signal_target = parser.toEventTarget(AbortSignal, s);
+
+            const scb = try allocator.create(SignalCallback);
```
sjorsdonkers, line 211: I suspect most eventlisteners will not have an abortsignal so allocating it as opposed to keeping it in every object I think is better (as is implemented here)

src/browser/events/event.zig
karlseguin: We use `_name` in a lot of places to wrap a !void.

https://github.com/lightpanda-io/browser/blob/main/src/browser/page.zig#L609

https://github.com/lightpanda-io/browser/blob/main/src/browser/dom/mutation_observer.zig#L248

https://github.com/lightpanda-io/browser/blob/main/src/browser/dom/mutation_observer.zig#L248

...

PR #857, opened 2025-07-08
Improve prototype resolution for native types

karlseguin:
Prototype resolution of Zig types previously had 2 limitations (bug?). The first was that the Zig prototype chain could only be 1 deep. You couldn't do A->B->C where each of those was a Zig type (but you could do A->B->C->D->E ... so long as every other type was a C opaque value).

The other limitation was that Zig prototypes only worked when the nested field was directly embedded in the struct (i.e. not a pointer). So you could do:

```zig
const X = struct {
   proto: XParent,
};
```

But not:

```zig
const X = struct {
   proto: *XParent,
};
```

This addresses both limitations. The first issue is solved by keeping track of the cumulative offset

### DIFF
--- a/src/runtime/js.zig
+++ b/src/runtime/js.zig
@@ -81,14 +81,14 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
     // that looks like:
     //
     // const TypeLookup = struct {
-    //     comptime cat: usize = TypeMeta{.index = 0, ...},
-    //     comptime owner: usize = TypeMeta{.index = 1, ...},
+    //     comptime cat: usize = 0,
+    //     comptime owner: usize = 1,
     //     ...
     // }
     //
     // So to get the template index of `owner`, we can do:
     //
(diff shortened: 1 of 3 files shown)


src/runtime/js.zig
```diff
@@ -1387,6 +1416,78 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
...
+                // ...unless, the proto is behind a pointer, then total_offset will
+                // get reset to 0, and our base_ptr will move to the address
+                // referenced by the proto field.
+                var base_ptr: usize = @intFromPtr(toa.ptr);
+
+                // search through the prototype tree
+                while (true) {
+                    const proto_offset = meta_lookup[type_index].proto_offset;
+                    if (proto_offset < 0) {
+                        base_ptr = @as(*align(1) usize, @ptrFromInt(base_ptr + total_offset + @as(usize, @intCast(-proto_offset)))).*;
```
krichprollsch, line 1472: this is black magic :magic_wand:

krichprollsch (approved):

PR #861, opened 2025-07-08
NodeIterator

sjorsdonkers:
https://developer.mozilla.org/en-US/docs/Web/API/Document/createNodeIterator
Does not reuse the TreeWalker implementation as there are several subtle differences

TreeWalker is likely not handling its edges cases correctly, those are not addressed in this PR except fixing, `get_filter()`, `_previous_node()` going beyond the root and functions returning `*parser.Node` instead of `Node.Union`

### DIFF
--- a/src/browser/dom/document.zig
+++ b/src/browser/dom/document.zig
@@ -266,7 +266,7 @@ pub const Document = struct {
         return try TreeWalker.init(root, what_to_show, filter);
     }
 
-    pub fn _createNodeIterator(_: *parser.Document, root: *parser.Node, what_to_show: ?u32, filter: ?TreeWalker.TreeWalkerOpts) !NodeIterator {
+    pub fn _createNodeIterator(_: *parser.Document, root: *parser.Node, what_to_show: ?u32, filter: ?NodeIterator.NodeIteratorOpts) !NodeIterator {
         return try NodeIterator.init(root, what_to_show, filter);
     }
 
--- a/src/browser/dom/document.zig
+++ b/src/browser/dom/document.zig
@@ -33,6 +33,7 @@ const Element = @import("element.zig").Element;
 const ElementUnion = @import("element.zig").Union;
 const TreeWalker = @import("tree_walker.zig").TreeWalker;
 const CSSStyleSheet = @import("../cssom/css_stylesheet.zig").CSSStyleSheet;
+const NodeIterator = @import("node_iterator.zig").NodeIterator;
 const Range = @import("range.zig").Range;
 
 const Env = @import("../env.zig").Env;
@@ -263,6 +264,10 @@ pub const Document = struct {
 
     pub fn _createTreeWalker(_: *parser.Document, root: *parser.Node, what_to_show: ?u32, filter: ?TreeWalker.TreeWalkerOpts) !TreeWalker {
         return try TreeWalker.init(root, what_to_show, filter);
+    }
+
+    pub fn _createNodeIterator(_: *parser.Document, root: *parser.Node, what_to_show: ?u32, filter: ?TreeWalker.TreeWalkerOpts) !NodeIterator {
+        return try NodeIterator.init(root, what_to_show, filter);
     }
 
     pub fn getActiveElement(self: *parser.Document, page: *Page) !?*parser.Element {
--- a/src/browser/dom/node_filter.zig
+++ b/src/browser/dom/node_filter.zig
@@ -19,6 +19,7 @@
 const std = @import("std");
 const parser = @import("../netsurf.zig");
 const Env = @import("../env.zig").Env;
+const Node = @import("node.zig").Node;
 
 pub const NodeFilter = struct {
     pub const _FILTER_ACCEPT: u16 = 1;
@@ -63,7 +64,7 @@ pub fn verify(what_to_show: u32, filter: ?Env.Function, node: *parser.Node) !Ver
 
     // Verify that we aren't filtering it out.
     if (filter) |f| {
-        const acceptance = try f.call(u16, .{node});
+        const acceptance = try f.call(u16, .{try Node.toInterface(node)});
         return switch (acceptance) {
             NodeFilter._FILTER_ACCEPT => .accept,
             NodeFilter._FILTER_REJECT => .reject,
--- a/src/browser/dom/node_filter.zig
+++ b/src/browser/dom/node_filter.zig
@@ -22,6 +22,7 @@ pub const NodeFilter = struct {
     pub const _FILTER_ACCEPT: u16 = 1;
     pub const _FILTER_REJECT: u16 = 2;
     pub const _FILTER_SKIP: u16 = 3;
+
     pub const _SHOW_ALL: u32 = std.math.maxInt(u32);
     pub const _SHOW_ELEMENT: u32 = 0b1;
     pub const _SHOW_ATTRIBUTE: u32 = 0b10;
--- a/src/browser/dom/node_filter.zig
+++ b/src/browser/dom/node_filter.zig
@@ -17,6 +17,8 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 const std = @import("std");
+const parser = @import("../netsurf.zig");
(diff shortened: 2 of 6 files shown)


sjorsdonkers:
Edit: Fixed
Warnings when fetching: `https://www.airbnb.com/`
image

Error is raised when calling this filter function:
```js
let e;
for (!r && document.body && (r = document.createNodeIterator(document.body, NodeFilter.SHOW_ELEMENT, (e => {
        return t = e, f.includes(t.tagName) || t.hasAttribute("data-lcp-candidate") ? NodeFilter.FILTER_ACCEPT : NodeFilter.FILTER_REJECT;
        var t
    }))); e = r?.nextNode();) h(e);
n(N)
```

TBD if this is related to the ~~TreeWalker~~ NodeIterator  impl or something unrelated

karlseguin:
I'll look at the PR more completely in a bit, but I do think the issue is likely related to the implementation of the NodeIterator.

When I debug this, I see that the callback is called with a Node (as well as other elements), despite the code using the `NodeFilter.SHOW_ELEMENT`; which I assume means that it should _only_ iterate element nodes. But, because an underlying Node is passed, the call to tagName returns undefined (which is ok), and the call to `hasAttribute` fails - which is where the code fails.

Without looking at the implementation, I assume it's either emitting a node when it shouldn't (because of the filter), or it's emitting an element as a node without going through the `toInterface` call.

src/browser/dom/document.zig
```diff
@@ -264,6 +265,10 @@ pub const Document = struct {
         return try TreeWalker.init(root, what_to_show, filter);
     }
 
+    pub fn _createNodeIterator(_: *parser.Document, root: *parser.Node, what_to_show: ?u32, filter: ?NodeIterator.NodeIteratorOpts) !NodeIterator {
+        return try NodeIterator.init(root, what_to_show, filter);
```
karlseguin, line 270: FYI, you can return without the try. 

The only time you need the explicit try, is when the return type is optional, e.g. `!?NodeIterator` would require a `return try ...`. 

Codebase uses both styles, so either is fine.

src/browser/dom/node_filter.zig
```diff
@@ -37,6 +40,39 @@ pub const NodeFilter = struct {
     pub const _SHOW_NOTATION: u32 = 0b100000000000;
 };
 
+const VerifyResult = enum { accept, skip, reject };
+
+pub fn verify(what_to_show: u32, filter: ?Env.Function, node: *parser.Node) !VerifyResult {
+    const node_type = try parser.nodeType(node);
+
+    // Verify that we can show this node type.
+    if (!switch (node_type) {
```
karlseguin, line 50: A Zig-ish alternative to this is `@bitCast` the u32 into a packed struct:

```zig
pub const WhatToShow = packed struct {
    element: bool,
    attribute: bool,
    text: bool,
    cdata: bool,
    entity_reference: bool,
    entity: bool,
    processing_instruction: bool,
    comment: bool,
    document: bool,
    document_type: bool,
    document_fragment: bool,
    notation: bool,
    _: u20,
};

pub fn verify(what_to_show: u32, filter: ?Env.Function, node: *parser.Node) !VerifyResult {
    const node_type = try parser.nodeType(node);
     // or WhatToShow.fromBitmask(what_to_show) if you want to encapsulate it
    const wts: WhatToShow = @bitCast(what_to_show);
   
   if (!switch (node_type) {
      .attribute => wts.atribute,
      , ....
     })
```

src/browser/dom/node_filter.zig
```diff
@@ -37,6 +40,39 @@ pub const NodeFilter = struct {
...
+        .entity => what_to_show & NodeFilter._SHOW_ENTITY != 0,
+        .entity_reference => what_to_show & NodeFilter._SHOW_ENTITY_REFERENCE != 0,
+        .notation => what_to_show & NodeFilter._SHOW_NOTATION != 0,
+        .processing_instruction => what_to_show & NodeFilter._SHOW_PROCESSING_INSTRUCTION != 0,
+        .text => what_to_show & NodeFilter._SHOW_TEXT != 0,
+    }) return .reject;
+
+    // Verify that we aren't filtering it out.
+    if (filter) |f| {
+        const acceptance = try f.call(u16, .{node});
```
karlseguin: I guess this is where the `toInterface` is missing

src/browser/dom/node_iterator.zig
```diff
@@ -0,0 +1,288 @@
...
+
+                    // Otherwise, this node is our previous one.
+                    self.reference_node = current;
+                    return current;
+                },
+                .reject, .skip => {
+                    // Get last child if it has one.
+                    if (try self.lastChild(current)) |child| {
+                        self.reference_node = child;
+                        return child;
```
karlseguin: If the current node is skip/reject, we still return it's last child?

src/browser/dom/node_iterator.zig
```diff
@@ -0,0 +1,288 @@
...
+        var current = self.reference_node;
+        while (try parser.nodePreviousSibling(current)) |previous| {
+            current = previous;
+
+            switch (try NodeFilter.verify(self.what_to_show, self.filter_func, current)) {
+                .accept => {
+                    // Get last child if it has one.
+                    if (try self.lastChild(current)) |child| {
+                        self.reference_node = child;
+                        return child;
```
karlseguin: Does this have to be verified?

src/browser/dom/node_iterator.zig
```diff
@@ -0,0 +1,288 @@
...
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+const std = @import("std");
+const parser = @import("../netsurf.zig");
+const Env = @import("../env.zig").Env;
+const NodeFilter = @import("node_filter.zig");
+
+// https://developer.mozilla.org/en-US/docs/Web/API/NodeIterator
+// While this is similar to TreeWalker it has it's own implementation as there are several suttle differences
```
karlseguin:
```suggestion
// While this is similar to TreeWalker it has its own implementation as there are several subtle differences
```

src/browser/dom/node_iterator.zig
sjorsdonkers: If lastChild finds one that is accepted, yes.

One of the main differences with TreeWalker is that `FILTER_REJECT ` is handled as a skip.

> For createNodeIterator, the values NodeFilter.FILTER_REJECT and NodeFilter.FILTER_SKIP are equivalent. This node will not be included in the list of iterable nodes, but its children will continue to be iterated over.

src/browser/dom/node_iterator.zig
sjorsdonkers: It is yes, lastChild will verify.

src/browser/dom/node_iterator.zig
sjorsdonkers: 😅 [93ea4ed](https://github.com/lightpanda-io/browser/pull/861/commits/93ea4ed51a8b6461c70219b70d0a2b2ffa23af56)

src/browser/dom/node_filter.zig
sjorsdonkers: It was great find, thanks!
Fixed: [93ea4ed](https://github.com/lightpanda-io/browser/pull/861/commits/93ea4ed51a8b6461c70219b70d0a2b2ffa23af56)

src/browser/dom/node_filter.zig
sjorsdonkers: Thanks! Good to know.
In this case I'm inclined to leave it as it was. Just fewer loc.
If you do feel strongly about I would also change the u32 stored in the NodeIterator and TreeWalker, and cast it back when that property is accessed. Otherwise I'll leave it as is.

src/browser/dom/document.zig
sjorsdonkers: Yup, it is intentional.
I got surprised a couple of times when reading code that didn't have the try, so I prefer it.

karlseguin:
LGTM

PR #864, opened 2025-07-09
add HTMLElementLink get/set href

karlseguin:


src/browser/html/elements.zig
```diff
@@ -1581,6 +1590,12 @@ test "Browser.HTML.Element" {
         .{ "document.createElement('a').focus()", null },
         .{ "document.activeElement === focused", "true" },
     }, .{});
+
+    try runner.testCases(&.{
+        .{ "let l2 = document.createElement('link');", null },
+        .{ "l2.href = 'https://lightpanda.io/'", null },
+        .{ "l2.href", "https://lightpanda.io/" },
```
sjorsdonkers: This maybe a slightly more interesting test 🤷
```suggestion
        .{ "l2.href", "" },
        .{ "l2.href = 15", null },
        .{ "l2.href", "https://lightpanda.io/15" },
```

sjorsdonkers: @karlseguin whoops should be `https://lightpanda.io/opensource-browser/15`

PR #872, opened 2025-07-09
Dynamic import

mookums:
This adds support for dynamic imports of modules (eg. `await import("xyz")`). This is in draft at the moment as it is dependent on https://github.com/lightpanda-io/zig-v8-fork/pull/83 as it requires the `isPromise()`.

### DIFF
--- a/src/runtime/js.zig
+++ b/src/runtime/js.zig
@@ -767,9 +767,8 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
             pub fn module(self: *JsContext, src: []const u8, url: []const u8, cacheable: bool) !?v8.Promise {
                 const arena = self.context_arena;
 
-                if (cacheable) {
-                    const value = self.module_cache.get(url);
-                    if (value != null) return null;
+                if (cacheable and self.module_cache.contains(url)) {
+                    return null;
                 }
                 errdefer _ = self.module_cache.remove(url);
 
@@ -780,7 +779,11 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
                 errdefer _ = self.module_identifier.remove(m.getIdentityHash());
 
                 if (cacheable) {
-                    try self.module_cache.put(arena, owned_url, PersistentModule.init(self.isolate, m));
+                    try self.module_cache.putNoClobber(
+                        arena,
+                        owned_url,
+                        PersistentModule.init(self.isolate, m),
+                    );
                 }
 
                 // resolveModuleCallback loads module's dependencies.
@@ -1552,14 +1555,14 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
             }
 
             fn _dynamicModuleCallback(
-                context: *JsContext,
+                self: *JsContext,
                 specifier: []const u8,
                 resolver: *const v8.PromiseResolver,
             ) !void {
-                const iso = context.isolate;
-                const ctx = context.v8_context;
+                const iso = self.isolate;
+                const ctx = self.v8_context;
 
-                const module_loader = context.module_loader;
+                const module_loader = self.module_loader;
                 const source = module_loader.func(module_loader.ptr, specifier) catch {
                     const error_msg = v8.String.initUtf8(iso, "Failed to load module");
                     _ = resolver.reject(ctx, error_msg.toValue());
@@ -1571,34 +1574,36 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
                 };
 
                 var try_catch: TryCatch = undefined;
-                try_catch.init(context);
+                try_catch.init(self);
                 defer try_catch.deinit();
 
-                const maybe_promise = context.module(source, specifier, true) catch {
+                const maybe_promise = self.module(source, specifier, true) catch {
                     log.err(.js, "module compilation failed", .{
                         .specifier = specifier,
-                        .exception = try_catch.exception(context.call_arena) catch "unknown error",
-                        .stack = try_catch.stack(context.call_arena) catch null,
+                        .exception = try_catch.exception(self.call_arena) catch "unknown error",
+                        .stack = try_catch.stack(self.call_arena) catch null,
                         .line = try_catch.sourceLineNumber() orelse 0,
                     });
                     const error_msg = if (try_catch.hasCaught()) blk: {
-                        const exception_str = try_catch.exception(context.call_arena) catch "Evaluation error";
+                        const exception_str = try_catch.exception(self.call_arena) catch "Evaluation error";
                         break :blk v8.String.initUtf8(iso, exception_str orelse "Evaluation error");
                     } else v8.String.initUtf8(iso, "Module evaluation failed");
                     _ = resolver.reject(ctx, error_msg.toValue());
                     return;
                 };
-                const new_module = context.module_cache.get(specifier).?.castToModule();
+                const new_module = self.module_cache.get(specifier).?.castToModule();
 
                 if (maybe_promise) |promise| {
                     // This means we must wait for the evaluation.
                     const EvaluationData = struct {
+                        specifier: []const u8,
                         module: v8.Persistent(v8.Module),
                         resolver: v8.Persistent(v8.PromiseResolver),
                     };
 
-                    const ev_data = try context.context_arena.create(EvaluationData);
+                    const ev_data = try self.context_arena.create(EvaluationData);
                     ev_data.* = .{
+                        .specifier = specifier,
                         .module = v8.Persistent(v8.Module).init(iso, new_module),
                         .resolver = v8.Persistent(v8.PromiseResolver).init(iso, resolver.*),
                     };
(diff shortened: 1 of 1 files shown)


src/runtime/js.zig
```diff
@@ -1355,6 +1358,209 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
...
+                    return @constCast(resolver.getPromise().handle);
+                };
+                const resource: v8.String = .{ .handle = resource_name.? };
+                const resource_str = jsStringToZig(context.call_arena, resource, iso) catch {
+                    const error_msg = v8.String.initUtf8(iso, "Failed to parse module resource");
+                    _ = resolver.reject(ctx, error_msg.toValue());
+                    return @constCast(resolver.getPromise().handle);
+                };
+
+                const referrer_full_url = blk: {
```
karlseguin: Is this necessary? I would have thought that you can use `resource_str` directly as the key to `context.module_identifier`.

src/runtime/js.zig
```diff
@@ -1355,6 +1358,209 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
...
+                };
+
+                const normalized_specifier = @import("../url.zig").stitch(
+                    context.context_arena,
+                    specifier_str,
+                    referrer_full_url,
+                    .{ .alloc = .if_needed },
+                ) catch unreachable;
+
+                _dynamicModuleCallback(context, resource_str, normalized_specifier, &resolver);
```
karlseguin: There's a few `unreachables` in `_dynamicModuleCallback`. Why not do:

```zig
 _dynamicModuleCallback(context, resource_str, normalized_specifier, &resolver) catch |err| {
  // todo log err
};
```

And then you can change the return type of `_dynamicModuleCallback` to `!void` and use `try` instead of `catch unreachable` (especially for code that isn't actually unreachable).

src/runtime/js.zig
```diff
@@ -1355,6 +1358,209 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
...
+                return @constCast(resolver.getPromise().handle);
+            }
+
+            fn _dynamicModuleCallback(context: *JsContext, resource: []const u8, specifier: []const u8, resolver: *const v8.PromiseResolver) void {
```
karlseguin: I think it's worth exploring whether `dynamicModuleCallback` can leverage the existing `module` function. `module` doesn't have its own try catch, so I think the rejection parts shouldn't be an issue:

```zig
var try_catch: TryCatch = undefined;
try_catch.init(context);
defer try_catch.deinit();

// change module to return the module so that you can evulate it, I gues
const m = self.module(....) catch |err| {
  var error_message: ?[]const u8 = null;
  if (try_catch.hasCaught()) {
      // ...
  } else switch (err) {
     // if you want to handle specific errors from module(...)
  }
}
```

Not having to duplicate the specifier normalization, cache management and module compilation/evaluation seems like a win.

src/runtime/js.zig
mookums: Sometimes using the resource works but other times like below, it doesn't.

```
INFO  js : dynamic import . . . . . . . . . . . . . . . . . . [+0ms]
      specifier = ./shreddit-overflow-menu-31f50e45.js
      resource = ./comment-client-js-4adc8893.js
      referrer_full = https://www.redditstatic.com/shreddit/en-US/comment-client-js-4adc8893.js
      normalized_specifier = ./shreddit-overflow-menu-31f50e45.js
 ```

File: src/runtime/js.zig (hunk 3 above)
Comment (mookums) [reply src/runtime/js.zig]: Ended up changing `module()` so it now returns `!?v8.Promise` where it returns null when the module is already in the cache and returns the evaluation result, which should always be a `v8.Promise`.

I ended up removing `moduleNoCache()` and adding the functionality to the normal `module()` because it felt like a lot was being duplicated across the two.

File: src/runtime/js.zig [hunk 4]
```diff
@@ -768,17 +771,14 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
...
+            // It returns null if the module is already compiled and in the cache.
+            // It returns a v8.Promise if the module must be evaluated.
+            pub fn module(self: *JsContext, src: []const u8, url: []const u8, cacheable: bool) !?v8.Promise {
                 const arena = self.context_arena;
 
-                const gop = try self.module_cache.getOrPut(arena, url);
-                if (gop.found_existing) {
-                    return;
+                if (cacheable) {
+                    const value = self.module_cache.get(url);
```
karlseguin:
```
if (self.module_cache.contains(url)) {
   return null;
}
```

src/runtime/js.zig
```diff
@@ -788,30 +788,22 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
                 try self.module_identifier.putNoClobber(arena, m.getIdentityHash(), owned_url);
                 errdefer _ = self.module_identifier.remove(m.getIdentityHash());
 
-                gop.key_ptr.* = owned_url;
-                gop.value_ptr.* = PersistentModule.init(self.isolate, m);
+                if (cacheable) {
+                    try self.module_cache.put(arena, owned_url, PersistentModule.init(self.isolate, m));
```
karlseguin: Can this be a `putNoClobber`?

src/runtime/js.zig
```diff
@@ -1516,6 +1508,184 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
...
+                            const cb_info = v8.FunctionCallbackInfo{ .handle = info.? };
+                            const cb_isolate = cb_info.getIsolate();
+                            const cb_context = cb_isolate.getCurrentContext();
+                            const data: *EvaluationData = @ptrCast(@alignCast(cb_info.getExternalValue()));
+                            const cb_module = data.module.castToModule();
+                            const cb_resolver = data.resolver.castToPromiseResolver();
+
+                            const namespace = cb_module.getModuleNamespace();
+                            log.info(.js, "dynamic import complete", .{
+                                .module = cb_module,
```
karlseguin: these are opaque values, it just prints:

```
      module = v8.Module{ .handle = cimport.struct_Module@12e09b500 }
      namespace = v8.Value{ .handle = cimport.struct_Data@12e04d208
```

src/runtime/js.zig
```diff
@@ -1516,6 +1508,184 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
...
+                    });
+                    // Must be rejected at this point
+                    // otherwise, we will just wait on a pending promise.
+                    std.debug.assert(resolver.getPromise().getState() == .kRejected);
+                };
+                return @constCast(resolver.getPromise().handle);
+            }
+
+            fn _dynamicModuleCallback(
+                context: *JsContext,
```
karlseguin:
```suggestion
                self: *JsContext,
```

?

src/runtime/js.zig
```diff
@@ -1516,6 +1508,184 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
...
+                        .line = try_catch.sourceLineNumber() orelse 0,
+                    });
+                    const error_msg = if (try_catch.hasCaught()) blk: {
+                        const exception_str = try_catch.exception(context.call_arena) catch "Evaluation error";
+                        break :blk v8.String.initUtf8(iso, exception_str orelse "Evaluation error");
+                    } else v8.String.initUtf8(iso, "Module evaluation failed");
+                    _ = resolver.reject(ctx, error_msg.toValue());
+                    return;
+                };
+                const new_module = context.module_cache.get(specifier).?.castToModule();
```
karlseguin: I don't see how it's possible, but when I tried this branch with reddit, `context.module_cache.get(specifier)` returns null (and this code crashed). But I don't see how `context.module()` called with `cacheable: true` can return without inserting that entry...

src/runtime/js.zig
karlseguin: We should look into this. I started to look into this, but the crash on 1620  (and, to a lesser extent, the logging of opaque values), kind of stopped me.

I _think_ the `resource` value is always supplied by us, and, for correct caching, it should always be the full url. We should find out where we're submitting the non-full script name `./comment-client-js-4adc8893.js` and fix it (maybe there's a combination of of base + url that `stitch` doesn't handle properly).

(It's possible the name is somehow derived by v8, and we have no choice, but that seems unlikely, and I rather make sure that, because it smells like a bug to me..and fixing it wouldn't just fix potentially weird caching issue, it would let you remove this block of code).

src/runtime/js.zig
karlseguin: The `call_arena` isn't safe to use, e.g.

```zig
 const specifier_str = jsStringToZig(context.call_arena, specifier, iso) catch {
```

The exact lifetime of the call_arena is a bit unspecific, but it's guaranteed to exist for at least the point where a WebAPI function is called (e.g. Window._setTimeout) to the point where data is handed to v8.

Loading a module is _a lot_ wider than that and likely includes many WebApi calls. So imagine you have a module:

```js
// awesome.js
console.log("hello");
// <-- should assume that the call_arena is reset at this point
```
Your specifier (and probably other variables) are created before this code executes, but those allocations will become invaild as soon as `console.log` returns.

src/runtime/js.zig
karlseguin: After fixing the crash, when I looked at this for reddit, the resource was always a full url, including for `comment-client-js-cc940b94.js`

src/runtime/js.zig
mookums: Logs the actual specifier now.

src/runtime/js.zig
mookums: I never really crashed running this but I switched from `call_arena` to`context_arena` and now the resource_str works so I assume that's also how you fixed the crash?

PR #874, opened 2025-07-10
add dummy document.get_styleSheets

karlseguin:


src/browser/dom/document.zig
```diff
@@ -295,6 +296,10 @@ pub const Document = struct {
     pub fn _createRange(_: *parser.Document, page: *Page) Range {
         return Range.constructor(page);
     }
+
+    pub fn get_styleSheets(_: *parser.Document) []CSSStyleSheet {
+        return &.{};
```
sjorsdonkers, line 302: It might be good for us to add warnings in dummy implementations. To leave some bread crumbs, when someone in the future is not crashing but getting unexpected values

sjorsdonkers (approved):

PR #875, opened 2025-07-10
Fix async https requests over a http forward proxy

karlseguin:
XHR requests to https (which is most XHR requests) currently don't work with the implementation proxy because of this.

### DIFF
--- a/src/http/client.zig
+++ b/src/http/client.zig
@@ -801,7 +801,7 @@ pub const Request = struct {
         }
 
         const is_proxy = self._client.isProxy();
-        if ((is_proxy and self._proxy_secure) or (!is_proxy and self._request_secure)) {
+        if ((is_proxy and self._proxy_secure) or (!is_forward_proxy and self._request_secure)) {
             if (self._connection_from_keepalive) {
                 // If the connection came from the keepalive pool, than we already
                 // have a TLS Connection.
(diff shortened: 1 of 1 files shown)


src/http/client.zig
```diff
@@ -792,8 +796,12 @@ pub const Request = struct {
...
         };
 
-        if (self._client.isConnectProxy() and self._proxy_secure) log.warn(.http, "ASYNC TLS CONNECT no impl.", .{});
-        if (self._request_secure) {
+        if (self._client.isConnectProxy() and self._proxy_secure) {
+            log.warn(.http, "not implemented", .{ .feature = "async tls connect" });
+        }
+
+        const is_proxy = self._client.isProxy();
+        if ((is_proxy and self._proxy_secure) or (!is_proxy and self._request_secure)) {
```
sjorsdonkers: If the connection is to a CONNECT proxy that is not secure, but the request is secure.
We are still making the handshake and such, so we could benefit from reusing that connection.
```suggestion
        if ((is_proxy and self._proxy_secure) or ((!is_proxy or is_connect_proxy) and self._request_secure)) {
```
or
```suggestion
        if ((is_proxy and self._proxy_secure) or (!is_forward_proxy and self._request_secure)) {
```

sjorsdonkers: In general I think we should revisit the connection manager more carefully

sjorsdonkers: Apologies, I didn't mean for the suggestions to be merged as is. I'll be clearer in the future.

sjorsdonkers: Compiling version of the change
```suggestion
        if ((is_proxy and self._proxy_secure) or (!self._client.isForwardProxy() and self._request_secure)) {
```

PR #881, opened 2025-07-11
add window.queueMicrotask

karlseguin:


krichprollsch:
I don't think it changes anything but maybe we should use [`EnqueueMicrotask`](https://v8.github.io/api/head/classv8_1_1MicrotaskQueue.html#a21df1c7eac1a206aa139d7d86dee5e7c ) ?

karlseguin:
I'd like to keep it as is. It makes `setTimeout`, `setInterval`, `requestAnimationFrame`, `requestIdleCallback` (TBD) and `queueMicrotask` all have the same implementation. It's also 1 less layer of indirection and keeps control of it in our code, rather than passing it to v8.

Practically, there isn't much difference, since the v8 microtask is itself run from a IO loop timeout and both will run on the next `run` call.

krichprollsch (approved):

Issue #866, opened 2025-07-09
chromedp.Dump doesn't work


feeops:
the code
```
package main

import (
	"context"
	"fmt"
	"github.com/chromedp/chromedp"
	"os"
)

func run() {

	t := "https://www.youtube.com"

	allocatorCtx, _ := chromedp.NewRemoteAllocator(context.Background(), "ws://127.0.0.1:9222")
	ctx, _ := chromedp.NewContext(allocatorCtx)
	var url string
	if err := chromedp.Run(ctx, []chromedp.Action{
		chromedp.Navigate(t),
		// chromedp.Sleep(10 * time.Second),

		chromedp.Location(&url),
		chromedp.Dump("html", os.Stdout),
	}...); err != nil {
		fmt.Println("t", err)
	}

	fmt.Println("URL:", url)

}

func main() {

	run()
}

```

output
```
t context canceled
URL: https://www.youtube.com/
```

```
 ./lightpanda version
bdc49a6
```

krichprollsch:
Hello @feeops,

Thanks for you issue.
It looks like we don't support `chromedp.Dump` for now.
I will take a look at it to see how it is implemented.

In the meanwhile, you can achieve an equivalent result with:

```go
package main

import (
	"context"
	"fmt"

	"github.com/chromedp/chromedp"
)

func run() {
	t := "https://www.youtube.com"
	var content string

	allocatorCtx, _ := chromedp.NewRemoteAllocator(context.Background(), "ws://127.0.0.1:9222")
	ctx, _ := chromedp.NewContext(allocatorCtx)
	var url string
	if err := chromedp.Run(ctx, []chromedp.Action{
		chromedp.Navigate(t),
		// chromedp.Sleep(10 * time.Second),

		chromedp.Location(&url),
		chromedp.OuterHTML("html", &content),
	}...); err != nil {
		fmt.Println("t", err)
	}

	fmt.Println(content)
	fmt.Println("t", url)

}

func main() {
	run()
}

```

karlseguin:
The original code using `chromedp.Dump` should be working.

However, I noticed that Chrome doesn't emit _every_ node, whereas we do. It looks like maybe it leaves out script tags. I need to investigate this more..but for now, i figure having too much data is still a step in the right direction.

PR #888, opened 2025-07-14
Add support for CDP's DOM.requestChildNodes

karlseguin:
https://github.com/lightpanda-io/browser/issues/866

### DIFF
--- a/src/cdp/Node.zig
+++ b/src/cdp/Node.zig
@@ -213,7 +213,7 @@ pub const Writer = struct {
 
     pub fn jsonStringify(self: *const Writer, w: anytype) !void {
         if (self.exclude_root) {
-            _ = self.writeChildren(self.root, 0, w) catch |err| {
+            _ = self.writeChildren(self.root, 1, w) catch |err| {
                 log.err(.cdp, "node writeChildren", .{ .err = err });
                 return error.OutOfMemory;
             };
@@ -467,7 +467,7 @@ test "cdp Node: Writer" {
         const node = registry.lookup_by_id.get(1).?;
         const json = try std.json.stringifyAlloc(testing.allocator, Writer{
             .root = node,
-            .depth = 0,
+            .depth = 1,
             .exclude_root = false,
             .registry = &registry,
         }, .{});
--- a/src/cdp/domains/dom.zig
+++ b/src/cdp/domains/dom.zig
@@ -442,6 +442,7 @@ fn requestChildNodes(cmd: anytype) !void {
         pierce: bool = false,
     })) orelse return error.InvalidParams;
 
+    if (params.depth == 0) return error.InvalidParams;
     const bc = cmd.browser_context orelse return error.BrowserContextNotLoaded;
     const session_id = bc.session_id orelse return error.SessionIdNotLoaded;
     const node = bc.node_registry.lookup_by_id.get(params.nodeId) orelse {


src/cdp/domains/dom.zig
```diff
@@ -433,6 +435,29 @@ fn getBoxModel(cmd: anytype) !void {
     } }, .{});
 }
 
+fn requestChildNodes(cmd: anytype) !void {
+    const params = (try cmd.params(struct {
+        nodeId: Node.Id,
+        depth: i32 = 1,
+        pierce: bool = false,
```
sjorsdonkers, line 442: May want to warn or so when pierce is set to true that this is not implemented yet

src/cdp/Node.zig
```diff
@@ -201,49 +201,69 @@ pub const Search = struct {
...
 pub const Writer = struct {
-    opts: Opts,
-    node: *const Node,
+    depth: i32,
+    exclude_root: bool,
+    root: *const Node,
     registry: *Registry,
 
-    pub const Opts = struct {};
+    pub const Opts = struct {
```
sjorsdonkers, line 209: `DOM.getDocument` I believe uses the undocumented value `3` by default. So may be good to update the other users in this pr as well, and connect the parameters.
For `DOM.describeNode` the default is `1`, however it would be good to wire it in.
`performSearch` and `querySelector` most likely use the default, but I did not check to confirm

src/cdp/Node.zig
```diff
@@ -201,49 +201,69 @@ pub const Search = struct {
...
+        depth: i32 = 0,
+        exclude_root: bool = false,
+    };
 
     pub fn jsonStringify(self: *const Writer, w: anytype) !void {
```
sjorsdonkers, line 214: Potentially a good place to check that `depth != 0`, alternatively can catch `0`s at every CDP entrance call that has the parameter

src/cdp/Node.zig
```diff
@@ -201,49 +201,69 @@ pub const Search = struct {
...
-            // error, we can only return a writer error. Kinda sucks.
-            log.err(.cdp, "json stringify", .{ .err = err });
-            return error.OutOfMemory;
-        };
+        if (self.exclude_root) {
+            _ = self.writeChildren(self.root, 0, w) catch |err| {
+                log.err(.cdp, "node writeChildren", .{ .err = err });
+                return error.OutOfMemory;
+            };
+        } else {
```
sjorsdonkers, line 220: I'm assume that the meaning of the depth value 1, 2, 3 has been confirmed for both cases

src/cdp/Node.zig
```diff
@@ -201,49 +201,69 @@ pub const Search = struct {
...
-                const child = (try parser.nodeListItem(child_nodes, @intCast(i))) orelse break;
-                const child_node = try registry.register(child);
+        try w.endObject();
+    }
+
+    fn writeChildren(self: *const Writer, node: *const Node, depth: usize, w: anytype) anyerror!usize {
+        var registry = self.registry;
+        const child_nodes = try parser.nodeGetChildNodes(node._node);
+        const child_count = try parser.nodeListLength(child_nodes);
+        const full_child = self.depth < 0 or self.depth < depth;
```
sjorsdonkers, line 247:
```suggestion
        const full_child = self.depth < 0 or depth < self.depth;
```
My understanding is that if  `self.depth` is 1 we only get one layer of nodes.
So in the `exclude_root == true` case that would be a single list of nodes without any of their children populated.
So I would expect this function to be called with the parameter `depth: usize` set to `1` (now called with `0`).

In the `exclude_root == false` I would expect `writeChildren` not to be called at all as just the root node would be written. So we'd need this if statement also in the `toJSON` function.

I would not be surpised if I'm off by one tho.

src/cdp/Node.zig
sjorsdonkers: I just now noticed the default value is set to 0, but zero is not a valid value.
With the first comment I meant that the `depth != 0` could be at the start of the `jsonStringify` function

karlseguin: depth == 0 is valid on some calls, i.e. DOM.getDocument, so I pushed the check to requestChildNodes

PR #891, opened 2025-07-14
Add childElementCount and children to DocumentFragment

karlseguin:
Also, when shadowRoot is re-attached to an element, clear all existing children (like we're supposed to)


mookums (approved):
looks good to me

PR #893, opened 2025-07-14
Add a --noscript option to "improve" --dump

karlseguin:
Currently, fetch --dump includes <script> tag (either inline or with src). I don't know what use-case this is the desired behavior. Excluding them, via the new --noscript option has benefit that if you --dump --noscript and open the resulting page in the browser, you don't re-execute JavaScript, which is likely to break the page.

For example, opening a --dump of github makes it look like the page is broken because it re-executes JavaScript that isn't meant to be re-executed.

Similarly, opening a --dump in a browser might execute JavaScript that lightpanda browser failed to execute, making it looks like it worked better than it did.

### DIFF
--- a/src/browser/dom/element.zig
+++ b/src/browser/dom/element.zig
@@ -110,13 +110,13 @@ pub const Element = struct {
 
     pub fn get_innerHTML(self: *parser.Element, page: *Page) ![]const u8 {
         var buf = std.ArrayList(u8).init(page.arena);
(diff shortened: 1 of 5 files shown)


karlseguin:
@francisbouvier Any thought on making this the default?

francisbouvier:
I agree with you, I don't see the use case to have JS script in the dump.
Do we make this the default? It will change the behaviour but it just makes more sense to me.

PR LGMT

PR #897, opened 2025-07-15
Add Element.animate and Animation

karlseguin:
These are dummy implementations, but they do expose the ready and finished promise, and do resolve the finished promise, so it should unblock basic cases.

### DIFF
--- a/src/browser/dom/Animation.zig
+++ b/src/browser/dom/Animation.zig
@@ -1,0 +1,126 @@
+// Copyright (C) 2023-2025s  Lightpanda (Selecy SAS)
+//
+// Francis Bouvier <francis@lightpanda.io>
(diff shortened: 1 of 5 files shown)


mookums:
looks fine to me. just curious what our general consensus is on the file-as-a-struct thing?

karlseguin:
I personally find it troublesome. Always seems like sooner or later, you'll want to convert the type to a generic, or a union. But I keep trying to use it, because I find `const SomeLongName = @import("some_long_name.zig").SomeLongName` insulting to the aesthetics of the code.

My logic has been that, for these WebAPIs at least, they should be razor-focused. There's no reason that `Animation` or `DataSet` should be anything but a POJO..and the file-as-a-struct thing helps make the point.

You?

PR #899, opened 2025-07-17
Add element.remove()  (needed by reddit)

karlseguin:


src/browser/dom/element.zig
```diff
@@ -508,6 +508,12 @@ pub const Element = struct {
         _ = opts;
         return Animation.constructor(effect, null);
     }
+
+    pub fn _remove(self: *parser.Element) !void {
+        const as_node: *parser.Node = @ptrCast(self);
+        const parent = try parser.nodeParentNode(as_node) orelse return;
+        _ = try Node._removeChild(parent, as_node);
```
sjorsdonkers, line 519: hmm, I doubt this will update all location where a pointer to the deleted element was stored.
For example:

Image

Code:
```js
document.body.innerHTML = `
  <div class="preference">
    <label for="cheese">Do you like cheese?</label>
    <input type="checkbox" name="cheese" id="cheese" />
  </div>
`;

const label = document.querySelector('label[for="cheese"]');
label.control

const input = document.getElementById('cheese');
if (input) {
  input.remove();
}

label.control
```
Maybe just leave a TODO

sjorsdonkers: Alternatively elements could have a tombstone flag

karlseguin: Isn't this a widespread issue for any part of the code that removes/moves nodes?

sjorsdonkers: It is, I personally prefer leaving breadcrumbs around to help our future debugging selves.

PR #900, opened 2025-07-17
support `depth` parameter for DOM.getDocument

karlseguin:


src/cdp/domains/dom.zig
```diff
@@ -60,17 +60,18 @@ pub fn processMessage(cmd: anytype) !void {
...
+    })) orelse return error.InvalidParams;
 
     const bc = cmd.browser_context orelse return error.BrowserContextNotLoaded;
     const page = bc.session.currentPage() orelse return error.PageNotLoaded;
     const doc = parser.documentHTMLToDocument(page.window.document);
 
     const node = try bc.node_registry.register(parser.documentToNode(doc));
-    return cmd.sendResult(.{ .root = bc.nodeWriter(node, .{}) }, .{});
+    return cmd.sendResult(.{ .root = bc.nodeWriter(node, .{ .depth = params.depth orelse 3 }) }, .{});
 }
```
sjorsdonkers: DOM.getDocument can nolonger be called without parameters. To fix that we should handle the null case:
```zig
fn getDocument(cmd: anytype) !void {
    const Params = struct {
        // CDP documentation implies that 0 isn't valid, but it _does_ work in Chrome
        depth: i32 = 3,
        pierce: bool = false,
    };
    const params = try cmd.params(Params) orelse Params{};
    if (params.pierce) return error.NotYetImplementedParams;

    const bc = cmd.browser_context orelse return error.BrowserContextNotLoaded;
    const page = bc.session.currentPage() orelse return error.PageNotLoaded;
    const doc = parser.documentHTMLToDocument(page.window.document);

    const node = try bc.node_registry.register(parser.documentToNode(doc));
    return cmd.sendResult(.{ .root = bc.nodeWriter(node, .{ .depth = params.depth }) }, .{});
}
```

src/cdp/domains/dom.zig
```diff
@@ -610,6 +611,7 @@ test "cdp.dom: getBoxModel" {
     try ctx.processMessage(.{ // Hacky way to make sure nodeId 0 exists in the registry
         .id = 3,
         .method = "DOM.getDocument",
+        .params = .{ .depth = 3 },
```
sjorsdonkers: Considering keeping the test without params as well

src/cdp/domains/dom.zig
```diff
@@ -60,17 +60,18 @@ pub fn processMessage(cmd: anytype) !void {
 
 // https://chromedevtools.github.io/devtools-protocol/tot/DOM/#method-getDocument
 fn getDocument(cmd: anytype) !void {
-    // const params = (try cmd.params(struct {
-    //     depth: ?u32 = null,
-    //     pierce: ?bool = null,
-    // })) orelse return error.InvalidParams;
+    const params = (try cmd.params(struct {
+        // CDP documentation implies that 0 isn't valid, but it _does_ work in Chrome
```
sjorsdonkers, line 64: I see, 0 means just this node, 1 means 1 layer of children

src/cdp/domains/dom.zig
karlseguin: I changed this. But I didn't add the `pierce` check (not here and not on `requestChildNodes`). The truth is that we only support the non-default `pierce == false`. I'm afraid to turn that check on for risk of breaking scripts (

PR #901, opened 2025-07-17
Rework URL.stitch, handle ../ (for yahoo)

karlseguin:
Also handle ./ anywhere in the path.

### DIFF
--- a/src/url.zig
+++ b/src/url.zig
@@ -153,18 +153,11 @@ pub const URL = struct {
                 std.debug.assert(out[out_i - 1] == '/');
 
                 out_i -= 2;
-                while (out_i > 1) {
-                    const next = out_i - 1;
-                    if (out[next] == '/') {
-                        // <= to deal with the hack-ish protocol_end which will be
-                        // off-by-one between http and https
-                        if (out_i <= protocol_end) {
-                            return error.InvalidURL;
-                        }
-                        break;
-                    }
-                    out_i = next;
+                while (out_i > 1 and out[out_i - 1] != '/') {
+                    out_i -= 1; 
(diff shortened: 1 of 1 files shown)


src/url.zig
```diff
@@ -100,52 +100,79 @@ pub const URL = struct {
...
         }
 
-        const protocol_end: usize = blk: {
-            if (std.mem.indexOf(u8, base, "://")) |protocol_index| {
-                break :blk protocol_index + 3;
-            } else {
-                break :blk 0;
-            }
-        };
+        // Quick hack becauste domains have to be at least 3 characters.
```
sjorsdonkers:
```suggestion
        // Quick hack because domains have to be at least 3 characters.
```

src/url.zig
```diff
@@ -100,52 +100,79 @@ pub const URL = struct {
...
+                    if (out[next] == '/') {
+                        // <= to deal with the hack-ish protocol_end which will be
+                        // off-by-one between http and https
+                        if (out_i <= protocol_end) {
+                            return error.InvalidURL;
+                        }
+                        break;
+                    }
+                    out_i = next;
+                }
```
sjorsdonkers:
```suggestion
                while (out_i > 1 and out[out_i - 1] != '/') {
                    out_i -= 1; 
                }
                // <= to deal with the hack-ish protocol_end which will be off-by-one between http and https
                if (out_i <= protocol_end) return error.InvalidURL;
```
Basically I think the `while (out_i > 1)` loop can be a bit simpler, it is just finding the previous separator

sjorsdonkers (approved):

PR #908, opened 2025-07-21
Prevent double-execution of script tags.

karlseguin:
Depends on https://github.com/lightpanda-io/libdom/pull/31


sjorsdonkers (approved):
4 layers of indirection to access a boolean.
I feel we should make a plan to improve the current state otherwise it's going to be to hard to change everything later on.

PR #913, opened 2025-07-23
Check Dynamically Cached Module First

mookums:
This just reorders how the dynamic import works. It checks the module cache first and fixes a crash experienced when a module is compiled while it is in the `.kEvaluating` state.

### DIFF
--- a/src/runtime/js.zig
+++ b/src/runtime/js.zig
@@ -1580,38 +1580,54 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
                 const iso = self.isolate;
                 const ctx = self.v8_context;
 
-                const module_loader = self.module_loader;
-                const source = module_loader.func(module_loader.ptr, specifier) catch {
-                    const error_msg = v8.String.initUtf8(iso, "Failed to load module");
-                    _ = resolver.reject(ctx, error_msg.toValue());
-                    return;
-                } orelse {
-                    const error_msg = v8.String.initUtf8(iso, "Module source not available");
-                    _ = resolver.reject(ctx, error_msg.toValue());
-                    return;
-                };
+                if (self.module_cache.get(specifier)) |cached_module| {
+                    const new_module = cached_module.castToModule();
+                    const status = new_module.getStatus();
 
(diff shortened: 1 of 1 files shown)


src/runtime/js.zig
```diff
@@ -1580,98 +1580,104 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
...
-                defer try_catch.deinit();
+                if (self.module_cache.get(specifier)) |cached_module| {
+                    const new_module = cached_module.castToModule();
+                    const status = new_module.getStatus();
+
+                    switch (status) {
+                        .kEvaluated, .kEvaluating => {
+                            const namespace = new_module.getModuleNamespace();
+                            log.info(.js, "dynamic import complete", .{
+                                .module = new_module,
```
karlseguin: these two are opaque values, they print meaningless data.

src/runtime/js.zig
```diff
@@ -1580,98 +1580,104 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
...
+                                .namespace = namespace,
+                            });
+                            _ = resolver.resolve(ctx, namespace);
+                            return;
+                        },
+                        else => {
```
karlseguin: This is still in the `if (cached)` block. If the module is cached, why are we loading it again?

src/runtime/js.zig
```diff
@@ -1580,98 +1580,104 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
...
+                                log.err(.js, "module evaluation is promise", .{
+                                    .specifier = specifier,
+                                    .line = try_catch.sourceLineNumber() orelse 0,
+                                });
+                                const error_msg = v8.String.initUtf8(iso, "Evaluation is a promise");
+                                _ = resolver.reject(ctx, error_msg.toValue());
+                                return;
+                            };
+                        },
+                    }
```
karlseguin: Am I following the indentation, wrong, or is this not the end of the `if (cached)` block, shouldn't there be if not cached handling?

krichprollsch:
@mookums what's the status of this PR? Can you rebase it please?

PR #916, opened 2025-07-23
Handle Null Event Listener

mookums:
This supersedes #912. It fixes the same behavior but in a more proper way, by allowing and handling null event listeners that are expected to clear the callback.


karlseguin:
Could have saved yourself some typing and passed the optional to the `fn register` and done the check there in 1 place.

LGTM.

PR #919, opened 2025-07-31
Create HTMLElement instead of pure Element

krichprollsch:
Our `toInterface` expect the given Element is always an HTMLElement.
~So I think it's better to create always HTMLElement.~ No b/c we could want to create a non-HTML tag, eg. svg

~But I would also like to detect if an Element is really an HTMLElement to avoid crash on `toInterface`, but I can't figure out one correct solution...~ done using `_dom_element_get_tag_name`  :heavy_check_mark: 

depends on https://github.com/lightpanda-io/libdom/pull/32


karlseguin:
I think you could stop it from crashing by calling `_dom_element_get_tag_name` (which I don't think we have exposed in netsurf.zig) instead of `_dom_html_element_get_tag_type` ?? You could then maybe handle non-HTML tag names (`svg`, ...??).

krichprollsch:
So as you suggested, as used `_dom_element_get_tag_name` + `Tag.fromString` instead of `_dom_html_element_get_tag_type` to retrieve the element's Tag.
I think it makes more sense this way: Node.toInterface => Element.toInterface => HTMLElement.toInterfaceFromTag

krichprollsch:
I pushed more changed to remove `netsurf.elementHTMLGetTagType`

But i have `html/dom/documents/dom-tree-accessors/document.title-09.html` crashing again (on `c._dom_html_document_create_element_ns`).

so still in progress...

krichprollsch:
@karlseguin WPT tests are passing now.

krichprollsch:
Hum, Maybe I should keep `try Element.toInterface` signature unchanged and create a `toInterfaceT(comptime T: type, e: *parser.Element) !T {` instead to avoid all these changes :thinking:

krichprollsch:
ok here is a cleaner version

src/browser/netsurf.zig
```diff
@@ -328,6 +328,24 @@ pub const Tag = enum(u8) {
             else => upperName(@tagName(tag)),
         };
     }
+
+    pub fn fromString(tagname: []const u8) !Tag {
```
karlseguin, line 332: I wish I could find the code in netsurf that does it, I _think_ the node / element tags are always upper cased:

```
const n = document.createElement('div');
n.tagName === 'DIV'
```

src/browser/netsurf.zig
```diff
@@ -2139,13 +2156,32 @@ pub inline fn documentCreateDocument(title: ?[]const u8) !*DocumentHTML {
     return doc_html;
 }
 
+pub inline fn documentCreateHTMLElement(doc: *Document, tag_name: []const u8) !*Element {
```
karlseguin: I know everything in netsurf.zig is inlined, but Andrew's not a fan:

karlseguin: Is there a reason to keep the originals? If our document is _always_ an html document, then it will _always_ create an HTMLElement. Only reason I can see to keep both is if we think we'll have a non-html document??

src/browser/netsurf.zig
krichprollsch: Maybe this one?
https://github.com/lightpanda-io/libdom/blob/master/src/html/html_document.c#L563

But the enum's field name is lower case. I tried to uppercase it at comptime, but I fall in a `evaluation exceeded 1000 backwards branches` error :/

src/browser/netsurf.zig
krichprollsch: You can create non-html document from scratch in JS and then add elements to them.
If I try to use only HTML element creation, WPT test is crashing on https://github.com/lightpanda-io/wpt/blob/main/html/dom/documents/dom-tree-accessors/document.title-09.html.

It's b/c the document isn't an HTML one.

So depending the document, we want to create HTMLElement or Element...
I could maybe detect the document's type and use the best function...

krichprollsch: I'm afraid I can't find a way to detect if a document is an HTML one...
The constructor doesn't set any base doc attribute with a specific value.
https://github.com/lightpanda-io/libdom/blob/master/src/html/html_document.c#L93-L116

krichprollsch: ok I did with https://github.com/lightpanda-io/libdom/pull/32

karlseguin:
LGTM

PR #920, opened 2025-08-03
Fix: Properly handle node ownership when using appendChild and insertBefore

SrikanthKumarC:
Fixes a bug where using appendChild or insertBefore with a node from a different ownerDocument would throw a WrongDocument Error. According to the specification, the new node being added should be implicitly adopted by the parent's ownerDocument.

This change updates these methods to first set the ownerDocument of the node being added, and recursively update the ownerDocument of its children to match that of the parent.

---
Currently, the behavior is as follows: an error is thrown when a node is inserted using appendChild or insertBefore if its ownerDocument differs from that of the parent. This PR fixes that.

Debug log: 
```
js : function call error . . . . . . . . . . . . . . .  [0m[+0ms]
      name = browser.dom.node.Node._insertBefore
      err = WrongDocument
      args = 
        1: #<HTMLDivElement> (object)
        2: #<HTMLParagraphElement> (object)
      stack = 
        <anonymous>:28
```

Javascript used: 
```js
console.log("--- Starting Automatic Cross-Document Insertion ---");


const parser = new DOMParser();
const newDoc = parser.parseFromString('<div id="new-node"><p>Hey</p></div>', 'text/html');
console.log("A new document has been created in memory using DOMParser:", newDoc);


const newNode = newDoc.getElementById('new-node');

console.log("The new node's initial ownerDocument is:", newNode.ownerDocument);
console.log("The main document is:", document);


const parent = document.getElementById('target-container');
const referenceNode = document.getElementById('reference-node');


parent.insertBefore(newNode, referenceNode);

console.log("\n--- Insertion Complete ---");
console.log("The new node has been inserted into the main document.");
console.log("The new node's updated ownerDocument is:", newNode.ownerDocument);
console.log("Is the new node now a child of the main document's parent?", parent.contains(newNode));

const k = document.getElementById('new-node');
const ptag = k.querySelector('p');
console.log("The new node's p tag owner is: ", ptag.ownerDocument);
console.log("P tags owner is same as main document?", ptag?.ownerDocument === document);

const statusMessage = document.getElementById('status-message');
statusMessage.textContent = "Success! The node was inserted automatically on load. Check the console for details.";

```

### DIFF
--- a/src/browser/dom/node.zig
+++ b/src/browser/dom/node.zig
@@ -201,13 +201,17 @@ pub const Node = struct {
         const self_owner = try parser.nodeOwnerDocument(self);
         const child_owner = try parser.nodeOwnerDocument(child);
 
+        // If the node to be inserted has a different ownerDocument than the parent node, 
+        // modern browsers automatically adopt the node and its descendants into 
+        // the parent's ownerDocument. 
+        // This process is known as adoption.
+        // (7.1) https://dom.spec.whatwg.org/#concept-node-insert 
         if (child_owner == null or (child_owner.? != self_owner.?)) {
             const w = Walker{};
-            child.owner = self_owner;
             var current = child;
-            while (try w.get_next(child, current)) |current_node| {
-                current_node.owner = self_owner;
-                current = current_node;
+            while (true) {
+                current.owner = self_owner;
+                current = try w.get_next(child, current) orelse break;
             }
         }
 
@@ -306,13 +310,17 @@ pub const Node = struct {
         const self_owner = try parser.nodeOwnerDocument(self);
         const new_node_owner = try parser.nodeOwnerDocument(new_node);
 
+        // If the node to be inserted has a different ownerDocument than the parent node, 
+        // modern browsers automatically adopt the node and its descendants into 
+        // the parent's ownerDocument. 
+        // This process is known as adoption.
+        // (7.1) https://dom.spec.whatwg.org/#concept-node-insert 
         if (new_node_owner == null or (new_node_owner.? != self_owner.?)) {
             const w = Walker{};
-            new_node.owner = self_owner;
             var current = new_node;
-            while (try w.get_next(new_node, current)) |current_node| {
-                current_node.owner = self_owner;
-                current = current_node;
+            while(true) {
+                current.owner = self_owner;
+                current = try w.get_next(new_node, current) orelse break;
             }
         }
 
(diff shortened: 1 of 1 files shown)


SrikanthKumarC:

krichprollsch (commented):
Hello @SrikanthKumarC

Thank you very much for you PR :pray: 
 
I would like to avoid the usage of an arena for traversing children.
Can you consider using the walker instead?

```zig
const w = Walker{};
while (true) {
    next = try w.get_next(child, next) orelse break;
    next.owner = self_owner;
}
```

existing example:  https://github.com/lightpanda-io/browser/blob/main/src/browser/dom/node.zig#L224-L234

SrikanthKumarC:
Oh yeah that's better. Just refactored it to use the walker instead. thanks

src/browser/dom/node.zig
```diff
@@ -198,6 +198,19 @@ pub const Node = struct {
...
+        const child_owner = try parser.nodeOwnerDocument(child);
+
+        if (child_owner == null or (child_owner.? != self_owner.?)) {
+            const w = Walker{};
+            child.owner = self_owner;
+            var current = child;
+            while (try w.get_next(child, current)) |current_node| {
+                current_node.owner = self_owner;
+                current = current_node;
+            }
```
krichprollsch: what about:
```suggestion
            var current = child;
            while (true) {
                current.owner = self_owner;
                current = try w.get_next(child, next) orelse break;
            }
```

krichprollsch (commented):
Thanks for the change.

Here is a little non-blocking suggestion. 

It would be also great to:
* add a comment before both `if` statements to explain the intention of changing the owner
* write a JS test case at the end of the file, similar to https://github.com/lightpanda-io/browser/blob/main/src/browser/polyfill/webcomponents.zig#L39-L65 You could use `Browser.DOM.node.owner` as test name. You can run it specifically with `T=Browser.DOM.node.owner make test`

SrikanthKumarC:
Thanks for the suggestion. I’ve added the tests and simplified the walker as you recommended. Sorry for the trouble, I'm still getting the hang of Zig

krichprollsch (approved):

krichprollsch:
No worries, thank you very much for your contribution :pray: 
I think a will create a little additional PR to avoid getting the owner twice when calling `__insertBefore` with a null `ref_node_`.

PR #930, opened 2025-08-07
request interception

sjorsdonkers:
Depends on: https://github.com/lightpanda-io/browser/pull/922
Also reenables: `http_request_start`

TODO:
- [ ] Enable: patterns
- [ ] continueWithAuth also Enable.handleAuthRequests
- [ ] fulfillRequest
- [ ] InterceptResponse, also in continueRequest
- [ ] continueRequest.postData
- [ ] continueRequest.headers
- [ ] Common headers and start_callback before intercept
- [ ] Confirm string lifetime of waiting Requests

Example playwright fragment:
```js

await page.route('**/*', (route, request) => {
  console.log(`Request ROUTE: ${request.method()} ${request.url()}`);
  route.continue({url: "http://lightpanda.io/", method: "GET"});// or route.abort();
});

await page.goto('/campfire-commerce/');
```

### DIFF
--- a/src/browser/browser.zig
+++ b/src/browser/browser.zig
@@ -52,6 +52,8 @@ pub const Browser = struct {
         errdefer env.deinit();
 
         const notification = try Notification.init(allocator, app.notification);
+        app.http.client.notification = notification;
+        app.http.client.next_request_id = 0; // Should we track ids in CDP only?
         errdefer notification.deinit();
 
         return .{
@@ -74,6 +76,7 @@ pub const Browser = struct {
         self.page_arena.deinit();
         self.session_arena.deinit();
         self.transfer_arena.deinit();
+        self.http_client.notification = null;
         self.notification.deinit();
         self.state_pool.deinit();
     }
--- a/src/browser/page.zig
+++ b/src/browser/page.zig
@@ -467,12 +467,15 @@ pub const Page = struct {
         const owned_url = try self.arena.dupeZ(u8, request_url);
         self.url = try URL.parse(owned_url, null);
 
+        var headers = try HttpClient.Headers.init();
+        if (opts.header) |hdr| try headers.add(hdr);
+
         self.http_client.request(.{
             .ctx = self,
             .url = owned_url,
             .method = opts.method,
+            .headers = headers,
             .body = opts.body,
-            .header = opts.header,
             .cookie = self.requestCookie(.{ .is_navigation = true }),
             .header_done_callback = pageHeaderDoneCallback,
             .data_callback = pageDataCallback,
--- a/src/browser/page.zig
+++ b/src/browser/page.zig
@@ -469,6 +469,7 @@ pub const Page = struct {
 
         var headers = try HttpClient.Headers.init();
         if (opts.header) |hdr| try headers.add(hdr);
+        try self.requestCookie(.{ .is_navigation = true }).headersForRequest(self.arena, owned_url, &headers);
 
         self.http_client.request(.{
             .ctx = self,
@@ -476,7 +477,7 @@ pub const Page = struct {
             .method = opts.method,
             .headers = headers,
             .body = opts.body,
-            .cookie = self.requestCookie(.{ .is_navigation = true }),
+            .cookie_jar = self.cookie_jar,
             .header_done_callback = pageHeaderDoneCallback,
             .data_callback = pageDataCallback,
             .done_callback = pageDoneCallback,
--- a/src/cdp/domains/fetch.zig
+++ b/src/cdp/domains/fetch.zig
@@ -1,4 +1,4 @@
-// Copyright (C) 2023-2024  Lightpanda (Selecy SAS)
+// Copyright (C) 2023-2025    Lightpanda (Selecy SAS)
 //
 // Francis Bouvier <francis@lightpanda.io>
 // Pierre Tachoire <pierre@lightpanda.io>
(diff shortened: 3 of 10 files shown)


src/cdp/domains/fetch.zig
```diff
@@ -17,13 +17,217 @@
...
+
+fn disable(cmd: anytype) !void {
+    const bc = cmd.browser_context orelse return error.BrowserContextNotLoaded;
+    bc.fetchDisable();
+    return cmd.sendResult(null, .{});
+}
+
+fn enable(cmd: anytype) !void {
+    const params = (try cmd.params(EnableParam)) orelse EnableParam{};
+    if (params.patterns.len != 0) std.debug.print("Fetch.enable: Not implemented patterns set\n", .{});
```
karlseguin: use logger

src/browser/browser.zig
```diff
@@ -52,6 +52,7 @@ pub const Browser = struct {
         errdefer env.deinit();
 
         const notification = try Notification.init(allocator, app.notification);
+        app.http.client.notification = notification;
```
karlseguin, line 55: Don't think `deinit` is ever called nor is it ever removed from the http client when the browser is removed.

src/cdp/domains/fetch.zig
```diff
@@ -17,13 +17,217 @@
...
+
+    const request_id = try id_from_request_id(params.requestId);
+    if (state.waiting.fetchSwapRemove(request_id) == null) return error.RequestNotFound;
+
+    log.info(.cdp, "Request aborted by intercept", .{params.errorReason});
+    return cmd.sendResult(null, .{});
+}
+
+// Get u64 from requestId which is formatted as: "INTERCEPT-{d}"
+fn id_from_request_id(request_id: []const u8) !u64 {
```
karlseguin:
```suggestion
fn idFromRequestId(request_id: []const u8) !u64 {
```

src/cdp/domains/fetch.zig
```diff
@@ -17,13 +17,217 @@
...
+    return std.fmt.parseInt(u64, request_id[10..], 10) catch |err| {
```
karlseguin: Worth doing a length check first

src/cdp/domains/fetch.zig
```diff
@@ -17,13 +17,217 @@
...
+    var state = &bc.cdp.intercept_state;
+    const params = (try cmd.params(struct {
+        requestId: []const u8, // "INTERCEPT-{d}"
+        errorReason: ErrorReason,
+    })) orelse return error.InvalidParams;
+
+    const request_id = try id_from_request_id(params.requestId);
+    if (state.waiting.fetchSwapRemove(request_id) == null) return error.RequestNotFound;
+
+    log.info(.cdp, "Request aborted by intercept", .{params.errorReason});
```
karlseguin:
```suggestion
    log.info(.cdp, "Request aborted by intercept", .{.reason = params.errorReason});
```

karlseguin (commented):
I know this is still in draft.

I like how little impact this has on the `http/*`.

Do you think it's possible to re-add the `http_request_fail` and `http_request_complete`? We could then merge this into `nonblocking_libcurl` and then into `main`, as these 2 (+`http_request_start`) are all that's missing.

sjorsdonkers:
> I like how little impact this has on the `http/*`.

My main concern, and what I got stuck on for while before ignoring it, is dealing with the start_callback. My understanding is that it can modify the headers and the body, but I'm not sure why it happens as a callback as opposed to when creating the Request.

> Do you think it's possible to re-add the `http_request_fail` and `http_request_complete`? We could then merge this into `nonblocking_libcurl` and then into `main`, as these 2 (+`http_request_start`) are all that's missing.

I'm OK with both adding them here or taking the `http_request_start` out and moving it into `nonblocking_libcurl`. We'll see whatever turns out more convenient.

karlseguin:
The only thing using the startCallback is XHR (ScriptManager is only using to log). I was going to say _we can remove it_..but...

XHR does two things with this:
1 - it sets the headers. This we could replace. The `Request` already allows a header to be passed.

2 - it gets a reference to the transfer, which it needs in case xhr.abort() is called. The only solution I can think for this is that `request()` would return some opaque identifier (probably the request id) and then we can cancel based on the request_id, instead of the transfer. Requires Client to maintain a list of id => transfer which is a bit unfortunate.

src/cdp/domains/fetch.zig
sjorsdonkers: [bac639a](https://github.com/lightpanda-io/browser/pull/930/commits/bac639a6b33e7b85620ea81b1ac4b6645183bd38)

src/cdp/domains/fetch.zig
sjorsdonkers: [bac639a](https://github.com/lightpanda-io/browser/pull/930/commits/bac639a6b33e7b85620ea81b1ac4b6645183bd38)

src/cdp/domains/fetch.zig
sjorsdonkers: [bac639a](https://github.com/lightpanda-io/browser/pull/930/commits/bac639a6b33e7b85620ea81b1ac4b6645183bd38)

src/browser/browser.zig
sjorsdonkers: it was deinitted, but not set to null: [bac639a](https://github.com/lightpanda-io/browser/pull/930/commits/bac639a6b33e7b85620ea81b1ac4b6645183bd38)

src/http/Http.zig
```diff
@@ -210,6 +210,59 @@ pub const Connection = struct {
...
+            const str = std.mem.span(@as([*:0]const u8, @ptrCast(node.*.data)));
+            const header = parseHeader(str) orelse return error.InvalidHeader;
+            list.putAssumeCapacity(header.name, header.value);
+            current = node.*.next;
+        }
+        return list;
+    }
+
+    fn parseHeader(header_str: []const u8) ?struct { name: []const u8, value: []const u8 } {
+        const colon_pos = std.mem.indexOf(u8, header_str, ":") orelse return null;
```
karlseguin, line 248: `indexOfScalar` removes 1 len check.

src/browser/page.zig
```diff
@@ -467,12 +467,15 @@ pub const Page = struct {
         const owned_url = try self.arena.dupeZ(u8, request_url);
         self.url = try URL.parse(owned_url, null);
 
+        var headers = try HttpClient.Headers.init();
```
karlseguin, line 470: A lot of internal failure cases will result in this leaking. But I think we can leave it as-is for now.

The difficulty in resource management here is that we need to be careful about allocations up to the point where `curl_multi_add_handle` is called. After that, it's handled by the `perform` loop. So a simple `errdefer` won't do, since you risk double-free depending on where the failure is. A simple fix is to not call` perform `in `makeRequest` - making `curl_multi_add_handle` the last-called function - but I feel like delaying `perform` until the next `tick` (especially since there's no reason to think the request can't be sent out immediately) is unnecessary latency.

I want to address it, but, at this point, after we merge everything.

PR #934, opened 2025-08-08
add a --with_base option to fetch

krichprollsch:


src/browser/dump.zig
```diff
@@ -91,6 +92,14 @@ pub fn writeNode(node: *parser.Node, opts: Opts, writer: anytype) anyerror!void
             // void elements can't have any content.
             if (try isVoid(parser.nodeToElement(node))) return;
 
+            // If we wrote the <header> and we want to include a <base>, add it
+            // now.
+            if (opts.include_base != null and tag_type == .head) {
```
karlseguin: I know `dump` isn't called often (although, `get_innerHTML` uses it), and this isn't a bottleneck, but....

did you consider appending a <base> element to the head before `dump` is called, so that you don't need the condition inside of here? It would just be a child of head.

src/browser/dump.zig
```diff
@@ -91,6 +92,14 @@ pub fn writeNode(node: *parser.Node, opts: Opts, writer: anytype) anyerror!void
             // void elements can't have any content.
             if (try isVoid(parser.nodeToElement(node))) return;
 
+            // If we wrote the <header> and we want to include a <base>, add it
```
karlseguin:
```suggestion
            // If we wrote the <head> and we want to include a <base>, add it
```

src/browser/dump.zig
krichprollsch: updated :heavy_check_mark:

PR #959, opened 2025-08-20
handle text content type with HTML

krichprollsch:
For text content type (and application/json) we create a pseudo HTML tree with the text value in a `<pre>` tag.

It allows CDP clients to interact with text content easily.

Relates with #824


src/browser/page.zig
```diff
@@ -207,6 +208,21 @@ pub const Page = struct {
                 return out.writeAll(buf.items);
             },
             .raw_done => |data| return out.writeAll(data),
+            .text => {
+                // processed below, along with .html
```
karlseguin: Commit is good, but don't think this comment is true.

krichprollsch: good catch, thanks

PR #960, opened 2025-08-20
WIP: auth required interception

krichprollsch:


src/cdp/domains/fetch.zig
```diff
@@ -276,6 +274,56 @@ fn continueRequest(cmd: anytype) !void {
     return cmd.sendResult(null, .{});
 }
 
+fn continueWithAuth(cmd: anytype) !void {
+    const bc = cmd.browser_context orelse return error.BrowserContextNotLoaded;
+    const params = (try cmd.params(struct {
+        requestId: []const u8, // "INTERCEPT-{d}"
+        authChallengeResponse: struct {
+            response: []const u8,
```
sjorsdonkers: Best to use an enum here for consistency/ validation:
`Default, CancelAuth, ProvideCredentials`

src/cdp/domains/fetch.zig
```diff
@@ -276,6 +274,56 @@ fn continueRequest(cmd: anytype) !void {
...
+            username: ?[]const u8 = null,
+            password: ?[]const u8 = null,
```
sjorsdonkers: Consider:
```suggestion
            username: []const u8 = "",
            password: []const u8 = "",
```

src/cdp/domains/fetch.zig
```diff
@@ -276,6 +274,56 @@ fn continueRequest(cmd: anytype) !void {
...
+        },
+    })) orelse return error.InvalidParams;
+
+    const page = bc.session.currentPage() orelse return error.PageNotLoaded;
+
+    var intercept_state = &bc.intercept_state;
+    const request_id = try idFromRequestId(params.requestId);
+    const transfer = intercept_state.remove(request_id) orelse return error.RequestNotFound;
```
sjorsdonkers, line 299: I think we should add a flag to make sure that it was a auth interception and not a normal intercept.
Otherwise the situation may occur that a user sends a continueWithAuth as a responce to a requestPaused.
In that case for example if we call abortAuthChallenge we do not do the correct cleanup. In general we do not know what could happen, so better to catch it.

src/http/Client.zig
```diff
@@ -365,6 +370,22 @@ fn perform(self: *Client, timeout_ms: c_int) !void {
...
+        if (transfer._auth_challenge != null) {
+            if (transfer.client.notification) |notification| {
+                var wait_for_interception = false;
+                notification.dispatch(.http_request_auth_required, &.{ .transfer = transfer, .wait_for_interception = &wait_for_interception });
+                if (wait_for_interception) {
+                    // the request is put on hold to be intercepted.
+                    // In this case we ignore callbacks for now.
+                    // Note: we don't deinit transfer on purpose: we want to keep
+                    // using it for the following request.
+                    self.endTransfer(transfer);
```
sjorsdonkers, line 383: I wonder whether we need to send a `transfer.resetCallback` as consumer may need to reset their state for the replaced intercept to be sent.
Or we can make it part of the contract that this is the case when `transfer.start_callback` is called for the same transfer, but we probably should document that perhaps near: `start_callback: ?*const fn (transfer: *Transfer) anyerror!void = null,` in Request.

| don't see issue with the current consumers implementation.

src/cdp/domains/fetch.zig
krichprollsch: but the client sends a string :thinking: Do we have a way to transform strings into enum on unserialisation?

sjorsdonkers: Yup, will just work :), see `ErrorReason` for example

src/cdp/domains/fetch.zig
krichprollsch: What about testing if the transfer has a auth challenge?
Or you do want a complete separate `intercept_state` list?

src/http/Client.zig
krichprollsch: Hum,  I suppose the point of the interception is to take place during the request lifecycle. So the following callbacks will be called depending the result of the interception and the optional request following.

src/cdp/domains/fetch.zig
sjorsdonkers: Checking and I guess then removing the auth challenge should work, then the opposite for continue/abort/fulfill

PR #963, opened 2025-08-21
Improves correctness of NodeIterator

karlseguin:
Minor improvement to correctness of TreeWalker.

Fun fact, this is the first time, that I've run into, where we have to default null and undefined to different values.

Also, tweaked the WPT test runner. WPT test results use | as a field delimiter. But a WPT test (and, I assume a message) can contain a |. So we had at least a few tests that were being reported as failed, only because the result line was weird / unexpected. No great robust way to parse this, but I changed it to look explicitly for |Pass or |Fail and use those positions as anchor points.

### DIFF
--- a/src/runtime/js.zig
+++ b/src/runtime/js.zig
@@ -946,7 +946,7 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
                             // a JsObject, not null. Consider a function,
                             //    _doSomething(arg: ?Env.JsObjet) void { ... }
                             //
-                            // And then these three calls:
+                            // And then these two calls:
                             //   doSomething();
                             //   doSomething(null);
                             //


src/runtime/js.zig
```diff
@@ -941,6 +941,27 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
             fn jsValueToZig(self: *JsContext, comptime named_function: NamedFunction, comptime T: type, js_value: v8.Value) !T {
                 switch (@typeInfo(T)) {
                     .optional => |o| {
+                        if (comptime isJsObject(o.child)) {
+                            // If type type is a ?JsObject, then we want to pass
+                            // a JsObject, not null. Consider a function,
+                            //    _doSomething(arg: ?Env.JsObjet) void { ... }
+                            //
+                            // And then these three calls:
```
```suggestion
                            // And then these two calls:
```

?

krichprollsch (approved):

PR #970, opened 2025-08-25
Remove the loop

karlseguin:
Previously, the IO loop was doing three things:
1 - Managing timeouts (either from scripts or for our own needs) 
2 - Handling browser IO events (page/script/xhr)
3 - Handling CDP events (accept, read, write, timeout)

With the libcurl merge, 1 was moved to an in-process scheduler and 2 was moved to libcurl's own event loop. That means the entire loop code, including the dependency on tigerbeetle-io existed for handling a single TCP client. Not only is that a lot of code, there was also friction between the two loops (the libcurl one and our IO loop), which would result in latency - while one loop is waiting for the events, any events on the other loop go un-processed.

This PR removes our IO loop. To accomplish this:

1 - The main accept loop is blocking. This is simpler and works perfectly well, given we only allow 1 active connection.
2 - The client socket is passed to libcurl - yes, libcurl's loop can take arbitrary FDs and poll them along with its own.

In addition to having one less dependency, the CDP code is quite a bit simpler, especially around shutdowns and writes. This also removes _some_ of the latency caused by the friction between page process and CDP processing. Specifically, when CDP now blocks for input, http page events (script loading, xhr, ...) will still be processed.

There's still friction. For one, the reverse isn't true: when the page is waiting for events, CDP events aren't going to be processed. But the page.wait already have some sensitivity to this (e.g. the page.request_intercepted flag). Also, when CDP waits, while we will process network events, page timeouts are still not processed. Because of both these remaining issues, we still need to jump between the two loops - but being able to block on CDP (even for a short time) WITHOUT stopping the page's network I/O, should reduce some latency.

### DIFF
--- a/src/server.zig
+++ b/src/server.zig
@@ -72,12 +72,9 @@ pub const Server = struct {
         self.listener = listener;
 
         try posix.setsockopt(listener, posix.SOL.SOCKET, posix.SO.REUSEADDR, &std.mem.toBytes(@as(c_int, 1)));
-        // TODO: Broken on darwin
-        // https://github.com/ziglang/zig/issues/17260  (fixed in Zig 0.14)
-        // if (@hasDecl(os.TCP, "NODELAY")) {
-        //  try os.setsockopt(socket.sockfd.?, os.IPPROTO.TCP, os.TCP.NODELAY, &std.mem.toBytes(@as(c_int, 1)));
-        // }
-        try posix.setsockopt(listener, posix.IPPROTO.TCP, 1, &std.mem.toBytes(@as(c_int, 1)));
+        if (@hasDecl(posix.TCP, "NODELAY")) {
+            try posix.setsockopt(listener, posix.IPPROTO.TCP, posix.TCP.NODELAY, &std.mem.toBytes(@as(c_int, 1)));
+        }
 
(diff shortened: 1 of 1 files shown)


src/server.zig
```diff
@@ -46,114 +38,143 @@ const MAX_HTTP_REQUEST_SIZE = 4096;
...
+        const flags = posix.SOCK.STREAM | posix.SOCK.CLOEXEC;
+        const listener = try posix.socket(address.any.family, flags, posix.IPPROTO.TCP);
+        self.listener = listener;
+
+        try posix.setsockopt(listener, posix.SOL.SOCKET, posix.SO.REUSEADDR, &std.mem.toBytes(@as(c_int, 1)));
+        // TODO: Broken on darwin
+        // https://github.com/ziglang/zig/issues/17260  (fixed in Zig 0.14)
+        // if (@hasDecl(os.TCP, "NODELAY")) {
+        //  try os.setsockopt(socket.sockfd.?, os.IPPROTO.TCP, os.TCP.NODELAY, &std.mem.toBytes(@as(c_int, 1)));
+        // }
```
sjorsdonkers: Perhaps a good time to re-enable this?

krichprollsch (approved):
:tada:

PR #972, opened 2025-08-25
Fetch + ReadableStream

mookums:
This implements `fetch` in Zig, allowing us to also gain access to the `ReadableStream` API.

### DIFF
--- a/src/browser/fetch/fetch.zig
+++ b/src/browser/fetch/fetch.zig
@@ -101,7 +101,7 @@ pub fn fetch(input: RequestInput, options: ?RequestInit, page: *Page) !Env.Promi
 
     try page.requestCookie(.{}).headersForRequest(arena, req.url, &headers);
 
-    const resolver = page.main_context.createPersistentPromiseResolver();
+    const resolver = try page.main_context.createPersistentPromiseResolver(.page);
 
     const fetch_ctx = try arena.create(FetchContext);
     fetch_ctx.* = .{
@@ -170,7 +170,6 @@ pub fn fetch(input: RequestInput, options: ?RequestInit, page: *Page) !Env.Promi
         .done_callback = struct {
             fn doneCallback(ctx: *anyopaque) !void {
                 const self: *FetchContext = @ptrCast(@alignCast(ctx));
-                defer self.promise_resolver.deinit();
                 self.transfer = null;
 
                 log.info(.fetch, "request complete", .{
@@ -187,7 +186,6 @@ pub fn fetch(input: RequestInput, options: ?RequestInit, page: *Page) !Env.Promi
         .error_callback = struct {
             fn errorCallback(ctx: *anyopaque, err: anyerror) void {
                 const self: *FetchContext = @ptrCast(@alignCast(ctx));
-                defer self.promise_resolver.deinit();
                 self.transfer = null;
 
                 log.err(.fetch, "error", .{
--- a/src/browser/streams/ReadableStreamDefaultReader.zig
+++ b/src/browser/streams/ReadableStreamDefaultReader.zig
@@ -56,7 +56,7 @@ pub fn _read(self: *const ReadableStreamDefaultReader, page: *Page) !Env.Promise
                 if (self.stream.reader_resolver) |rr| {
                     return rr.promise();
                 } else {
-                    const persistent_resolver = page.main_context.createPersistentPromiseResolver();
+                    const persistent_resolver = try page.main_context.createPersistentPromiseResolver(.page);
                     self.stream.reader_resolver = persistent_resolver;
                     return persistent_resolver.promise();
                 }
--- a/src/browser/streams/ReadableStream.zig
+++ b/src/browser/streams/ReadableStream.zig
@@ -78,11 +78,15 @@ const QueueingStrategy = struct {
 pub fn constructor(underlying: ?UnderlyingSource, _strategy: ?QueueingStrategy, page: *Page) !*ReadableStream {
     const strategy: QueueingStrategy = _strategy orelse .{};
 
-    const cancel_resolver = page.main_context.createPersistentPromiseResolver();
-    const closed_resolver = page.main_context.createPersistentPromiseResolver();
+    const cancel_resolver = try page.main_context.createPersistentPromiseResolver(.self);
+    const closed_resolver = try page.main_context.createPersistentPromiseResolver(.self);
 
     const stream = try page.arena.create(ReadableStream);
-    stream.* = ReadableStream{ .cancel_resolver = cancel_resolver, .closed_resolver = closed_resolver, .strategy = strategy };
+    stream.* = ReadableStream{
+        .cancel_resolver = cancel_resolver,
+        .closed_resolver = closed_resolver,
+        .strategy = strategy,
+    };
 
     const controller = ReadableStreamDefaultController{ .stream = stream };
 
@@ -108,10 +112,7 @@ pub fn constructor(underlying: ?UnderlyingSource, _strategy: ?QueueingStrategy, 
 pub fn destructor(self: *ReadableStream) void {
     self.cancel_resolver.deinit();
     self.closed_resolver.deinit();
-
-    if (self.reader_resolver) |*rr| {
-        rr.deinit();
-    }
+    // reader resolver is scoped to the page lifetime and is cleaned up by it.
 }
 
 pub fn get_locked(self: *const ReadableStream) bool {
--- a/src/browser/streams/ReadableStreamDefaultController.zig
+++ b/src/browser/streams/ReadableStreamDefaultController.zig
@@ -40,7 +40,6 @@ pub fn _close(self: *ReadableStreamDefaultController, _reason: ?[]const u8, page
 
     // Resolve the Reader Promise
     if (self.stream.reader_resolver) |*rr| {
-        defer rr.deinit();
         try rr.resolve(ReadableStreamReadResult{ .value = .empty, .done = true });
         self.stream.reader_resolver = null;
     }
@@ -62,7 +61,6 @@ pub fn _enqueue(self: *ReadableStreamDefaultController, chunk: []const u8, page:
     const duped_chunk = try page.arena.dupe(u8, chunk);
 
     if (self.stream.reader_resolver) |*rr| {
-        defer rr.deinit();
         try rr.resolve(ReadableStreamReadResult{ .value = .{ .data = duped_chunk }, .done = false });
         self.stream.reader_resolver = null;
     }
@@ -75,7 +73,6 @@ pub fn _error(self: *ReadableStreamDefaultController, err: Env.JsObject) !void {
     self.stream.state = .{ .errored = err };
 
     if (self.stream.reader_resolver) |*rr| {
-        defer rr.deinit();
         try rr.reject(err);
         self.stream.reader_resolver = null;
     }
--- a/src/runtime/js.zig
+++ b/src/runtime/js.zig
@@ -677,6 +677,11 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
             // we now simply persist every time persist() is called.
             js_object_list: std.ArrayListUnmanaged(PersistentObject) = .empty,
 
+            // Various web APIs depend on having a persistent promise resolver. They
+            // require for this PromiseResolver to be valid for a lifetime longer than
+            // the function that resolves/rejects them.
+            persisted_promise_resolvers: std.ArrayListUnmanaged(v8.Persistent(v8.PromiseResolver)) = .empty,
+
             // When we need to load a resource (i.e. an external script), we call
             // this function to get the source. This is always a reference to the
             // Page's fetchModuleSource, but we use a function pointer
@@ -730,6 +735,10 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
                 }
 
                 for (self.js_object_list.items) |*p| {
+                    p.deinit();
+                }
+
+                for (self.persisted_promise_resolvers.items) |*p| {
                     p.deinit();
                 }
 
@@ -1261,11 +1270,20 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
                 };
             }
 
-            pub fn createPersistentPromiseResolver(self: *JsContext) PersistentPromiseResolver {
-                return .{
-                    .js_context = self,
-                    .resolver = v8.Persistent(v8.PromiseResolver).init(self.isolate, v8.PromiseResolver.init(self.v8_context)),
-                };
+            // creates a PersistentPromiseResolver, taking in a lifetime parameter.
+            // If the lifetime is page, the page will clean up the PersistentPromiseResolver.
+            // If the lifetime is self, you will be expected to deinitalize the PersistentPromiseResolver.
+            pub fn createPersistentPromiseResolver(
+                self: *JsContext,
+                lifetime: enum { self, page },
+            ) !PersistentPromiseResolver {
+                const resolver = v8.Persistent(v8.PromiseResolver).init(self.isolate, v8.PromiseResolver.init(self.v8_context));
+
+                if (lifetime == .page) {
+                    try self.persisted_promise_resolvers.append(self.context_arena, resolver);
+                }
+
+                return .{ .js_context = self, .resolver = resolver };
             }
 
             // Probing is part of trying to map a JS value to a Zig union. There's


src/browser/fetch/Headers.zig
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    return .{
+        .headers = headers,
+    };
+}
+
+pub fn _get(self: *const Headers, header: []const u8, page: *Page) !?[]const u8 {
+    const arena = page.arena;
+    const key = try std.ascii.allocLowerString(arena, header);
```
karlseguin: can use the `call_arena` here, as you don't need the key to outlive this function call.

src/browser/fetch/Headers.zig
```diff
@@ -0,0 +1,88 @@
...
+
+    const value = (self.headers.getEntry(key) orelse return null).value_ptr.*;
+    return try arena.dupe(u8, value);
```
karlseguin: You don't need to dupe this value at all, the string will get copied by v8. you can `return self.headers.get(key)` directly.

src/browser/streams/ReadableStreamDefaultReader.zig
```diff
@@ -0,0 +1,100 @@
...
+pub fn _cancel(self: *ReadableStreamDefaultReader) Env.Promise {
+    return self.stream._cancel();
+}
+
+pub const ReadableStreamReadResult = struct {
+    value: ?[]const u8,
+    done: bool,
+
+    pub fn get_value(self: *const ReadableStreamReadResult, page: *Page) !?[]const u8 {
+        return if (self.value) |value| try page.arena.dupe(u8, value) else null;
```
karlseguin: Probably don't need to dupe this:

```zig
return self.value;
```

src/browser/streams/ReadableStreamDefaultReader.zig
```diff
@@ -0,0 +1,100 @@
...
+
+    const resolver = Env.PromiseResolver{
+        .js_context = page.main_context,
+        .resolver = v8.PromiseResolver.init(page.main_context.v8_context),
+    };
+
+    switch (stream.state) {
+        .readable => {
+            if (stream.queue.items.len > 0) {
+                const data = self.stream.queue.orderedRemove(0);
```
karlseguin, line 49: This is O(N), right?  I have no clue what these streams are used for, so we know how big the queues (a) typically are and (b) can grow to?

src/browser/streams/ReadableStreamDefaultReader.zig
```diff
@@ -0,0 +1,100 @@
...
+                try resolver.resolve(ReadableStreamReadResult{ .value = data, .done = false });
+            } else {
+                // TODO: need to wait until we have more data
+                try resolver.reject("TODO!");
+                return error.Todo;
+            }
+        },
+        .closed => |_| {
+            if (stream.queue.items.len > 0) {
+                const data = try page.arena.dupe(u8, self.stream.queue.orderedRemove(0));
```
karlseguin: If the value is dupe'd using the page.arena when being inserted, I don't think you'll ever need to dupe it again (certainly not with the same arena).

krichprollsch:
I tried to fetch http://127.0.0.1:1234/campfire-commerce/ and I get a `response.json is not a function` error.
We will have to implement https://developer.mozilla.org/en-US/docs/Web/API/Response/json

src/browser/fetch/Headers.zig
```diff
@@ -0,0 +1,188 @@
...
+const Page = @import("../page.zig").Page;
+
+// https://developer.mozilla.org/en-US/docs/Web/API/Headers
+const Headers = @This();
+
+// Case-Insensitive String HashMap.
+// This allows us to avoid having to allocate lowercase keys all the time.
+const HeaderHashMap = std.HashMapUnmanaged([]const u8, []const u8, struct {
+    pub fn hash(_: @This(), s: []const u8) u64 {
+        var hasher = std.hash.Wyhash.init(s.len);
```
karlseguin, line 37: If you batch this, it'll be faster. I benchmarked it and saw a 2x improvement:
```
var buf: [64]u8 = undefined;
var hasher = std.hash.Wyhash.init(s.len);

var key = s;  
while (key.len >= 64) {
  const lower = std.ascii.lowerString(buf[0..], key[0..64]);
  hasher.update(lower);
  key = key[64..];
}

if (key.len > 0) {
  const lower = std.ascii.lowerString(buf[0..key.len], key);
  hasher.update(lower);
}
return hasher.final();
```

src/browser/fetch/Headers.zig
```diff
@@ -0,0 +1,188 @@
...
+        for (s) |c| {
+            hasher.update(&.{std.ascii.toLower(c)});
+        }
+
+        return hasher.final();
+    }
+    pub fn eql(_: @This(), a: []const u8, b: []const u8) bool {
+        if (a.len != b.len) return false;
```
karlseguin: `return std.ascii.eqlIgnoreCase(a, b);`

src/browser/fetch/Headers.zig
```diff
@@ -0,0 +1,188 @@
...
+                    const key = try page.arena.dupe(u8, pair[0]);
+                    const value = try page.arena.dupe(u8, pair[1]);
+
+                    try headers.put(arena, key, value);
+                }
+            },
+            .headers => |hdrs| {
+                var iter = hdrs.headers.iterator();
+                while (iter.next()) |entry| {
+                    const key = try page.arena.dupe(u8, entry.key_ptr.*);
```
karlseguin: I'm not sure these need to be duped. If they come from an existing `Headers`, then they're already owned by the page arena...I think.

src/browser/fetch/Headers.zig
```diff
@@ -0,0 +1,188 @@
...
+pub fn clone(self: *Headers, allocator: std.mem.Allocator) !Headers {
+    return Headers{
+        .headers = try self.headers.clone(allocator),
+    };
+}
+
+pub fn _append(self: *Headers, name: []const u8, value: []const u8, page: *Page) !void {
+    const arena = page.arena;
+
+    if (self.headers.getEntry(name)) |entry| {
```
karlseguin: `getOrPut` lets you do this with a single hash lookup. As-is, if the entry isn't found, it takes 2.

src/browser/fetch/Headers.zig
```diff
@@ -0,0 +1,188 @@
...
+pub fn _has(self: *const Headers, name: []const u8) bool {
+    return self.headers.contains(name);
+}
+
+// TODO: keys iterator
+
+pub fn _set(self: *Headers, name: []const u8, value: []const u8, page: *Page) !void {
+    const arena = page.arena;
+
+    if (self.headers.getEntry(name)) |entry| {
```
karlseguin: Same, `getOrPut` will let you do this in a single hash lookup.

src/browser/streams/ReadableStreamDefaultReader.zig
mookums: Yes it is. Will circle back on this, mainly was waiting for us to go to `0.15.1` because a lot of the FIFO stuff was removed. Would be nice if we had our own nice Queue impl (or maybe I just use the PriorityQueue and always return `.eq` :shrug:)

krichprollsch:
I'm testing the branch and comparing results w/ `main`.
It works quite well :clap: 

I found a difference fetching https://fr.pinterest.com/
I have the following error on this branch I don't have with main:
```
INFO  js : function call error . . . . . . . . . . . . . . .  [+1095ms]
      name = browser.html.window.Window._fetch
      err = InvalidArgument
      args =
        1: https://fr.pinterest.com/_/_/storage_report/ (string)
        2: {"method":"POST","headers":{"Content-Type":"application/json"},"body":"{\"storage-report\":{\"get\":[\"theme\"],\"set\":[],\"remove\":[]}}"} (object)
      stack =
        fetch:1
        <anonymous>:1
        d:18
        <anonymous>:18
        <anonymous>:18
        d:1
        a:1
        <anonymous>:1
        Promise:1
        <anonymous>:1
        v:1
        <anonymous>:1
        r:1
        <anonymous>:1
        d:18
        <anonymous>:18
        <anonymous>:18
        d:1
        a:1
        <anonymous>:1
        Promise:1
        <anonymous>:1
        b:1
        w:1
        value:1
        p:1
        237193:1
        n:1
        264788:1
        n:1
```

src/browser/fetch/Request.zig
```diff
@@ -0,0 +1,298 @@
...
+        .resolver = v8.PromiseResolver.init(page.main_context.v8_context),
+    };
+
+    const p = std.json.parseFromSliceLeaky(
+        std.json.Value,
+        page.arena,
+        self.body,
+        .{},
+    ) catch |e| {
+        log.warn(.browser, "invalid json", .{ .err = e, .source = "Request" });
```
krichprollsch: You should create and use a `.fetch` domain for all fetch's logs

krichprollsch:
Same with https://jobs.lightpanda.io/

```
INFO  js : function call error . . . . . . . . . . . . . . .  [+799ms]
      name = browser.html.window.Window._fetch
      err = InvalidArgument
      args =
        1: /api/v3/getStatsigResults (string)
        2: {"method":"POST","credentials":"same-origin","headers":{"notion-client-version":"23.13.0.4890","notion-audit-log-platform":"web","x-notion-active-user-header":"","Content-Type":"application/json"},"body":"{\"deviceId\":\"9928347c-20cc-4366-aa1f-c00d3080f870
\",\"device\":{\"clientVersion\":\"23.13.0.4890\",\"isElectron\":false,\"isMobileNative\":false,\"isMobileBeta\":false,\"isMobileBrowser\":false,\"isBrowser\":true,\"isMobile\":false,\"isDesktopBrowser\":true,\"isTablet\":false,\"os\":\"unknown\",\"browserName\":\"und
efined\"},\"stableID\":\"d9ec7133-17de-4ba0-a206-c8ad22c49127\",\"locale\":\"en-US\",\"shouldStringifyInitialValues\":true,\"clientSdkApiKey\":\"client-Tgza5wNFa8dVt9BdeUfG6Vkm29bHxX10MhoztTMzLBB\"}"} (object)
      stack =
        m:54
        <anonymous>:54
```

mookums:
Both of those resolved with https://github.com/lightpanda-io/browser/pull/972/commits/4413f68a4b0f20837baa0acc87ea73ea6539be23.

src/browser/fetch/Headers.zig
```diff
@@ -0,0 +1,267 @@
...
+                    };
+                    const value_string = try value.toString(arena);
+
+                    try headers.put(arena, name, value_string);
+                }
+            },
+            .headers => |hdrs| {
+                var iter = hdrs.headers.iterator();
+                while (iter.next()) |entry| {
+                    const key = try arena.dupe(u8, entry.key_ptr.*);
```
karlseguin: Are you sure these two dupes are necessary?

src/browser/fetch/Headers.zig
```diff
@@ -0,0 +1,267 @@
...
+                    const value = try arena.dupe(u8, pair[1]);
+
+                    try headers.put(arena, key, value);
+                }
+            },
+            .mappings => |obj| {
+                var iter = obj.nameIterator();
+                while (try iter.next()) |name_value| {
+                    const name = try name_value.toString(arena);
+                    const value = Env.Value{
```
karlseguin: Is this right? I feel like you're just re-recreating the value that represents the name? Pretty sure this is storing name => name.

See how the URL constructor does it.  Could add an object.iterator() that emits a key, value pair too, since we now have 2 places that do this.

src/browser/fetch/Headers.zig
```diff
@@ -0,0 +1,267 @@
...
+            return null;
+        }
+    }
+};
+
+pub fn _entries(self: *const Headers, page: *Page) !HeaderEntryIterator {
+    const iter = try page.arena.create(HeaderHashMap.Iterator);
+    iter.* = self.headers.iterator();
+
+    return .{ .slot = try page.arena.alloc([]const u8, 2), .iter = iter };
```
karlseguin: Is there any difference between doing this allocation, and defining the field as `slot: [2][]const u8` ?

src/browser/fetch/Headers.zig
```diff
@@ -0,0 +1,267 @@
...
+}
+
+pub fn _forEach(self: *Headers, callback_fn: Env.Function, this_arg: ?Env.JsObject) !void {
+    var iter = self.headers.iterator();
+
+    if (this_arg) |this| {
```
karlseguin: I think this will give you the same without having to repeat the loop

```zig
const cb = if (this_arg) |this| callback_fn.withThis(this) else callback_fn;
while (iter.next()) |entry| {
    try cb.call(
        void,
        .{ entry.key_ptr.*, entry.value_ptr.*, self },
    );
}

File: src/browser/fetch/Headers.zig [hunk 16]
```diff
@@ -0,0 +1,267 @@
...
+            self.slot[1] = entry.value_ptr.*;
+            return self.slot;
+        } else {
+            return null;
+        }
+    }
+};
+
+pub fn _entries(self: *const Headers, page: *Page) !HeaderEntryIterator {
+    const iter = try page.arena.create(HeaderHashMap.Iterator);
```
karlseguin: Does this really have to be on the heap?

src/browser/fetch/Headers.zig
```diff
@@ -0,0 +1,267 @@
...
+        if (self.iter.next()) |key| {
+            return key.*;
+        } else {
+            return null;
+        }
+    }
+};
+
+pub fn _keys(self: *const Headers, page: *Page) !HeaderKeyIterator {
+    const iter = try page.arena.create(HeaderHashMap.KeyIterator);
```
karlseguin: Does this really have to be on the heap?

src/browser/fetch/Headers.zig
```diff
@@ -0,0 +1,267 @@
...
+};
+
+pub fn _values(self: *const Headers, page: *Page) !HeaderValueIterator {
+    const iter = try page.arena.create(HeaderHashMap.ValueIterator);
+    iter.* = self.headers.valueIterator();
+    return .{ .iter = iter };
+}
+
+const testing = @import("../../testing.zig");
+test "fetch: headers" {
```
karlseguin: If you rebase, it would be nice if to use the new htmlRunner. Plenty of examples in there now, though I haven't converted them all.

src/browser/fetch/Headers.zig
```diff
@@ -0,0 +1,267 @@
...
+    _ = self.headers.remove(name);
+}
+
+pub const HeaderEntryIterator = struct {
+    slot: [][]const u8,
+    iter: *HeaderHashMap.Iterator,
+
+    // TODO: these SHOULD be in lexigraphical order but I'm not sure how actually
+    // important that is.
+    pub fn _next(self: *HeaderEntryIterator) !?[]const []const u8 {
```
karlseguin: Don't think this can fail.

src/browser/fetch/Headers.zig
```diff
@@ -0,0 +1,267 @@
...
+    const arena = page.arena;
+
+    const gop = try self.headers.getOrPut(arena, name);
+    gop.value_ptr.* = try arena.dupe(u8, value);
+}
+
+pub const HeaderValueIterator = struct {
+    iter: *HeaderHashMap.ValueIterator,
+
+    pub fn _next(self: *HeaderValueIterator) !?[]const u8 {
```
karlseguin: Don't think this can fail.

src/browser/fetch/Request.zig
karlseguin: This should be a `debug` or `info`. It's almost certainly a valid JS error that we can't do anything about. warn+ should be for things we can action.

src/browser/fetch/Request.zig
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    const resolver = Env.PromiseResolver{
+        .js_context = page.main_context,
+        .resolver = v8.PromiseResolver.init(page.main_context.v8_context),
+    };
+
+    const p = std.json.parseFromSliceLeaky(
+        std.json.Value,
+        page.arena,
```
karlseguin: `call_arena` is safe here, I think. Seems like `p` is just being transferred to v8.

src/browser/fetch/Response.zig
```diff
@@ -0,0 +1,191 @@
...
+
+pub fn _clone(self: *const Response, page: *Page) !Response {
+    if (self.body_used) {
+        return error.TypeError;
+    }
+
+    const arena = page.arena;
+
+    return Response{
+        .body = try arena.dupe(u8, self.body),
```
karlseguin: Does anything really need to be duped?  I assume if we're here, the body, url, mime, etc are all already owned by the page arena. 

Duping might be necessary to protect them from unitentially mutation, but I don't think any of these can be mutated anyways.

mookums: Both `Request` and `Response` fields are read only so yeah. Added a note about it and I just have them return the same fields.

mookums:
Ready for review by whoever, just don't merge yet as I'm going to migrate the tests to the new htmlRunner.

krichprollsch:
We have a hug memory regression:
> Peak resident set size: 50120
https://github.com/lightpanda-io/browser/actions/runs/17554979612/job/49962587005?pr=972

mookums:
> We have a hug memory regression:
> 
> > Peak resident set size: 50120
> > https://github.com/lightpanda-io/browser/actions/runs/17554979612/job/49962587005?pr=972

Fixed I believe, was a Persistent `PromiseResolver` not being deinitalized.

src/browser/streams/ReadableStreamDefaultReader.zig
krichprollsch: @mookums Can you address this change now?

src/browser/fetch/fetch.zig
```diff
@@ -0,0 +1,231 @@
...
+    };
+
+    var headers = try Http.Headers.init();
+
+    // Copy our headers into the HTTP headers.
+    var header_iter = req.headers.headers.iterator();
+    while (header_iter.next()) |entry| {
+        // This is fine because curl/headers copies it internally.
+        const combined = try std.fmt.allocPrintSentinel(
+            page.call_arena,
```
karlseguin: I _think_ this might have to be `page.arena`, because the http client has a queue, so the headers might not be passed off to libcurl within this function call.

src/browser/fetch/fetch.zig
```diff
@@ -0,0 +1,236 @@
...
+
+// https://developer.mozilla.org/en-US/docs/Web/API/Window/fetch
+pub fn fetch(input: RequestInput, options: ?RequestInit, page: *Page) !Env.Promise {
+    const arena = page.arena;
+
+    const req = try Request.constructor(input, options, page);
+
+    const resolver = Env.PromiseResolver{
+        .js_context = page.main_context,
+        .resolver = v8.PromiseResolver.init(page.main_context.v8_context),
```
karlseguin: No other part of the WebAPI that needs to dig this deep into the JS implementation. The fetch code does this 10 times. Surely there's a way to clean this up and make it easier and less v8-specific to create promises

src/browser/fetch/fetch.zig
```diff
@@ -0,0 +1,236 @@
...
+        .promise_resolver = v8.Persistent(v8.PromiseResolver).init(
+            page.main_context.isolate,
+            resolver.resolver,
+        ),
+        .method = req.method,
+        .url = req.url,
+    };
+
+    // Add destructor callback for FetchContext.
+    try page.main_context.destructor_callbacks.append(arena, Env.DestructorCallback.init(fetch_ctx));
```
karlseguin: This isn't ideal. Really reaching into an implementation details. Do we really need to reject the promise on page shutdown?

The other issue, that we talked about, is that if you don't cancel the transfer, then when the page does it, the errorCallback is called, and at that point, the context is aborted. But I think we can make this safer. (Although, I notice now that the errorCallback doesn't seem to be doing anything)

src/browser/fetch/fetch.zig
```diff
@@ -0,0 +1,236 @@
...
+                    .js_context = self.js_ctx,
+                    .resolver = self.promise_resolver.castToPromiseResolver(),
+                };
+
+                try promise_resolver.resolve(response);
+            }
+        }.doneCallback,
+        .error_callback = struct {
+            fn errorCallback(ctx: *anyopaque, err: anyerror) void {
+                const self: *FetchContext = @ptrCast(@alignCast(ctx));
```
karlseguin, line 188: Shouldn't this reject the promise?

src/browser/fetch/fetch.zig
mookums: Changed this so it only rejects the promise if we have an error that isn't `error.Abort` which is what we throw when the page is closed (I think). This allows us to just ditch using the destructor callbacks entirely.

src/browser/fetch/fetch.zig
mookums: I changed it so all instances of this pattern now use the `Env.PersistentPromiseResolver` which should make it less v8 specific to use these.

krichprollsch (approved):

PR #973, opened 2025-08-25
Handle response without body

sjorsdonkers:


src/browser/page.zig
```diff
@@ -695,6 +695,18 @@ pub const Page = struct {
         self.clearTransferArena();
 
         switch (self.mode) {
+            .pre => {
+                // Received a response without a body like: https://httpbin.io/status/200
+                // We assume we have received an OK status (checked in Client.headerCallback)
+                // so we load a blank document to navigate away from any prior page.
+                self.mode = .{ .parsed = {} };
+
+                var fbs = std.io.fixedBufferStream("");
```
karlseguin: There's a `parser.documentHTMLParseFromStr($STR)` that does this for you - wraps $STR in a fixedBufferStream

src/browser/page.zig
```diff
@@ -695,6 +695,18 @@ pub const Page = struct {
         self.clearTransferArena();
 
         switch (self.mode) {
+            .pre => {
```
karlseguin, line 698: just in case you missed it, build fails because .pre case is already defined as the last case of the switch.

src/browser/page.zig
sjorsdonkers: changed: [e2d47e1](https://github.com/lightpanda-io/browser/pull/973/commits/e2d47e1c8690a7fce77d6ba14e6990766953161a)

src/browser/page.zig
sjorsdonkers: Ah didn't expect the merge conflict: [e2d47e1](https://github.com/lightpanda-io/browser/pull/973/commits/e2d47e1c8690a7fce77d6ba14e6990766953161a)

PR #974, opened 2025-08-26
Removes the log for unknown script tags

karlseguin:
Some sites have a lot of text/template or application/json, and it just adds noise to the logs.

### DIFF
--- a/src/browser/ScriptManager.zig
+++ b/src/browser/ScriptManager.zig
@@ -136,14 +136,10 @@ pub fn addFromElement(self: *ScriptManager, element: *parser.Element) !void {
         if (std.ascii.eqlIgnoreCase(script_type, "module")) {
             break :blk .module;
         }
-        if (std.ascii.eqlIgnoreCase(script_type, "application/json")) {
-            return;
-        }
-        if (std.ascii.eqlIgnoreCase(script_type, "application/ld+json")) {
-            return;
-        }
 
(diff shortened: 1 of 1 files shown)


src/browser/ScriptManager.zig
```diff
@@ -136,14 +136,10 @@ pub fn addFromElement(self: *ScriptManager, element: *parser.Element) !void {
         if (std.ascii.eqlIgnoreCase(script_type, "module")) {
             break :blk .module;
         }
-        if (std.ascii.eqlIgnoreCase(script_type, "application/json")) {
-            return;
-        }
-        if (std.ascii.eqlIgnoreCase(script_type, "application/ld+json")) {
-            return;
-        }
```
krichprollsch, line 144: I'm not sure I understand why you remove these kind?
The won't create more log warning then?

krichprollsch: oh you remove the warning too :facepalm: my bad.
What about using debug instead?

krichprollsch (approved):

src/browser/ScriptManager.zig
karlseguin: Thought about it. On the fence. 

On the one hand I thought: good to log in case there's a type we don't handle that we should (which is the intention behind the original code). On the other hand, some sites have a lot of text/template, and I felt the noise outweighed the maybe-one-day-it-will-be-useful

PR #978, opened 2025-08-27
Make the CDP read buffer heap allocated & dynamic

karlseguin:
Rather than stack-allocating MAX_MESSAGE_SIZE upfront, we now allocate 32KB and grow the buffer as needed for larger messages, up to MAX_MESSAGE_SIZE.

This will reduce memory usage for drivers that don't send huge payloads (like playwright does).

While not implemented, this would also enable us to set the MAX_MESSAGE_SIZE at runtime (e.g. via a command line option).

### DIFF
--- a/src/server.zig
+++ b/src/server.zig
@@ -36,7 +36,7 @@ const MAX_HTTP_REQUEST_SIZE = 4096;
 // max message size
 // +14 for max websocket payload overhead
 // +140 for the max control packet that might be interleaved in a message
-const MAX_MESSAGE_SIZE = 512 * 1024 + 14;
+const MAX_MESSAGE_SIZE = 512 * 1024 + 14 + 140;
 
 pub const Server = struct {
     app: *App,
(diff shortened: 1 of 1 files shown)


sjorsdonkers (approved):
LGTM, any specific reason for why the number 32 was chosen?

karlseguin:
32KB for the default buffer? No. Any suggestions?

krichprollsch:
I think we can be more conservative and use only 16kb to start.
the max we have if I ignore playwright with our e2e test is ~10kb

src/server.zig
```diff
@@ -613,9 +625,17 @@ fn Reader(comptime EXPECT_MASK: bool) type {
                     }
                 } else if (message_len > MAX_MESSAGE_SIZE) {
                     return error.TooLarge;
-                }
-
-                if (buf.len < message_len) {
+                } else if (message_len > self.buf.len) {
+                    const new_buf = try self.allocator.alloc(u8, message_len);
```
krichprollsch: I'm wondering if we should take a bit more than the msg len.
In our e2e playwright script I can see a resize from 284314 to 284315...
So taking directly something like 25% more is maybe useful. WDYT?

krichprollsch: Moreover if we start with a small buffer

krichprollsch: And a debug log on buffer resize could be helpful

PR #983, opened 2025-08-28
exit the browser on SIGINT signal

krichprollsch:
In the docker container, the browser doesn't exit on `CRTL-C` (`SIGINT`).
Catching the signal and exiting fixes the issue.

## Reproduce the issue
 You can reproduce the issue with the existing version of the docker image

```
$ docker run -ti --rm --name lpd  lightpanda/browser:nightly

# the container starts the browser with the serve command.
# try CTRL-C: the container doesn't stop.
# use docker docker kill lpd to kill the container. 
```

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -41,6 +41,21 @@ pub fn main() !void {
         if (gpa.detectLeaks()) std.posix.exit(1);
     };
 
+    // Exit the program on SIGINT signal. When running the browser in a Docker
+    // container, sending a CTRL-C (SIGINT) signal is catched but doesn't exit
+    // the program. Here we force exiting on SIGINT.
+    std.posix.sigaction(std.posix.SIG.INT, &std.posix.Sigaction{
+        .handler = .{
+            .handler = struct {
+                pub fn handler(_: c_int) callconv(.c) void {
+                    std.posix.exit(0);
+                }
+            }.handler,
+        },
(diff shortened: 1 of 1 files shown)


src/main.zig
```diff
@@ -41,6 +41,21 @@ pub fn main() !void {
         if (gpa.detectLeaks()) std.posix.exit(1);
     };
 
+    // Exit the program on SIGINT signal. When running the browser in a Docker
+    // container, sending a CTRL-C (SIGINT) signal is catched but doesn't exit
+    // the program. Here we force exiting on SIGINT.
+    std.posix.sigaction(std.posix.SIG.INT, &std.posix.Sigaction{
+        .handler = .{
+            .handler = struct {
+                pub fn handler(_: c_int) callconv(.C) void {
```
karlseguin: `.c` for Zig 0.15 compatibility.

src/main.zig
```diff
@@ -41,6 +41,21 @@ pub fn main() !void {
...
+                    std.posix.exit(0);
```
karlseguin, line 70: If it's worth it, we can create a global ?App and ?Server, and try to shutdown them cleanly.

The one argument that I can make for going through the trouble is that, on `deinit`, the app flushes the telemetry queue. Can also do it in a separate PR.

krichprollsch: good point, it would be nice, I can try that.

PR #992, opened 2025-08-31
Pre-size the destination buffer when we know the response content length

karlseguin:


src/browser/xhr/xhr.zig
```diff
@@ -438,6 +438,13 @@ pub const XMLHttpRequest = struct {
 
         self.state = .loading;
         self.dispatchEvt("readystatechange");
+
+        if (transfer.getContentLength()) |cl| {
+            if (cl > 100 * 1024 * 1024) {
```
krichprollsch: Shouldn't we use a common const for this limit accross xhr, scriptmanager and future fetch?

src/http/Client.zig
```diff
@@ -1040,6 +1040,16 @@ pub const Transfer = struct {
 
         try req.done_callback(req.ctx);
     }
+
+    pub fn getContentLength(self: *const Transfer) ?u32 {
+        // It's possible for this to be null even with correct code, due to
+        // request fulfillment. If transfer.fulfill is called, we won't have
+        // a handle.
+        const handle = self._handle orelse return null;
```
krichprollsch: It works correctly into the header callback but this function will work only if the request is not fulfill and is not done. Maybe we should a comment to specify this function is useful only during this callback.

karlseguin: I added a comment and I made it work on fufilled requests.

src/browser/xhr/xhr.zig
karlseguin: I decided to remove these checks. It's easier to enforce this at a higher level (e.g. in systemd or some container). As I did it, the check would only apply to scripts and XHR with content-length. It wouldn't apply to the page itself, or any script/xhr without a content-length. If we _did_ decide to do this, it should probably be configurable anyways.

PR #995, opened 2025-09-01
Looking for feedback on new test runner

karlseguin:
Follows up on https://github.com/lightpanda-io/browser/pull/994 and replaces
the jsRunner with a new page.navigation-based test runner.

Currently only implemented for the Window tests, looking for feedback and
converting every existing test will take time - so for a while, newRunner (to be
renamed) will sit side-by-side with jsRunner.

In addition to the benefits outlined in 994, largely around code simplicity and
putting more of the actual code under tests, I think our WebAPI tests
particularly benefit from:
1 - No need to recompile when modifying the html tests
2 - Much better assertions, e.g. you can assert that something is actually an
    array, not just a string representation of an array
3 - Ability to test some edge cases (e.g. dynamic script loading)

I've put some effort into testing.js to make sure that, if the encapsulating
zig test passes, it's because it actually passed, not because it didn't run.

For the time being, console tests are removed. I think it's more useful to have
access to the console within tests, than it is to test the console (which is
just a wrapper around log, which is both tested and heavily used).

### DIFF
--- a/src/browser/tests/window/window.html
+++ b/src/browser/tests/window/window.html
@@ -1,3 +1,4 @@
+<script src="../testing.js"></script>
 <body></body>
 <script id=aliases>
   testing.expectEqual(window, window.self);
--- a/src/browser/tests/window/frames.html
+++ b/src/browser/tests/window/frames.html
@@ -1,3 +1,5 @@
+<script src="../testing.js"></script>
+
 <body>
  <iframe src="https://httpbin.io/html" title="iframea"></iframe>
  <iframe src="https://httpbin.io/html" title="iframeb"></iframe>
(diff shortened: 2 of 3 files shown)


src/browser/tests/window/window.html
```diff
@@ -0,0 +1,104 @@
...
+    atob('b');
+  });
+</script>
+
+<script id=scroll>
+  let scroll = false;
+  let scrolend = false
+  window.addEventListener('scroll', () => {scroll = true});
+  document.addEventListener('scrollend', () => {scrollend = true});
+  window.scrollTo(0);
```
sjorsdonkers, line 86: When this HTML is put into chrome we get:
Image
as with 1 parameter it is expecting to pass in Option instead of coordinates.

```suggestion
  window.scrollTo(0, 0);
```

PR #1007, opened 2025-09-04
Limit serve timeout to 1 week

karlseguin:


src/main.zig
```diff
@@ -137,7 +137,9 @@ fn run(alloc: Allocator) !void {
             const server = &_server.?;
             defer server.deinit();
 
-            server.run(address, opts.timeout * 1000) catch |err| {
+            // max timeout of 1 week.
+            const timeout = if (opts.timeout > 604_800) 604_800_000 else @as(i32, @intCast(opts.timeout)) * 1000;
```
krichprollsch: We should to either:
* add in the help the indication we take a max of 1 week timeout
* error the command if the value is bigger than 1 week.

krichprollsch: Isn't `@as(i32, opts.timeout)` enough? why do we need `@intCast`?

karlseguin: I expanded on the --help text. I removed the @intCast

PR #1008, opened 2025-09-04
Emit networkIdle and networkAlmostIdle Page.lifecycleEvent

karlseguin:
Most CDP drivers have a mechanism to wait for idle network, or an almost idle network (sometimes called networkIdle2). These are events the browser must emit.

The page will now emit `networkIdle` when we are reasonably sure there's no more network activity (this requires some slight changes to request interception, since, I believe, intercepted requests should be considered).

`networkAlmostIdle` is currently _always_ emitted prior to emitting `networkIdle`. We should tweak this but I can't, at a glance, think of a great heuristic for when this should be emitted.

### DIFF
--- a/src/browser/page.zig
+++ b/src/browser/page.zig
@@ -336,7 +336,7 @@ pub const Page = struct {
 
                     const http_active = http_client.active;
                     if (http_active == 0 and exit_when_done) {
-                        // we don't need to concider http_client.intercepted here
+                        // we don't need to consider http_client.intercepted here
                         // because exit_when_done is true, and that can only be
                         // the case when interception isn't possible.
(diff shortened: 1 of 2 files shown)


src/browser/page.zig
```diff
@@ -329,7 +334,13 @@ pub const Page = struct {
                         return error.JsError;
                     }
 
-                    if (http_client.active == 0 and exit_when_done) {
+                    const http_active = http_client.active;
+                    if (http_active == 0 and exit_when_done) {
+                        // we don't need to concider http_client.intercepted here
```
krichprollsch:
```suggestion
                        // we don't need to consider http_client.intercepted here
```

krichprollsch (approved):

PR #1012, opened 2025-09-05
Don't assume that page events means the BrowserContext has a page

karlseguin:
CDP currently assumes that if we get a page-related notification (like a request interception, or page lifecycle event), then we must have a session and page.

But, Target.detachFromTarget can remove the session from the BrowserContext while still having the page run (I wonder if we should stop the page at this point??). So, remove these assumptions and make sure we have a page/session in the handling of page events.


krichprollsch:
> I wonder if we should stop the page at this point?

I think we could detach and re-attach immediately to the same target.
So keeping the page is a good thing.

I agree to return if session_id is missing, It think it's the good way to do.

krichprollsch (approved):

PR #1019, opened 2025-09-05
jsValueToZig for fixed sized arrays

sjorsdonkers:
- Only accepts the value if the Js array size matches exactly

Note: tests failing due to fetch branch issue

### DIFF
--- a/src/runtime/js.zig
+++ b/src/runtime/js.zig
@@ -331,17 +331,15 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
 
         fn promiseRejectCallback(v8_msg: v8.C_PromiseRejectMessage) callconv(.c) void {
             const msg = v8.PromiseRejectMessage.initFromC(v8_msg);
-            const isolate =  msg.getPromise().toObject().getIsolate();
+            const isolate = msg.getPromise().toObject().getIsolate();
             const v8_context = isolate.getCurrentContext();
             const context: *JsContext = @ptrFromInt(v8_context.getEmbedderData(1).castTo(v8.BigInt).getUint64());
 
             const value =
-                if (msg.getValue()) |v8_value| valueToString(context.call_arena, v8_value, isolate, v8_context) catch |err| @errorName(err)
-                else "no value";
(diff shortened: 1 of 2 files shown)


src/runtime/js.zig
```diff
@@ -1110,6 +1108,19 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
...
+                        // Retrieve fixed-size array as slice then copy it
+                        const slice_type = []arr.child;
+                        const slice_value = try self.jsValueToZig(named_function, slice_type, js_value);
+                        if (slice_value.len != arr.len) {
+                            // Exact length match, we could allow smaller arrays, but we would not be able to communicate how many were written
+                            return error.InvalidArgument;
+                        }
+
+                        var result: T = undefined;
+                        @memcpy(&result, slice_value[0..arr.len]);
```
karlseguin: I think you can avoid this copy. I'm thinking:

```zig
return @as(*T, @ptrCast(slice_value.ptr)).*;
```

sjorsdonkers: Fixed: [1a064b3](https://github.com/lightpanda-io/browser/pull/1019/commits/1a064b39c3a870b7143c5d349388451d45378897)

src/runtime/js.zig
```diff
@@ -1393,6 +1404,36 @@ pub fn Env(comptime State: type, comptime WebApis: type) type {
                         },
                         else => {},
                     },
+                    .array => |arr| {
+                        // Retrieve fixed-size array as slice then probe
+                        const slice_type = []arr.child;
+                        switch (try self.probeJsValueToZig(named_function, slice_type, js_value)) {
+                            .value => |slice_value| {
+                                if (slice_value.len == arr.len) {
+                                    return .{ .ok = {} };
```
karlseguin: Should consider returning a `.{.value =  ...}` here. If probe returns a value, that means there was a cost to creating it that it doesn't want you to have to repeat. In this case, the probe allocated + populated the slice). If you turn that `.value` into an `.ok`, it'll have to do that same allocation + copy again.

sjorsdonkers: Fixed: [6631292](https://github.com/lightpanda-io/browser/pull/1019/commits/6631292b1ec872d777b11ae4dedd4f5f4580c920)
.value currently only used for 0 length, but code may change.

PR #1033, opened 2025-09-09
Support for CustomEvent in document.createEvent

nikneym:
Also adds `CustomEvent.initCustomEvent` function. Marks `netsurf.strFromData` as public in order to implement it, let me know if it should be done in some other way.

### DIFF
--- a/src/browser/events/custom_event.zig
+++ b/src/browser/events/custom_event.zig
@@ -64,7 +64,7 @@ pub const CustomEvent = struct {
         event_type: []const u8,
         can_bubble: bool,
         cancelable: bool,
-        detail: ?JsObject,
+        maybe_detail: ?JsObject,
     ) !void {
         // This function can only be called after the constructor has called.
         // So we assume proto is initialized already by constructor.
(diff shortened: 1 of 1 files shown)


src/browser/events/custom_event.zig
```diff
@@ -55,6 +56,25 @@ pub const CustomEvent = struct {
...
+        detail: ?JsObject,
+    ) !void {
+        // This function can only be called after the constructor has called.
+        // So we assume proto is initialized already by constructor.
+        self.proto.type = try netsurf.strFromData(event_type);
+        self.proto.bubble = can_bubble;
+        self.proto.cancelable = cancelable;
+        self.proto.is_initialised = true;
+        // Detail is stored separately.
+        self.detail = detail;
```
karlseguin: When v8 calls a Zig function, the data it passes is all owned/maintained by v8. When we convert these to Zig types in js.zig, we create copies - either implicitly because the value exists on the stack (bools/ints/floats, ...) or because we allocate memory on the heap.

For JsObject, the object is just passed as-is, so it continues to be owned/managed by v8. V8 doesn't know that we're referencing it, so it could free it at any point. V8 has the concept of a Global and Persistent which essentially add external reference counting. You need to "persist" this object so V8 won't free it from under you:

```
self.detail = detail.persist();
```

This does 2 things: 
- is informs v8 that there's an external reference to the object, so v8 won't free it
- it adds it to our list of persisted objects, so that when the JsContext is freed, we can "release" the persistent object which allows v8 to free it.

karlseguin (commented):
Looks good.

nikneym:

nikneym:
recheck

PR #1040, opened 2025-09-11
Increase event timeStamp resolution

karlseguin:
Depends on https://github.com/lightpanda-io/libdom/pull/36

The spec says this should be a High Definition timestamp. But browsers avoid that in order to avoid fingerprinting. By default, FireFox rounds to 2ms (which is what this PR does).

Previously, the timestamp was seconds, so you'd think: isn't that better? Well, it's pretty far off the spec and what browsers do, but more importantly, it crashes our WPT test. If you look at `Event-timestamp-safe-resolution.html` you'll see that it's trying to find the delta between two timestamps, in an endless loop (without a loop of many iterations). With second-resolution, it just takes too long (and crashes..memory).


karlseguin:
It's a bit funny, but the reason this started to crash WPT is because [this PR](https://github.com/lightpanda-io/browser/pull/998) fixed the `event.timestamp` getter...there was a typo before: `fn get_timestamp` which was fixed to `fn get_timeStamp` (uppercase 'S')..which then unlocked the test in question.

PR #1041, opened 2025-09-11
KeyboardEvent support

nikneym:
This PR adds support for `KeyboardEvent`. It doesn't support deprecated `charCode`, `keyCode` and `which` options in it's constructor. I believe approach taken in `CustomEvent` may be preferred in order to support them.

### DIFF
--- a/src/browser/netsurf.zig
+++ b/src/browser/netsurf.zig
@@ -952,7 +952,7 @@ pub fn keyboardEventDestroy(evt: *KeyboardEvent) void {
     c._dom_keyboard_event_destroy(evt);
 }
 
-pub inline fn keyboardEventKeyIsSet(
+pub fn keyboardEventKeyIsSet(
     evt: *KeyboardEvent,
     comptime key: enum { ctrl, alt, shift, meta },
 ) bool {


src/browser/events/keyboard_event.zig
```diff
@@ -0,0 +1,160 @@
...
+            return (get_ctrlKey(self) or get_metaKey(self));
+        }
+
+        // TODO: Add support for "CapsLock", "ScrollLock".
+        return false;
+    }
+
+    // Getters.
+
+    pub fn get_altKey(self: *Self) bool {
```
karlseguin, line 124: These can probably all be `*const Self`. Might not change much in terms of how it gets compiled, but it does document the non-mutating behavior.

src/browser/netsurf.zig
```diff
@@ -388,7 +388,7 @@ pub const DOMError = error{
 
 const DOMException = c.dom_exception;
 
-fn DOMErr(except: DOMException) DOMError!void {
+pub fn DOMErr(except: DOMException) DOMError!void {
```
karlseguin: Don't think this is necessary? And if it was, I'd be suspicious. We don't leak this part of the libdom API anywhere else, why now?

src/browser/netsurf.zig
```diff
@@ -951,15 +952,44 @@ pub fn keyboardEventDestroy(evt: *KeyboardEvent) void {
     c._dom_keyboard_event_destroy(evt);
 }
 
-const KeyboardEventOpts = struct {
-    key: []const u8,
-    code: []const u8,
+pub inline fn keyboardEventKeyIsSet(
```
karlseguin: I know [virtually] everything in netsurf.zig is inlined, but I think it's a mistake. Your decision if you think it's a good idea. I tend to fallback to:

src/browser/netsurf.zig
```diff
@@ -951,15 +952,44 @@ pub fn keyboardEventDestroy(evt: *KeyboardEvent) void {
...
+) bool {
+    var is_set: bool = false;
+    const err = switch (key) {
+        .ctrl => c._dom_keyboard_event_get_ctrl_key(evt, &is_set),
+        .alt => c._dom_keyboard_event_get_alt_key(evt, &is_set),
+        .shift => c._dom_keyboard_event_get_shift_key(evt, &is_set),
+        .meta => c._dom_keyboard_event_get_meta_key(evt, &is_set),
+    };
+    // None of the earlier can fail.
+    std.debug.assert(err == c.DOM_NO_ERR);
```
karlseguin, line 967: I strongly approve of this when we know the netsurf library can't fail. There are a lot of cases where this is true, though it can be hard to figure out in some cases (where the libdom path is deep/complicated); we should do it where we're reasonably sure it's correct.

src/tests/events/keyboard.html
```diff
@@ -0,0 +1,87 @@
...
+  testing.expectEqual(true, event.getModifierState("AltGraph"));
+  testing.expectEqual(true, event.getModifierState("Control"));
+  testing.expectEqual(true, event.getModifierState("Shift"));
+  testing.expectEqual(true, event.getModifierState("Meta"));
+  testing.expectEqual(true, event.getModifierState("OS"));
+  testing.expectEqual(true, event.getModifierState("Accel"));
+</script>
+
+<script id=keyDownListener>
+event = new KeyboardEvent("keydown", { key: "z" });
```
karlseguin: Indent is inconsistent...some <script> are indented (which is the norm), some aren't.

src/browser/events/keyboard_event.zig
nikneym: I agree that it'd help while reading code but libdom functions expect mutable pointers (while doing no mutation) so we can either do this:
```zig
pub fn keyboardEventKeyIsSet(
    evt: *const KeyboardEvent,
    comptime key: enum { ctrl, alt, shift, meta },
) bool {
    var is_set: bool = false;
    const err = switch (key) {
        .ctrl => c._dom_keyboard_event_get_ctrl_key(@constCast(evt), &is_set),
        .alt => c._dom_keyboard_event_get_alt_key(@constCast(evt), &is_set),
        .shift => c._dom_keyboard_event_get_shift_key(@constCast(evt), &is_set),
        .meta => c._dom_keyboard_event_get_meta_key(@constCast(evt), &is_set),
    };
    // None of the earlier can fail.
    std.debug.assert(err == c.DOM_NO_ERR);

    return is_set;
}
```

Or maybe something like this:
```zig
pub fn get_altKey(self: *const Self) bool {
    return netsurf.keyboardEventKeyIsSet(@constCast(self), .alt);
}
```

I think either solution would be fine.

src/browser/netsurf.zig
nikneym: I've found couple of other places where `DOM_NO_ERR` is the only case being returned as you've claimed so I'll just do assertions if it makes sense. Here what it does is checking if a bit is set:
```c
/**
 * Get the ctrl key state
 *
 * \param evt  The Event object
 * \param key  Whether the Control key is pressed down
 * \return DOM_NO_ERR.
 */
dom_exception _dom_keyboard_event_get_ctrl_key(dom_keyboard_event *evt,
		bool *key)
{
	*key = ((evt->modifier_state & DOM_MOD_CTRL) != 0);

	return DOM_NO_ERR;
}
```

In fact we can just do this in Zig but a change in libdom affecting here is a possibility.

src/browser/events/keyboard_event.zig
karlseguin: I missed that these were going into libdom. Fine as-is.

Issue #1045, opened 2025-09-12
Handle ISO-8859-9 encoding while fetching HTTP response


nikneym:
You _may_ reproduce via https://github.com/lightpanda-io/demo/pull/69.

The HTML parsing failing if `Content-Type` is given `ISO-8859-9`.

```
* Request completely sent off
< HTTP/2 200
< content-type: text/html; charset=ISO-8859-9
< content-security-policy: object-src 'none';base-uri 'self';script-src 'nonce-Q6q-_mo0CXrGYWoedX1mGg' 'strict-dynamic' 'report-sample' 'unsafe-eval' 'unsafe-inline' https: http:;report-uri https://csp.withgoogle.com/csp/gws/xsrp
< accept-ch: Sec-CH-Prefers-Color-Scheme
< p3p: CP="This is not a P3P policy! See g.co/p3phelp for more info."
< content-encoding: gzip
< date: Fri, 12 Sep 2025 11:42:39 GMT
< server: gws
< x-xss-protection: 0
< x-frame-options: SAMEORIGIN
< expires: Fri, 12 Sep 2025 11:42:39 GMT
< cache-control: private
< set-cookie: NID=525=eMD6244YFmPpoAbacf5t97406d-9mOAmAqqe3F67zHCogH4vwWhGbMaJ7SWOFUR2KC1KXvh0RW_wRR76KPoATQvMG-Hb9nJD1Uo6D9PxSZlUxQQvQqMSsIQ5jdsq97695Xrcd1Vz6UmGdxqSgvKe9YuvbS5ewVWn81JipwQJGWUtf257DiB5v-awDJux_8JKs2h_B7kgSKU806AoKd4; expires=Sat, 14-Mar-2026 11:42:38 GMT; path=/; domain=.google.com; HttpOnly
< alt-svc: h3=":443"; ma=2592000,h3-29=":443"; ma=2592000
<
ERROR http : data_callback . . . . . . . . . . . . . . . . .  [+5598ms]
      err = BadEncoding
      req = GET https://www.google.com/search?ie=ISO-8859-1&hl=tr&source=hp&biw=&bih=&q=lightpanda&iflsig=AOw8s4IAAAAAaMQVPg2w3BwseditS2s1ZbjXM82_cZLX&gbv=2

* client returned ERROR on write of 2330 bytes
* Connection #1 to host www.google.com:443 left intact
ERROR http : navigate failed . . . . . . . . . . . . . . . .  [+5598ms]
[...]

PR #1048, opened 2025-09-15
Mime: charset identification changes

nikneym:
`Mime` is changed to store the received charset value and use it as-is. We used to match the charset value for pre-known cases. Seem to be solving #1045.

As a follow-up commit, we can add support for charset value validation as specified in [RFC 2978](https://datatracker.ietf.org/doc/rfc2978/).

### DIFF
--- a/src/browser/mime.zig
+++ b/src/browser/mime.zig
@@ -123,8 +123,8 @@ pub const Mime = struct {
 
                     const attribute_value = try parseCharset(value);
                     @memcpy(charset[0..attribute_value.len], attribute_value);
-                    // Fill the rest with zeroes.
-                    @memset(charset[attribute_value.len..], 0);
+                    // Null-terminate right after attribute value.
+                    charset[attribute_value.len] = 0;
                 },
             }
         }
--- a/src/browser/mime.zig
+++ b/src/browser/mime.zig
@@ -52,6 +52,95 @@ pub const Mime = struct {
         other: struct { type: []const u8, sub_type: []const u8 },
     };
 
+    /// Removes quotes of value if quotes are given.
+    ///
+    /// Currently we don't validate the charset.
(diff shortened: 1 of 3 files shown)


src/browser/mime.zig
```diff
@@ -52,6 +54,100 @@ pub const Mime = struct {
...
+        var offset: usize = 1;
+
+        // Charset values are not so large; 128-bit registers should be
+        // more than enough.
+        const vec_size = 16;
+        const Vec = @Vector(vec_size, u8);
+        const UInt = std.meta.Int(.unsigned, vec_size);
+        const block_size = @sizeOf(u64);
+
+        const charset = blk: {
```
karlseguin: If the value is quoted, is the point to escape special characters, and thus this should un-escape them? Which I don't think it is.

I think this is just finding the (non-matching) closing quote, and stripping it off. If that's all we want, surely the faster and simplest thing to do is just look at the last character of the value?

karlseguin: `try parseCharset("\"ut\\\"f-8\"")`  as a test, should return `ut"f-8`   but this returns `ut\`

nikneym: My idea was to find the earliest double quote after the first one since `boundary` attribute can also have its value inside double quotes. Though in the code, params are already splitted by `;` which would eliminate the need for double quote search... I've missed that part.

We might want to escape special characters too but I was unsure since the older approach doesn't do it.

karlseguin: The older approach didn't care because it was looking for specific values that don't have escape values.

karlseguin: It's fine if we don't escape it [until we have a use case that needs it]. But I have to imagine a simple reverse loop will be faster for the _very_ common case that `"` will be the last character.

PR #1054, opened 2025-09-16
iOS Build Support

carson-katri:
This adds support for building dependencies for iOS. It depends on https://github.com/lightpanda-io/zig-v8-fork/pull/96 for v8.

Lightpanda can be built for iOS with these commands:

```sh
make get-v8
OS=ios make build-v8

export OS=iphonesimulator
export ARCH=arm64
export SDK=$(xcrun --sdk $OS --show-sdk-path)
export CC=$(xcrun --sdk "$SDK" -f clang)
export CXX=$(xcrun --sdk "$SDK" -f clang++)
export CPP="$CC -E"
export AR="$(xcrun --sdk "$SDK" -f ar)"
export RANLIB="$(xcrun --sdk "$SDK" -f ranlib)"
export CFLAGS="-isysroot \"$SDK\" -arch $ARCH -mios-simulator-version-min=16.0"
export LDFLAGS="-isysroot \"$SDK\" -arch $ARCH -mios-simulator-version-min=16.0"
export MAKE="make CC=$CC AR=$AR RANLIB=$RANLIB"
export CMAKE="cmake -DCMAKE_SYSTEM_NAME=iOS -DCMAKE_OSX_ARCHITECTURES=$ARCH -DCMAKE_OSX_SYSROOT=\"$SDK\" -DCMAKE_OSX_DEPLOYMENT_TARGET=16.0"
export AUTOCONF_FLAGS="--disable-shared --host=arm-apple-darwin CC=\"$CC\" CXX=\"$CXX\" CPP=\"$CPP\" AR=\"$AR\" RANLIB=\"$RANLIB\" CFLAGS=\"$CFLAGS\" CPPFLAGS=\"$CFLAGS\" LDFLAGS=\"$LDFLAGS\""
export SKIP_EXAMPLES=1
make install

SDK=$(xcrun --sdk iphonesimulator --show-sdk-path) zig build -Dtarget=aarch64-ios-simulator -Dcpu=apple_m1 --release=safe
```

> [!NOTE]
> The lightpanda executable cannot be built for iOS. `build.zig` should be modified to produce a `liblightpanda.a` file that can be included in an xcframework.


carson-katri:

src/app.zig
```diff
@@ -96,6 +96,11 @@ fn getAndMakeAppDir(allocator: Allocator) ?[]const u8 {
     if (@import("builtin").is_test) {
         return allocator.dupe(u8, "/tmp") catch unreachable;
     }
+
+    if (@import("builtin").os.tag == .ios) {
+        return null; // getAppDataDir is not available on iOS
```
krichprollsch: Do we have an alternative dir to use on ios?

build.zig
```diff
@@ -494,7 +522,7 @@ fn buildMbedtls(b: *Build, m: *Build.Module) !void {
     mbedtls.addIncludePath(b.path(root ++ "include"));
     mbedtls.addIncludePath(b.path(root ++ "library"));
 
-    mbedtls.addCSourceFiles(.{ .flags = &.{}, .files = &.{
+    mbedtls.addCSourceFiles(.{ .flags = &.{"-Wno-nullability-completeness"}, .files = &.{
```
krichprollsch: I'm not sure about changing this flag for all build.
Maybe it could be conditional for ios only for now? Or is there another reason to use it?

src/app.zig
carson-katri: I inlined the macOS version.

build.zig
carson-katri: I adjusted it to only add that flag for iOS builds.

carson-katri:
I added a `zig build static-lib` step based on `lib.zig`. Right now this is empty, but it can include `extern` definitions. We're building out the API we need for liveview-native this way.

We're going to want this static lib on other platforms too, so I didn't restrict it to iOS.

build.zig
```diff
@@ -149,6 +149,25 @@ pub fn build(b: *Build) !void {
         const build_step = b.step("build-v8", "Build v8");
         build_step.dependOn(&build_v8.step);
     }
+
+    {
+        // static lib
+        // -------
+        const static_lib_module = b.addModule("lightpanda", .{
+            .root_source_file = b.path("src/lib.zig"),
```
krichprollsch, line 157: I have an error b/c this file is missing when I'm trying to run `zig build static-lib`

krichprollsch (commented):
Please can you rebase the PR to remove the `.DS_Store` file :pray: 

May I suggest you to configure a global `.gitgnore` file to avoid the risk of committing these kind of files. We don't want to add this into the project's gitignore because it relates to the dev workspace.

krichprollsch:
I created #1193 follow up this PR with rebase + ios nightly build

PR #1056, opened 2025-09-17
Convert more DOM_NO_ERR cases to assertions

karlseguin:
There is some risk to this change. The first is that I made a mistake. The other is that one of the APIs that doesn't currently return an error changes in the future.

### DIFF
--- a/src/browser/dom/attribute.zig
+++ b/src/browser/dom/attribute.zig
@@ -25,24 +25,24 @@ pub const Attr = struct {
     pub const prototype = *Node;
     pub const subtype = .node;
 
-    pub fn get_namespaceURI(self: *parser.Attribute) !?[]const u8 {
-        return try parser.nodeGetNamespace(parser.attributeToNode(self));
+    pub fn get_namespaceURI(self: *parser.Attribute) ?[]const u8 {
+        return parser.nodeGetNamespace(parser.attributeToNode(self));
     }
 
-    pub fn get_prefix(self: *parser.Attribute) !?[]const u8 {
-        return try parser.nodeGetPrefix(parser.attributeToNode(self));
+    pub fn get_prefix(self: *parser.Attribute) ?[]const u8 {
+        return parser.nodeGetPrefix(parser.attributeToNode(self));
     }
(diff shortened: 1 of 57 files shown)


src/browser/dom/attribute.zig
```diff
@@ -25,24 +25,24 @@ pub const Attr = struct {
...
+        return parser.nodeGetNamespace(parser.attributeToNode(self));
     }
 
-    pub fn get_prefix(self: *parser.Attribute) !?[]const u8 {
-        return try parser.nodeGetPrefix(parser.attributeToNode(self));
+    pub fn get_prefix(self: *parser.Attribute) ?[]const u8 {
+        return parser.nodeGetPrefix(parser.attributeToNode(self));
     }
 
     pub fn get_localName(self: *parser.Attribute) ![]const u8 {
```
krichprollsch, line 36: you don't remove the error return here on purpose? (same for the 2 following)

src/main.zig
```diff
@@ -22,12 +22,10 @@ const Allocator = std.mem.Allocator;
 
 const log = @import("log.zig");
 const App = @import("app.zig").App;
-const Http = @import("http/Http.zig");
 const Server = @import("server.zig").Server;
 const Browser = @import("browser/browser.zig").Browser;
 
 const build_config = @import("build_config");
-const parser = @import("browser/netsurf.zig");
```
krichprollsch, line 30: Do you use something auto to cleanup imports? or it's manual?

krichprollsch (approved):
2 non-blocking comments

src/main.zig
karlseguin: I used https://github.com/DonIsaac/zlint   I was hoping it would help me find error return types that didn't need to be, but it doesn't detect that I guess. And I couldn't help myself but also remove these imports when I saw it.

src/browser/dom/attribute.zig
karlseguin: The underlying netsurf function still returns an error, I just removed the redundant `try`.

PR #1063, opened 2025-09-18
Remove JSRunner

karlseguin:
It only had a few fetch tests still using it. But now everything is migrated to the htmlRunner


krichprollsch (approved):
:tada:

PR #1065, opened 2025-09-18
Add setImmediate, clearImmediate

nikneym:
Adds mentioned functions by wrapping relevant timeout functions.


nikneym:
I've broke the matchMedia while stashing things; sorry for that.

PR #1066, opened 2025-09-18
Relaxed `MessagePort.postMessage`

nikneym:
Removes error return in `postMessage`, we still missing the implementation for `transfer` & `options` but function seem to work for other cases.


krichprollsch:
What about checking if `transfer` is given to return error only in this case?
I think it's easier to have an error (+log) knowing we don't support something.

karlseguin:
We think this is for transfer an object to a different JS context. A frame (what I saw) or a web worker.

If the script does:

```js
port.postMessage(null, [object_to_transfer]);
// we should continue to execute code
```

I don't think it's worth stopping execution (at that point anyways) because the "transfer" to a frame/webworker failed.

PR #1071, opened 2025-09-19
Add `element.dir` getter & setter

nikneym:


karlseguin:
test?

src/tests/dom/element.html
```diff
@@ -263,3 +263,18 @@
   $('#to-remove').remove();
   testing.expectEqual(null, $('#to-remove'));
 </script>
+
+<script id=elementDir>
```
karlseguin, line 269: Probably the most important behavior is that you can get the `dir` from a parsed element. So above, change:

```html
<div id="content">
```

to

```
<div id="content" dir=ltr>
```

And then check that `content.dir == 'ltr' ` and `p1.dir == ''` (i assume empty is the correct default?). Both `content` and `p1` are alreayd in scope.

PR #1085, opened 2025-09-23
Refactor `mime.zig`

nikneym:
This PR refactors `Mime` and introduces different parsing logic (prefers integer comparisons where possible) while giving up readability a bit.

### DIFF
--- a/src/browser/mime.zig
+++ b/src/browser/mime.zig
@@ -19,16 +19,16 @@
 const std = @import("std");
 
 pub const Mime = struct {
+    /// MIME type.
     content_type: ContentType,
-    params: []const u8 = "",
-    // IANA defines max. charset value length as 40.
-    // We keep 41 for null-termination since HTML parser expects in this format.
+    /// IANA defines max. charset value length as 40.
+    /// We keep 41 for null-termination since HTML parser expects in this format.
     charset: [41]u8 = default_charset,
 
     /// String "UTF-8" continued by null characters.
     pub const default_charset = .{ 'U', 'T', 'F', '-', '8' } ++ .{0} ** 36;
 
-    /// Mime with unknown Content-Type, empty params and empty charset.
+    /// Mime with unknown Content-Type, empty params and default charset.
     pub const unknown = Mime{ .content_type = .{ .unknown = {} } };
 
     pub const ContentTypeEnum = enum {
@@ -53,6 +53,8 @@ pub const Mime = struct {
         other: struct { type: []const u8, sub_type: []const u8 },
     };
 
+    pub const ParseError = error{Invalid};
+
     /// Returns the null-terminated charset value.
     pub fn charsetString(mime: *const Mime) [:0]const u8 {
         return @ptrCast(&mime.charset);
@@ -63,10 +65,10 @@ pub const Mime = struct {
     /// Currently we don't validate the charset.
     /// See section 2.3 Naming Requirements:
     /// https://datatracker.ietf.org/doc/rfc2978/
-    fn parseCharset(value: []const u8) error{ CharsetTooBig, Invalid }![]const u8 {
+    fn parseCharset(value: []const u8) ParseError![]const u8 {
         // Cannot be larger than 40.
         // https://datatracker.ietf.org/doc/rfc2978/
-        if (value.len > 40) return error.CharsetTooBig;
+        if (value.len > 40) return error.Invalid;
 
         // If the first char is a quote, look for a pair.
         if (value[0] == '"') {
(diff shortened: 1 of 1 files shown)


src/browser/mime.zig
```diff
@@ -81,58 +83,250 @@ pub const Mime = struct {
...
-        if (input.len > 255) {
-            return error.TooBig;
+    /// Matches the first 2 characters of data with given characters.
+    inline fn match2(data: *const [2]u8, c0: u8, c1: u8) bool {
+        return @as(u16, @bitCast(data.*)) == @as(u16, @bitCast([_]u8{ c0, c1 }));
+    }
+
+    /// Matches the first 3 characters of data with given characters.
+    inline fn match3(data: *const [3]u8, c0: u8, c1: u8, c2: u8) bool {
+        return data[0] == c0 and data[1] == c1 and data[2] == c2;
```
karlseguin, line 93: Why not use a `u24` here? (genuine question)

src/browser/mime.zig
```diff
@@ -81,58 +83,250 @@ pub const Mime = struct {
...
+        const rem = std.mem.trimStart(u8, normalized[mime_end..], &.{ ' ', '\t', ';' });
+
+        var iterator = std.mem.splitScalar(u8, rem, ';');
+        while (iterator.next()) |attr| {
+            // Skip.
+            if (attr.len < 8) {
+                continue;
+            }
+
+            const charset_: u64 = @bitCast([_]u8{ 'c', 'h', 'a', 'r', 's', 'e', 't', '=' });
```
karlseguin, line 131: Dunno if it's valid, but go parses "charset = hello"  (space after charset)

src/browser/mime.zig
```diff
@@ -81,58 +83,250 @@ pub const Mime = struct {
...
+                    }
 
-        var it = std.mem.splitScalar(u8, params, ';');
-        while (it.next()) |attr| {
-            const i = std.mem.indexOfScalarPos(u8, attr, 0, '=') orelse return error.Invalid;
-            const name = trimLeft(attr[0..i]);
+                    return parseOther(normalized);
+                },
+                // Perfect cases.
+                text_xml => break :blk .{ .text_xml = {} },
```
karlseguin, line 197: will this match `text/xmlover9000!` ?

src/browser/mime.zig
```diff
@@ -81,58 +83,250 @@ pub const Mime = struct {
...
+                    if (rem.len < 7) {
+                        @branchHint(.unlikely);
+                        return parseOther(normalized);
+                    }
 
-            const value = trimRight(attr[i + 1 ..]);
-            if (value.len == 0) {
-                return error.Invalid;
-            }
+                    if (match4(rem[0..4], 'a', 's', 'c', 'r') and match3(rem[4..7], 'i', 'p', 't')) {
```
karlseguin, line 205: Same, would match `text/javascript_is_java_written_in_cursive` ?

src/browser/mime.zig
```diff
@@ -81,58 +83,250 @@ pub const Mime = struct {
...
+
+        var m = Mime{ .content_type = mime_type };
+        var iterator = std.mem.splitScalar(u8, rem, ';');
+        while (iterator.next()) |attr| {
+            // Skip.
+            if (attr.len < 8) {
+                continue;
+            }
+
+            const charset_: u64 = @bitCast([_]u8{ 'c', 'h', 'a', 'r', 's', 'e', 't', '=' });
```
karlseguin, line 309: Enough repeated logic here that I'd be tempted to try to DRY it if possible.

src/browser/mime.zig
nikneym: I thought the same but the format specified as:
```
  parameter       = parameter-name "=" parameter-value
  parameter-name  = token
  parameter-value = ( token / quoted-string )
  
  token          = 1*tchar

  tchar          = "!" / "#" / "$" / "%" / "&" / "'" / "*"
                 / "+" / "-" / "." / "^" / "_" / "`" / "|" / "~"
                 / DIGIT / ALPHA
                 ; any VCHAR, except delimiters
```
I've also tested it in the wild, web servers seem to comply.

src/browser/mime.zig
nikneym: It Fails since after `text/xml`, following bytes should contain `charset=` though `text/xmlcharset=utf-8` passes sadly...

src/browser/mime.zig
nikneym: Same as before, `text/javascriptcharset=utf-8` passes, will handle semicolon case.

src/browser/mime.zig
nikneym: I was unsure to prefer arbitrary bit widths, it can definitely suit here.

nikneym:
> I approve of changes for the sake of changes, but can you benchmark it to compare? At least we'll have a better idea of further changes like this is worth it.

I've benchmarked both implementations in a separate environment with large data, I think the performance gain compared to code complexity isn't worth it:

```
Benchmark 1: ./zig-out/bin/new
  Time (mean ± σ):       0.3 ms ±   0.0 ms    [User: 0.2 ms, System: 0.1 ms]
  Range (min … max):     0.2 ms …   0.4 ms    3929 runs

Benchmark 2: ./zig-out/bin/old
  Time (mean ± σ):       0.3 ms ±   0.0 ms    [User: 0.2 ms, System: 0.1 ms]
  Range (min … max):     0.2 ms …   0.5 ms    3605 runs

Summary
  './zig-out/bin/new' ran
    1.00 ± 0.13 times faster than './zig-out/bin/old'
```

Closing in favor of this.

PR #1086, opened 2025-09-24
Implement `History` WebAPI.

mookums:
This implements the entire `History` WebAPI.

There are also some modifications to our testing runner because History navigates the page and needs a second `page.wait()` to rerun and validate that the state persisted.


src/browser/html/History.zig
```diff
@@ -0,0 +1,220 @@
...
+
+fn _dispatchPopStateEvent(
+    state: ?[]const u8,
+    page: *Page,
+) !void {
+    var evt = try PopStateEvent.constructor("popstate", .{ .state = state });
+
+    _ = try parser.eventTargetDispatchEvent(
+        @as(*parser.EventTarget, @ptrCast(&page.window)),
+        @as(*parser.Event, @ptrCast(&evt)),
```
karlseguin: I don't think this is safe. But I did the same thing with MessageEvent. I feel like they should both be:
```suggestion
        &evt.proto
```

src/browser/html/History.zig
```diff
@@ -0,0 +1,220 @@
...
+    );
+}
+
+pub fn _pushState(self: *History, state: Env.JsObject, _: ?[]const u8, _url: ?[]const u8, page: *Page) !void {
+    const arena = page.session.arena;
+
+    const url = if (_url) |u| try arena.dupe(u8, u) else try arena.dupe(u8, page.url.raw);
+    const json = try state.toJson(page.session.arena);
```
karlseguin:
```suggestion
    const json = try state.toJson(arena);
```

karlseguin: Is the idea that we can't store it as a `Env.JsObject` because it has to surirve a JSContext? If so, probably worth a small comment. Took me a second to figure out why were were obj -> json , json -> obj.

karlseguin: If this can fail due to bad input, I'd move it ahead of the `url` duping. But if v8 can serialize any v8.Value, and this would only fail in an extreme case (OOM), then it doesn't matter.

src/browser/html/History.zig
```diff
@@ -0,0 +1,220 @@
...
+    if (index_s < 0 or index_s > self.stack.items.len - 1) {
+        return;
+    }
+
+    const index = @as(usize, @intCast(index_s));
+    const entry = self.stack.items[index];
+    self.current = index;
+
+    if (try page.isSameOrigin(entry.url)) {
+        if (entry.state) |state| {
```
karlseguin: If is unnecessary.

```
History.dispatchPopStateEvent(entry.state, page);
```

src/browser/html/History.zig
```diff
@@ -0,0 +1,220 @@
...
+
+    const url = if (_url) |u| try arena.dupe(u8, u) else try arena.dupe(u8, page.url.raw);
+    const json = try state.toJson(page.session.arena);
+    const entry = HistoryEntry{ .state = json, .url = url };
+    try self.stack.append(arena, entry);
+    self.current = self.stack.items.len - 1;
+}
+
+pub fn _replaceState(self: *History, state: Env.JsObject, _: ?[]const u8, _url: ?[]const u8, page: *Page) !void {
+    const arena = page.session.arena;
```
karlseguin, line 127: Thoughts on just storing the session arena in a field?

mookums: Not opposed to it but we already take `page: *Page` in a few places because we need to access things like the `page.url` so it feels like duplication.

PR #1087, opened 2025-09-24
fix pointer parameter into KeyboardEvent contructor

krichprollsch:


nikneym:
I was going to mention `MouseEvent` also! Let's wait for CI to become green.

PR #1092, opened 2025-09-25
Support `Element#insertAdjacentHTML`

nikneym:
This PR adds support for `insertAdjacentHTML`. It currently missing XML parsing and `TrustedHTML` support, yet sites seem to prefer HTML and string variant only.

### DIFF
--- a/src/browser/dom/element.zig
+++ b/src/browser/dom/element.zig
@@ -273,12 +273,22 @@ pub const Element = struct {
             if (std.mem.eql(u8, position, "beforebegin")) {
                 // The node must have a parent node in order to use this variant.
                 const parent = parser.nodeParentNode(self_node) orelse return error.NoModificationAllowed;
+                // Parent cannot be Document.
+                // Should have checks for document_fragment and document_type?
+                if (parser.nodeType(parent) == .document) {
+                    return error.NoModificationAllowed;
+                }
+
                 break :blk .{ parent, self_node };
             }
 
             if (std.mem.eql(u8, position, "afterend")) {
                 // The node must have a parent node in order to use this variant.
                 const parent = parser.nodeParentNode(self_node) orelse return error.NoModificationAllowed;
+                // Parent cannot be Document.
+                if (parser.nodeType(parent) == .document) {
+                    return error.NoModificationAllowed;
+                }
                 // Get the next sibling or null; null indicates our node is the only one.
                 const sibling = parser.nodeNextSibling(self_node);
                 break :blk .{ parent, sibling };
--- a/src/browser/dom/element.zig
+++ b/src/browser/dom/element.zig
@@ -251,11 +251,11 @@ pub const Element = struct {
         // None of the following can be null.
         const maybe_html = parser.nodeFirstChild(fragment_node);
         std.debug.assert(maybe_html != null);
-        const html = maybe_html.?;
+        const html = maybe_html orelse return;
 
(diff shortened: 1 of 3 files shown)


src/browser/dom/element.zig
```diff
@@ -231,6 +231,80 @@ pub const Element = struct {
...
+
+        // Parse the fragment.
+        // Should return error.Syntax on fail?
+        const fragment = try parser.documentParseFragmentFromStr(doc, input);
+        const fragment_node = parser.documentFragmentToNode(fragment);
+
+        // We always get it wrapped like so:
+        // <html><head></head><body>{ ... }</body></html>
+        // None of the following can be null.
+        const html = parser.nodeFirstChild(fragment_node).?;
```
krichprollsch: I think it's better to return in case of null value returned (same for the following).

```suggestion
        const html = parser.nodeFirstChild(fragment_node) orelse return;
```

src/browser/dom/element.zig
```diff
@@ -231,6 +231,80 @@ pub const Element = struct {
         }
     }
 
+    /// Parses the given `input` string and inserts it's children to an element at given `position`.
```
karlseguin:
```suggestion
    /// Parses the given `input` string and inserts its children to an element at given `position`.
```

src/browser/dom/element.zig
karlseguin: If you do that, then we should add an assertion:

```zig
std.debug.assert(first_child != null);
```

Then at least in debug mode, if our understanding of libdom proves wrong, we'll be able to learn from it.

src/tests/dom/element.html
```diff
@@ -289,3 +289,40 @@
...
+<div id="insert-adjacent-html-outer-wrapper">
+    <div id="insert-adjacent-html-inner-wrapper">
+        <span></span>
+        <p>content</p>
+    </div>
+</div>
+
+<script id=insertAdjacentHTML>
+  // Insert "beforeend".
+  const wrapper = document.querySelector("div#insert-adjacent-html-inner-wrapper");
```
karlseguin: the id should be unique, the tag name seems unnecessary.

`$` is a wrapper in testing.js for `document.querySelector` and `$$` for `querySelectorAll`..if you want - using the helper of the original full method is fine.

```js
const wrapper = $('#insert-adjacent-html-inner-wrapper');
```

src/browser/dom/element.zig
```diff
@@ -231,6 +231,85 @@ pub const Element = struct {
...
+        // Should return error.Syntax on fail?
+        const fragment = try parser.documentParseFragmentFromStr(doc, input);
+        const fragment_node = parser.documentFragmentToNode(fragment);
+
+        // We always get it wrapped like so:
+        // <html><head></head><body>{ ... }</body></html>
+        // None of the following can be null.
+        const maybe_html = parser.nodeFirstChild(fragment_node);
+        std.debug.assert(maybe_html != null);
+        const html = maybe_html.?;
```
karlseguin: I think the combination of our comments would have been:

```zig
const maybe_html = parser.nodeFirstChild(fragment_node);
std.debug.assert(maybe_html != null);
const html = maybe_html orelse return;
```

So that, in debug mode, this crashes if our expectation turns out to be wrong. But in release mode, it doesn't crash (but it fails to insert). I think as you've written it, it fails in debug and is undefined in release.

PR #1100, opened 2025-09-29
Add mimalloc allocator

nikneym:
This PR adds wraps mimalloc as a Zig Allocator and uses it instead of c_allocator in main.


karlseguin:
That's interesting. Is the goal better performance and/or memory usage? Any comparison?

nikneym:
For the future readers, it seem to be using more memory than we had with c_allocator, you can inspect [here](https://github.com/lightpanda-io/browser/actions/runs/18091051657/job/51479032415). Closing in favor.

PR #1115, opened 2025-10-02
Small URL & Location changes

nikneym:
This PR;
* adds `href` setter to location,
* removes allocation for `protocol` getter of url,
* adds an invalid (which should not be passing) URL test.

### DIFF
--- a/src/browser/html/location.zig
+++ b/src/browser/html/location.zig
@@ -29,6 +29,10 @@ pub const Location = struct {
         return "";
     }
 
+    pub fn set_href(_: *const Location, href: []const u8, page: *Page) !void {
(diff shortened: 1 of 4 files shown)


src/browser/html/location.zig
```diff
@@ -29,8 +29,12 @@ pub const Location = struct {
         return "";
     }
 
-    pub fn get_protocol(self: *Location, page: *Page) ![]const u8 {
-        if (self.url) |*u| return u.get_protocol(page);
+    pub fn set_href(_: *const Location, href: []const u8, page: *Page) !void {
+        return page.navigate(href, .{ .reason = .script });
```
karlseguin: You always want to use `navigateFromWebAPI`.

PR #1119, opened 2025-10-03
Emit Log.addEntry

karlseguin:
Currently, this hooks a single log.Interceptor into the logging framework, but changing it to take a list shouldn't be too hard. Biggest issue is who will own it, as we'd need an allocator to maintain a list / lookup (which log doesn't currently have).

Uses logFmt format, and, for now, always filters out debug messages and a few particularly verbose scopes.


krichprollsch (commented):
LGTM :+1:

Issue #1121, opened 2025-10-03
Panic due to InvalidURI exception in JavaScript module on nytimes.com


prefrontalsys:
Description

  The LightPanda browser panics and crashes with a "reached unreachable code" error when attempting to load
  https://www.nytimes.com. The crash appears to be triggered by an InvalidURI exception within the browser's
  JavaScript engine while parsing a script from static01.nyt.com.

  Impact

  The panic causes the lightpanda process to hang indefinitely.
   * When run from the command line, the shell hangs and requires a ^C (Ctrl+C) to terminate the process and
     return to the prompt.
   * When run as a subprocess by the Magg MCP proxy server, the lp browser crash causes the Magg server to also
     hang. A ^C in the Magg terminal is required, which then terminates both the Magg server and the hung
     lightpanda process.

  Steps to Reproduce

   1. Launch the LightPanda browser from the command line:
       - lightpanda https://www.nytimes.com
   2. The browser process will crash and the terminal will hang.

  Expected Behavior

  The page should either load successfully or fail gracefully without crashing or hanging the browser process.

  Actual Behavior

  The browser process panics and hangs, requiring a manual ^C to terminate.

  System Information

   * Operating System: macOS 26.0.1 (Tahoe)
   * Architecture: aarch64 (Apple Silicon, MacBook Air M4)

  Crash Log

    1 $time=6040384 $scope=console $level=warn $msg=warn args="; 1: limitedAds is not set, defaulting
      to false"
    2 $time=6040385 $scope=console $level=warn $msg=warn args="; 1: userinfo data unavailable"
    3 $time=6040561 $scope=js $level=warn $msg="compile resolved module" specifier=./identify-user-NlI
      stack=null src=null line=0 exception=InvalidURI
    4 $time=6040561 $scope=user_script $level=warn $msg="eval script"
      url=https://static01.nyt.com/video-static/betamax/player-2G-OVVMI.js err=unknown cacheable=true
    5 thread 124698 panic: reached unreachable code
    6 ???:?:?: 0x102405f2b in _browser.js.Context.resolveModuleCallback (???)
    7 ???:?:?: 0x101090c73 in _locale_charset (???)
    8 ???:?:?: 0x10105d09f in _locale_charset (???)
    9 ???:?:?: 0x100b6e3d7 in _locale_charset (???)
   10 ???:?:?: 0x100af51df in _locale_charset (???)
[...]

PR #1123, opened 2025-10-06
Remove the single-blocking-import restrictions

karlseguin:
Remove the is_blocking variable (and checks) from the ScriptManager. This is a holdover to when blocking imports had a dedicated connections, and thus couldn't be loaded concurrently.

This changes two obvious things (and probably a few subtle ones). The first is that plain async scripts are now free to be executed as soon as they are completed. As far as I can tell, this is a safe behavior, is simpler and should be a bit faster.

More importantly, it allows for chains of imports. normal import (A) -> dynamic import (B) -> normal import (C). This would previously fail an assertion. The superficial issue was that dynamic import handling didn't respect the `is_blocking` flag. But if we did respect it, than module B would never be able to load module C, which would block module A from ever completing. By removing the flag, module C will now be properly evaluated, which unblocks module B which allows module A to unblock.

(1) I believe this is the issue seen here:
    https://github.com/lightpanda-io/browser/issues/1121

### DIFF
--- a/src/browser/ScriptManager.zig
+++ b/src/browser/ScriptManager.zig
@@ -38,21 +38,12 @@ page: *Page,
 // used to prevent recursive evalutaion
 is_evaluating: bool,
 
-// used to prevent executing scripts while we're doing a blocking load
-is_blocking: bool = false,
-
 // Only once this is true can deferred scripts be run
 static_scripts_done: bool,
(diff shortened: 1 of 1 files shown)


krichprollsch (approved):
Looks simpler this way.

PR #1127, opened 2025-10-07
Nuke `std.Uri`, integrate Ada URL

nikneym:
This PR does a `URL` overhaul by introducing [ada-url](https://github.com/ada-url/ada) to the codebase.

* Removes all `std.Uri` usage.
* Href strings are normalized as WHATWG requires.
* Refactors `HTMLAnchorElement` by introducing `object_data` in `Page`.

The CDP changes require extra attention; this is the first time I'm making a change in that part of the code. 👀

### DIFF
--- a/src/tests/html/link.html
+++ b/src/tests/html/link.html
@@ -16,8 +16,10 @@
   testing.expectEqual('https://lightpanda.io', link.origin);
 
   link.host = 'lightpanda.io:443';
-  testing.expectEqual('lightpanda.io:443', link.host);
-  testing.expectEqual('443', link.port);
+  // Port is omitted if its the default one for the scheme.
+  testing.expectEqual('lightpanda.io', link.host);
+  // Port is omitted if its the default one for the scheme.
+  testing.expectEqual('', link.port);
   testing.expectEqual('lightpanda.io', link.hostname);
 
   link.host = 'lightpanda.io';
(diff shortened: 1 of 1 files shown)


src/browser/page.zig
```diff
@@ -122,12 +127,30 @@ pub const Page = struct {
         complete,
     };
 
+    const ObjectDataMap = std.HashMapUnmanaged(
```
karlseguin, line 130: Is this being replaced with the `state`, e.g. `getOrCreateNodeState` ?

src/url.zig
```diff
@@ -1,82 +1,233 @@
...
+        return href.data[0..href.length];
+    }
 
-        return .{
-            .uri = uri,
-            .raw = str,
-        };
+    pub fn getHostname(self: URL) []const u8 {
+        const hostname = ada.getHostnameNullable(self.internal);
+        return hostname.data[0..hostname.length];
```
karlseguin, line 100: presumably this can be null?

src/browser/page.zig
nikneym: No, my idea was to keep `URL` needed for `HTMLAnchorElement` only. I think it should live inside `State`.

Edit: `State` seem to be holding many things... I'm not sure which part should be taken.

anonrig (approved):

PR #1133, opened 2025-10-09
Add a fast path for validating cookie strings

nikneym:
This prefers `suggestVectorLength` in order to pick a vector size; for cookie strings shorter than, say 64, this might cause it to fallback to slow path on architectures that support larger vector sizes (like AVX-512). We may also add checks for smaller vector sizes if desired in the future.

### DIFF
--- a/src/browser/storage/cookie.zig
+++ b/src/browser/storage/cookie.zig
@@ -315,7 +315,7 @@ pub const Cookie = struct {
     const ValidateCookieError = error{ Empty, InvalidByteSequence };
 
     /// Returns an error if cookie str length is 0
-    /// or contains characters between 32...126.
+    /// or contains characters outside of the ascii range 32...126.
     fn validateCookieString(str: []const u8) ValidateCookieError!void {
         if (str.len == 0) {
             return error.Empty;


src/browser/storage/cookie.zig
```diff
@@ -321,6 +312,56 @@ pub const Cookie = struct {
...
+
+                // Got match.
+                if (reduced != 0) {
+                    return error.InvalidByteSequence;
+                }
+            }
+
+            // Means str.len % size == 0; we also know str.len != 0.
+            // Cookie is valid.
+            if (offset == str.len) {
```
karlseguin, line 351: This seems rare enough to not be worth it?

src/browser/storage/cookie.zig
```diff
@@ -321,6 +312,56 @@ pub const Cookie = struct {
         };
     }
 
+    const ValidateCookieError = error{ Empty, InvalidByteSequence };
+
+    /// Returns an error if cookie str length is 0
+    /// or contains characters between 32...126.
```
karlseguin: reversed

```suggestion
    /// or contains characters outside of the ascii range 32...126.
```

src/browser/storage/cookie.zig
nikneym: True but we have to return early since `std.mem.minMax` asserts len != 0; acts like a guard for that.

nikneym: Maybe we can do a scalar search instead; that would allow us to remove the check.

karlseguin: Arg..ok..guess we keep it in.

PR #1135, opened 2025-10-09
Importmap support

### LINKED ISSUE #1129: JS Import maps
Ruby on Rails applications use import maps: https://caniuse.com/import-maps and so we end up with a script tag like this:

```html

<script type="importmap" data-turbo-track="reload">{
--
"imports": {
"application": "/assets/application-bfcdf840.js",
"@hotwired/turbo-rails": "/assets/turbo.min-c1e82531.js",
"@hotwired/stimulus": "/assets/stimulus.min-4b1e420e.js",
"@hotwired/stimulus-loading": "/assets/stimulus-loading-1fc53fe7.js",
"controllers/application": "/assets/controllers/application-3affb389.js",
"controllers/hello_controller": "/assets/controllers/hello_controller-708796bd.js",
"controllers": "/assets/controllers/index-ee64e1f1.js"
}
}</script>

```

From what we can tell lightpanda doesn't know how to handle this. We can try to PR a fix for this soon but I wasn't sure if there is already someone else looking into it?

krichprollsch:
fix #1129

### DIFF
--- a/src/browser/ScriptManager.zig
+++ b/src/browser/ScriptManager.zig
@@ -72,7 +72,8 @@ sync_modules: std.StringHashMapUnmanaged(*SyncModule),
 
 // Mapping between module specifier and resolution.
 // see https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/script/type/importmap
-importmap: std.StringHashMapUnmanaged([]const u8),
+// importmap contains resolved urls.
+importmap: std.StringHashMapUnmanaged([:0]const u8),
 
 const OrderList = std.DoublyLinkedList;
 
@@ -124,7 +125,7 @@ pub fn reset(self: *ScriptManager) void {
     self.sync_modules.clearRetainingCapacity();
     // Our allocator is the page arena, it's been reset. We cannot use
     // clearAndRetainCapacity, since that space is no longer ours
-    self.importmap.clearAndFree(self.page.arena);
+    self.importmap = .empty;
 
     self.clearList(&self.asyncs);
     self.clearList(&self.scripts);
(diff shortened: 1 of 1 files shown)


src/tests/html/script/importmap.html
```diff
@@ -0,0 +1,24 @@
...
+      }
+    }
+</script>
+
+<script id=use_importmap type=module>
+  import * as im from 'core';
+  testing.expectEqual('hello', im.greeting);
+</script>
+
+<script id=cached_importmpa type=module>
```
bcardarella: typo with `cached_importmpa` ?

src/browser/ScriptManager.zig
```diff
@@ -115,6 +122,9 @@ pub fn reset(self: *ScriptManager) void {
         self.sync_module_pool.destroy(value_ptr.*);
     }
     self.sync_modules.clearRetainingCapacity();
+    // Our allocator is the page arena, it's been reset. We cannot use
+    // clearAndRetainCapacity, since that space is no longer ours
+    self.importmap.clearAndFree(self.page.arena);
```
karlseguin:
```suggestion
    self.importmap = .empty
```

src/browser/ScriptManager.zig
```diff
@@ -248,6 +261,22 @@ pub fn addFromElement(self: *ScriptManager, element: *parser.Element, comptime c
     });
 }
 
+// Resolve a module specifier to an valid URL.
+pub fn resolveSpecifier(self: *ScriptManager, arena: Allocator, _specifier: []const u8, base: []const u8) ![:0]const u8 {
+    var specifier = _specifier;
+    // If the specifier is mapped in the importmap, use it to resolve the path.
+    if (self.importmap.get(specifier)) |s| {
```
karlseguin, line 268: I would expect the URLs in the import map to be resolved based on the document's URL and not the referrer of the  module. If that's the case, then

1 - When populating `importmap`, the imports need to be resolved against page.url
2 - In this function, when found, it should be returned directly from the map

krichprollsch: good point

krichprollsch: Should be fixed now

src/tests/html/script/importmap.html
krichprollsch: good catch, thanks

PR #1140, opened 2025-10-10
remove invalid errdefer

krichprollsch:


karlseguin:
I figured if there was an error, we'd try again..but I agree, it might not make sense. Is it causing an issue?

krichprollsch:
Yes, I have a crash sometimes with https://github.com/lightpanda-io/demo/pull/81
and once removed, I have another one with a debug log.

Issue #1116, opened 2025-10-02
Intermittent segfault crash when loading Amazon.pl


PawiX25:
**Environment:** `lightpanda/browser:nightly` on ARM64 (Raspberry Pi), Docker, Playwright Python

**Reproduce:**
```python
import asyncio
from playwright.async_api import async_playwright

async def test():
    playwright = await async_playwright().start()
    browser = await playwright.chromium.connect_over_cdp("ws://IP:9222")
    context = browser.contexts[0] if browser.contexts else await browser.new_context()
    page = await context.new_page()
    
    await page.goto("http://example.com")  # Works
    await page.goto("https://www.amazon.pl/")  # Crashes intermittently
    
    await browser.close()
    await playwright.stop()

asyncio.run(test())
```

**Error (Playwright side):**
```
playwright._impl._errors.TargetClosedError: Page.goto: Target page, context or browser has been closed
```

**Error (Docker logs):**
```
TypeError: document.write is not a function
Segmentation fault at address 0x32dcd2c1
/home/runner/work/.../hash_map.zig:1155:39: in createTimeout
```

**Result:** Container exits code 133, CDP connection closes.

**Note:** Sometimes loads successfully (205KB page), but usually crashes. Reproducible with fresh restart.

krichprollsch:
Hello @PawiX25, we fixed some crashes recently.
I tried your script with the last version, it blocks, but it doesn't crash.
Can you make a retry please?
Thanks!

krichprollsch:
My bad, I just get the crash 🤦

krichprollsch:
I can't really reproduce the crash :/ But I found an issue blocking the load fixed in #1150 
@PawiX25 can you make some retries please?

PawiX25:
It works fine, I can now fetch successfully. Thanks for the help :)

PR #1150, opened 2025-10-15
isDone must be run after script's deinit

krichprollsch:
before this change, isDone was run before the pending script deinit. So the `asyncs` list wasn't yet cleared and the document never completed...

I was abe to reproduce by fetching amazon.pl

Relates with #1116


karlseguin:
I believe that https://github.com/lightpanda-io/browser/pull/1197, which was merged into main and later into zigdom, reworked this so that it isn't an issue anymore (and thus this behavior should be working correctly in zigdom without this explicit change).

Issue #1148, opened 2025-10-14
page.fill: TypeError: input.select is not a function


smarlhens:
# page.fill: TypeError: input.select is not a function

Hey 👋🏻

I hope you are well and thanks for your amazing work on Lightpanda Browser !

I'm trying to use Lightpanda with Playwright (`1.56.0`) and I'm wondering if I'm using it the right way. I'm running the browser using Docker Compose with the following configuration:

```yaml
lightpanda:
  image: lightpanda/browser:latest
  ports:
    - "9222:9222"
  tty: true
  restart: always
  deploy:
    resources:
      limits:
        memory: 2G
        cpus: '2'
```

## Issue Description
When using `page.fill()` on an input element, I get the following error:

```
page.fill: TypeError: input.select is not a function
    at InjectedScript.selectText (<anonymous>:7094:13)
    at InjectedScript.fill (<anonymous>:7085:10)
    at eval (eval at evaluate (:290:30), <anonymous>:7:25)
    at async <anonymous>:316:30
```

## Call log
```
- waiting for locator('input[name="q"]')
  - locator resolved to <input value="" name="q" size="57" class="lst" maxlength="2048" autocomplete="off" title="Recherche Google"/>
  - fill("lightpanda")
- attempting fill action
  - waiting for element to be visible, enabled and editable
```

## Reproduction code

```typescript
import type { Browser, Page } from 'playwright';
import { chromium } from 'playwright';

const main = async () => {
  const browser: Browser = await chromium.connectOverCDP('ws://127.0.0.1:9222');
  const context = await browser.newContext();
  const page: Page = await context.newPage();
  await page.goto('https://www.google.com/');
  await page.fill('input[name="q"]', 'lightpanda');
  await browser.close();
};

main().catch(err => {
  console.error(err);
  process.exit(1);
});
```

## Expected behavior
The input field should be filled with the text "lightpanda".

## Actual behavior
An error is thrown: `TypeError: input.select is not a function`

krichprollsch:
Hello @smarlhens ,
Thanks for this documented issue 🙏 
We merged #1152 which should fix the issue.
Could you make a retry please?

smarlhens:
Thanks a lot @krichprollsch 🙏 for the quick fix and great reactivity!
I tested with a Docker image built locally, and it seems to be working perfectly now 🎉

krichprollsch:
Good news! Thank you for the feedback 🙏

PR #1152, opened 2025-10-15
cdp: add input.insertText

krichprollsch:
Relates with #1148


karlseguin:
can do it separately, but I sometimes see `el.blur()` being called, which could impact how/where this behaves. (I think in DDG)

krichprollsch:
I agree, we have work to do regarding that:
* dispatch events on text selection, focus and blur
* listen these events to maintain a document's (or window) state of active element and selected text (I know active_element is in the document's state)
* implement `Selection` API

I will write an issue for all of that.

PR #1158, opened 2025-10-17
handle multiple waiters for the same module

krichprollsch:
relates with #1121


karlseguin:
@krichprollsch Do you remember how reliable the crash was in https://github.com/lightpanda-io/browser/issues/1121  I can't reproduce the issue in zigdom, but I haven't added the multiple waiters code. I feel like it shouldn't be possible for this case to happen - the first waiter should return a promise that all other waiters use.

krichprollsch:
oh sounds like a way better solution :eyes:

Issue #1160, opened 2025-10-19
WS Connection error: Ws(Protocol(ResetWithoutClosingHandshake))


holmofy:
Thanks to the awesome lightpanda. I'm using the Rust community's [chromiumoxide](https://github.com/mattsse/chromiumoxide) to connect to lightpanda, but I can't connect to lightpanda. 

Below is my code repository: 

* https://github.com/AutoWDS/autowds-instance

https://github.com/mattsse/chromiumoxide/issues/267

krichprollsch:
Hello @holmofy,
Thanks for your issue.

I'm using a CDP proxy to dump the messages exchanged between your project and Lightpanda.
A first issue is chromiumxide uses `Security.setIgnoreCertificateErrors` which is not implemented on Lightpanda.
I will try to fix it first and see if it unblocks you.

```
> {"id":10,"method":"Security.setIgnoreCertificateErrors","sessionId":"SID-1","params":{"ignore":true}}
< {"id":10,"error":{"code":-31998,"message":"UnknownMethod"},"sessionId":"SID-1"}
```

krichprollsch:
Another issue I can see is we don't support `url` in createTarget for now (see https://github.com/lightpanda-io/browser/blob/main/src/cdp/domains/target.zig#L112).

I will take a look at this 👀 

```
> {"id":1,"method":"Target.createTarget","params":{"url":"https://en.wikipedia.org"}}
```

krichprollsch:
A third issue is when we run `find_element`:
```
2025-10-21T15:58:29.397592Z ERROR autowds_instance: 💥 Browser session error: Node is either not visible or not an HTMLElement
```

It's due to a `false` result of this JS script eval:

```
[...]

PR #1169, opened 2025-10-21
cdp: implement Security.setIgnoreCertificateErrors

krichprollsch:
In relation with #1160

I assume I can change `CURLOPT_SSL_VERIFYHOST`, `CURLOPT_SSL_VERIFYPEER`, `CURLOPT_PROXY_SSL_VERIFYHOST` and `CURLOPT_PROXY_SSL_VERIFYPEER` with active connections. Not sure if it's safe :thinking:

### DIFF
--- a/src/cdp/domains/security.zig
+++ b/src/cdp/domains/security.zig
@@ -35,15 +35,34 @@ fn setIgnoreCertificateErrors(cmd: anytype) !void {
         ignore: bool,
     })) orelse return error.InvalidParams;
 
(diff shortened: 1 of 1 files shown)


karlseguin:
I don't think it's safe...except for the fact that CDP generally does all of its setup before navigating to the page.

You could check if `self.in_use.first  != null` and either log a message or reject the change.

krichprollsch:
PR updated

PR #1171, opened 2025-10-21
support url on createTarget and send lifecycle events

krichprollsch:
Relates with #1160

* Support `url` parameter on createTarget. we now navigate on `createTarget` to dispatch events correctly, even in case of `about:blank`.
* Send lifecycle events when enabling the feature: chrome dispatch events on activation if the page is loaded
* Remove inflight connections check on enable/disable TLS b/c [chromiumoxide](https://github.com/mattsse/chromiumoxide) calls the command during navigate and Curl seems to accept it...

### DIFF
--- a/src/cdp/domains/target.zig
+++ b/src/cdp/domains/target.zig
@@ -180,7 +180,6 @@ fn createTarget(cmd: anytype) !void {
 
     try page.navigate(params.url, .{
         .reason = .address_bar,
-        .cdp_id = cmd.input.id,
     });
 
     try cmd.sendResult(.{


src/cdp/domains/page.zig
```diff
@@ -18,6 +18,7 @@
 
 const std = @import("std");
 const Page = @import("../../browser/page.zig").Page;
+const timestampF = @import("../../datetime.zig").timestamp;
```
karlseguin, line 21: `timestampF` because it collides with existing `timestamp` variables?

src/cdp/domains/page.zig
```diff
@@ -82,11 +83,33 @@ fn setLifecycleEventsEnabled(cmd: anytype) !void {
...
+
+    // Enable lifecycle events.
+    try bc.lifecycleEventsEnable();
+
+    // When we enable lifecycle events, we must dispatch events for all
+    // attached targets.
+    const page = bc.session.currentPage() orelse return error.PageNotLoaded;
+
+    if (page.load_state == .complete) {
+        try sendPageLifecycle(bc, "DOMContentLoaded", timestampF());
```
karlseguin, line 100: I like that all this is in the CDP layer!

karlseguin:
demo's `node puppeteer/cdp.js` hangs on this branch, which is probably the same thing as the failing action.

src/cdp/domains/page.zig
krichprollsch: indeed, do you have a better suggestion?

karlseguin: Nope, was just curious.

krichprollsch:
I fixed puppeteer/playwright issues

krichprollsch:
~~Still have an issue with chromedp :detective:~~
It works in fact...

PR #1173, opened 2025-10-22
renderer: set a default box size of 5 pixels

krichprollsch:
Relates with #1160

[chromiumoxide](https://github.com/mattsse/chromiumoxide) expects a box bigger than 1x1 to select an element.

This PR changes the default element size in the flat renderer to move from 1px to 5px.


karlseguin (approved):
LGTM

karlseguin:
Renderer in Zigdom has changed, but it's still using 5x5 width/height in most cases (it tries to respect basic css visibility rules and width/height attributes).

Issue #1176, opened 2025-10-24
Use v8 Snapshot to inject polyfills


krichprollsch:
We use JS [polyfills](https://github.com/lightpanda-io/browser/tree/main/src/browser/polyfill) to supports some web APIs.
These polyfills are injected into v8 during page [initialization](https://github.com/lightpanda-io/browser/blob/main/src/browser/page.zig#L150-L151).

The polyfill loader [executes the pre-load part](https://github.com/lightpanda-io/browser/blob/main/src/browser/polyfill/polyfill.zig#L86-L100) of the polyfill, and it [lazy loads](https://github.com/lightpanda-io/browser/blob/main/src/browser/polyfill/polyfill.zig#L52-L84) the reset by [implementing a V8's callback](https://github.com/lightpanda-io/browser/blob/main/src/browser/js/js.zig#L218-L240) on  missing elements.

This lazy load is a hacking and not perfect way to inject polyfill, and we would prefer loading them all the time but faster.

Since we know these scripts are always the same, we would like to use a faster way to inject them into v8, one idea was to use the snapshot feature from v8. We had an attempt to implement them previously, but it didn't work. See https://github.com/lightpanda-io/zig-v8-fork/pull/82 and #845

karlseguin:
We no longer polyfill anything (And we do use snapshots,just not for this specific reason).

PR #1187, opened 2025-10-28
Provides an PoC of using V8 snapshots

mrdimidium:
Provides a proof of concept on how to use snapshots for #1176. Depends on PR in v8 repository.: https://github.com/lightpanda-io/zig-v8-fork/pull/110

As stated in the documentation, snapshots are serialized state of the V8 heap. We simply execute JavaScript code and save the snapshot (it can be saved to disk if desired). In the PR above, you can see how a snapshot is created at browser startup, which provides the `LightpandaSnapshotPoC` function, later accessible in tests.

Since the polyfill is still JavaScript code, it can only be executed and serialized along with the initialization of window* and other global objects, like this:
- at browser startup, create a clean context
- add global WebAPIs to the context
- add the polyfills to the context
- serialize the snapshot for subsequent pages
- (optional) save the snapshot to disk

Currently, this isn't very convenient to do in code, as the snapshot must be ready before creating the Env. I'd like to discuss the implementation in advance:
- when creating the first page, save a snapshot for subsequent pages in env;
- change Env so that it proactively prepares a snapshot at startup, and only loads Page from the snapshot and completely skips preparing window*

@krichprollsch, wdyt?

### DIFF
--- a/src/browser/js/Env.zig
+++ b/src/browser/js/Env.zig
@@ -3,6 +3,7 @@ const js = @import("js.zig");
 const v8 = js.v8;
 
 const log = @import("../../log.zig");
+const polyfill = @import("../polyfill/polyfill.zig");
 
 const types = @import("types.zig");
 const Types = types.Types;
@@ -34,6 +35,8 @@ isolate: v8.Isolate,
 // just kept around because we need to free it on deinit
 isolate_params: *v8.CreateParams,
 
+isolate_snaphost: v8.StartupData,
+
 // Given a type, we can lookup its index in TYPE_LOOKUP and then have
 // access to its TunctionTemplate (the thing we need to create an instance
 // of it)
@@ -54,6 +57,21 @@ context_id: usize,
 const Opts = struct {};
 
 pub fn init(allocator: Allocator, platform: *const Platform, _: Opts) !*Env {
+    const env = try allocator.create(Env);
+    errdefer allocator.destroy(env);
+
+    env.* = .{
+        .context_id = 0,
+        .isolate = undefined,
+        .isolate_params = undefined,
+        .isolate_snaphost = undefined,
+        .platform = platform,
+        .templates = undefined,
+        .allocator = allocator,
+        .meta_lookup = undefined,
+        .prototype_lookup = undefined,
+    };
+
     // var params = v8.initCreateParams();
     var params = try allocator.create(v8.CreateParams);
     errdefer allocator.destroy(params);
@@ -63,8 +81,14 @@ pub fn init(allocator: Allocator, platform: *const Platform, _: Opts) !*Env {
     params.array_buffer_allocator = v8.createDefaultArrayBufferAllocator();
     errdefer v8.destroyArrayBufferAllocator(params.array_buffer_allocator.?);
 
+    env.isolate_snaphost = generateIsolateSnaphot(params);
+    params.snapshot_blob = &env.isolate_snaphost;
+
     var isolate = v8.Isolate.init(params);
     errdefer isolate.deinit();
+
+    env.isolate = isolate;
+    env.isolate_params = params;
 
     // This is the callback that runs whenever a module is dynamically imported.
     isolate.setHostImportModuleDynamicallyCallback(Context.dynamicModuleCallback);
@@ -79,20 +103,6 @@ pub fn init(allocator: Allocator, platform: *const Platform, _: Opts) !*Env {
     var temp_scope: v8.HandleScope = undefined;
     v8.HandleScope.init(&temp_scope, isolate);
     defer temp_scope.deinit();
-
-    const env = try allocator.create(Env);
-    errdefer allocator.destroy(env);
-
-    env.* = .{
-        .context_id = 0,
-        .platform = platform,
-        .isolate = isolate,
-        .templates = undefined,
-        .allocator = allocator,
-        .isolate_params = params,
-        .meta_lookup = undefined,
-        .prototype_lookup = undefined,
-    };
 
     // Populate our templates lookup. generateClass creates the
     // v8.FunctionTemplate, which we store in our env.templates.
@@ -536,4 +546,22 @@ fn generateUndetectable(comptime Struct: type, template: v8.ObjectTemplate) void
         }
         template.markAsUndetectable();
     }
+}
+
+fn generateIsolateSnaphot(params: *v8.CreateParams) v8.StartupData {
+    var snapshot_creator = v8.SnapshotCreator{};
+    snapshot_creator.init(params);
+    defer snapshot_creator.deinit();
+
+    const isolate = snapshot_creator.getIsolate();
+
+    var temp_scope: v8.HandleScope = undefined;
+    v8.HandleScope.init(&temp_scope, isolate);
+    defer temp_scope.deinit();
+
+    const context = v8.Context.init(isolate, null, null);
+    context.enter();
+    defer context.exit();
+
+    return snapshot_creator.createSnapshotDataBlob("function LightpandaSnapshotPoC() { return 73; }");
 }
--- a/src/tests/polyfill/webcomponents.html
+++ b/src/tests/polyfill/webcomponents.html
@@ -13,6 +13,8 @@
     }
   }
 
+  testing.expectEqual(73, LightpandaSnapshotPoC());
+
   window.customElements.define("lightpanda-test", LightPanda);
   const main = document.getElementById('main');
   main.appendChild(document.createElement('lightpanda-test'));


mrdimidium:

mrdimidium:
@krichprollsch, to pass the build and tests, need to build with PR in v8 repo

krichprollsch:
@mrdimidium,
Thanks for the PR.

> (it can be saved to disk if desired)

Do you know if the data is independent from the platform? Could we generate the blob in a file and commit it along the project?

Regarding the loading strategy, what about changing the polyfill to be independent from any global object, and only run a small JS when need to create what we need with global objects? I didn't look at the JS code for now, so idk if it's hard to do, but it could worth the try.

WDYT?

mrdimidium:
@krichprollsch, it's not portable across platforms, but it can be generated during the build.  For example, Chromium builds the snapshot in GN and packages it into an `apk/dmg/apk`. This also allows them to avoid having to include extra snapshot generation code at runtime.

I was thinking about changing the JS, but I couldn't immediately figure out how to handle the "something in the window was accessed" event before the window actually appears. Maybe something with globalThis?

It will also be slower. If we take a snapshot after the JS has executed, we'll get an already initialized window in the snapshot. If we use JS wrappers, it will need to be executed at runtime on the page.

krichprollsch:
Got it. I'm not sure about the good balance here. I'm afraid we will have to generate the snapshot too often according the way we are creating js env & globals right now. But in the other hand that could force us to refacto for a better way.

krichprollsch:
Thinking a bit more about that, I'm not sure it worth it, I'm afraid we will have to generate the snapshot at the same moment we currently inject the polyfill. So there will be no gain at all.

mrdimidium:
I think this could work if the global API preparation were separated from the page. In that case, a snapshot with the WebAPI (including polyfills) could be prepared in advance. This should be done once during the build and completely removed from runtime.

But it seems like this would require significant refactoring.

krichprollsch:
I think I will take a look a the polyfill first to see if we can extract the global deps first.

krichprollsch:
I tried the branch locally (ubuntu x86_64), the webcomponent test is working correctly.
```
$ F='Browser: Polyfill.WebComponents' make test
Browser: Polyfill.WebComponents (113.85ms)

1 of 1 test passed

Slowest 1 test:
  113.85ms      Browser: Polyfill.WebComponents


Build Summary: 7/7 steps succeeded
test success
└─ run test success 154ms MaxRSS:62M
   └─ compile test Debug native cached 15ms MaxRSS:39M
      ├─ options cached
      ├─ WriteFile ada_c.h cached
      ├─ compile lib ada Debug native cached 20ms MaxRSS:38M
      ├─ compile lib ada Debug native (reused)
      └─ options cached
```

But I have a crash when running the whole test suite :thinking: 

```
Client: http invalid request (0.47ms)
Client: http invalid handshake (2.01ms)


#
# Fatal error in ../../../src/src/heap/read-only-spaces.cc, line 96
# Check failed: read_only_blob_checksum_ == snapshot_checksum (<unprintable> vs. 987505134).
#
#
#
#FailureMessage Object: 0x732cebff4500
==== C stack trace ===============================

    ./.zig-cache/o/82a9ee220bf2145336ad999ef7b4f2bc/test() [0x3ecc943]
    ./.zig-cache/o/82a9ee220bf2145336ad999ef7b4f2bc/test() [0x3eca91d]
    ./.zig-cache/o/82a9ee220bf2145336ad999ef7b4f2bc/test() [0x3f3e3d4]
    ./.zig-cache/o/82a9ee220bf2145336ad999ef7b4f2bc/test() [0x4f735db]
    ./.zig-cache/o/82a9ee220bf2145336ad999ef7b4f2bc/test() [0x41c4b36]
    ./.zig-cache/o/82a9ee220bf2145336ad999ef7b4f2bc/test() [0x408d4a3]
    ./.zig-cache/o/82a9ee220bf2145336ad999ef7b4f2bc/test() [0x408ef71]
    ./.zig-cache/o/82a9ee220bf2145336ad999ef7b4f2bc/test() [0x4f574bb]
    ./.zig-cache/o/82a9ee220bf2145336ad999ef7b4f2bc/test() [0x4f5d121]
    ./.zig-cache/o/82a9ee220bf2145336ad999ef7b4f2bc/test() [0x3fd4798]
    ./.zig-cache/o/82a9ee220bf2145336ad999ef7b4f2bc/test() [0x3ec1826]
    ./.zig-cache/o/82a9ee220bf2145336ad999ef7b4f2bc/test() [0x32fd130]
    ./.zig-cache/o/82a9ee220bf2145336ad999ef7b4f2bc/test() [0x328165e]
    ./.zig-cache/o/82a9ee220bf2145336ad999ef7b4f2bc/test() [0x327602e]
    ./.zig-cache/o/82a9ee220bf2145336ad999ef7b4f2bc/test() [0x3287525]
    ./.zig-cache/o/82a9ee220bf2145336ad999ef7b4f2bc/test() [0x3287aa0]
    ./.zig-cache/o/82a9ee220bf2145336ad999ef7b4f2bc/test() [0x32889f4]
    ./.zig-cache/o/82a9ee220bf2145336ad999ef7b4f2bc/test() [0x328a28d]
    ./.zig-cache/o/82a9ee220bf2145336ad999ef7b4f2bc/test() [0x328b426]
    ./.zig-cache/o/82a9ee220bf2145336ad999ef7b4f2bc/test() [0x328bd2c]
    ./.zig-cache/o/82a9ee220bf2145336ad999ef7b4f2bc/test() [0x328bee0]
    ./.zig-cache/o/82a9ee220bf2145336ad999ef7b4f2bc/test() [0x328c696]
    ./.zig-cache/o/82a9ee220bf2145336ad999ef7b4f2bc/test() [0x328d4fd]
    ./.zig-cache/o/82a9ee220bf2145336ad999ef7b4f2bc/test() [0x3cb714d]
    ./.zig-cache/o/82a9ee220bf2145336ad999ef7b4f2bc/test() [0x3cb2fe3]
    ./.zig-cache/o/82a9ee220bf2145336ad999ef7b4f2bc/test() [0x3cb0bd4]
    /lib/x86_64-linux-gnu/libc.so.6(+0x9caa4) [0x732cf709caa4]
    /lib/x86_64-linux-gnu/libc.so.6(+0x129c6c) [0x732cf7129c6c]
test
└─ run test failure
```

mrdimidium:
Interesting, It's an debug assert, but I don't see it. @krichprollsch, are you building on a Mac or Linux?

krichprollsch:
linux, x86_64 (ubuntu 24.04)

mrdimidium:
Give me some time to figure this out. You can temporarily disable dcheck in args.gn.

krichprollsch:
Regarding decoupling the polyfill from window, this experiment is working: https://github.com/lightpanda-io/browser/commit/5a1564dee1cc71ccea60f55b09fa2c990113d56f
But it has drawbacks:
* I still have to use the lazy loading to pass tests, not sure why, but it's not relative to snapshot/decouple from window changes, b/c loading the original polyfill all the time doesn't work either
* we expose the `__polyfillWebcomponents` function in the global state

krichprollsch:
Hey @mrdimidium, thinking a bit more about the feature, and here is what I would like to do:

I want to be able to build a snapshot binary file at compile time.
The constraint will be the snapshot will take only javascript independent from any native pointers. So it must not rely on window or anything else.
I think it's enough to pre-load most of the JS we need this way.

We can create a dedicated snapshot directory in `src/` with a zig file to declare the js to load and js files to snapshot (similar to `polyfill/`).

Ideally we will generate the snapshot then use it to compile the program, all-in-one. I suppose comptime will help here, but we must be sure the zig cache works correctly, I don't want to re-build that all the time on dev...

If it's too complex, we can create  `zig build v8-snapshot` separated from the build dedicated to generate the snapshot.

I want to keep a live polyfill loader like it exists. This will be useful to initiate some small JS at runtime. Maybe we can also imagine in the future be able to load customer's JS script at runtime too.

WDYT @mrdimidium?

mrdimidium:
That sounds reasonable. In fact, creating a snapshot is an inexpensive operation, taking about a second. This is noticeable at runtime, but shouldn't cause any problems during the build.

I agree that the current mechanism should be retained, at least to compare with snapshots under real-world load.

Here's how I could implement it:
- `zig-v8` adds the mksnapshot build target (a regular C++ executable) used by v8
- the browser has a `src/polyfill.zig` file, which is used by build.zig to create the snapshot's JS file
- `build.zig` runs `mksnapshot`, which creates a `.bin` file in zig-out, which is added to the browser's builtins
- if the browser's builtin is available, use it during browser initialization.

How do you like this structure?

krichprollsch:
sounds good to me :+1:

krichprollsch:
closed in favor of #1279

PR #1190, opened 2025-10-30
`Blob` support

nikneym:
This PR adds `Blob` API and moves file API related code to their own folders.

I'll also add SIMD version of [scanning endings](https://developer.mozilla.org/en-US/docs/Web/API/Blob/Blob#endings) in a separate PR.

### DIFF
--- a/src/browser/file/Blob.zig
+++ b/src/browser/file/Blob.zig
@@ -139,9 +139,7 @@ fn writeBlobParts(
 /// Returns a Promise that resolves with the contents of the blob
 /// as binary data contained in an ArrayBuffer.
 pub fn _arrayBuffer(self: *const Blob, page: *Page) !js.Promise {
-    const resolver = page.js.createPromiseResolver(.none);
-    try resolver.resolve(js.ArrayBuffer{ .values = self.slice });
-    return resolver.promise();
+    return page.js.resolvePromise(js.ArrayBuffer{ .values = self.slice });
 }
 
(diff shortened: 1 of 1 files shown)


src/browser/file/Blob.zig
```diff
@@ -0,0 +1,240 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+const std = @import("std");
+const Writer = std.Io.Writer;
+
+const builtin = @import("builtin");
+const is_windows = builtin.os.tag == .windows;
```
karlseguin: not used?

src/browser/file/Blob.zig
```diff
@@ -0,0 +1,240 @@
...
+        // `the quick brown\rfox`
+        // `the quick brown\r\nfox`
+        try writer.writeAll(part[start..end]);
+    }
+}
+
+/// Returns a Promise that resolves with the contents of the blob
+/// as binary data contained in an ArrayBuffer.
+pub fn _arrayBuffer(self: *const Blob, page: *Page) !js.Promise {
+    const resolver = page.js.createPromiseResolver(.none);
```
karlseguin: You can use

```
return page.js.resolvePromise(js.ArrayBuffer{ .values = self.slice })
```

which is a wrapper for what you have.

PR #1198, opened 2025-11-04
Trim CR and LF characters from both ends

### LINKED ISSUE #1194: UrlMalformat error when parsing https://cdn.jsdelivr.net/npm/@mediakind/mkplayer@1.12.0/mkplayer.min.js
When fetching https://www.etro.com/us-en/cotton-jacquard-waistcoat-MRKF0100AQ197S9820.html I get an UrlMalformat error.

```
$ zig build run -- fetch 'https://www.etro.com/us-en/cotton-jacquard-waistcoat-MRKF0100AQ197S9820.html'
```

```
WARN  http : script fetch error . . . . . . . . . . . . . . . [+216ms]
      req = https://cdn.jsdelivr.net/npm/@mediakind/mkplayer@1.12.0/mkplayer.min.js

      err = UrlMalformat
```

nikneym:
Fixes #1194.

### DIFF
--- a/src/url.zig
+++ b/src/url.zig
@@ -81,31 +81,31 @@ pub const URL = struct {
     /// For URLs without a path, it will add src as the path.
     pub fn stitch(
         allocator: Allocator,
-        path: []const u8,
+        raw_path: []const u8,
         base: []const u8,
         comptime opts: StitchOpts,
     ) !StitchReturn(opts) {
-        const trimmed_path = std.mem.trim(u8, path, &.{ '\n', '\r' });
+        const path = std.mem.trim(u8, raw_path, &.{ '\n', '\r' });
 
-        if (base.len == 0 or isCompleteHTTPUrl(trimmed_path)) {
-            return simpleStitch(allocator, trimmed_path, opts);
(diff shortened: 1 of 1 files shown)


src/url.zig
```diff
@@ -85,25 +85,27 @@ pub const URL = struct {
         base: []const u8,
         comptime opts: StitchOpts,
     ) !StitchReturn(opts) {
-        if (base.len == 0 or isCompleteHTTPUrl(path)) {
-            return simpleStitch(allocator, path, opts);
+        const trimmed_path = std.mem.trim(u8, path, &.{ '\n', '\r' });
```
krichprollsch: could save a bit of changes by renaming the `path` parameter and use a `const path = ` here instead.

krichprollsch: Should `base` be trimmed also? 🤔

nikneym: > Should `base` be trimmed also? 🤔

I think it depends; if the given string is not provided from JS environment, it'd be fine since`base`s we provide don't include line endings. We use this function in places where `base` string come from JS too, though.

krichprollsch: ok, let's keep it as it for now

PR #1200, opened 2025-11-05
add `set_hash` to Location

mookums:
This adds `set_hash` to the `Location` WebAPI.

### DIFF
--- a/src/browser/html/location.zig
+++ b/src/browser/html/location.zig
@@ -17,8 +17,6 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 const std = @import("std");
-const Uri = std.Uri;
-
 const Page = @import("../page.zig").Page;
 const URL = @import("../url/url.zig").URL;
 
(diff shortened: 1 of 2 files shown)


src/browser/html/location.zig
```diff
@@ -16,7 +16,8 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-const Uri = @import("std").Uri;
+const std = @import("std");
+const Uri = std.Uri;
```
nikneym: `Uri` seem to be not used.

src/browser/html/location.zig
```diff
@@ -42,6 +43,15 @@ pub const Location = struct {
         return page.navigateFromWebAPI(href, .{ .reason = .script }, .{ .push = null });
     }
 
+    pub fn set_hash(_: *const Location, hash: []const u8, page: *Page) !void {
+        const normalized_hash = if (hash[0] == '#')
```
nikneym: Maybe we can add boundary check before matching characters; just in case.

PR #1209, opened 2025-11-13
Dummy `WebGLRenderingContext`

nikneym:
This adds the bare minimum `WebGLRenderingContext` to execute various websites.

### DIFF
--- a/src/browser/canvas/WebGLRenderingContext.zig
+++ b/src/browser/canvas/WebGLRenderingContext.zig
@@ -61,7 +61,7 @@ pub const Extension = union(enum) {
     WEBGL_debug_shaders: void,
     WEBGL_depth_texture: void,
     WEBGL_draw_buffers: void,
-    WEBGL_lose_context: void,
+    WEBGL_lose_context: Type.WEBGL_lose_context,
     WEBGL_multi_draw: void,
     WEBGL_polygon_mode: void,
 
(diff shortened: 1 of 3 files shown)


src/browser/canvas/WebGLRenderingContext.zig
```diff
@@ -0,0 +1,140 @@
...
+    WEBGL_debug_renderer_info: Type.WEBGL_debug_renderer_info,
+    WEBGL_debug_shaders: void,
+    WEBGL_depth_texture: void,
+    WEBGL_draw_buffers: void,
+    WEBGL_lose_context: void,
+    WEBGL_multi_draw: void,
+    WEBGL_polygon_mode: void,
+
+    /// Reified enum type from the fields of this union.
+    const Kind = blk: {
```
karlseguin, line 69: Is this doing the same thing as `const Kind = std.meta.FieldEnum(Extension);` ?

src/browser/canvas/WebGLRenderingContext.zig
```diff
@@ -0,0 +1,140 @@
...
+    pub const Type = struct {
+        pub const WEBGL_debug_renderer_info = packed struct(u64) {
+            UNMASKED_RENDERER_WEBGL: u32 = 0,
+            UNMASKED_VENDOR_WEBGL: u32 = 0,
+        };
+    };
+};
+
+/// An array of supported WebGL extensions.
+const extension_array = blk: {
```
karlseguin: same as `const extension_array = std.meta.fieldNames(Extension.Kind)`  (I think)

src/browser/canvas/WebGLRenderingContext.zig
nikneym: Requires higher branch quota for evaluation when done like so, should I change it?

karlseguin: meh..sounds like that would slow down compilation (probably by not much, but I'd rather make it incrementally faster than incrementally slower).

karlseguin (approved):

PR #1213, opened 2025-11-14
Remove `_TYPED_ARRAY_ID_KLUDGE` hack

nikneym:
This replaces `_TYPED_ARRAY_ID_KLUDGE` usage with actual types we use for `TypedArray`.


src/browser/js/Context.zig
```diff
@@ -780,55 +780,55 @@ pub fn jsValueToZig(self: *Context, comptime named_function: NamedFunction, comp
...
+    return switch (T) {
+        js.Function => {
+            if (!js_value.isFunction()) {
+                return null;
+            }
 
-    if (comptime T == js.Object) {
+            return try self.createFunction(js_value);
+        },
+        // zig fmt: off
```
karlseguin, line 791: TIL

krichprollsch: same 👀

PR #1229, opened 2025-11-21
Port `Blob` to zigdom

nikneym:


src/browser/webapi/Blob.zig
```diff
@@ -0,0 +1,294 @@
...
+
+const std = @import("std");
+const Writer = std.Io.Writer;
+
+const js = @import("../../js/js.zig");
+const Page = @import("../../Page.zig");
+
+/// https://w3c.github.io/FileAPI/#blob-section
+/// https://developer.mozilla.org/en-US/docs/Web/API/Blob
+const Blob = @This();
```
karlseguin, line 27: I think you want to add a _type field.  When it's created directly, the `_type` will be `.generic`.

```zig
_type: Type,

const Type = union(enum) {
   generic,
   file: *@import("File.zig"),
};
```

(you can optionally add the `as` and `is` helpers, see Event.zig, if you think they're needed, but I guess, right now, they aren't).

src/browser/webapi/File.zig
```diff
@@ -0,0 +1,49 @@
...
+const js = @import("../js/js.zig");
+
+const File = @This();
+
+/// `File` inherits `Blob`.
+_proto: *Blob,
+
+// TODO: Implement File API.
+pub fn init() File {
+    return .{ ._proto = undefined };
```
karlseguin: You need to set the proto. If you see it unset in other examples, it's becaue they're going through specialized factory methods (e.g. factory.eventTarget(...)) which builds it. That's also an option, you can create a factory.blob(...), but if this is the only Blob sub-type, it doesn't matter..(but the blob has to be set either way)

src/browser/webapi/file/Blob.zig
```diff
@@ -0,0 +1,294 @@
...
+        };
+
+        return .{ .slice = slice[start..end], .mime = mime };
+    }
+
+    return .{ .slice = slice, .mime = mime };
+}
+
+/// Returns the size of the Blob in bytes.
+pub fn size(self: *const Blob) usize {
```
karlseguin: for accessors, like `size` and `type`, I've been using getSize/setSize and getType/setType. Not necessary, but I find it avoids variable conflicts..e.g. now you can't have a variable named `size` anywhere in the file.

nikneym:
> Don't understand why File.zig in in the root of webapi, but Blob.zig is under the file folder?

I was unsure about the folder structure, should everything related to File API go under directory `file`? If, for example, `File.zig` is entrypoint to the API, should it be at the top-level or under the respective folder?

src/browser/webapi/Blob.zig
nikneym: So the `Type` should contain the types inherit from this type in all cases, right?

src/browser/webapi/File.zig
nikneym: Bit confused here, should it be something like this perhaps?
```zig
return .{ .proto = page._factory.create(Blob{ .slice = slice, .mime = mime });
```

karlseguin:
> I was unsure about the folder structure, should everything related to File API go under directory `file`? If, for example, `File.zig` is entrypoint to the API, should it be at the top-level or under the respective folder?

There's no hard rule, but since Blob is the parent of File, I was surprised to see `File.zig` in the root, and `file/Blob.zig`. If anything, I would have expected the opposite, `Blob.zig` and `blob/File.zig` If it's just those two, I'd be tempted to stick them both in webapi/.

But if you think there's more, like ReadableStream and WriteableStream and ??  Then, we can have a file/ folder or whatever else you think is best. I'd probably stick them all in there though....having the child (File.zig) be UP a directory than it's parent (Blob.zig) is weird to me.

src/browser/webapi/Blob.zig
karlseguin: Yes.

And the way this is allocated, they each have to be allocated separately (which is part of what the Factory does). So when you allocate a File, you also have to allocate the Blob. It's a tradeoff largely between performance and memory size...though I'm hoping  at some point we could optimize the allocator to reduce the performance cost we're paying)

src/browser/webapi/File.zig
karlseguin: Blob might be a bit unique 'cuz there's only 1 child?  But you could keep it the same, in which case you'd create a helper in Factory.zig:

```zig
pub fn blob(self: *Factory, child: anytype) !*@TypeOf(child) {
    const child_ptr = try self.createT(@TypeOf(child));
    child_ptr.* = child;

    const b = try self.createT(Blob);
    child_ptr._proto = b;
    b.* = .{ ._type = unionInit(Blob.Type, child_ptr) };
    return child_ptr;
}
```

which you'd then use here as: `return page._factory.blob(File{_proto = undefined});`

The `blob` function creates the `Blob` and fills in the `_proto`.  But you'll need to tweak things a bit since it looks like your Blob can't be initialized with default values. So maybe you also have to pass the `slice` and `mime` to the factory.

(Or, if File is the only child of blob, you can just do)

```zig
return page._factory.create(File{
   ._proto = try page._factory.create(Blob{....}),
});
```

(Using the factory here has limited purposes, but I'm hoping we can one day do something fancier with it)

PR #1232, opened 2025-11-24
element.prefix

krichprollsch:


src/browser/webapi/Element.zig
```diff
@@ -64,6 +64,7 @@ pub const Namespace = enum(u8) {
 _type: Type,
 _proto: *Node,
 _namespace: Namespace = .html,
+_namespace_prefix: ?[]const u8 = null,
```
karlseguin: This field adds 16 bytes to every element. If you take an Amazon product page that has ~7500 elements (`document.querySelectorAll('*').length`), it isn't a trivial amount. If you expect this to be null in most cases, consider what we do for element.styles, element.dataset, element.shadowRoot and element.classList - store them in the page.

So in Page, you could do:

```zig
// existing element data lookup
_element_shadow_roots: Element.ShadowRootLookup = .{},
_element_namespace_prefixes: Element.NamespacePrefixLookup = .{}
```

And then use the lookup to store/get the per-element prefix. There's a balance with this approach. But if
1 - We expect element.prefix not to be called too often, and
2 - Expect element.prefix to be null most of the time

The overhead (both in terms of performance and memory) of the lookup should result in less memory usage with little performance impact.

src/browser/parser/Parser.zig
```diff
@@ -188,6 +188,10 @@ fn createElementCallback(ctx: *anyopaque, data: *anyopaque, qname: h5e.QualName,
 }
 fn _createElementCallback(self: *Parser, data: *anyopaque, qname: h5e.QualName, attributes: h5e.AttributeIterator) !*anyopaque {
     const page = self.page;
+    // (PT) According with https://docs.rs/markup5ever/latest/markup5ever/interface/struct.QualName.html
```
karlseguin: I think html5ever is correct here. If you open the element.html test in chrome, you'll see what html5ever is saying: prefix is null and the tag name is `CUSTOM:DIV`. That's correct for an HTML document.

I think we need to deduce if the document is HTML or XML and, if XML, use a dedicated XML parser. Do we need to parse XML?

src/browser/Page.zig
```diff
@@ -907,29 +907,41 @@ pub fn createElement(self: *Page, ns_: ?[]const u8, name: []const u8, attribute_
         break :blk .html;
     };
 
+    var name = name_;
+    var ns_prefix_: ?[]const u8 = null;
+    if (std.mem.indexOfPos(u8, name_, 0, ":")) |pos| {
```
karlseguin: indexOfScalarPos with a `':'` avoids 1 length check...but, see above comment about whether this is correct for an HTML document.

src/browser/parser/Parser.zig
```diff
@@ -190,7 +190,13 @@ fn _createElementCallback(self: *Parser, data: *anyopaque, qname: h5e.QualName,
     const page = self.page;
     const name = qname.local.slice();
     const namespace = qname.ns.slice();
-    const node = try page.createElement(namespace, name, attributes);
+    const ns_prefix = blk: {
+        if (qname.prefix.unwrap()) |s| {
+            break :blk try page.dupeString(s.slice());
```
krichprollsch: I'm duplicating the string here instead of into `createElement` or `createHtmlElementT`. Not sure if it's the right place.

src/browser/webapi/Element.zig
krichprollsch: updated

src/browser/parser/Parser.zig
krichprollsch: ok, I had a confusion between localname, tagname and prefix.

image

krichprollsch: So I can use the actual `name` to get the `prefix` and the `localName` directly.

src/browser/parser/Parser.zig
karlseguin: I don't think it matters too much now, although nothing else in Parser does this. The main difference is really that multiple places call `page.createElement`. If we had a 2nd place that also passed the prefix and it also needed the page to own the string, doing it in page.createElement might make more sense.

Thinking more about it, we also now have a short string optimization (string.zig). Your prefix could be a `string.String` rather than a `[]const u8`. If you expect short prefixes (<= 12 characters), using that might be worth it (no allocation). In that case, it probably makes sense to pass the non-duped value into page.createElement..the details of using `[]const u8` or `string.String` don't really belong in the Parser.

krichprollsch: Thanks for the precision.
According with https://github.com/lightpanda-io/browser/pull/1232#discussion_r2559019717, I will finally avoid storing anything but using only `name` to return prefix and `localName`.

krichprollsch:
@karlseguin another try :)

PR #1233, opened 2025-11-24
`zigdom` Factory Slab Allocator

mookums:
This introduces a `SlabAllocator` that can be used by the Factory to do basically the same thing, managing reusable memory pools depending on size and alignment that are backed by our page arena. This one is entirely dynamic, directing allocation requests to distinct slabs.

From my fairly limited testing, there was a minimal memory usage increase (48kB when running a memory test like the one we do in CI) and the `make tests` actually had an improvement performance wise, likely due to the batched chunk allocations in each slab.

### DIFF
--- a/src/browser/Factory.zig
+++ b/src/browser/Factory.zig
@@ -24,7 +24,7 @@ const IS_DEBUG = builtin.mode == .Debug;
 const log = @import("../log.zig");
 const String = @import("../string.zig").String;
 
-const SlabAllocator = @import("../slab.zig").SlabAllocator(16);
+const SlabAllocator = @import("../slab.zig").SlabAllocator;
 
 const Page = @import("Page.zig");
 const Node = @import("webapi/Node.zig");
@@ -53,7 +53,7 @@ _slab: SlabAllocator,
 pub fn init(page: *Page) Factory {
     return .{
         ._page = page,
-        ._slab = SlabAllocator.init(page.arena),
+        ._slab = SlabAllocator.init(page.arena, 128),
     };
 }
 
--- a/src/slab.zig
+++ b/src/slab.zig
@@ -4,394 +4,415 @@ const assert = std.debug.assert;
 const Allocator = std.mem.Allocator;
 const Alignment = std.mem.Alignment;
 
-pub fn SlabAllocator(comptime slot_count: usize) type {
-    comptime assert(std.math.isPowerOfTwo(slot_count));
+const Slab = struct {
+    alignment: Alignment,
+    item_size: usize,
+    max_slot_count: usize,
 
-    const SlabKey = struct {
-        size: usize,
-        alignment: Alignment,
-    };
+    bitset: std.bit_set.DynamicBitSetUnmanaged,
+    chunks: std.ArrayListUnmanaged([]u8),
 
-    const Slab = struct {
-        const Slab = @This();
-        const chunk_shift = std.math.log2_int(usize, slot_count);
-        const chunk_mask = slot_count - 1;
-
+    pub fn init(
+        allocator: Allocator,
         alignment: Alignment,
         item_size: usize,
+        max_slot_count: usize,
+    ) !Slab {
+        return .{
+            .alignment = alignment,
+            .item_size = item_size,
+            .bitset = try .initFull(allocator, 0),
+            .chunks = .empty,
+            .max_slot_count = max_slot_count,
+        };
+    }
 
-        bitset: std.bit_set.DynamicBitSetUnmanaged,
-        chunks: std.ArrayListUnmanaged([]u8),
+    pub fn deinit(self: *Slab, allocator: Allocator) void {
+        self.bitset.deinit(allocator);
 
-        pub fn init(
-            allocator: Allocator,
-            alignment: Alignment,
-            item_size: usize,
-        ) !Slab {
-            return .{
-                .alignment = alignment,
-                .item_size = item_size,
-                .bitset = try .initFull(allocator, 0),
-                .chunks = .empty,
-            };
+        for (self.chunks.items) |chunk| {
+            allocator.rawFree(chunk, self.alignment, @returnAddress());
         }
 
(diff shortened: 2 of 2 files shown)


karlseguin:
I like this. Pretty nice maintainability upgrade and could help performance too by pre-heating allocations. I think it's worth merging as-is, but..there are 3 things I think we should think about:

1 - A few items are large, probably aren't too numerous, and aren't likely to share a size with anything else. Window immediately comes to mind, but so does XMLHttpRequest. We should potentially special case these types...why allocate 16*@sizeOf(Window) when we only need 1?

2 - At some point, it would be nice to get some metrics out of this, might give us some insight. Low priority though.

3 - I was wondering if it made sense to special-case html -> element -> node -> EventTarget. Currently, every html element requires those 4 allocations, and they're all going to be spread out (the slab probably makes this worse). I was thinking instead of 4 allocations, we could make 1  allocation`@sizeOf(Html) + @sizeOf(Element) + @sizeOf(Node) + @sizeOf(EventTarget)`, and then parcel out the memory to each component. Slab can stay as is, and the Factory could handle that. Thoughts? (could be worth doing for text -> cdata -> node -> eventtarget which is also quite common)

mookums:
Added some slab statistics just to help with development. This slab (at a slot count of 16) generally has a fairly low fragmentation rate, usually less than 5% of the allocated memory is considered not utilized. There are some slabs that are consistently cold, like (size 96, alignment 3) which I assume is like Window or these other objects that we only create one of.

This is pretty good regardless but the problem is that the higher we move the slot count (which reduces the number of allocations), the higher our fragmentation becomes. Ideally though, we'd be able to have larger slot counts because the hot paths (especially size 16, alignment 3) could really benefit from having larger batched allocations to work with.

I've updated the `SlabAllocator` and this new version uses dynamic chunk sizing (exponential with a cap) to keep cold slabs small but to allow hot ones to take increasing large slices of memory, generally amortizing our allocations.

I decided to try this out on a fairly reasonable real-world example, the old reddit link we use for integration tests. Below are the results with the SlabAllocator with a `slot_count` of 16, this generally results in the best fragmentation with anything above this wasting more memory.

```
=== Slab Allocator Statistics ===
Overall Memory:
  Total allocated: 514176 bytes (0.49 MB)
  In use:          495472 bytes (0.47 MB)
  Free:            18704 bytes (0.02 MB)

Overall Structure:
  Slab Count:    16
  Total chunks:    853
  Total slots:     13648
  Slots in use:    13497
  Slots free:      151

Overall Efficiency:
  Utilization:     96.4%
  Fragmentation:   3.6%

Per-Slab Breakdown:
   Size | Algn | Chunks |  Slots |  InUse |      Bytes |  Util%
  ------+------+--------+--------+--------+------------+-------
      8 |    3 |      1 |     16 |      8 |        128 |  50.0%
     80 |    3 |      1 |     16 |      1 |       1280 |   6.3%
     56 |    3 |    148 |   2368 |   2357 |     132608 |  99.5%
     16 |    3 |    248 |   3968 |   3954 |      63488 |  99.6%
     48 |    3 |      3 |     48 |     43 |       2304 |  89.6%
    232 |    3 |      1 |     16 |      1 |       3712 |   6.3%
     32 |    3 |     96 |   1536 |   1528 |      49152 |  99.5%
     40 |    3 |    149 |   2384 |   2380 |      95360 |  99.8%
     48 |    4 |    183 |   2928 |   2914 |     140544 |  99.5%
    128 |    3 |      1 |     16 |     14 |       2048 |  87.5%
     96 |    4 |      1 |     16 |     16 |       1536 | 100.0%
     24 |    3 |      1 |     16 |      8 |        384 |  50.0%
     64 |    4 |      1 |     16 |     10 |       1024 |  62.5%
    648 |    3 |      1 |     16 |      2 |      10368 |  12.5%
     32 |    4 |     17 |    272 |    261 |       8704 |  96.0%
     96 |    3 |      1 |     16 |      0 |       1536 |   0.0%
```

So you can see that we do a total of 853 allocations (each chunk in total chunks represents an allocation) with a fragmentation of 3.6%  You can also see how we have certain pools that perform considerably worse than others. 

```
=== Slab (Exponential) Allocator Statistics ===
Overall Memory:
  Total allocated: 511200 bytes (0.49 MB)
  In use:          495472 bytes (0.47 MB)
  Free:            15728 bytes (0.01 MB)

Overall Structure:
  Slab Count:    16
  Total chunks:    175
  Total slots:     13850
  Slots in use:    13497
  Slots free:      353

Overall Efficiency:
  Utilization:     96.9%
  Fragmentation:   3.1%

Per-Slab Breakdown:
   Size | Algn | Chunks |  Slots |  InUse |      Bytes |  Util%
  ------+------+--------+--------+--------+------------+-------
      8 |    3 |      4 |     15 |      8 |        120 |  53.3%
     80 |    3 |      1 |      1 |      1 |         80 | 100.0%
     56 |    3 |     25 |   2431 |   2357 |     136136 |  97.0%
     16 |    3 |     37 |   3967 |   3954 |      63472 |  99.7%
     48 |    3 |      6 |     63 |     43 |       3024 |  68.3%
    232 |    3 |      1 |      1 |      1 |        232 | 100.0%
     32 |    3 |     18 |   1535 |   1528 |      49120 |  99.5%
     40 |    3 |     25 |   2431 |   2380 |      97240 |  97.9%
     48 |    4 |     29 |   2943 |   2914 |     141264 |  99.0%
    128 |    3 |      4 |     15 |     14 |       1920 |  93.3%
     96 |    4 |      5 |     31 |     16 |       2976 |  51.6%
     24 |    3 |      4 |     15 |      8 |        360 |  53.3%
     64 |    4 |      4 |     15 |     10 |        960 |  66.7%
    648 |    3 |      2 |      3 |      2 |       1944 |  66.7%
     32 |    4 |      9 |    383 |    261 |      12256 |  68.1%
     96 |    3 |      1 |      1 |      0 |         96 |   0.0%
```

With this new version, we can crank the `max_slot_count` up to 128. This results in only 20% of the allocations that the previous version made while having comparable memory fragmentation.

PR #1234, opened 2025-11-25
Add a build step for `html5ever` in `build.zig`

nikneym:
This PR removes these `Makefile` rules:
```
make install-html5ever
make install-html5ever-dev
```

And introduces these `build.zig` steps:
```
zig build html5ever
zig build html5ever -Doptimize=[enum]
```

Optimize options build different targets of html5ever:
* `ReleaseFast`, `ReleaseSmall` => `release` target
* `Debug`, `ReleaseSafe` => `debug` target

## Why?
Currently we have 2 build systems in the project; `build.zig` and `Makefile`. This separates how we build dependencies and creates the need for 2 different optimize options (`-Doptimize` in `build.zig` and `*-dev` suffix in `Makefile`). We also prefer different locations for build artifacts depending on build system whereas `build.zig` already offer `zig-out/` which we can use.

I think the obvious way forward is to stick with `build.zig` since its more declarative and, well, we need `build.zig` to build the project anyway.

In fact, at some point (if I can get this PR merged 👀), we can have a single step to build the whole project for dev and prod. Something like:
```
zig build lightpanda -Doptimize=[enum]
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 lightpanda.id
 /v8/
 /build/
+/src/html5ever/target/


karlseguin:
I'm generally a fan of this...BUT...I've been exploring the exact opposite for our curl dependency. Newer versions of curl (and some of its dependencies) changed significantly, and the build got more complicated - in particular mbedtls. mbedtls now requires python, with specific dependencies, and some of the source files are now generated as part of the build script. I thought using bash to execute their build script was a lot easier.

Would you create a venv, install the `requirements.txt` and so on from build.zig ?

nikneym:
> Would you create a venv, install the requirements.txt and so on from build.zig ?

Well, that might not be the most ideal and make things extra harder to maintain. I haven't checked what are the requirements for curl but I think we can try some other TLS library (AFAIK curl supports multiple) if there're no specific benefits regarding to MbedTLS.

karlseguin:
We're not really vendoring those files. Those are our bindings.

PR #1238, opened 2025-11-26
Prototype Chain Allocations

mookums:
This explores the idea of allocating a prototype chain as a single element and then splitting the memory internally, allowing for us to create things like `Element` as a single allocation instead of 4. This does remove the optimization of storing the leaf/terminal nodes inline as structs but it provides us with better allocation behavior in exchange.

This method is faster than the previous multi-allocation chain but how much better depends on the webpage and what actions it is taking. We lose the ability to reuse sub-elements in the allocator like reusing an EventTarget for another object, since each slab is sized based on the prototype chain of top level element it is creating.

This is the previous multi-allocation Factory:
```
=== Slab Allocator Statistics ===
Overall Memory:
  Total allocated: 42032 bytes (0.04 MB)
  In use:          29424 bytes (0.03 MB)
  Free:            12608 bytes (0.01 MB)

Overall Structure:
  Slab Count:    12
  Total chunks:    52
  Total slots:     1174
  Slots in use:    817
  Slots free:      357

Overall Efficiency:
  Utilization:     70.0%
  Fragmentation:   30.0%

Per-Slab Breakdown:
   Size | Algn | Chunks |  Slots |  InUse |      Bytes |  Util%
  ------+------+--------+--------+--------+------------+-------
      8 |    3 |      2 |      3 |      2 |         24 |  66.7%
     80 |    3 |      1 |      1 |      1 |         80 | 100.0%
     56 |    3 |      8 |    255 |    189 |      14280 |  74.1%
     16 |    3 |      9 |    383 |    272 |       6128 |  71.0%
     32 |    3 |      7 |    127 |     74 |       4064 |  58.3%
    216 |    3 |      1 |      1 |      1 |        216 | 100.0%
     48 |    3 |      2 |      3 |      2 |        144 |  66.7%
     40 |    3 |      8 |    255 |    188 |      10200 |  73.7%
     48 |    4 |      7 |    127 |     76 |       6096 |  59.8%
     32 |    4 |      4 |     15 |      9 |        480 |  60.0%
    128 |    3 |      1 |      1 |      1 |        128 | 100.0%
     64 |    4 |      2 |      3 |      2 |        192 |  66.7%
```

This is the new chained Factory:
```=== Slab Allocator Statistics ===
Overall Memory:
  Total allocated: 40176 bytes (0.04 MB)
  In use:          33960 bytes (0.03 MB)
  Free:            6216 bytes (0.01 MB)

Overall Structure:
  Slab Count:    13
  Total chunks:    40
  Total slots:     469
  Slots in use:    353
  Slots free:      116

Overall Efficiency:
  Utilization:     84.5%
  Fragmentation:   15.5%

Per-Slab Breakdown:
   Size | Algn | Chunks |  Slots |  InUse |      Bytes |  Util%
  ------+------+--------+--------+--------+------------+-------
    184 |    3 |      1 |      1 |      1 |        184 | 100.0%
     32 |    3 |      1 |      1 |      1 |         32 | 100.0%
    256 |    3 |      1 |      1 |      1 |        256 | 100.0%
     48 |    3 |      1 |      1 |      1 |         48 | 100.0%
      8 |    3 |      1 |      1 |      1 |          8 | 100.0%
    168 |    3 |      6 |     63 |     62 |      10584 |  98.4%
     16 |    3 |      7 |    127 |     82 |       2032 |  64.6%
    136 |    3 |      7 |    127 |    115 |      17272 |  90.6%
     48 |    4 |      7 |    127 |     76 |       6096 |  59.8%
    192 |    4 |      4 |     15 |      9 |       2880 |  60.0%
    208 |    3 |      1 |      1 |      1 |        208 | 100.0%
    288 |    3 |      1 |      1 |      1 |        288 | 100.0
```

I also ran hyperfine against the demo page we have up. 
```
➜ hyperfine --warmup 3 '../lightpanda-old fetch https://demo-browser.lightpanda.io/campfire-commerce' '../lightpanda-new fetch https://demo-browser.lightpanda.io/campfire-commerce'
Benchmark 1: ../lightpanda-old fetch https://demo-browser.lightpanda.io/campfire-commerce
  Time (mean ± σ):     462.6 ms ±   8.1 ms    [User: 63.6 ms, System: 16.4 ms]
  Range (min … max):   445.4 ms … 473.9 ms    10 runs

Benchmark 2: ../lightpanda-new fetch https://demo-browser.lightpanda.io/campfire-commerce
  Time (mean ± σ):     458.8 ms ±  10.7 ms    [User: 66.7 ms, System: 14.2 ms]
  Range (min … max):   441.1 ms … 474.8 ms    10 runs

Summary
  ../lightpanda-new fetch https://demo-browser.lightpanda.io/campfire-commerce ran
    1.01 ± 0.03 times faster than ../lightpanda-old fetch https://demo-browser.lightpanda.io/campfire-commerce
```
The performance benefit seems to be directly related with how many operations we do as I've seen page with heavier JS perform up to 10% better.

TODO: Fix the SVG element

### DIFF
--- a/src/browser/Factory.zig
+++ b/src/browser/Factory.zig
@@ -241,17 +241,22 @@ pub fn htmlElement(self: *Factory, child: anytype) !*@TypeOf(child) {
 
 pub fn svgElement(self: *Factory, tag_name: []const u8, child: anytype) !*@TypeOf(child) {
     const allocator = self._slab.allocator();
+    const ChildT = @TypeOf(child);
 
-    // will never allocate, can't fail
-    const tag_name_str = String.init(self._page.arena, tag_name, .{}) catch unreachable;
+    if (ChildT == Element.Svg) {
+        return self.element(child);
+    }
 
     const chain = try PrototypeChain(
-        &.{ EventTarget, Node, Element, Element.Svg, @TypeOf(child) },
+        &.{ EventTarget, Node, Element, Element.Svg, ChildT },
     ).allocate(allocator);
 
     chain.setRoot(EventTarget.Type);
     chain.setMiddle(1, Node.Type);
     chain.setMiddle(2, Element.Type);
+
+    // will never allocate, can't fail
+    const tag_name_str = String.init(self._page.arena, tag_name, .{}) catch unreachable;
 
     // Manually set Element.Svg with the tag_name
     chain.set(3, .{
--- a/src/browser/Page.zig
+++ b/src/browser/Page.zig
@@ -1177,22 +1177,21 @@ pub fn createElement(self: *Page, ns_: ?[]const u8, name: []const u8, attribute_
         else => {},
     }
 
-    // TODO: uncomment
-    // if (namespace == .svg) {
-    //     const tag_name = try String.init(self.arena, name, .{});
-    //     if (std.ascii.eqlIgnoreCase(name, "svg")) {
-    //         return self.createSvgElementT(Element.Svg, name, attribute_iterator, .{
-    //             ._proto = undefined,
-    //             ._type = .svg,
-    //             ._tag_name = tag_name,
-    //         });
-    //     }
+    if (namespace == .svg) {
+        const tag_name = try String.init(self.arena, name, .{});
+        if (std.ascii.eqlIgnoreCase(name, "svg")) {
+            return self.createSvgElementT(Element.Svg, name, attribute_iterator, .{
+                ._proto = undefined,
+                ._type = .svg,
+                ._tag_name = tag_name,
+            });
+        }
 
-    //     // Other SVG elements (rect, circle, text, g, etc.)
-    //     const lower = std.ascii.lowerString(&self.buf, name);
-    //     const tag = std.meta.stringToEnum(Element.Tag, lower) orelse .unknown;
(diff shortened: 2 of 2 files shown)


src/browser/webapi/element/Html.zig
```diff
@@ -67,36 +67,36 @@ pub fn construct(page: *Page) !*Element {
 }
 
 pub const Type = union(enum) {
-    anchor: Anchor,
-    body: Body,
-    br: BR,
-    button: Button,
+    anchor: *Anchor,
```
karlseguin, line 70: This change is fine. The goal was to remove an unnecessary allocation and keep everything close together, but with your change, it's still 1 allocation instead of 3 for these...and they're still grouped together, so it's still an improvement for these cases.

karlseguin:
Fancy.

I think some of these were only ever used internally, e.g. I don't think `factory.element` is used now (i know, it shouldn't have been public in the first place then).

In other places, we use `(@hasField(S, "_type") and @typeInfo(@TypeOf(value._type)) == .@"union"` to determine if something is a root or not.  I don't mind having a more explicit field, but I don't like having both.

I don't think it's an issue, but just FYI, some "roots" can also be stand-alone objects. E.g. `Event.zig` can both be a root but it can also exist  as-is (in that case, `_type = .generic` which has a `void` type). This is something the previous cleanup code had to be aware of.

PR #1242, opened 2025-12-03
add `insertAdjacentHTML` and other variants

nikneym:
This PR ports `insertAdjacentHTML` and adds `insertAdjacentText` and `insertAdjacentElement` APIs to zigdom.

### DIFF
--- a/src/browser/tests/document/insert_adjacent_text.html
+++ b/src/browser/tests/document/insert_adjacent_text.html
@@ -1,0 +1,49 @@
+<!DOCTYPE html>
+<head id="the_head">
+  <title>Test Document Title</title>
+  <script src="../testing.js"></script>
+</head>
+
+<body>
+  <!-- This structure will get mutated by insertAdjacentText test -->
+  <div id="insert-adjacent-text-outer-wrapper">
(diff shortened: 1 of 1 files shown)


src/browser/webapi/Element.zig
```diff
@@ -360,6 +360,65 @@ pub fn attachShadow(self: *Element, mode_str: []const u8, page: *Page) !*ShadowR
...
+
+    const maybe_body_node = html_node.lastChild();
+    std.debug.assert(maybe_body_node != null);
+    const body = maybe_body_node orelse return;
+
+    const target_node, const prev_node = try self.asNode().findAdjacentNodes(position);
+
+    var iter = body.childrenIterator();
+    while (iter.next()) |child_node| {
+        _ = try target_node.insertBefore(child_node, prev_node, page);
```
karlseguin, line 397: There's probably some optimization opportunity by using `page.insertNodeRelative` directly. You'd be able to call `page.domChanged();` just once (before the loop), and you could skip the call to `page.removeNode` (which checks mutationobservers), since you know the children are all newly created (would need to manually set their `_parent = null` before calling). There might be other small tweaks.  Not sure if this API is called enough to warrant it though.

PR #1243, opened 2025-12-03
Add a synchronous signal handler for graceful shutdown

mrdimidium:
Signals in Unix can be asynchronous (interrupting one of the threads and calling the `sighandle` callback) or synchronous (blocking the process on the sigwait call).

Asynchronous signals seem easier to use, but they have a problem: since the system switches to sigaction at an arbitrary moment, the process can be stopped at an arbitrary point (for example, while acquiring a mutex or writing to a global buffer). Therefore, posix defines a specific [list of functions](https://man7.org/linux/man-pages/man7/signal-safety.7.html) that are safe to use in a handler. Using any other code (even the simplest ones like print) results in UB.

The code below circumvents this problem using blocking signal handling:
- set a mask as early as possible to prevent receiving signals on the main thread (this property will be inherited when creating new threads);
- create a new thread that specifies that it can receive SIGINT/SIGTERM;
- block the new thread on sigwait, and call cleanup callbacks when a signal is received.

Since the thread is blocked on sigwait, it consumes no resources (except a few MB for its stack). If the program terminates, the thread will be killed when the main thread terminates.

Using atomic to get shutdown is necessary regardless of the thread I create here, posix states that the signal sent to the process can be handled on any of the threads.

### DIFF
--- a/src/sighandler.zig
+++ b/src/sighandler.zig
@@ -1,0 +1,88 @@
+//! This structure processes operating system signals (SIGINT, SIGTERM)
+//! and runs callbacks to clean up the system gracefully.
+//!
+//! The structure does not clear the memory allocated in the arena,
+//! clear the entire arena when exiting the program.
+const std = @import("std");
+const assert = std.debug.assert;
+const Allocator = std.mem.Allocator;
+
+const log = @import("log.zig");
+
+pub const SigHandler = struct {
+    arena: Allocator,
+
+    sigset: std.posix.sigset_t = undefined,
+    handle_thread: ?std.Thread = null,
+
+    attempt: u32 = 0,
+    listeners: std.ArrayList(Listener) = .empty,
+
+    pub const Listener = struct {
+        args: []const u8,
+        start: *const fn (context: *const anyopaque) void,
+    };
+
+    pub fn install(self: *SigHandler) !void {
+        // Block SIGINT and SIGTERM for the current thread and all created from it
+        self.sigset = std.posix.sigemptyset();
+        std.posix.sigaddset(&self.sigset, std.posix.SIG.INT);
+        std.posix.sigaddset(&self.sigset, std.posix.SIG.TERM);
+        std.posix.sigaddset(&self.sigset, std.posix.SIG.QUIT);
+        std.posix.sigprocmask(std.posix.SIG.BLOCK, &self.sigset, null);
+
+        self.handle_thread = try std.Thread.spawn(.{ .allocator = self.arena }, SigHandler.sighandle, .{self});
+        self.handle_thread.?.detach();
+    }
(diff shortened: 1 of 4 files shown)


src/sighandler.zig
```diff
@@ -0,0 +1,86 @@
+//! This structure processes operating system signals (SIGINT, SIGTERM)
+//! and runs callbacks to clean up the system gracefully.
+//!
+//! The structure does not clear the memory allocated in the arena,
+//! clear the entire arena when exiting the program.
+const std = @import("std");
+const assert = std.debug.assert;
+const Allocator = std.mem.Allocator;
+
+const Self = @This();
```
karlseguin: We don't use Self for normal struct, only for generics.

src/sighandler.zig
```diff
@@ -0,0 +1,86 @@
...
+
+    self.handle_thread = try std.Thread.spawn(.{ .allocator = self.arena }, Self.sighandle, .{self});
+    self.handle_thread.?.detach();
+}
+
+pub fn on(self: *Self, func: anytype, args: std.meta.ArgsTuple(@TypeOf(func))) !void {
+    assert(@typeInfo(@TypeOf(func)).@"fn".return_type.? == void);
+
+    const Args = @TypeOf(args);
+    const TypeErased = struct {
```
karlseguin: This is fancy, and it's fine. But it would also be ok to have a `shutdown` function in App/Server/whatever that takes an `*anyopaque` and knows to cast it back to its own type. I guess you wouldn't need an allocation anymore.

mrdimidium: This trick is only needed for type checking. You should copy the pointer to anyopaque anyway, so you won't waste any extra memory with `.{server}`. Since memory is allocated in the arena, the allocation here is almost equivalent to a pointer next to a pointer to a function.

krichprollsch:
Does it work on macos? I tested it on my linux and in docker, it looks good to me in these cases.

mrdimidium:
@krichprollsch, sorry, I didn't have MacOS at hand, I'll set it up this evening and check.

krichprollsch:
np, we can also wait for a test from @karlseguin or @nikneym

nikneym:
Just tested on MacOS, seems I have to run `kill` 2 times after to properly exit.

mrdimidium:
Interesting, this is actually a problem on macos, it doesn't interrupt accept after calling posix.shutdown, it needs an explicit posix.close.

mrdimidium:
Sorry for the delay, I fixed it for Mac. The issue was related to different shutdown behavior on Linux and Mac (the handler itself works fine), so I left a comment in the code.

Screenshot 2025-12-04 at 19 18 32

PR #1246, opened 2025-12-05
Support `isEqualNode`

nikneym:
This is the bare minimum required for a specific website, I'll add support for other variants once needed.

### DIFF
--- a/src/browser/webapi/Element.zig
+++ b/src/browser/webapi/Element.zig
@@ -117,6 +117,49 @@ pub fn className(self: *const Element) []const u8 {
     };
 }
 
+pub fn attributesEql(self: *const Element, other: *Element) bool {
+    if (self._attributes) |attr_list| {
+        const other_list = other._attributes orelse return false;
+        return attr_list.eql(other_list);
+    }
+    // Make sure no attrs in both sides.
+    return other._attributes == null;
+}
+
+/// TODO: localName and prefix comparison.
+pub fn isEqualNode(self: *Element, other: *Element) bool {
+    const self_tag = self.getTagNameDump();
+    const other_tag = other.getTagNameDump();
+    // Compare namespaces and tags.
+    const dirty = self._namespace != other._namespace or !std.mem.eql(u8, self_tag, other_tag);
+    if (dirty) {
+        return false;
+    }
+
+    // Compare attributes.
+    if (!self.attributesEql(other)) {
+        return false;
+    }
+
+    // Compare children.
+    var self_iter = self.asNode().childrenIterator();
+    var other_iter = other.asNode().childrenIterator();
+    var self_count: usize = 0;
+    var other_count: usize = 0;
+    while (self_iter.next()) |self_node| : (self_count += 1) {
+        const other_node = other_iter.next() orelse return false;
+        other_count += 1;
+        if (self_node.isEqualNode(other_node)) {
+            continue;
+        }
+
+        return false;
+    }
+
+    // Make sure both have equal number of children.
+    return self_count == other_count;
+}
+
 pub fn getTagNameLower(self: *const Element) []const u8 {
     switch (self._type) {
         .html => |he| switch (he._type) {
--- a/src/browser/webapi/Node.zig
+++ b/src/browser/webapi/Node.zig
@@ -286,6 +286,27 @@ pub fn getNodeType(self: *const Node) u8 {
     };
 }
 
+pub fn isEqualNode(self: *Node, other: *Node) bool {
+    // Make sure types match.
+    if (self.getNodeType() != other.getNodeType()) {
+        return false;
+    }
+
+    // TODO: Compare `localName` and prefix.
+    return switch (self._type) {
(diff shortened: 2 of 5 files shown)


src/browser/webapi/Node.zig
```diff
@@ -227,6 +227,81 @@ pub fn nodeType(self: *const Node) u8 {
...
 }
 
+pub fn isEqualNode(self: *Node, other: *Node, page: *Page) bool {
+    // Make sure types match.
+    if (self.nodeType() != other.nodeType()) {
+        return false;
+    }
+
+    // TODO: Compare `localName` and prefix.
+    switch (self._type) {
```
karlseguin: Could push this logic into each type, e.g. 

```zig
if (self.as(Element).isEqualNode(other.as(Element)) == false) {
  return false;
}
```

src/browser/webapi/Node.zig
```diff
@@ -227,6 +227,81 @@ pub fn nodeType(self: *const Node) u8 {
...
+        .element => {
+            const e1 = self.as(Element);
+            const e2 = other.as(Element);
+
+            const e1_tag = e1.getTagNameSpec(&page.buf);
```
karlseguin: I think you can use `getTagNameDump()` instead, which doesn't require a buffer

src/browser/webapi/Node.zig
```diff
@@ -227,6 +227,81 @@ pub fn nodeType(self: *const Node) u8 {
...
+                iter2_count += 1;
+
+                if (attrib1._name.eql(attrib2._name) and attrib1._value.eql(attrib2._value)) {
+                    continue;
+                }
+
+                return false;
+            }
+
+            if (e1._namespace != e2._namespace or !std.mem.eql(u8, e1_tag, e2_tag) or iter1_count != iter2_count) {
```
karlseguin: Both more efficient and easier to read if you group these a little better.  Check the tags names and namespaces where you get them, before iterating through all the attributes and comparing htem.

src/browser/webapi/Node.zig
```diff
@@ -227,6 +227,81 @@ pub fn nodeType(self: *const Node) u8 {
...
+            const a2 = other.as(Element.Attribute);
+            return std.mem.eql(u8, a1.getName(), a2.getName()) and std.mem.eql(u8, a1.getValue(), a2.getValue());
+        },
+        .cdata => {
+            const c1 = self.as(CData);
+            const c2 = other.as(CData);
+
+            return std.mem.eql(u8, c1.getData(), c2.getData());
+        },
+        else => @panic("NIY"),
```
karlseguin:
```
log.warn(.browser, "not implemented", .{
    .type = self._type,
    .feature = "Node.isEqualNode",
});
return false;
```

src/browser/webapi/Node.zig
```diff
@@ -227,6 +227,81 @@ pub fn nodeType(self: *const Node) u8 {
...
+    }
+
+    // `Node.childNodes` allocate memory, this seems like a better option.
```
karlseguin:
```suggestion
```

karlseguin: Yes, childrenIterator is meant for internal iteration within the Zig code, specifically to avoid the overhead of childNodes necessary to fulfill the webapi of that type.

src/browser/webapi/Node.zig
nikneym: I thought the same but was unsure to do, will make this change.

src/browser/webapi/element/Attribute.zig
```diff
@@ -96,6 +100,8 @@ pub const JsApi = struct {
     pub const value = bridge.accessor(Attribute.getValue, null, .{});
     pub const namespaceURI = bridge.accessor(Attribute.getNamespaceURI, null, .{});
     pub const ownerElement = bridge.accessor(Attribute.getOwnerElement, null, .{});
+
+    pub const isEqualNode = bridge.function(Attribute.isEqualNode, .{});
```
karlseguin: You don't need these. I guess it's probably a bit more efficient (though I don't really know how v8 resolves this)..but it isn't typically how we do it. The implementation on Node which dispatches into the type is good enough.

src/browser/webapi/Element.zig
```diff
@@ -117,6 +117,57 @@ pub fn className(self: *const Element) []const u8 {
...
+        return false;
+    }
+
+    // Compare attributes.
+    {
+        var self_iter = self.attributeIterator();
+        var other_iter = other.attributeIterator();
+        var self_count: usize = 0;
+        var other_count: usize = 0;
+        while (self_iter.next()) |a1| : (self_count += 1) {
```
karlseguin: Probably can't assume they're in order.

karlseguin:
Tests?

src/browser/webapi/element/Attribute.zig
nikneym: I was after the little performance diff but you're right, using parent's method is more conventional.

PR #1248, opened 2025-12-06
textContent and innerText adjustements

krichprollsch:
This PR splits `textContent` and `innerText`.

`textContent` must return the raw content of the data but it must ignore comments

`innertText` must return rendered text: it must ignore `style`, `script` content, replace `br` with `\n`, trim white spaces.
:warning: trimming is not correct is this PR, more work is need to improve from here.

### DIFF
--- a/src/browser/tests/element/inner.html
+++ b/src/browser/tests/element/inner.html
@@ -172,5 +172,5 @@
 
   // TODO innerText is not rendered correctly for now.
   //testing.expectEqual("This is a\ntext", d2.innerText);
-  testing.expectEqual("  This is a \n  text", d2.innerText);
+  testing.expectEqual(" This is a \n text ", d2.innerText);
 </script>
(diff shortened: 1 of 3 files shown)


src/browser/webapi/Node.zig
```diff
@@ -171,7 +171,20 @@ pub fn childNodes(self: *const Node, page: *Page) !*collections.ChildNodes {
 
 pub fn getTextContent(self: *Node, writer: *std.Io.Writer) error{WriteFailed}!void {
     switch (self._type) {
-        .element => |el| return el.getInnerText(writer),
+        .element => {
+            var it = self.childrenIterator();
+            while (it.next()) |child| {
+                // ignore comments and TODO processing instructions.
+                switch (child._type) {
```
karlseguin: As an alternative:

```zig
if (child.is(CData.Comment) != null) {
   // ignore comments
   continue;
}
```

PR #1249, opened 2025-12-06
fix comment formatting

krichprollsch:


karlseguin:
FYI, `format` is only used internally for debugging

```zig
std.debug.print("what are you: {f}", .{node});
```

PR #1138, opened 2025-10-10
add `Navigation` WebAPI

mookums:
This is still pretty rough around the edges for now (and is incomplete)

This implements the `Navigation` WebAPI. `History` is now a compatibility layer over `Navigation`, allowing them to share the same underlying entries. `testing` has been changed, using `fetchWait` and loading a secondary testing html file that checks for the appropriate conditions when we need to load a secondary page (such as with `History` or `Navigation`. the `navigateFromWebAPI` has also been changed, adding a field, that allows it so serve as the central way to navigate with proper hooks into `Navigation`.

There are a few Events that need to be completed so far, such as `NavigateEvent`, and a lot of conditions that need extra testing (such as reloading).

---

Looking to get this merged soon, just to prevent this from drifting off too far. There is still a good chunk of work left to be fully WPT compliant (such as firing events at the proper time), but a lot of these WPTs depend on other browser features (like `pageshow`) that have to be implemented separately. I'm going to continue working on Navigation, specifically on getting most of the tests to pass.

PR #1253, opened 2025-12-08
Backport Navigation (and friends)

mookums:
This backports `Navigation` mostly and all of the related changes that I made across various PRs.

This backports a number of PRs (including various with zigdom_backport labels)
- https://github.com/lightpanda-io/browser/pull/1086
- https://github.com/lightpanda-io/browser/pull/1138
- https://github.com/lightpanda-io/browser/pull/1182
- https://github.com/lightpanda-io/browser/pull/1192
- https://github.com/lightpanda-io/browser/pull/1200
- https://github.com/lightpanda-io/browser/pull/1206
- https://github.com/lightpanda-io/browser/pull/1207
- https://github.com/lightpanda-io/browser/pull/1215

This also brings over the URL stitch (now resolve) fix and the regression test that was solved by https://github.com/lightpanda-io/browser/pull/1113.

### DIFF
--- a/src/browser/URL.zig
+++ b/src/browser/URL.zig
@@ -272,14 +272,10 @@ pub fn getHost(raw: [:0]const u8) []const u8 {
 
 // Returns true if these two URLs point to the same document.
 pub fn eqlDocument(first: [:0]const u8, second: [:0]const u8) bool {
-    if (!std.mem.eql(u8, getProtocol(first), getProtocol(second))) return false;
-    if (!std.mem.eql(u8, getHost(first), getHost(second))) return false;
-    if (!std.mem.eql(u8, getPort(first), getPort(second))) return false;
-    if (!std.mem.eql(u8, getPathname(first), getPathname(second))) return false;
-    if (!std.mem.eql(u8, getSearch(first), getSearch(second))) return false;
-    // hashes are allowed to be different.
-
-    return true;
+    // First '#' signifies the start of the fragment.
+    const first_hash_index = std.mem.indexOfScalar(u8, first, '#') orelse first.len;
+    const second_hash_index = std.mem.indexOfScalar(u8, second, '#') orelse second.len;
+    return std.mem.eql(u8, first[0..first_hash_index], second[0..second_hash_index]);
 }
 
 const KnownProtocol = enum {


src/browser/webapi/event/NavigationCurrentEntryChangeEvent.zig
```diff
@@ -0,0 +1,79 @@
+// Copyright (C) 2023-2024  Lightpanda (Selecy SAS)
```
karlseguin:
```suggestion
// Copyright (C) 2023-2025  Lightpanda (Selecy SAS)
```

src/browser/URL.zig
```diff
@@ -268,6 +270,18 @@ pub fn getHost(raw: [:0]const u8) []const u8 {
...
 }
 
+// Returns true if these two URLs point to the same document.
+pub fn eqlDocument(first: [:0]const u8, second: [:0]const u8) bool {
+    if (!std.mem.eql(u8, getProtocol(first), getProtocol(second))) return false;
+    if (!std.mem.eql(u8, getHost(first), getHost(second))) return false;
+    if (!std.mem.eql(u8, getPort(first), getPort(second))) return false;
+    if (!std.mem.eql(u8, getPathname(first), getPathname(second))) return false;
+    if (!std.mem.eql(u8, getSearch(first), getSearch(second))) return false;
+    // hashes are allowed to be different.
```
karlseguin: Tempted to look for '#' and slice them based on that?  Seems simple enough, unless `#` can be in the url, unencoded, outside of the hash ?

src/browser/webapi/event/PageTransitionEvent.zig
```diff
@@ -0,0 +1,63 @@
...
+
+pub fn asEvent(self: *PageTransitionEvent) *Event {
+    return self._proto;
+}
+
+pub fn getPersisted(self: *PageTransitionEvent) bool {
+    return self._persisted;
+}
+
+const PageTransitionKind = enum { show, hide };
```
karlseguin: Not used?

src/browser/webapi/event/PopStateEvent.zig
```diff
@@ -0,0 +1,72 @@
...
+}
+
+pub fn getState(self: *PopStateEvent, page: *Page) !?js.Value {
+    if (self._state == null) return null;
+
+    const value = try js.Value.fromJson(page.js, self._state.?);
+    return value;
+}
+
+pub fn getUAVisualTransition(_: *PopStateEvent) bool {
```
karlseguin:
```suggestion
pub fn hasUAVisualTransition(_: *PopStateEvent) bool {
```

src/browser/webapi/navigation/NavigationHistoryEntry.zig
```diff
@@ -0,0 +1,106 @@
...
+//         NavigationHistoryEntryEventTarget{ ._proto = undefined },
+//     );
+// }
+
+pub fn id(self: *const NavigationHistoryEntry) []const u8 {
+    return self._id;
+}
+
+pub fn index(self: *const NavigationHistoryEntry, page: *Page) i32 {
+    const navigation = page._session.navigation;
```
karlseguin: I think this is a fat copy?

src/browser/webapi/Window.zig
```diff
@@ -24,6 +24,7 @@ const log = @import("../../log.zig");
 const Page = @import("../Page.zig");
 const Console = @import("Console.zig");
 const History = @import("History.zig");
+const Navigation = @import("../webapi/navigation/Navigation.zig");
```
karlseguin:
```suggestion
const Navigation = @import("navigation/Navigation.zig");
```

src/browser/Page.zig
```diff
@@ -393,6 +421,9 @@ pub fn documentIsComplete(self: *Page) void {
 fn _documentIsComplete(self: *Page) !void {
     self.document._ready_state = .complete;
 
+    self._session.browser.runMicrotasks();
```
karlseguin: Are you sure this is still necessary? It should currently be run after every script (inline, dynamic, module, doesn't matter) is done evaluating (in addition to other places). I don't think it's  possible for `documentIsComplete` to have been called without `runMicrotasks` called before.

src/browser/webapi/navigation/root.zig
```diff
@@ -0,0 +1,72 @@
...
+    reload,
+};
+
+pub const NavigationState = struct {
+    source: enum { history, navigation },
+    value: ?[]const u8,
+};
+
+// https://developer.mozilla.org/en-US/docs/Web/API/NavigationActivation
+pub const NavigationActivation = struct {
```
karlseguin: Doesn't seem used nor ported to zigdom-esque style.

src/browser/webapi/navigation/Navigation.zig
```diff
@@ -0,0 +1,410 @@
...
+const NavigationCurrentEntryChangeEvent = @import("../event/NavigationCurrentEntryChangeEvent.zig");
+const NavigationEventTarget = @import("NavigationEventTarget.zig");
+
+_proto: *NavigationEventTarget = undefined,
+_arena: std.mem.Allocator,
+_current_navigation_kind: ?NavigationKind = null,
+
+_index: usize = 0,
+// Need to be stable pointers, because Events can reference entries.
+_entries: std.ArrayListUnmanaged(*NavigationHistoryEntry) = .empty,
```
karlseguin:
```suggestion
_entries: std.ArrayList(*NavigationHistoryEntry) = .empty,
```

src/browser/webapi/navigation/Navigation.zig
```diff
@@ -0,0 +1,410 @@
...
+pub fn init(arena: std.mem.Allocator) Navigation {
+    return Navigation{ ._arena = arena };
+}
+
+fn asEventTarget(self: *Navigation) *EventTarget {
+    return self._proto.asEventTarget();
+}
+
+pub fn onRemovePage(self: *Navigation) void {
+    self._proto = undefined;
```
karlseguin, line 59: Wasn't expecting that...Is there a reason for having these `onRemovePage` and `onNewPage` at all? Can the Navigation instance not be created once (in the Session)...nothing here references the `*Page` or anything tied to the page lifetime (e.g. window) as far as I can see.

src/browser/webapi/navigation/Navigation.zig
```diff
@@ -0,0 +1,410 @@
...
+pub fn getCanGoBack(self: *const Navigation) bool {
+    return self._index > 0;
+}
+
+pub fn getCanGoForward(self: *const Navigation) bool {
+    return self._entries.items.len > self._index + 1;
+}
+
+pub fn getCurrentEntry(self: *Navigation) *NavigationHistoryEntry {
+    return self._entries.items[self._index];
```
karlseguin: Pretty sure this can crash if called when empty (and `self._index == 0`)

src/browser/webapi/navigation/Navigation.zig
mookums: The `Navigation` has an EventTarget so with libdom, we had to create a new EventTarget whenever a new Page was created (or we'd get a segfault). With zigdom (from my understanding), since we create a new Factory whenever we get a new page, we need to create a new EventTarget for it whenever we create a new page.

src/browser/Page.zig
mookums: Yeah this doesn't seem to be needed anymore, just ported it over because it was an issue we previously had.

src/browser/webapi/navigation/Navigation.zig
mookums: It should be impossible for this to be called without an entry existing. Whenever the page loads, an entry should be created before we run any scripts on the page. I added an assert and a note about it.

src/browser/URL.zig
mookums: ya seems to be fine to just find first '#' and do a comparison. this was just directly ported from the previous version where we were using the other URL obj

src/browser/webapi/navigation/Navigation.zig
karlseguin: It can be created outside the factory, or the factory methods could take an `opts` that allows an explicit allocator to be passed in (though I don't love that approach, since it's an `if` check on every create).

karlseguin:
This looks good. Up to you if you think it's worth moving the Navigation instance to the session-owned arena so that we don't to reset between page navigation events - but I do think it would be cleaner. I'm trying to think if there are any implications to doing it (since nothing else does)...can't think of any.

mookums:
> This looks good. Up to you if you think it's worth moving the Navigation instance to the session-owned arena so that we don't to reset between page navigation events - but I do think it would be cleaner. I'm trying to think if there are any implications to doing it (since nothing else does)...can't think of any.

I'm happy keeping it where it is now, will probably make a PR in the near future regarding those `Factory` changes.

Issue #1239, opened 2025-11-28
How to get response headers


wh5938316:
The return value of `await page.goto` is null. If I want to get the response headers of that website's document, how should I do it?

``` typescript
const response = await page.goto(url, { waitUntil: 'load', timeout: timeout });

// echo `null`
console.log(response);

// I want to get headers
const headers = response.headers();
```

I'm sure it's not due to a timeout, because the subsequent `page.evaluate` will execute normally.

krichprollsch:
Hello @wh5938316, thanks for the issue.
What library are you using? playwright?

wh5938316:
I'm using puppeteer@^24.4.0; I haven't tried Playwright.

krichprollsch:
@wh5938316 just merged #1265 which should fix the issue.
It will be available in the nightly build tomorrow.

PR #1265, opened 2025-12-10
cdp: improve network's events

krichprollsch:
Working on #1239, I'm trying to figure out why Playwright returns a `null` response.

I did find yet, but I have added some changes in Network events.

### DIFF
--- a/src/browser/mime.zig
+++ b/src/browser/mime.zig
@@ -24,6 +24,7 @@ pub const Mime = struct {
     // IANA defines max. charset value length as 40.
     // We keep 41 for null-termination since HTML parser expects in this format.
     charset: [41]u8 = default_charset,
+    charset_len: usize = 5,
 
     /// String "UTF-8" continued by null characters.
     pub const default_charset = .{ 'U', 'T', 'F', '-', '8' } ++ .{0} ** 36;
@@ -53,7 +54,7 @@ pub const Mime = struct {
         other: struct { type: []const u8, sub_type: []const u8 },
     };
 
-    pub fn contentTypeString(mime: *const Mime) [:0]const u8 {
+    pub fn contentTypeString(mime: *const Mime) []const u8 {
         return switch (mime.content_type) {
             .text_xml => "text/xml",
             .text_html => "text/html",
@@ -66,8 +67,12 @@ pub const Mime = struct {
     }
 
     /// Returns the null-terminated charset value.
-    pub fn charsetString(mime: *const Mime) [:0]const u8 {
-        return @ptrCast(&mime.charset);
+    pub fn charsetStringZ(mime: *const Mime) [:0]const u8 {
+        return mime.charset[0..mime.charset_len :0];
+    }
+
+    pub fn charsetString(mime: *const Mime) []const u8 {
+        return mime.charset[0..mime.charset_len];
     }
 
     /// Removes quotes of value if quotes are given.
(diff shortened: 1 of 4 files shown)


krichprollsch:
I added an important change in e2682ab9fe8d318348c1584359aa34489d0aac6c
I now return the `Page.navigate` response **after** the navigation.

Claude mentioned that timing issue:
```
  Timing Issue

  Chrome sequence:
  1. Line 91: Page.navigate sent
  2. Line 96: Network.requestWillBeSent received ← BEFORE navigate returns
  3. Line 99: Network.responseReceived received
  4. Line 100: Page.navigate returns

  Lightpanda sequence:
  1. Line 71: Page.navigate sent
  2. Line 77: Page.navigate returns ← BEFORE requestWillBeSent!
  3. Line 82: Network.requestWillBeSent received ← TOO LATE
  4. Line 83: Network.responseReceived received
  
[...]

  3. Send Network.requestWillBeSent BEFORE Page.navigate returns
    - The CDP event must be dispatched to Playwright before the Page.navigate command response is sent
    - This allows Playwright to associate the pending request with the navigation's documentId
```

:warning: But it doesn't fix the actual issue, so I'm not sure if I should keep it...

krichprollsch:
TODO The final problem is Playwright expects `requestId` equals `loaderId`

```
   The requestId and loaderId still don't match!

  Looking at the Playwright code again in crNetworkManager.ts:348-349:

  const isNavigationRequest = requestWillBeSentEvent.requestId === requestWillBeSentEvent.loaderId &&
                              requestWillBeSentEvent.type === 'Document';
  const documentId = isNavigationRequest ? requestWillBeSentEvent.loaderId : undefined;

  For a request to be recognized as a navigation request, BOTH conditions must be true:
  1. ✅ type === 'Document' - You fixed this!
  2. ❌ requestId === loaderId - Still broken!

  Since "REQ-1" !== "LID-2", the request is NOT recognized as a navigation request, so documentId remains undefined, and the
  request is never associated with the navigation.

  Required Fix

  In line 76, change:
  "requestId": "REQ-1",
  "loaderId": "LID-2",

  To use the same value for both:
  "requestId": "LID-2",
  "loaderId": "LID-2",

  Or alternatively:
  "requestId": "REQ-1",
  "loaderId": "REQ-1",

  The key is: for the main document navigation request, requestId must equal loaderId.

  Compare with Chrome's log (line 96 of chrome.log):
  {
    "requestId": "0FEB712DC451FB8C15552FA23FAB3B16",
    "loaderId": "0FEB712DC451FB8C15552FA23FAB3B16"  // ← Same value!
  }

  This is the final piece needed for Playwright to return a non-null response from page.goto().
  ```

krichprollsch:
@karlseguin I added the change to make `loaderId` == `requestId` for all events relative to a given request.

If you want to review this PR again please :pray:

karlseguin:
PR looks fine.  Just to document past knowledge ,puppeteer relies on the loaderId to change in order to notice that a navigation has happened. Its `waitForNavigation` waits until the old loaderId != new loaderId.  Looking at this PR, it seems like the loaderId will still changing on such internal navigations, so it should be fine.

krichprollsch:
>  Looking at this PR, it seems like the loaderId will still changing on such internal navigations, so it should be fine.

Indeed, I kept the loader id generator for these cases.

PR #1270, opened 2025-12-11
CSS parsing

mrdimidium:


src/browser/css/Tokenizer.zig
```diff
@@ -0,0 +1,820 @@
...
+const Tokenizer = @This();
+
+pub const Token = union(enum) {
+    /// A `<ident-token>`
+    Ident: []const u8,
+
+    /// A `<function-token>`
+    ///
+    /// The value (name) does not include the `(` marker.
+    Function: []const u8,
```
karlseguin: We tend to follow Zig's naming guidelines, which has tag names as snake_case.

src/browser/css/Tokenizer.zig
```diff
@@ -0,0 +1,820 @@
...
+};
+
+input: []const u8,
+
+/// Counted in bytes, not code points. From 0.
+position: usize = 0,
+
+// If true, the input has at least `n` bytes left *after* the current one.
+// That is, `Lexer.byte_at(n)` will not panic.
+fn has_at_least(self: *const Tokenizer, n: usize) bool {
```
karlseguin: And function names are `camelCase`

src/browser/css/Tokenizer.zig
```diff
@@ -0,0 +1,820 @@
...
+}
+
+/// https://drafts.csswg.org/css-syntax/#consume-string-token
+fn consume_string(self: *Tokenizer, single_quote: bool) Token {
+    self.advance(1); // Skip the initial quote
+
+    // start_pos is at code point boundary, after " or '
+    const start_pos = self.position;
+
+    while (self.is_eof()) {
```
karlseguin: Is this right?  I think it shoud be `while (!self.is_eof()) {`    good indication of a missing test.

src/browser/css/Tokenizer.zig
```diff
@@ -0,0 +1,820 @@
...
+
+        if (self.has_at_least(2) and (self.byte_at(1) == '+' or self.byte_at(1) == '-') and std.ascii.isDigit(self.byte_at(2))) {
+            mul = switch (self.byte_at(1)) {
+                '-' => -1.0,
+                '+' => 1.0,
+                else => unreachable,
+            };
+
+            self.advance(2);
+        } else if (self.has_at_least(1) and std.ascii.isDigit(self.byte_at(2))) {
```
karlseguin: If we're checking `byte_at(2)`, should this be `self.has_at_least(2)` ?

src/browser/css/Tokenizer.zig
```diff
@@ -0,0 +1,820 @@
...
+        else => codepoint >= '\u{10000}',
+    };
+}
+
+fn is_ident_start(self: *Tokenizer) bool {
+    if (self.is_eof()) return false;
+
+    var b = self.next_byte_unchecked();
+    if (b == '-') {
+        b = self.next_byte() orelse return false;
```
karlseguin: Isn't `b` guaranteed to be `'-'` here?

src/browser/css/Tokenizer.zig
```diff
@@ -0,0 +1,820 @@
...
+    return .{ .Number = .{
+        .has_sign = has_sign,
+        .int_value = int_value,
+        .value = @as(f32, @floatCast(value)),
+    } };
+}
+
+fn consume_unquoted_url(self: *Tokenizer) ?Token {
+    _ = self;
+    @panic("unimplemented");
```
karlseguin: If this isn't something we plan on implementing now, we should log, not panic.

src/browser/css/Tokenizer.zig
```diff
@@ -0,0 +1,820 @@
...
+
+fn slice_from(self: *const Tokenizer, start_pos: usize) []const u8 {
+    return self.slice(start_pos, self.position);
+}
+
+// Advance over N bytes in the input.  This function can advance
+// over ASCII bytes (excluding newlines), or UTF-8 sequence
+// leaders (excluding leaders for 4-byte sequences).
+fn advance(self: *Tokenizer, n: usize) void {
+    if (builtin.mode == .Debug) {
```
karlseguin, line 231: I'm not sure this is super useful...It's not really adding checks in debug for **our code**, it's adding checks in debug for external code (css). If the assertion fails, what are we supposed to do about it?

src/browser/css/Tokenizer.zig
```diff
@@ -0,0 +1,820 @@
...
+fn is_ident_start(self: *Tokenizer) bool {
+    if (self.is_eof()) return false;
+
+    var b = self.next_byte_unchecked();
+    if (b == '-') {
+        b = self.next_byte() orelse return false;
+    }
+
+    return switch (b) {
+        'a'...'z', 'A'...'Z', '_', 0x0 => true,
```
karlseguin, line 290: is `0x0` really valid here (or anywhere else)

src/browser/css/Tokenizer.zig
mrdimidium: No, if `hasAtLeast(n)` is true, the input has at least `n` bytes left *after* the current one.

src/browser/css/Tokenizer.zig
mrdimidium: This is an invariant of the `advance` function; it should never be called for multiple lines (no matter how valid the CSS is). This is always a bug.

src/browser/css/Tokenizer.zig
mrdimidium: Not really, for example, when processing `# token` there will not be `-`

src/browser/css/Tokenizer.zig
mrdimidium: Yes, the specification says they can occur: https://www.w3.org/TR/css-syntax-3/#input-preprocessing.

The original idea of ​​the specification was to always work with code points (and replace 0x0 with �), but real parsers just explicitly work with bytes.

src/browser/css/Tokenizer.zig
karlseguin: So `nextByte() == nexByteUnchecked()` isn't always true? Neither advances the`position`, so I'm struggling to see how that's true?

karlseguin:
It hits the panic on existing tests,`zig build test`.  It also outputs debug statements.

PR #1271, opened 2025-12-11
Inherit Prototype Event Options

mookums:
This provides a mechanism for Events to inherit all of the options in their Event Prototype Chain. Previously, there was a lot of duplication with child Events storing fields such as bubbles. This just creates a options struct at compile time that includes all of the prototype options as well.


mookums:
CI is failing but should be fixed by #1272.

PR #1274, opened 2025-12-13
implement document.write

### LINKED ISSUE #1266: TypeError: document.write is not a function
Hi, thanks for implementing LightPanda which I'm hoping to be able to use with my web-accessibility proxy so that blind users will have the option of using text-mode browsers that don't support Javascript themselves.

However, I noticed that any attempt to use `document.write` (during page load) gives `TypeError: document.write is not a function` whereas it's one of the earliest DOM-modifying functions around (it predates innerHTML).  Is this just "not implemented yet" or was there a deliberate decision not to support it for some reason?

Thanks.

krichprollsch:
fix #1266

### DIFF
--- a/src/browser/dom/document.zig
+++ b/src/browser/dom/document.zig
@@ -341,9 +341,8 @@ pub const Document = struct {
     }
 
     pub fn _write(self: *parser.Document, str: []const u8, page: *Page) !void {
-        if (!page.open) {
-            _ = try _open(self, page);
-        }
+        _ = try _open(self, page);
+
         const document = parser.documentHTMLToDocument(page.window.document);
         const fragment = try parser.documentParseFragmentFromStr(document, str);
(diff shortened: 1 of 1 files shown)


src/browser/dom/document.zig
```diff
@@ -319,9 +319,68 @@ pub const Document = struct {
...
+
+        return page.window.document;
+    }
+
+    pub fn _close(_: *parser.Document, page: *Page) !void {
+        page.open = false;
+    }
+
+    pub fn _write(self: *parser.Document, str: []const u8, page: *Page) !void {
+        if (!page.open) {
```
karlseguin: I think you can call ` _ = try _open(self, page)` directly here (without the `if` check), since it does that `if` check for you. Seems better for `write` not to know about `page.open` if possible.

PR #1279, opened 2025-12-18
Enable v8 snapshots

karlseguin:
There are two layers here. The first is that, on startup, a v8 SnapshotCreator is created, and a snapshot-specific isolate/context is setup with our browser environment. This contains most of what was in `Env.init` and a good chunk of what was in `ExecutionWorld.createContext`. From this, we create a `v8.StartupData` which is used for the creation of all subsequent contexts. The snapshot sits at the application level, above the `Env` - it's re-used for all envs/isolates, so this gives a nice performance boost for both 1 connection opening multiple pages or multiple connections opening 1 page.

The second layer is that the Snapshot data can be embedded into the binary, so that it doesn't have to be created on startup, but rather created at build-time. This improves the startup time (though, I'm not really sure how to measure that accurately...).

The first layer is the big win (and just works as-is without any build / usage changes).

with snapshot
total runs 1000
total duration (ms) 7527
avg run duration (ms) 7
min run duration (ms) 5
max run duration (ms) 41

without snapshot
total runs 1000
total duration (ms) 9350
avg run duration (ms) 9
min run duration (ms) 8
max run duration (ms) 42

To embed a snapshot into the binary, we first need to create the snapshot file:

```
zig build -Doptimize=ReleaseFast snapshot_creator -- src/snapshot.bin
```

And then build using the new snapshot_path argument:

```
zig build -Dsnapshot_path=../../snapshot.bin -Doptimize=ReleaseFast
```

The paths are weird, I know...since it's embedded, it needs to be inside the project path, hence we put it in src/snapshot.bin. And since it's embedded relative to the embedder (src/browser/js/Snapshot.zig) the path has to be relative to that, hence ../../snapshot.bin. I'm open to suggestions on improving this.

### DIFF
--- a/src/browser/js/ExecutionWorld.zig
+++ b/src/browser/js/ExecutionWorld.zig
@@ -84,7 +84,7 @@ pub fn createContext(self: *ExecutionWorld, page: *Page, enter: bool) !*Context 
         if (comptime IS_DEBUG) {
             // Getting this into the snapshot is tricky (anything involving the
             // global is tricky). Easier to do here, and in debug more, we're
-            // find with paying the small perf hit.
+            // fine with paying the small perf hit.
             const js_global = v8.FunctionTemplate.initDefault(isolate);
             const global_template = js_global.getInstanceTemplate();
 


src/browser/js/Env.zig
```diff
@@ -88,50 +81,43 @@ pub fn init(allocator: Allocator, platform: *const Platform, _: Opts) !*Env {
 
     isolate.setHostInitializeImportMetaObjectCallback(Context.metaObjectCallback);
 
-    var temp_scope: v8.HandleScope = undefined;
-    v8.HandleScope.init(&temp_scope, isolate);
-    defer temp_scope.deinit();
+    // // Allocate templates array dynamically to avoid comptime dependency on JsApis.len
```
krichprollsch:
```suggestion
    // Allocate templates array dynamically to avoid comptime dependency on JsApis.len
```

src/browser/js/ExecutionWorld.zig
```diff
@@ -72,82 +70,32 @@ pub fn deinit(self: *ExecutionWorld) void {
...
-            if (@hasDecl(JsApi.Meta, "name")) {
-                const class_name = if (@hasDecl(JsApi.Meta, "constructor_alias")) JsApi.Meta.constructor_alias else JsApi.Meta.name;
-                const v8_class_name = v8.String.initUtf8(isolate, class_name);
-                global_template.set(v8_class_name.toName(), templates[i], v8.PropertyAttribute.None);
-            }
-        }
+        if (comptime IS_DEBUG) {
+            // Getting this into the snapshot is tricky (anything involving the
+            // global is tricky). Easier to do here, and in debug more, we're
+            // find with paying the small perf hit.
```
krichprollsch:
```suggestion
            // fine with paying the small perf hit.
```
?

src/browser/js/Snapshot.zig
```diff
@@ -0,0 +1,471 @@
...
+const v8 = js.v8;
+const JsApis = bridge.JsApis;
+const Allocator = std.mem.Allocator;
+
+const Snapshot = @This();
+
+const embedded_snapshot_blob = if (@import("build_config").snapshot_path) |path| @embedFile(path) else "";
+
+// When creating our Snapshot, we use local function templates for every Zig type.
+// You cannot, from what I can tell, create persisted FunctoinTemplates at
```
krichprollsch:
```suggestion
// You cannot, from what I can tell, create persisted FunctionTemplates at
```

src/browser/js/Snapshot.zig
```diff
@@ -0,0 +1,471 @@
...
+// snapshot creation time. But you can embedd those templates (or any other v8
+// Data) so that it's available to contexts created from the snapshot. This is
+// the starting index of those function templtes, which we can extract. At
```
krichprollsch:
```suggestion
// the starting index of those function templates, which we can extract. At
```

build.zig
```diff
@@ -29,10 +29,12 @@ pub fn build(b: *Build) !void {
 
     const git_commit = b.option([]const u8, "git_commit", "Current git commit");
     const prebuilt_v8_path = b.option([]const u8, "prebuilt_v8_path", "Path to prebuilt libc_v8.a");
+    const snapshot_path = b.option([]const u8, "snapshot_path", "Path to v8 snapshot");
 
     var opts = b.addOptions();
     opts.addOption([]const u8, "version", manifest.version);
     opts.addOption([]const u8, "git_commit", git_commit orelse "dev");
+    opts.addOption(?[]const u8, "snapshot_path", snapshot_path);
```
krichprollsch, line 37: AFAIK I need to pass an absolute path to make it work?

krichprollsch: oh I didn't red your comment my bad
```
zig build -Dsnapshot_path=../../snapshot.bin
```

PR #1282, opened 2025-12-19
Support `PerformanceObserver` API

nikneym:
This PR implements `PerformanceObserver` for `mark` and `measure` performance types. I'll add tests also but would like to receive reviews before that.

### DIFF
--- a/src/browser/tests/performance_observer/performance_observer.html
+++ b/src/browser/tests/performance_observer/performance_observer.html
@@ -1,0 +1,38 @@
+<!DOCTYPE html>
+<script src="../testing.js"></script>
+
+<script id="observe_performance_mark">
+{
+  const observer = new PerformanceObserver((list, observer) => {
+    testing.expectEqual(true, list instanceof PerformanceObserverEntryList);
+    testing.expectEqual(true, observer instanceof PerformanceObserver);
+
+    const entries = list.getEntries();
+    testing.expectEqual(true, entries instanceof Array);
+    testing.expectEqual(2, entries.length);
+
+    {
+      const { name, startTime, duration, entryType } = entries[0];
+      testing.expectEqual("operationStart", name);
+      testing.expectEqual(20, startTime);
+      testing.expectEqual(0, duration);
+      testing.expectEqual("mark", entryType);
+    }
+
+    {
+      const { name, startTime, duration, entryType } = entries[1];
+      testing.expectEqual("operationEnd", name);
+      testing.expectEqual(34.0, startTime);
+      testing.expectEqual(0, duration);
+      testing.expectEqual("mark", entryType);
+    }
+
+    observer.disconnect();
+  });
+
+  // Look for performance marks.
+  observer.observe({ type: "mark" });
+  performance.mark("operationStart", { startTime: 20.0 });
+  performance.mark("operationEnd", { startTime: 34.0 });
+}
+</script>


src/ring_buffer.zig
mookums, line 1: Doesn't seem like this file is used anywhere in this PR?

src/browser/Page.zig
```diff
@@ -51,6 +51,8 @@ const DocumentFragment = @import("webapi/DocumentFragment.zig");
 const ShadowRoot = @import("webapi/ShadowRoot.zig");
 const Performance = @import("webapi/Performance.zig");
 const Screen = @import("webapi/Screen.zig");
+const HtmlScript = @import("webapi/Element.zig").Html.Script;
```
karlseguin: Not used?   (Can always use `Element.Html.Script`)

src/browser/Page.zig
```diff
@@ -1046,6 +1054,54 @@ pub fn getElementByIdFromNode(self: *Page, node: *Node, id: []const u8) ?*Elemen
...
+    }
+
+    if (self._performance_delivery_scheduled == false) {
+        self._performance_delivery_scheduled = true;
+        try self.scheduler.add(
+            self,
+            struct {
+                fn run(_page: *anyopaque) anyerror!?u32 {
+                    const page: *Page = @ptrCast(@alignCast(_page));
+                    page._performance_delivery_scheduled = true;
```
karlseguin: This should be set to `false`?

src/browser/Page.zig
```diff
@@ -1046,6 +1054,54 @@ pub fn getElementByIdFromNode(self: *Page, node: *Node, id: []const u8) ?*Elemen
...
+            return;
+        }
+    }
+}
+
+/// Updates performance observers with the new entry.
+/// This doesn't emit callbacks but rather fills the queues of observers.
+pub fn notifyPerformanceObservers(self: *Page, entry: *Performance.Entry) !void {
+    for (self._performance_observers.items) |observer| {
+        if (observer.interested(entry)) {
```
karlseguin, line 1062: I might encapsulate this in the observer, so just:

```zig
observer.newEntry(entry)
```

and let it decide how/if to handle it.

src/browser/Page.zig
```diff
@@ -1046,6 +1054,54 @@ pub fn getElementByIdFromNode(self: *Page, node: *Node, id: []const u8) ?*Elemen
...
+        self._performance_delivery_scheduled = true;
+        try self.scheduler.add(
+            self,
+            struct {
+                fn run(_page: *anyopaque) anyerror!?u32 {
+                    const page: *Page = @ptrCast(@alignCast(_page));
+                    page._performance_delivery_scheduled = true;
+                    // Dispatch performance observer events.
+                    for (page._performance_observers.items) |observer| {
+                        if (observer.hasRecords()) {
```
karlseguin: again, maybe just call `observer.dispatch()` and let it deal with what (if anything) to do.

src/browser/webapi/PerformanceObserver.zig
```diff
@@ -60,13 +151,35 @@ pub const JsApi = struct {
...
+
+    pub const JsApi = struct {
+        pub const bridge = js.Bridge(EntryList);
+
+        pub const Meta = struct {
+            pub const name = "PerformanceEntryList";
+            pub const prototype_chain = bridge.prototypeChain();
+            pub var class_id: bridge.ClassId = undefined;
+
+            pub const getEntries = bridge.function(EntryList.getEntries, .{});
```
karlseguin: This should be defined in JsApi, not Meta. This tells me there might be some tests missing.

PR #1290, opened 2025-12-24
Zigdom request interception

karlseguin:
Zigdom broke request interception. It isn't zigdom specifically, but in zigdom
we properly block the parser when executing a normal (not async, not defer)
script. This does not work well with request interception, because an
intercepted request isn't blocked on HTTP data, it's blocked on a message from
CDP. Generally, neither our Page nor ScriptManager are CDP-aware. And, even if
they were, it would be hard to break out of our parsing and return control to
the CDP server.

To fix this, we expand on the HTTP Client's basic awareness of CDP (via its
extra_socket field). The HTTP client is now able to block until an intercepted
request is continued/aborted/fulfilled. it does this by being able to ask the
CDP client to read/process data.

### DIFF
--- a/src/http/Client.zig
+++ b/src/http/Client.zig
@@ -756,7 +756,7 @@ pub const Request = struct {
     credentials: ?[:0]const u8 = null,
 
     // This is only relevant for intercepted requests. If a request is flagged
-    // as blocking AND is interecepted, then it'll be up to us to wait until
+    // as blocking AND is intercepted, then it'll be up to us to wait until
     // we receive a response to the interception. This probably isn't ideal,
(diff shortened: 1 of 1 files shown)


src/http/Client.zig
```diff
@@ -668,6 +755,13 @@ pub const Request = struct {
     resource_type: ResourceType,
     credentials: ?[:0]const u8 = null,
 
+    // This is only relevant for intercepted requests. If a request is flagged
+    // as blocking AND is interecepted, then it'll be up to us to wait until
```
krichprollsch:
```suggestion
    // as blocking AND is intercepted, then it'll be up to us to wait until
```

PR #1292, opened 2025-12-24
Run microtasks after each script execution

nikneym:
This don't change the behavior for async and deferred scripts.

### DIFF
--- a/src/browser/ScriptManager.zig
+++ b/src/browser/ScriptManager.zig
@@ -801,8 +801,13 @@ pub const Script = struct {
             log.debug(.browser, "executed script", .{ .src = url, .success = success, .on_load = script_element._on_load != null });
         }
 
-        // We should run microtasks even if script execution fails.
-        defer page.js.runMicrotasks();
+        defer {
+            // We should run microtasks even if script execution fails.
+            page.js.runMicrotasks();
+            _ = page.scheduler.run() catch |err| {
+                log.err(.page, "scheduler", .{ .err = err });
+            };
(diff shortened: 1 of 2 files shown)


src/browser/ScriptManager.zig
```diff
@@ -801,7 +801,11 @@ pub const Script = struct {
             log.debug(.browser, "executed script", .{ .src = url, .success = success, .on_load = script_element._on_load != null });
         }
 
-        defer page.tick();
```
karlseguin, line 806: This'll run microtasks, scheduled tasks, and microtasks again, is that right?

Maybe it should just be:

```zig
    _ = self.scheduler.run() catch |err| {
        log.err(.page, "scheduler", .{ .err = err });
    };
    self.js.runMicrotasks();
```

and remove page.tick.

nikneym: I think that makes more sense. I thought running macrotasks after a script execution is necessary.

src/browser/ScriptManager.zig
```diff
@@ -803,7 +801,8 @@ pub const Script = struct {
...
+        // We should run microtasks even if script execution fails.
+        defer page.js.runMicrotasks();
```
karlseguin:
```suggestion
defer {
    self.js.runMicrotasks();
    _ = self.scheduler.run() catch |err| {
        log.err(.page, "scheduler", .{ .err = err });
    };
} 
```

and you can remove the tick function from Page (I think)

Issue #1214, opened 2025-11-14
Context canceled for JS loading site


suntong:
And here's an example for chromedp:

```go
// The CSS selector for the element you must wait for
const selectorToWaitFor = "a" 

var htmlContent string
const targetURL = "https://example.com"

log.Printf("Navigating to %s, waiting for '%s', and retrieving source...", targetURL, selectorToWaitFor)

err := chromedp.Run(ctx,
    // 1. Navigate to the target URL
    chromedp.Navigate(targetURL),
    
    // 2. Add the waiting condition here
    // This pauses execution until the element is visible in the DOM
    chromedp.WaitVisible(selectorToWaitFor, chromedp.ByQuery),
    
    // 3. Retrieve the OuterHTML of the entire 'html' element
    chromedp.OuterHTML(`html`, &htmlContent, chromedp.ByQuery),
)
```

It works for example.com, but not another site (with the _**same**_ code) which uses JS to load its content (`targetURL = "https://so.gitee.com/?q=Vue%20go"`, `selectorToWaitFor = "a[aria-label=Next]"`) --

I'm using the latest github.com/chromedp/chromedp (v0.14.2), and

I set a generous overall timeout using context.WithTimeout (300 seconds) for the entire operation. However, during execution of specific actions (chromedp.WaitVisible), the program consistently fails with a "context canceled" error after approximately 7 seconds. The actual body returned is (for sure the element is not found):

```html
 <div class="d-flex justify-content-center align-middle" style="height: 80vh;"><div class="m-auto"><div class="spinner-grow spinner-grow-sm text-success" role="status"><span class="visually-hidden">Loading...</span></div> <div class="spinner-grow spinner-grow-sm text-success" role="status"><span class="visually-hidden">Loading...</span></div> <div class="spinner-grow spinner-grow-sm text-success" role="status"><span class="visually-hidden">Loading...</span></div></div></div></body></html>
```

The site is heavy JS based, the initial load takes 6~7 seconds. then the browser should run JS to finish loading.

Using `go-rod`, and waiting for the _**same**_ element, it can retrieve the whole JS loaded content.

The issue might be a lightpanda cloud browser issue, as I am using Google Chrome browser using `browser=chrome` parameter in the url for `go-rod`, yet when adding `browser=chrome` parameter in the url for chromedp, it failed with the following error (so I'm not sure where exactly the problem is):

```
Visiting 'https://example.com' (& wait for 'a') ...
[...]

PR #1293, opened 2025-12-24
Use V8 to parse JSON with fetch/xhr

krichprollsch:
Using Zig JSON parser + zig to js conversion led to invalid type for some values.

For example, value like `1765867200000` was converted into BigInt but JSON.parse generates Number.

This PR replaces the standard Zig JSON parser with the V8 JSON parser to be sure to get the correct types.

Relates with #1214

### DIFF
--- a/src/browser/fetch/Request.zig
+++ b/src/browser/fetch/Request.zig
@@ -258,8 +258,9 @@ pub fn _json(self: *Response, page: *Page) !js.Promise {
             log.info(.browser, "invalid json", .{ .err = e, .source = "Request" });
             return error.SyntaxError;
         };
+        const pvalue = try value.persist(page.js);
 
-        return page.js.resolvePromise(value);
+        return page.js.resolvePromise(pvalue);
     }
     return page.js.resolvePromise(null);
 }
--- a/src/browser/fetch/Response.zig
+++ b/src/browser/fetch/Response.zig
@@ -183,8 +183,9 @@ pub fn _json(self: *Response, page: *Page) !js.Promise {
             log.info(.browser, "invalid json", .{ .err = e, .source = "Response" });
             return error.SyntaxError;
         };
+        const pvalue = try value.persist(page.js);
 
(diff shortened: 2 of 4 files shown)


krichprollsch:
Open question: in `js.zig` should we convert int64 into BigInt as currently, or should we generate Number instead?

krichprollsch:
I had to persist the v8.Value returned by fromJson to fix the test. I added another list into context to track these persisted values and deinit them properly.
I'm not sure if I did that correctly and if I should use existing persisted object list instead :thinking: 
see beef458c3c8df4448a7a99a879dc1fe37093f9e1

krichprollsch:
@mookums I saw History uses `fromJson`. I'm not sure if the js value returned should be persisted or not in this case. ie: should I have a `fromJsonPersistent` to keep local value w/ History, or my change could fix potential issues with History?

karlseguin:
> I had to persist the v8.Value returned by fromJson to fix the test. I added another list into context to track these persisted values and deinit them properly. I'm not sure if I did that correctly and if I should use existing persisted object list instead 🤔 see [beef458](https://github.com/lightpanda-io/browser/commit/beef458c3c8df4448a7a99a879dc1fe37093f9e1)

It's right as-is.

karlseguin:
> @mookums I saw History uses `fromJson`. I'm not sure if the js value returned should be persisted or not in this case. ie: should I have a `fromJsonPersistent` to keep local value w/ History, or my change could fix potential issues with History?

Looks like history always returns a transient value, always calling `toJson` on a value that's immediately returned to V8. there's `createPromiseResolver` in Context which is an example that takes  a `comptime` enum to control the persisted nature of the value. 

The other option is to take the function/object approach, and have an explicit `persist()` function on js.Value. So that you fromJson -> persist as 2 separate / explicit steps.  I think I prefer this, as it allows for values to be persisted in other cases.


Otherwise ,PR looks good.

PR #1295, opened 2025-12-24
Fix crashes with History/Navigation legacy tests

mookums:
This fixes the `History` and `Navigation` legacy tests in zigdom. This depends on #1294.


karlseguin:
This doesn't actually fix the crash, it just masks it.

You'll want to rebase so you grab https://github.com/lightpanda-io/browser/commit/f266dbc171b0a43f249ba50057d631da863de919

and then use `page.scheduleNavigation` rather than `page.navigate`. It takes a 3rd parameter, `priority`. You'll want to add a `history` and `navigate` entry to the `NavigationPriority` and update `canScheduleNavigation` accordingly.

Gemini says (which would make our handling of anchor clicks wrong, but we can fix that separately).

1 (Highest) | User-Initiated / Form | form.submit(), <a> clicks | High priority as they often involve data transmission (POST).
2 | Programmatic Navigation | location.href, location.assign() | Standard script-based URL changes.
3 | History API | history.pushState(), history.replaceState() | Synchronous URL updates that do not trigger a page reload.
4 (Special) | Navigation API | navigation.navigate() | Modern, promise-based, and can intercept or cancel other navigations.

PR #1300, opened 2025-12-29
backport set active html element on click

krichprollsch:
backport of #1231

### DIFF
--- a/src/browser/Page.zig
+++ b/src/browser/Page.zig
@@ -2472,8 +2472,8 @@ pub fn handleClick(self: *Page, target: *Node) !void {
                 return self.submitForm(element, button.getForm(self));
             }
         },
-        .select, .textarea => self.window._document._active_element = element,
-        else => {},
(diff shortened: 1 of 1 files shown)


src/browser/Page.zig
```diff
@@ -2462,8 +2472,8 @@ pub fn handleClick(self: *Page, target: *Node) !void {
                 return self.submitForm(element, button.getForm(self));
             }
         },
-        .select, .textarea => self.window._document._active_element = element,
-        else => {},
+        // by default set the focus on the clicked HTML element.
+        else => self.window._document._active_element = element,
```
karlseguin, line 2476: I don't think this is right. Only elements that can be focused, or those with a tabindex attribute can become active.

I found this answer for a list of focusable elements: https://stackoverflow.com/a/1600194

krichprollsch: Oh good catch, actually chromiumoxide clicks on the input text element.
I backported the changes w/o thinking...

PR #1305, opened 2025-12-29
prefer `DoublyLinkedList` for storing `MutationObserver`s in `Page`

nikneym:


src/browser/Page.zig
```diff
@@ -1025,17 +1025,12 @@ pub fn getElementByIdFromNode(self: *Page, node: *Node, id: []const u8) ?*Elemen
     return null;
 }
 
-pub fn registerMutationObserver(self: *Page, observer: *MutationObserver) !void {
-    try self._mutation_observers.append(self.arena, observer);
+pub inline fn registerMutationObserver(self: *Page, observer: *MutationObserver) !void {
```
karlseguin: I think it's worth avoiding `inline` unless we can prove it's worth it.

src/browser/webapi/MutationObserver.zig
```diff
@@ -53,6 +55,7 @@ pub const ObserveOptions = struct {
 pub fn init(callback: js.Function, page: *Page) !*MutationObserver {
     return page._factory.create(MutationObserver{
         ._callback = callback,
+        .node = .{},
```
karlseguin: kind of redundant with the default field value...we don't set any of the other fields with default.

PR #1244, opened 2025-12-04
cdp: add AXTree

krichprollsch:
fixes #1220 
relates with #1217

PR #1217, opened 2025-11-18
Stagehand compatibility

krichprollsch:
https://docs.stagehand.dev/

```js
import "dotenv/config";
import { Stagehand } from "@browserbasehq/stagehand";
import { z } from "zod/v3";

async function main() {
 const stagehand = new Stagehand({
    env: "LOCAL",
	localBrowserLaunchOptions: {
		cdpUrl: "ws://127.0.0.1:9222"
	},
    model: "ollama/qwen3:1.7b",
  });

  await stagehand.init();
  const page = stagehand.context.pages()[0];

  await page.goto("https://demo-browser.lightpanda.io/");

  // Act on the page
  await stagehand.act("Click the Campfire link")
  const price = await stagehand.extract("product price", z.string())

    console.log(price);

  await stagehand.close();
}

main().catch((err) => {
  console.error(err);
  process.exit(1);
});
```

PR #1308, opened 2025-12-30
cdp: add AXTree

### LINKED ISSUE #1220: Implement Accessibility tree
[Accessibility tree](https://developer.mozilla.org/en-US/docs/Glossary/Accessibility_tree) 

Create an accessibility tree based on the DOM tree, which is used by platform-specific Accessibility APIs to provide a representation that can be understood by assistive technologies, such as screen readers.

* [MDN Accessibility tree Glossary](https://developer.mozilla.org/en-US/docs/Glossary/Accessibility_tree)
* [Accessible Name and Description Computation spec](https://w3c.github.io/accname/)
* [CDP Accessibility.getFullAXTree](https://chromedevtools.github.io/devtools-protocol/tot/Accessibility/#method-getFullAXTree)

blocks #1217

krichprollsch:
Backport of #1244 for `zigdom`

fixes #1220
relates with #1217

### DIFF
--- a/src/cdp/AXNode.zig
+++ b/src/cdp/AXNode.zig
@@ -1,4 +1,4 @@
-// Copyright (C) 2023-2024  Lightpanda (Selecy SAS)
+// Copyright (C) 2023-2026  Lightpanda (Selecy SAS)
 //
 // Francis Bouvier <francis@lightpanda.io>
 // Pierre Tachoire <pierre@lightpanda.io>
@@ -60,8 +60,7 @@ pub const Writer = struct {
 
     fn writeNodeChildren(self: *const Writer, parent: AXNode, w: anytype) !void {
         // Add ListMarker for listitem elements
-        if (parent.dom.is(DOMNode.Element) != null) {
-            const parent_el = parent.dom.as(DOMNode.Element);
+        if (parent.dom.is(DOMNode.Element)) |parent_el| {
             if (parent_el.getTag() == .li) {
                 try self.writeListMarker(parent.dom, w);
             }
@@ -94,9 +93,7 @@ pub const Writer = struct {
     fn writeListMarker(self: *const Writer, li_node: *DOMNode, w: anytype) !void {
         // Find the parent list element
         const parent = li_node._parent orelse return;
-        if (parent.is(DOMNode.Element) == null) return;
-
-        const parent_el = parent.as(DOMNode.Element);
+        const parent_el = parent.is(DOMNode.Element) orelse return;
         const list_type = parent_el.getTag();
 
         // Only create markers for actual list elements
@@ -627,7 +624,7 @@ pub const AXRole = enum(u8) {
                         // zig fmt: off
                         .password, .@"datetime-local", .hidden, .month, .color,
                         .week, .time, .file, .date => .none,
-                        // zig fmt: ofn
+                        // zig fmt: on
                     };
                 },
                 .textarea => .textbox,
@@ -768,7 +765,6 @@ fn writeName(axnode: AXNode, w: anytype, page: *Page) !?AXSource {
 
                 // Parse space-separated list of IDs and concatenate their text content
                 var it = std.mem.splitScalar(u8, labelledby, ' ');
-                var first = true;
                 var has_content = false;
 
                 var buf = std.Io.Writer.Allocating.init(page.call_arena);
@@ -777,13 +773,9 @@ fn writeName(axnode: AXNode, w: anytype, page: *Page) !?AXSource {
                     if (trimmed_id.len == 0) continue;
 
                     if (doc.getElementById(trimmed_id, page)) |referenced_el| {
-                        if (!first) {
-                            try buf.writer.writeByte(' ');
-                        }
-                        first = false;
-
                         // Get the text content of the referenced element
                         try referenced_el.getInnerText(&buf.writer);
+                        try buf.writer.writeByte(' ');
                         has_content = true;
                     }
                 }
@@ -932,14 +924,12 @@ fn isIgnore(self: AXNode, page: *Page) bool {
     const role_attr = self.role_attr;
 
     // Don't ignore non-Element node: CData, Document...
-    if (node.is(DOMNode.Element) == null) {
-        return false;
-    }
+    const elt = node.is(DOMNode.Element) orelse return false;
     // Ignore non-HTML elements: svg...
-    if (node.is(DOMNode.Element.Html) == null) {
+    if (elt._type != .html) {
         return true;
     }
-    const elt = node.as(DOMNode.Element);
+
     const tag = elt.getTag();
     switch (tag) {
         // zig fmt: off


src/cdp/AXNode.zig
```diff
@@ -0,0 +1,864 @@
...
+
+    fn writeAXValue(self: *const Writer, value: AXValue, w: anytype) !void {
+        try w.beginObject();
+        try w.objectField("type");
+        try w.write(@tagName(value.type));
+
+        if (value.value) |v| {
+            try w.objectField("value");
+            switch (v) {
+                .uint => try w.write(v.uint),
```
karlseguin: I think this would work:

```zig
switch (v) {
  inline else => |v| v try w.write(v),
}
```

src/cdp/AXNode.zig
```diff
@@ -0,0 +1,864 @@
...
+            // error, we can only return a writer error. Kinda sucks.
+            log.err(.cdp, "node toJSON stringify", .{ .err = err });
+            return error.WriteFailed;
+        };
+    }
+
+    fn toJSON(self: *const Writer, node: *const Node, w: anytype) !void {
+        try w.beginArray();
+        const root = try AXNode.fromNode(node.dom);
+        if (try self.writeNode(node.id, root, w)) {
```
karlseguin, line 55: I would probably reverse this, so that writeNode returns `true` if the children should be written, so that you can just do:


```zig
if (try self.writeNode(node.id, root, w)) {
    try self.writeNodeChildren(root, w);
}
return w.endArray();
```

src/cdp/AXNode.zig
```diff
@@ -0,0 +1,864 @@
...
+                return error.InvalidTag;
+            },
+        };
+    }
+};
+
+dom: *DOMNode,
+role_attr: ?[]const u8,
+
+pub fn fromNode(dom: *DOMNode) !AXNode {
```
karlseguin: This can't return an error. If you remove the error union here, then you can remove it from `isIgnore`

src/cdp/AXNode.zig
```diff
@@ -0,0 +1,864 @@
...
+        value: AXValue,
+    };
+
+    fn writeAXProperties(self: *const Writer, axnode: AXNode, w: anytype) !void {
+        const dom_node = axnode.dom;
+        const page = self.page;
+        switch (dom_node._type) {
+            .document => |document| {
+                const uri = document.getURL(page);
+                try self.writeAXProperty(.{ .name = .url, .value = .{ .type = .string, .value = .{ .string = uri } } }, w);
```
karlseguin: Isn't the `type` field and union a redundant? Could you make it work with:

```zig
.{ .name = .url, .value = .{.string = uri } }
```

And use the std.meta.activeTag(value) in `writeAXValue` ? (renaming the `uint` tag to `integer`)

src/cdp/AXNode.zig
```diff
@@ -0,0 +1,864 @@
...
+        },
+        else => {
+            try w.write("");
+            return null;
+        },
+    };
+}
+
+fn isHidden(elt: *DOMNode.Element) bool {
+    if (elt.getAttributeSafe("aria-hidden")) |value| {
```
karlseguin, line 860: Outside this PR, we could potentially optimize checking/fetching multiple attributes. (I think there's a number of small optimizations we might be able to do for attributes, and this multi-attribute check use-case is worth thinking about)

krichprollsch:
@karlseguin I implemented your suggestions + added more support.
I think it's good (and big) anough to be merge if you are OK with it.

src/cdp/AXNode.zig
```diff
@@ -0,0 +1,1140 @@
+// Copyright (C) 2023-2024  Lightpanda (Selecy SAS)
```
karlseguin:
```suggestion
// Copyright (C) 2023-2026  Lightpanda (Selecy SAS)
```

src/cdp/AXNode.zig
```diff
@@ -0,0 +1,1140 @@
...
+        const root = AXNode.fromNode(node.dom);
+        if (try self.writeNode(node.id, root, w)) {
+            try self.writeNodeChildren(root, w);
+        }
+        return w.endArray();
+    }
+
+    fn writeNodeChildren(self: *const Writer, parent: AXNode, w: anytype) !void {
+        // Add ListMarker for listitem elements
+        if (parent.dom.is(DOMNode.Element) != null) {
```
karlseguin:
```zig
if (parent.dom.is(DOMNode.Element)) |parent_el| {
    if (parent_el.getTag() == .li) {
        try self.writeListMarker(parent.dom, w);
    }
}
```

src/cdp/AXNode.zig
```diff
@@ -0,0 +1,1140 @@
...
+            if (try self.writeNode(node.id, axn, w)) {
+                try self.writeNodeChildren(axn, w);
+            }
+        }
+    }
+
+    fn writeListMarker(self: *const Writer, li_node: *DOMNode, w: anytype) !void {
+        // Find the parent list element
+        const parent = li_node._parent orelse return;
+        if (parent.is(DOMNode.Element) == null) return;
```
karlseguin:
```
const parent_el = parent.is(DOMNode.Element) orelse return;
```

src/cdp/AXNode.zig
```diff
@@ -0,0 +1,1140 @@
...
+                var first = true;
+                var has_content = false;
+
+                var buf = std.Io.Writer.Allocating.init(page.call_arena);
+                while (it.next()) |id| {
+                    const trimmed_id = std.mem.trim(u8, id, &std.ascii.whitespace);
+                    if (trimmed_id.len == 0) continue;
+
+                    if (doc.getElementById(trimmed_id, page)) |referenced_el| {
+                        if (!first) {
```
karlseguin: If this is a hot-path, you can rewrite this to remove the check from the loop. On write, always append a trailing space. Then, outside the loop, if `has_content`, you can `buf.writer.undo(1)` to strip out the last trailing whitespace. I think.

src/cdp/AXNode.zig
```diff
@@ -0,0 +1,1140 @@
...
+        else => false,
+    };
+}
+
+fn isIgnore(self: AXNode, page: *Page) bool {
+    const node = self.dom;
+    const role_attr = self.role_attr;
+
+    // Don't ignore non-Element node: CData, Document...
+    if (node.is(DOMNode.Element) == null) {
```
karlseguin: Couple different ways to rewrite this to avoid the `is(DOMNode.Element)` and then a few lines later `as(DOMNode.Element)`:


```zig
const elt = node.is(DOMNode.Element) orelse return false;
if (elt._type != .html) {
   return true;
}
```

src/cdp/AXNode.zig
```diff
@@ -0,0 +1,1140 @@
...
+
+    const role_implicit = try AXRole.fromNode(self.dom);
+
+    return @tagName(role_implicit);
+}
+
+// Replace successives whitespaces with one withespace.
+// Trims left and right according to the options.
+// Returns true if the string ends with a trimmed whitespace.
+fn writeString(s: []const u8, w: anytype) !void {
```
karlseguin: I'm not 100% following but...are we not worried about possible double quotes in `s`?

PR #1309, opened 2026-01-03
Anchor Click

mookums:
This properly supports navigation on anchor tag clicks, using `scheduleNavigation` using the href value.

### DIFF
--- a/src/browser/Page.zig
+++ b/src/browser/Page.zig
@@ -2723,11 +2723,6 @@ pub fn handleClick(self: *Page, target: *Node) !void {
                 return;
             }
 
-            if (std.mem.startsWith(u8, href, "#")) {
-                // Hash-only links (#foo) should be handled as same-document navigation
-                return;
-            }
-
             if (std.mem.startsWith(u8, href, "javascript:")) {
                 return;
             }


karlseguin:
This is already handled in the Page.zig `handleClick` function. Is there a case where that wasn't working where this is? As-is, I think this code is a no-op, as the `handleClick` will execute after and last-navigate wins.  There's a `click.js` test in demo that's passing with zigdom.

mookums:
The `navigation-api/currententrychange-event/anchor-click.html`is dependent on the named access PR but it will fail unless we have the changes in this PR.

```html
<!doctype html>
<script src="/resources/testharness.js"></script>
<script src="/resources/testharnessreport.js"></script>
<a id="a" href="#foo"></a>
<script>
test(t => {
  let start_index = navigation.currentEntry.index;

  let oncurrententrychange_called = false;
  navigation.oncurrententrychange = t.step_func(e => {
    oncurrententrychange_called = true;
    assert_equals(e.from, navigation.entries()[start_index]);
    assert_equals(e.from.index, start_index);
    assert_equals(e.navigationType, "push");
    assert_equals(navigation.currentEntry.index, start_index + 1);
  });
  a.click();
  assert_true(oncurrententrychange_called);
}, "currententrychange fires for link click");
</script>
```

Maybe they use different click mechanisms?

karlseguin:
`handleClick` has this check:

```zig
if (std.mem.startsWith(u8, href, "#")) {
    // Hash-only links (#foo) should be handled as same-document navigation
    return;
}
```

I guess this guard should either be removed (in which case I guess it's handled by the `URL.eqlDocument` case of `scheduleNavigation` ?) or instead of `return` we should hook into the navigation directly.

Issue #1210, opened 2025-11-13
Named access on the `Window` object


mookums:
https://html.spec.whatwg.org/multipage/nav-history-apis.html#named-access-on-the-window-object

We need to support the ability to get elements directly using their ID. This is used in various Web Platform tests where we end up having something like:

```js
<iframe id="i"></iframe>

<script>
    i.onload = () => { console.log("loaded") }
</script>
```

This currently causes us to get a `ReferenceError` for `i` not existing.

mookums:
Completed with #1310.

PR #1310, opened 2026-01-03
Named Access on the Window Object

mookums:
This adds support for named access on the `Window` object, closing https://github.com/lightpanda-io/browser/issues/1210.

### DIFF
--- a/src/browser/js/ExecutionWorld.zig
+++ b/src/browser/js/ExecutionWorld.zig
@@ -88,7 +88,7 @@ pub fn createContext(self: *ExecutionWorld, page: *Page, enter: bool) !*Context 
         global_template.setNamedProperty(v8.NamedPropertyHandlerConfiguration{
             .getter = unknownPropertyCallback,
             .flags = v8.PropertyHandlerFlags.NonMasking | v8.PropertyHandlerFlags.OnlyInterceptStrings,
-        }, v8.External.init(isolate, page));
+        }, null);
 
         const context_local = v8.Context.init(isolate, global_template, null);
         const v8_context = v8.Persistent(v8.Context).init(isolate, context_local).castToContext();
@@ -157,10 +157,6 @@ pub fn resumeExecution(self: *const ExecutionWorld) void {
 pub fn unknownPropertyCallback(c_name: ?*const v8.C_Name, raw_info: ?*const v8.C_PropertyCallbackInfo) callconv(.c) u8 {
     const info = v8.PropertyCallbackInfo.initFromV8(raw_info);
 
-    const data_value = info.getData();
-    const external = data_value.castTo(v8.External);
-    const page: *Page = @ptrCast(@alignCast(external.get()));
-
     const context = Context.fromIsolate(info.getIsolate());
     const maybe_property: ?[]u8 = context.valueToString(.{ .handle = c_name.? }, .{}) catch null;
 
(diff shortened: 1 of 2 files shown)


karlseguin:
This fails:

```js
  const a = document.createElement('a');
  a.id = 'dune';
  document.body.appendChild(a);
  testing.expectEqual(a, window.dune);
  ```

Also, the named access should work for `name` attributes on certain element types; if multiple elements match, it ought to return an HTMLCollection (which is live). In firefox, it looks like it looks for `id` first, then fallsback to `name.

I feel like this is better implemented with a `namedIndex` on window which, in order:
1 - calls document.getElementById
2 - if (1) returns `null` then calls documnet.getElementsByName

**Comment by mookums** [comment]
Yeah, need to fix the actual management of the properties on id change.

I tried `namedIndex` but it didn't work for behavior like below (which is the more important one for a lot of the WPT tests):
```js
  const a = document.createElement('a');
  a.id = 'dune';
  document.body.appendChild(a);
  testing.expectEqual(a, dune);
```

karlseguin:
> I tried namedIndex but it didn't work  ...

I think that's what we have to fix; you'll have a hard time getting the live HTMLCollection working by setting it directly on the JS object.

I believe it's related to how we setup the unknown property callback on the global: https://github.com/lightpanda-io/browser/blob/25366f0e479dfa7a4959ee5e201177def715941c/src/browser/js/ExecutionWorld.zig#L90

Prior to zigdom, this was _always_ registered and how we did polyfills. In zigdom, as you can see, it's only enabled in debug builds as it's only used for logging unknown globals. Looks like we need to re-enable this and hook it into a window.unknwonGetter  or something which can apply the id/name getter (and if there are none, then log it in debug mode).

src/browser/js/ExecutionWorld.zig
```diff
@@ -159,9 +156,13 @@ pub fn resumeExecution(self: *const ExecutionWorld) void {
 
 pub fn unknownPropertyCallback(c_name: ?*const v8.C_Name, raw_info: ?*const v8.C_PropertyCallbackInfo) callconv(.c) u8 {
     const info = v8.PropertyCallbackInfo.initFromV8(raw_info);
-    const context = Context.fromIsolate(info.getIsolate());
 
-    const property = context.valueToString(.{ .handle = c_name.? }, .{}) catch "???";
+    const data_value = info.getData();
+    const external = data_value.castTo(v8.External);
+    const page: *Page = @ptrCast(@alignCast(external.get()));
```
karlseguin: You don't have to do this. Once you have the `*Context` (next line), you can access the page via `context.page`.

src/browser/js/Function.zig
```diff
@@ -115,6 +115,8 @@ pub fn tryCallWithThis(self: *const Function, comptime T: type, this: anytype, a
 
 pub fn callWithThis(self: *const Function, comptime T: type, this: anytype, args: anytype) !T {
     const context = self.context;
+    context.call_depth += 1;
```
karlseguin:
```suggestion
    // When we're calling a function from within JavaScript itself, this isn't necessary. We're within a Caller
    // instantiation, which will already have incremented the call_depth and it won't decrement it until the
    // Caller is done.
    // But some JS functions are initiated from Zig code, and not v8. For example, Observers, some events
    // and window callbacks. In those cases, we need to increase the call_depth so that the call_arena
    // remains valid for the duration of the function call. If we don't do this, the call_arena will be reset after
    // each statement of the function which executes Zig code.
    context.call_depth += 1;
```


I didn't understand why this was necessary [but it is]. Think the comment (or something like it) might help.

PR #1311, opened 2026-01-04
Backport dummy canvas APIs

nikneym:
This is missing [getContext](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/getContext) method since I'm not sure how custom element methods are implemented in zigdom.

### DIFF
--- a/src/browser/webapi/canvas/CanvasRenderingContext2D.zig
+++ b/src/browser/webapi/canvas/CanvasRenderingContext2D.zig
@@ -29,11 +29,11 @@ const Page = @import("../../Page.zig");
 const CanvasRenderingContext2D = @This();
 /// Fill color.
 /// TODO: Add support for `CanvasGradient` and `CanvasPattern`.
-fill_style: color.RGBA = color.RGBA.Named.black,
+_fill_style: color.RGBA = color.RGBA.Named.black,
 
 pub fn getFillStyle(self: *const CanvasRenderingContext2D, page: *Page) ![]const u8 {
     var w = std.Io.Writer.Allocating.init(page.call_arena);
-    try self.fill_style.format(&w.writer);
+    try self._fill_style.format(&w.writer);
     return w.written();
 }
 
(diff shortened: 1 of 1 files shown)


karlseguin:
I don't understand how this is related to custom elements?

nikneym:
> I don't understand how this is related to custom elements?/

Sorry for the misunderstanding, what I was trying to mean is to provide `getContext` function to `HTMLCanvasElement`.

karlseguin:
There's a `fn getContext()` in https://github.com/lightpanda-io/browser/blob/zigdom/src/browser/webapi/element/html/Canvas.zig. You can remove the dummy `RenderingContext2D` from that file. Remove the `registerTypes` function (since you're explicitly adding `RenderingContext2D` to the bridge)..and that's it.

nikneym:
I didn't realize there's a `CanvasRenderingContext2D` in zigdom, I've combine them together, anyways, this should be ready to merge now.

src/browser/webapi/css/CSSStyleDeclaration.zig
```diff
@@ -246,6 +246,24 @@ fn isInlineTag(tag_name: []const u8) bool {
     return false;
 }
 
+pub fn isHexColor(value: []const u8) bool {
```
karlseguin: If the only thing using `isHexColor` is `color.zig`, maybe it should go in `color.zig` ?

src/browser/webapi/canvas/CanvasRenderingContext2D.zig
```diff
@@ -0,0 +1,183 @@
...
+const color = @import("../../color.zig");
+const Page = @import("../../Page.zig");
+
+/// This class doesn't implement a `constructor`.
+/// It can be obtained with a call to `HTMLCanvasElement#getContext`.
+/// https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D
+const CanvasRenderingContext2D = @This();
+/// Fill color.
+/// TODO: Add support for `CanvasGradient` and `CanvasPattern`.
+fill_style: color.RGBA = color.RGBA.Named.black,
```
karlseguin: In the DOM, fields are prefixed with underscore. Isn't an issue here, but it's consistent with the rest of the API to avoid naming conflicts (which are more common with the file-as-a-struct pattern being used).

PR #1312, opened 2026-01-05
Stagehand zigdom

krichprollsch:
Following #1308 add Stagehand compatibility in `zigdom`

Replace #1217


krichprollsch:
ok, we are close, but now we need to implement [Document.evaluate](https://developer.mozilla.org/en-US/docs/Web/API/Document/evaluate)

PR #1313, opened 2026-01-05
Support XML parsing

nikneym:


src/browser/webapi/DOMParser.zig
```diff
@@ -21,41 +21,67 @@ const std = @import("std");
...
+        }
+
+        return .{ .html_document = doc };
     }
 
-    // Parse HTML into the document
-    const Parser = @import("../parser/Parser.zig");
-    var parser = Parser.init(page.arena, doc.asNode(), page);
-    parser.parse(normalized);
+    if (std.mem.eql(u8, mime_type, "text/xml")) {
```
karlseguin: I think we can do this for all the rest of the allowed values:

- text/xml
- application/xml
- application/xhtml+xml
- image/svg+xml

PR #1314, opened 2026-01-05
Fix `Document.currentScript.html` WPT

mookums:
This should fix the crash with the `html/dom/documents/dom-tree-accessors/Document.currentScript.html` by not double de-initializing Scripts in the `errorCallback`.

### DIFF
--- a/src/browser/ScriptManager.zig
+++ b/src/browser/ScriptManager.zig
@@ -728,17 +728,17 @@ pub const Script = struct {
 
         const manager = self.manager;
         manager.scriptList(self).remove(&self.node);
-        if (manager.shutdown) {
-            self.deinit(true);
-            return;
-        }
 
         if (self.mode == .import) {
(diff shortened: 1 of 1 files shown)


karlseguin:
I see that the issue is happening in `deinit` -> `reset` of the  `ScriptManager`, cleaning out its lists, but the errorCallback should remove the script from the list via the call to `manager.scriptList(self).remove(&self.node);`.

karlseguin:
I looked more into this. What seems to be happening is that the errorCallback is fired _before_ the script is added to the list. This likely happens because, on Client.request, we [most of the time] immediately pump curl's message loop to get the request going ASAP.

I think the correct fix is to change the ScriptManager's `addFromElement` to make sure the node is in the list before the request is made.

karlseguin:
Closed in favor of https://github.com/lightpanda-io/browser/pull/1319

PR #1328, opened 2026-01-07
add outerHTML setter

karlseguin:

PR #1329, opened 2026-01-07
fix `dom/nodes/Document-getElementById` WPT

mookums:
This fixes the `dom/nodes/Document-getElementById` WPT by properly handling the id argument. Two of the other failing tests are also resolved by having `setOuterHTML` from https://github.com/lightpanda-io/browser/pull/1328.


karlseguin:
In slack I flagged that I was looking into this. https://github.com/lightpanda-io/browser/pull/1328 already handles this one case also.

PR #1331, opened 2026-01-07
Adds case insensitivity to `Element.querySelector`

mookums:
This adds support for case insensitivity to the `selector/Parser.zig`. This fixes one of the two failures with the `dom/nodes/ParentNode-querySelector-case-insensitive.html` WPT.

### DIFF
--- a/src/browser/webapi/selector/List.zig
+++ b/src/browser/webapi/selector/List.zig
@@ -468,7 +468,7 @@ fn matchesAttribute(el: *Node.Element, attr: Selector.Attribute) bool {
         },
         .word => |expected| {
             // Space-separated word match (like class names)
-            var it = std.mem.tokenizeScalar(u8, value, ' ');
+            var it = std.mem.tokenizeAny(u8, value, &std.ascii.whitespace);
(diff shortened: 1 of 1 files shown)


src/browser/webapi/selector/List.zig
```diff
@@ -436,38 +436,64 @@ fn matchesPart(el: *Node.Element, part: Part, scope: *Node, page: *Page) bool {
...
+        },
+        .ends_with => |expected| {
+            return if (attr.case_insensitive)
+                std.ascii.endsWithIgnoreCase(value, expected)
+            else
+                std.mem.endsWith(u8, value, expected);
+        },
+        .word => |expected| {
+            // Space-separated word match (like class names)
+            var it = std.mem.tokenizeScalar(u8, value, ' ');
```
karlseguin: Is this change correct? It used to be ` std.mem.tokenizeAny(u8, value, &std.ascii.whitespace)`

mookums: Uses `tokenizeAny` now

PR #1334, opened 2026-01-07
Prefer `RAND_bytes` from BoringSSL for random

nikneym:
Zig will integrate random to `Io` interface soon; I think this is cleaner to do since we already have such symbol.


karlseguin:
I think it would be better to consider this change alongside a holistic approach to deal with Zig 0.16.

PR #1353, opened 2026-01-09
Use thread per connection

mrdimidium:
Adds the ability to launch a thread per CDP connection to the server, thereby handling multiple clients simultaneously.

Wraps some APIs with mutexes to allow access from different threads, but the network uses a separate one for each thread.

Creates a v8 session on a shared Isolate (which is probably not the best option). In the future, it would be worth maintaining a separate Isolate for each tab and a pool of pre-initialized ones to speed up startup.

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -122,8 +122,8 @@ jobs:
     needs: zig-build-release
 
     env:
-      MAX_VmHWM: 26000 # 26MB (KB)
-      MAX_CG_PEAK: 6000 # 6MB (KB)
+      MAX_VmHWM: 28000 # 28MB (KB)
+      MAX_CG_PEAK: 8000 # 8MB (KB)
       MAX_AVG_DURATION: 17
       LIGHTPANDA_DISABLE_TELEMETRY: true
 


src/App.zig
```diff
@@ -16,23 +16,24 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
+//! Container for global state / objects that various parts of the system might need.
+//! It is assumed that all fields of this structure are thread-safe.
```
karlseguin: I would assume the opposite. Nothing is thread-safe.

Platform isn't thread-safe, but it should only be used from the main thread.

Snapshot isn't thread-safe, the setup should only be used from the main thread, but can you concurrently create contexts from different threads from the same snapshot? Probably, but I don't know

Notification is definitely not thread safe, and should not be shared between threads, it's like http.

Telemetry is definitely not thread safe, but could b emade thread safe and shared between threads.

src/browser/Browser.zig
```diff
@@ -47,15 +48,15 @@ session_arena: ArenaAllocator,
 transfer_arena: ArenaAllocator,
 notification: *Notification,
 
-pub fn init(app: *App) !Browser {
+pub fn init(app: *App, http: *Http) !Browser {
```
karlseguin: I would add a new type, `Tenant` or  `Client` or something, because you'll need to inject more than Http (at the very least, also Notification, but it isn't hard to imagine a per-connection config too).

src/Server.zig
```diff
@@ -248,6 +273,7 @@ pub const Client = struct {
             .mode = .{ .http = {} },
             .socket_flags = socket_flags,
             .send_arena = ArenaAllocator.init(server.allocator),
+            .http = try Http.init(server.allocator, &server.app.config),
```
karlseguin: I believe the CA bundle is non-trivial to setup (both in terms of memory and time). I wonder if we could move the `c.curl_blob` to the application (like StartupData), and then re-use that for each http instance?

src/App.zig
```diff
@@ -16,23 +16,24 @@
...
 const Platform = @import("browser/js/Platform.zig");
 
 const Notification = @import("Notification.zig");
 const Telemetry = @import("telemetry/telemetry.zig").Telemetry;
 
-// Container for global state / objects that various parts of the system
-// might need.
 const App = @This();
 
-http: Http,
```
karlseguin, line 35: I know what you did is simpler/safer/logical, but I'm curious if you considered trying to share this across all connections? If each connection is a distinct tenant, than the obvious risk would be 1 tenant consuming all the resources. But if this is just 1 tenant using multi-threading to get greater performance, we could potentially optimize for that case by sharing more between them.

src/Server.zig
```diff
@@ -118,23 +136,29 @@ pub fn run(self: *Server, address: net.Address, timeout_ms: u32) !void {
...
-            posix.close(s);
-            self.client = null;
-        };
-
-        if (log.enabled(.app, .info)) {
-            var client_address: std.net.Address = undefined;
-            var socklen: posix.socklen_t = @sizeOf(net.Address);
-            try std.posix.getsockname(socket, &client_address.any, &socklen);
-            log.info(.app, "client connected", .{ .ip = client_address });
+        if (self.config.max_connections > self.connections.run_queue.len()) {
```
karlseguin: `run_queue` are pending tasks? If `getCpuCount() == 8`, does this mean I can have 8 concurrent connections + 8 waiting connections? I think we need to decide how to handle this? Reject any connection if the thread_pool is full? If we do allow pending connections, they might need their own timeout.

krichprollsch:
We can't build in ReleaseXXX due ot an error:
```
src/telemetry/lightpanda.zig:133:14: error: root source file struct 'App' has no member named 'RunMode'
    mode: App.RunMode,
          ~~~^~~~~~~~
```

src/browser/Browser.zig
```diff
@@ -40,35 +41,33 @@ const Browser = @This();
 
 env: js.Env,
 app: *App,
```
krichprollsch, line 42: I think you can also remove `app` field from browser.
The last reference I see is in `src/browser/webapi/Navigator.zig` where we should use `page._session.browser.http_client.user_agent;`

src/browser/Browser.zig
```diff
@@ -26,7 +26,8 @@ const log = @import("../log.zig");
 const App = @import("../App.zig");
 
 const ArenaPool = App.ArenaPool;
-const HttpClient = App.Http.Client;
+const Http = App.Http;
+const HttpClient = Http.Client;
```
krichprollsch:
```suggestion
const HttpClient = App.Http.Client;
```

src/ArenaPool.zig
```diff
@@ -29,6 +29,7 @@ free_list_len: u16 = 0,
 free_list: ?*Entry = null,
 free_list_max: u16,
 entry_pool: std.heap.MemoryPool(Entry),
+mutex: std.Thread.Mutex = .{},
```
krichprollsch, line 32:
```suggestion
mutex: std.Thread.Mutex,
```

src/browser/Browser.zig
```diff
@@ -40,35 +41,33 @@ const Browser = @This();
...
 notification: *Notification,
 
-pub fn init(app: *App) !Browser {
-    const allocator = app.allocator;
-
+pub fn init(allocator: Allocator, app: *App, http_client: *HttpClient) !Browser {
     var env = try js.Env.init(allocator, &app.platform, &app.snapshot);
     errdefer env.deinit();
 
     const notification = try Notification.init(allocator, app.notification);
```
karlseguin: This will dispatch on `app.notification` which isn't thread safe

src/ThreadPool.zig
```diff
@@ -0,0 +1,257 @@
...
+allocator: Allocator,
+
+active: u16,
+shutdown: bool,
+max_threads: u16,
+
+lock: std.Thread.RwLock,
+threads: std.DoublyLinkedList,
+
+const Func = struct {
```
karlseguin: Given that the threadpool is used to run 1 function (known at compile time) I think this can be simplified / optimized.

If I'm following along, this also doesn't keep a pool of threads alive to be re-used from connection to connection? There's no "min" setting? "ThreadPool" might not be the best name, since I think that describes something different.

src/LimitedAllocator.zig
```diff
@@ -0,0 +1,75 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+const std = @import("std");
+const Allocator = std.mem.Allocator;
+
+const LimitedAllocator = @This();
```
karlseguin: We don't have a holistic view of memory - v8 is doing its own thing.  The consequences of limiting Zig memory but not v8 memory isn't clear to me. There's also curl memory that we don't control.

karlseguin:
Would be nice if you could get thread sanitizer worker. I tried recently, but ran into build conflicts: https://github.com/lightpanda-io/browser/actions/runs/21083881975/job/60643343485

karlseguin:
Since I keep finding things that probably need to be made thread-safe, here's a checklist:

- [ ] telemetry/lightpanda.zig `mem_pool` isn't currently under the mutex
- [ ] notification.zig isn't thread safe. I believe this only has to be thread-safe because of the way telemetry hooks into individual notifications. I think a simpler solution is to make notification not have to be thread-safe, remove nested notifiations, and just trigger telemetry directly via something like `app.telemetry.record(...)`
- [ ] log.interceptors - @krichprollsch suggests simply removing this feature. It isn't clear how this is supposed to behave in a multi-connection environment anyways
- [ ] Can v8.Platform functions be (a) called from different threads (b) be called concurrently, e.g. Env.pumpMessageLoop

src/Config.zig
```diff
@@ -30,6 +30,13 @@ pub const RunMode = enum {
     version,
 };
 
+pub const MAX_HTTP_REQUEST_SIZE = 4096;
```
karlseguin: Hate to pick on names, but `MAX_HTTP_REQUEST_SIZE` makes it sound like that's the largest HTTP request size our browser supports.  This and `MAX_MESSAGE_SIZE` should be `CDP_...` .

src/Config.zig
```diff
@@ -144,7 +165,6 @@ pub const Serve = struct {
     timeout: u31 = 10,
     max_connections: u16 = 16,
     max_tabs_per_connection: u16 = 8,
-    max_memory_per_tab: u64 = 512 * 1024 * 1024,
```
karlseguin, line 161: Above this, `max_tabs_per_connection` doesn't appear to be used either.

src/Config.zig
```diff
@@ -131,6 +138,20 @@ pub fn userAgentSuffix(self: *const Config) ?[]const u8 {
     };
 }
 
+pub fn maxConnections(self: *const Config) u16 {
```
karlseguin, line 155: Because these command line argument names are exposed, they're hard to changed after the fact without breaking people's scripts.  We have different types of "connections", so which is this limiting?

In fairness, this is a problem we already have, e.g. --timeout is a timeout for what?  But if we going multi-threaded, it isn't crazy to think that we'll also want to configure limites to outgoing connections (globally, per host, per client, ..). So maybe it's time we start being a bit more verbose in our flags, `--cdp_max_connections`

src/log.zig
```diff
@@ -146,7 +146,14 @@ fn logTo(comptime scope: Scope, level: Level, comptime msg: []const u8, data: an
     }
     out.flush() catch return;
 
-    const interceptor = _interceptor orelse return;
+    // Copy the interceptor under lock, then release before doing I/O
+    const interceptor = blk: {
```
karlseguin: I think removing the feature is better than this. This is unexpected. You're getting logs, then there's another connections so you stop getting logs. Should either be a list of interceptors, or removed since @krichprollsch said he was ok with that.

Part of the question about this is how Chrome actually behaves

src/Server.zig
```diff
@@ -28,43 +28,51 @@ const ArenaAllocator = std.heap.ArenaAllocator;
...
-client: ?posix.socket_t,
 listener: ?posix.socket_t,
 json_version_response: []const u8,
 
-pub fn init(app: *App, address: net.Address) !Server {
-    const allocator = app.allocator;
+// Thread management
+active_threads: std.atomic.Value(u32) = .init(0),
+clients: std.ArrayListUnmanaged(*Client) = .{},
+clients_mu: std.Thread.Mutex = .{},
```
karlseguin: `clients_mutex`

src/Server.zig
```diff
@@ -81,17 +89,18 @@ pub fn stop(self: *Server) void {
 }
 
 pub fn deinit(self: *Server) void {
+    self.joinThreads();
```
karlseguin, line 99: `stop` has to be called before deinit? (maybe add a comment if that's the case)

src/Server.zig
```diff
@@ -119,97 +132,99 @@ pub fn run(self: *Server, address: net.Address, timeout_ms: u32) !void {
...
+fn spawnWorker(self: *Server, socket: posix.socket_t, timeout_ms: u32) !void {
+    if (self.shutdown.load(.acquire)) {
+        return error.ShuttingDown;
     }
+
+    if (self.active_threads.load(.monotonic) >= self.app.config.maxConnections()) {
+        return error.MaxThreadsReached;
+    }
+
+    _ = self.active_threads.fetchAdd(1, .monotonic);
```
karlseguin: There's a tiny window where multiple threads could pass the maxConn check, resulting in more connections than max?  This this has to be a `cmpxchg` or `active_threads` guarded by a mutex.

Or maybe you can `fetchAdd` first, and compare the return value to maxConnection (and decrement it back on `MaxThreadsReached`)

src/Server.zig
```diff
@@ -119,97 +132,99 @@ pub fn run(self: *Server, address: net.Address, timeout_ms: u32) !void {
...
-    var http = &self.app.http;
-    http.addCDPClient(.{
-        .socket = socket,
-        .ctx = client,
-        .blocking_read_start = Client.blockingReadStart,
-        .blocking_read = Client.blockingRead,
-        .blocking_read_end = Client.blockingReadStop,
-    });
-    defer http.removeCDPClient();
+    self.registerClient(client);
```
karlseguin, line 172: Is there a case where `server.stop()` is called before this? If so, I think that would block `server.deinit` forever.

src/browser/js/Inspector.zig
```diff
@@ -38,19 +38,21 @@ const IS_DEBUG = @import("builtin").mode == .Debug;
 const Inspector = @This();
 
 unique_id: i64,
+allocator: Allocator,
 isolate: *v8.Isolate,
 handle: *v8.Inspector,
 client: *v8.InspectorClientImpl,
 default_context: ?v8.Global,
-session: ?Session,
+sessions: std.ArrayListUnmanaged(*Session),
```
karlseguin: I'm not sure I understand this change. This would imply multiple BrowserContexts per CDP connection, which isn't something we support and would require more than just this

francisbouvier:
Just linking a few resources from the rusty (V8 rust port) project:

- This PR (2020, merged) removing the `v8::Locker` and considering `v8::Isolate` as _single threaded creature_ https://github.com/denoland/rusty_v8/pull/272
- This other PR (2020, merged) adding a `v8::IsolateHandle`, a thread-safe handle for Isolate https://github.com/denoland/rusty_v8/pull/274, see also doc https://docs.rs/rusty_v8/latest/rusty_v8/struct.IsolateHandle.html
- This issue (2021; not implemented AFAIK) trying to reintroduce it https://github.com/denoland/rusty_v8/issues/643

src/browser/js/Inspector.zig
mrdimidium: This is only needed in the future, I will delete it for this PR

src/Server.zig
mrdimidium: I added an explicit call to `stop` in `deinit` in case it wasn't called explicitly. `stop`  is idempotent.

.github/workflows/e2e-test.yml
```diff
@@ -122,7 +122,7 @@ jobs:
     needs: zig-build-release
 
     env:
-      MAX_MEMORY: 26000
+      MAX_MEMORY: 28000 # 28MB
```
mrdimidium: The measurements here are approximate (in reality, this PR should use slightly more memory, given the stack overhead of the new thread). Related: https://github.com/lightpanda-io/browser/pull/1510

karlseguin: Flagging @krichprollsch for this. Just from past experience, there's some sensitively to jumps in memory, as it potentially impacts our marketing material.

Run export LPD_VmHWM=`cat LPD.VmHWM`
Peak resident set size: 26328

src/browser/js/Env.zig
```diff
@@ -78,6 +86,9 @@ pub const InitOpts = struct {
 };
 
 pub fn init(app: *App, opts: InitOpts) !Env {
+    // Initialize class IDs once before any V8 work
+    class_id_once.call();
```
karlseguin, line 102: 👍

src/log.zig
```diff
@@ -49,8 +49,9 @@ const Opts = struct {
 
 pub var opts = Opts{};
 
-// synchronizes writes to the output
-var out_lock: Thread.Mutex = .{};
+// synchronizes access to interceptors
+var lock: Thread.RwLock = .{};
+var interceptors: std.ArrayListUnmanaged(Interceptor) = .{};
```
karlseguin:
```suggestion
var interceptors: std.ArrayList(Interceptor) = .{};
```

We're removed all of the [deprecated]`std.ArrayListUnmanaged` from the codebase recently.

src/log.zig
```diff
@@ -146,11 +147,25 @@ fn logTo(comptime scope: Scope, level: Level, comptime msg: []const u8, data: an
     }
     out.flush() catch return;
 
-    const interceptor = _interceptor orelse return;
-    if (interceptor.writer(interceptor.ctx, scope, level)) |iwriter| {
-        try logLogfmt(scope, level, msg, data, iwriter);
-        try iwriter.flush();
-        interceptor.done(interceptor.ctx, scope, level);
+    // Read lock allows multiple threads to log simultaneously
```
karlseguin: I don't think this is safe. There's no thread-safety within the LogInterceptor. It has a single std.io.Writer.Allocating field that it writes to.

And it eventually ends up calling Client.sendJSON which also isn't thread-safe.

But I agree with the general pattern, just easier and safer to use a plain Mutex for now I think.

src/main_legacy_test.zig
```diff
@@ -46,8 +46,11 @@ pub fn main() !void {
     var test_arena = std.heap.ArenaAllocator.init(allocator);
     defer test_arena.deinit();
 
-    var browser = try lp.Browser.init(app, .{});
-    const notification = try lp.Notification.init(app.allocator);
+    const http_client = try app.http.createClient(allocator);
+    defer http_client.deinit();
+
+    var browser = try lp.Browser.init(app, .{ .http_client = http_client });
```
karlseguin, line 52: Minor, but since you're going to do other tweaks, can we group the init/deinit?  Just easier to read/reason about. I saw this in lightpanda.zig fetch and it confused me for a second. (I updated it there).

```zig
const notification = try lp.Noti..
defer notification.deinit()

const browser = try lp.Browser...
defer browser.deinit()
```

src/Server.zig
```diff
@@ -81,17 +92,22 @@ pub fn stop(self: *Server) void {
 }
 
 pub fn deinit(self: *Server) void {
+    if (!self.shutdown.load(.acquire)) {
```
karlseguin, line 95: Calling `stop()` directly here seems safe.

src/Server.zig
```diff
@@ -81,17 +92,22 @@ pub fn stop(self: *Server) void {
...
-    // *if* server.run is running, we should really wait for it to return
-    // before existing from here.
+    self.clients.deinit(self.allocator);
+    self.clients_pool.deinit();
     self.allocator.free(self.json_version_response);
 }
 
 pub fn run(self: *Server, address: net.Address, timeout_ms: u32) !void {
-    const flags = posix.SOCK.STREAM | posix.SOCK.CLOEXEC;
+    const flags = posix.SOCK.STREAM | posix.SOCK.CLOEXEC | posix.SOCK.NONBLOCK;
```
karlseguin, line 110: Why?

You had to add a busy loop because of this. For the shutdown? Why doesn't closing the socket (macos) or shutting down (linux) unblock it?

src/Server.zig
```diff
@@ -119,97 +139,124 @@ pub fn run(self: *Server, address: net.Address, timeout_ms: u32) !void {
...
+    };
+    defer client.deinit();
+
+    self.registerClient(client);
+    defer self.unregisterClient(client);
+
+    // Check shutdown after registering to avoid missing stop() signal.
+    // If stop() already iterated over clients, this client won't receive stop()
+    // and would block joinThreads() indefinitely.
+    if (self.shutdown.load(.acquire)) {
```
karlseguin, line 178: 👍

karlseguin:
I believe this issue:
* [ ]  telemetry/lightpanda.zig `mem_pool` isn't currently under the mutex

And this question:
* [ ]  Can v8.Platform functions be (a) called from different threads (b) be called concurrently, e.g. Env.pumpMessageLoop

still need to be addressed.

.github/workflows/e2e-test.yml
karlseguin: Also, do we know why? Is it just the overhead of the thread's default stack size?

src/browser/Robots.zig
```diff
@@ -58,12 +58,16 @@ pub const RobotStore = struct {
...
     allocator: std.mem.Allocator,
     map: RobotsMap,
+    mutex: std.Thread.Mutex = .{},
 
     pub fn init(allocator: std.mem.Allocator) RobotStore {
         return .{ .allocator = allocator, .map = .empty };
     }
 
     pub fn deinit(self: *RobotStore) void {
+        self.mutex.lock();
```
karlseguin, line 121: @mookums  FYI.  We need to consider whether the `pending_robots_queue` should be pushed down from the Client to here, so that  N connections all share the same preflight queue. Personally, I think it's ideal, but not without its challenge. (Those requests can disappear while the robots fetch is happening).

.github/workflows/e2e-test.yml
mrdimidium: Yes, at least the stack size for one more thread. However, this is an approximate measurement, and we need to check the memory consumption more precisely in #1510

src/Server.zig
mrdimidium: The difference is what happens if you `recv() / send() / accept()` after `shutdown`. For a blocking socket, this can lead to blocking (since there is no data), whereas with a non-blocking socket, this is guaranteed (EAGAIN).

karlseguin: But this is only for the accepting socket (the client socket is nonblocking already).  Have we run into this blocking behavior?  Would using a `poll` with only the accepting socket be a better option?

krichprollsch:
Regarding the logs, I made some tests with https://github.com/lightpanda-io/demo/pull/109
It appears each CDP connection in Chrome has different session ids, but the target ids are shared.

```json
{"method":"Target.attachedToTarget","params":{"sessionId":"C24C495C022C331A0CAA0C54E286EF83","targetInfo":{"targetId":"9B80BFFA2710643F4EE94D85A44818A5","type":"page","title":"","url":"about:blank","attached":true,"canAccessOpener":false,"browserContextId":"D10DFC3E4D711EF82C36A72264898C82"},"waitingForDebugger":false}}
```

```json
{"method":"Target.attachedToTarget","params":{"sessionId":"B2423DA3CF0FCED5D1B354F1B53F53E1","targetInfo":{"targetId":"9B80BFFA2710643F4EE94D85A44818A5","type":"page","title":"","url":"about:blank","attached":true,"canAccessOpener":false,"browserContextId":"D10DFC3E4D711EF82C36A72264898C82"},"waitingForDebugger":false}}
```

Calling `xxx.Enable`,  on a given debugger session has no effect on another. So each CDP connection/session can get events independently from other connections/sessions.

With this PR, we will dispatch logs messages to every connections if one of them activate the logs. This is definitively not correct. But is it temporary acceptable? I think we can let as it for now if it doesn't break existing script. As I said, removing the feature would be OK too.

karlseguin:
I guess there could be security concerns. But not so much if it's just 1 customer/client.  Anyone actually logging the output will end up getting a bunch of duplicates (the same message for each connection). I don't think it's a huge deal, but it should be addressed promptly in a follow up PR.

krichprollsch:
About the memory usage change:

| branch | cgroup |diff| virtual |diff|
|-|-|-|-|-|
|main|5280||24928||
|PR|7140 |+1860/+35%|26764 |+1836/+7%|

* main ref: https://github.com/lightpanda-io/browser/actions/runs/21990514749/job/63536856449
* PR ref: https://github.com/lightpanda-io/browser/actions/runs/21989271390/job/63532674144?pr=1353

PR #1356, opened 2026-01-11
Initial support for `SubtleCrypto` API

nikneym:
This PR adds initial implementation for `SubtleCrypto` API. `sign`, `verify` and `exportKey` are implemented for HMAC; `deriveBits` is implemented for X25519.

### DIFF
--- a/src/browser/webapi/SubtleCrypto.zig
+++ b/src/browser/webapi/SubtleCrypto.zig
@@ -295,6 +295,7 @@ pub const CryptoKey = struct {
                 if (std.mem.eql(u8, "X25519", name)) {
                     return initX25519(extractable, key_usages, page);
                 }
+                log.warn(.not_implemented, "CryptoKey.init", .{ .name = name });
                 return error.NotSupported;
             },
             .object => |object| {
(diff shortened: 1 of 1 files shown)


src/browser/webapi/SubtleCrypto.zig
```diff
@@ -0,0 +1,304 @@
...
+            ._type = .hmac,
+            ._extractable = extractable,
+            ._usages = 0,
+            ._key = key,
+            ._digest = digest,
+        });
+    }
+
+    fn signHMAC(self: *const CryptoKey, data: []const u8, page: *Page) !js.Promise {
+        const buffer = try page.arena.alloc(u8, crypto.EVP_MD_size(self._digest));
```
karlseguin: this can be the `call_arena`?

nikneym: We're returning this memory to JS context, I thought `call_arena` lifetime won't be sufficient.

karlseguin: The `call_arena` is guaranteed to live until any data from the current call is transferred to v8. "call" in "call_arena" refers to to v8 calling a zig function.  It looks like the buffer doesn't outlive this single call.

krichprollsch:
Would be helpful to add logs when hitting a not implemented algorithm/option.
similar to: https://github.com/lightpanda-io/browser/blob/main/src/browser/webapi/Document.zig#L175

krichprollsch:
I think this PR is ready to merge, are you agree @karlseguin?

PR #1359, opened 2026-01-12
Improve crash handling

karlseguin:
This adds a crash handler which reports a crash (if telemetry is enabled). On a crash, this looks for `curl` (using the PATH env), and forks the process to then call execve. This relies on a new endpoint to be setup to accept the "report". Also, we include very little data..I figured just knowing about crashes would be a good place to start.

A panic handler is provided, which override's Zig default handler and hooks into the crash handler.

An `assert` function is added and hooks into the crash handler. This is currently only used in one place (Session.zig) to demonstrate its use. In addition to reporting a failed assert, the assert aborts execution in ReleaseFast (as opposed to an undefined behavior with std.debug.assert).

I want to hook this into the v8 global error handler, but only after direct_v8 is merged.

Much of this is inspired by bun's code. They have their own assert (1) and a [more sophisticated] crashHandler (2).
:

(1) https://github.com/oven-sh/bun/blob/beccd01647ec4fcb5895be5c5fb88f5254dc1e8d/src/bun.zig#L2987 (2) https://github.com/oven-sh/bun/blob/beccd01647ec4fcb5895be5c5fb88f5254dc1e8d/src/crash_handler.zig#L198

### DIFF
--- a/src/crash_handler.zig
+++ b/src/crash_handler.zig
@@ -79,7 +79,7 @@ fn report(reason: []const u8) !void {
     var args_buffer: [4096]u8 = undefined;
     var writer: std.Io.Writer = .fixed(&args_buffer);
 
-    try writer.print("https://crash.lightpanda.io/?v={s}&r=", .{lp.build_config.git_commit});
+    try writer.print("https://crash.lightpanda.io/c?v={s}&r=", .{lp.build_config.git_commit});
     for (reason) |b| {
         switch (b) {
             'A'...'Z', 'a'...'z', '0'...'9', '-', '.', '_' => try writer.writeByte(b),


src/crash_handler.zig
```diff
@@ -0,0 +1,129 @@
...
+        return;
+    }
+
+    var curl_path: [2048]u8 = undefined;
+    const curl_path_len = curlPath(&curl_path) orelse return;
+
+    var args_buffer: [4096]u8 = undefined;
+    var writer: std.Io.Writer = .fixed(&args_buffer);
+
+    try writer.print("https://crash.lightpanda.io/?v={s}&r=", .{lp.build_config.git_commit});
```
krichprollsch: I will share the http server w/ the telemetry one. I will configure the subdomain `crash.lightpanda.io` on it, but it would be helpful to use a specific endpoit also.

I suggest `/c` for crash.

```suggestion
    try writer.print("https://crash.lightpanda.io/c?v={s}&r=", .{lp.build_config.git_commit});
```

PR #1365, opened 2026-01-13
Try catch caught

karlseguin:
Improves the ergonomics of TryCatch and Function.tryCall.

You now use:

```zig
if (try_catch.caught(allocator)) |caught| {
   log.err(.js, "MutationObserver", .{.caught = caught});
}
```

The key change is that all fields (stack, exception, line) are captured by "caught" and it can be formatted as-is in the logs. Making it easier to use and providing more consistent output.

The Function.tryCall now takes a `*js.TryCatch.Caught`, rather than a `*js.Function.Result`.


krichprollsch:
This change looks great. But I think the rebase is not correct, I guess I should'nt see 25 commits in the PR :eyes:

karlseguin:
I cleaned it up, sorry.

PR #1376, opened 2026-01-16
Add simplified `setAttributeNS` and `getAttributeNS`

nikneym:
This ignores namespaces for now, we have to come up with a solution if it becomes a necessity.


karlseguin:
Could do like Document.createAttributeNS and log not implemneted (and then just treat it like HTML, like you're doing):

```zig
pub fn createAttributeNS(_: *const Document, namespace: []const u8, name: []const u8, page: *Page) !?*Element.Attribute {
    if (std.mem.eql(u8, namespace, "http://www.w3.org/1999/xhtml") == false) {
        log.warn(.not_implemented, "document.createAttributeNS", .{ .namespace = namespace });
    }
   ...
}
```

If a script uses it, and things end up broken, it might help us identity the issue.

PR #1377, opened 2026-01-16
Enable ThreadSanitizer (tsan) in CI run

karlseguin:
Fix a couple test-specific threading issues. Not enabled in the `make test` target because it adds 3-4 seconds to shutdown time.

Just a small thing to enable before any multithreading lands.


karlseguin:
Closing this - don't want to deal with build issues.

PR #1383, opened 2026-01-19
Add XHR finalizer and ArenaPool

karlseguin:
Any object we return from Zig to V8 becomes a v8::Global that we track in our `ctx.identity_map`. V8 will not free such objects. On the flip side, on its own, our Zig code never knows if the underlying v8::Object of a global can still be used from JS. Imagine an XHR request where we fire the last readyStateChange event..we might think we no longer need that XHR instance, but nothing stops the JavaScript code from holding a reference to it and calling a property on it, e.g. `xhr.status`.

What we can do is tell v8 that we're done with the global and register a callback. We make our reference to the global weak. When v8 determines that this object cannot be reached from JavaScript, it _may_ call our registered callback. We can then clean things up on our side and free the global (we actually _have_ to free the global).

v8 makes no guarantee that our callback will ever be called, so we need to track these finalizable objects and free them ourselves on context shutdown. Furthermore there appears to be some possible timing issues, especially during context shutdown, so we need to be defensive and make sure we don't double-free (we can use the existing identity_map for this).

An type like XMLHttpRequest can be re-used. After a request succeeds or fails, it can be re-opened and a new request sent. So we also need a way to revert a "weak" reference back into a "strong" reference. These are simple v8 calls on the v8::Global, but it highlights how sensitive all this is. We need to mark it as weak when we're 100% sure we're done with it, and we need to switch it to strong under any circumstances where we might need it again on our side.

Finally, none of this makes sense if there isn't something to free. Of course, the finalizer lets us release the v8::Global, and we can free the memory for the object itself (i.e. the `*XMLHttpRequest`). This PR also adds an ArenaPool. This allows the XMLHTTPRequest to be self-contained and not need the `page.arena`. On init, the `XMLHTTPRequest` acquires an arena from the pool. On finalization it releases it back to the pool. So we now have:

- page.call_arena: short, guaranteed for 1 v8 -> zig -> v8 flow
- page.arena long: lives for the duration of the entire page
- page.arena_pool: ideally lives for as long as needed by its instance (but no guarantees from v8 about this, or the script might leak a lot of global, so worst case, same as page.arena)

### DIFF
--- a/src/http/Client.zig
+++ b/src/http/Client.zig
@@ -606,7 +606,7 @@ fn processMessages(self: *Client) !bool {
                     continue;
                 };
                 if (!proceed) {
-                    self.requestFailed(transfer, error.Abort);
+                    self.requestFailed(transfer, error.Abort, true);
                     break :blk;
                 }
             }


krichprollsch:
> page.arena_pool: ideally lives for as long as needed by its instance (but no guarantees from v8 about this, or the script might leak a lot of global, so worst case, same as page.arena)

Do you mean the arena_pool is released at the end of the page? b/c I didn't see in the corresponding code :thinking:

src/browser/webapi/net/XMLHttpRequest.zig
```diff
@@ -79,19 +79,25 @@ const ResponseType = enum {
 };
 
 pub fn init(page: *Page) !*XMLHttpRequest {
-    return page._factory.xhrEventTarget(XMLHttpRequest{
+    const arena = try page.getArena("XMLHttpRequest");
```
krichprollsch: On review, I was suprised by this string parameter. I was wondering if is was a kind of `id` for the pool, then I saw it's for debugging only.

Maybe it would be clearer to change the signature for something like `page.getArena(.{.debug = "XMLHttpRequest"})` to be more explicit.

So it could be also optional...

src/browser/webapi/net/XMLHttpRequest.zig
```diff
@@ -148,6 +154,7 @@ pub fn send(self: *XMLHttpRequest, body_: ?[]const u8) !void {
     if (self._ready_state != .opened) {
         return error.InvalidStateError;
     }
+    self._page.js.strongRef(self);
```
krichprollsch: I'm confused about when we must call `strongRef`.
ie. Why we don't do that on `open` too?

Also if the js call `open` but never `send`, we never set the js as weak. Does it mean v8's GC will never be able to finalize it?

karlseguin: XMLHttpRequest starts off "weak" because JsApi.Meta has `weak = true`:

```zig
pub const weak = true;
pub const finalizer = bridge.finalizer(XMLHttpRequest.deinit);
```

Maybe that should be the default, need to do more types to figure it out.

You need a strong reference whenever you want to block v8 from finalizing it. If you take something like TextEncoder, we'll never call `strongRef` on it. We _never_ call a function on a TextEncoder from Zig.  But for XMLHttpRequest, you can have:

{
   const xhr = new XMLHttpRequest();
   xhr.open('get', '....');
   xhr.send();
}

Here, xhr goes out of scope, so, from the JS's point of view, it can be freed. But send started an HTTP requests, and those http callbacks reference the XMLHttpRequest object. IF the v8 finalizer was to execute after send() but while the request is stil being procesed, we'd get a use-after-free in our HTTP callbacks. So we prevent that by calling `strongRef` and only make it weak again when the http callbacks complete (on success or on error).

src/browser/webapi/net/XMLHttpRequest.zig
```diff
@@ -79,19 +79,24 @@ const ResponseType = enum {
...
         ._request_headers = try Headers.init(null, page),
     });
 }
 
 pub fn deinit(self: *XMLHttpRequest) void {
-    if (self.transfer) |transfer| {
-        transfer.abort(error.Abort);
-        self.transfer = null;
+    if (self._transfer) |transfer| {
+        transfer.terminate();
```
krichprollsch: I'm not sure about `terminate` vs `abort` here. Since the end of request event could be used by a pure zig listener too. I mean, what about in flight request interception?

karlseguin: You're right. The code has been changed to:

```zig
pub fn deinit(self: *XMLHttpRequest, comptime shutdown: bool) void {
    if (self._transfer) |transfer| {
        if (shutdown) {
            transfer.terminate();
        } else {
            transfer.abort(error.Abort);
        }
        self._transfer = null;
    }
    self._page.releaseArena(self._arena);
    self._page._factory.destroy(self);
}
```

During a Context shutdown, it isn't safe to do certain things (e.g. executing new JavaScript, such as a on_abort callback). 

PLUS `terminate` is now closer to abort. It _will_ send notifications, but it _will not_ execute the registered error_callback.

PR #1389, opened 2026-01-20
`Image` & `Style`: Dispatch `load` event

nikneym:


karlseguin:
No tests?

nikneym:
Removed `hasListener` and added a test.

src/browser/webapi/element/html/Image.zig
```diff
@@ -47,6 +48,11 @@ pub fn getSrc(self: *const Image, page: *Page) ![]const u8 {
 
 pub fn setSrc(self: *Image, value: []const u8, page: *Page) !void {
     try self.asElement().setAttributeSafe("src", value, page);
+
+    // We don't actually fetch the media, here we fake the load call.
+    const event_target = self.asNode().asEventTarget();
+    const event = try Event.initTrusted("load", .{}, page);
+    return page._event_manager.dispatch(event_target, event);
```
krichprollsch: I do think it would be better to dispatch the event asynchronously.

krichprollsch:
Here are references:
* https://html.spec.whatwg.org/multipage/images.html
* https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/load_event

1. I do think we should trigger the event for `img` tags initially in the HTML. (same for other tags. Synchronously b/c they must block `window.load` event.

2. I guess also we should track  dynamic insertion of these tags in the DOM to also dispatch the events correctly on them. It's something we do for script already if I'm not wrong.

3. I think we should dispatch `load` on `body` element just after dispatching the `window,onload` event (TBC).

src/browser/tests/element/html/image.html
```diff
@@ -97,3 +97,28 @@
...
+    testing.expectEqual(img, target);
+  });
+
+  for (let i = 0; i < 3; i++) {
+    img.src = "https://cdn.lightpanda.io/website/assets/images/docs/hn.png";
+    testing.expectEqual("https://cdn.lightpanda.io/website/assets/images/docs/hn.png", img.src);
+  }
+
+  // Should work since we synchronously exeute the listener.
+  testing.expectEqual(3, count);
```
karlseguin: Doesn't pass in FF. Now, sometimes our test don't pass in FF/Chrome, and that's ok. But I _think_ this behavior should be consistent with FF/Chrome?

nikneym: Correct; that was because callback is being executed synchronously, I'll use `Scheduler` to simulate asynchrony instead.

src/browser/webapi/element/html/Image.zig
```diff
@@ -115,6 +127,124 @@ pub const JsApi = struct {
...
+}
+
+pub const Build = struct {
+    pub fn created(node: *Node, page: *Page) !void {
+        const self = node.as(Image);
+        const image = self.asElement();
+        // Exit if src not set. We might want to check if src point to valid image.
+        const src = image.getAttributeSafe("src") orelse return;
+        // We can at least check if this is a valid URL.
+        if (!URL.isCompleteHTTPUrl(src)) {
```
karlseguin: Don't think these checks are correct, `/img.png`  should still get loaded.

src/browser/webapi/element/html/Image.zig
```diff
@@ -115,6 +127,124 @@ pub const JsApi = struct {
...
+        //
+        // I'm well aware that such scenario is possible:
+        // imageElement.onload = () => console.warn("HA HA HA overwritten!");
+        //
+        // this is allowed both in Chrome and FF. Honestly it makes the logic
+        // quite easier, which might be the reason.
+        return page.scheduler.add(
+            args,
+            dispatchLoadEvent,
+            25,
```
karlseguin: Don't know about this...in the case of google news, it would just delay the loading of the page by 25ms...

nikneym:
> the incorrect use of `isCompleteHTTPUrl`

I agree, since we don't actually make a request to specified `src`, I've put that instead. I'll get rid of it.

> 1 - Should this be applied to other element types, and if so, can the solution be generalized

I'd like to introduce a queue on `Scheduler` for that. We can have a queue of background tasks for elements that require `load` event execution (multimedia, `<link>`, `<style>` etc.). That'd also reduce allocations since we can pass `*Page` as argument when executing such tasks.

I'll come up with implementation on a separate PR.

src/browser/EventManager.zig
```diff
@@ -42,11 +43,26 @@ list_pool: std.heap.MemoryPool(std.DoublyLinkedList),
...
+///
+/// Or:
+///
+/// ```js
+/// img.onload = () => { ... };
+/// ```
+inline_lookup: std.AutoHashMapUnmanaged(usize, js.Function.Global),
 
 pub fn init(page: *Page) EventManager {
+    lp.assert(@alignOf(EventTarget) == 8, "EventManager.init", .{ .event_target_alignment = @alignOf(EventTarget) });
```
karlseguin: I would move this as a comptime check in `createLookupKey`, since that's where it's important and it kind of helps document what you're doing.

src/browser/EventManager.zig
```diff
@@ -42,11 +43,26 @@ list_pool: std.heap.MemoryPool(std.DoublyLinkedList),
...
+/// ```html
+/// <img onload="(() => { ... })()" />
+/// ```
+///
+/// Or:
+///
+/// ```js
+/// img.onload = () => { ... };
+/// ```
+inline_lookup: std.AutoHashMapUnmanaged(usize, js.Function.Global),
```
karlseguin: This idea here is to not have a `?js.Function.Global` on Image directly because most images don't have this and you want to avoid bloating _every_ image with those 16 bytes, right? Probably worth mentioning that.

My gut feeling would be to put this on the page, along with the the 7 other similar lookups, e.g. `_element_styles` (which would make it a bit more obvious what the purpose is, since it's grouped with other fields that serve the same purpose).

src/browser/EventManager.zig
```diff
@@ -436,6 +478,18 @@ fn findListener(list: *const std.DoublyLinkedList, typ: []const u8, callback: Ca
     return null;
 }
 
+/// Creates a lookup key to use with `inline_lookup`.
+fn createLookupKey(event_target: *EventTarget, event_type: Listener.Type) usize {
+    const ptr = @intFromPtr(event_target) >> 3;
+    lp.assert(ptr < (1 << 57), "createLookupKey: pointer overflow", .{ .ptr = ptr });
+    return ptr | (@as(u64, @intFromEnum(event_type)) << 57);
```
karlseguin: How sure are you that this is safe to do across all supported platforms? Does it break canonical addresses?

src/browser/webapi/element/html/Image.zig
```diff
@@ -115,6 +126,118 @@ pub const JsApi = struct {
...
+        _page.js.localScope(&ls);
+        defer ls.deinit();
+
+        try _page._event_manager.dispatchWithFunction(
+            event_target,
+            event,
+            _on_load.local(&ls.local),
+            .{ .context = "Image.onload" },
+        );
+        return null;
```
karlseguin: Is this right? It should still dispatch the load event, no?

src/browser/EventManager.zig
nikneym: > This idea here is to not have a ?js.Function.Global on Image directly because most images don't have this and you want to avoid bloating every image with those 16 bytes, right? Probably worth mentioning that.

Indeed, these are rare kind of event listeners. I'll add a comment for that.

> My gut feeling would be to put this on the page, along with the the 7 other similar lookups, e.g. _element_styles (which would make it a bit more obvious what the purpose is, since it's grouped with other fields that serve the same purpose).

Yes, we can extend the usage of it for other kinds of resources.

src/browser/EventManager.zig
nikneym: This assumes we're on virtual address space; in which most 64-bit systems use 48 to 57 bits for pointers. Known limitation is that this won't work for systems like s390x. Should be compatible with AMD64 and ARM64. Hence the reason I've added the assertion to check if we're overflowing bits.

I'm thinking of adding a fallback implementation for systems that use full bits.

karlseguin: My understanding is that modern cpus can verify those top "unused" bits (specifically to prevent this type of thing in the name of future proofing), and a correct solution possibly requires more steps.

https://muxup.com/2023q4/storing-data-in-pointers

nikneym: > My understanding is that modern cpus can verify those top "unused" bits (specifically to prevent this type of thing in the name of future proofing), and a correct solution possibly requires more steps.
>
> https://muxup.com/2023q4/storing-data-in-pointers

I think statements made here could be a problem if we'd use the key to get a pointer back; we're only using it to generate the key though (one-way), I should have noted that.

karlseguin: That's true.

Can we 
1 - Move `inline_lookup` to something like Page._element_listeners` or something
2 - Move the alignment assertion into `createLookupKey`, do it at comptime, and add a bit of documentation around what's going on in there.
3 - Fix, what I think is an issue with the early return in Image.zig on line 155.

Then we can merge.

nikneym:
Implemented inline event listeners in #1430; this PR will base it instead soon.

src/browser/Page.zig
```diff
@@ -699,15 +707,34 @@ pub fn documentIsComplete(self: *Page) void {
...
     self.js.localScope(&ls);
     defer ls.deinit();
 
+    // Dispatch `_to_load` events before window.load.
+    for (self._to_load.items) |element| {
+        const maybe_inline_listener = self.getAttrListener(element, .onload);
+
+        try self._event_manager.dispatchWithFunction(
+            element.asEventTarget(),
+            event,
```
karlseguin, line 722: Do they really all share the same Event with each other and with the window.load?

karlseguin:
Did you test to see if it fixed the google news issue?

nikneym:
I did, I don't think this is the only issue (if it ever was a specific issue for this) though. I'm looking for what else is missing.

src/browser/Page.zig
nikneym: I believe only difference is the target which is handled?

PR #1391, opened 2026-01-20
call env.lowMemoryNotification() during page reset

krichprollsch:
calling env.lowMemoryNotification() on page reset encourages v8 to free memory and keep low usage.

On the crawler benchmark, we peak at ~70MB vs ~22MB with this change.


karlseguin:
Could be interesting to see if Inspector.destroyContext has any impact on this once we do it.

PR #1400, opened 2026-01-22
Call `contextDestroyed` to signal context removal

nikneym:
This PR adds `contextDestroyed` calls just before an isolate be removed.

relates with https://github.com/lightpanda-io/project/issues/185


krichprollsch:
I'm closing this one for now. According with #1408 I think calling directly `Inspector.contextDestroyed` is useless since the inspector is already notified via tasks.

Moreover, If we decide to implement `Inspector.contextDestroyed` I think it would be better to use some callbacks on context creation and context remove.

PR #1404, opened 2026-01-23
Capture the stack trace on the crash handler report

karlseguin:


src/crash_handler.zig
```diff
@@ -68,32 +68,51 @@ pub noinline fn crash(
...
-    try writer.print("https://crash.lightpanda.io/c?v={s}&r=", .{lp.build_config.git_commit});
-    for (reason) |b| {
-        switch (b) {
-            'A'...'Z', 'a'...'z', '0'...'9', '-', '.', '_' => try writer.writeByte(b),
-            ' ' => try writer.writeByte('+'),
-            else => try writer.writeByte('!'), // some weird character, that we shouldn't have, but that'll we'll replace with a weird (bur url-safe) character
+    var url_buffer: [4096]u8 = undefined;
+    const url = blk: {
+        var writer: std.Io.Writer = .fixed(&url_buffer);
+        try writer.print("http://localhost:1234/c?v={s}&r=", .{lp.build_config.git_commit});
```
krichprollsch: localhost?

src/crash_handler.zig
```diff
@@ -68,32 +68,51 @@ pub noinline fn crash(
...
+        written[written.len] = 0;
+        break :blk written[0 .. written.len + 1];
+    };
 
     var argv = [_:null]?[*:0]const u8{
         curl_path[0..curl_path_len :0],
         "-fsSL",
+        "-H",
+        "Content-Type: application/octet-stream",
+        "--data-binary",
```
krichprollsch, line 114: Does this change the method used? Or is it still a `GET`?

src/crash_handler.zig
karlseguin: Ughh..I was debugging. Updated.

src/crash_handler.zig
karlseguin: It's a POST now.

krichprollsch: Ok, so let's wait until I upgraded the telemetry to support that too.

src/crash_handler.zig
```diff
@@ -68,32 +68,51 @@ pub noinline fn crash(
...
+        var writer: std.Io.Writer = .fixed(stack_buffer[0..4095]); // reserve 1 space
+        std.debug.dumpCurrentStackTraceToWriter(begin_addr, &writer) catch {};
+        var written = writer.buffered();
+        if (written.len == 0) {
+            break :blk "???";
+        }
+        // overwrite the last character with our null terminator, safest since
+        // our buffer could be full at this point
+        written[written.len] = 0;
+        break :blk written[0 .. written.len + 1];
```
krichprollsch: Should'nt be? it crashes here...

```suggestion
        written[written.len-1] = 0;
        break :blk written[0 .. written.len];
```

krichprollsch:
The telemetry server is ready :heavy_check_mark:

src/crash_handler.zig
karlseguin: Actually, I meant:

```
stack_buffer[written.len] = 0;
break :blk stack_buffer[0 .. written.len + 1];
```

Which is why I pass 0..4095 instead of 0..4096 into the Writer. But your version and this result in the same output.

PR #1407, opened 2026-01-23
Support for `robots.txt`

mookums:
This adds support for `robots.txt` (https://datatracker.ietf.org/doc/html/rfc9309) in the Lightpanda Browser.

### DIFF
--- a/src/http/Client.zig
+++ b/src/http/Client.zig
@@ -436,6 +436,12 @@ fn robotsShutdownCallback(ctx_ptr: *anyopaque) void {
         ctx.robots_url,
     ) orelse @panic("Client.robotsErrorCallback empty queue");
     defer queued.value.deinit(ctx.client.allocator);
+
+    for (queued.value.items) |queued_req| {
+        if (queued_req.shutdown_callback) |shutdown_cb| {
+            shutdown_cb(queued_req.ctx);
+        }
+    }
 }
 
 fn waitForInterceptedResponse(self: *Client, transfer: *Transfer) !bool {


src/browser/URL.zig
```diff
@@ -778,3 +788,31 @@ test "URL: concatQueryString" {
         try testing.expectEqual("https://www.lightpanda.io/index?1=2&a=b", url);
     }
 }
+
+test "URL: getRobotsUrl" {
```
karlseguin, line 792: Minor: Could add a test for the `error.NoOrigin` case.

src/http/Client.zig
```diff
@@ -247,6 +283,107 @@ pub fn request(self: *Client, req: Request) !void {
...
+        .done_callback = robotsDoneCallback,
+        .error_callback = robotsErrorCallback,
+    });
+}
+
+fn robotsHeaderCallback(transfer: *Http.Transfer) !bool {
+    const ctx: *RobotsRequestContext = @ptrCast(@alignCast(transfer.ctx));
+
+    if (transfer.response_header) |hdr| {
+        log.debug(.browser, "robots status", .{ .status = hdr.status });
```
karlseguin: useful to include the url here.

src/http/Client.zig
```diff
@@ -247,6 +283,107 @@ pub fn request(self: *Client, req: Request) !void {
...
+const RobotsRequestContext = struct {
+    client: *Client,
+    req: Request,
+    robots_url: [:0]const u8,
+    buffer: std.ArrayList(u8),
+    status: u16 = 0,
+};
+
+fn fetchRobotsThenProcessRequest(self: *Client, robots_url: [:0]const u8, req: Request) !void {
+    const ctx = try self.allocator.create(RobotsRequestContext);
```
karlseguin: As a follow-up PR, I think we should leverage the ArenaPool here. Only reason not to do it in this PR is that it isn't readily accessible from this point.  But ultimately, it would look like;

```zig
const arena =  try client.arena_pool.acquire();
errdefer client.arena_pool.release(arena);

// use the arena to create RobotsRequestContext and to populate the buffer
// release the arena when done
```

src/http/Client.zig
```diff
@@ -247,6 +283,107 @@ pub fn request(self: *Client, req: Request) !void {
...
+}
+
+fn robotsDataCallback(transfer: *Http.Transfer, data: []const u8) !void {
+    const ctx: *RobotsRequestContext = @ptrCast(@alignCast(transfer.ctx));
+    try ctx.buffer.appendSlice(ctx.client.allocator, data);
+}
+
+fn robotsDoneCallback(ctx_ptr: *anyopaque) !void {
+    const ctx: *RobotsRequestContext = @ptrCast(@alignCast(ctx_ptr));
+    defer ctx.client.allocator.destroy(ctx);
```
karlseguin: `defer ctx.deinit()`  just because you have the same code in  ErrorCallback

src/http/Client.zig
```diff
@@ -247,6 +283,107 @@ pub fn request(self: *Client, req: Request) !void {
...
+        );
+
+        try ctx.req.robots.put(ctx.robots_url, robots);
+
+        const path = URL.getPathname(ctx.req.url);
+        allowed = robots.isAllowed(path);
+    }
+
+    // If not found, store as Not Found.
+    if (ctx.status == 404) {
```
karlseguin:
```suggestion
    else if (ctx.status == 404) {
```

src/http/Client.zig
```diff
@@ -247,6 +283,107 @@ pub fn request(self: *Client, req: Request) !void {
...
+fn robotsDoneCallback(ctx_ptr: *anyopaque) !void {
+    const ctx: *RobotsRequestContext = @ptrCast(@alignCast(ctx_ptr));
+    defer ctx.client.allocator.destroy(ctx);
+    defer ctx.buffer.deinit(ctx.client.allocator);
+    defer ctx.client.allocator.free(ctx.robots_url);
+
+    var allowed = true;
+
+    if (ctx.status >= 200 and ctx.status < 400 and ctx.buffer.items.len > 0) {
+        const robots = try ctx.req.robots.robotsFromBytes(
```
karlseguin: I think this is a bit problematic. From what I understand, `robotsFromBytes` can return a recoverable error, e.g. `error.UnexpectedRule`.  So if there's an invalid robots file, all requests to that domain will fail?

I believe the correct solution is to make `robotsFromBytes` only return an error in some unrecoverable case (e.g an arraylist append that fails, signalling OOM). In which case, failing here also makes sense. I also think this is "correct" because I don't think 1 invalid line is supposed to invalid the entire file?

src/browser/Robots.zig
```diff
@@ -0,0 +1,870 @@
...
+                        errdefer allocator.free(duped_value);
+                        try rules.append(allocator, .{ .allow = duped_value });
+                    },
+                    .in_other_entry => {},
+                    .in_wildcard_entry => {
+                        const duped_value = try allocator.dupe(u8, value);
+                        errdefer allocator.free(duped_value);
+                        try wildcard_rules.append(allocator, .{ .allow = duped_value });
+                    },
+                    .not_in_entry => return error.UnexpectedRule,
```
karlseguin: See comment in client.zig: https://github.com/lightpanda-io/browser/pull/1407/changes#r2761960975

src/browser/Robots.zig
```diff
@@ -0,0 +1,870 @@
...
+        }
+    }
+}
+
+fn parseRulesWithUserAgent(
+    allocator: std.mem.Allocator,
+    user_agent: []const u8,
+    bytes: []const u8,
+) ![]const Rule {
+    var rules: std.ArrayList(Rule) = .empty;
```
karlseguin, line 126: Could use an arena from the ArenaPool for the lifetime of this function for all the temp allocations which are done.

src/browser/Robots.zig
```diff
@@ -0,0 +1,870 @@
...
+                        const duped_value = try allocator.dupe(u8, value);
+                        errdefer allocator.free(duped_value);
+                        try wildcard_rules.append(allocator, .{ .disallow = duped_value });
+                    },
+                    .not_in_entry => return error.UnexpectedRule,
+                }
+            },
+        }
+    }
+
```
karlseguin, line 235: Oh man, this confused me. I need a comment

// If we collected rules for our specific user-agent, than those becomes the rules we use for this site
// Otherwise, we fallback to using the rules for the wildcard ('*') user-agent.

src/browser/webapi/net/XMLHttpRequest.zig
```diff
@@ -206,6 +206,7 @@ pub fn send(self: *XMLHttpRequest, body_: ?[]const u8) !void {
         .headers = headers,
         .body = self._request_body,
         .cookie_jar = &page._session.cookie_jar,
+        .robots = &page._session.browser.app.robots,
```
karlseguin: Is there a reason that we pass this into every request rather than having it passed into Http.init?

ALso, this will impact multi-threading work  @mrdimidium

src/browser/Robots.zig
```diff
@@ -0,0 +1,870 @@
...
+                if (matchPattern(pattern, path)) |len| {
+                    // Longest or Last Wins.
+                    if (len >= longest_match_len) {
+                        longest_match_len = len;
+                        is_allowed_result = true;
+                    }
+                }
+            },
+            .disallow => |pattern| {
+                if (pattern.len == 0) continue;
```
karlseguin, line 323: can these just be filtered out when we're building the rule list?

src/browser/Robots.zig
```diff
@@ -0,0 +1,870 @@
...
+pub fn isAllowed(self: *const Robots, path: []const u8) bool {
+    const rules = self.rules;
+
+    var longest_match_len: usize = 0;
+    var is_allowed_result = true;
+
+    for (rules) |rule| {
+        switch (rule) {
+            .allow => |pattern| {
+                if (matchPattern(pattern, path)) |len| {
```
karlseguin, line 314: Is it even necessary to check a pattern if the `pattern.len < longest_match_len` ?

src/browser/Robots.zig
```diff
@@ -0,0 +1,870 @@
...
+    return .{ .rules = rules };
+}
+
+pub fn deinit(self: *Robots, allocator: std.mem.Allocator) void {
+    freeRulesInList(allocator, self.rules);
+    allocator.free(self.rules);
+}
+
+fn matchPatternRecursive(pattern: []const u8, path: []const u8, exact_match: bool) bool {
+    if (pattern.len == 0) return true;
```
karlseguin, line 258: For a deny rule, that's the 3rd time we do this check on the initial pattern.

src/browser/Robots.zig
```diff
@@ -0,0 +1,870 @@
...
+
+    return false;
+}
+
+/// There are rules for how the pattern in robots.txt should be matched.
+///
+/// * should match 0 or more of any character.
+/// $ should signify the end of a path, making it exact.
+/// otherwise, it is a prefix path.
+fn matchPattern(pattern: []const u8, path: []const u8) ?usize {
```
karlseguin, line 293: The pattern matching isn't the most efficient (possibly worth a TODO). I'm wondering if we should pre-process the the robots file beyond just string patterns? There's a memory/time tradeoff here, but as we add multi-threading, but say you were to scrape 1000 product pages from the same website, you'd be matching each line tens of thousands of time. 

Would you prefer to do that in a separate PR?

karlseguin:
I think we need to implement some singleflight mechanism. If I load a page that imports 10 js files from http://www.example.com/ I'm pretty sure it'll fetch robots.txt 10 times (because it isn't in the cache until it's loaded).

I get memory leaks fetching an amazon product page (and, that's a good example where it tries to fetch https://m.media-amazon.com/robots.txt 37 times).

src/browser/webapi/net/XMLHttpRequest.zig
mrdimidium: If this is an immutable value in the new Config that is passed by const reference, then this should not create any new problems.

mrdimidium:
@karlseguin, I don't think we should even try to load anything until we've checked robots.txt. That is, when loading the page, we get `robots.txt` and only then go for index.html and resources.

src/browser/webapi/net/XMLHttpRequest.zig
karlseguin: It's not. It's mutated.

karlseguin:
The implementation in this PR doesn't load anything until robots.txt is loaded, but it's non-blocking. So if you have

```html
<script type=module src="https://www.example.com/s1.js"></script>
<script type=module src="https://www.example.com/s2.js"></script>
```

It won't load s1.js or s2.js until the robots.txt has be loaded, but it will load the robots.txt file twice. And I don' think you can make it globally blocking, because there could be other resources, say on a different domain, that you can make progress towards.

Which is why you'll need a singleflight. A lookup of `robots_url -> []Request`, The first request for the robots_url gets/parses the file and subsequent ones queue their request. When robots is fetched, you don't just proceed to s1.js, but you process all queued requests (s1.js, s2.js, ...).

mrdimidium:
Ah, you mean accessing resources from another domain when parsing HTML. I don't think we should check robots.txt in that case. We should only check robots.txt for the root domain, and I'm sure browsers do that.

Otherwise, it will simply break the loading logic. What happens if the site we're opening approves us in robots, and the CDN doesn't like crawlers?

karlseguin:
I don't have a strong opinion, but at the standup a couple weeks ago both @krichprollsch and @mookums  made the case for `--obey_robots` checking on every domain (which I can kind of understand, it's a signal to behave more like a crawler than a browser). It could be configurable`--obey_robots origin` vs `--obey_robots all`.

I agree if you're only doing `--obey_robots origin` the issue of concurrently fetching the same robots.txt multiple times goes away. But, as is, it's an issue.

src/browser/Robots.zig
mookums: Yeah, a separate PR would be best so this one is just focused on getting the robots functionality up.

src/http/Client.zig
mookums: changed this just to `log.warn` when we have an unexpected rule and it just continues.

mookums:
Didn't touch any of the `matchPattern` stuff because I will do a secondary PR after this one that can focus on fixing any performance issues with it. I want this to be focused on getting proper `robots.txt` behavior.

src/browser/Robots.zig
```diff
@@ -0,0 +1,879 @@
...
+        pub fn hash(_: Context, value: []const u8) u32 {
+            var hasher = std.hash.Wyhash.init(value.len);
+            for (value) |c| {
+                std.hash.autoHash(&hasher, std.ascii.toLower(c));
+            }
+            return @truncate(hasher.final());
+        }
+
+        pub fn eql(_: Context, a: []const u8, b: []const u8) bool {
+            if (a.len != b.len) return false;
```
karlseguin: Looks like `std.ascii.eqlIgnoreCase` already does this check.

src/browser/Robots.zig
```diff
@@ -0,0 +1,879 @@
...
+    const RobotsEntry = union(enum) {
+        present: Robots,
+        absent,
+    };
+
+    pub const RobotsMap = std.HashMapUnmanaged([]const u8, RobotsEntry, struct {
+        const Context = @This();
+
+        pub fn hash(_: Context, value: []const u8) u32 {
+            var hasher = std.hash.Wyhash.init(value.len);
```
karlseguin, line 47: Can be done in chunks. In my test this was about 3x faster, though it'll obviously depend on the length of value:

```zig
    var key = value;
    var buf: [128]u8 = undefined;
    var h = std.hash.Wyhash.init(0);
    while (key.len >= 128) {
      const lower = std.ascii.lowerString(buf[0..], key[0..128]);
      h.update(lower);
      key = key[128..];
    }

    if (key.len > 0) {
      const lower = std.ascii.lowerString(buf[0..key.len], key);
      h.update(lower);
    }
    return @truncate(h.final());
```

src/browser/Robots.zig
karlseguin: TBD in a follow up PR to optimize perfomance

src/browser/Robots.zig
karlseguin: TBD in a follow up PR to look at performance

src/browser/Robots.zig
karlseguin: TBD in a follow up PR to look at performance

src/http/Client.zig
```diff
@@ -246,6 +296,134 @@ pub fn request(self: *Client, req: Request) !void {
...
+        try ctx.client.robot_store.put(ctx.robots_url, robots);
+
+        const path = URL.getPathname(ctx.req.url);
+        allowed = robots.isAllowed(path);
+    } else if (ctx.status == 404) {
+        log.debug(.http, "robots not found", .{ .url = ctx.robots_url });
+        try ctx.client.robot_store.putAbsent(ctx.robots_url);
+    }
+
+    const queued = ctx.client.pending_robots_queue.getPtr(ctx.robots_url) orelse unreachable;
```
karlseguin: You can use `fetchRemove` here (and in robotsErrorCallback). 

Also, `unreachable` has caused us problems, and I'm slowly trying to remove them. Use `@panic("Client.robotsDoneCallback empty queue")` or an lp.assert. Then if it does happen in production, there's a chance we get an error report. And I agree that this _is_ unreachable, but I've been wrong about this too often now to trust myself (or future changes which might break this assumption).

src/http/Client.zig
```diff
@@ -246,6 +296,134 @@ pub fn request(self: *Client, req: Request) !void {
...
+    defer {
+        queued.deinit(ctx.client.allocator);
+        _ = ctx.client.pending_robots_queue.remove(ctx.robots_url);
+    }
+
+    for (queued.items) |queued_req| {
```
karlseguin: Alternative:
```zig
if (allowed) {
    for (queued.items) |req| {
        ctx.client.processRequest(req) catch |e| {
            req.error_callback(req.ctx, e);
        };
    }
} else {
    for (queued.items) |req| {
        log.warn(.http, "blocked by robots", .{ .url = req.url });
        req.error_callback(req.ctx, error.RobotsBlocked);
    }
}
```

karlseguin:
@mookums I've left a few more comments, and resolved some performance things that can be addressed in a follow up PR. All of it can be addressed (or not) in a separate PR; but I'm still getting a memory leak.

The singleflight seems to be working.

krichprollsch:
> Otherwise, it will simply break the loading logic. What happens if the site we're opening approves us in robots, and the CDN doesn't like crawlers?

~It's not 100% clear what to do.~ => After all I do think we must check for each request.

If we follow strictly the [RFC](https://datatracker.ietf.org/doc/html/rfc9309), I understand we must check the robots.txt for all resources.  But I guess this standard has been written when crawlers didn't load page's resources like a browser.

When looking at [Google's interpretation](https://developers.google.com/crawling/docs/robots-txt/robots-txt-spec) they provide this snippet:
```
# This robots.txt file controls crawling of URLs under https://example.com.
# All crawlers are disallowed to crawl files in the "includes" directory, such
# as .css, .js, but Google needs them for rendering, so Googlebot is allowed
# to crawl them.
User-agent: *
Disallow: /includes/

User-agent: Googlebot
Allow: /includes/
```

I understand that in the way Google checks `robots.txt` for all resources. So, if you want Google loads your CSS, you must explicitly allow it.

PR #1411, opened 2026-01-26
more small strings (string.String)

karlseguin:
`Attribute.List` was internally using string.String but only exposing `[]const u8` at its boundaries. This expands the use of string.String to much of the code that uses Attributes.

The main advantage is that WebAPIs can take a `string.String` directly as a parameter. For inputs with a length of <= 12, this avoids any allocation from v8->zig. As a convenience, WebAPIs can also return a `string.String`.

A new discriminatory type, `string.Global` (or `String.Global` if you're already importing `String`) can be used as an input from a v8 call to use the `arena` rather than a page.call_arena.  Previously, functions that wanted to persist a string would need to dupe it. This resulted in 2 copies v8 -> call_arena -> arena. By using a `String.Global` this becomes v8 -> arena  (or, if len <= 12, it is copied directly in the SSO with no allocations).

`String.init` is still used to allocate (if needed) a String.
`String.wrap` wraps an existing string without allocation, thus inheriting the underlying string's lifetime. Importantly, for a string literal <= 12, this can be called at comptime to avoid any runtime overhead.

```zig
pub fn getId(self: *const Element) []const u8 {
    return self.getAttributeSafe(comptime .wrap("id")) orelse "";
}
```

### DIFF
--- a/src/cdp/Node.zig
+++ b/src/cdp/Node.zig
@@ -344,7 +344,7 @@ test "cdp Node: Registry register" {
     var doc = page.window._document;
 
     {
-        const dom_node = (try doc.querySelector(testing.newString("#a1"), page)).?.asNode();
+        const dom_node = (try doc.querySelector(.wrap("#a1"), page)).?.asNode();
         const node = try registry.register(dom_node);
         const n1b = registry.lookup_by_id.get(1).?;
         const n1c = registry.lookup_by_node.get(node.dom).?;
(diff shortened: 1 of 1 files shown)


src/string.zig
```diff
@@ -34,6 +36,48 @@ pub const String = packed struct {
...
+    // literal. This is because, at comptime, we do not have a ptr for data and
+    // thus can't store it.
+    pub fn wrap(input: anytype) String {
+        if (@inComptime()) {
+            const l = input.len;
+            if (l > 12) {
+                @compileError("Comptime string must be <= 12 bytes (SSO only): " ++ input);
+            }
+
+            var content: [12]u8 = @splat(0);
```
nikneym, line 54: Small nitpick; do we need `@splat` here? It'll be overwritten and since the length is also stored, there's no need for sentinel either.

karlseguin: I think we do. When we compare, we compare the first 64 bits..so we need [at least part of] content to be reset.

PR #1429, opened 2026-01-28
update required deps for build from sources

krichprollsch:


krichprollsch:
ok, I goes a bit too fast, some deps are required for v8 build :facepalm: 
I will create another PR...


#1434

PR #1430, opened 2026-01-28
Support HTML inline event listeners

nikneym:
This PR adds support for parsing, setting and getting HTML event listeners; dispatching is not implemented, it will require some changes in `EventManager`. Some commits are cherry-picked from `Image` branch, which led to more broad changes here.

This is now parsed as function and stored in `_element_attr_listeners` on `Page`:
```html
<img onload="(() => console.warn('loaded!'))()" />
```

Getter/setter fashion:
```js
const div = document.getElementById("my-awesome-div");
div.onclick = "...";

div.onclick();
```

### DIFF
--- a/src/browser/EventManager.zig
+++ b/src/browser/EventManager.zig
@@ -22,7 +22,6 @@ const builtin = @import("builtin");
 const log = @import("../log.zig");
 const String = @import("../string.zig").String;
 
-const lp = @import("lightpanda");
 const js = @import("js/js.zig");
 const Page = @import("Page.zig");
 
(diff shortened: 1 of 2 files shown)


src/browser/EventManager.zig
```diff
@@ -22,6 +22,7 @@ const builtin = @import("builtin");
 const log = @import("../log.zig");
 const String = @import("../string.zig").String;
 
+const lp = @import("lightpanda");
```
karlseguin:
```suggestion
```

src/browser/webapi/Element.zig
```diff
@@ -48,6 +48,128 @@ pub const ClassListLookup = std.AutoHashMapUnmanaged(*Element, *collections.DOMT
 pub const RelListLookup = std.AutoHashMapUnmanaged(*Element, *collections.DOMTokenList);
 pub const ShadowRootLookup = std.AutoHashMapUnmanaged(*Element, *ShadowRoot);
 pub const AssignedSlotLookup = std.AutoHashMapUnmanaged(*Element, *Html.Slot);
+/// Better to discriminate it since not directly a pointer int.
+///
+/// See `calcAttrListenerKey` to obtain one.
+pub const AttrListenerKey = u64;
```
```suggestion
const AttrListenerKey = u64;
```

karlseguin (commented):
Is the goal to merge this first, then use the changes here for the Image?

nikneym:
> Is the goal to merge this first, then use the changes here for the Image?

Indeed, that's the idea. Changes made here were not in scope of Image anyway.

PR #1440, opened 2026-01-29
Lazily evaluate inline event listeners

nikneym:


src/browser/webapi/element/Html.zig
```diff
@@ -20,6 +20,7 @@ const std = @import("std");
 const lp = @import("lightpanda");
 const js = @import("../../js/js.zig");
 const reflect = @import("../../reflect.zig");
+const log = @import("../../../log.zig");
```
karlseguin, line 22:
```suggestion
```

karlseguin:
Whenever an element is created (via the parser or dynamically), we, check each attribute to see if it's one of ~90 events. We then dupe the value and store it in a lookup. If that attribute is requested, we fetch it from that lookup.

But we already have a lookup of attribute=>value. So why do we need a second one?

nikneym:
> Whenever an element is created (via the parser or dynamically), we, check each attribute to see if it's one of ~90 events.

Our current code first checks if attribute starts with `on` prefix to reduce number of checks:
```zig
const has_on_prefix = @as(u16, @bitCast([2]u8{ name.ptr[0], name.ptr[1 % name.len] })) == asUint("on");
// We may have found an event handler.
if (has_on_prefix) {
    // Large number of checks happen...
}
```

> But we already have a lookup of attribute=>value. So why do we need a second one?

I agree, our initial discussion was to prefer attribute list for not-evaulated inline handlers. Browsers have some interesting behavior when it comes to inline handlers though, consider such example:
```js
div.onload = () => console.warn("via setter");
console.warn(div.getAttribute("onload"));
```

If we evaluate `onload` lazily, meaning store it in attribute list and create function later, The `getAttribute` part should be printing the function. FF and Chrome print `null` though since both use different map for storage.

karlseguin:
I don't understand. What test cases are we missing? This is what we talked about, and it passes the current tests:
https://github.com/lightpanda-io/browser/compare/main..26387db56e9eb67b60cae512f3b4b0076b8efcbf

src/browser/webapi/Element.zig
```diff
@@ -59,101 +59,101 @@ pub const AttrListenerLookup = std.AutoHashMapUnmanaged(AttrListenerKey, js.Func
 /// Enum of known event listeners; increasing the size of it (u7)
 /// can cause `AttrListenerKey` to behave incorrectly.
 pub const KnownListener = enum(u7) {
-    abort,
```
karlseguin, line 62: If `KnownListener` and `calcAttrListenerKey` are in `Elemenet.zig` then I would expect all the getters and setters to also be in `Element.zig`.

If the getters and setters correctly belong on `HTMLElement`, why aren't KnownListener and `calcAttrListenerKey` defined there? 

From what I can understand, these are "GlobalEventHandlers", and they specifically _do not_ relate to element? They relate to HtmlElement, Window and Document?

src/browser/webapi/element/Html.zig
```diff
@@ -332,764 +333,798 @@ pub fn click(self: *HtmlElement, page: *Page) !void {
     try page._event_manager.dispatch(self.asEventTarget(), event.asEvent());
 }
 
+fn getAttributeFunction(
+    self: *HtmlElement,
+    comptime listener_type: Element.KnownListener,
```
karlseguin: Is there any value in making this comptime?

src/browser/webapi/element/Html.zig
```diff
@@ -332,764 +333,798 @@ pub fn click(self: *HtmlElement, page: *Page) !void {
...
+) ?js.Function.Global {
+    const element = self.asElement();
+    if (page.getAttrListener(element, listener_type)) |cached| {
+        return cached;
+    }
+
+    const attr = element.getAttributeSafe(.wrap(@tagName(listener_type))) orelse return null;
+    const callback = page.js.stringToPersistedFunction(attr) catch |err| {
+        // Not a valid expression; log this to find out if its something we should be supporting.
+        log.warn(.unknown_prop, "Page.getAttrListener", .{
```
karlseguin: it isn't an "unknown_prop", is it?  Isn't this just user javascript that doesn't compile?  Should be under the `.js` category.

src/browser/webapi/element/Html.zig
```diff
@@ -332,764 +333,798 @@ pub fn click(self: *HtmlElement, page: *Page) !void {
...
+            .expression = attr,
+            .err = err,
+        });
+
+        return null;
+    };
+
+    page.setAttrListener(element, listener_type, callback) catch {
+        // This is fine :tm: We're out of memory for cache.
+        // I don't want to make all getters "!?" just because of this honestly.
```
karlseguin: Why not?

src/browser/webapi/Element.zig
nikneym: There are event handlers that're only for `Element` in that enum; namely `onfullscreenchange` and `onfullscreenerror` (incorrectly put in `HTMLElement` ATM). So this isn't only tied to `HTMLElement` actually. I can offer 3 options, pick your poison:

- Move everything to `HTMLElement` anyway (`KnownListener`, `AttrListenerLookup`, `calcAttrListenerKey` etc.).
- Since `KnownListener` list is a unified list of all possible global handlers, move everything related to it back to `Page`.
- Have separate `KnownListener`s for each type (for `Element`, `HTMLElement` and in `XMLHttpRequest` at some point for example) and `calcAttrListenerKey` takes `u7` instead; issues would arise if 2 types with `KnownListener` are in same prototype chain so types should keep track if such enum value taken.

karlseguin: I suggest we create a new file `global_event_handlers.zig`, since that's what they are, and they aren't specific to element/htmlElement/document/window.  But ya, it would help if they were defined in the correct place because, as-is, it's definedin Element, but that's the one place it _isn't_ used.

PR #1444, opened 2026-01-30
Rework Inspector usage

karlseguin:
V8's inspector world is made up of 4 components: Inspector, Client, Channel and Session. Currently, we treat all 4 components as a single unit which is tied to the lifetime of CDP BrowserContext - or, loosely speaking, 1 "Inspector Unit" per page / v8::Context.

According to https://web.archive.org/web/20210622022956/https://hyperandroid.com/2020/02/12/v8-inspector-from-an-embedder-standpoint/ and conversation with Gemini, it's more typical to have 1 inspector per isolate. The general breakdown is the Inspector is the top-level manager, the Client is our implementation which control how the Inspector works (its function we expose that v8 calls into). These should be tied to the Isolate. Channels and Sessions are more closely tied to Context, where the Channel is v8->zig and the Session us zig->v8.

This PR does a few things
1 - It creates 1 Inspector and Client per Isolate (Env.js) 2 - It creates 1 Session/Channel per BrowserContext 3 - It merges v8::Session and v8::Channel into Inspector.Session 4 - It moves the Inspector instance directly into the Env 5 - BrowserContext interacts with the Inspector.Session, not the Inspector

4 is arguably unnecessary with respect to the main goal of this commit, but the end-goal is to tighten the integration. Specifically, rather than CDP having to inform the inspector that a context was created/destroyed, the Env which manages Contexts directly (https://github.com/lightpanda-io/browser/pull/1432) and which now has direct access to the Inspector, is now equipped to keep this in sync.

### DIFF
--- a/src/browser/js/Inspector.zig
+++ b/src/browser/js/Inspector.zig
@@ -23,99 +23,76 @@ const v8 = js.v8;
 const TaggedOpaque = @import("TaggedOpaque.zig");
 
 const Allocator = std.mem.Allocator;
-const RndGen = std.Random.DefaultPrng;
 
 const CONTEXT_GROUP_ID = 1;
 const CLIENT_TRUST_LEVEL = 1;
+const IS_DEBUG = @import("builtin").mode == .Debug;
 
+// Inspector exists for the lifetime of the Isolate/Env. 1 Isolate = 1 Inspector.
+// It combines the v8.Inspector and the v8.InspectorClientImpl. The v8.InspectorClientImpl
+// is our own implementation that fulfills the InspectorClient API, i.e. it's the
+// mechanism v8 provides to let us tweak how the inspector works. For example, it
+// Below, you'll find a few pub export fn v8_inspector__Client__IMPL__XYZ functions
+// which is our implementation of what the v8::Inspector requires of our Client
+// (not much at all)
 const Inspector = @This();
 
-handle: *v8.Inspector,
+unique_id: i64,
 isolate: *v8.Isolate,
-client: Client,
-channel: Channel,
-session: Session,
-rnd: RndGen = RndGen.init(0),
+handle: *v8.Inspector,
+client: *v8.InspectorClientImpl,
 default_context: ?v8.Global,
+session: ?Session,
 
-// We expect allocator to be an arena
-// Note: This initializes the pre-allocated inspector in-place
-pub fn init(self: *Inspector, isolate: *v8.Isolate, ctx: anytype) !void {
-    const ContextT = @TypeOf(ctx);
+pub fn init(allocator: Allocator, isolate: *v8.Isolate) !*Inspector {
+    const self = try allocator.create(Inspector);
+    errdefer allocator.destroy(self);
 
-    const Container = switch (@typeInfo(ContextT)) {
-        .@"struct" => ContextT,
-        .pointer => |ptr| ptr.child,
-        .void => NoopInspector,
-        else => @compileError("invalid context type"),
-    };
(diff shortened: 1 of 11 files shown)


src/cdp/cdp.zig
```diff
@@ -678,18 +681,6 @@ pub fn BrowserContext(comptime CDP_T: type) type {
             };
         }
 
-        // debugger events
-
-        pub fn onRunMessageLoopOnPause(_: *anyopaque, _: u32) void {
```
krichprollsch, line 683: I know this is empty and not used yet, but I expect implement them at some point to add chrome inspector compatibility. So I think we could keep these placeholder.

src/browser/js/Inspector.zig
```diff
@@ -254,20 +186,109 @@ pub const RemoteObject = struct {
...
+        defer v8.v8__HandleScope__DESTRUCT(&hs);
+
         v8.v8_inspector__Session__dispatchProtocolMessage(
             self.handle,
             isolate,
             msg.ptr,
             msg.len,
         );
+
+        v8.v8__Isolate__PerformMicrotaskCheckpoint(isolate);
```
krichprollsch, line 245: 👍

src/browser/js/Inspector.zig
```diff
@@ -254,20 +186,109 @@ pub const RemoteObject = struct {
...
+            .ctx = ctx,
+            .handle = handle,
+            .channel = channel,
+            .inspector = inspector,
+            .onResp = Container.onInspectorResponse,
+            .onNotif = Container.onInspectorEvent,
+        };
+    }
+
+    fn deinit(self: *const Session) void {
```
krichprollsch, line 227: If a session is tight to a ctx, maybe we should call contextDestroyed here.

src/cdp/cdp.zig
```diff
@@ -409,27 +413,28 @@ pub fn BrowserContext(comptime CDP_T: type) type {
...
-            self.cdp.browser.env.runMicrotasks();
+            browser.env.runMicrotasks();
 
             // resetContextGroup detach the inspector from all contexts.
             // It append async tasks, so we make sure we run the message loop
             // before deinit it.
-            self.inspector.resetContextGroup();
-            self.session.browser.runMessageLoop();
-            self.inspector.deinit();
+            browser.env.inspector.?.resetContextGroup();
```
krichprollsch, line 435: In the future, if we have multiple browser contexts for a CDP conn, calling resetContext here could have impact to others inspector's sessions...

src/cdp/cdp.zig
karlseguin: FWIW, there's still the C hook in https://github.com/lightpanda-io/browser/pull/1444/changes#diff-ae2549a80312068902a9682af627e9c17952b772fbef4a87b95463e97ab85872R388

src/cdp/cdp.zig
karlseguin: I *think* that's where the context group id comes into play? Instead of hard-coding 1, we'd use an incrementing id per browser context...(question mark)

src/browser/js/Inspector.zig
karlseguin: Yes, with this change, and the previously merged removal of ExecutionWorld (https://github.com/lightpanda-io/browser/pull/1432), I want to try to cleanup the createContext (and, hopefully, destroyContext becomes trivial).

src/cdp/cdp.zig
krichprollsch: oh you are right I think.

krichprollsch (approved):

PR #1446, opened 2026-01-30
When loading a module, make sure we have a module_promise

karlseguin:
Currently, when loading a module, if the module is found in the cache, it's immediately returned. However, that can result in a timing issue where the module is cached, but not evaluated, and doesn't have an associated promise.

This commit tries to ensure a module is always evaluated and that the cache entry has a module promise.

This might fix an crash handler issue. I couldn't reproduce the issue though. I believe it requires specific timing which is hard to reproduce in a test.

### DIFF
--- a/src/browser/js/Context.zig
+++ b/src/browser/js/Context.zig
@@ -324,10 +324,22 @@ pub fn module(self: *Context, comptime want_result: bool, local: *const js.Local
             if (gop.found_existing) {
                 if (gop.value_ptr.module) |cache_mod| {
                     if (gop.value_ptr.module_promise == null) {
+                        // This an usual case, but it can happen if a module is
+                        // first asynchronously requested and then synchronously
+                        // requested as a child of some root import. In that case,
+                        // the module may not be instantiated yet (so we have to
+                        // do that). It might not be evaluated yet. So we have
+                        // to do that too. Evaluation is particularly important
+                        // as it sets up our cache entry's module_promise.
+                        // It appears that v8 handles potential double-instantiated
+                        // and double-evaluated modules safely. The 2nd instantiation
+                        // is a no-op, and the second evaluation returns the same
+                        // promise.
                         const mod = local.toLocal(cache_mod);
-                        if (mod.getStatus() == .kInstantiated) {
-                            return self.evaluateModule(want_result, mod, url, true);
+                        if (mod.getStatus() == .kUninstantiated and try mod.instantiate(resolveModuleCallback) == false) {
+                            return error.ModuleInstantiationError;
                         }
+                        return self.evaluateModule(want_result, mod, url, true);
                     }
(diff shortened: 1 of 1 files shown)


src/browser/js/Context.zig
```diff
@@ -322,7 +322,13 @@ pub fn module(self: *Context, comptime want_result: bool, local: *const js.Local
         if (cacheable) {
             gop = try self.module_cache.getOrPut(arena, url);
             if (gop.found_existing) {
-                if (gop.value_ptr.module != null) {
+                if (gop.value_ptr.module) |cache_mod| {
+                    if (gop.value_ptr.module_promise == null) {
+                        const mod = local.toLocal(cache_mod);
+                        if (mod.getStatus() == .kInstantiated) {
```
krichprollsch: So I guess it happens when a module is required again during its evaluation?
Can't we have the risk to evaluate it twice then? Or having a kind of loop 🤔

Or does the status check protect us from this?

karlseguin: If it was being evaluated, the state would be `.kEvaluating`. I believe by limiting this evaluation to `.kInstantiated`, it will prevent the case that you're talking about.

krichprollsch: but in this case, won't you get the same crash then?

karlseguin: The flow that I suspect is:

1 - The module is dynamically imported, this creates a cache entry with no module and no module_promise, and starts an async fetch.

2 - Before `dynamicModuleSourceCallback` fires (from step 1 above), the same module is imported as a child of a call graph, i.e. via `resolveModuleCallback`. Here the module is compiled, but never evaluated (we only evaluate the root module). This is where things start to go sour. Our cache entry now has a `module`, but no `module_promise`.

3 - The async fetch completes and calls `dynamicModuleSourceCallback` which call `Context.module`. This returns early (the module is already cached thanks to step 2). But it then calls `resolveDynamicModule` which (a) has a module and (b) no module_promise.

Our fix works because, if Context.module finds the cached module (from step 2), it now also checks for the `module_promise`. If it doesn't find it, it evaluates the module (which sets it).

I've since expanded the code to handle more intermediary states.

The original PR had:

```zig
if (gop.value_ptr.module_promise == null) {
    const mod = local.toLocal(cache_mod);
    if (mod.getStatus() == .kInstantiated) {
        return self.evaluateModule(want_result, mod, url, true);
    }
}
```

But now the code is:

```zig
if (gop.value_ptr.module_promise == null) {
    const mod = local.toLocal(cache_mod);
    if (mod.getStatus() == .kUninstantiated and try mod.instantiate(resolveModuleCallback) == false) {
        return error.ModuleInstantiationError;
    }
    return self.evaluateModule(want_result, mod, url, true);
}
```

It seems that v8 handles double-instantiation and double-evaluations safely.

PR #1456, opened 2026-02-02
Rework `Scheduler`

nikneym:
This PR reworks `Scheduler` by;
- replacing priority queue with [pairing heap](https://en.wikipedia.org/wiki/Pairing_heap),
- introducing `Scheduler.once` and `Scheduler.after` functions,
- adding a memory pool to recycle tasks,
- using same `now` timestamp for both trees,
- introducing `ScheduleInterface`; which is kind of a DSL to write scheduler actions and finalizers.

If a task is being scheduled via `Scheduler.once`, the passed type must have `action` callback implemented. Here's the function signature:
```zig
pub fn action(scheduler: *Scheduler, context: *T) !void {
    // ...
}
```

If a task is being scheduled via `Scheduled.afer` instead, the passed type must have `action` callback with this signature:
```zig
pub fn action(scheduler: *Scheduler, context: *T) !AfterAction {
    // ...
    // Repeat this `action` after 200ms:
    return .repeat(200);
    // Don't repeat the action.
    return .dont_repeat;
}
```

Both variants can also have `finalize` callback:
```zig
pub fn finalize(context: *T) void {
    // ...
}
```

The "finalizers" will be fired before `Scheduler` itself is deinitialized, as before.

### DIFF
--- a/src/browser/Scheduler.zig
+++ b/src/browser/Scheduler.zig
@@ -273,13 +273,14 @@ pub fn after(
 }
 
 pub fn run(self: *Scheduler) !?u64 {
-    self.runTasks(.low);
-    return self.runTasks(.high);
+    const now = milliTimestamp(.monotonic);
+
+    self.runTasks(.low, now);
+    return self.runTasks(.high, now);
 }
 
 /// Runs events of the desired tree.
-fn runTasks(self: *Scheduler, comptime prio: Priority) if (prio == .low) void else ?u64 {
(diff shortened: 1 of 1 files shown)


krichprollsch:
I suppose #1426 will break this one.

nikneym:
> I suppose https://github.com/lightpanda-io/browser/pull/1426 will break this one.

So far both play well; other PR moves `Scheduler` from `Page` to `Context`, which is fine for this one.

PR #1459, opened 2026-02-03
Add a dedicated browser_context and page_arena to CDP.

karlseguin:
The BrowserContext currently uses 3 arenas:
1 - Command-specific, which is like the call_arena, but for the processing of a
    single CDP command
2 - Notification-specific, which is similar, but for the processing of a single
    internal notification event
3 - Arena, which is just the session arena and lives for the duration of the
    BrowseContext/Session

This is pretty coarse and can results in significant memory accumulation if a browser context is re-used for multiple navigations.

This commit introduces 3 changes:
1 - Rather than referencing Session.arena, the BrowerContext.arena is now its
    own arena. This doesn't really change anything, but it does help keep things
    a bit better separated.

2 - Introduces a page_arena (not to be confused with Page.arena). This arena
    exists for the duration of a 1 page, i.e. it's cleared when the
    BrowserContext receives the page_created internal notification. The
    `captured_responses` now uses this arena, which means captures only exist
    for the duration of the current page. This appears to be consistent with
    how chrome behaves (In fact, Chrome seems even more aggressive and doesn't
    appear to make any guarantees around captured responses). CDP refers to this
    lifetime as a "renderer" and has an experimental message, which we don't
    support, `Network.configureDurableMessages` to control this.

3 - Isolated Worlds are now more self contained with an arena from the ArenaPool.

There are currently 2 places where the BrowserContext.arena is still used: 1 - the isolated_world list
2 - the custom headers

Although this could be long lived, I believe the above is ok. We should just really think twice whenever we want to use it for anything else.

### DIFF
--- a/src/cdp/cdp.zig
+++ b/src/cdp/cdp.zig
@@ -760,10 +760,7 @@ const IsolatedWorld = struct {
     grant_universal_access: bool,
 
     pub fn deinit(self: *IsolatedWorld) void {
-        if (self.context) |ctx| {
-            self.browser.env.destroyContext(ctx);
-            self.context = null;
-        }
+        self.removeContext() catch {};
         self.browser.arena_pool.release(self.arena);
     }
 


src/cdp/cdp.zig
```diff
@@ -734,20 +753,23 @@ pub fn BrowserContext(comptime CDP_T: type) type {
...
     name: []const u8,
-    env: *js.Env,
     context: ?*js.Context = null,
     grant_universal_access: bool,
 
     pub fn deinit(self: *IsolatedWorld) void {
         if (self.context) |ctx| {
-            self.env.destroyContext(ctx);
+            self.browser.env.destroyContext(ctx);
             self.context = null;
```
krichprollsch: What about calling `self.removeContext()` instead of duplicating?

PR #1461, opened 2026-02-03
Fix [I hope] blocking auth interception

karlseguin:
On a blocking request that requires authentication, we now handle the two cases correctly:
1 - if the request is aborted, we don't continue processing (if we did, that
    would result in (a) transfer.deinit being called twice and (b) the callbacks
    being called twice

2 - if the request is "continue", we queue the transfer to be re-issued, as
    opposed to just processing it as-is. We have to queue it because we're
    currently inside a process loop and it [probaby] isn't safe to re-enter it.
    By using the queue, we wait until the next call to `tick` to re-issue the
    request.

### DIFF
--- a/src/http/Client.zig
+++ b/src/http/Client.zig
@@ -17,6 +17,8 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 const std = @import("std");
+const lp = @import("lightpanda");
+
 const log = @import("../log.zig");
 const builtin = @import("builtin");
 
@@ -235,7 +237,7 @@ pub fn request(self: *Client, req: Request) !void {
     if (req.blocking == false) {
         // The request was interecepted, but it isn't a blocking request, so we
         // dont' need to block this call. The request will be unblocked
-        // asynchronously via eitehr continueTransfer or abortTransfer
+        // asynchronously via either continueTransfer or abortTransfer
         return;
     }
 
@@ -458,7 +460,7 @@ pub fn disableTlsVerify(self: *const Client) !void {
     }
 }
 
-fn makeRequest(self: *Client, handle: *Handle, transfer: *Transfer) !void {
+fn makeRequest(self: *Client, handle: *Handle, transfer: *Transfer) anyerror!void {
     const conn = handle.conn;
     const easy = conn.easy;
     const req = &transfer.req;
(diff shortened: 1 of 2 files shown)


src/http/Client.zig
```diff
@@ -587,7 +589,18 @@ fn processMessages(self: *Client) !bool {
                         continue;
                     }
 
-                    _ = try self.waitForInterceptedResponse(transfer);
+                    self.endTransfer(transfer);
```
krichprollsch: so we can directly call `endTransfer` when entering in `if (wait_for_interception) {` isn't it?

karlseguin: You mean calling `endTransfer(transfer)`  and then calling `waitForInterceptedResponse` ? 

Yes, `endTransfer` removes the easy from the multi and puts the easy back into the pool and `waitForInterceptedResponse` just blocks on the CDP socket waiting for a message from the client. When we queue the transfer back up, it'll eventually (on `tick`) re-associate a transfer with a handle.

krichprollsch: I mean we can remove the  ` if (!transfer.req.blocking) {` test like:

```zig
                if (wait_for_interception) {
                    self.intercepted += 1;
                    if (comptime IS_DEBUG) {
                        log.debug(.http, "wait for auth interception", .{ .intercepted = self.intercepted });
                    }
                    transfer._intercept_state = .pending;

                    // the request is put on hold to be intercepted.
                    // In this case we ignore callbacks for now.
                    // Note: we don't deinit transfer on purpose: we want to keep
                    // using it for the following request
                    self.endTransfer(transfer);

                    if (try self.waitForInterceptedResponse(transfer)) {
                        // we've been asked to continue with the request
                        // we can't process it here, since we're already inside
                        // a process, so we need to queue it and wait for the
                        // next tick.
                        self.queue.append(&transfer._node);
                    } else {
                        // aborted, already cleaned up
                    }

                    continue;
                }
```

krichprollsch: oh my bad we still have to continue on non-blocking request right?
```zig
                if (wait_for_interception) {
                    self.intercepted += 1;
                    if (comptime IS_DEBUG) {
                        log.debug(.http, "wait for auth interception", .{ .intercepted = self.intercepted });
                    }
                    transfer._intercept_state = .pending;

                    // the request is put on hold to be intercepted.
                    // In this case we ignore callbacks for now.
                    // Note: we don't deinit transfer on purpose: we want to keep
                    // using it for the following request
                    self.endTransfer(transfer);

                    if (!transfer.req.blocking) {
                        continue;
                    }

                    if (try self.waitForInterceptedResponse(transfer)) {
                        // we've been asked to continue with the request
                        // we can't process it here, since we're already inside
                        // a process, so we need to queue it and wait for the
                        // next tick.
                        self.queue.append(&transfer._node);
                    } else {
                        // aborted, already cleaned up
                    }

                    continue;
                }

krichprollsch (approved):

PR #1465, opened 2026-02-03
Telemetry without notifications

mrdimidium:
To ensure notifications work in a multithreaded environment, this PR moves Notification from the global scope to the BrowserContext level (which roughly corresponds to a CDP connection).

Since there can now be multiple notifications, I've removed telemetry from events and instead write events directly to the global Telemetry.


karlseguin:
I don't know why I it via notifications initially...Calling telemetry directly is so much simpler.

This PR appears to contain another set of changes: moving the notification to the session and injecting it in the request. I think the 2nd part of that is nice, but why the first part?  It's to eventually support multiple pages per browser, yes? I've heard this mentioned a few times, but it isn't clear why/why/if we're ever going to need/want that.

mrdimidium:
@karlseguin, I moved Notification from the global scope to ensure it wouldn't be accessed by multiple connections, even accidentally.

For multiple pages, this works transparently: one Notification per BrowserContext for one CDP connection, and all nested tabs.

But for now it's necessary to clearly separate ownership

karlseguin:
Can be merged when the conflict is resolved.

PR #1475, opened 2026-02-05
Change the listener lookup from element, to (element,type).

karlseguin:
Right now, to do anything with listeners, we need to (a) lookup the listeners registered with the element and (b) walk the linked list to find listeners of the correct type.

There's no operations that behave on all of an element's listeners, e.g. there's no element.removeAllListeners(). So, we can optimize the code, and generally make it simpler to read, by changing our key to include the type (along with the element).

This is an optimization on its own (and makes the code simpler), but it also makes it more palatable to do a pre-listener check to avoid creating events when no listener even exists; if we decide to implement that.

### DIFF
--- a/src/browser/EventManager.zig
+++ b/src/browser/EventManager.zig
@@ -34,20 +34,20 @@ const Allocator = std.mem.Allocator;
 const IS_DEBUG = builtin.mode == .Debug;
 
 const EventKey = struct {
-    element: usize,
+    event_target: usize,
     type_string: String,
 };
 
 const EventKeyContext = struct {
     pub fn hash(_: @This(), key: EventKey) u64 {
         var hasher = std.hash.Wyhash.init(0);
-        hasher.update(std.mem.asBytes(&key.element));
+        hasher.update(std.mem.asBytes(&key.event_target));
         hasher.update(key.type_string.str());
         return hasher.final();
     }
 
     pub fn eql(_: @This(), a: EventKey, b: EventKey) bool {
-        return a.element == b.element and a.type_string.eql(b.type_string);
+        return a.event_target == b.event_target and a.type_string.eql(b.type_string);
     }
 };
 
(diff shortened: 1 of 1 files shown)


src/browser/EventManager.zig
```diff
@@ -33,13 +33,36 @@ const Allocator = std.mem.Allocator;
...
 
+const EventKey = struct {
+    element: usize,
+    type_string: String,
+};
+
+const EventKeyContext = struct {
+    pub fn hash(_: @This(), key: EventKey) u64 {
+        var hasher = std.hash.Wyhash.init(0);
+        hasher.update(std.mem.asBytes(&key.element));
```
nikneym: There's a chance of making this even smaller (because of unused bits) I'm fine with changes though.

In fact, you can even embed some of the `Listener` fields if desired.

src/browser/EventManager.zig
```diff
@@ -33,13 +33,36 @@ const Allocator = std.mem.Allocator;
 
 const IS_DEBUG = builtin.mode == .Debug;
 
+const EventKey = struct {
+    element: usize,
```
nikneym: I think we can extend this beyond `Element` since there are objects that can dispatch events (like `XMLHttpRequest`) that are not elements.

nikneym:
Not a concern of this PR but do you think `list_pool` is benefitable? `std.DoublyLinkedList` itself is 16 bytes, internal `Node` of `heap.MemoryPool` adds 16 bytes (`?*next`) to chain nodes so it ends up being 32 bytes each; essentially double the size.

src/browser/EventManager.zig
karlseguin: `element` is a dumb field name. Renamed to `event_target`

karlseguin:
> Not a concern of this PR but do you think `list_pool` is benefitable? `std.DoublyLinkedList` itself is 16 bytes, internal `Node` of `heap.MemoryPool` adds 16 bytes (`?*next`) to chain nodes so it ends up being 32 bytes each; essentially double the size.

You mean, why use `list_pool` for such a small amount of memory? I'm working under the vague impression that virtually any pressure removed from the general allocator (which, I'm pretty sure, has its own per-allocation overhead) is a win...less fragmentation, less thread-safety concern, and possibly less overhead. Though I don't know this for sure.

src/browser/EventManager.zig
karlseguin: Too clever.

PR #1479, opened 2026-02-06
Add Finalizers to events

karlseguin:
At a high level, this does for Events what was recently done for XHR, Fetch and Observers. Events are self-contained in their own arena from the ArenaPool and are registered with v8 to be finalized.

But events are more complicated than those other types. For one, events have a prototype chain. (XHR also does, but it's always the top-level object that's created, whereas it's valid to create a base Event or something that inherits from Event). But the _real_ complication is that Events, unlike previous types, can be created from Zig or from V8.

This is something that Fetch had to deal with too, because the Response is only given to V8 on success. So in Fetch, there's a period of time where Zig is solely responsible for the Response, until it's passed to v8. But with events it's a lot more subtle.

There are 3 possibilities:
1 - An Event is created from v8. This is the simplest, and it simply becomes a
    a weak reference for us. When v8 is done with it, the finalizer is called.
2 - An Event is created in Zig (e.g. window.load) and dispatched to v8. Again
    we can rely on the v8 finalizer.
3 - An event is created in Zig, but not dispatched to v8 (e.g. there are no
    listeners), Zig has to release the event.

(It's worth pointing out that one thing that still keeps this relatively straightforward is that we never hold on to Events past some pretty clear point)

Now, it would seem that #3 is the only issue we have to deal with, and maybe we can do something like:

```
if (event_manager.hasListener("load", capture)) {
   try event_manager.dispatch(event);
} else {
   event.deinit();
}
```

In fact, in many cases, we could use this to optimize not even creating the event:

```
if (event_manager.hasListener("load, capture)) {
   const event = try createEvent("load", capture);
   try event_manager.dispatch(event);
}
```

And that's an optimization worth considering, but it isn't good enough to properly manage memory. Do you see the issue? There could be a listener (so we think v8 owns it), but we might never give the value to v8. Any failure between hasListener and actually handing the value to v8 would result in a leak.

To solve this, the bridge will now set a _v8_handover flag (if present) once it has created the finalizer_callback entry. So dispatching code now becomes:

```
const event = try createEvent("load", capture);
defer if (!event._v8_handover) event.deinit(false);
try event_manager.dispatch(event);
```

The v8 finalizer callback was also improved. Previously, we just embedded the pointer to the zig object. In the v8 callback, we could cast that back to T and call deinit. But, because of possible timing issues between when (if) v8 calls the finalizer, and our own cleanup, the code would check in the context to see if the ptr was still valid. Wait, what? We're using the ptr to get the context to see if the ptr is valid?

We now store a pointer to the FinalizerCallback which contains the context. So instead of something stupid like:

```
// note, if the identity_map doesn't contain the value, then value is likely
// invalid, and value.page will segfault
value.page.js.identity_map.contains(@intFromPtr(value))
```

We do:
```
if (fc.ctx.finalizer_callbacks.contains(@intFromPtr(fc.value)) {
  // fc.value is safe to use
}
```

### DIFF
--- a/src/browser/js/Local.zig
+++ b/src/browser/js/Local.zig
@@ -206,7 +206,7 @@ pub fn mapZigInstanceToJs(self: *const Local, js_obj_handle: ?*const v8.Object, 
                 //    if (resolved.finalizer_from_v8) |finalizer| {
                 // But that's a runtime check.
                 // Instead, we check if the base has finalizer. The assumption
-                // here is that if a resolve type has a finalizer, than the base
+                // here is that if a resolve type has a finalizer, then the base
                 // should have a finalizer too.
                 const fc = try ctx.createFinalizerCallback(gop.value_ptr.*, resolved.ptr, resolved.finalizer_from_zig.?);
                 {


src/browser/js/Local.zig
```diff
@@ -198,21 +198,28 @@ pub fn mapZigInstanceToJs(self: *const Local, js_obj_handle: ?*const v8.Object,
...
-                    std.debug.assert(resolved.class_id == JsApi.Meta.class_id);
+                // It would be great if resolved knew the resolved type, but I
+                // can't figure out how to make that work, since it depends on
+                // the [runtime] `value`.
+                // We need the resolved finalizer, which we have in resolved.
+                // The above if statement would be more clear as:
+                //    if (resolved.finalizer_from_v8) |finalizer| {
+                // But that's a runtime check.
+                // Instead, we check if the base has finalizer. The assumption
+                // here is that if a resolve type has a finalizer, than the base
```
krichprollsch:
```suggestion
                // here is that if a resolve type has a finalizer, then the base
```

krichprollsch (approved):

PR #1482, opened 2026-02-06
Add failing `document.documentElement` `load` event test

nikneym:


krichprollsch:
Are we agree this test should actually fail?

nikneym:
Once #1487 merged, this test should pass.

Though we have to add `testing.expectEqual(true, true)` at the end to satisfy sync nature of testing framework.

PR #1486, opened 2026-02-06
Change hash generation of global event handlers

nikneym:
Small change inspired from #1475.

### DIFF
--- a/src/browser/webapi/global_event_handlers.zig
+++ b/src/browser/webapi/global_event_handlers.zig
@@ -23,6 +23,8 @@ const js = @import("../js/js.zig");
 
 const EventTarget = @import("EventTarget.zig");
 
+const IS_DEBUG = @import("builtin").mode == .Debug;
+
 const Key = struct {
     target: *EventTarget,
     handler: Handler,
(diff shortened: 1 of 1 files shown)


src/browser/webapi/global_event_handlers.zig
```diff
@@ -23,13 +23,41 @@ const js = @import("../js/js.zig");
...
+    handler: Handler,
 
-/// Use `calculateKey` to create a key.
-pub const Lookup = std.AutoHashMapUnmanaged(Key, js.Function.Global);
+    /// Fuses `target` pointer and `handler` enum; used at hashing.
+    /// NEVER use a fusion to retrieve a pointer back. Portability is not guaranteed.
+    /// See `Context.hash`.
+    fn fuse(self: *const Key) u64 {
+        // Check if we have 3 bits available from alignment of 8.
+        lp.assert(@alignOf(EventTarget) == 8, "Key.fuse: incorrect alignment", .{
```
karlseguin: I always struggle with what is called comptime vs runtime.

I know the equality check will be done at comptime. But does that mean at runtime we end up with:

`lp.assert(true, "Key.fuse: incorrect alignment", ...)` ?  And does the compiler then see `lp.assert` is it `(!ok) {...}` and strip out the entire call?

nikneym: > does the compiler then see lp.assert is it (!ok) {...} and strip out the entire call?

That's intuitively what I also expect... We can always wrap these in `comptime if` if you like, though.

PR #1487, opened 2026-02-06
Changes to trigger `documentElement`s load event

nikneym:
This fixes the triggering issue; though the order of execution is wrong...

### DIFF
--- a/src/browser/Page.zig
+++ b/src/browser/Page.zig
@@ -698,17 +698,12 @@ pub fn documentIsComplete(self: *Page) void {
 fn _documentIsComplete(self: *Page) !void {
     self.document._ready_state = .complete;
 
-    // Should this be different for each?
     var ls: JS.Local.Scope = undefined;
     self.js.localScope(&ls);
     defer ls.deinit();
 
     // Dispatch `_to_load` events before window.load.
     for (self._to_load.items) |element| {
-        if (comptime IS_DEBUG) {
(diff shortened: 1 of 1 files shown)


src/browser/Page.zig
```diff
@@ -698,25 +698,30 @@ pub fn documentIsComplete(self: *Page) void {
 fn _documentIsComplete(self: *Page) !void {
     self.document._ready_state = .complete;
 
+    // Should this be different for each?
```
karlseguin: Will work either way. I don't know what's a good general rule.

src/browser/Page.zig
```diff
@@ -698,25 +698,30 @@ pub fn documentIsComplete(self: *Page) void {
...
-            event,
-            ls.toLocal(maybe_inline_listener),
-            .{ .context = "Page dispatch load events" },
-        );
-
         if (comptime IS_DEBUG) {
             log.debug(.page, "load event for element", .{ .element = element });
         }
+
+        const event = try Event.initTrusted("load", .{}, self);
```
karlseguin, line 707: If this is something we find ourself needing more, we should build the functionality into event_manager. Either changing the existing dispatchWithFunction or adding a new one.

karlseguin (approved):

src/browser/Page.zig
nikneym: Agree.

PR #1497, opened 2026-02-08
Reduce some debug logs

karlseguin:
1 - Remove the double logging of "load" dispatch

2 - On even registration, just log the event target type, not the full node 

3 - Most significantly, rather than logging unknown properties (both global and
    per object) as they happen, collect them and log the on context ends. This
    log includes a count of how often it happened. On some pages, this reduces
    the number of unknown property messages by thousands.


krichprollsch (approved):
🔥

PR #1498, opened 2026-02-09
Convert more comptime-getters to fast getters

karlseguin:
Follow up to https://github.com/lightpanda-io/browser/pull/1495 which introduced the concept of a fast getter. This commit expands the new behavior to all comptime-known scalar getters.

It also leverages the new `v8__FunctionTemplate__New__Config` to 
1 - flag fast getters as having no side effect
2 - set the length (arity) on all functions

### DIFF
--- a/src/browser/js/bridge.zig
+++ b/src/browser/js/bridge.zig
@@ -63,12 +63,23 @@ pub fn Builder(comptime T: type) type {
         }
 
         pub fn property(value: anytype, opts: Property.Opts) Property {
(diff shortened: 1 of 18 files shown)


src/browser/js/bridge.zig
```diff
@@ -150,6 +161,7 @@ pub const Constructor = struct {
 
 pub const Function = struct {
     static: bool,
+    arity: usize,
```
krichprollsch, line 164: TIL

PR #1508, opened 2026-02-09
Support `selectionchange` Event

mookums:
This properly fires the `selectionchange` event on [Document, HTMLInputElement, HTMLTextAreaElement] whenever their internal selection is modified.

### DIFF
--- a/src/browser/tests/element/html/input.html
+++ b/src/browser/tests/element/html/input.html
@@ -188,7 +188,7 @@
     const input = document.createElement('input');
     input.value = 'Hello World';
     document.body.appendChild(input);
-    
+
     let eventCount = 0;
     let lastEvent = null;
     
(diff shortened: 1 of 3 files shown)


src/browser/webapi/element/html/Input.zig
```diff
@@ -83,6 +84,26 @@ _selection_start: u32 = 0,
...
+
+pub fn setOnSelectionChange(self: *Input, listener: ?js.Function) !void {
+    if (listener) |listen| {
+        self._on_selectionchange = try listen.persistWithThis(self);
+    } else {
+        self._on_selectionchange = null;
+    }
+}
+
+fn dispatchSelectionChangeEvent(self: *Input, page: *Page) anyerror!void {
```
karlseguin: is `anyerror` necessary here?

src/browser/webapi/element/html/TextArea.zig
```diff
@@ -35,6 +36,26 @@ _selection_start: u32 = 0,
...
+
+pub fn setOnSelectionChange(self: *TextArea, listener: ?js.Function) !void {
+    if (listener) |listen| {
+        self._on_selectionchange = try listen.persistWithThis(self);
+    } else {
+        self._on_selectionchange = null;
+    }
+}
+
+fn dispatchSelectionChangeEvent(self: *TextArea, page: *Page) anyerror!void {
```
karlseguin: is anyerror necessary here?

karlseguin (commented):
Besides the anyerror, this looks good.

The 3 test files, `input.html`, `textarea.html` and `selection.html` all have new cases which fail in Firefox. Is this expected? input.html and selection.html already had failing cases..but now there are more.

(I double checked, the previously failing input.html test has to do with a different exception message, so somewhat expected, but worth fixing (at some point)..so ya, can you double check the tests/behavior with respect to them failing in Firefox).

src/browser/webapi/element/html/Input.zig
mookums: Doesn't seem to compile without this one specifically. It's fine without it in the TextArea which is a little weird?

karlseguin: I fixed it in https://github.com/lightpanda-io/browser/pull/1508/changes/97d53b81a7a67400485065f7cd8ef1489cb3d861

It _is_ weird. Whether or not it can ever be recursive, I'm not sure..but I do know there isn't enough info for the compiler to rule it out.

PR #1510, opened 2026-02-10
Use cgroups for RAM mesurement

mrdimidium:
I was currently investigating why CI was crashing in the multithread due to a memory usage error in CI.

The current approach uses RSS (`Maximum Resident Set Size`). This metric measures the number of pages, not actual memory usage. These numbers are close, but don't necessarily match – memory pages can be swapped out and/or deduplicated with other processes. This also doesn't take into account short peaks, virtual memory, or page cache.

A more reliable way to measure per-process memory usage (or something else in future) is `cgroups v2`. It allows you to explicitly separate a process (and its child processes) and measure memory usage with sufficient accuracy.

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -215,8 +215,9 @@ jobs:
         run: |
           export AVG_DURATION=`cat puppeteer.out|grep 'avg run'|sed 's/avg run duration (ms) //'`
           export TOTAL_DURATION=`cat puppeteer.out|grep 'total duration'|sed 's/total duration (ms) //'`
-          export LPD_MEM_PEAK_KB=$(( $(cat $CG/memory.peak) / 1024 ))
-          echo "{\"duration_total\":${TOTAL_DURATION},\"duration_avg\":${AVG_DURATION},\"mem_peak\":${LPD_MEM_PEAK_KB}}" > bench.json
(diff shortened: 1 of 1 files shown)


mrdimidium:
@krichprollsch, PR cannot succeed because it requires permission to create cgroups on the self-hosted runer.

.github/workflows/e2e-test.yml
```diff
@@ -122,10 +122,18 @@ jobs:
     needs: zig-build-release
 
     env:
-      MAX_MEMORY: 26000
+      MAX_MEMORY: 26000 # 26MB (KB)
```
krichprollsch: The current result is `memory.peak_kb=5412`
So having a limit of `26000` isn't meaningful... What is the "correct" limit now? `6000`?

krichprollsch:
I think I would like to keep both measures.
I cgroup seems interesting, more accurate, but I want to keep the previous one to track the whole mem footprint, including resident mem, which is closer to what user will see. Then we could lax a bit the constraint on the test.

krichprollsch:
Relates with https://github.com/lightpanda-io/perf-fmt/pull/27

.github/workflows/e2e-test.yml
krichprollsch: Let's use 6MB to start.

PR #1513, opened 2026-02-10
Fix double-free of XHR on double client abort

karlseguin:
It's possible (in fact normal) for client.abort to be called twice on a schedule navigation. We immediately abort any pending requests once a secondary navigation is called (is that right?), and then again when the page shuts down.

The first abort will kill the transfer, so the XHR object has to null this value so that, on context shutdown, when the finalizer is called, we don't try to free it again.


nikneym:
> We immediately abort any pending requests once a secondary navigation is called (is that right?)

I think this is correct; can reproduce on other browsers.

PR #1522, opened 2026-02-11
Remove Page.reset

karlseguin:
Page.reset exists for 1 use case: multiple calls to the Page.navigate CDP method. At an extreme, something like this in puppeteer:

```
await page.goto(baseURL + '/campfire-commerce/');
await page.goto(baseURL + '/campfire-commerce/');
```

Rather than handling this generically in Page, we now handle this case specifically at the CDP layer. If the page isn't in its initial load state, i.e. page._load_state != .waiting, then we reload the page from the session.

For reloading, my initial inclination was to do session.removePage then session.createPage(). This behavior still seems potentially correct to me, but compared to our `reset`, this would trigger extra notifications, namely:

self.notification.dispatch(.page_remove, .{});

and

self.notification.dispatch(.page_created, page);

Bacause of https://github.com/lightpanda-io/browser/pull/1265/ I guess that could have side effects. So, to keep the behavior as close to the current "reset", a new `session.replacePage()` has been added which behaves a lot like removePage + createPage, but without the notifications being sent.

While I generally think this is just cleaner, this was largely driven by some planning for frame support. The entity for a Frame will share a lot with the Page (we'll extract that logic), so simplifying the Page, especially around initialization, helps simplify frame support.

### DIFF
--- a/src/browser/webapi/Screen.zig
+++ b/src/browser/webapi/Screen.zig
@@ -32,13 +32,6 @@ const Screen = @This();
 _proto: *EventTarget,
 _orientation: ?*Orientation = null,
 
-pub fn init(page: *Page) !*Screen {
-    return page._factory.eventTarget(Screen{
-        ._proto = undefined,
-        ._orientation = null,
-    });
-}
-
 pub fn asEventTarget(self: *Screen) *EventTarget {
     return self._proto;
 }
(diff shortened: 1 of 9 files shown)


src/browser/webapi/Screen.zig
```diff
@@ -32,13 +32,6 @@ const Screen = @This();
 _proto: *EventTarget,
 _orientation: ?*Orientation = null,
 
-pub fn init(page: *Page) !*Screen {
```
krichprollsch, line 35: Does this change relates w/ reset remove?

karlseguin: Only tangentially. The Screen instance is (was) created in page.reset. Every frame will have a distinct window, but those windows will reference the same screen. This inlines the creation, making the page the authority, and [eventually] ensuring no one else (e.g. a frame) can [easily] create a Screen.

Issue #1527, opened 2026-02-11
Segfault from RecursiveApiCall when fetching a16z.com


hammeiam:
Lightpanda segfaults when fetching https://a16z.com. The crash is reproducible on every attempt.

The logs show a RecursiveApiCall error in the HTTP client immediately before the segfault:
```
$time=1770849265444 $scope=js $level=warn $msg="eval script" url=https://cdn.parsely.com/keys/a16z.com/p.js
caught.exception="Cannot read properties of undefined (reading 'navigationStart')" caught.stack="TypeError: Cannot read
properties of undefined (reading 'navigationStart')\n    at e.now (https://cdn.parsely.com/keys/a16z.com/p.js:7:27212)\n
at a.slots.discoverSlots (https://cdn.parsely.com/keys/a16z.com/p.js:7:38925)\n    at
https://cdn.parsely.com/keys/a16z.com/p.js:7:38940\n    at https://cdn.parsely.com/keys/a16z.com/p.js:7:39354" caught.line=7
caught.caught=true cacheable=true
$time=1770849265560 $scope=http $level=fatal $msg="Failed to remove handle" err=RecursiveApiCall
[1]    88779 segmentation fault  ./lightpanda serve --host 127.0.0.1 --port 9222
```

The JS exception itself is caught (caught.caught=true) — it's a third-party analytics script (Parsely) trying to access
performance.timing.navigationStart. The segfault appears to come from the HTTP client's curl handle cleanup after the
RecursiveApiCall error.

Steps to reproduce:

`./lightpanda serve --host 127.0.0.1 --port 9222`

Then in another terminal (using Playwright):

```js
from playwright.sync_api import sync_playwright

with sync_playwright() as p:
    browser = p.chromium.connect_over_cdp("ws://127.0.0.1:9222")
    context = browser.new_context(ignore_https_errors=True)
    page = context.new_page()
    page.goto("https://a16z.com", wait_until="commit", timeout=30000)
    print(page.content())
```

Environment:
- Nightly build 28815a0a (downloaded 2026-02-10)
- macOS 15.7, arm64 (Apple Silicon)

karlseguin:
Hi. Sorry about that. Was able to reproduce and is fixed in the nightly.

PR #1528, opened 2026-02-12
Remove potential recursive abort call in curl

karlseguin:
Curl doesn't like recursive calls. For example, you can't call curl_multi_remove_handle from within a dataCallback.

This specifically means that, as-is, transfer.abort() calls aren't safe to be called during a libcurl callback. Consider this code:

```
req.open('GET', 'http://127.0.0.1:9582/xhr');
req.onreadystatechange = (e) => {
  req.abort();
}
req.send();
```

onreadystatechange is triggered by network events, i.e. it executes in libcurl callback. Thus, the above code fails to truly "abort" the request with `curl_multi_remove_handle` error, saying it's a recursive call.

To solve this, transfer.abort() now sets an `aborted = true` flag. Callbacks can now use this flag to signal to libcurl to stop the transfer.

A test was added which reproduced this issue, but this comes from: https://github.com/lightpanda-io/browser/issues/1527  which I wasn't able to reliably reproduce. I did see it happen regularly, just not always. It seems like this commit fixes that issue.

### DIFF
--- a/src/http/Client.zig
+++ b/src/http/Client.zig
@@ -111,6 +111,10 @@ config: *const Config,
 
 cdp_client: ?CDPClient = null,
 
+// keep track of when curl_multi_perform is happening so that we can avoid
+// recursive calls into curl (which it will fail)
+performing: bool = false,
+
 // libcurl can monitor arbitrary sockets, this lets us use libcurl to poll
 // both HTTP data as well as messages from an CDP connection.
 // Furthermore, we have some tension between blocking scripts and request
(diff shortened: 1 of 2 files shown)


src/http/Client.zig
```diff
@@ -1224,10 +1226,18 @@ pub const Transfer = struct {
...
         }
-        self.deinit();
+
+        // abort can be called from a libcurl callback, e.g. we get data, we
+        // do the header done callback, the client aborts.
+        // libcurl doesn't support re-entrant calls. We can't remove the
+        // handle from the multi during a callback. Instead, we flag this as
+        // aborted, which will signal the callbacks to stop processing the
+        // transfer
+        self.aborted = true;
```
krichprollsch: Before this change, we always deinit the transfer on abort.
Now, if the transfer is processing, it will be deinit in a further callback?

karlseguin: Ya, that's not right. 

The code now checks if `client.performing` is true. When true, it flags the abort (which the callbacks will handle). When false, it frees it like it used. to.

performing is set when multi_perform is called:
```zig
{
        self.performing = true;
        defer self.performing = false;
        try errorMCheck(c.curl_multi_perform(multi, &running));
}
```

krichprollsch (approved):

PR #1538, opened 2026-02-13
Remove custom-arenas, use ArenaPool instead

karlseguin:
This removes the browser-specific arenas (session, transfer, page, call) in favor of the arena pool.

This is a bit of a win-lose commit. It exists as (the last?) step before I can really start working on frames. Frames will require their own "page" and "call" arenas, so there isn't just 1 per browser now, but rather N, where N is the number of frames + 1 page. This change was already done for Contexts when ExecutionWorld was removed, and the idea is the same: making these units more self contained so to support cases where we break out of the "1" model we currently have (1 browser, 1 session, 1 page, 1 context, ...).

But it's a bit of a step backwards because the ArenaPool is dumb and just resets everything to a single hard-coded (for now) value: 16KB. But in my mind, an arena that's used for 1 thing (e.g. the page or call arenas) is more likely to be well-sized for that specific role in the future, even on a different page/navigate.

I think ultimately, we'll move to an ArenaPool that has different levels, e.g. acquire() and acquireLarge() which can reset to different sizes, so that a page arena can use acquireLarge() and retain a larger amount of memory between use.

### DIFF
--- a/src/ArenaPool.zig
+++ b/src/ArenaPool.zig
@@ -85,3 +85,8 @@ pub fn release(self: *ArenaPool, allocator: Allocator) void {
     self.free_list_len = free_list_len + 1;
     self.free_list = entry;
 }
+
+pub fn reset(_: *const ArenaPool, allocator: Allocator, retain: usize) void {
+    const arena: *std.heap.ArenaAllocator = @ptrCast(@alignCast(allocator.ptr));
+    _ = arena.reset(.{ .retain_with_limit = retain });
+}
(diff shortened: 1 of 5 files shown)


krichprollsch:
What about acquiring arena from the pool by category, let say: xhr, session, page, event... then attribute an prev arena using this category if exists?

karlseguin:
> What about acquiring arena from the pool by category, let say: xhr, session, page, event... then attribute an prev arena using this category if exists?

Yes, something like that would work.

PR #1542, opened 2026-02-13
Add option to override dns servers

krichprollsch:
Add `--dns_servers` option to override default's DNS servers. The option expects a comma separated IP list.


krichprollsch:
I close this PR b/c curl requires [c-ares](https://github.com/c-ares/c-ares) dependency to support `CURLOPT_DNS_SERVERS` option.

The drawback of this change: we will not support `/etc/hosts` resolution by default.

PR #1551, opened 2026-02-14
`simpleZigValueToJs`: support `Uint8ClampedArray`

nikneym:
Needed this to implement `ImageData#data` getter. This works differently than other typed arrays since returned object can be mutated from both Zig and JS ends.

### DIFF
--- a/src/browser/js/js.zig
+++ b/src/browser/js/js.zig
@@ -223,6 +223,22 @@ pub fn simpleZigValueToJs(isolate: Isolate, value: anytype, comptime fail: bool,
                     }
                     return @ptrCast(v8.v8__Uint8ClampedArray__New(array_buffer, 0, len));
                 },
+                Uint8ClampedArray(.copy) => {
+                    const values = value.values;
+                    const len = values.len;
+                    var array_buffer: *const v8.ArrayBuffer = undefined;
+                    if (len == 0) {
+                        array_buffer = v8.v8__ArrayBuffer__New(isolate.handle, 0).?;
+                    } else {
+                        const backing_store = v8.v8__ArrayBuffer__NewBackingStore(isolate.handle, len);
+                        const data: [*]u8 = @ptrCast(@alignCast(v8.v8__BackingStore__Data(backing_store)));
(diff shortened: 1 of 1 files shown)


src/browser/js/js.zig
```diff
@@ -194,6 +202,24 @@ pub fn simpleZigValueToJs(isolate: Isolate, value: anytype, comptime fail: bool,
                     // but this can never be valid.
                     @compileError("Invalid TypeArray type: " ++ @typeName(value_type));
                 },
+                Uint8ClampedArray(.mutable) => {
```
karlseguin: Shouldn't there be some handling for `Uint8ClampedArray(.immutable)` somewhere in the mapping code?

src/browser/js/js.zig
```diff
@@ -77,6 +77,14 @@ pub const ArrayBuffer = struct {
     }
 };
 
+/// `mutable` indicates bytes are not copied by `simpleZigValueToJs`;
+/// instead, `BackingStore` takes ownership of already allocated `values`.
```
karlseguin: > `BackingStore` takes ownership of already allocated `values`.

Which is a bit confusing when you see the provided deleter does is no-op. In combination, it looks like we're leaking `values`. So I understand that this is the definition of a backingstore, but with an empty/no-op deleter, it's kind of the exact opposite. It doesn't take ownership, it merely references it.  And, I assume this is all "ok", because we assume value will be in an arena.

Think the comment is worth fixing since it's kind of saying the opposite of the intent.

src/browser/js/js.zig
nikneym: I intentionally omitted that since not needed for now; let me also add it.

src/browser/js/js.zig
nikneym: > It doesn't take ownership, it merely references it. And, I assume this is all "ok", because we assume value will be in an arena.

Indeed, I'll update the comment.
BTW we can introduce reference counting for `deleter`; that way, instead of relying on arena, we can free it once not needed on JS side.

PR #1553, opened 2026-02-15
Basic support for `ImageData`

nikneym:
This PR adds support for `ImageData`.

https://developer.mozilla.org/en-US/docs/Web/API/ImageData

### DIFF
--- a/src/browser/js/js.zig
+++ b/src/browser/js/js.zig
@@ -78,16 +78,16 @@ pub const ArrayBuffer = struct {
 };
 
 pub const ArrayType = enum(u8) {
-    Int8,
-    Uint8,
-    Uint8Clamped,
-    Int16,
-    Uint16,
-    Int32,
-    Uint32,
-    Float16,
-    Float32,
-    Float64,
+    int8,
+    uint8,
+    uint8_clamped,
+    int16,
+    uint16,
+    int32,
+    uint32,
+    float16,
+    float32,
+    float64,
 };
 
 pub fn ArrayBufferRef(comptime kind: ArrayType) type {
@@ -95,15 +95,15 @@ pub fn ArrayBufferRef(comptime kind: ArrayType) type {
         const Self = @This();
 
         const BackingInt = switch (kind) {
-            .Int8 => i8,
-            .Uint8, .Uint8Clamped => u8,
-            .Int16 => i16,
-            .Uint16 => u16,
-            .Int32 => i32,
-            .Uint32 => u32,
-            .Float16 => f16,
-            .Float32 => f32,
-            .Float64 => f64,
+            .int8 => i8,
+            .uint8, .uint8_clamped => u8,
+            .int16 => i16,
+            .uint16 => u16,
+            .int32 => i32,
+            .uint32 => u32,
+            .float16 => f16,
+            .float32 => f32,
+            .float64 => f64,
         };
 
         handle: *const v8.Value,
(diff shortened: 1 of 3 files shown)


nikneym:
> The code isn't included in anything. It's unused. If you try to run this test explicitly, it isn't found.

Forgot to include change for `bridge.zig` since I have couple of modified files here and there, should be fine now.

> Also, the image_data.html fails in firefox. In some cases, that is expected. But is it in this case?

It is, the reason why is FF doesn't support [`settings`](https://developer.mozilla.org/en-US/docs/Web/API/ImageData/ImageData#settings) argument. Should be passing on Chrome.

src/browser/js/Context.zig
```diff
@@ -1045,6 +1045,12 @@ pub const FinalizerCallback = struct {
     }
 };
 
+/// Creates a new typed array. Memory is owned by JS context.
+/// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Typed_arrays
+pub fn createTypedArray(self: *Context, comptime array_type: js.ArrayType, size: usize) js.ArrayBufferRef(array_type) {
```
karlseguin, line 1050: Just by zig convention, type names are always PascalCase.

src/browser/js/js.zig
```diff
@@ -77,14 +77,69 @@ pub const ArrayBuffer = struct {
...
-/// `ref` indicates bytes are not copied by `simpleZigValueToJs`;
-/// instead, `values` references an already allocated memory. Note that
-/// this variant assumes memory is (de)allocated by an arena allocator.
-///
-/// `copy` behaves the same as `TypedArray(T)`.
-pub fn Uint8ClampedArray(comptime state: enum(u1) { ref, copy }) type {
-    return struct {
-        values: if (state == .ref) []u8 else []const u8,
+pub const ArrayType = enum(u8) {
+    Int8,
```
karlseguin: and enum feilds are just like struct fields, `snake_case`. Some flexibility here though, e.g. if the goal is to do some meta programming with them like with `enumFromString`,

src/browser/js/js.zig
```diff
@@ -77,14 +77,69 @@ pub const ArrayBuffer = struct {
...
+    Uint16,
+    Int32,
+    Uint32,
+    Float16,
+    Float32,
+    Float64,
+};
+
+pub fn ArrayBufferRef(comptime kind: ArrayType) type {
+    return extern struct {
```
karlseguin: why `extern`?

src/browser/js/js.zig
```diff
@@ -77,14 +77,69 @@ pub const ArrayBuffer = struct {
...
+            .Int16 => i16,
+            .Uint16 => u16,
+            .Int32 => i32,
+            .Uint32 => u32,
+            .Float16 => f16,
+            .Float32 => f32,
+            .Float64 => f64,
+        };
+
+        internal: *const v8.Value,
```
karlseguin: we call this `handle` pretty consistently in other wrappers.

src/browser/js/Context.zig
nikneym: I didn't consider that as a type name though, that'd be `ArrayBufferRef`. We can, however, name this `TypedArray` in `Context.zig` if you like.

karlseguin: err..right, it isn't a type, it's an enum value that represents a type. Does `ArrayType` have to exists?

What's the difference between `js.createTypedArray(.UInt16, 0)` and `js.createTypedArray(u16, 0)` ?

nikneym: Actually nothing, enum solely exists for clamped array since u8 would be used by `Uint8Array`.

karlseguin: okok, you win.

PR #1560, opened 2026-02-16
Add support for dumping output to markdown

arrufat:
This PR extends the `--dump` flag to accept an argument
- `html`: same as the old `--dump` flag
- `markdown`: dumps a markdown version of the site (converted from the internal HTML representation)

The main goal is to [reduce token usage up to 80%](https://blog.cloudflare.com/markdown-for-agents/).

### DIFF
--- a/src/browser/markdown.zig
+++ b/src/browser/markdown.zig
@@ -152,7 +152,7 @@ fn renderElement(el: *Element, state: *State, writer: *std.Io.Writer, page: *Pag
         },
         .li => {
             const indent = if (state.list_depth > 0) state.list_depth - 1 else 0;
-            try writeIndentation(indent, writer);
+            for (0..indent) |_| try writer.writeAll("  ");
 
             if (state.list_depth > 0) {
                 const current_list = &state.list_stack[state.list_depth - 1];
@@ -396,13 +396,6 @@ fn escapeMarkdown(writer: *std.Io.Writer, text: []const u8) !void {
             },
             else => try writer.writeByte(c),
         }
-    }
-}
-
-fn writeIndentation(level: usize, writer: *std.Io.Writer) !void {
-    var i: usize = 0;
-    while (i < level) : (i += 1) {
-        try writer.writeAll("  ");
     }
 }
 


src/Config.zig
```diff
@@ -546,15 +552,23 @@ fn parseFetchArgs(
...
+    var dump_mode: ?DumpFormat = null;
     var withbase: bool = false;
     var url: ?[:0]const u8 = null;
     var common: Common = .{};
     var strip: dump.Opts.Strip = .{};
 
     while (args.next()) |opt| {
         if (std.mem.eql(u8, "--dump", opt)) {
-            fetch_dump = true;
+            const str = args.next() orelse {
```
karlseguin: For backwards compatibility, would be nice if `html` was default if no option is given.

src/browser/markdown.zig
```diff
@@ -0,0 +1,518 @@
...
+pub fn dump(node: *Node, opts: Opts, writer: *std.Io.Writer, page: *Page) !void {
+    _ = opts;
+    var state = State{};
+    try render(node, &state, writer, page);
+    if (!state.last_char_was_newline) {
+        try writer.writeByte('\n');
+    }
+}
+
+fn render(node: *Node, state: *State, writer: *std.Io.Writer, page: *Page) anyerror!void {
```
karlseguin: If you change this 1 `anyerror` to `error{WriteFailed}`, then all the other cases can use an inferred error set.

src/browser/markdown.zig
```diff
@@ -0,0 +1,518 @@
...
+            try writer.writeAll(")");
+            state.last_char_was_newline = false;
+            return; // Treat as void
+        },
+        .anchor => {
+            try writer.writeByte('[');
+            state.last_char_was_newline = false;
+        },
+        .input => {
+            if (el.getAttributeSafe(comptime .wrap("type"))) |t| {
```
karlseguin: This can stay as-is, a page won't have thousands of input boxes, so it doesn't really matter. But:

```zig
const input = el.as(Element.Html.Input);
```

And now you have access to `input._input_type` and `input._checked`, which avoids the attribute lookup (our attributes are a linked list).

src/browser/markdown.zig
```diff
@@ -0,0 +1,518 @@
...
+                    state.last_char_was_newline = false;
+                }
+            }
+            return;
+        },
+        else => {},
+    }
+
+    // --- Render Children ---
+    try renderChildren(el.asNode(), state, writer, page);
```
karlseguin, line 268: did you consider inlining some of this?

like, take `.anchor`. In prefixes, you write "['. Now you come here and render children. Then in suffixes you write `)($href)`. 

Instead, you could one-shot it:

```zig
.anchor => {
    try writer.writeByte('[');
    try renderChildren(el.asNode(), state, writer, page);
    try writer.writeAll("](");
    if (el.getAttributeSafe(comptime .wrap("href"))) |href| {
      try writer.writeAll(href);
    }
    try writer.writeByte(')');
},
```

You have to duplicate the call to `try renderChildren(el.asNode(), state, writer, page);` in every case, but you potentially need to preserve less state, and subjectively, it's easier to just understand how the anchor is being rendered from that one cohesive block.

I don't have a strong opinion about it. I think your solution has a certain elegance to it too. Just wanted to throw it out there.

Maybe it doesn't work well in call cases, and then having a mix of the two approaches would be worse.

src/browser/markdown.zig
```diff
@@ -0,0 +1,518 @@
...
+        .i, .em => {
+            try writer.writeAll("*");
+            state.last_char_was_newline = false;
+        },
+        .s, .del => {
+            try writer.writeAll("~~");
+            state.last_char_was_newline = false;
+        },
+        .blockquote => {
+            state.in_blockquote = false;
```
karlseguin: Don't think `in_blockquote` was ever used for anything.

src/browser/markdown.zig
```diff
@@ -0,0 +1,518 @@
...
+        if (text.len > 0 and std.ascii.isWhitespace(text[text.len - 1])) {
+            try writer.writeByte(' ');
+        }
+    }
+}
+
+fn escapeMarkdown(writer: *std.Io.Writer, text: []const u8) !void {
+    for (text) |c| {
+        switch (c) {
+            '\\', '`', '*', '_', '{', '}', '[', ']', '(', ')', '#', '+', '-', '!', '|', '<', '>' => {
```
karlseguin: Claude doesn't like the inclusion of '<' and '>'. It thinks its (a) unnecessary (snice escapeMarkdown is never called at the start of a line) and (b) could cause issues with renders.

src/Config.zig
arrufat: Right, I thought about it, but didn't want to add the extra complexity of handling an optional argument. Fixed.

src/browser/markdown.zig
arrufat: Good catch!

src/browser/markdown.zig
arrufat: Fixed

src/browser/markdown.zig
arrufat: Fixed

src/browser/markdown.zig
arrufat: Yeah, I like this one, too.

karlseguin: I used anchor as an example, I do think it could be applied to more/most?

arrufat: Right, it can also be applied to `.input`

arrufat:
@karlseguin thank you for the thorough code review :)

As a side note, on my personal projects I use the latest Zig dev version, which has all the new Io changes, the juicy main, etc, so it was weird to write old Zig (from 3 months ago).

It's going to be fun to port lightpanda to Zig 0.16.0...

krichprollsch:
Hey @arrufat thank you for the PR! This is a great new feature for the browser :tada: 

I'm wondering how we could improve a bit more the output in some specific cases, for example, when testing on lightpanda.io, the "Developers are talking about Lightpanda" is not really good, mainly due to the html itself:
```markdown
## Developers are talking about Lightpanda
[![@asher\_on\_it\_all twitter avatar](https://cdn.lightpanda.io/website/assets/images/avatars/wzjKMbU7.jpg)

##### Asher \| 16

@asher\_on\_it\_all

@lightpanda\_io is going to change the world. I'm calling it\!\!\! I don't understand why i don't hear anything about their browser. \- CDP support \- most features you need are already supported \- most websites already work \- insanely low memory use \- insanely fast WHAT?
](https://x.com/i/status/1992203799885226259)
```


```html 
                <h2 class="text-center mb-8 text-3xl lg:text-5xl lg:leading-tight font-bold">Developers are talking about Lightpanda</h2>
                <div class="block relative p-6 -m-6 mb-12 overflow-x-visible transition-all max-h-[700px] overflow-y-hidden">
                    <div class="relative flex flex-row flex-nowrap gap-4 w-full transition-all">
                        <div class="flex flex-col gap-4 h-fit w-full animate-[marquee_40000ms_linear_both_infinite] will-change-transform transition-transform"><a target="_blank" href="https://x.com/i/status/1992203799885226259">
                                <div class="bg-surface-75 border group-hover/testimonial-card:border-background transition-colors rounded-lg p-6 drop-shadow-sm backdrop-blur-[2px] hover:backdrop-blur-lg shadow-[0px_0px_8px_rgba(108,199,226,1)] hover:shado
w-[0px_0px_16px_rgba(108,199,226,1)] transition-shadow">
                                    <div class="relative">
                                        <div class="flex items-start justify-between gap-2 w-full">
                                            <div class="flex flex-row items-center gap-3 w-fit">
                                                <div class="h-10 w-10 overflow-hidden rounded-full border-2 border-primary border-control"><img src="https://cdn.lightpanda.io/website/assets/images/avatars/wzjKMbU7.jpg" class="w-10 h-10" alt="@asher_on_it_
all twitter avatar"></div>
                                                <div class="flex flex-col">
                                                    <h5 class="text-foreground text-sm font-semibold">Asher | 16</h5>
                                                    <p class="text-gray-300 text-xs font-normal font-mono">@asher_on_it_all</p>
                                                </div>
                                            </div>
                                            <div class="flex h-6 w-6 items-center justify-center rounded-full bg-black"><svg class="h-[12px] w-[12px]" fill="white" viewBox="0 0 24 24" aria-hidden="true">
                                                    <path d="M18.244 2.25h3.308l-7.227 8.26 8.502 11.24H16.17l-5.214-6.817L4.99 21.75H1.68l7.73-8.835L1.254 2.25H8.08l4.713 6.231zm-1.161 17.52h1.833L7.084 4.126H5.117z"></path>
                                                </svg></div>
                                        </div>
                                    </div>
                                    <p class="text-foreground-lighter mt-3 text-base font-normal whitespace-pre-line">@lightpanda_io is going to change the world. I'm calling it!!!

                                        I don't understand why i don't hear anything about their browser.

                                        - CDP support
                                        - most features you need are already supported
                                        - most websites already work
                                        - insanely low memory use
                                        - insanely fast

                                        WHAT? </p>
                                </div>
                            </a>
```

image


The markdown rendering is not very nice:

image




But it's something we can discuss and improve in following PRs.

karlseguin:
FWIW, I had the same thought about the output based on the output for https://www.openmymind.net/  but I couldn't think of how you'd fix it. It's, as you say, the HTML itself.  All the text just appears like a gigantic block because the HTML has no spacing, just tags.  So while in html It might look like: 

```
<link>          date
```

In Markdown they're glued:

```
<link>date
```

And I assume adding spaces around markup would fix this case, but break others.

PR #1562, opened 2026-02-16
dispatch .page_navigated event on page error callback and create HTML page

krichprollsch:
When a CDP client navigates to a page and the page generates an error, it blocks waiting for the .page_navigated event.

It currently happens w/ robots.txt denied page.
Example: https://httpbin.io/deny

EDIT: others clients like chromedp expect to have an HTML page on error too. So We create an error HTML page in case of pageErrorCallback.

### DIFF
--- a/src/browser/Page.zig
+++ b/src/browser/Page.zig
@@ -708,7 +708,10 @@ fn pageDataCallback(transfer: *Http.Transfer, data: []const u8) !void {
                 try arr.appendSlice(self.arena, "<html><head><meta charset=\"utf-8\"></head><body><pre>");
                 self._parse_state = .{ .text = arr };
             },
-            else => self._parse_state = .{ .raw = .{} },
+            .image_jpeg, .image_gif, .image_png, .image_webp => {
+                self._parse_state = .{ .image = .empty };
+            },
+            else => self._parse_state = .{ .raw = .empty },
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/browser/Page.zig
```diff
@@ -799,6 +799,23 @@ fn pageErrorCallback(ctx: *anyopaque, err: anyerror) void {
     var self: *Page = @ptrCast(@alignCast(ctx));
     self.clearTransferArena();
     self._parse_state = .{ .err = err };
+    defer self.documentIsComplete();
+
+    // Generate a pseudo HTML page indicating the navigation falilure.
```
karlseguin:
```suggestion
    // Generate a pseudo HTML page indicating the navigation failure.
```

src/browser/Page.zig
```diff
@@ -799,6 +799,23 @@ fn pageErrorCallback(ctx: *anyopaque, err: anyerror) void {
     var self: *Page = @ptrCast(@alignCast(ctx));
     self.clearTransferArena();
     self._parse_state = .{ .err = err };
+    defer self.documentIsComplete();
```
karlseguin: Is this necessary? Shouldn't this ultimately result in `pageDoneCallback` being called, which will call this?

krichprollsch: It is to dispatch events correctly. But I agree with you, moving that part into `pageDoneCallback` makes more sense.

krichprollsch:
I updated the code to:
* have an HTML for all `pageDoneCallback` cases
* support images

PR #1565, opened 2026-02-17
Initial support for frames

karlseguin:
Adds basic frame support. Initially explored adding a BrowsingContext and embedding it in Page, with the goal of also having it embedded in a to-be created Frame. But it turns out that 98% of Page _was_ BrowsingContext and introducing a BrowsingContext as the primary interaction unit broke pretty much _every_ single WebAPI. So Page was expanded:

- Added `_parent: ?*Page`, which is `null` for "root" page.
- Added `frame: ?*IFrame`, which is `null` for the "root" page. This is the HTMLIFrameElement for frame-pages.
- Added a _type: enum{root, frame}, which is currently only used to improve the logs
- Added a frames: std.ArrayList(*Page). This is a list of frames for the page. Note that a "frame-page" can itself haven nested frames.

Besides the above, there were 3 "big" changes.

1 - Adding frames (dynamically, parsed) has to create a new page, start
    navigation, track it (in the frames list). Part of this was just
    piggybacking off of code that handles <script>

2 - The page "load" event blocks on the frame "load" event. This cascades.
    when a page triggers it's load, it can do:
```zig
      if (self._parent) |p| {
        p.iframeLoaded(self);
      }
```
   Pages need to keep track of how many iframes they're waiting to load. When
   all iframes (and all scripts) are loaded, it can then triggers its own load
   event.

3 - Our JS execution expects 1 primary entered context (the pages). But we now
    have multiple page contexts, and we need to be in the correct one based
    on where javascript is being executed. There is no more an default entered
    context. Creating a Local.Scope enters the context, and ls.deinit() exits
    the context.

### DIFF
--- a/src/browser/js/Env.zig
+++ b/src/browser/js/Env.zig
@@ -175,7 +175,7 @@ pub fn init(app: *App, opts: InitOpts) !Env {
             .data = null,
             .flags = v8.kOnlyInterceptStrings | v8.kNonMasking,
         });
-        // I don' 100% understand this. We actually set this up in the snapshot,
+        // I don't 100% understand this. We actually set this up in the snapshot,
         // but for the global instance, it doesn't work. SetIndexedHandler and
         // SetNamedHandler are set on the Instance template, and that's the key
         // difference. The context has its own global instance, so we need to set


karlseguin:
Based on https://github.com/lightpanda-io/browser/pull/1582, so we should merge that first.

Navigation of the main page works. Navigation within frames doesn't (but it doesn't break the main page navigation, like with the initial commits). Also, I don't think we handle src-less frames properly, but I'm not sure that will actually be an issue, and if it is, I rather deal with it in a separate commit.

src/browser/js/Env.zig
```diff
@@ -175,8 +175,23 @@ pub fn init(app: *App, opts: InitOpts) !Env {
             .data = null,
             .flags = v8.kOnlyInterceptStrings | v8.kNonMasking,
         });
+        // I don' 100% understand this. We actually set this up in the snapshot,
```
krichprollsch:
```suggestion
        // I don't 100% understand this. We actually set this up in the snapshot,
```

PR #1570, opened 2026-02-17
Dispatch play, pause, and emptied events from HTMLMediaElement

egrs:
## Summary
- `play()` now dispatches `play` and `playing` events
- `pause()` now dispatches `pause` event
- `load()` now dispatches `emptied` event
- Follows the existing event dispatch pattern (Event.init + page._event_manager.dispatch)
- Added test for event listener callbacks on play/pause/load

## Test plan
- [x] All 238 unit tests pass (`make test`)
- [x] New `play_pause_events` test verifies event listener callbacks
- [x] CDP integration test confirms events fire on patched build
- [x] Official container confirmed to NOT fire events (baseline)

### DIFF
--- a/src/browser/tests/element/html/media.html
+++ b/src/browser/tests/element/html/media.html
@@ -76,21 +76,21 @@
     audio.pause();
     testing.expectEqual('play,playing,pause', events.join(','));
 
-    // Third play: resume from pause, fires play only (not playing)
+    // Third play: resume from pause, fires play + playing (verified in Chrome)
     audio.play();
-    testing.expectEqual('play,playing,pause,play', events.join(','));
+    testing.expectEqual('play,playing,pause,play,playing', events.join(','));
 
     // Pause again
     audio.pause();
-    testing.expectEqual('play,playing,pause,play,pause', events.join(','));
+    testing.expectEqual('play,playing,pause,play,playing,pause', events.join(','));
 
     // Load: resets state, fires emptied
     audio.load();
-    testing.expectEqual('play,playing,pause,play,pause,emptied', events.join(','));
+    testing.expectEqual('play,playing,pause,play,playing,pause,emptied', events.join(','));
 
-    // Play after load: fires play + playing again (fresh start)
+    // Play after load: fires play + playing
     audio.play();
-    testing.expectEqual('play,playing,pause,play,pause,emptied,play,playing', events.join(','));
+    testing.expectEqual('play,playing,pause,play,playing,pause,emptied,play,playing', events.join(','));
   }
 </script>
 
--- a/src/browser/webapi/element/html/Media.zig
+++ b/src/browser/webapi/element/html/Media.zig
@@ -61,7 +61,6 @@ _playback_rate: f64 = 1.0,
 _ready_state: ReadyState = .HAVE_NOTHING,
 _network_state: NetworkState = .NETWORK_EMPTY,
 _error: ?*MediaError = null,
-_playing: bool = false,
 
 pub fn asElement(self: *Media) *Element {
     return self._proto._proto;
@@ -145,10 +144,7 @@ pub fn play(self: *Media, page: *Page) !void {
     self._network_state = .NETWORK_IDLE;
     if (was_paused) {
         try self.dispatchEvent("play", page);
-        if (!self._playing) {
-            self._playing = true;
-            try self.dispatchEvent("playing", page);
-        }
+        try self.dispatchEvent("playing", page);
     }
 }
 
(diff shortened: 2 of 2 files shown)


src/browser/webapi/element/html/Media.zig
```diff
@@ -136,25 +137,32 @@ fn isMaybeSupported(mime_type: []const u8) bool {
     return false;
 }
 
-pub fn play(self: *Media) void {
+pub fn play(self: *Media, page: *Page) !void {
     self._paused = false;
     self._ready_state = .HAVE_ENOUGH_DATA;
     self._network_state = .NETWORK_IDLE;
-    // TODO: Could dispatch 'play' and 'playing' events
+    try self.dispatchEvent("play", page);
```
krichprollsch: MDN says:
> The play event is fired when the paused property is changed from true to false
https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/play_event

So you must trigger the event only if `self._paused` was `true` before the change..

src/browser/webapi/element/html/Media.zig
```diff
@@ -136,25 +137,32 @@ fn isMaybeSupported(mime_type: []const u8) bool {
...
     self._current_time = 0;
     self._ready_state = .HAVE_NOTHING;
     self._network_state = .NETWORK_LOADING;
     self._error = null;
-    // TODO: Could dispatch events
+    try self.dispatchEvent("emptied", page);
+}
+
+fn dispatchEvent(self: *Media, name: []const u8, page: *Page) !void {
+    const event = try Event.init(name, null, page);
```
krichprollsch: Since these events are not cancellables nor bubbles, I would explicitly set this options:
 
```suggestion
    const event = try Event.init(name, .{ .bubbles = false, .cancelable = false }, page);
```

src/browser/webapi/element/html/Media.zig
```diff
@@ -136,25 +137,32 @@ fn isMaybeSupported(mime_type: []const u8) bool {
...
-    // TODO: Could dispatch 'play' and 'playing' events
+    try self.dispatchEvent("play", page);
+    try self.dispatchEvent("playing", page);
 }
 
-pub fn pause(self: *Media) void {
+pub fn pause(self: *Media, page: *Page) !void {
     self._paused = true;
-    // TODO: Could dispatch 'pause' event
+    try self.dispatchEvent("pause", page);
```
krichprollsch: I think it's the same as `play`: dispatch the event only if `self._paused` was `false`.

src/browser/webapi/element/html/Media.zig
```diff
@@ -136,25 +137,32 @@ fn isMaybeSupported(mime_type: []const u8) bool {
...
 }
 
-pub fn play(self: *Media) void {
+pub fn play(self: *Media, page: *Page) !void {
     self._paused = false;
     self._ready_state = .HAVE_ENOUGH_DATA;
     self._network_state = .NETWORK_IDLE;
-    // TODO: Could dispatch 'play' and 'playing' events
+    try self.dispatchEvent("play", page);
+    try self.dispatchEvent("playing", page);
```
krichprollsch: A bit harder for this one:
> The playing event is fired after playback is first started, and whenever it is restarted.
https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/playing_event

I guess we need another state to know if the event must be dispatched.

src/browser/webapi/element/html/Media.zig
```diff
@@ -136,25 +137,37 @@ fn isMaybeSupported(mime_type: []const u8) bool {
...
 }
 
-pub fn pause(self: *Media) void {
-    self._paused = true;
-    // TODO: Could dispatch 'pause' event
+pub fn pause(self: *Media, page: *Page) !void {
+    if (!self._paused) {
+        self._paused = true;
+        try self.dispatchEvent("pause", page);
+    }
```
krichprollsch, line 155:
```suggestion
    if (self._paused) return
    self._paused = true;
    try self.dispatchEvent("pause", page);
```

src/browser/tests/element/html/media.html
```diff
@@ -50,6 +50,38 @@
...
+    testing.expectEqual('play,playing', events.join(','));
+
+    // Pause: playing -> paused, fires pause
+    audio.pause();
+    testing.expectEqual('play,playing,pause', events.join(','));
+
+    // Second pause: already paused, no event
+    audio.pause();
+    testing.expectEqual('play,playing,pause', events.join(','));
+
```
krichprollsch, line 78: If you add this test, it will fail.

```suggestion
    // third play: already paused, play only
    audio.play();
    testing.expectEqual('play,playing,pause,play', events.join(','));
    
```

We are always dispatching a playing event after play, but we should to only one.
Add a `_playing: bool = false` field into Media.zig and check it before dispaching the playing event:
```zig
    if (!self._playing) {
        self._playing = true;
        try self.dispatchEvent("playing", page);
    }
```

krichprollsch: So chrome dispatch a `playing` event after pause->play transition, let's do the same.

PR #1571, opened 2026-02-17
Persisted typed arrays

nikneym:
This introduces persisted typed arrays; there was a dangling pointer issue introduced in #1551.

### DIFF
--- a/src/browser/js/Context.zig
+++ b/src/browser/js/Context.zig
@@ -1045,13 +1045,6 @@ pub const FinalizerCallback = struct {
     }
 };
 
-/// Creates a new typed array. Memory is owned by JS context.
-/// If storing the type in a Zig type is desired, prefer `.global` state to persist.
-/// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Typed_arrays
-pub fn createTypedArray(self: *const Context, comptime array_type: js.ArrayType, size: usize) js.ArrayBufferRef(array_type) {
-    return .init(self.local.?, size);
-}
-
 // == Profiler ==
 pub fn startCpuProfiler(self: *Context) void {
     if (comptime !IS_DEBUG) {
--- a/src/browser/js/Local.zig
+++ b/src/browser/js/Local.zig
@@ -75,6 +75,12 @@ pub fn newArray(self: *const Local, len: u32) js.Array {
     };
 }
 
+/// Creates a new typed array. Memory is owned by JS context.
+/// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Typed_arrays
+pub fn createTypedArray(self: *const Local, comptime array_type: js.ArrayType, size: usize) js.ArrayBufferRef(array_type) {
+    return .init(self, size);
+}
+
 pub fn runMicrotasks(self: *const Local) void {
     self.isolate.performMicrotasksCheckpoint();
 }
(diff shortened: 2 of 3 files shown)


src/browser/js/Context.zig
```diff
@@ -1046,8 +1046,14 @@ pub const FinalizerCallback = struct {
 };
 
 /// Creates a new typed array. Memory is owned by JS context.
+/// If storing the type in a Zig type is desired, prefer `.global` state to persist.
 /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Typed_arrays
-pub fn createTypedArray(self: *Context, comptime array_type: js.ArrayType, size: usize) js.ArrayBufferRef(array_type) {
+pub fn createTypedArray(
+    self: *const Context,
+    comptime array_type: js.ArrayType,
+    comptime state: js.ArrayBufferState,
```
karlseguin: I think all the code is now consistent with how we do this:
1 - create the local
2 - call .persist()  on it.

I'd prefer to keep that pattern unless there's a reason we can't.

src/browser/js/js.zig
```diff
@@ -138,7 +147,23 @@ pub fn ArrayBufferRef(comptime kind: ArrayType) type {
                 .float64 => @ptrCast(v8.v8__Float64Array__New(array_buffer, 0, size).?),
             };
 
-            return .{ .handle = handle };
+            switch (comptime state) {
+                .local => return .{ .handle = local_handle },
+                .global => {
+                    // We need a global handle if state is `global`.
+                    var global_handle: v8.Global = undefined;
+                    v8.v8__Global__New(isolate.handle, local_handle, &global_handle);
```
karlseguin: If you create a global,it almost certainly needs to be added to the context's `global_values` list so that, on shutdown, we can free them. See any of the existing `persist()` functions.

src/browser/js/js.zig
```diff
@@ -90,7 +90,13 @@ pub const ArrayType = enum(u8) {
     float64,
 };
 
-pub fn ArrayBufferRef(comptime kind: ArrayType) type {
+pub const ArrayBufferState = enum(u1) { local, global };
+
+/// If `state` is `global`; a persisted, global typed array is created.
+pub fn ArrayBufferRef(
+    comptime kind: ArrayType,
+    comptime state: ArrayBufferState,
```
karlseguin: Again, would prefer to to keep it consistent with the other code...ArrayBufferRef is always local. ArrayBufferRef.Global is the global. call persist() on `ArrayBufferRef` to get a global, and call `local` on ArrayBufferRef.Global to get the local.

(Unless there's a reason that doesn't work, e.g. Private is _always_ a global, but that's because it's owned by the isolate, not the context)

src/browser/js/Context.zig
```diff
@@ -1046,9 +1046,10 @@ pub const FinalizerCallback = struct {
 };
 
 /// Creates a new typed array. Memory is owned by JS context.
+/// If storing the type in a Zig type is desired, prefer `.global` state to persist.
 /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Typed_arrays
-pub fn createTypedArray(self: *Context, comptime array_type: js.ArrayType, size: usize) js.ArrayBufferRef(array_type) {
-    return .init(self.isolate, size);
+pub fn createTypedArray(self: *const Context, comptime array_type: js.ArrayType, size: usize) js.ArrayBufferRef(array_type) {
```
karlseguin: If you need a local, then the function should almost certainly be defined on `Local`. Context.local is often null. Having the api be `page.js.createTypedArray(...)` hides the need for the local and will likely result in a null reference. If it's defined on `Local` then it's clear (and required) that the caller have a local.

PR #1577, opened 2026-02-17
Add `modify` to Selection

mookums:
This adds support for `modify` into the Selection WebAPI. This allows us to pass a couple more WPT tests (`selection/modify.tentative.html` and `selection/modify-extend-word-trailing-inline-block.tentative.html`).

`modify` currently only supports `word` and `character` as the others are either layout dependent (like paragraph and line) or not all implemented across the major browsers.

### DIFF
--- a/src/browser/webapi/Selection.zig
+++ b/src/browser/webapi/Selection.zig
@@ -354,6 +354,30 @@ fn nextTextNode(node: *Node) ?*Node {
     }
 }
 
+fn nextTextNodeAfter(node: *Node) ?*Node {
+    var current = node;
+    while (true) {
+        if (current.nextSibling()) |sib| {
+            current = sib;
+        } else {
+            while (true) {
+                const parent = current.parentNode() orelse return null;
+                if (parent.nextSibling()) |uncle| {
+                    current = uncle;
+                    break;
+                }
+                current = parent;
+            }
+        }
+
+        var descend = current;
+        while (true) {
+            if (isTextNode(descend)) return descend;
+            descend = descend.firstChild() orelse break;
+        }
+    }
+}
+
 fn prevTextNode(node: *Node) ?*Node {
     var current = node;
 
@@ -413,10 +437,13 @@ fn modifyByCharacter(self: *Selection, alter: ModifyAlter, forward: bool, range:
                     new_node = t;
                     new_offset = 0;
                 }
+            } else if (nextTextNodeAfter(focus_node)) |next| {
+                new_node = next;
+                new_offset = 1;
             }
         } else {
+            // backward element-node case
             var idx = focus_offset;
-
             while (idx > 0) {
                 idx -= 1;
                 const child = focus_node.getChildAt(idx) orelse break;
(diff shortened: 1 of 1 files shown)


src/browser/webapi/Selection.zig
```diff
@@ -312,18 +305,251 @@ pub fn modify(
...
+                if (isTextNode(bottom)) {
+                    new_node = bottom;
+                    new_offset = bottom.getLength();
+                    break;
+                }
+            }
+        }
+    }
+
+    switch (alter) {
```
karlseguin: I think this is all a duplicate of what `applyWordModify` does? Could maybe just call that?

src/browser/webapi/Selection.zig
```diff
@@ -312,18 +305,251 @@ pub fn modify(
...
-    });
+fn applyWordModify(self: *Selection, alter: ModifyAlter, new_node: *Node, new_offset: u32, page: *Page) !void {
+    switch (alter) {
+        .move => {
+            const new_range = try Range.init(page);
+            try new_range.setStart(new_node, new_offset);
+            try new_range.setEnd(new_node, new_offset);
+            self._range = new_range;
+            self._direction = .none;
+            try dispatchSelectionChangeEvent(page);
```
karlseguin, line 566: It's possible that the selection wasn't actually changed, right? Should it still dispatch the event? Is it worth checking for that case?

src/browser/webapi/Selection.zig
```diff
@@ -312,18 +305,251 @@ pub fn modify(
...
+                new_node = prev;
+                new_offset = prevWordStart(prev.getData(), @intCast(prev.getData().len));
+            }
+        }
+    } else {
+        // Search and apply rules on the next Text Node.
+        // This is either next (on forward) or previous (on backward).
+
+        if (forward) {
+            const child = focus_node.getChildAt(focus_offset) orelse {
```
karlseguin, line 525: Here's what claude says about this:

When getChildAt(focus_offset) returns null (meaning the cursor is logically past all children of the element), the code calls:

  if (nextTextNode(focus_node)) |next| { ... }

  But nextTextNode starts by diving into focus_node.firstChild(). This finds text nodes inside the element — i.e., positions before the current focus. For a forward movement, we should be searching for text nodes after focus_node in document order.

  Concrete example: if the focus is (p1, 1) where p1 has exactly one text child, getChildAt(1) returns null, nextTextNode(p1) dives into p1's first child and returns that same text node, which is behind the focus position. A "move forward" would land backwards.

  The same class of issue exists in modifyByCharacter's non-text forward case when getChildAt returns null — new_node/new_offset remain as the element, which is incorrect.

  This is not exercised by the current tests because all four new tests collapse to an explicit text node, never an element node.

src/browser/webapi/Selection.zig
mookums: Whenever we assign a new range like `self._range = new_range`, we should fire the event.

src/browser/webapi/Selection.zig
mookums: Should be fixed by the latest commit.

PR #1579, opened 2026-02-18
return CSSStyleSheet from HTMLStyleElement.sheet

egrs:
## Summary

- Connected `<style>` elements with `type="text/css"` (or default) now return a `CSSStyleSheet` from the `.sheet` property
- Disconnected elements or non-CSS types correctly return `null`
- Sheet is lazily created and cached on first access

Per [MDN spec](https://developer.mozilla.org/en-US/docs/Web/API/HTMLStyleElement/sheet), a `StyleSheet` is always associated with an `HTMLStyleElement` unless its `type` attribute is not `text/css`.

## Test plan

- [x] Unit test: disconnected → null, connected → CSSStyleSheet, non-CSS type → null, same instance on repeat
- [x] CDP integration test: dev build passes all checks, official nightly returns null (confirms new feature)
- [x] All 238 unit tests pass

### DIFF
--- a/src/browser/webapi/element/html/Style.zig
+++ b/src/browser/webapi/element/html/Style.zig
@@ -92,7 +92,7 @@ pub fn getSheet(self: *Style, page: *Page) !?*CSSStyleSheet {
     }
 
     if (self._sheet) |sheet| return sheet;
-    const sheet = try CSSStyleSheet.init(page);
+    const sheet = try CSSStyleSheet.initWithOwner(self.asElement(), page);
     self._sheet = sheet;
     return sheet;
 }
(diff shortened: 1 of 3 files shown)


src/browser/webapi/element/html/Style.zig
```diff
@@ -75,9 +77,24 @@ pub fn setDisabled(self: *Style, disabled: bool, page: *Page) !void {
...
+    }
+    const t = self.getType();
+    if (t.len != 0 and !std.ascii.eqlIgnoreCase(t, "text/css")) {
+        self._sheet = null;
+        return null;
+    }
+
+    if (self._sheet) |sheet| return sheet;
+    const sheet = try CSSStyleSheet.init(page);
+    self._sheet = sheet;
```
karlseguin, line 96: We can do it in a separate PR, but I think this is bidirectional .  CSSStyleSheet.ownerNode should return `self`.  Right now, it's hard-coded to return null (and the return type is wrong, but it doesn't matter, since we return null).

PR #1580, opened 2026-02-18
add HTMLImageElement.complete property

egrs:
## Summary

- Add the `complete` read-only property to `HTMLImageElement`
- Returns `true` for all images, since the headless browser does not fetch image resources
- Per [MDN spec](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/complete), `complete` is true when no src is set, the image is fully fetched, or the image is broken — all of which apply in a headless context

## Test plan

- [x] Unit test: no src, with src, createElement, Image constructor — all return true
- [x] CDP integration test: dev build passes, official nightly returns undefined (confirms new property)
- [x] All 238 unit tests pass


karlseguin:
Thank you!

PR #1583, opened 2026-02-18
add window stubs, HTMLElement hidden/tabIndex, input attribute reflections

egrs:
## Summary

Adds commonly expected properties across three areas:

**Window** (headless stubs)
- `alert()`, `confirm()`, `prompt()` — no-op stubs (alert returns undefined, confirm returns false, prompt returns null)
- `devicePixelRatio` — returns 1

**HTMLElement** (base class, applies to all elements)
- `hidden` — boolean attribute reflection
- `tabIndex` — integer attribute reflection (default -1)

**HTMLInputElement**
- `placeholder`, `min`, `max`, `step` — string attribute reflections
- `multiple` — boolean attribute reflection
- `autocomplete` — string attribute reflection

All follow existing codebase patterns for attribute reflection. Tests included for each property.

This can also be split up into multiple smaller PRs if preferred.

### DIFF
--- a/src/browser/tests/window/stubs.html
+++ b/src/browser/tests/window/stubs.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<script src="testing.js"></script>
+<script src="../testing.js"></script>
 
 <script id="alert">
   {
--- a/src/browser/webapi/Window.zig
+++ b/src/browser/webapi/Window.zig
@@ -704,14 +704,6 @@ fn getFunctionFromSetter(setter_: ?FunctionSetter) ?js.Function.Global {
     };
 }
 
-// Headless browser stubs: alert/confirm/prompt are no-ops
-fn jsAlert(_: *Window, _: ?[]const u8) void {}
-fn jsConfirm(_: *Window, _: ?[]const u8) bool {
-    return false;
-}
-fn jsPrompt(_: *Window, _: ?[]const u8, _: ?[]const u8) ?[]const u8 {
-    return null;
-}
 
 pub const JsApi = struct {
     pub const bridge = js.Bridge(Window);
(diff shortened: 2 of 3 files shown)


src/browser/tests/window/stubs.html
```diff
@@ -0,0 +1,34 @@
+<!DOCTYPE html>
```
karlseguin, line 1: This file should go in tests/window/stubs.html

src/browser/webapi/element/Html.zig
```diff
@@ -342,6 +342,37 @@ pub fn click(self: *HtmlElement, page: *Page) !void {
...
+            .anchor, .area, .button, .input, .select, .textarea, .iframe => 0,
+            else => -1,
+        };
+    };
+    return std.fmt.parseInt(i32, attr, 10) catch -1;
+}
+
+pub fn setTabIndex(self: *HtmlElement, value: i32, page: *Page) !void {
+    var buf: [12]u8 = undefined;
+    const str = std.fmt.bufPrint(&buf, "{d}", .{value}) catch return;
```
karlseguin: I'd probably `catch unreachable` here. But I'm fine with this also. It's a question of how much you trust the stdlib api to introduce a breaking change.

src/browser/webapi/Window.zig
```diff
@@ -704,6 +704,15 @@ fn getFunctionFromSetter(setter_: ?FunctionSetter) ?js.Function.Global {
     };
 }
 
+// Headless browser stubs: alert/confirm/prompt are no-ops
+fn jsAlert(_: *Window, _: ?[]const u8) void {}
```
karlseguin: I don't think there's any reason to prefix these with `js`. If you really don't want them exposed on the window, you can define them like:

```zig
pub const alert = bridge.function(struct {
   fn alert(_: *const Window, _: ?[]const u8) void {}
}.alert, .{});
```

src/browser/webapi/Window.zig
```diff
@@ -790,13 +782,23 @@ pub const JsApi = struct {
...
+            return null;
+        }
+    }.prompt, .{});
 };
 
 const testing = @import("../../testing.zig");
 test "WebApi: Window" {
     try testing.htmlRunner("window", .{});
-    try testing.htmlRunner("window-stubs.html", .{});
+    try testing.htmlRunner("window/stubs.html", .{});
```
karlseguin, line 803: Since the line above is a path to a folder, it'll automatically run all the files directly nested in it. I think if you leave this line in, it'll actually end up running the test twice.

karlseguin: I removed it in a separate PR: https://github.com/lightpanda-io/browser/pull/1591

Issue #1161, opened 2025-10-19
Improve focus/blur and text selection


krichprollsch:
We currently have a `active_element` in Document's state for the active input.
But it would be great to improve it to support also `blur`.

Moreover we need an additional state for the text selection.

* dispatch events on text selection, focus and blur
* listen these events to maintain a document's (or window) state of active element and selected text (I know active_element is in the document's state)
* implement `Selection` API
* take care of the current selection when filling a text input

Follows #1152

egrs:
Work done towards this issue:

- **#1584** — fix textarea text insertion to respect selection range
- **#1585** — dispatch focusin/focusout events with relatedTarget
- **#1588** — click on focusable elements calls focus() with events
- **#1589** — dispatch select event from input/textarea select()
- **#1590** — fix Range.toString() for cross-container and element ranges, uncomment ~800 lines of Range tests

karlseguin:
I believe this is all implemented. @krichprollsch

PR #1585, opened 2026-02-18
dispatch focusin/focusout events with relatedTarget

egrs:
## Summary

- `Element.focus()` and `Element.blur()` now dispatch all four spec-required FocusEvents in the correct order: `blur` → `focusout` → `focus` → `focusin`
- `focusin` and `focusout` bubble (per spec), while `focus` and `blur` do not
- All four events carry the correct `relatedTarget` (the other element involved in the focus change)
- Added `FocusEvent.initTrusted()` to create trusted FocusEvent instances (matching the pattern used by KeyboardEvent, ErrorEvent, etc.)
- Events are now proper `FocusEvent` instances instead of plain `Event`s

Relates to #1161

## Test plan

- [x] Unit tests pass (246/246)
- [x] New tests for focusin/focusout event order, bubbling behavior, and relatedTarget values
- [x] CDP integration test: 5/5 pass (event order, bubbling, relatedTarget)

### DIFF
--- a/src/browser/webapi/Element.zig
+++ b/src/browser/webapi/Element.zig
@@ -799,6 +799,8 @@ pub fn focus(self: *Element, page: *Page) !void {
         try page._event_manager.dispatch(old_target, focusout_event.asEvent());
     }
 
+    // Must be set after blur/focusout and before focus/focusin —
+    // event dispatch can reset _active_element if set earlier.
     if (self.asNode().isConnected()) {
         page.document._active_element = self;
     }


egrs:
Checked the tests against the spec — they're correct.

**Event order** follows what all browsers actually implement (blur → focusout → focus → focusin). The UI Events spec technically defines a [different order](https://w3c.github.io/uievents/#events-focusevent-event-order) with focusin/focusout firing *before* blur/focus, but no browser does that — the [MDN FocusEvent page](https://developer.mozilla.org/en-US/docs/Web/API/FocusEvent#order_of_events) notes this discrepancy explicitly.

**Event properties** are all correct per spec:
- All four are `FocusEvent` instances (not plain `Event`)
- `composed: true` for all four
- `cancelable: false` for all four
- `bubbles`: true for focusin/focusout, false for focus/blur
- `relatedTarget`: points to the element gaining/losing focus respectively

**Firefox failures**: Firefox is [known to not fire focus events when the window doesn't have OS focus](https://bugzilla.mozilla.org/show_bug.cgi?id=1398111) — programmatic `.focus()` becomes a no-op. This explains the `got: ""` result. The existing `focus_events` test at line 25 has the same pattern and would fail under the same conditions.

So the difference with Firefox is expected — it's a Firefox window-focus quirk, not a test correctness issue.

PR #1590, opened 2026-02-18
fix Range.toString() for cross-container and element ranges

egrs:
## Summary

- Fix `Range.toString()` to properly handle cross-container ranges and element container ranges (previously returned empty string)
- Implement spec-compliant tree walking: partial start text, fully-contained text nodes in tree order, partial end text
- Exclude Comment and ProcessingInstruction nodes per spec (only Text/CDATASection)
- Uncomment ~800 lines of Range tests that were disabled during the v8 childNodes caching refactor (f4a5f73a)
- Uncomment cross-node deleteContents/extractContents/cloneContents tests
- Add 5 new toString tests: same-text, same-element, cross-container siblings, cross-container nested, comment exclusion

Contributes to #1161

## Test plan

- [x] `make test` — 246/246 pass
- [x] CDP integration test against dev container — 6/6 pass
- [x] Nightly container — 4/6 fail as expected (cross-container not supported yet)
- [x] Perf baseline — 0.00 KB/cycle (no persistent allocations, toString uses call_arena)

### DIFF
--- a/src/browser/tests/range.html
+++ b/src/browser/tests/range.html
@@ -7,7 +7,7 @@
   <span id="s1">Span content</span>
 </div>
 
-<!-- <script id=basic>
+<script id=basic>
 {
   const range = document.createRange();
   testing.expectEqual('object', typeof range);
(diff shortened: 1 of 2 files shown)


karlseguin:
Thanks.

PR #1596, opened 2026-02-19
Improve Animation support: async update from idle => running => finished

krichprollsch:


src/browser/webapi/animation/Animation.zig
```diff
@@ -72,26 +145,87 @@ pub fn setTimeline(self: *Animation, timeline: ?js.Object.Global) !void {
     self._timeline = timeline;
 }
 
+pub fn getStartTime(self: *const Animation) ?f64 {
+    return self._startTime;
+}
+
+pub fn setStartTime(self: *Animation, value: ?f64, page: *Page) !void {
+    self._startTime = value;
+    return self.play(page);
```
karlseguin, line 165: Is calling `play` correct when `value` is null?

src/browser/webapi/animation/Animation.zig
```diff
@@ -72,26 +145,87 @@ pub fn setTimeline(self: *Animation, timeline: ?js.Object.Global) !void {
...
+pub fn getOnFinish(self: *const Animation) ?js.Function.Temp {
+    return self._onFinish;
+}
+
+pub fn deinit(self: *Animation, _: bool) void {
+    self._page.releaseArena(self._arena);
+}
+
+// callback function transitionning from a state to another
+fn update(ctx: *anyopaque) anyerror!?u32 {
```
karlseguin:
```suggestion
fn update(ctx: *anyopaque) !?u32 {
```

(I tried it, is fine)

src/browser/webapi/animation/Animation.zig
```diff
@@ -16,40 +16,113 @@
...
+
+pub fn play(self: *Animation, page: *Page) !void {
+    if (self._playState == .running) {
+        return;
+    }
+
+    // transition to running.
+    self._playState = .running;
+
+    // Schedule the transition from .idle => .running in 10ms.
```
karlseguin: It's already in running now, not in 10ms?

src/browser/webapi/animation/Animation.zig
```diff
@@ -72,26 +145,87 @@ pub fn setTimeline(self: *Animation, timeline: ?js.Object.Global) !void {
...
+pub fn setStartTime(self: *Animation, value: ?f64, page: *Page) !void {
+    self._startTime = value;
+    return self.play(page);
+}
+
+pub fn getOnFinish(self: *const Animation) ?js.Function.Temp {
+    return self._onFinish;
+}
+
+pub fn deinit(self: *Animation, _: bool) void {
```
karlseguin: Super minor, but...can you move this under `init`? 

Also, the `self` itself isn't freed. In init, either use the arena to create the animation, or here call factory.destroy. I probably prefer the 1st one (I think it's what we do in most of these cases).

PR #1598, opened 2026-02-19
input value sanitization per WHATWG spec

egrs:
## Summary

Implements proper value sanitization for `<input>` types per the WHATWG HTML spec, fixing 13 WPT test files (147 new passing subtests).

**Changes (all in `Input.zig`):**

- **number/range**: Replace `std.fmt.parseFloat` with WHATWG "valid floating-point number" grammar validator — rejects `"+1"`, `"1."`, `"Infinity"`, `"NaN"`, overflow (`"2e308"`), leading whitespace, trailing junk
- **date**: Validate `YYYY-MM-DD` with month/day range checks and leap year support
- **time**: Validate `HH:MM[:SS[.s{1,3}]]` with hour/minute/second range checks
- **month**: Validate `YYYY-MM`
- **week**: Validate `YYYY-Www` with ISO 8601 max-week calculation
- **datetime-local**: Validate `date + (T|space) + time`, normalize to shortest form with `T` separator
- **color**: Normalize uppercase hex to lowercase per spec
- **checkbox/radio**: Return `"on"` as default value per "default/on" value mode
- **created()**: Sanitize initial value from HTML attributes after input type is set

## WPT results

13 test files flipped from failing to fully passing:

| Test file | Before | After |
|-----------|--------|-------|
| `number.html` | 27/33 | **33/33** |
| `type-change-state.html` | 329/380 | **380/380** |
| `type-change-state-weekmonth.html` | 38/82 | **82/82** |
| `datetime-local.html` | 2/17 | **17/17** |
| `color.tentative.html` | 21/23 | **23/23** |
| `week.html` | 7/16 | **16/16** |
| `time-2.html` | 10/17 | **17/17** |
| `telephone.html` | 10/13 | **13/13** |
| `date.html` | 4/6 | **6/6** |
| `url.html` | 2/4 | **4/4** |
| `valueMode-weekmonth.html` | 0/4 | **4/4** |
| `input-type-checkbox.html` | 6/7 | **7/7** |
| `datetime-local-trailing-zeros.html` | 0/1 | **1/1** |

6 additional files improved (not yet fully passing):

| Test file | Before | After |
|-----------|--------|-------|
| `month.html` | 5/15 | 14/15 |
| `valueMode.html` | 32/40 | 38/40 |
| `range.html` | 10/25 | 11/25 |
| `time.html` | 8/32 | 9/32 |

## Test plan

- [x] Unit tests: 247/247 pass
- [x] WPT tests: 13 files newly passing, 0 regressions
[...]

PR #1599, opened 2026-02-19
Improve and fix sanitized value ownership.

karlseguin:
1 - Fix an issue where build would persist a value in the call_arena 
2 - Remove double allocation (call_arena -> page_arena) 
3 - Improve ergonomics of sanitizeValue with a comptime value indicating whether or not to always dupe the value.


karlseguin:
Better to merge https://github.com/lightpanda-io/browser/pull/1598 first then reply these changes to that.

PR #1603, opened 2026-02-19
spec compliance: missing validation guards

egrs:
## Summary
- **Event.preventDefault()** and **returnValue setter** now respect `cancelable=false` per spec (no-op when not cancelable)
- **MutationObserver.observe()** validates options: throws TypeError when none of childList/attributes/characterData are true, or when attributeOldValue/attributeFilter conflict with attributes=false, or characterDataOldValue conflicts with characterData=false. Correctly distinguishes "omitted" (implies true) from "explicitly false".
- **Detached checkbox/radio** click no longer fires input/change events (spec requires element to be connected)
- **Doctype insertion** into non-Document parent now throws HierarchyRequestError
- **error.TypeError** now maps to a proper JS TypeError instance (not generic Error)
- **dom_exception flag** enabled on Element and DocumentFragment mutation methods (prepend, append, before, after, replaceWith, replaceChildren)

## WPT impact
- 6 test files newly passing (+29 subtests, 0 regressions)
- `dom/events/Event-defaultPrevented.html` 6/8 → 8/8
- `dom/events/Event-returnValue.html` 5/7 → 7/7
- `dom/events/Event-dispatch-detached-input-and-change.html` 8/12 → 12/12
- `dom/nodes/MutationObserver-sanity.html` 9/13 → 13/13
- `dom/nodes/ParentNode-prepend.html` 20/22 → 22/22
- `navigation-api/currententrychange-event/not-on-load.html` (bonus)

## Test plan
- [x] Unit tests: 260/260 pass
- [x] WPT: 6 files newly passing, +29 subtests, 0 real regressions

### DIFF
--- a/src/browser/webapi/Element.zig
+++ b/src/browser/webapi/Element.zig
@@ -1594,7 +1594,7 @@ pub const JsApi = struct {
     fn _attachShadow(self: *Element, init: ShadowRootInit, page: *Page) !*ShadowRoot {
         return self.attachShadow(init.mode, page);
     }
-    pub const replaceChildren = bridge.function(Element.replaceChildren, .{});
+    pub const replaceChildren = bridge.function(Element.replaceChildren, .{ .dom_exception = true });
     pub const replaceWith = bridge.function(Element.replaceWith, .{ .dom_exception = true });
     pub const remove = bridge.function(Element.remove, .{});
     pub const append = bridge.function(Element.append, .{ .dom_exception = true });


src/browser/webapi/Element.zig
```diff
@@ -1595,12 +1595,12 @@ pub const JsApi = struct {
         return self.attachShadow(init.mode, page);
     }
     pub const replaceChildren = bridge.function(Element.replaceChildren, .{});
```
karlseguin: From the PR description, this should be `.{ . dom_exception = true }` also.

PR #1604, opened 2026-02-19
Introduce NodeType enum for Node.getNodeType()

arrufat:
I thought I'd introduce an enum for the node type instead of just having numbers.
It clutters the code base a bit with casts, so feel free to reject.

### DIFF
--- a/src/browser/markdown.zig
+++ b/src/browser/markdown.zig
@@ -75,15 +75,16 @@ pub fn dump(node: *Node, opts: Opts, writer: *std.Io.Writer, page: *Page) !void 
 }
 
 fn render(node: *Node, state: *State, writer: *std.Io.Writer, page: *Page) error{WriteFailed}!void {
-    switch (node._type) {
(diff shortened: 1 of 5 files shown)


src/browser/markdown.zig
```diff
@@ -75,15 +75,16 @@ pub fn dump(node: *Node, opts: Opts, writer: *std.Io.Writer, page: *Page) !void
 }
 
 fn render(node: *Node, state: *State, writer: *std.Io.Writer, page: *Page) error{WriteFailed}!void {
-    switch (node._type) {
+    switch (node.getNodeType()) {
```
karlseguin, line 78: Nice PR. But this one seems a bit unnecessary, it's adding an additional switch to turn it from on enum to another, and then switch on the second one...and you lose the nice (and I assume, more efficient then then having to .as(X) again) union capture.

arrufat: Yeah, I was just experimenting with the codebase. Closing.

PR #1606, opened 2026-02-19
More pseudo-seletors

karlseguin:
:invalid, :valid and :indeterminate implementation

Fix Element.closest with :scope pseudo-selector.

### DIFF
--- a/src/browser/webapi/selector/List.zig
+++ b/src/browser/webapi/selector/List.zig
@@ -531,11 +531,45 @@ fn matchesPseudoClass(el: *Node.Element, pseudo: Selector.PseudoClass, scope: *N
         .enabled => {
             return el.getAttributeSafe(comptime .wrap("disabled")) == null;
         },
-        .indeterminate => return false,
+        .indeterminate => {
+            const input = el.is(Node.Element.Html.Input) orelse return false;
+            return switch (input._input_type) {
+                .checkbox => input.getIndeterminate(),
+                else => false,
+            };
+        },
 
         // Form validation
-        .valid => return false,
-        .invalid => return false,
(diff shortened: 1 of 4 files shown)


src/browser/webapi/selector/List.zig
```diff
@@ -531,11 +531,45 @@ fn matchesPseudoClass(el: *Node.Element, pseudo: Selector.PseudoClass, scope: *N
...
+            };
+        },
 
         // Form validation
-        .valid => return false,
-        .invalid => return false,
+        .valid => {
+            if (el.is(Node.Element.Html.Input)) |input| {
+                return switch (input._input_type) {
+                    .hidden, .submit, .reset, .button => false,
```
krichprollsch, line 546: These are always invalid? Shouldn't be `true` instead?

karlseguin: :valid matches an "elements that are [candidates for constraint validation](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#candidate-for-constraint-validation)"

And type=hidden is "If an [input](https://html.spec.whatwg.org/multipage/input.html#the-input-element) element's [type](https://html.spec.whatwg.org/multipage/input.html#attr-input-type) attribute is in the [Hidden](https://html.spec.whatwg.org/multipage/input.html#hidden-state-(type=hidden)) state, it is [barred from constraint validation](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#barred-from-constraint-validation)"

same for the other ones.

It isn't so much that they are "valid" or "invalid", it's that validity doesn't apply to them anymore than it applies to a p tag.

PR #1607, opened 2026-02-19
markdown: handle block-level and standalone anchors

arrufat:
Adds logic to detect if an anchor contains block descendants or is a standalone element within a layout block. These are now rendered with appropriate spacing and link formatting. Also adds `.main` to the list of block elements.

Follow up of:
- #1560

### DIFF
--- a/src/browser/markdown.zig
+++ b/src/browser/markdown.zig
@@ -19,6 +19,7 @@
 const std = @import("std");
 
 const Page = @import("Page.zig");
+const CData = @import("webapi/CData.zig");
 const Element = @import("webapi/Element.zig");
 const Node = @import("webapi/Node.zig");
 
@@ -68,24 +69,23 @@ fn isLayoutBlock(tag: Element.Tag) bool {
 fn isStandaloneAnchor(el: *Element) bool {
     const node = el.asNode();
     const parent = node.parentNode() orelse return false;
+    const parent_el = parent.is(Element) orelse return false;
 
-    if (parent._type != .element) return false;
-    const parent_el = parent.as(Element);
     if (!isLayoutBlock(parent_el.getTag())) return false;
 
     var prev = node.previousSibling();
     while (prev) |p| : (prev = p.previousSibling()) {
         if (isSignificantText(p)) return false;
-        if (p._type == .element) {
-            if (isVisibleElement(p.as(Element))) break;
+        if (p.is(Element)) |pe| {
+            if (isVisibleElement(pe)) break;
         }
     }
 
     var next = node.nextSibling();
     while (next) |n| : (next = n.nextSibling()) {
         if (isSignificantText(n)) return false;
-        if (n._type == .element) {
-            if (isVisibleElement(n.as(Element))) break;
+        if (n.is(Element)) |ne| {
+            if (isVisibleElement(ne)) break;
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/browser/markdown.zig
```diff
@@ -58,6 +58,74 @@ fn shouldAddSpacing(tag: Element.Tag) bool {
...
+        .main, .section, .article, .nav, .aside, .header, .footer, .div, .ul, .ol => true,
+        else => false,
+    };
+}
+
+fn isStandaloneAnchor(el: *Element) bool {
+    const node = el.asNode();
+    const parent = node.parentNode() orelse return false;
+
+    if (parent._type != .element) return false;
```
karlseguin: const parent_el = parent.is(Element) orelse return false;

src/browser/markdown.zig
```diff
@@ -58,6 +58,74 @@ fn shouldAddSpacing(tag: Element.Tag) bool {
...
+    const parent_el = parent.as(Element);
+    if (!isLayoutBlock(parent_el.getTag())) return false;
+
+    var prev = node.previousSibling();
+    while (prev) |p| : (prev = p.previousSibling()) {
+        if (isSignificantText(p)) return false;
+        if (p._type == .element) {
```
karlseguin:
```zig
if (p.is(Element)) |pe| {
   if (isVisibleElement(pe)) break;
}
```

src/browser/markdown.zig
```diff
@@ -58,6 +58,74 @@ fn shouldAddSpacing(tag: Element.Tag) bool {
...
+            if (isVisibleElement(p.as(Element))) break;
+        }
+    }
+
+    var next = node.nextSibling();
+    while (next) |n| : (next = n.nextSibling()) {
+        if (isSignificantText(n)) return false;
+        if (n._type == .element) {
```
karlseguin: can also use `.is` here

src/browser/markdown.zig
```diff
@@ -58,6 +58,74 @@ fn shouldAddSpacing(tag: Element.Tag) bool {
...
+            if (isVisibleElement(n.as(Element))) break;
+        }
+    }
+
+    return true;
+}
+
+fn isSignificantText(node: *Node) bool {
+    if (node._type != .cdata) return false;
```
karlseguin:
```
const text = node.is(Node.Data.Text) orelse return false;
return !isAllWhitespace(test._proto.getData());
```

(we should probably have a getData() on Text directly, or at least a `asCData()`..but since we don't right now, need to use `_proto` directly)

src/browser/markdown.zig
```diff
@@ -58,6 +58,74 @@ fn shouldAddSpacing(tag: Element.Tag) bool {
...
+fn isAllWhitespace(text: []const u8) bool {
+    return for (text) |c| {
+        if (!std.ascii.isWhitespace(c)) break false;
+    } else true;
+}
+
+fn hasBlockDescendant(node: *Node) bool {
+    var it = node.childrenIterator();
+    while (it.next()) |child| {
+        if (child._type == .element) {
```
karlseguin: Same here.

```zig
if (child.is(Element)) |el| {
   ....
```

src/browser/markdown.zig
arrufat: It turns out there's already `Text.getWholeText()`

arrufat:
Thank you for the code review. I will follow the idioms more closely next time.

PR #1608, opened 2026-02-19
add DOMException legacy error code constants

egrs:
## Summary

- Add all 25 legacy error code constants (`INDEX_SIZE_ERR` through `DATA_CLONE_ERR`) to DOMException via `bridge.property(N, .{ .template = true })`, exposing them on both the constructor and prototype.
- Enables WPT `assert_throws_dom` checks that compare `e.code` against `DOMException.NOT_FOUND_ERR` etc.

## Test plan

- [x] `make test` — 260/260 pass
- [x] CI green (zig fmt, zig test, e2e)

### DIFF
--- a/src/browser/webapi/DOMException.zig
+++ b/src/browser/webapi/DOMException.zig
@@ -195,6 +195,33 @@ pub const JsApi = struct {
     pub const name = bridge.accessor(DOMException.getName, null, .{});
     pub const message = bridge.accessor(DOMException.getMessage, null, .{});
     pub const toString = bridge.function(DOMException.toString, .{});
+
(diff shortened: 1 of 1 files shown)


karlseguin:
I like this PR, but when I saw the logic a 3rd time, I figured a discriminatory type would be better. I think I prefer the approach in https://github.com/lightpanda-io/browser/pull/1610 for the stringification portion of this PR; mostly because I suspect that this should be a common pattern used in multiple places.

If you agree, can you update (or close this PR and open a new one) with only the DOMException change?

egrs:
Good for me, updated this PR with your suggestions. Let's go with your approach with #1610.

PR #1612, opened 2026-02-20
fix ChildNode after() and replaceWith() sibling ordering

egrs:
## Summary

- `after()` captured `node.nextSibling()` once, which went stale when that sibling was in the insertion list. Now uses `viableNextSibling()` per DOM spec to find the first following sibling not in the nodes list.
- `replaceWith()` in CData had the same stale-reference bug and also removed self before inserting (unlike Element.replaceWith). Now uses self as the insertion anchor: insert before self, remove self at end.
- Added `NodeOrText.viableNextSibling()` helper to Node.zig.

## WPT results

|---|---|---|
| ChildNode-after.html | 33/45 | **45/45** |
| ChildNode-replaceWith.html | 27/33 | **33/33** |
| ChildNode-before.html | 45/45 | 45/45 (no regression) |

## Test plan

- [x] `make test` — 260/260 pass
- [x] WPT ChildNode-after: 45/45
- [x] WPT ChildNode-replaceWith: 33/33
- [x] WPT ChildNode-before: 45/45 (regression check)

### DIFF
--- a/src/browser/webapi/Node.zig
+++ b/src/browser/webapi/Node.zig
@@ -1069,6 +1069,25 @@ pub const NodeOrText = union(enum) {
             .text => |txt| page.createTextNode(txt),
         };
     }
+
+    /// DOM spec: first following sibling of `node` that is not in `nodes`.
+    pub fn viableNextSibling(node: *Node, nodes: []const NodeOrText) ?*Node {
+        var sibling = node.nextSibling() orelse return null;
+        blk: while (true) {
+            for (nodes) |n| {
+                switch (n) {
+                    .node => |nn| if (sibling == nn) {
+                        sibling = sibling.nextSibling() orelse return null;
+                        continue :blk;
+                    },
+                    .text => {},
+                }
+            } else {
+                return sibling;
+            }
+        }
+        return null;
+    }
 };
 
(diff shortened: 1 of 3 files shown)


src/browser/webapi/Node.zig
```diff
@@ -1074,6 +1069,26 @@ pub const NodeOrText = union(enum) {
             .text => |txt| page.createTextNode(txt),
         };
     }
+
+    /// DOM spec: first following sibling of `node` that is not in `nodes`.
+    pub fn viableNextSibling(node: *Node, nodes: []const NodeOrText) ?*Node {
```
karlseguin, line 1074: You can avoid that 1 condition by immediately continuing the loop when you've found a match:

```zig
    pub fn viableNextSibling(node: *Node, nodes: []const NodeOrText) ?*Node {
        var sibling = node.nextSibling() orelse return null;
        blk: while (true) {
            for (nodes) |n| {
                switch (n) {
                    .node => |nn| if (sibling == nn) {
                        sibling = sibling.nextSibling() orelse return null;
                        continue :blk;
                    },
                    .text => {},
                }
            } else {
                return sibling;
            }
        }
        return null;
    }
```

PR #1613, opened 2026-02-20
implement Node.lookupNamespaceURI() and isDefaultNamespace()

egrs:
## Summary

- Implement `Node.lookupNamespaceURI()` and `Node.isDefaultNamespace()` per DOM spec
- Store custom namespace URIs in a page lookup for elements created via `createElementNS` with unknown namespaces
- Fix `setAttributeNS` to preserve qualified names for xmlns namespace declarations (e.g. `xmlns:bar` instead of just `bar`)
- Move `_prefix()` helper from JsApi to Element level for reuse by lookup algorithm

## Test plan

- [x] `make test` — 280/280 pass
- [x] WPT `dom/nodes/Node-lookupNamespaceURI.html` — 75/75 pass (was 0/75)
- [x] Broader `dom/nodes/` WPT suite — no regressions

### DIFF
--- a/src/browser/webapi/Document.zig
+++ b/src/browser/webapi/Document.zig
@@ -160,6 +160,14 @@ pub fn createElementNS(self: *Document, namespace: ?[]const u8, name: []const u8
     const normalized_name = if (ns == .html) std.ascii.lowerString(&page.buf, name) else name;
     const node = try page.createElementNS(ns, normalized_name, null);
 
+    // Store original URI for unknown namespaces so lookupNamespaceURI can return it
+    if (ns == .unknown) {
+        if (namespace) |uri| {
+            const duped = try page.dupeString(uri);
+            try page._element_namespace_uris.put(page.arena, node.as(Element), duped);
+        }
+    }
+
     // Track owner document if it's not the main document
     if (self != page.document) {
         try page.setNodeOwnerDocument(node, self);
--- a/src/browser/webapi/Element.zig
+++ b/src/browser/webapi/Element.zig
@@ -49,6 +49,7 @@ pub const ClassListLookup = std.AutoHashMapUnmanaged(*Element, *collections.DOMT
 pub const RelListLookup = std.AutoHashMapUnmanaged(*Element, *collections.DOMTokenList);
 pub const ShadowRootLookup = std.AutoHashMapUnmanaged(*Element, *ShadowRoot);
 pub const AssignedSlotLookup = std.AutoHashMapUnmanaged(*Element, *Html.Slot);
+pub const NamespaceUriLookup = std.AutoHashMapUnmanaged(*Element, []const u8);
 
 pub const ScrollPosition = struct {
     x: u32 = 0,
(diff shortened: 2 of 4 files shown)


src/browser/webapi/Document.zig
```diff
@@ -160,6 +160,14 @@ pub fn createElementNS(self: *Document, namespace: ?[]const u8, name: []const u8
     const normalized_name = if (ns == .html) std.ascii.lowerString(&page.buf, name) else name;
     const node = try page.createElementNS(ns, normalized_name, null);
 
+    // Store original URI for unknown namespaces so lookupNamespaceURI can return it
+    if (ns == .unknown) {
+        if (namespace) |uri| {
+            const duped = try page.arena.dupe(u8, uri);
```
karlseguin:
```suggestion
            const duped = try page.dupeString(uri);
```

A chance to be interned (we should increase our pool of interned strings at some point)

src/browser/webapi/Element.zig
```diff
@@ -364,6 +365,65 @@ pub fn getNamespaceURI(self: *const Element) ?[]const u8 {
...
+        if (match) return ns_uri;
+    }
+
+    // Step 2: search xmlns attributes
+    if (self._attributes) |attrs| {
+        var iter = attrs.iterator();
+        while (iter.next()) |entry| {
+            const name = entry._name.str();
+            if (prefix == null) {
+                if (std.mem.eql(u8, name, "xmlns")) {
```
karlseguin: `entry._name.eql(comptime .wrap("xmlns"))` could be a bit faster..you could move `const name = entry._name.str();` into the `else` block.

src/browser/webapi/Element.zig
```diff
@@ -364,6 +365,65 @@ pub fn getNamespaceURI(self: *const Element) ?[]const u8 {
...
+pub fn lookupNamespaceURIForElement(self: *Element, prefix: ?[]const u8, page: *Page) ?[]const u8 {
+    // Hardcoded reserved prefixes
+    if (prefix) |p| {
+        if (std.mem.eql(u8, p, "xml")) return "http://www.w3.org/XML/1998/namespace";
+        if (std.mem.eql(u8, p, "xmlns")) return "http://www.w3.org/2000/xmlns/";
+    }
+
+    // Step 1: check element's own namespace/prefix
+    const ns_uri = self.getNamespaceUri(page);
+    if (ns_uri != null) {
```
```zig
if (self.getNamespaceUri(page)) |ns_url| {
  ....
}
```

The more narrowly scoped variable makes it easier to read. I don't have to keep `ns_uri` in the tiny context that my brain supports.

src/browser/webapi/Document.zig
```diff
@@ -160,6 +160,14 @@ pub fn createElementNS(self: *Document, namespace: ?[]const u8, name: []const u8
...
+            try page._element_namespace_uris.put(page.arena, node.as(Element), duped);
```
karlseguin, line 167: I should have thought of that 😊

PR #1536, opened 2026-02-12
wpt: use official wpt http server

krichprollsch:
Adapt wpt test runner to request official wpt HTTP server and run all the testharness tess.

It first parses the `MANIFEST.json` file to list all the tests to run.

See `README` change for installation instructions.

PR #1633, opened 2026-02-23
remove WPT specific code

krichprollsch:
Using both lightpanda-io/wpt and lightpanda-io/demo/wptrunner remove the need for code specific to run WPT from browser.

Depends on https://github.com/lightpanda-io/demo/pull/117
Replaces #1536 

* [x] Update readme
* [x] Remove code
* [x] Update CI

### DIFF
--- a/.github/workflows/wpt.yml
+++ b/.github/workflows/wpt.yml
@@ -107,7 +107,7 @@ jobs:
         run: |
           ./wpt serve 2> /dev/null & echo $! > WPT.pid
           sleep 10s
-          ./wptrunner -lpd-path ./lightpanda -json -concurrency 5 > wpt.json
+          ./wptrunner -lpd-path ./lightpanda -json -concurrency 1 > wpt.json
           kill `cat WPT.pid`
 
       - name: write commit


.github/workflows/wpt.yml
```diff
@@ -30,11 +30,85 @@ jobs:
...
+          name: wptrunner
+
+      - run: chmod a+x ./wptrunner
 
       - name: run test with json output
-        run:  zig-out/bin/lightpanda-wpt --json > wpt.json
+        run: |
+          ./wpt serve & echo $! > WPT.pid
+          sleep 10s
+          ./lightpanda serve & echo $! > LPD.pid
```
karlseguin: I think we need `--insecure_disable_tls_host_verification` here, because some tests connect to the python server over TLS, and probably also need `--log_level error`   because some tests console.log literally millions of items.

karlseguin (changes requested):

karlseguin:
Besides the minor change to the arguments for starting lightpanda, my concern is getting a reliable WPT report/run. I get a number of failures, e.g. RecvError, that (a) don't seem to consistently happen and (b) seem related to the test running environment and not something specific to browser (though I could be wrong). Plus I haven't gotten a full run completed on the new runner yet.

(Not that I mind removing this now and working on the new runner...we've gone long periods of time without working WPT reports before).

krichprollsch:
I pushed a change into https://github.com/lightpanda-io/demo/pull/117 to take an optional `--lpd-path` which enable autorestart browser on crash.
Using this option in the CI warranty the test suite will run until the end, even if the browser crash during the run.
I set `crash` result for the failed inflight tests during crash. But It's only an approximation b/c we don't know exactly which one caused the crash.

With this change, I can run the whole suite.

PR #1640, opened 2026-02-24
Dispatch `load` events that're attached after `documentIsComplete`

nikneym:


krichprollsch:
I think you should also call `page.linkAddedCallback` in `Link.setHref` in the same way we do on `Script.setSrc`.

Can you add tests into `link.html` to test load event dispatch?

PR #1644, opened 2026-02-24
Prefer `hive` for storing timers

nikneym:
_I'd totally understand if this won't get merged, had to scratch an itch of mine._

This PR changes the way we store timers/timer-likes by introducing a new map-like type. Pages create numerous `ScheduleCallback` objects; the goal of this PR is to optimize this; `hive` lazily-allocates blocks of `ScheduleCallback`s, taking the stress off of allocators and introduces better cache locality. It also reuses the same memory if `pooling` mode is active. Lastly, it allows reuse of a timer ID if its no longer owned by any. The tradeoff is more memory being used.

### DIFF
--- a/src/browser/webapi/Window.zig
+++ b/src/browser/webapi/Window.zig
@@ -19,6 +19,7 @@
 const std = @import("std");
 const js = @import("../js/js.zig");
 const builtin = @import("builtin");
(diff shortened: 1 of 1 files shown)


karlseguin:
I don't think the complexity of a new allocator is worth it for the scheduler.

PR #1646, opened 2026-02-25
Add url encoding option to URL.resolve

karlseguin:
Given:

a.href = "over 9000!"

Then:

a.href === BASE_URL + '/over%209000!';

This commits adds an escape: bool option to URL.resolve which will escape the path, query and fragment when true.

Also changes the Anchor, Image, Link and IFrame getSrc to escape. Escaping is also used when navigating a frame.


karlseguin:
@krichprollsch I didn't do it in this PR, but should the main URL we pass to navigate also be escaped?  If you:

```
curl "https://lightpanda.io/hello world"
```

You'll get an error, if you `./lightpanda fetch "https://lightpanda.io/hello world"`  you get that same error.

If we decide _not_ to escape the main navigation url, then we need to change https://github.com/lightpanda-io/demo/pull/117/changes#diff-15ff6b6f810cbccf0c86f043b8815c1ca9e84110c983dba1958ac3f993f311b5R286 to escape the URL..because, right now, many WPT tests are failing to navigate because it's an invalid URL.

krichprollsch:
I will update the runner to escape the URL, seems OK to refuse invalid URL input.

krichprollsch:
Hum, I tested with Chrome and cdpcli, and calling `cdpcli dump "https://httpbin.io/get?foo bar"` on chrome, it works... So I guess we should escape URLs after all...

But it's not that easy, we musn't double escape the url:

```
$ cdpcli dump "https://httpbin.io/get?foo bar"
// ...
  "url": "http://httpbin.io/get?foo%20bar"
```

```
$ cdpcli dump "https://httpbin.io/get?foo%20bar"
// ...
  "url": "http://httpbin.io/get?foo%20bar"
```

And this is not a feature from the client:
```
> {"method":"Page.navigate","params":{"url":"https://httpbin.io/get?foo bar","frameId":"768705AEC2AD8746E3F1AD4AAC574E3C"},"id":29,"sessionId":"FF654D0CB6DF14FD02F5EB14FF9FEECF"}
< {"method":"Page.frameStartedNavigating","params":{"frameId":"768705AEC2AD8746E3F1AD4AAC574E3C","url":"https://httpbin.io/get?foo%20bar","loaderId":"04DA9A2AF5CD30B261875EB8D0542772"
,"navigationType":"differentDocument"},"sessionId":"FF654D0CB6DF14FD02F5EB14FF9FEECF"}
```

PR #1649, opened 2026-02-25
fix: handle trailing slash on /json/version CDP endpoint

ireydiak:
# Summary

Some CDP clients (e.g. `playwright-go`) request `/json/version` with a trailing slash. The server previously only matched `/json/version` exactly, causing these clients to receive no response.
This fix adds handling for the trailing slash variant using the same logic and response as the existing `/json/version` handler.

### DIFF
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -521,7 +521,7 @@ pub const Client = struct {
             return true;
         }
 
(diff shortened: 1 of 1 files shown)


ireydiak:

karlseguin:
Thanks!

PR #1650, opened 2026-02-25
Updates dependencies and their build

mrdimidium:
I needed to update libcurl and then everything was a blur :/

This PR updates `curl`, `zlib`, `brotli`, and `nghttp2` to the latest versions and updates the build.

Build changes:
- dependencies are installed using zig fetch instead of submodules.
- each dependency is built in a separate module (to fix an issue with the global macro space)
- libcurl uses the official config file instead of manually installing defines
- html5ever uses an explicit file dependency instead of `dependOn(opts)`

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -642,7 +642,7 @@ fn buildCurl(
             "-DBUILDING_LIBCURL",
         },
         .files = &.{
-            // You can include all files from lib, libcurl uses an #ifdef-guards to exclude code for disabled functions
+            // You can include all files from lib, libcurl uses #ifdef-guards to exclude code for disabled functions
             "altsvc.c",              "amigaos.c",              "asyn-ares.c",
             "asyn-base.c",           "asyn-thrdd.c",           "bufq.c",
             "bufref.c",              "cf-h1-proxy.c",          "cf-h2-proxy.c",


karlseguin:
I don't have a strong opinion or, frankly, understanding of builds.  All I can say is that it works on MacOS and the ReleaseFast binary is the same-ish size.  I'm fine with anything that builds.

krichprollsch:
What is the impact on v8 build? I have to recompile fully v8, I didn't expect that :thinking:

mrdimidium:
@krichprollsch, I'm not sure what this is related to. The V8 build changed the way the inspector_subtype is passed, which could have called `gn gen`. Please check that this is a one-time issue and that rebuilds with this PR work as expected.

build.zig
```diff
@@ -171,561 +174,539 @@ pub fn build(b: *Build) !void {
...
-            root ++ "lib/vauth/spnego_sspi.c",
-            root ++ "lib/vauth/vauth.c",
-            root ++ "lib/vtls/cipher_suite.c",
-            root ++ "lib/vtls/openssl.c",
-            root ++ "lib/vtls/hostcheck.c",
-            root ++ "lib/vtls/keylog.c",
-            root ++ "lib/vtls/vtls.c",
-            root ++ "lib/vtls/vtls_scache.c",
-            root ++ "lib/vtls/x509asn1.c",
+            // You can include all files from lib, libcurl uses an #ifdef-guards to exclude code for disabled functions
```
nikneym:
```suggestion
            // You can include all files from lib, libcurl uses #ifdef-guards to exclude code for disabled functions
```

krichprollsch:
My bad, the rebuild was due to a change in `main`, not relative to this branch.

PR #1651, opened 2026-02-25
Run the MessageLoop [a lot] more.

karlseguin:
Depends on: https://github.com/lightpanda-io/zig-v8-fork/pull/152

We previously ran the message loop every 250ms. This commit changes it to run on every tick (much more frequently). It also runs microtasks after draining the message loop (since it can generate microtasks).

Also, we use to run microtasks after each script execution. Now we drain the message Loop + microtasks.

We still only drain the microtasks when executing v8 callbacks.

As part of this change, we also adjust our wait time based on whether or not there are pending background tasks in v8 in order to try to execute them (in general) and in a timely manner.

The goal is to ensure that tasks v8 enqueued on the foreground thread are executed promptly.

This change is particularly useful for calls to webassembly as compilation happens in the background and eventually requires the message loop to be drained to continue.

Previously, if a script did `await WebAssembly.instantiate(....)`, there was a good chance we'd never finish the code - we'd wait too long to run the message loop AND, after running it, we wouldn't necessarily resolve the promise.

### DIFF
--- a/.github/actions/install/action.yml
+++ b/.github/actions/install/action.yml
@@ -13,7 +13,7 @@ inputs:
   zig-v8:
     description: 'zig v8 version to install'
     required: false
-    default: 'v0.3.0'
+    default: 'v0.3.1'
   v8:
     description: 'v8 version to install'
     required: false
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM debian:stable-slim
 ARG MINISIG=0.12
 ARG ZIG_MINISIG=RWSGOq2NVecA2UPNdBUZykf1CCb147pkmdtYxgb3Ti+JO/wCYvhbAb/U
 ARG V8=14.0.365.4
-ARG ZIG_V8=v0.3.0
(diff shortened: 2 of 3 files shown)


src/browser/Session.zig
```diff
@@ -292,8 +300,14 @@ fn _wait(self: *Session, page: *Page, wait_ms: u32) !WaitResult {
                     // an cdp_socket registered with the http client).
                     // We should continue to run lowPriority tasks, so we
                     // minimize how long we'll poll for network I/O.
-                    const ms_to_wait = @min(200, @min(ms_remaining, ms_to_next_task orelse 200));
-                    if (try http_client.tick(ms_to_wait) == .cdp_socket) {
+                    var ms_to_wait = @min(200, ms_to_next_task orelse 200);
+                    if (ms_to_wait > 10 and browser.hasBackgroundTasks()) {
+                        // if we have bakcground tasks, we don't want ot wait too
```
krichprollsch:
```suggestion
                        // if we have background tasks, we don't want to wait too
```

PR #1653, opened 2026-02-25
add dummy PerformanceTiming

krichprollsch:


src/browser/webapi/Performance.zig
```diff
@@ -449,6 +456,69 @@ pub const Measure = struct {
...
+    pub const JsApi = struct {
+        pub const bridge = js.Bridge(PerformanceTiming);
+
+        pub const Meta = struct {
+            pub const name = "PerformanceTiming";
+            pub const prototype_chain = bridge.prototypeChain();
+            pub var class_id: bridge.ClassId = undefined;
+        };
+
+        pub const navigationStart = bridge.accessor(PerformanceTiming.getNavigationStart, null, .{});
```
karlseguin:
```suggestion
        pub const navigationStart = bridge.property(0.0, .{ .template = false, .readonly = true });
```

You can do this for all of them. It'll avoid having to create a function template for each one (keeping the snapshot smaller and the startup time fast) and be faster when called.

src/browser/webapi/Performance.zig
```diff
@@ -449,6 +456,69 @@ pub const Measure = struct {
...
+    pub fn getLoadEventStart(_: *const PerformanceTiming) f64 { return 0; }
+    pub fn getLoadEventEnd(_: *const PerformanceTiming) f64 { return 0; }
+
+    pub const JsApi = struct {
+        pub const bridge = js.Bridge(PerformanceTiming);
+
+        pub const Meta = struct {
+            pub const name = "PerformanceTiming";
+            pub const prototype_chain = bridge.prototypeChain();
+            pub var class_id: bridge.ClassId = undefined;
```
karlseguin, line 480: Add

```
pub const empty_with_no_proto = true;
```

Since there's no state, no state to preserve when v8 calls back into Zig, it'll avoid creating the TaggedOpaque for every instance of this type.

krichprollsch:
updated

PR #1658, opened 2026-02-25
Move curl C API to type-safe wrapper

mrdimidium:
I've already made several stupid mechanical errors in the `curl_easy_setopt` parameters and don't want to do it again.

This PR moves all direct libcurl calls into a small, type-safe wrapper. Constants are collected in enums, options are checked at compile time, and callbacks are not tied to callconv, numeric types are cast to c_long. The C api is private and not used externally.

The interface doesn't accept allocators and doesn't perform additional copies. Enums are cast directly with bitflags, with comptime checking that the memory layout matches.

### DIFF
--- a/src/Net.zig
+++ b/src/Net.zig
@@ -21,10 +21,7 @@ const builtin = @import("builtin");
 const posix = std.posix;
 const Allocator = std.mem.Allocator;
 const ArenaAllocator = std.heap.ArenaAllocator;
-
-pub const c = @cImport({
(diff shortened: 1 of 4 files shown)


karlseguin:
My only [very much bikeshedding] comment is the introduction of a "sys" folder with a single file in it. What makes something "sys"? I think I'd like a src/net/<everything_net_related_here>, including breaking out some of Net.zig, etc.  But that design / change can emerge as this code goes through more changes.

PR #1664, opened 2026-02-26
implement storage size limit per origin

krichprollsch:


src/browser/webapi/storage/storage.zig
```diff
@@ -69,6 +72,11 @@ pub const Lookup = struct {
     pub fn setItem(self: *Lookup, key_: ?[]const u8, value: []const u8, page: *Page) !void {
         const k = key_ orelse return;
 
+        if (self._size + value.len > max_size) {
```
karlseguin, line 75: you could calculate `self._size + value.len` just once

self._size should be set to 0 on clear. and self._size should be updated on removeItem.

krichprollsch: oups :sweat_smile:  fixed

PR #1667, opened 2026-02-27
On Client.stop, terminate the isolate

karlseguin:
Shutdown on MacOS doesn't work properly. The process appears to shutdown, but will continue to run in the background. It can run infinitely if it's stuck in a JS loop. To help it along, Client.stop now force-terminates the isolate.

I also don't think shutdown is working as intended on Linux, but the problem seems less serious there. On Linux, it appears to properly kill the process (which is the important thing), but I don't think it necessarily does a clean shutdown.

### DIFF
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -296,6 +296,10 @@ pub const Client = struct {
     }
 
     fn stop(self: *Client) void {
+        switch (self.mode) {
+            .http => {},
+            .cdp => |*cdp| cdp.browser.env.terminate(),
+        }
         self.ws.shutdown();
     }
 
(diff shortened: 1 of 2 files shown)


mrdimidium:
This seems logical, but the problem seems to lie somewhere deeper? Do we exit main and the background thread continues to hang, or are we blocked somewhere in main?

karlseguin:
On MacOS with this commit it behaves like I'd expect. Specifically, on MacOS, the program _seems_ to stop, but it continues to run. `client.stop()` just shuts down the socket, but if we're in a JS loop, it'll run _forever_ in the background. This prevents that.

On Linux, it shuts down the process whether or not it's in a JS loop, which I guess is good, but from what I can tell, teh sighandler never fires. Neither @krichprollsch nor I see "Received termination signal...". It seems to only work on Linux, because it's ignoring the sighandler. I think if the sighandler was working, it would behave like MacOS - waiting for JS to terminate.

PR #1673, opened 2026-02-27
Add Model Context Protocol (MCP) server support

arrufat:
Still work in progress, but it works. Bear with me while I get familiar with the codebase.

I've only tried with Gemini CLI.
You can add this to your current working directory `.gemini/settings.json`:
```json
{
  "mcpServers": {
    "lightpanda": {
      "command": "/home/adria/Projects/lightpanda/browser/zig-out/bin/lightpanda",
      "args": ["mcp"]
    }
  }
}
```

Then, start `gemini` and
- run: `/mcp list` to make sure `lightpanda` is available
- ask: _use lightpanda to get the markdown of archlinux.org_

You need to explicitly mention `lightpanda` because Gemini and other agents have a dedicated web tool, so they might try to fetch the URL by themselves, and pipe it through pandoc or similar to get the markdown. Explicitly mentioning lightpanda avoids that.

### DIFF
--- a/src/mcp/tools.zig
+++ b/src/mcp/tools.zig
@@ -24,19 +24,6 @@ pub const tool_list = [_]protocol.Tool{
         ),
     },
     .{
-        .name = "search",
-        .description = "Use a search engine to look for specific words, terms, sentences. The search page will then be loaded in memory.",
-        .inputSchema = protocol.minify(
-            \\{
-            \\  "type": "object",
-            \\  "properties": {
-            \\    "text": { "type": "string", "description": "The text to search for, must be a valid search query." }
-            \\  },
-            \\  "required": ["text"]
-            \\}
-        ),
-    },
-    .{
         .name = "markdown",
         .description = "Get the page content in markdown format. If a url is provided, it navigates to that url first.",
         .inputSchema = protocol.minify(
@@ -74,19 +61,6 @@ pub const tool_list = [_]protocol.Tool{
             \\}
         ),
     },
-    .{
-        .name = "over",
-        .description = "Used to indicate that the task is over and give the final answer if there is any. This is the last tool to be called in a task.",
-        .inputSchema = protocol.minify(
-            \\{
-            \\  "type": "object",
-            \\  "properties": {
-            \\    "result": { "type": "string", "description": "The final result of the task." }
-            \\  },
-            \\  "required": ["result"]
-            \\}
-        ),
-    },
 };
 
 pub fn handleList(server: *Server, arena: std.mem.Allocator, req: protocol.Request) !void {
@@ -98,17 +72,9 @@ const GotoParams = struct {
     url: [:0]const u8,
 };
 
-const SearchParams = struct {
-    text: [:0]const u8,
-};
-
 const EvaluateParams = struct {
     script: [:0]const u8,
     url: ?[:0]const u8 = null,
-};
-
-const OverParams = struct {
-    result: [:0]const u8,
 };
 
 const ToolStreamingText = struct {
@@ -155,21 +121,17 @@ const ToolStreamingText = struct {
 const ToolAction = enum {
     goto,
     navigate,
-    search,
     markdown,
     links,
     evaluate,
-    over,
 };
 
 const tool_map = std.StaticStringMap(ToolAction).initComptime(.{
     .{ "goto", .goto },
     .{ "navigate", .navigate },
-    .{ "search", .search },
     .{ "markdown", .markdown },
     .{ "links", .links },
     .{ "evaluate", .evaluate },
-    .{ "over", .over },
 });
 
 pub fn handleCall(server: *Server, arena: std.mem.Allocator, req: protocol.Request) !void {
@@ -192,11 +154,9 @@ pub fn handleCall(server: *Server, arena: std.mem.Allocator, req: protocol.Reque
 
     switch (action) {
         .goto, .navigate => try handleGoto(server, arena, req.id.?, call_params.arguments),
-        .search => try handleSearch(server, arena, req.id.?, call_params.arguments),
         .markdown => try handleMarkdown(server, arena, req.id.?, call_params.arguments),
         .links => try handleLinks(server, arena, req.id.?, call_params.arguments),
         .evaluate => try handleEvaluate(server, arena, req.id.?, call_params.arguments),
-        .over => try handleOver(server, arena, req.id.?, call_params.arguments),
     }
 }
 
@@ -205,24 +165,6 @@ fn handleGoto(server: *Server, arena: std.mem.Allocator, id: std.json.Value, arg
     try performGoto(server, args.url, id);
 
     const content = [_]protocol.TextContent([]const u8){.{ .text = "Navigated successfully." }};
-    try server.sendResult(id, protocol.CallToolResult([]const u8){ .content = &content });
-}
-
-fn handleSearch(server: *Server, arena: std.mem.Allocator, id: std.json.Value, arguments: ?std.json.Value) !void {
-    const args = try parseArguments(SearchParams, arena, arguments, server, id, "search");
-
-    const component: std.Uri.Component = .{ .raw = args.text };
-    var url_aw = std.Io.Writer.Allocating.init(arena);
-    component.formatQuery(&url_aw.writer) catch {
-        return server.sendError(id, .InternalError, "Internal error formatting query");
-    };
-    const url = std.fmt.allocPrintSentinel(arena, "https://duckduckgo.com/?q={s}", .{url_aw.written()}, 0) catch {
-        return server.sendError(id, .InternalError, "Internal error formatting URL");
-    };
-
-    try performGoto(server, url, id);
-
-    const content = [_]protocol.TextContent([]const u8){.{ .text = "Search performed successfully." }};
     try server.sendResult(id, protocol.CallToolResult([]const u8){ .content = &content });
 }
 
@@ -289,13 +231,6 @@ fn handleEvaluate(server: *Server, arena: std.mem.Allocator, id: std.json.Value,
     const str_result = js_result.toStringSliceWithAlloc(arena) catch "undefined";
 
     const content = [_]protocol.TextContent([]const u8){.{ .text = str_result }};
-    try server.sendResult(id, protocol.CallToolResult([]const u8){ .content = &content });
-}
-
-fn handleOver(server: *Server, arena: std.mem.Allocator, id: std.json.Value, arguments: ?std.json.Value) !void {
-    const args = try parseArguments(OverParams, arena, arguments, server, id, "over");
-
-    const content = [_]protocol.TextContent([]const u8){.{ .text = args.result }};
     try server.sendResult(id, protocol.CallToolResult([]const u8){ .content = &content });
 }
 


src/mcp/Server.zig
```diff
@@ -0,0 +1,140 @@
...
+        self.allocator.destroy(self.browser);
+        self.notification.deinit();
+        self.http_client.deinit();
+
+        self.allocator.destroy(self);
+    }
+
+    pub fn start(self: *Self) !void {
+        self.is_running.store(true, .seq_cst);
+        self.io_thread = try std.Thread.spawn(.{}, ioWorker, .{self});
```
karlseguin: Is there a reason to do it this way? From what I understand, we have 1 thread reading from stdin and queueing messages. And another thread (the main thread) processing these messages from the queue.  Why not just read and process on the main thread? I think this would simplify the code and better leverage the OS caching and back-pressure mechanism.

src/mcp/Server.zig
```diff
@@ -0,0 +1,140 @@
+const std = @import("std");
+
+const lp = @import("lightpanda");
+
+const App = @import("../App.zig");
+const HttpClient = @import("../http/Client.zig");
+
+pub const McpServer = struct {
```
karlseguin: Since the filename is PascalCase, this should be a file-as-a-struct.  So either the file should be renamed to `server.zig`, or the Server struct should be the file. I prefer the second one.

src/mcp/router.zig
```diff
@@ -0,0 +1,92 @@
...
+    const parsed = std.json.parseFromSliceLeaky(protocol.Request, arena, msg, .{
+        .ignore_unknown_fields = true,
+    }) catch |err| {
+        log.err(.app, "MCP JSON Parse Error", .{ .err = err, .msg = msg });
+        return;
+    };
+
+    if (parsed.id == null) {
+        // It's a notification
+        if (std.mem.eql(u8, parsed.method, "notifications/initialized")) {
```
karlseguin: We're sure these are always lowercase?

src/mcp/router.zig
```diff
@@ -0,0 +1,92 @@
...
+            };
+        }
+    }
+}
+
+fn handleMessage(server: *McpServer, arena: std.mem.Allocator, msg: []const u8) !void {
+    const parsed = std.json.parseFromSliceLeaky(protocol.Request, arena, msg, .{
+        .ignore_unknown_fields = true,
+    }) catch |err| {
+        log.err(.app, "MCP JSON Parse Error", .{ .err = err, .msg = msg });
```
karlseguin: Use the new `.mcp` log category?  I'm also weary of logging invalid user-input at `error` level, because there's nothing we can do about it and it trains us to dismiss `error`-level messages as things we can't fix. I'd make this a `warn`.

Although, maybe you're assuming it would be _extremely_ unlikely that an MCP user is sending invalid JSON and this more than likely represents an issue in the app..in which case, leave it as is.

src/mcp/router.zig
```diff
@@ -0,0 +1,92 @@
...
+const resources = @import("resources.zig");
+const tools = @import("tools.zig");
+
+pub fn processRequests(server: *McpServer) void {
+    while (server.is_running.load(.seq_cst)) {
+        if (server.getNextMessage()) |msg| {
+            defer server.allocator.free(msg);
+
+            // Critical: Per-request Arena
+            var arena = std.heap.ArenaAllocator.init(server.allocator);
```
karlseguin: Move this outside the while loop and reset/reuse it:

```
// I don't know what's a reasonable size of memory to retain between messages, 4K? 32K, 1MB?
defer _ = arena.reset(.{ .retain_with_limit = 4 * 1024 });
```

src/mcp/resources.zig
```diff
@@ -0,0 +1,101 @@
...
+    };
+
+    try sendResult(server, req.id.?, result);
+}
+
+const ReadParams = struct {
+    uri: []const u8,
+};
+
+pub fn handleRead(server: *McpServer, arena: std.mem.Allocator, req: protocol.Request) !void {
```
karlseguin: Hmmm..I was going to suggest that you could write directly to stdout, avoiding the triple buffering (the `aw` buffer, the Zig stdout buffer that you're using, the OS buffer)..but I see why you first have to buffer into `aw` - that stream is just a field in an outer envelope. I still think it's possible to flatten that all out, but I don't think it would be pretty. I only bring it up because it would be worth considering if you expect this to be on the hot path, but I suspect you don't. I guess this is largely used to load a handful of pages sequentially...not something that'll loop a million times over and over.

src/mcp/tools.zig
```diff
@@ -0,0 +1,301 @@
...
+            .description = "Navigate to a specified URL and load the page in memory so it can be reused later for info extraction.",
+            .inputSchema = std.json.parseFromSliceLeaky(std.json.Value, arena,
+                \\{
+                \\  "type": "object",
+                \\  "properties": {
+                \\    "url": { "type": "string", "description": "The URL to navigate to, must be a valid URL." }
+                \\  },
+                \\  "required": ["url"]
+                \\}
+            , .{}) catch unreachable,
```
karlseguin: Unlikely, but could OOM. Better to `try`.

src/mcp/tools.zig
```diff
@@ -0,0 +1,301 @@
...
+const log = lp.log;
+const js = lp.js;
+
+const Element = @import("../browser/webapi/Element.zig");
+const Selector = @import("../browser/webapi/selector/Selector.zig");
+const String = @import("../string.zig").String;
+const McpServer = @import("Server.zig").McpServer;
+const protocol = @import("protocol.zig");
+
+pub fn handleList(server: *McpServer, arena: std.mem.Allocator, req: protocol.Request) !void {
```
karlseguin: How often is this called? `tools` could be generated on server startup (I think it's static).

src/mcp/tools.zig
```diff
@@ -0,0 +1,301 @@
...
+        if (call_params.arguments) |args_raw| {
+            if (std.json.parseFromValueLeaky(LinksParams, arena, args_raw, .{})) |args| {
+                if (args.url) |u| {
+                    performGoto(server, arena, u) catch {
+                        return sendError(server, req.id.?, -32603, "Internal error during navigation");
+                    };
+                }
+            } else |_| {}
+        }
+        const list = Selector.querySelectorAll(server.page.document.asNode(), "a[href]", server.page) catch {
```
karlseguin: In a commit made yesterday, this has to be freed, `defer list.deinit(server.page)`.

src/mcp/tools.zig
```diff
@@ -0,0 +1,301 @@
...
+            return sendError(server, req.id.?, -32603, "Internal error querying selector");
+        };
+
+        var aw = std.Io.Writer.Allocating.init(arena);
+        var first = true;
+        for (list._nodes) |node| {
+            if (node.is(Element)) |el| {
+                if (el.getAttributeSafe(String.wrap("href"))) |href| {
```
karlseguin:
```suggestion
                if (el.getAttributeSafe(comptime .wrap("href"))) |href| {
```

The`String` can be inferred, which means you no longer have to import String atop. And, because the value is <= 12 bytes long, you can do it at comptime.

Note that the `href` attribute will be as it appears in the attribute, e.g. it could be relative.  If you want it resolved based on the page's URL, you can either call URL.resolve yourself, or (since this must be an Anchor element, given the query selector):

```zig
var anchor = node.as(Element.Html.Anchor);
const resolved_url = try anchor.getHref(server.page);
```

src/mcp/tools.zig
```diff
@@ -0,0 +1,301 @@
...
+        tools: []const protocol.Tool,
+    }{
+        .tools = &tools,
+    };
+
+    try sendResult(server, req.id.?, result);
+}
+
+const GotoParams = struct {
+    url: []const u8,
```
karlseguin:
```suggestion
    url: [:0]const u8,
```

I **think** if you do this, it'll deserialize it for you null terminated, and then you won't have to dupeZ it later. I could be wrong.

src/mcp/Server.zig
arrufat: Yes, me too. I pushed here, but it was not really ready for a review yet. I just wanted to give some visibility of my progress :)

src/mcp/Server.zig
arrufat: Thanks for the suggestion! This is indeed way simpler.

src/mcp/router.zig
arrufat: Yes, MCP follows the JSON-RPC 2.0 specification
https://www.jsonrpc.org/specification

> _All member names exchanged between the Client and the Server that are considered for matching of any kind should be considered to be case-sensitive._

src/mcp/resources.zig
arrufat: I've implemented a JSON streaming approach to address the triple buffering you mentioned. Memory usage should be constant, and maybe a good thing, since LLMs maybe be used to scrape very large and DOM-heavy sites. Have a look when you have time. It's the first time I implement an interface with a VTable in Zig...

src/mcp/tools.zig
arrufat: Thank you!

src/mcp/tools.zig
arrufat: Great point. I've refactored the MCP server to pre-initialize both tools and resources in Server.init instead of reconstructing them on every request.

src/mcp/tools.zig
arrufat: You're right! I've updated the tool parameter structs to use `[:0]const u8`: `std.json.parseFromValueLeaky` respects the sentinel, so we now get null-terminated strings automatically. I've also updated `performGoto` to take `[:0]const u8` directly, which allowed me to remove the redundant `dupeZ` calls.

src/mcp/Server.zig
```diff
@@ -0,0 +1,90 @@
...
+const HttpClient = @import("../http/Client.zig");
+const protocol = @import("protocol.zig");
+const Self = @This();
+
+allocator: std.mem.Allocator,
+app: *App,
+
+http_client: *HttpClient,
+notification: *lp.Notification,
+browser: *lp.Browser,
```
karlseguin: Minor, but this can own the browser:


```suggestion
browser: lp.Browser,
```

src/mcp/tools.zig
```diff
@@ -0,0 +1,323 @@
...
+        try jw.beginWriteRaw();
+        try jw.writer.writeByte('"');
+        var escaped = protocol.JsonEscapingWriter.init(jw.writer);
+        const w = &escaped.writer;
+        switch (self.action) {
+            .markdown => lp.markdown.dump(self.server.page.document.asNode(), .{}, w, self.server.page) catch |err| {
+                log.err(.mcp, "markdown dump failed", .{ .err = err });
+            },
+            .links => {
+                if (Selector.querySelectorAll(self.server.page.document.asNode(), "a[href]", self.server.page)) |list| {
```
karlseguin, line 94: against main, `list` has to be freed (deinit)

arrufat:
Actually, I am getting an error now, where the server runs well if I pipe stuff into it, but it gives an error in the Gemini CLI. Try to figure out what I broke. I guess yesterday I forgot to restart the server while refactoring stuff, and broke it without noticing.

src/mcp/tools.zig
arrufat: Done!

src/mcp/router.zig
```diff
@@ -0,0 +1,145 @@
...
+            .version = "0.1.0",
+        },
+    };
+
+    try server.sendResult(req.id.?, result);
+}
+
+const testing = @import("../testing.zig");
+
+test "handleMessage - synchronous unit tests" {
```
karlseguin: You might want to give your tests somewhat namespace'd names. 

test "MCP.router - handleMessage - synchronous unit tests"   or something. It'll let you filter them better using `TEST_FILTER`. Right now, there's no easy way to run all MCP tests.

`TEST_FITLER="MCP" zig build test"`  or `TEST_FITLER="MCP.router" zig build test"`

(or, if you prefer a few less keystrokes)

`make test F="MCP.router"`

src/mcp/router.zig
```diff
@@ -0,0 +1,145 @@
...
+    const allocator = testing.allocator;
+    const app = testing.test_app;
+
+    var out_alloc = std.io.Writer.Allocating.init(allocator);
```
karlseguin: If you have an need for test-specific allocations that you don't care about tracking, you can do something like:

```
defer testing.reset(); // clears the testing arena
std.io.Writer.Allocating.init(testing.arena_allocator);
```

Allocations made against this allocator won't report any leaks, it's meant for test data. So you [probably] don't want to use it in Server.init (since you _would_ care to catch leaks), but from what I understand of test, could use it for `out_alloc` and in your `handleMessage`

src/mcp/router.zig
```diff
@@ -0,0 +1,145 @@
...
+
+    var arena = std.heap.ArenaAllocator.init(allocator);
+    defer arena.deinit();
+    const aa = arena.allocator();
+
+    // 1. Valid handshake
+    try handleMessage(server, aa,
+        \\{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"test-client","version":"1.0.0"}}}
+    );
+    try testing.expect(std.mem.indexOf(u8, out_alloc.writer.buffered(), "\"id\":1") != null);
```
karlseguin: you might prefer expectJson for all of these `indexOf` checks.

```
try testing.expectJson(.{.id = 1}, out_alloc.writer.buffered());
```

The first parameter, `.{.id = 1}`, will "drive" the comparison. It's called expectJson because it tried to convert the parameters to JSON. If you give it a string, it assumes that's a JSON string and parses it.

src/mcp/router.zig
```diff
@@ -0,0 +1,145 @@
...
+        \\{"jsonrpc":"2.0","id":3,"method":"unknown_method"}
+    );
+    try testing.expect(std.mem.indexOf(u8, out_alloc.writer.buffered(), "\"id\":3") != null);
+    try testing.expect(std.mem.indexOf(u8, out_alloc.writer.buffered(), "\"code\":-32601") != null);
+    out_alloc.writer.end = 0;
+
+    // 4. Parse error
+    {
+        const old_filter = log.opts.filter_scopes;
+        log.opts.filter_scopes = &.{.mcp};
```
karlseguin, line 137: nice

src/mcp/Server.zig
```diff
@@ -0,0 +1,105 @@
...
+pub fn deinit(self: *Self) void {
+    self.browser.deinit();
+    self.notification.deinit();
+    self.http_client.deinit();
+
+    self.allocator.destroy(self);
+}
+
+pub fn sendResponse(self: *Self, response: anytype) !void {
+    var aw: std.io.Writer.Allocating = .init(self.allocator);
```
karlseguin: This can't be called concurrently, can it? I'd be tempted to find a way to re-use the memory. It would mean storing the `std.io.Writer` a field in `Server` and then calling `clearRetainingCapacity` on it after the flush.

src/mcp/tools.zig
```diff
@@ -0,0 +1,328 @@
...
+
+    if (args.url) |url| {
+        try performGoto(server, url, id);
+    }
+
+    var ls: js.Local.Scope = undefined;
+    server.page.js.localScope(&ls);
+    defer ls.deinit();
+
+    const js_result = ls.local.compileAndRun(args.script, null) catch {
```
karlseguin: If you wrap this in a TryCatch, can you get more meaningful error? I really don't know.

```zig
var try_catch: js.TryCatch = undefined;
try_catch.init(&ls.local);
defer try_catch.deinit();

const js_result = ls.local.compileAndRun(args.script, null) catch |err| {
  const caught = try_catch.caughtOrError(arena, err);
  // caught has a "format" method that you can look at to figure ou thow to serialize
  // we could add a "pub fn jsonStringify" so that it can be serialized directly to JSON
}
```

src/mcp/router.zig
arrufat: Nice!

src/mcp/router.zig
arrufat: Cool!

src/mcp/tools.zig
arrufat: Yes, I think it can be improved :)

karlseguin (approved):

src/mcp/tools.zig
```diff
@@ -0,0 +1,370 @@
...
+            \\  "properties": {
+            \\    "script": { "type": "string" },
+            \\    "url": { "type": "string", "description": "Optional URL to navigate to before evaluating." }
+            \\  },
+            \\  "required": ["script"]
+            \\}
+        ),
+    },
+    .{
+        .name = "over",
```
krichprollsch: Please can you remove this tool? It was added in gomcp for a specific case which is I think obsolete.

src/mcp/tools.zig
```diff
@@ -0,0 +1,370 @@
...
+            \\  "type": "object",
+            \\  "properties": {
+            \\    "url": { "type": "string", "description": "The URL to navigate to, must be a valid URL." }
+            \\  },
+            \\  "required": ["url"]
+            \\}
+        ),
+    },
+    .{
+        .name = "search",
```
krichprollsch: I know we added that in gomcp, but I'm not sure we should include it with duckduckgo directly.
LLMs can go to duckduckgo directly.
I would remove it too.

src/mcp/tools.zig
arrufat: Yes, it was bothering me when using the agent, it wanted to use it at the end of each task.

PR #1693, opened 2026-03-02
`ArenaPool`: make init configurable + add tests

nikneym:


karlseguin:
As we move more to them, making them user-configurable via arguments would probably make sense.

There are still other things that can be moved to these. Pretty much every other ArenaAllocator.init should be scrutinized.

One thing that's been briefly discussed is having different tiers, small, medium, large. For example, you could have 32 "large" pools that retain 256K. And we could use these where we expect large memory usage (e.g. page arena). As a single config, 32 large pools, and 256 small pools could both result in fewer allocations and less memory than a compromising 288 "medium" pools. But it isn't 100% clear if something fancier than "large" and "small" is needed / beneficial.  Too many categories, and you kind of defeat the purpose.

Also, we should probably consider moving the ScriptManager BufferPool to these.  Right now that Buffer pool has the advantage of being specialized, but  it's per-page with no re-use between pages.

PR #1696, opened 2026-03-02
Add TextEncoderStream and TextDecoderStream implementation

krichprollsch:
This PR has been created with heavy assistance of Claude Code.
I have to review it more deeply myself. It can have issues/inconsistencies.

Reviewing commit per commit could be simpler.

### DIFF
--- a/src/browser/webapi/streams/TransformStream.zig
+++ b/src/browser/webapi/streams/TransformStream.zig
@@ -99,11 +99,7 @@ pub fn transformWrite(self: *TransformStream, chunk: js.Value, page: *Page) !voi
 
         try ls.toLocal(transform_fn).call(void, .{ chunk, self._controller });
     } else {
-        // Default transform: pass through
-        if (chunk.isString()) |str| {
-            const slice = try str.toSlice();
-            try self._readable._controller.enqueue(.{ .string = slice });
-        }
+        try self._readable._controller.enqueue(.{ .string = try chunk.toStringSlice() });
     }
 }
 


src/browser/webapi/streams/ReadableStream.zig
```diff
@@ -233,6 +235,190 @@ pub fn cancel(self: *ReadableStream, reason: ?[]const u8, page: *Page) !js.Promi
...
+    }
+
+    fn finish(state: *PipeState, local: *const js.Local) void {
+        state.reader.releaseLock();
+        if (state.resolver) |r| {
+            local.toLocal(r).resolve("pipe finished", {});
+        }
+    }
+
+    fn newFunctionWithData(
```
krichprollsch: This is a copy of `Context.newFunctionWithData` which is private, and used only for `resolveDynamicModule`.

I guess we should move these calls to v8 somewhere into `js/`

src/browser/webapi/encoding/TextDecoderStream.zig
```diff
@@ -0,0 +1,110 @@
...
+    const opts = opts_ orelse InitOpts{};
+    const transform = try TransformStream.initWithZigTransform(&decodeTransform, page);
+    return .{
+        ._transform = transform,
+        ._fatal = opts.fatal,
+        ._ignore_bom = opts.ignoreBOM,
+    };
+}
+
+fn decodeTransform(controller: *TransformStream.DefaultController, chunk: js.Value) !void {
```
karlseguin: `chunk` can be directly declared as a `js.TypedArray(u8)`. I think there might be a few cases where we take the js.Value and `toZig` it to a single type, but that's usually because we want to keep the js.Value as-is also. But that isn't the case here.

src/browser/webapi/streams/TransformStream.zig
```diff
@@ -0,0 +1,202 @@
...
+
+const ReadableStream = @import("ReadableStream.zig");
+const ReadableStreamDefaultController = @import("ReadableStreamDefaultController.zig");
+const WritableStream = @import("WritableStream.zig");
+
+const TransformStream = @This();
+
+pub const DefaultController = TransformStreamDefaultController;
+
+pub const ZigTransformFn = *const fn (*TransformStreamDefaultController, js.Value) anyerror!void;
```
karlseguin, line 30:
```suggestion
const ZigTransformFn = *const fn (*TransformStreamDefaultController, js.Value) anyerror!void;
```

src/browser/webapi/encoding/TextDecoderStream.zig
```diff
@@ -0,0 +1,110 @@
...
+    // chunk should be a Uint8Array; decode it as UTF-8 string
+    const typed_array = try chunk.toZig(js.TypedArray(u8));
+    var input = typed_array.values;
+
+    // Strip UTF-8 BOM if present
+    if (std.mem.startsWith(u8, input, &.{ 0xEF, 0xBB, 0xBF })) {
```
karlseguin: Based on TextDecoder, should this only happen if `_ignore_bom == false`?

src/browser/webapi/streams/TransformStream.zig
```diff
@@ -0,0 +1,202 @@
...
+const Transformer = struct {
+    start: ?js.Function = null,
+    transform: ?js.Function.Global = null,
+    flush: ?js.Function.Global = null,
+};
+
+pub fn init(transformer_: ?Transformer, page: *Page) !*TransformStream {
+    const readable = try ReadableStream.init(null, null, page);
+
+    const self = try page._factory.create(TransformStream{
```
karlseguin, line 45:
```zig
const transformer = transformer_ orelse .{};
```

will clean up the code under it.  

```zig
    const transform_controller = try TransformStreamDefaultController.init(
        self,
        transformer.transform else null,
        transformer.flush
        null,
        page,
    );
```

and

```
    if (transformer_) |transformer| { // can remove this outer check
        if (transformer.start) |start| {
            try start.call(void, .{transform_controller});
        }
    }
```

src/browser/webapi/streams/TransformStream.zig
```diff
@@ -0,0 +1,202 @@
...
+
+    if (self._controller._transform_fn) |transform_fn| {
+        var ls: js.Local.Scope = undefined;
+        page.js.localScope(&ls);
+        defer ls.deinit();
+
+        try ls.toLocal(transform_fn).call(void, .{ chunk, self._controller });
+    } else {
+        // Default transform: pass through
+        if (chunk.isString()) |str| {
```
karlseguin: I think you have this `isString()` check somewhere else. Are you sure? I feel like it might enforce a stricter type checking than is correct. I don't really understand this feature, but generally, JavaScript code doesn't care if something is an actual string, it just wants a string representation of it. e.g., should this work if the chunk is the number `9001` ? If so, should go directly to a string via `js_val.toString()`

karlseguin (commented):
I made a larger change in:
https://github.com/lightpanda-io/browser/pull/1699

src/browser/webapi/encoding/TextDecoderStream.zig
krichprollsch: Shouldn't be apply only if `_ignore_bom == true`?

krichprollsch: > true if the [byte order mark](https://www.w3.org/International/questions/qa-byte-order-mark) will be included in the decoded text; false if it will be skipped over when decoding and omitted.
https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder/ignoreBOM

:ok_hand:

src/browser/webapi/streams/TransformStream.zig
krichprollsch: I'm not sure :sweat: I changed to the stringSlice.
I'm wondering if I should create a `js_string` in the `Chunk` enum to avoid extra allocation on dupe :thinking:

karlseguin: Could this function just take a `Chunk` directly?  If not, I think it's probably fine as-is.

krichprollsch: ok, let's keep it as is.

PR #1698, opened 2026-03-02
use a pool arena with ReadableStream

krichprollsch:


src/browser/webapi/streams/ReadableStream.zig
```diff
@@ -67,9 +69,13 @@ const QueueingStrategy = struct {
 pub fn init(src_: ?UnderlyingSource, strategy_: ?QueueingStrategy, page: *Page) !*ReadableStream {
     const strategy: QueueingStrategy = strategy_ orelse .{};
 
+    const arena = try page.getArena(.{ .debug = "Animation" });
```
karlseguin: Could use the arena to create the ReadableStream insteance.

karlseguin (commented):
I don't think this will work. As soon as the type becomes a field in another struct, it becomes tricky. ReadableStream is referenced by ~3 other types.

```js
let it = null;
{
    let rs = new ReadableStream(....);
    it = rs[Symbol.asyncIterator]();
}
console.log(it.next());
```

`rs` can be released by v8 when the scope exits, but then our iterator would still be referencing it.

To make this work:

1 - add a `_rc: usize = 0` field to ReadableStream
2 - Add an `acquireRef` function to ReadableStream. This will be automatically called by the bridge
```zig
pub fn acquireRef(self: *ReadableStream) void {
    self._rc += 1;
}
```
3 - Change `deinit` to decrement _rc and only release it when _rc == 0
4 - Add a similar `acquireRef` to every type that references a RS: 
```zig
    pub fn acquireRef(self: *AsyncIterator) void {
        self._stream.acquireRef();
    }
```
5 - Add a similar deinit / finalizer to every type that referneces a RS:
```zig
    pub fn deinit(self: *AsyncIterator, shutdown: bool, page: *Page) void {
        self._stream.deinit(shutdown, page);
    }
```

Like I said above `acquireRef` is automatically called by the bridge, but if it's a reference we hold internally, then we need to manually call `acquireRef`. I don't think that's ever the case of ReadableStream though.

There is potentially a simpler way to do all of this by using v8's internal fields. We can store the v8::Object that represents the ReadableStream inside an internal field of the v8::Object that represents the AsycIteratr, and then v8 will know that the reference is still alive. BUT: (a) I haven't found a good way to express that in JsApi.Meta (because the ownership can get complicated) and (b) it doens't work great for things like Events where we want both Zig and V8 to have a reference.

krichprollsch:
Ok I made a try with reference counting, but I'm not sure it's correct and solid :/
Maybe it isn't worth the change...

karlseguin:
I'm ok with the change, but I agree the whole thing isn't great / feels risky. It's why I've only done it on high value targets, namely Events and NodeList  which are both frequently used and rather large.

PR #1717, opened 2026-03-05
Improve markdown links

arrufat:
Since the article is coming soon, let's polish this a bit more.

- The Markdown link text (`[text]`) is determined using the following priority:

1.  **Visible Content**: Uses inner text or nested `` tags if present.
2.  **Semantic Fallback**: If no content is found (e.g., icon-only links), it falls back to the `aria-label` or `title` attributes.
3.  **Empty Placeholder**: If both are missing but an `href` exists, it renders as `[]`. This preserves the link for LLMs to follow without introducing "hallucinated" filler text like `[Link]`, while saving on tokens.

- Additionally, all links are now resolved to **absolute URLs** using the page's base URI.

My intuition tells me that the proposed approach uses more input tokens, which are the cheapest ones, but less thinking/output tokens, which are more expensive. So we make the task more straightforward for the LLM.

```console
lightpanda fetch --dump markdown https://lightpanda.io/
```

```diff
< [](/)
< - [Cloud offer](/#cloud-offer)
< - [Docs](/docs)
< - [Blog](/blog)
< - [Jobs](/jobs)
---
> [](https://lightpanda.io/)
> - [Cloud offer](https://lightpanda.io/#cloud-offer)
> - [Docs](https://lightpanda.io/docs)
> - [Blog](https://lightpanda.io/blog)
> - [Jobs](https://lightpanda.io/jobs)
```

### DIFF
--- a/src/browser/markdown.zig
+++ b/src/browser/markdown.zig
@@ -298,7 +298,7 @@ fn renderElement(el: *Element, state: *State, writer: *std.Io.Writer, page: *Pag
             }
             try writer.writeAll("](");
             if (el.getAttributeSafe(comptime .wrap("src"))) |src| {
-                const absolute_src = URL.resolve(page.call_arena, page.base(), src, .{}) catch src;
+                const absolute_src = URL.resolve(page.call_arena, page.base(), src, .{ .encode = true }) catch src;
                 try writer.writeAll(absolute_src);
             }
             try writer.writeAll(")");
@@ -313,7 +313,7 @@ fn renderElement(el: *Element, state: *State, writer: *std.Io.Writer, page: *Pag
             if (!has_content and label == null and href_raw == null) return;
 
             const has_block = hasBlockDescendant(el.asNode());
-            const href = if (href_raw) |h| URL.resolve(page.call_arena, page.base(), h, .{}) catch h else null;
+            const href = if (href_raw) |h| URL.resolve(page.call_arena, page.base(), h, .{ .encode = true }) catch h else null;
 
             if (has_block) {
                 try renderChildren(el.asNode(), state, writer, page);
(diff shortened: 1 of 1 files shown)


src/browser/markdown.zig
```diff
@@ -119,6 +124,21 @@ fn hasBlockDescendant(node: *Node) bool {
     } else false;
 }
 
+fn hasVisibleContent(node: *Node) bool {
+    var it = node.childrenIterator();
```
karlseguin, line 128: I think you should keep this code as-is, but there is a TreeWalker that could be used here.  It  would let you remove 1 line of code - the recursive check.

But the reason I think you should keep this code as-is, is that TreeWalker walks in document order (something you don't care about here) and because of that it's going to be less efficient than what you're doing here.

Just an FYI.

src/browser/markdown.zig
```diff
@@ -278,20 +298,29 @@ fn renderElement(el: *Element, state: *State, writer: *std.Io.Writer, page: *Pag
             }
             try writer.writeAll("](");
             if (el.getAttributeSafe(comptime .wrap("src"))) |src| {
-                try writer.writeAll(src);
+                const absolute_src = URL.resolve(page.call_arena, page.base(), src, .{}) catch src;
```
karlseguin:
```suggestion
                const absolute_src = URL.resolve(page.call_arena, page.base(), src, .{.encode = true}) catch src;
```
It's a new option, which is why it isn't on by default, but I imagine you do want it here if the goal is to generate valid URLs.

src/browser/markdown.zig
```diff
@@ -278,20 +298,29 @@ fn renderElement(el: *Element, state: *State, writer: *std.Io.Writer, page: *Pag
...
         },
         .anchor => {
+            const has_content = hasVisibleContent(el.asNode());
+            const label = getAnchorLabel(el);
+            const href_raw = el.getAttributeSafe(comptime .wrap("href"));
+
+            if (!has_content and label == null and href_raw == null) return;
+
             const has_block = hasBlockDescendant(el.asNode());
+            const href = if (href_raw) |h| URL.resolve(page.call_arena, page.base(), h, .{}) catch h else null;
```
karlseguin: same as above, .encode = true

src/browser/markdown.zig
arrufat: I love this kind of feedback, thank you!

PR #1718, opened 2026-03-05
Enhance TreeWalker

arrufat:
Follow-up of https://github.com/lightpanda-io/browser/pull/1717#discussion_r2887341944

### DIFF
--- a/src/browser/markdown.zig
+++ b/src/browser/markdown.zig
@@ -20,6 +20,7 @@ const std = @import("std");
 
 const Page = @import("Page.zig");
 const URL = @import("URL.zig");
+const TreeWalker = @import("webapi/TreeWalker.zig");
 const CData = @import("webapi/CData.zig");
 const Element = @import("webapi/Element.zig");
 const Node = @import("webapi/Node.zig");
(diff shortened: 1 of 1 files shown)


src/browser/webapi/TreeWalker.zig
```diff
@@ -47,37 +48,74 @@ pub fn TreeWalker(comptime mode: Mode) type {
...
+            return node;
+        }
+
+        pub fn skipChildren(self: *Self) void {
+            if (comptime mode == .children) return;
+            const current = self._current orelse return;
+            self._next = self.computeNextSiblingOrUncle(current);
+        }
+
+        pub fn nextSibling(self: *Self) ?*Node {
```
karlseguin, line 68: Is there a reason to add this, if it isn't being used?

src/browser/webapi/TreeWalker.zig
```diff
@@ -47,37 +48,74 @@ pub fn TreeWalker(comptime mode: Mode) type {
...
-                    }
-                    current = parent._parent;
-                } else {
-                    self._next = null;
+                self._next = self.computeNextInDocumentOrder(sibling);
+            }
+            return sibling;
+        }
+
+        pub fn previousSibling(self: *Self) ?*Node {
```
karlseguin, line 81: Is there a reason to add this, if it isn't being used?

src/browser/webapi/TreeWalker.zig
arrufat: I just thought it would be nice to have them to complete the TreeWalker API.
YAGNI, we can add them if we need them later to keep the API leaner.

PR #1719, opened 2026-03-05
DRAFT: common libcurl for all cdp conns.

mrdimidium:
Implements a common runtime for libcurl and the accept loop (a common ws will come later).

A notable issue I encountered and haven't fully resolved yet: network callbacks must be executed on the worker thread because they touch the V8 API, but moving header parsing out of the callback isn't possible due to cookies being set during redirects.

There's also a rare test failure (`exception: testing is not defined`), which is likely due to incorrect processing order.

### DIFF
--- a/src/browser/HttpClient.zig
+++ b/src/browser/HttpClient.zig
@@ -65,14 +65,16 @@ active: usize,
 // 'networkAlmostIdle' Page.lifecycleEvent in CDP).
 intercepted: usize,
 
-// Our easy handles, managed by a curl multi.
-handles: Net.Handles,
+worker_id: Network.WorkerId,
 
 // Use to generate the next request ID
 next_request_id: u32 = 0,
 
-// When handles has no more available easys, requests get queued.
+// When no connections are available, requests get queued.
 queue: TransferQueue,
+
+// Transfers that have been submitted to Runtime (in-flight).
+active_transfers: TransferQueue = .{},
 
 // The main app allocator
 allocator: Allocator,
@@ -125,23 +127,17 @@ pub fn init(allocator: Allocator, network: *Network) !*Client {
     const client = try allocator.create(Client);
     errdefer allocator.destroy(client);
 
-    var handles = try Net.Handles.init(allocator, network.ca_blob, network.config);
-    errdefer handles.deinit(allocator);
-
-    // Set transfer callbacks on each connection.
-    for (handles.connections) |*conn| {
-        try conn.setCallbacks(Transfer.headerCallback, Transfer.dataCallback);
-    }
-
     const http_proxy = network.config.httpProxy();
+
+    const worker_id = try network.registerWorker();
 
     client.* = .{
         .queue = .{},
         .active = 0,
         .intercepted = 0,
-        .handles = handles,
         .allocator = allocator,
         .network = network,
+        .worker_id = worker_id,
         .http_proxy = http_proxy,
         .use_proxy = http_proxy != null,
         .transfer_pool = transfer_pool,
@@ -152,7 +148,7 @@ pub fn init(allocator: Allocator, network: *Network) !*Client {
 
 pub fn deinit(self: *Client) void {
     self.abort();
-    self.handles.deinit(self.allocator);
+    self.network.unregisterWorker(self.worker_id);
 
     self.transfer_pool.deinit();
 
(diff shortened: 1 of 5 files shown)


src/network/Runtime.zig
```diff
@@ -0,0 +1,491 @@
...
+
+const net_http = @import("http.zig");
+const RobotStore = @import("Robots.zig").RobotStore;
+
+const Runtime = @This();
+
+const Listener = struct {
+    socket: posix.socket_t,
+    ctx: *anyopaque,
+    onAccept: *const fn (ctx: *anyopaque, socket: posix.socket_t) void,
```
karlseguin, line 37:
```suggestion
    on_accept: *const fn (ctx: *anyopaque, socket: posix.socket_t) void,
```

src/network/Runtime.zig
```diff
@@ -0,0 +1,491 @@
...
+    try posix.setsockopt(listener, posix.SOL.SOCKET, posix.SO.REUSEADDR, &std.mem.toBytes(@as(c_int, 1)));
+    if (@hasDecl(posix.TCP, "NODELAY")) {
+        try posix.setsockopt(listener, posix.IPPROTO.TCP, posix.TCP.NODELAY, &std.mem.toBytes(@as(c_int, 1)));
+    }
+
+    try posix.bind(listener, &address.any, address.getOsSockLen());
+    try posix.listen(listener, self.config.maxPendingConnections());
+
+    const listener_fds = self.pollfds[1..1 + Config.MAX_LISTENERS];
+    for (&self.listeners, listener_fds) |*slot, *pfd| {
```
karlseguin, line 192: This seems complicated. If I understand correctly, why not treat "accept" like you do the wake-up pipe? pollfds[0] = wakeup, pollfds[1] == accept.

....

Actually, I just got to the `run` function and I don't understand what listeners are. Can you explain it?

src/network/Runtime.zig
```diff
@@ -0,0 +1,491 @@
...
+    _ = self.workers.remove(id);
+}
+
+/// Thread-safe: called by workers to submit a transfer.
+pub fn submit(self: *Runtime, node: *std.DoublyLinkedList.Node) void {
+    self.mutex.lock();
+    self.submit_queue.append(node);
+    self.mutex.unlock();
+
+    _ = posix.write(self.wakeup_pipe[1], &.{1}) catch {};
```
karlseguin, line 309: There are a lot of `catch {}`...they should all at least include a log, but cases like this seem particularly problematic. If writing to the pipe fail, things are kind of in a bad state, no?

src/browser/HttpClient.zig
```diff
@@ -989,6 +992,13 @@ pub const Transfer = struct {
 
     // for when a Transfer is queued in the client.queue
     _node: std.DoublyLinkedList.Node = .{},
+    // for submitting to Runtime's submit/completion queues
+    _request: Network.Request = .{ .worker_id = 0, .conn = undefined },
```
karlseguin, line 996: Transfer has a `req` and a `_request`. A bit confusing.

src/network/Runtime.zig
```diff
@@ -0,0 +1,491 @@
...
+            lp.log.err(.http, "curl_multi_add_handle", .{ .err = err });
+            request.err = err;
+            self.notifyWorker(request);
+        };
+    }
+}
+
+fn notifyWorker(self: *Runtime, request: *Request) void {
+    self.mutex.lock();
+    const slot = self.workers.getPtr(request.worker_id);
```
karlseguin, line 335: Http.Client gets a `worker_id` on init, which it then passes to every Transfer. The point is to be able to lookup the slot in the workers lookup.

Can you allocate the slot on the heap, and make the worker_id the address? You avoid the lock and lookup:

```zig
const slot: *Slot = @ptrFromInt(request.worker_id)
```

src/network/Runtime.zig
```diff
@@ -0,0 +1,491 @@
...
+    self.mutex.unlock();
+
+    if (slot) |s| {
```
karlseguin, line 338: Can this ever not be true?

karlseguin (commented):
Initial comments from just looking at a few files. I'll try to go through it in more detail this weekend.

src/browser/HttpClient.zig
```diff
@@ -227,40 +213,49 @@ fn _abort(self: *Client, comptime abort_all: bool, frame_id: u32) void {
...
+        const conn = self.network.getConnection() orelse {
+            self.queue.prepend(&transfer._node);
+            break;
+        };
         try self.makeRequest(conn, transfer);
     }
-    return self.perform(@intCast(timeout_ms));
+
+    // Wait for completions from the shared Runtime event loop.
+    const slot = self.getWorkerSlot() orelse return .normal;
```
karlseguin, line 230: For a given instance of an Http.Client, does the slot instance change?

src/browser/HttpClient.zig
```diff
@@ -585,6 +576,12 @@ pub fn fulfillTransfer(self: *Client, transfer: *Transfer, status: u16, headers:
     transfer._intercept_state = .fulfilled;
 }
 
+fn getWorkerSlot(self: *Client) ?*Network.WorkerSlot {
+    self.network.mutex.lock();
```
karlseguin, line 580: Encapsulate in runtime? Http.Client shouldn't know about network mutexes.

src/network/Runtime.zig
mrdimidium: Listeners are a thin wrapper around bind/listen. Since the runtime must be present on everyone, it by default assumes there can be more than one listener (for example, for mcp).

karlseguin: Unless this is a feature that's been talked about, I'm not sure it's worth supporting multple listeners considering how much complexity it adds.

PR #1720, opened 2026-03-05
Improve frame sub-navigation

karlseguin:
This makes frame sub-navigation "work" for all page navigations (click, form submit, location.top...) as well as setting the iframe.src.

Fixes at least 2 WPT crashes.

BUT, the implementation still isn't 100% correct, with two known issues:

1. Navigation currently happens in the context where it's called, not the context of the frame. So if Page1 accesses Frame1 and causes it to navigate, e.g. f1.contentDocument.querySelector('#link').click(), it's Page1 that will be navigated, since the JS is being executed in the Page1 context. This should be relatively easy to fix.

2. There are particularly complicated cases in WPT where a frame is navigated inside of its own load, creating an endless loop. There's some partial support for this as-is, but it doesn't work correctly and it currently is defensive and likely will not continue to navigate. This is particularly true when sub-navigation is done to about:blank within the frame's on load event. (Which is probably not a real concern, but an issue for some WPT tests)

Although it shares a lot with the original navigation code, there are many more
edge cases here, possibly due to being developed along side WPT tests. The
source of most of the complexity is the synchronous handling of "about:blank"
in page.navigate, which can result in a scheduled navigation synchronously
causing more scheduled navigation. (Specifically because
`self.documentIsComplete();` is called from page.navigate in that case). It
might be worth seeing if something can be done about that, to simplify this new
code (removing the double queue, removing the flag, simplifying pre-existing
schedule checks ,...)

### DIFF
--- a/src/browser/Session.zig
+++ b/src/browser/Session.zig
@@ -44,13 +44,11 @@ const Session = @This();
 browser: *Browser,
 notification: *Notification,
 
-queued_navigation: std.ArrayList(*QueuedNavigation),
-// It's possible (but unlikely) that a queued navigation happens when we're
-// processessing queued navigations (thank you WPT). This causes a lot of issues
-// including possibly invalidating `queued_navigation` and endless loops.
-// We use a double queue to avoid this.
-processing_queued_navigation: bool,
-queued_queued_navigation: std.ArrayList(*QueuedNavigation),
+queued_navigation: std.ArrayList(*Page),
+// Temporary buffer for about:blank navigations during processing.
+// We process async navigations first (safe from re-entrance), then sync
+// about:blank navigations (which may add to queued_navigation).
+queued_queued_navigation: std.ArrayList(*Page),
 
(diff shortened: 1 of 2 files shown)


src/browser/Session.zig
```diff
@@ -339,42 +349,147 @@ fn _wait(self: *Session, page: *Page, wait_ms: u32) !WaitResult {
...
-        browser.http_client.abort();
-        self.removePage();
+    if (current_page._parent_notified) {
+        // we already notified the parent that we had loaded
+        parent._pending_loads += 1;
+    }
 
-        break :blk .{ frame_id, parent };
+    const frame_id = current_page._frame_id;
+    defer current_page.deinit(true);
```
karlseguin, line 440: This is incorrect, as it will abort all frame requests _after_ we start to navigate the new page with that same frame id. It's fixed in https://github.com/lightpanda-io/browser/pull/1729

PR #1725, opened 2026-03-05
initialize all App fields after allocator.create

egrs:
## Summary

Same pattern as 3dea554e (mcp/Server.zig fix): `allocator.create(App)` returns undefined memory, and the struct field default `shutdown: bool = false` is not applied when fields are set individually — only struct literal syntax applies defaults.

Uses `app.* = .{...}` to ensure all fields are initialized, including `shutdown` which is read atomically in `deinit()`.

## Test plan

- [x] `make test` — 316/316 pass on aarch64 Docker Desktop
- [x] CI passes on x86_64

### DIFF
--- a/src/App.zig
+++ b/src/App.zig
@@ -47,10 +47,17 @@ pub fn init(allocator: Allocator, config: *const Config) !*App {
     const app = try allocator.create(App);
     errdefer allocator.destroy(app);
 
-    app.config = config;
-    app.allocator = allocator;
-
-    app.robots = RobotStore.init(allocator);
+    app.* = .{
+        .config = config,
+        .allocator = allocator,
+        .robots = RobotStore.init(allocator),
+        .http = undefined,
+        .platform = undefined,
+        .snapshot = undefined,
+        .app_dir_path = undefined,
+        .telemetry = undefined,
+        .arena_pool = undefined,
+    };
 
     app.http = try Http.init(allocator, &app.robots, config);
     errdefer app.http.deinit();


karlseguin:
Works on MacOS, seems ok on the CI. @egrs ok to merge this?

egrs:
I'm still not sure about this one, having trouble reproducing the hang reliably. I'll continue investigating for now.

egrs:
Thanks! After more investigation, I should clarify the scope of this change.

I did an A/B test on aarch64 Docker Desktop (Apple Silicon) — the WebApi tests (which use TestHTTPServer) pass cleanly on both the original and the new code. The full test suite hang on aarch64 is actually in the MCP test path, not in TestHTTPServer, so this PR does **not** fix that.

I've reframed the PR accordingly:
- The `poll()` change is a robustness improvement — removes the platform-specific `shutdown`/`close` workaround in `stop()` in favor of a simpler, portable polling loop
- Added a fix for a listener fd leak in `deinit()` (previously it just set the field to null without closing the socket)

Keeping as draft for now.

karlseguin:
It could be related to this: https://github.com/lightpanda-io/browser/pull/1721

egrs:
Restructured: dropped the TestHTTPServer changes (separate concern) and focused on the real root cause.

**Root cause**: \`allocator.create(Self)\` returns undefined memory. When \`init()\` sets fields individually, struct field defaults like \`mutex: std.Thread.Mutex = .{}\` are not applied — they only work with struct literal syntax. The uninitialized mutex caused \`sendResponse()\` to deadlock on \`mutex.lock()\` on aarch64 where the memory contained non-zero bytes.

**Fix**: Use \`self.* = .{...}\` to initialize all fields, ensuring defaults are applied. Applied the same fix to \`App.zig\` which had \`shutdown: bool = false\` never explicitly set.

Single commit, two files, minimal diff.

egrs:
> It could be related to this: #1721

Indeed it was, thanks for the fix ;-)
Applying it to App.

PR #1740, opened 2026-03-08
execute dynamically inserted inline script elements

egrs:
## Summary

- Scripts created via `createElement('script')` with inline content (`textContent`, `.text`, or `innerHTML`) and inserted into the DOM now execute per the HTML spec
- Previously all dynamically inserted scripts without a `src` attribute were unconditionally skipped in `nodeIsReady()`, breaking most JS framework hydration patterns (Vue/Nuxt, React/Next.js, analytics/consent scripts)
- The fix narrows the guard to only skip scripts with neither `src` nor inline content (`firstChild() == null`)

## Details

The root cause was in `Page.zig:nodeIsReady()` — the guard at line 2910 returned early for **all** dynamic scripts without `src`, including ones with inline content that should execute per the [HTML spec](https://html.spec.whatwg.org/multipage/scripting.html#script-processing-model).

| Pattern | Should execute? | Before | After |
|---|---|---|---|
| `createElement + textContent + appendChild` | Yes (sync) | No | Yes |
| `createElement + type=module + textContent + appendChild` | Yes (async) | No | Yes |
| `innerHTML = '<script>...'` (fragment parsing) | No | No | No |
| `createElement + src + appendChild` | Yes (async) | Yes | Yes |

The downstream code (`ScriptManager.addFromElement`) already handled inline source correctly — the only issue was the guard preventing it from being reached.

## A/B test on lequipe.fr

| Metric | Before | After |
|---|---|---|
| Vue islands hydrated | 16/17 | **17/17** |
| Style normalization (Vue re-render) | No | Yes |

## Test plan

- [x] 6 new unit tests in `dynamic_inline.html` (textContent, .text, innerHTML, no-double-execute, empty-script-skip, module inline)
- [x] 316/316 unit tests pass
- [x] `zig fmt` clean
- [x] A/B tested on lequipe.fr — confirms Vue island hydration improvement

### DIFF
--- a/src/browser/Page.zig
+++ b/src/browser/Page.zig
@@ -2911,7 +2911,7 @@ fn nodeIsReady(self: *Page, comptime from_parser: bool, node: *Node) !void {
             // Script was added via JavaScript without a src attribute.
             // Only skip if it has no inline content either — scripts with
             // textContent/text should still execute per spec.
-            if (script.asConstElement().asConstNode().firstChild() == null) {
+            if (node.firstChild() == null) {
                 return;
             }
         }


src/browser/Page.zig
```diff
@@ -2908,9 +2908,12 @@ fn nodeIsReady(self: *Page, comptime from_parser: bool, node: *Node) !void {
     }
     if (node.is(Element.Html.Script)) |script| {
         if ((comptime from_parser == false) and script._src.len == 0) {
-            // script was added via JavaScript, but without a src, don't try
-            // to execute it (we'll execute it if/when the src is set)
-            return;
+            // Script was added via JavaScript without a src attribute.
+            // Only skip if it has no inline content either — scripts with
+            // textContent/text should still execute per spec.
+            if (script.asConstElement().asConstNode().firstChild() == null) {
```
karlseguin:
```suggestion
            if (node.firstChild() == null) {
```

PR #1742, opened 2026-03-09
Context origins

karlseguin:
Enable same-origin cross-frame scripting.

Depends on https://github.com/lightpanda-io/zig-v8-fork/pull/153

This PR allows two frames on the same origin to have access to each others object. There are a couple large commits in this PR which are well documented, so at a high level:

When `window.top` is called, we nee to return the v8::Object that is already associated with the root page's window. For this to work, the identity_map needs to be shared across frames of the same origin. A js.Origin has been added which is shared by every js.Context with the same origin. Rather than being tied to a Context, the identity_map, along with Globals and Temps are tied to the js.Context. The Context is reference counted.

The impact of this change was more significant for finalizers. A object created on Page1 can now be finalized _after_ Page1 goes away. It no longer makes sense for Finalizers to take a *Page and it doesn't make sense for Arenas to live on the Page.

There's a mismatch between our existing containers and these new lifetimes. The new `js.Origin` captures the lifetime of v8 objects well, but what captures the lifetime of a `js.Origin`?  This problem has existed since work on frames began. The "root" Page creates the Factory and the page.arena, and frames reference these. But now this has expanded to include the origins and individual arenas (which can outlive the page they were created on). 

I think renaming `Page` -> `Frame`, and then adding a `Page` which is largely a container for frames and all the things frames share would make sense.. But, for now, I expanded the responsibility of Session. Session now has a `page_arena`, a `factory` and an `origins` lookup. With this change, a Session is managing 2 distinct lifetimes. The first is the lifetime of everything shared across page.navigates - the cookie jar, the history, essentially what Session has always been used for. But the Session is now also aware of the "root" page lifecycle and resets anything that should _not_ be shared across root navigation. It's not a perfect match, but Session was already root-page aware and it's the only entry-point into root-page navigation, so it does fit here without too much change.

(The biggest API change is that all finalizers now take a *Session, rather than a *Page, because the Zig instance and its underlying v8::Object are no longer tied to a Page/Context and can outlive it).

### DIFF
--- a/src/browser/Page.zig
+++ b/src/browser/Page.zig
@@ -332,12 +332,11 @@ pub fn deinit(self: *Page, abort_http: bool) void {
         session.releaseArena(qn.arena);
     }
 
-    const is_root = self.parent == null;
-    session.browser.env.destroyContext(self.js, is_root);
+    session.browser.env.destroyContext(self.js);
 
     self._script_manager.shutdown = true;
 
-    if (is_root) {
+    if (self.parent == null) {
         session.browser.http_client.abort();
     } else if (abort_http) {
         // a small optimization, it's faster to abort _everything_ on the root
--- a/src/browser/Session.zig
+++ b/src/browser/Session.zig
@@ -276,7 +276,7 @@ pub fn replacePage(self: *Session) !*Page {
 
     var current = self.page.?;
     const frame_id = current._frame_id;
-    current.deinit(false);
+    current.deinit(true);
 
     self.resetPageResources();
     self.browser.env.memoryPressureNotification(.moderate);
--- a/src/browser/js/Env.zig
+++ b/src/browser/js/Env.zig
@@ -342,7 +342,7 @@ pub fn createContext(self: *Env, page: *Page) !*Context {
     return context;
 }
 
-pub fn destroyContext(self: *Env, context: *Context, is_root: bool) void {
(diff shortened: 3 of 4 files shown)


src/browser/js/Origin.zig
```diff
@@ -0,0 +1,228 @@
...
+        .origin = self,
+        .session = session,
+        .ptr = ptr,
+        .global = global,
+        .finalizerFn = finalizerFn,
+    };
+    return fc;
+}
+
+pub fn transferTo(self: *Origin, dest: *Origin) !void {
```
krichprollsch, line 186: Claude said you don't transfer `finalizer_callbacks`, is that intentional?

```
  1. transferTo does not transfer finalizer_callbacks (Origin.zig:202-228)

  When Context.setOrigin() is called, origin.transferTo(new_origin) moves globals, temps, and identity_map entries, but NOT finalizer_callbacks. Then the old origin is deinited, which calls all remaining finalizers on objects that were just transferred to
   the new origin.

  This is safe only if no finalizer-bearing objects exist at transfer time. The transfer happens in pageHeaderDoneCallback (before document parsing), so likely only the Window object is in the identity_map — and Window has no finalizer. But this invariant
   is fragile and undocumented.

  Suggestion: Add std.debug.assert(self.finalizer_callbacks.count() == 0) in transferTo to make this invariant explicit.
```

PR #1744, opened 2026-03-09
add Range.getBoundingClientRect and getClientRects

egrs:
## Summary

- Implements `Range.getBoundingClientRect()` returning a zero-valued `DOMRect` (headless — no layout)
- Implements `Range.getClientRects()` returning an empty list
- Per [CSSOM View spec](https://developer.mozilla.org/en-US/docs/Web/API/Range/getBoundingClientRect)

Fixes `getBoundingClientRect is not a function` errors on sites where layout code calls this on Range objects (e.g. airbnb.com React layout).

## Test plan

- [x] Tests added to existing `src/browser/tests/range.html`
- [x] 316/316 unit tests pass
- [x] MDN spec verified

### DIFF
--- a/src/browser/tests/range.html
+++ b/src/browser/tests/range.html
@@ -1023,7 +1023,7 @@
 }
 </script>
 
-<script id=getBoundingClientRect>
+<script id=getBoundingClientRect_collapsed>
 {
   const range = new Range();
   const rect = range.getBoundingClientRect();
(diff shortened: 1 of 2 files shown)


karlseguin:
This is tricky. I think it's better to use `self._proto.getCommonAncestorContainer` and, if it's an Element, delegate to `Element.getBoundingClientRect` and `Element.getClientRects`.

If it _isn't_ an Element but the parent is an element, then you can use that (i.e. so that it works with text nodes too).

karlseguin:
I guess the other case where 0,0,0,0 is correct is when the range is collapsed.

egrs:
Updated — now delegates to `Element.getBoundingClientRect` / `getClientRects` via the common ancestor container. If the container is a text node, walks up to the parent element. Returns zeros only when the range is collapsed or has no element ancestor.

PR #1747, opened 2026-03-09
ci: add e2e tests w/ web bot auth

krichprollsch:


mookums:
I don't think we need this anymore?

PR #1752, opened 2026-03-09
build: add code formatting check

arrufat:
Not sure this is something we want, but I add it to all my projects, so compilation fails early if the code is not properly formatted.

Since #1751 has not been merged, checks should fail here (to test the feature is working)


arrufat:
Ah @karlseguin was too fast merging #1751 haha

PR #1755, opened 2026-03-09
cdp: add a dummy Page.getLayoutMetrics

krichprollsch:


karlseguin (commented):
Various properties in the browser report 1920x1080, e.g. window.innerWidth and screen.height.

PR #1757, opened 2026-03-09
add LP.getInteractiveElements CDP command

egrs:
## What

New `LP.getInteractiveElements` CDP command that returns every actionable element on a page in a single call: buttons, links, inputs, ARIA widgets, contenteditable regions, and elements with registered event listeners.

## Why

AI web agents today (Browser-Use, WebVoyager, Stagehand…) determine "what can I click" by taking screenshots, overlaying numbered markers, and sending them to a vision LLM. This is slow, expensive, and error-prone.

Lightpanda can answer this question natively because it already tracks all event listeners internally. The `EventManager` HashMap and inline handler registry give us something no other browser exposes to automation code: a complete map of which elements have `addEventListener` registrations and which event types they listen for. Chrome's `DOMDebugger.getEventListeners` is debug-only and requires per-element calls — LP does it in one shot across the entire DOM.

Combined with `LP.getMarkdown`, this gives agents a complete semantic picture of a page without screenshots or pixel-level reasoning.

## What's returned

```json
{
  "elements": [
    {
      "tagName": "button",
      "role": "button",
      "name": "Submit",
      "type": "native",
      "listeners": ["click", "mousedown"],
      "tabIndex": 0,
      "id": "submit-btn",
      "class": "btn primary"
    },
    {
      "tagName": "a",
      "role": "link",
      "name": "Home",
      "type": "native",
      "tabIndex": 0,
      "href": "/home"
    },
    {
      "tagName": "div",
      "role": "button",
      "name": "Custom action",
      "type": "listener",
      "listeners": ["click"],
      "tabIndex": -1
    }
  ],
  "nodeIds": [42, 43, 44]
}
```

Each element includes: tag name, implicit/explicit ARIA role, accessible name (aria-label > alt > title > placeholder > value > text content), interactivity type, registered listener event types, disabled state, tab index, and key attributes (id, class, href, inputType, value, name, placeholder). Null fields are omitted from the JSON.

The `nodeIds` array maps 1:1 to elements — all nodes are registered in the CDP node registry so agents can immediately use them in follow-up calls (DOM.focus, DOM.querySelector, Input.dispatchMouseEvent, etc).

## Interactivity classification

Elements are classified into 5 types, checked in priority order:

| Type | Criteria |
|------|----------|
| `native` | button, a\[href\], area\[href\], input (except hidden), select, textarea, details, summary |
| `aria` | `role` attribute matching an interactive widget role (button, link, tab, menuitem, checkbox, radio, slider, spinbutton, searchbox, combobox, option, treeitem, switch, menuitemcheckbox, menuitemradio) |
| `contenteditable` | `contenteditable=""` or `contenteditable="true"` |
| `listener` | Has `addEventListener` or inline handler (onclick, etc.) registrations |
| `focusable` | Explicit `tabindex >= 0` on an otherwise non-interactive element |

Non-visual elements (script, style, link, meta, head, noscript, template) are skipped.

## Performance

Listener detection is O(1) per element. A target-to-event-types map is pre-built in a single pass over the EventManager lookup + inline handler registry, then both classification ("has any listener?") and type collection ("which event types?") are simple map lookups.

## Known v1 simplifications

These are intentional scope choices, not bugs — can be added in follow-up PRs:

- **Missing ARIA widget roles**: `textbox`, `scrollbar`, `gridcell`, `separator` (most are edge cases — `role="textbox"` divs usually also have `contenteditable`, which IS detected)
- **Missing native interactive elements**: `label` (forwards clicks to associated input, which IS listed), `audio/video[controls]` (browser-rendered controls), `embed`, `iframe`
- **`<select>` implicit role**: Always returns "combobox" — should be "listbox" when `multiple` or `size>1`
- **Accessible name**: Simplified priority chain — doesn't resolve `aria-labelledby` references
- **`isDisabled`**: Checks attribute only, not `fieldset[disabled]` ancestor propagation
- **Text content**: Returns first non-empty text node, not full concatenation (e.g. `<button>Save <span>changes</span></button>` returns "Save")

## Files

| File | Change |
|---|---|
| `src/browser/interactive.zig` | New — core detection, classification, collection (~480 LOC incl. 12 unit tests) |
| `src/cdp/domains/lp.zig` | Add command handler + CDP integration test |
| `src/lightpanda.zig` | Register module |

## Test plan

- [x] 330/330 unit tests passing (12 new interactive tests + 1 CDP test + all existing)
- [x] `zig fmt` clean
- [x] CDP integration test on example.com (1 link detected) and custom HTML page (13 elements, all types)
- [x] Performance: all allocations are arena-based (cmd.arena / call_arena), no per-call leaks

### DIFF
--- a/src/browser/interactive.zig
+++ b/src/browser/interactive.zig
@@ -19,6 +19,7 @@
 const std = @import("std");
 
 const Page = @import("Page.zig");
+const URL = @import("URL.zig");
 const TreeWalker = @import("webapi/TreeWalker.zig");
 const Element = @import("webapi/Element.zig");
 const Node = @import("webapi/Node.zig");
@@ -125,8 +126,8 @@ pub const InteractiveElement = struct {
 /// Collect all interactive elements under `root`.
 pub fn collectInteractiveElements(
     root: *Node,
-    page: *Page,
     arena: Allocator,
+    page: *Page,
 ) ![]InteractiveElement {
     // Pre-build a map of event_target pointer → event type names,
     // so classify and getListenerTypes are both O(1) per element.
@@ -134,7 +135,7 @@ pub fn collectInteractiveElements(
 
     var results: std.ArrayList(InteractiveElement) = .empty;
 
-    var tw = TreeWalker.FullExcludeSelf.init(root, .{});
+    var tw = TreeWalker.Full.init(root, .{});
     while (tw.next()) |node| {
         const el = node.is(Element) orelse continue;
         const html_el = el.is(Element.Html) orelse continue;
@@ -163,7 +164,10 @@ pub fn collectInteractiveElements(
             .tab_index = html_el.getTabIndex(),
             .id = el.getAttributeSafe(comptime .wrap("id")),
             .class = el.getAttributeSafe(comptime .wrap("class")),
-            .href = el.getAttributeSafe(comptime .wrap("href")),
+            .href = if (el.getAttributeSafe(comptime .wrap("href"))) |href|
+                URL.resolve(arena, page.base(), href, .{ .encode = true }) catch href
+            else
+                null,
             .input_type = getInputType(el),
             .value = getInputValue(el),
             .element_name = el.getAttributeSafe(comptime .wrap("name")),
@@ -348,7 +352,34 @@ fn getTextContent(node: *Node) ?[]const u8 {
 }
 
 fn isDisabled(el: *Element) bool {
-    return el.getAttributeSafe(comptime .wrap("disabled")) != null;
+    if (el.getAttributeSafe(comptime .wrap("disabled")) != null) return true;
+    return isDisabledByFieldset(el);
+}
+
+/// Check if an element is disabled by an ancestor <fieldset disabled>.
+/// Per spec, elements inside the first <legend> child of a disabled fieldset
+/// are NOT disabled by that fieldset.
+fn isDisabledByFieldset(el: *Element) bool {
+    const element_node = el.asNode();
+    var current: ?*Node = element_node._parent;
+    while (current) |node| {
+        current = node._parent;
+        const ancestor = node.is(Element) orelse continue;
+
+        if (ancestor.getTag() == .fieldset and ancestor.getAttributeSafe(comptime .wrap("disabled")) != null) {
+            // Check if element is inside the first <legend> child of this fieldset
+            var child = ancestor.firstElementChild();
+            while (child) |c| {
+                if (c.getTag() == .legend) {
+                    if (c.asNode().contains(element_node)) return false;
+                    break;
+                }
+                child = c.nextElementSibling();
+            }
+            return true;
+        }
+    }
+    return false;
 }
 
 fn getInputType(el: *Element) ?[]const u8 {
@@ -381,7 +412,7 @@ fn testInteractive(html: []const u8) ![]InteractiveElement {
     const div = try doc.createElement("div", null, page);
     try page.parseHtmlAsChildren(div.asNode(), html);
 
-    return collectInteractiveElements(div.asNode(), page, page.call_arena);
+    return collectInteractiveElements(div.asNode(), page.call_arena, page);
 }
 
 test "browser.interactive: button" {
@@ -452,6 +483,20 @@ test "browser.interactive: disabled" {
     const elements = try testInteractive("<button disabled>Off</button>");
     try testing.expectEqual(1, elements.len);
     try testing.expect(elements[0].disabled);
+}
+
+test "browser.interactive: disabled by fieldset" {
+    const elements = try testInteractive(
+        \\<fieldset disabled>
+        \\  <button>Disabled</button>
+        \\  <legend><button>In legend</button></legend>
+        \\</fieldset>
+    );
+    try testing.expectEqual(2, elements.len);
+    // Button outside legend is disabled by fieldset
+    try testing.expect(elements[0].disabled);
+    // Button inside first legend is NOT disabled
+    try testing.expect(!elements[1].disabled);
 }
 
 test "browser.interactive: non-interactive div" {
--- a/src/cdp/domains/lp.zig
+++ b/src/cdp/domains/lp.zig
@@ -71,7 +71,7 @@ fn getInteractiveElements(cmd: anytype) !void {
     else
         page.document.asNode();
 
-    const elements = try interactive.collectInteractiveElements(root, page, cmd.arena);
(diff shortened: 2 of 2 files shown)


src/browser/interactive.zig
```diff
@@ -0,0 +1,481 @@
...
+                const content = std.mem.trim(u8, text.getWholeText(), &std.ascii.whitespace);
+                if (content.len > 0) return content;
+            }
+        }
+    }
+    return null;
+}
+
+fn isDisabled(el: *Element) bool {
+    return el.getAttributeSafe(comptime .wrap("disabled")) != null;
```
karlseguin: FWIW, this wouldn't catch if the field is disabled by a parent fieldset. There is a `isDisabledByFieldset` in FormData, but it isn't public. We could expose it, or maybe there needs to be AI-specific utilities class?  There's also code above that skips non-visual elements and maybe what is "non-visual" should be standard across the codebase? cc @arrufat

src/browser/interactive.zig
```diff
@@ -0,0 +1,481 @@
...
+            .tag_name = el.getTagNameLower(),
+            .role = getRole(el),
+            .name = getAccessibleName(el),
+            .interactivity_type = itype,
+            .listener_types = listener_types,
+            .disabled = isDisabled(el),
+            .tab_index = html_el.getTabIndex(),
+            .id = el.getAttributeSafe(comptime .wrap("id")),
+            .class = el.getAttributeSafe(comptime .wrap("class")),
+            .href = el.getAttributeSafe(comptime .wrap("href")),
```
karlseguin: This will be the attribute as it appears in the source, i.e. it could be relative. Maybe that's ok. If not, it needs to be resolve against the page.url:

```zig
if (el.getAttributeSafe(comptime .wrap("href")) |href| {
   try URL.resolve(arena page.base(), href, .{ .encode = true })
}
```

src/browser/interactive.zig
arrufat: Oh, I have something very similar in my local semantic-tree branch!
I just added an `isMetadata()` method to `Element.Tag` that covers that exact same list (`script`, `style`, `meta`, etc.).

https://github.com/lightpanda-io/browser/blob/a318c6263dfc5c681fd2b132d8a28794feebff41/src/browser/webapi/Element.zig#L1607-L1612

We can definitely share the logic.

Great catch on the fieldset disabled edge case too. I completely agree. Exposing `isDisabledByFieldset` from `FormData.zig` (probably moving it to `Element.zig` so it's globally available) is he right move rather than burying it in an AI utility class.

arrufat (commented):
In `src/browser/interactive.zig`, around line 140:

```zig
var tw = TreeWalker.FullExcludeSelf.init(root, .{});
```

Currently, `collectInteractiveElements` uses `TreeWalker.FullExcludeSelf`. This works perfectly in the unit tests because the test HTML is wrapped in a dummy `<div>`, so the dummy `<div>` is skipped and the actual test elements are evaluated.

However, in the CDP command `LP.getInteractiveElements`, if an AI agent passes a specific `nodeId` (e.g., the ID of a `<button>`), that button becomes the `root` passed to the `TreeWalker`. Because of `FullExcludeSelf`, the walker starts at the button's *first child* and completely skips evaluating the `<button>` itself. The agent will get an empty array back, entirely missing the interactivity of the node they explicitly queried.

Switching it to `TreeWalker.Full` fixes this!
- If a specific `nodeId` is passed, it evaluates the target element itself.
- If no `nodeId` is passed (the whole page), the `root` is the `Document` node, which safely fails the `node.is(Element)` check and is skipped anyway.
- The unit tests will still pass (the dummy `<div>` will just be evaluated as non-interactive).

src/browser/interactive.zig
```diff
@@ -0,0 +1,481 @@
...
+            try jw.objectField("placeholder");
+            try jw.write(v);
+        }
+
+        try jw.endObject();
+    }
+};
+
+/// Collect all interactive elements under `root`.
+pub fn collectInteractiveElements(
```
karlseguin, line 127: As a convention, Allocator is the first parameter (or the 2nd one if it's a method and the first has to be the receiver) and *Page is the last one.

arrufat: I also read somewhere, that with the new `Io` stuff, the `io` parameter takes preference over the allocator. So it should be:

0. self
1. io
2. allocator

src/browser/interactive.zig
```diff
@@ -0,0 +1,481 @@
...
+    page: *Page,
+    arena: Allocator,
+) ![]InteractiveElement {
+    // Pre-build a map of event_target pointer → event type names,
+    // so classify and getListenerTypes are both O(1) per element.
+    const listener_targets = try buildListenerTargetMap(page, arena);
+
+    var results: std.ArrayList(InteractiveElement) = .empty;
+
+    var tw = TreeWalker.FullExcludeSelf.init(root, .{});
```
arrufat: Currently, `collectInteractiveElements` uses `TreeWalker.FullExcludeSelf`. This works perfectly in the unit tests because the test HTML is wrapped in a dummy `<div>`, so the dummy `<div>` is skipped and the actual test elements are evaluated.

However, in the CDP command `LP.getInteractiveElements`, if an AI agent passes a specific `nodeId` (e.g., the ID of a `<button>`), that button becomes the `root` passed to the `TreeWalker`. Because of `FullExcludeSelf`, the walker starts at the button's *first child* and completely skips evaluating the `<button>` itself. The agent will get an empty array back, entirely missing the interactivity of the node they explicitly queried.

Switching it to `TreeWalker.Full` fixes this!
- If a specific `nodeId` is passed, it evaluates the target element itself.
- If no `nodeId` is passed (the whole page), the `root` is the `Document` node, which safely fails the `node.is(Element)` check and is skipped anyway.
- The unit tests will still pass (the dummy `<div>` will just be evaluated as non-interactive).

arrufat:
After thoroughly checking this PR, I think it will be very useful in my upcoming [semantic-tree](https://github.com/lightpanda-io/browser/tree/semantic-tree) branch.
So I'll wait until this is merged, and apply my changes on top of this.

egrs:
Thanks for the thorough review! All points addressed in dc395835:

- **TreeWalker.Full** — good catch, `FullExcludeSelf` would silently return empty results when querying a specific nodeId. Fixed.
- **href resolution** — now resolved to absolute URLs via `URL.resolve`, same pattern as `markdown.zig`.
- **fieldset disabled** — `isDisabled` now walks ancestors and respects the first-`<legend>` exemption per spec. Added a dedicated test.
- **Parameter ordering** — allocator before `*Page`, fixed.

PR #1759, opened 2026-03-10
Network poll runtime

mrdimidium:
Extracted a known working part from #1719. It introduces a NetworkRuntime entity similar in concept to Io from zig16—an abstract implementation that hides the event loop internally and is passed as a dependency to other modules (e.g., HttpClient).

I also followed Karl's advice in one of the previous PRs and split Net.zig into several separate files.

### DIFF
--- a/src/network/Runtime.zig
+++ b/src/network/Runtime.zig
@@ -104,6 +104,8 @@ pub fn deinit(self: *Runtime) void {
         self.allocator.free(data[0..ca_blob.len]);
     }
 
+    self.robot_store.deinit();
+
     globalDeinit();
 }
 


src/browser/Page.zig
```diff
@@ -3052,7 +3052,7 @@ pub const NavigateReason = enum {
 pub const NavigateOpts = struct {
     cdp_id: ?i64 = null,
     reason: NavigateReason = .address_bar,
-    method: Http.Method = .GET,
+    method: Net.Method = .GET,
```
karlseguin: I know what an `Http.Method` is, but what's a `Net.Method`? Does it relate to the `method` in a cdp message..it's become less clear/obvious.

You can improve this by exposing Method (and Header) on HttpClient...HttpClient.Method, HttpClient.Headers.

src/network/Runtime.zig
```diff
@@ -0,0 +1,273 @@
...
+
+    try posix.setsockopt(listener, posix.SOL.SOCKET, posix.SO.REUSEADDR, &std.mem.toBytes(@as(c_int, 1)));
+    if (@hasDecl(posix.TCP, "NODELAY")) {
+        try posix.setsockopt(listener, posix.IPPROTO.TCP, posix.TCP.NODELAY, &std.mem.toBytes(@as(c_int, 1)));
+    }
+
+    try posix.bind(listener, &address.any, address.getOsSockLen());
+    try posix.listen(listener, self.config.maxPendingConnections());
+
+    for (&self.listeners, &self.pollfds) |*slot, *pfd| {
```
karlseguin: Multiple listeners doesn't seem neccesary, unless the ability to run a CDP server and an MCP server (which doesn't exist yet) within the same process is something that's been discussed.

src/network/Runtime.zig
```diff
@@ -0,0 +1,273 @@
...
+    };
+}
+
+pub fn deinit(self: *Runtime) void {
+    if (self.ca_blob) |ca_blob| {
+        const data: [*]u8 = @ptrCast(ca_blob.data);
+        self.allocator.free(data[0..ca_blob.len]);
+    }
+
+    global_deinit_once.call();
```
karlseguin: If deinit is called multiple times, the above free is invalid. Is there a reason to use `std.once` in this specific place but nowhere else?

src/network/Runtime.zig
mrdimidium: The multi-listener interface is closer to Io in ZIG 16 (which we'll probably want to move to in the future). If you're strongly opposed, I'll simplify the code.

karlseguin: strongly oppose.

karlseguin:
>     * Tests hang on MacOS (but not Linux)

No combination of shutdown/close seems to unblock MacOS. `stop` is called, but `poll` never returns.

Do you want to bring in the pipe trick from the other PR ?

src/network/Runtime.zig
mrdimidium: Thinking about it, I'm not so sure. The curl documentation says that cleanup should be called exactly the same number of times as global init was called. The once restriction makes Runtime implicitly a singleton, although this isn't specified anywhere. I think we can simply call init/clenup within init/deinit, without any wrappers.

karlseguin:
Latest commit fixes the MacOS hang.

PR #1765, opened 2026-03-10
add LP.getStructuredData CDP command

egrs:
## Summary

New custom CDP command `LP.getStructuredData` that extracts all machine-readable structured data from a page in a single call.

This is the third command in the LP domain, alongside `LP.getMarkdown` and `LP.getInteractiveElements` (#1757). Together they give AI agents a complete, efficient view of any web page:

- **getMarkdown** — what the page *says* (content)
- **getInteractiveElements** — what the page *does* (actions)
- **getStructuredData** — what the page *is* (metadata + semantics)

### Why this matters for Lightpanda

Every headless browser can extract meta tags with injected JavaScript. But `LP.getStructuredData` does it natively in a single CDP call — no JS execution, no round-trips, no script injection overhead. For AI agents processing thousands of pages, this matters.

JSON-LD extraction is particularly valuable: it gives agents structured Schema.org data (products, articles, events, reviews, FAQs, breadcrumbs) without any parsing or heuristics. 53% of websites include JSON-LD, and it's the format Google explicitly recommends.

### What it extracts

| Format | Source | Adoption |
|--------|--------|----------|
| JSON-LD | `<script type="application/ld+json">` | 53% |
| Open Graph | `<meta property="og:*">` | 70% |
| Twitter/X Cards | `<meta name="twitter:*">` | 56% |
| HTML meta | `<title>`, `<meta name="...">`, charset | ~100% |
| Link elements | `<link rel="canonical\|icon\|manifest\|alternate">` | ~100% |

### Example output (bbc.co.uk)

```json
{
  "jsonLd": ["{\"@context\":\"http://schema.org\",\"@type\":\"CollectionPage\",...}"],
  "openGraph": {
    "title": "BBC - Home",
    "type": "website",
    "url": "https://www.bbc.co.uk/",
    "image": "https://static.files.bbci.co.uk/.../poster-1024x576.png",
    "image:alt": "BBC logo",
    "description": "The best of the BBC...",
    "site_name": "BBC"
  },
  "twitterCard": {
    "card": "summary_large_image",
    "site": "@BBC",
    "creator": "@BBC",
    "title": "BBC - Home"
  },
  "meta": {
    "language": "en-GB",
    "charset": "utf-8",
    "title": "BBC - Home",
    "description": "The best of the BBC...",
    "viewport": "width=device-width, initial-scale=1",
    "theme-color": "#FFFFFF",
    "robots": "max-image-preview:large"
  },
  "links": {
    "canonical": "https://www.bbc.co.uk/",
    "icon": "https://static.files.bbci.co.uk/.../favicon-32.png",
    "manifest": "https://static.files.bbci.co.uk/.../manifest.json"
  },
  "alternate": [
    {"href": "https://www.bbc.com/", "hreflang": "en"},
    {"href": "https://www.bbc.co.uk/", "hreflang": "en-gb"}
  ]
}
```

### Implementation

- Single-pass TreeWalker over the DOM, ~300 LOC in `src/browser/structured_data.zig`
- Link hrefs resolved to absolute URLs via `URL.resolve`
- OG sub-namespaces handled: `article:`, `profile:`, `book:`, `music:`, `video:`
- JSON-LD returned as string array (consumer parses) — avoids corrupting CDP response if page has malformed JSON-LD
- 10 unit tests + 2 CDP integration tests

### Known v1 simplifications

- Microdata (`itemscope`/`itemprop`) not extracted — 22% adoption, declining, overlaps with JSON-LD
- RDFa not extracted — complex to parse, largely redundant with JSON-LD
- `<link rel="next/prev">` not included (could add for pagination use cases)

## Test plan

- [x] 329/329 unit tests pass (`zig build test`)
- [x] CDP integration verified on example.com (minimal metadata) and bbc.co.uk (rich structured data)
- [x] `zig fmt` clean

### DIFF
--- a/src/browser/structured_data.zig
+++ b/src/browser/structured_data.zig
@@ -93,11 +93,41 @@ pub const StructuredData = struct {
     }
 };
 
+/// Serializes properties as a JSON object. When a key appears multiple times
+/// (e.g. multiple og:image tags), values are grouped into an array.
+/// Alternatives considered: always-array values (verbose), or an array of
+/// {key, value} pairs (preserves order but less ergonomic for consumers).
 fn writeProperties(jw: anytype, properties: []const Property) !void {
     try jw.beginObject();
-    for (properties) |prop| {
+    for (properties, 0..) |prop, i| {
+        // Skip keys already written by an earlier occurrence.
+        var already_written = false;
+        for (properties[0..i]) |prev| {
+            if (std.mem.eql(u8, prev.key, prop.key)) {
+                already_written = true;
+                break;
+            }
+        }
+        if (already_written) continue;
+
+        // Count total occurrences to decide string vs array.
+        var count: usize = 0;
+        for (properties) |p| {
+            if (std.mem.eql(u8, p.key, prop.key)) count += 1;
+        }
+
         try jw.objectField(prop.key);
-        try jw.write(prop.value);
+        if (count == 1) {
+            try jw.write(prop.value);
+        } else {
+            try jw.beginArray();
+            for (properties) |p| {
+                if (std.mem.eql(u8, p.key, prop.key)) {
+                    try jw.write(p.value);
+                }
+            }
+            try jw.endArray();
+        }
     }
     try jw.endObject();
 }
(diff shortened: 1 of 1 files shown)


src/browser/structured_data.zig
```diff
@@ -0,0 +1,435 @@
...
+    };
+    for (relevant_rels) |known| {
+        if (std.ascii.eqlIgnoreCase(rel, known)) {
+            try links.append(arena, .{ .key = known, .value = href });
+            return;
+        }
+    }
+}
+
+fn startsWith(haystack: []const u8, prefix: []const u8) bool {
```
arrufat: I think we can use `std.mem.startsWith`.

arrufat:
Great work on this PR! The native Zig implementation is very clean and the performance benefits are clear.

One minor observation: regarding the JSON serialization in `src/browser/structured_data.zig`. In the `writeProperties` function, if a page has multiple tags with the same property (e.g., multiple `<meta property="og:image">`), it will output duplicate keys in the resulting JSON object.

While technically valid JSON, many JSON parsers will silently overwrite previous values and only keep the last one. It might be worth grouping duplicate keys into an array (either during collection or serialization) so consumers don't lose data.

Otherwise, this looks really solid, and will simplify my semantic-tree branch even further.

egrs:
Thanks for the review! Addressed both points in f4ca5313:
- Replaced custom `startsWith` with `std.mem.startsWith`
- `writeProperties` now groups duplicate keys into arrays (e.g. multiple `og:image` → JSON array), single-occurrence keys stay as strings

If you'd prefer a different approach — always-array values, or an array of {key, value} pairs — happy to switch.

arrufat:
The current approach (`string | array`) is perfect.

Since the primary consumers for this CDP command are AI frameworks written in Python or TypeScript (Browser-Use,Stagehand), ergonomics are more important than strict type homogeneity.

Having to index [0] for a simple <title> (in the always-array approach) would be annoying DX, and an array of objects breaks direct dictionary lookups.

Dynamic languages handle `string | string[]` easily, so this current implementation strikes
the perfect balance.

PR #1771, opened 2026-03-10
Prefer `ScriptCompiler::CompileFunction` to compile attribute listeners

nikneym:
This changes our approach on attribute listeners by taking advantage of [`CompileFunction`](https://v8.github.io/api/head/classv8_1_1ScriptCompiler.html#a3a15bb5a7dfc3f998e6ac789e6b4646a). Reference usage for it can be found [in Chromium's source](https://sourcegraph.com/r/github.com/chromium/chromium/-/blob/third_party/blink/renderer/bindings/core/v8/js_event_handler_for_content_attribute.cc?L218).

### DIFF
--- a/src/browser/tests/window/body_onload1.html
+++ b/src/browser/tests/window/body_onload1.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<body onload=func1></body>
+<body onload="func1(event)"></body>
 <script src="../testing.js"></script>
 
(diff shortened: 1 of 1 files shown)


nikneym:
I believe CI can succeed once https://github.com/lightpanda-io/demo/pull/139 merged.

karlseguin:
Looks fine, except for the newly failing integration test, which seems like a real bug.

PR #1774, opened 2026-03-10
update live ranges after CharacterData and DOM mutations

egrs:
## Summary

- Track all live Range objects on Page via intrusive linked list, updating boundary offsets when CharacterData content changes (insertData, deleteData, replaceData, splitText) or when nodes are inserted/removed from the tree
- Fix pre-existing bugs in Range.deleteContents (loop read shrinking end_offset) and Range.insertNode (non-collapsed double-increment)
- Route textContent/nodeValue/data setters through replaceData so range updates fire consistently

Spec: https://dom.spec.whatwg.org/#concept-range (steps in insert, remove, replace data, split text algorithms)

## Test plan

- [x] 346/346 unit tests pass (`make test`)
- [x] 16 new range_mutations.html tests covering insert/delete/replace/split/appendChild/insertBefore/removeChild
- [x] 9 WPT Range-mutations test files pass at 100% (~1330 new passing subtests)
- [x] CDP integration test (dev vs nightly)
- [x] Performance regression test (new linked list iteration in mutation paths)

### DIFF
--- a/src/browser/webapi/Range.zig
+++ b/src/browser/webapi/Range.zig
@@ -680,10 +680,6 @@ fn getContainerElement(self: *const Range) ?*Node.Element {
     return parent.is(Node.Element);
 }
 
-pub fn deinit(self: *Range, _: bool, page: *Page) void {
-    page._live_ranges.remove(&self._proto._range_link);
-}
-
 pub const JsApi = struct {
     pub const bridge = js.Bridge(Range);
 
@@ -691,8 +687,6 @@ pub const JsApi = struct {
         pub const name = "Range";
         pub const prototype_chain = bridge.prototypeChain();
         pub var class_id: bridge.ClassId = undefined;
-        pub const weak = true;
-        pub const finalizer = bridge.finalizer(Range.deinit);
     };
 
     // Constants for compareBoundaryPoints


src/browser/webapi/Range.zig
```diff
@@ -351,9 +356,10 @@ pub fn insertNode(self: *Range, node: *Node, page: *Page) !void {
         _ = try container.insertBefore(node, ref_child, page);
     }
 
-    // Update range to be after the inserted node
-    if (self._proto._start_container == self._proto._end_container) {
-        self._proto._end_offset += 1;
+    // Per spec step 11: if range was collapsed, extend end to include inserted node.
+    // Non-collapsed ranges are already handled by the live range update in the insert path.
+    if (was_collapsed and self._proto._start_container == self._proto._end_container) {
```
karlseguin: Pretty sure if `was_collapsed` is true, then  `self._proto._start_container == self._proto._end_container` has to be true too (that's part of what makes it collapsed)

src/browser/Page.zig
```diff
@@ -2867,6 +2897,121 @@ pub fn childListChange(
...
+/// Update all live ranges after a replaceData mutation on a CharacterData node.
+/// Per DOM spec: insertData = replaceData(offset, 0, data),
+///               deleteData = replaceData(offset, count, "").
+/// All parameters are in UTF-16 code unit offsets.
+pub fn updateRangesForCharacterDataReplace(self: *Page, target: *Node, offset: u32, count: u32, data_len: u32) void {
+    var it: ?*std.DoublyLinkedList.Node = self._live_ranges.first;
+    while (it) |link| : (it = link.next) {
+        const ar: *AbstractRange = @fieldParentPtr("_range_link", link);
+
+        if (ar._start_container == target) {
```
karlseguin: I think I'd prefer to have all this range-specific logic in AbstractRange:

```
const ar: *AbstractRange = @fieldParentPtr("_range_link", link);
ar.updateForCharacterDataReplace(node, offset, count, data_len);
```

Same for `updateRangesForSplitText`, `updateRangesForNodeInsertion` and  `updateRangesForNodeRemoval`.

karlseguin (commented):
The biggest issue is that, once created, a range is never removed and thus we keep scanning . I can try to address that in a follow-up PR.

egrs:
Addressed both review comments:

- **Simplified collapsed check** in `insertNode` — removed redundant `start_container == end_container` (collapsed implies it)
- **Moved per-range logic to AbstractRange** — `Page` now just iterates the list and delegates to `ar.updateForCharacterDataReplace()`, `ar.updateForSplitText()`, etc.

**Re: live-range cleanup** — attempted a GC finalizer (`weak = true` + `bridge.finalizer(Range.deinit)`) but it hit a Zig module type mismatch in CI (`*Page` vs `*Session`) that doesn't affect any other finalizer in the codebase. Reverted and opened #1779 to track separately.

PR #1776, opened 2026-03-11
Add native Semantic Tree extraction engine for AI agents

arrufat:
This Pull Request introduces a native Semantic Tree extraction engine to Lightpanda. It is specifically designed to provide AI agents and LLM-based frameworks (like Stagehand and Browser-Use) with a high-performance, noise-free representation of web pages.

### Key Capabilities
*   **Semantic Tree Extraction Engine:** Provides a high-signal, cleanly structured accessibility tree natively computed in Zig, avoiding the overhead of heavy JavaScript injection.
*   **New CDP Commands:** Integrates `LP.getSemanticTree` directly into the CDP to enable seamless drop-in support for agent workflows.
*   **New CLI Dump Modes:** Adds `--dump semantic_tree` (JSON) and `--dump semantic_tree_text` (Text) for straightforward CLI testing and pipeline integration.
*   **Agent-Optimized Traversal:** Features an allocation-efficient tree walker that intelligently handles fragmented text concatenation, explicit ARIA labels, and structural pruning to prevent massive token bloat in LLM context windows.

### Verification
*   Empirically validated against dense architectures (Wikipedia), highly nested legacy layouts (HackerNews), and modern web-component frameworks (GitHub, Lightpanda.io) to ensure accurate semantic grouping and zero text redundancy.

```console
 ./zig-out/bin/lightpanda fetch --dump semantic_tree_text https://www.openmymind.net/

[1] RootWebArea: Karl Seguin
  [5] link: bsky
  [6] link: rss
  [7] link: contact
  [8] link: github
  [9] link: dark
  [11] link: Is Zig's New Writer Unsafe? Sep 20 2025
  [12] link: Everything is a []u8 Sep 07 2025
  [13] link: I'm too dumb for Zig's new IO interface Aug 22 2025
  [14] link: Zig's new Writer Jul 17 2025
  [15] link: Zig's new LinkedList API (it's time to learn @fieldParentPtr) Apr 10 2025
  [16] link: Allocator.resize Mar 27 2025
  [17] link: ArenaAllocator.free and Nested Arenas Mar 15 2025
  [18] link: Zig's dot star syntax (value.*) Mar 07 2025
  [19] link: GetOrPut With String Keys Feb 27 2025
  [20] link: Comparing Strings as Integers with @bitCast Feb 20 2025
  [21] link: Switching on Strings in Zig Feb 13 2025
  [22] link: Using Generics to Inject Stubs when Testing Feb 07 2025
  [23] link: In Zig, What's a Writer? Jan 28 2025
  [24] link: Using SIMD to Tell if the High Bit is Set Jan 21 2025
  .
  .
  .
```


src/browser/EventManager.zig
```diff
@@ -98,6 +98,14 @@ pub const Callback = union(enum) {
     object: js.Object,
 };
 
+pub fn hasListener(self: *EventManager, target: *EventTarget, typ: []const u8) bool {
```
karlseguin: This isn't used? (And you probably also want to check the inlineHandlers?)

src/browser/interactive.zig
```diff
@@ -344,13 +347,18 @@ fn getTextContent(node: *Node) ?[]const u8 {
...
             }
         }
     }
-    return null;
-}
 
+    if (chunks.items.len == 0) return null;
+    if (chunks.items.len == 1) return chunks.items[0];
+
+    return try std.mem.join(arena, " ", chunks.items);
```
karlseguin: I think I'd build it up directly to avoid the double allocation, it's more or less the same amount of code anyways:

```zig
if (content.len > 0) {
  try arr.appendSlice(arena, content);
  try arr.append(' ');
}
...
if (arr.items.len == 0) return null;
// strip out trailing space
return arr.items[0..arr.items.len - 1];
```

src/cdp/domains/lp.zig
```diff
@@ -18,25 +18,67 @@
...
+    var st = SemanticTree{
+        .dom_node = dom_node,
+        .registry = &bc.node_registry,
+        .page = page,
+        .arena = cmd.arena,
+        .prune = params.prune orelse false,
+    };
+
+    if (params.format) |format| {
+        if (std.mem.eql(u8, format, "text")) {
```
karlseguin: Will std parse into an enum???

src/cdp/AXNode.zig
```diff
@@ -738,6 +742,49 @@ const AXSource = enum(u8) {
...
+                try std.fmt.format(w.aw.writer, "{s}", .{val});
+            } else {
+                // Ignore unexpected types (e.g. booleans) to avoid garbage output
+            }
+        }
+
+        // Mock JSON Stringifier lifecycle methods
+        pub fn beginWriteRaw(_: @This()) !void {}
+        pub fn endWriteRaw(_: @This()) void {}
+        pub fn objectField(_: @This(), _: []const u8) !void {}
```
karlseguin: You only need the first 2. The last 5 aren't used. You're better off not including them..if someone changes what writeName does and _does_ include them, then the build will fail, which I think is better.

src/cdp/AXNode.zig
```diff
@@ -738,6 +742,49 @@ const AXSource = enum(u8) {
...
+    };
+
+    const w = TextCaptureWriter{ .aw = &aw, .writer = &aw.writer };
+
+    const source = try self.writeName(w, page);
+    if (source != null) {
+        // Remove literal quotes inserted by writeString.
+        var raw_text = std.mem.trim(u8, aw.written(), "\"");
+        raw_text = std.mem.trim(u8, raw_text, &std.ascii.whitespace);
+        return try allocator.dupe(u8, raw_text);
```
karlseguin, line 777: Is this dupe necessary? Isn't that the same allocator the writer is created with?

src/SemanticTree.zig
```diff
@@ -0,0 +1,452 @@
...
+        // Check visibility using the engine's checkVisibility which handles CSS display: none
+        if (!el.checkVisibility(self.page)) {
+            return;
+        }
+
+        if (el.is(Element.Html)) |html_el| {
+            if (html_el.getHidden()) return;
+        }
+    } else if (node.is(CData.Text) != null) {
+        const text_node = node.is(CData.Text).?;
```
karlseguin:
```suggestion
        const text_node = node.as(CData.Text);
```

src/SemanticTree.zig
```diff
@@ -0,0 +1,452 @@
...
+
+    pub fn leave(self: *JsonVisitor) !void {
+        try self.jw.endArray();
+        try self.jw.endObject();
+    }
+};
+
+fn isStructuralRole(role: []const u8) bool {
+    // zig fmt: off
+    return std.mem.eql(u8, role, "none") or
```
karlseguin: StaticStringMap ?

src/SemanticTree.zig
karlseguin: Wait, does a capture not work with else if?

```zig
} else if (node.is(CData.Text) != null) |text_node| {
```

If this works, it should be updated in all the places.

src/cdp/domains/lp.zig
arrufat: yes!

src/cdp/AXNode.zig
arrufat: I think we need it, because that memory is in the writer buffer, right? so when we `deinit` it at the end of the scope it will be freed.

src/SemanticTree.zig
arrufat: It works, too.

PR #1777, opened 2026-03-11
mcp: add interactiveElements and structuredData tools

arrufat:
I am wondering if these should also be expose in the CLI through `--dump`...


krichprollsch:
This is a nice addition to MCP tools.

I'm not sure it's really useful for the `fetch` command. You can't directly use the interactive elements list (except for links by you have them in the markdown)

The structured data could be useful in a JSON format, but it's maybe hard to explain :thinking:

PR #1778, opened 2026-03-11
Use zig allocator for libcurl

mrdimidium:
To detect memory leaks from libcurl, this PR setup a custom allocator for libcurl.

The header and size quirks are due to the fact that zig requires the size of the memory block it allocates/frees, so it needs to be stored somewhere. On the other hand, the C specification requires malloc to align memory according to `_Alignof(max_align_of)` (`C11 Standard, §7.22.3`), so we also align the block header too.

### DIFF
--- a/src/network/Runtime.zig
+++ b/src/network/Runtime.zig
@@ -51,8 +51,120 @@ wakeup_pipe: [2]posix.fd_t = .{ -1, -1 },
 
 shutdown: std.atomic.Value(bool) = .init(false),
 
-fn globalInit() void {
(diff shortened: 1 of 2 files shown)


krichprollsch:
Should we use the zig allocator for curl in release mode too?

mrdimidium:
@krichprollsch, I think so for two reasons:
- it's a more universal behavior, which we tested;
- it allows for more flexible handling of memory allocated in C (for example, manually clearing memory allocated in libcurl).

karlseguin:
My only contribution is that it works fine on mac.

krichprollsch (approved):

PR #1781, opened 2026-03-11
Use global connections poll

mrdimidium:
Replaces multiple separate connection pools with one shared one. The peculiarity is that the handle must be cleared before returning to the pool, since the proxy/auth/tls settings may differ.

### DIFF
--- a/src/browser/HttpClient.zig
+++ b/src/browser/HttpClient.zig
@@ -648,12 +648,6 @@ fn requestFailed(transfer: *Transfer, err: anyerror, comptime execute_callback: 
 // can be changed at any point in the easy's lifecycle.
 pub fn changeProxy(self: *Client, proxy: [:0]const u8) !void {
     try self.ensureNoActiveConnection();
-
-    var it = self.in_use.first;
-    while (it) |node| : (it = node.next) {
-        const conn: *Net.Connection = @fieldParentPtr("node", node);
-        try conn.setProxy(proxy);
-    }
     self.http_proxy = proxy;
     self.use_proxy = true;
 }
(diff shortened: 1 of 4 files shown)


src/browser/HttpClient.zig
```diff
@@ -656,31 +663,39 @@ pub fn changeProxy(self: *Client, proxy: [:0]const u8) !void {
...
+    self.use_proxy = self.http_proxy != null;
 }
 
 // Enable TLS verification on all connections.
 pub fn enableTlsVerify(self: *Client) !void {
     // Remove inflight connections check on enable TLS b/c chromiumoxide calls
     // the command during navigate and Curl seems to accept it...
 
-    for (self.handles.connections) |*conn| {
+    var it = self.in_use.first;
```
karlseguin, line 669: This behavior is very different, no? I don't think it's right. It should apply to future connections too, which won't work if they're shared?

mrdimidium: Future connections will set a parameters when sending a request: https://github.com/lightpanda-io/browser/pull/1781/changes#diff-5d7b298f04470765d710790cc4abbe20cf22ff32a757e584548e3cd20df2aa15R715

src/browser/HttpClient.zig
```diff
@@ -752,7 +774,19 @@ pub const PerformStatus = enum {
 };
 
 fn perform(self: *Client, timeout_ms: c_int) !PerformStatus {
+    self.performing = true;
     const running = try self.handles.perform();
+    self.performing = false;
```
karlseguin: never gets reset if `handles.perform()` fails

src/browser/HttpClient.zig
```diff
@@ -871,11 +905,27 @@ fn processMessages(self: *Client) !bool {
...
+        self.releaseConn(conn);
+    } else |_| {
+        // Can happen if we're in a perform() call, so we'll queue this
+        // for cleanup later.
+        self.dirty.append(&conn.node);
+    }
+}
+
+fn releaseConn(self: *Client, conn: *Net.Connection) void {
+    conn.reset() catch {};
```
karlseguin: If I understand, when you get a connection, you're responsible for setting it up. Hence makeREequest calls `conn.setCallbacks`, `conn.setProxy`, etc... Could we just not call `reset` here?  That `catch {}` tells me it doesn't really matter of this code completes or not. Either reset is important, and we need to handle failures, or it isn't, so why call it?

src/browser/HttpClient.zig
```diff
@@ -645,9 +649,12 @@ fn requestFailed(transfer: *Transfer, err: anyerror, comptime execute_callback:
 pub fn changeProxy(self: *Client, proxy: [:0]const u8) !void {
     try self.ensureNoActiveConnection();
 
-    for (self.handles.connections) |*conn| {
-        try conn.setProxy(proxy.ptr);
+    var it = self.in_use.first;
```
karlseguin: If we called `ensureNoActiveConnection` above, doesn't that mean `in_use.first` will always be null?

src/sys/libcurl.zig
```diff
@@ -703,6 +706,15 @@ pub fn curl_multi_poll(
     try errorMCheck(c.curl_multi_poll(multi, raw_fds, @intCast(extra_fds.len), timeout_ms, numfds));
 }
 
+pub fn curl_multi_waitfds(multi: *CurlM, ufds: []CurlWaitFd, fd_count: *c_uint) ErrorMulti!void {
```
karlseguin: not used?

mrdimidium: Sorry, it came from another PR

src/browser/HttpClient.zig
mrdimidium: Yeah, good point, for TLS it's dead code.

src/browser/HttpClient.zig
mrdimidium: Cleanup is necessary because, for example, telemetry or other future code like mcp might take a connection from this pool. Judging by the libcurl API, an error can only occur here if there's an error. I think the return-to-pool function is better suited for reliable cleanup.

PR #1785, opened 2026-03-11
Add structuredClone() global function

krichprollsch:
Next.js hydration requires structuredClone(), which was not implemented. Uses a JSON stringify/parse round-trip via V8 to deep-clone values, covering all JSON-serializable types (objects, arrays, primitives).

### DIFF
--- a/src/browser/webapi/Window.zig
+++ b/src/browser/webapi/Window.zig
@@ -412,6 +412,18 @@ pub fn atob(_: *const Window, input: []const u8, page: *Page) ![]const u8 {
     return decoded;
 }
 
+pub fn structuredClone(_: *const Window, value: js.Value) !js.Value {
+    // Simplified structured clone using JSON round-trip.
+    // Handles JSON-serializable types (objects, arrays, strings, numbers, booleans, null).
+    const local = value.local;
+    const str_handle = js.v8.v8__JSON__Stringify(local.handle, value.handle, null) orelse return error.DataCloneError;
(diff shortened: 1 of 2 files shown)


src/browser/webapi/Window.zig
```diff
@@ -412,6 +412,18 @@ pub fn atob(_: *const Window, input: []const u8, page: *Page) ![]const u8 {
     return decoded;
 }
 
+pub fn structuredClone(_: *const Window, value: js.Value) !js.Value {
+    // Simplified structured clone using JSON round-trip.
+    // Handles JSON-serializable types (objects, arrays, strings, numbers, booleans, null).
+    const local = value.local;
+    const str_handle = js.v8.v8__JSON__Stringify(local.handle, value.handle, null) orelse return error.DataCloneError;
+    const cloned_handle = js.v8.v8__JSON__Parse(local.handle, str_handle) orelse return error.DataCloneError;
```
krichprollsch, line 420: Maybe this way is too simple and I should use something else from v8...

karlseguin: v8::ValueSerializer and v8::ValueDeserialize.   Do you want to merge this as-is, and i can look at using the ValueSerializer/Deserializer?  I think JSON will have three issues: 
1 - It'll fail on circular dependencies
2 - It'll lose data for types like Date and Regex
3 - It won't work for our own types (https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#webapi_types)

krichprollsch:
closed in favor of #1790

PR #1796, opened 2026-03-12
Use common network runtime for telemetry messages

mrdimidium:
Adds support for libcurl requests to the network runtime. The first consumer is telemetry, which now serves as a thin event broker. Since Runtime doesn't support timeouts/intervals, events are reset either upon receiving an event or upon server termination.

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -115,6 +115,8 @@ fn run(allocator: Allocator, main_arena: Allocator) !void {
             };
             defer server.deinit();
 
+            try sighandler.on(lp.Server.shutdown, .{server});
+
             app.network.run();
         },
         .fetch => |opts| {
--- a/src/network/Runtime.zig
+++ b/src/network/Runtime.zig
@@ -491,7 +491,7 @@ fn acceptConnections(self: *Runtime) void {
                 error.ConnectionAborted => {
                     lp.log.warn(.app, "accept connection aborted", .{});
                     continue;
-            },
+                },
                 else => {
                     lp.log.err(.app, "accept error", .{ .err = err });
                     continue;
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -64,17 +64,17 @@ pub fn init(app: *App, address: net.Address) !*Server {
     return self;
 }
 
-pub fn deinit(self: *Server) void {
-    // Stop all active clients
-    {
-        self.client_mutex.lock();
-        defer self.client_mutex.unlock();
+pub fn shutdown(self: *Server) void {
+    self.client_mutex.lock();
+    defer self.client_mutex.unlock();
 
-        for (self.clients.items) |client| {
-            client.stop();
-        }
+    for (self.clients.items) |client| {
+        client.stop();
     }
+}
 
+pub fn deinit(self: *Server) void {
+    self.shutdown();
     self.joinThreads();
     self.clients.deinit(self.allocator);
     self.clients_pool.deinit();


src/telemetry/lightpanda.zig
```diff
@@ -2,140 +2,95 @@ const std = @import("std");
...
+}
+
+pub fn deinit(self: *LightPanda) void {
+    self.flush();
+}
 
-    pub fn send(self: *LightPanda, iid: ?[]const u8, run_mode: Config.RunMode, raw_event: telemetry.Event) !void {
-        const event = try self.mem_pool.create();
-        event.* = .{
+pub fn send(self: *LightPanda, iid: ?[]const u8, run_mode: Config.RunMode, raw_event: telemetry.Event) !void {
```
karlseguin: `send` is what the rest of the code ends up calling. On `main`, all this does is create an event, queues it and signals the worker thread.

In this version, once the buffer is full, it serializes json, has to acquire 3 mutexes and writes to the pipe.  Quite a bit more that can slow down a caller.

src/telemetry/lightpanda.zig
```diff
@@ -2,140 +2,95 @@ const std = @import("std");
...
+    const pending_count = blk: {
+        self.mutex.lock();
+        defer self.mutex.unlock();
+
+        self.pending[self.pcount] = .{
```
karlseguin: Pretty sure there's a window where this can overflow.

src/telemetry/lightpanda.zig
```diff
@@ -2,140 +2,95 @@ const std = @import("std");
...
-            into[i] = @fieldParentPtr("node", node);
-            i += 1;
-            if (i == MAX_BATCH_SIZE) {
-                break;
-            }
-        }
-        return into[0..i];
-    }
-};
+    const conn = self.runtime.getConnection() orelse return;
```
karlseguin: If there's no connection available, we lose the data?

src/network/Runtime.zig
```diff
@@ -55,6 +58,11 @@ wakeup_pipe: [2]posix.fd_t = .{ -1, -1 },
 
 shutdown: std.atomic.Value(bool) = .init(false),
 
+// Async HTTP requests (e.g. telemetry). Created on demand.
+multi: ?*libcurl.CurlM = null,
```
karlseguin, line 68: I'm having a hard time understanding why this is created on demand? It doesn't seem necessary, it seems like it'll be null only for the first few milliseconds of the program, but it requires the null check everywhere.

src/network/Runtime.zig
```diff
@@ -275,41 +287,52 @@ pub fn bind(
...
-            if (self.shutdown.load(.acquire)) break;
+            while (true)
+                _ = posix.read(self.wakeup_pipe[0], &drain_buf) catch break;
         }
 
-        // check new connections;
-        if (self.pollfds[1].revents == 0) continue;
-        self.pollfds[1].revents = 0;
+        // accept new connections
+        if (self.pollfds[1].revents != 0) {
```
karlseguin: Minor, but I think `self.pollfds[0]` and `self.pollfds[1]` can be stored in a local outside the loop.

src/network/Runtime.zig
```diff
@@ -275,41 +287,52 @@ pub fn bind(
...
+        if (self.multi) |multi| {
+            // Drive transfers and process completions.
+            libcurl.curl_multi_perform(multi, &running_handles) catch |err| {
+                lp.log.err(.app, "curl perform", .{ .err = err });
+            };
+            self.processCompletions(multi);
+        }
 
-        listener.onAccept(listener.ctx, socket);
+        if (self.shutdown.load(.acquire) and running_handles == 0)
```
karlseguin: So if we're shutting down but there are handles, we'll continue to process the handles? Is that right?

src/main.zig
```diff
@@ -137,11 +139,27 @@ fn run(allocator: Allocator, main_arena: Allocator) !void {
...
 
-            try lp.mcp.router.processRequests(mcp_server, &stdin.interface);
+            app.network.run();
         },
         else => unreachable,
     }
 }
+
+fn fetchThread(app: *App, url: [:0]const u8, fetch_opts: lp.FetchOpts) void {
+    defer app.network.stop();
```
karlseguin, line 167: Don't see any advantage to have this in the thread. Would be more self contained in the main thread:

```zig
app.network.run();
defer app.network.stop();
var worker_thread = try std.Thread.spawn(.{}, fetchThread, .{ app, url, fetch_opts });
worker_thread.join();
```

Same with the mcp path

src/network/Runtime.zig
```diff
@@ -275,41 +287,52 @@ pub fn bind(
...
         };
 
-        // check wakeup socket
+        // check wakeup pipe
         if (self.pollfds[0].revents != 0) {
             self.pollfds[0].revents = 0;
-
-            // If we were woken up, perhaps everything was cancelled and the iteration can be completed.
-            if (self.shutdown.load(.acquire)) break;
+            while (true)
```
karlseguin, line 387: I'm not sure, but doesn't checking for shutdown when `self.pollfds[0].revents != 0` make the most sense?  You check it on acceptConnection,once, and at the end of the loop, but if it's true, it'll be true here first. The other two checks don't seem necessary if you check here, and I think it better documents the flow, e.g. pollfds[0] is used to signal shutdown.

src/network/Runtime.zig
mrdimidium: This is a silly workaround :) `multi` is a fairly heavy structure; it stores a lot of SSL and can take up to 2MB of RAM in total.

Our performance tests are run without telemetry, meaning the cost of the telemetry thread isn't taken into account in the benchmark.

Without this workaround, the benchmark will exceed memory consumption.

After removing multi from HttpClient, it can be made persistent.

src/network/Runtime.zig
mrdimidium: The idea is that we should try to send buffered telemetry events when we receive `sigterm`, but we shouldn't accept new connections.

src/main.zig
mrdimidium: `app.network.run()` block thread, you cann't run anything after this

src/telemetry/lightpanda.zig
mrdimidium: The event buffer is twice the batch size, so we'll keep trying to send events. But when the buffer runs out, yes, we'll start losing events.

src/network/Runtime.zig
mrdimidium: I'm not sure I understand. They're set in init/bind, but revents are changed from poll at each iteration. Or am I misunderstanding the idea?

src/telemetry/lightpanda.zig
mrdimidium: Yeah. I think a better solution is `setInterval`, which is called from main and flush on main thread, regardless of when the events are posted. I simplify it because serializing 20 events seems cheap. But if you worries let me fix it.

src/telemetry/lightpanda.zig
karlseguin: I don't think that's right? pfcount is reset to 0 a few lines above, so if there's no connection, there's no retry for those events.

src/network/Runtime.zig
karlseguin: It deserves a comment.

src/network/Runtime.zig
karlseguin:
```zig
const poll_fd = &self.pollfds[0];
const listen_fd = &self.pollfds[1];
while (true) {
    self.drainQueue();
```
I guess without bound-checking in ReleaseMode, it isn't a huge win.

src/network/Runtime.zig
karlseguin: Hmm..that makes sense. It deserves a comment. But it would also "hang" the shutdown until any other network activity completes, right? Not just telemetry.  Like if I have a website that downloads a script, in our donecallback, we execute it, it could download another script and repeat forever. That would block the shutdown?

mrdimidium: I plan to stop the cdp threads when we receive a sigterm, and they'll explicitly cancel their requests.

SigHandler has logic for resuscitation, so if the user presses ctrlc again, we'll kill the process. But yes, it could potentially hang. We can add a timeout to the sighandler to kill everyone after a minute.

src/telemetry/lightpanda.zig
mrdimidium: You're right, it's a regression. Connection should be taken first.

src/telemetry/lightpanda.zig
karlseguin: It's not so much the JSON serialization (the original version retained the allocation for the writer/buffer). It's just all those extra opaque network things that could block, either now or in the future.  Very easy to change, say, getConnection and not realize "oh, this will impact all callers of telemetry).

src/network/Runtime.zig
mrdimidium: I'm not sure I understand the ide. We don't terminate the loop when we receive a shutdown because we're trying to send scheduled telemetry but are no longer accepting new connections.

karlseguin: Yes, now that I understand that the shutdown doesn't want to shutdown right away, I understand why it's how it is. Hence why I think this lines needs a comment:

```
if (self.shutdown.load(.acquire) and running_handles == 0)
```

karlseguin:
Also, can we clean up the names. I know I'm guilty of this too, but  as much as possible I'd like namespaces to match the file names.

```
const Network = @import("network/Runtime.zig");
```

Rename `Runtime.zig` to `Network.zig` ? When I see Network.doSomething, and I want to look at `doSomething` I _always_ try to quick-nav to `Network.zig`

```
const net_http = @import("../network/http.zig");
```

rename the imports to `http`? 

There's also 1 place where the http.zig is imported as Net: `const Net = @import("../network/http.zig");` and 1 place where the Runtime is imported as Runtime: `const Runtime = @import("../network/Runtime.zig");`

src/telemetry/lightpanda.zig
mrdimidium: The last commit implement interval flush on the main thread. Now the c dp thread only acquires a mutex and stores the event in the buffer.

karlseguin:
I haven't done a full re-review, because I think the design needs more work. It just has too much potential to lose messages. If I run `go run crawler/main.go -pool 10 https://demo-browser.lightpanda.io/` I get a steady stream of "telemetry buffer exhausted". Out of curiosity, I tried to set the FLUSH_INTERVAL to 10, and that seems to freeze it  (maybe it's spending all of its time trying to run the schedule?). The higher the concurrency, the more message we'll lose.

In the first iteration, a call to `telemetry.record()` had a chance to block. Notwithstanding the data-loss issue, it _seems_ like the current approach is on the right track, because `telemetry.record()` won't block (it's essentially like it was before). But thinking about it more, the processing now happens on the main thread, so it isn't telemetry.record() that can block, it's the main thread, which doesn't sound better.

(The flush on deinit also has me a little uneasy)

src/telemetry/lightpanda.zig
```diff
@@ -2,140 +2,113 @@ const std = @import("std");
...
-            .thread = null,
-            .running = true,
-            .allocator = allocator,
-            .connection = connection,
-            .config = app.config,
-            .mem_pool = std.heap.MemoryPool(LightPandaEvent).init(allocator),
-        };
-    }
+// const URL = "https://telemetry.lightpanda.io";
+const URL = "http://localhost:9876";
```
karlseguin: 🙈

src/telemetry/lightpanda.zig
```diff
@@ -2,140 +2,113 @@ const std = @import("std");
...
-        }
-        return into[0..i];
+    if (dropped > 0) {
+        const wrapped = LightPandaEvent{ .iid = iid, .mode = self.run_mode, .event = .{ .buffer_overflow = .{ .dropped = dropped } } };
+        try std.json.Stringify.value(&wrapped, .{ .emit_null_optional_fields = false }, &writer.writer);
+        try writer.writer.writeByte('\n');
     }
-};
+
+    const conn = self.runtime.getConnection() orelse {
```
karlseguin: Makes sense to make sure we can get a connection before serializing? Not sure, your call.

src/telemetry/telemetry.zig
```diff
@@ -105,43 +102,43 @@ fn getOrCreateId(app_dir_path_: ?[]const u8) ?[36]u8 {
 pub const Event = union(enum) {
     run: void,
     navigate: Navigate,
+    buffer_overflow: BufferOverflow,
```
karlseguin, line 105: @krichprollsch do we need to change anything to accept a new telemetry type?

src/telemetry/lightpanda.zig
```diff
@@ -2,140 +2,113 @@ const std = @import("std");
...
-        }
+    if (h == t and dropped == 0) return;
+    errdefer _ = self.dropped.fetchAdd(dropped, .monotonic);
+
+    var writer = std.Io.Writer.Allocating.init(self.allocator);
+    defer writer.deinit();
+
+    const iid: ?[]const u8 = if (self.iid) |*id| id else null;
+
+    for (h..t) |i| {
```
karlseguin, line 96: server has a max body size of 500KB (last I heard), so we might need to limit this to ~50 ?

src/network/Runtime.zig
```diff
@@ -285,42 +299,105 @@ pub fn bind(
...
-                else => {
-                    lp.log.err(.app, "accept", .{ .err = err });
-                },
-            }
-            continue;
-        };
+        self.fireTicks();
 
-        listener.onAccept(listener.ctx, socket);
+        if (self.shutdown.load(.acquire) and running_handles == 0) {
```
karlseguin, line 407: When I tested it, the batch size was always 1, which is good because it's on the main thread, but a bit less efficient in other ways (including on the server, cc @krichprollsch).  But, when I hit ctrl-c, the program takes a long time to exit. It just keeps processing CDP requests, and it keeps sending batches of 1. It took maybe 1 minute to properly shut down.

Is there  away to improve this? killing a process but having it run for 1 minute firing off telemetry isn't great. Do we have any way to disconnect all other clients, and synchronously flush the telemetry?

src/telemetry/telemetry.zig
krichprollsch: Yes I have to update the telemetry side. But we can merge,we will just loose these specific events in the meantime.

src/telemetry/lightpanda.zig
krichprollsch: Indeed: https://github.com/lightpanda-io/telemetry/blob/main/api/metrics.go#L51

mrdimidium: Different events have different sizes, it's easier to explicitly write events to the writer until we fit within 500 KB.

krichprollsch: The limit in the server side can be changed, 500KB is conservative. we can safely go higher.

src/telemetry/telemetry.zig
krichprollsch: ok telemetry is up to date to get the new event :heavy_check_mark:

src/telemetry/lightpanda.zig
mrdimidium: I don't think it's important, the main thing is that the limits on the backend and in the code match.

PR #1797, opened 2026-03-12
Implement CSSOM and Enhanced Visibility Filtering

arrufat:
This PR upgrades our CSS Object Model (CSSOM) to support stylesheets and dynamic rules, enabling the Semantic Tree to accurately prune hidden elements.

#### Core Improvements:
*   **CSSOM Support:** Implemented `insertRule`, `deleteRule`, `replace`, and `replaceSync`.
*   **Automatic Parsing:** Standard `<style>` tags now automatically parse their content and register rules with the document upon insertion.
*   **Visibility Engine:** Updated `Element.checkVisibility` to match elements against all active document stylesheets, not just inline styles.

#### Impact on Semantic Tree & `test_semantic_css.html`:
Previously, elements hidden via CSS classes or `<style>` tags were incorrectly included in the Semantic Tree. With these changes:
1.  **Correct Pruning:** Elements like `.hidden-via-style-tag { display: none; }` are now correctly identified as invisible and removed from the tree.
2.  **Dynamic JS Support:** Rules added via `sheet.insertRule()` (common in modern web apps) are now respected by the tree walker.
3.  **Cleaner Output:** Significantly reduces token noise by eliminating "phantom" elements that are technically in the DOM but hidden from the user.

**Verification:** New JS tests added for `CSSStyleSheet` APIs and verified pruning in `test_semantic_css.html`.
<details>
<summary>test_semantic_css.html</summary>

```html
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <title>CSS Semantic Tree Showcase</title>
</head>
<body>

  <h1>Visible Content</h1>
  <p>This paragraph is visible and should be in the tree.</p>

  <div class="hidden-via-style-tag">
    <p>This should NOT be in the semantic tree because it's hidden via a style tag!</p>
    <button>Secret Button</button>
  </div>

  <div id="hidden-via-js">
    <p>This should NOT be in the semantic tree because it's hidden via JS insertRule!</p>
    <a href="https://example.com">[Secret Link](https://example.com/)</a>
  </div>

  <div class="hidden-via-visibility">
    <p>This should NOT be in the semantic tree because it has visibility: hidden!</p>
    <button>Hidden Button</button>
  </div>

  <div class="hidden-via-opacity">
    <p>This should NOT be in the semantic tree because it has opacity: 0!</p>
    <input type="text" value="Secret Input">
  </div>

  <div class="non-interactive-via-pointer-events">
    <p>This should be in the tree, but the button should NOT be interactive!</p>
    <button>Unclickable Button</button>
  </div>

  <script>
    // 1. Hide using a standard <style> tag.
    // Our new CSS parser automatically reads the text content of this tag
    // and populates the document's stylesheet list.
    const styleEl = document.createElement('style');
    styleEl.textContent = `
      .hidden-via-style-tag { display: none; }
      .hidden-via-visibility { visibility: hidden; }
      .hidden-via-opacity { opacity: 0; }
      .non-interactive-via-pointer-events { pointer-events: none; }
    `;
    document.head.appendChild(styleEl);

    // 2. Hide using JavaScript sheet.insertRule().
    // This API now correctly parses the rule and adds it to the list,
    // where checkVisibility can now find it.
    const dynamicStyle = document.createElement('style');
    document.head.appendChild(dynamicStyle);
    dynamicStyle.sheet.insertRule('#hidden-via-js { display: none; }', 0);
  </script>
</body>
</html>
```
</details>

### DIFF
--- a/src/browser/webapi/css/CSSStyleSheet.zig
+++ b/src/browser/webapi/css/CSSStyleSheet.zig
@@ -9,6 +9,14 @@ const Parser = @import("../../css/Parser.zig");
 
 const CSSStyleSheet = @This();
 
+pub const CSSError = error{
+    OutOfMemory,
+    IndexSizeError,
+    WriteFailed,
+    StringTooLarge,
+    SyntaxError,
+};
+
 _href: ?[]const u8 = null,
 _title: []const u8 = "",
 _disabled: bool = false,
@@ -46,8 +54,17 @@ pub fn setDisabled(self: *CSSStyleSheet, disabled: bool) void {
 
 pub fn getCssRules(self: *CSSStyleSheet, page: *Page) !*CSSRuleList {
     if (self._css_rules) |rules| return rules;
+
     const rules = try CSSRuleList.init(page);
     self._css_rules = rules;
+
+    if (self.getOwnerNode()) |owner| {
+        if (owner.is(Element.Html.Style)) |style| {
+            const text = try style.asNode().getTextContentAlloc(page.call_arena);
+            try self.replaceSync(text, page);
+        }
+    }
+
     return rules;
 }
 
@@ -84,12 +101,12 @@ pub fn deleteRule(self: *CSSStyleSheet, index: u32, page: *Page) !void {
     page._style_manager.sheetModified();
 }
 
-pub fn replace(self: *CSSStyleSheet, text: []const u8, page: *Page) !js.Promise {
+pub fn replace(self: *CSSStyleSheet, text: []const u8, page: *Page) CSSError!js.Promise {
     try self.replaceSync(text, page);
     return page.js.local.?.resolvePromise(self);
 }
 
-pub fn replaceSync(self: *CSSStyleSheet, text: []const u8, page: *Page) !void {
+pub fn replaceSync(self: *CSSStyleSheet, text: []const u8, page: *Page) CSSError!void {
     const rules = try self.getCssRules(page);
     rules.clear();
 
--- a/src/browser/StyleManager.zig
+++ b/src/browser/StyleManager.zig
@@ -45,7 +45,7 @@ pub const PointerEventsCache = std.AutoHashMapUnmanaged(*Element, bool);
 const StyleManager = @This();
 
 const Tag = Element.Tag;
-const RuleList = std.ArrayList(VisibilityRule);
+const RuleList = std.MultiArrayList(VisibilityRule);
 
 page: *Page,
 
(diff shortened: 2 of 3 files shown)


karlseguin:
Didn't go over it in detail, since it's still pending, but two thoughts:

1 - This only handles inline styles, not external. That seems of limited use, but a good first step. Is the goal to eventually support external style sheets too?

2 - checkVisibility can be called a lot, normally through other APIs, e.g. getBoundingClientRect, so if we go down this route, do we need a better-for-querying in-memory data representation of the "computed" styles?  It's called +20K times on an amazon product page. (wow, now I want to see if we can optimize the current code).

krichprollsch:
I agree not loading external stylesheets limits the benefit of the change, but yes, it's a good start.

The question will be: how important is it to load external stylesheets, I suppose it is for accuracy, but it will slow the browser loading...

And yes, thinking of perf for visiblity check is important. I guess it can be done in a following PR?

So since here we are talking only of inline CSS + pure JS changes, the impact of the PR is limited.

src/browser/tests/css/stylesheet.html
```diff
@@ -419,3 +419,66 @@
...
+{
+    const style = document.createElement('style');
+    document.head.appendChild(style);
+    const sheet = style.sheet;
+
+    testing.expectEqual(0, sheet.cssRules.length);
+
+    sheet.insertRule('.test { color: green; }', 0);
+    testing.expectEqual(1, sheet.cssRules.length);
+    console.warn("constructor:", sheet.cssRules[0].constructor.name);
```
karlseguin:
```suggestion
```

src/browser/webapi/css/CSSStyleSheet.zig
```diff
@@ -54,30 +56,48 @@ pub fn getOwnerRule(self: *const CSSStyleSheet) ?*CSSRule {
...
+
+    const style_rule = try CSSStyleRule.init(page);
+    try style_rule.setSelectorText(parsed_rule.selector, page);
+
+    const style_props = try style_rule.getStyle(page);
+    const style = style_props.asCSSStyleDeclaration();
+    try style.setCssText(parsed_rule.block, page);
+
+    const rules = try self.getCssRules(page);
+    try rules.insert(index, style_rule._proto, page);
```
karlseguin, line 88: Looks like index should be optional and defaults to 0

src/browser/css/Parser.zig
```diff
@@ -293,3 +293,129 @@ fn isBang(token: Tokenizer.Token) bool {
...
+pub const Rule = struct {
+    selector: []const u8,
+    block: []const u8,
+};
+
+pub fn parseStylesheet(input: []const u8) RulesIterator {
+    return RulesIterator.init(input);
+}
+
+pub const RulesIterator = struct {
```
karlseguin, line 306: Maybe unit tests for this?

src/browser/webapi/css/CSSStyleSheet.zig
```diff
@@ -54,30 +56,48 @@ pub fn getOwnerRule(self: *const CSSStyleSheet) ?*CSSRule {
 }
 
 pub fn insertRule(self: *CSSStyleSheet, rule: []const u8, index: u32, page: *Page) !u32 {
-    _ = self;
-    _ = rule;
-    _ = index;
-    _ = page;
-    return 0;
+    var it = Parser.parseStylesheet(rule);
+    const parsed_rule = it.next() orelse return error.SyntaxError;
```
karlseguin, line 78: Not a huge deal, but insertRule seems pretty strict, like if I do this:
```js
s.insertRule('*{};')
```
it complains (because of the semicolon). If that's too much trouble to add, we can skip it.

src/browser/webapi/css/CSSStyleSheet.zig
```diff
@@ -54,30 +56,48 @@ pub fn getOwnerRule(self: *const CSSStyleSheet) ?*CSSRule {
...
-    _ = page;
+    const rules = try self.getCssRules(page);
+    try rules.remove(index);
 }
 
 pub fn replace(self: *CSSStyleSheet, text: []const u8, page: *Page) !js.Promise {
-    _ = self;
-    _ = text;
-    // TODO: clear self.css_rules
+    try self.replaceSync(text, page);
```
karlseguin, line 105: Docs say that this promise "resolves with the CSSStyleSheet."

src/browser/webapi/Element.zig
```diff
@@ -1041,22 +1041,141 @@ pub fn parentElement(self: *Element) ?*Element {
...
+};
 
-    while (current) |el| {
-        if (el.getStyle(page)) |style| {
-            const display = style.asCSSStyleDeclaration().getPropertyValue("display", page);
-            if (std.mem.eql(u8, display, "none")) {
-                return false;
+pub const CssCache = std.AutoHashMapUnmanaged(*Element, CssProperties);
+
+pub fn getCssProperties(el: *Element, page: *Page, doc_sheets: ?*StyleSheetList, cache: ?*CssCache) CssProperties {
```
```suggestion
fn getCssProperties(el: *Element, page: *Page, doc_sheets: ?*StyleSheetList, cache: ?*CssCache) CssProperties {
```

src/browser/webapi/Element.zig
```diff
@@ -1041,22 +1041,141 @@ pub fn parentElement(self: *Element) ?*Element {
...
+    var current: ?*Element = self;
+    while (current) |el| {
+        const props = getCssProperties(el, page, doc_sheets, cache);
+        if (props.pointer_events_none) return true;
+        current = el.parentElement();
+    }
+    return false;
+}
+
+pub fn hasPointerEventsNone(self: *Element, page: *Page) bool {
```
karlseguin: I don't think this is used.

src/browser/webapi/Element.zig
```diff
@@ -1041,22 +1041,141 @@ pub fn parentElement(self: *Element) ?*Element {
...
+    }
+
+    return props;
+}
+
+pub fn hasPointerEventsNoneCached(self: *Element, page: *Page, cache: ?*CssCache) bool {
+    const doc_sheets = page.document.getStyleSheets(page) catch null;
+    var current: ?*Element = self;
+    while (current) |el| {
+        const props = getCssProperties(el, page, doc_sheets, cache);
```
```suggestion
        const props = el.getCssProperties(page, doc_sheets, cache);
```

src/browser/webapi/Element.zig
karlseguin: Also, we usually put the page as the last argument.

karlseguin (commented):
If I run

```
./pandpanda fetch \
 --http_proxy "http://127.0.0.1:3000" \
 --insecure_disable_tls_host_verification \
 "https://www.amazon.com/A_PRODUCT_PAGe"
```

On `main`, it takes ~1.5 - 2 seconds, I get a number of errors, but it seems to dump properly.

If I run it on this branch, I get fewer errors, but it takes 5-6 seconds. I'm using our implementation proxy to reduce the impact of random network differences.

I don't _think_ this is a result of checkVisibility now being much slower. I profiled it, and while it shows up, I don't think it explains a 3x slowdown. Given that we get fewer errors now, I imagine we're doing a better job. Which is good, but...Can anyone replicate my numbers and, if so, it would e really nice to know what's going on.

arrufat:
I am observing a similar behavior: 5 vs 8 seconds, but with fewer errors. Isn't it expected (to a certain point) because we're executing more JS/CSS?

Because we fixed `CSSStyleSheet.replace()` returning the sheet itself and `insertRule()` handling edge-cases, Amazon's complex JS bundles are no longer crashing early. They are now successfully executing complex logic and layout queries (like `getBoundingClientRect`).

Because `Element.checkVisibility()` currently passes `null` for the CSS cache, every layout query triggers a full, un-cached linear scan of every CSS rule in the document to run `el.matches()`.

We should definitely try to close the gap, though, probably by implementing a proper layout-scoped `CssCache` that invalidates on DOM/CSS changes.

PR #1814, opened 2026-03-13
Make `body.onload` getter/setter alias to `window.onload`

nikneym:
We already do this in `HTMLBodyElement`'s `Build`; this PR continues the compat by applying the same to getter and setter.


nikneym:
Once this merged, tests added in #1772 should pass so it can also be merged.

PR #1821, opened 2026-03-14
Ignore partitionKey in cookie operations to fix Puppeteer page.setCookie()

### LINKED ISSUE #1818: Puppeteer page.setCookie() fails because Network.deleteCookies rejects partitionKey
Puppeteer's `page.setCookie()` doesn't work with Lightpanda because of how Puppeteer internally handles cookie setting.

When you call `page.setCookie({ name: 'test', value: 'hello', domain: 'example.com' })`, Puppeteer sends two `Network.deleteCookies` calls before setting the cookie. The first one works fine:

```json
{ "name": "test", "domain": ".example.com", "url": "https://example.com/" }
```

But the second call includes a `partitionKey` field for CHIPS (Chrome's partitioned cookies):

```json
{ "name": "test", "domain": ".example.com", "url": "https://example.com/", "partitionKey": { "topLevelSite": "https//example.com", "hasCrossSiteAncestor": false } }
```

This hits the early return at `src/cdp/domains/network.zig:120`:

```zig
if (params.partitionKey != null) return error.NotImplemented;
```

So the whole `page.setCookie()` call fails even though the actual cookie deletion logic underneath works perfectly.

Repro:

```javascript

const page = await browser.newPage();
await page.goto('https://example.com', { waitUntil: 'domcontentloaded' });

// fails: Protocol error (Network.deleteCookies): NotImplemented
await page.setCookie({ name: 'test', value: 'hello', domain: 'example.com' });
```

Since Lightpanda's cookie jar doesn't have partitioned cookie support, one option might be to just ignore `partitionKey` when it's present and proceed with the delete by name/domain/path. But you might have other plans for this, so wanted to surface it.

The same `partitionKey` check also exists in `setCdpCookie` in `storage.zig` (~line 131), so `Network.setCookie` has the same limitation if `partitionKey` is passed.

Our workaround for now is calling `Network.setCookies` directly via the CDP client, which bypasses the `deleteCookies` call:

```javascript
const client = page._client();
await client.send('Network.setCookies', {
  cookies: [{ name: 'test', value: 'hello', domain: 'example.com', path: '/' }]
});
```

Tested on nightly (`d3dad772`) and v0.2.5, Puppeteer 24.9.0, macOS aarch64.

hobostay:
## Summary
- Fix `Network.deleteCookies` and `Network.setCookie` to ignore `partitionKey` parameter instead of returning `NotImplemented`
- This enables Puppeteer's `page.setCookie()` to work correctly

## Problem
Puppeteer's `page.setCookie()` internally calls `Network.deleteCookies` twice before setting a cookie. The second call includes a `partitionKey` field for CHIPS (Chrome's partitioned cookies):

```json
{ "name": "test", "domain": ".example.com", "url": "https://example.com/", "partitionKey": { "topLevelSite": "https//example.com", "hasCrossSiteAncestor": false } }
```

This caused Lightpanda to return `NotImplemented`, making `page.setCookie()` fail even though the actual cookie deletion/setting logic works fine.

## Solution
Since Lightpanda doesn't support partitioned cookies (CHIPS), we now silently ignore the `partitionKey` parameter and proceed with cookie operations based on name/domain/path matching.

## Changes
- `src/cdp/domains/network.zig`: Log debug message when partitionKey is present in deleteCookies, but don't return error
- `src/cdp/domains/storage.zig`: Log debug message when partitionKey is present in setCdpCookie, but don't return error

## Test plan
Reproduction from issue #1818:
```javascript
const page = await browser.newPage();
await page.goto('https://example.com', { waitUntil: 'domcontentloaded' });
// This now works instead of throwing NotImplemented
await page.setCookie({ name: 'test', value: 'hello', domain: 'example.com' });
```

Fixes #1818

### DIFF
--- a/src/cdp/domains/network.zig
+++ b/src/cdp/domains/network.zig
@@ -19,6 +19,7 @@
 const std = @import("std");
 const lp = @import("lightpanda");
 const Allocator = std.mem.Allocator;
+const log = @import("../../log.zig");
 
 const CdpStorage = @import("storage.zig");
 
--- a/src/cdp/domains/storage.zig
+++ b/src/cdp/domains/storage.zig
@@ -132,7 +132,6 @@ pub fn setCdpCookie(cookie_jar: *CookieJar, param: CdpCookie) !void {
     // This allows Puppeteer's page.setCookie() to work, which may send cookies with
     // partitionKey as part of its cookie-setting workflow.
     if (param.partitionKey != null) {
-        const log = @import("../../log.zig");
         log.debug(.storage, "partitionKey ignored in setCdpCookie", .{});
     }
     // Still reject unsupported features


src/cdp/domains/network.zig
```diff
@@ -117,7 +117,12 @@ fn deleteCookies(cmd: anytype) !void {
         path: ?[]const u8 = null,
         partitionKey: ?CdpStorage.CookiePartitionKey = null,
     })) orelse return error.InvalidParams;
-    if (params.partitionKey != null) return error.NotImplemented;
+    // Silently ignore partitionKey since we don't support partitioned cookies (CHIPS).
+    // This allows Puppeteer's page.setCookie() to work, which sends deleteCookies
+    // with partitionKey as part of its cookie-setting workflow.
+    if (params.partitionKey != null) {
+        log.debug(.network, "partitionKey ignored in deleteCookies", .{});
```
karlseguin, line 125:
```suggestion
        lp.log.info(.not_implemented, "cookie partitionKey", .{.method = "network.deleteCookies" });
```

src/cdp/domains/storage.zig
```diff
@@ -128,7 +128,15 @@ pub const CdpCookie = struct {
 };
 
 pub fn setCdpCookie(cookie_jar: *CookieJar, param: CdpCookie) !void {
-    if (param.priority != .Medium or param.sameParty != null or param.sourceScheme != null or param.partitionKey != null) {
+    // Silently ignore partitionKey since we don't support partitioned cookies (CHIPS).
+    // This allows Puppeteer's page.setCookie() to work, which may send cookies with
+    // partitionKey as part of its cookie-setting workflow.
+    if (param.partitionKey != null) {
+        const log = @import("../../log.zig");
```
karlseguin: Pretty sure `log` is already imported in this file, so just the 1 line:

```zig
log.info(.not_implemented, "cookie partitionKey", .{.method = "storage.setCdpCookie" });
```

src/cdp/domains/network.zig
hobostay: Done, thanks for the suggestion!

src/cdp/domains/storage.zig
hobostay: Done, thanks for the suggestion!

PR #1824, opened 2026-03-14
fix: resolve memory leak in Option.getText() by using page arena

sjhddh:
This PR fixes a minor memory leak when calling `getText()` on `<option>` elements and resolves an existing `TODO: use proper allocator` in `src/browser/webapi/element/html/Option.zig`.

Previously, `std.heap.page_allocator` was used and never freed (e.g. when called inside `SemanticTree.zig`). By introducing `page: *Page` to the signature of `getText` and using `page.call_arena`, the memory lifecycle is safely scoped and collected, keeping headless usage slim.

- Fixes memory leak on Option text extraction
- Eliminates the TODO in `Option.zig`
- Auto-injected parameter in JS bridge logic means no JS engine wiring needs updating

### DIFF
--- a/src/SemanticTree.zig
+++ b/src/SemanticTree.zig
@@ -231,7 +231,7 @@ fn extractSelectOptions(node: *Node, page: *Page, arena: std.mem.Allocator) ![]O
         if (child.is(Element)) |el| {
             if (el.getTag() == .option) {
                 if (el.is(Element.Html.Option)) |opt| {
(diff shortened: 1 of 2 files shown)


sjhddh:

karlseguin:
Thanks! I'm glad `c_allocator` and `page_allocator` don't show up anywhere else they shouldn't!

PR #1825, opened 2026-03-14
fix: only increment TrackingAllocator reallocation_count on successful resizes

sjhddh:
Fixes #1820 

This small commit resolves the `// TODO: only if result is not null?` note in `TrackingAllocator` (`src/test_runner.zig`). 

Previously, `self.reallocation_count += 1` was incremented unconditionally after attempting to resize/remap, causing failed allocations to still count towards successful reallocations. This guards the increment behind a success check (`if (result)` for `rawResize` and `if (result != null)` for `rawRemap`).


sjhddh:

PR #1827, opened 2026-03-14
fix(dom): return parsererror document on XML parse failure

sjhddh:
This PR resolves a `TODO` in `src/browser/webapi/DOMParser.zig` regarding XML parsing behavior.

According to the [DOMParser spec](https://html.spec.whatwg.org/multipage/dynamic-markup-insertion.html#dom-domparser-parsefromstring), when `parseFromString` is called with an XML mime type and the parsing fails (or returns an empty document without a root element), it should not throw an exception. Instead, it must return an `XMLDocument` containing a `<parsererror>` element.

### Changes
- When `parser.err != null` or `doc_node.firstChild() == null` during XML parsing, we now generate and return a fallback `XMLDocument` containing the proper `<parsererror xmlns="http://www.mozilla.org/newlayout/xml/parsererror.xml">error</parsererror>` markup rather than throwing `error.JsException`.

### DIFF
--- a/src/browser/webapi/DOMParser.zig
+++ b/src/browser/webapi/DOMParser.zig
@@ -86,15 +86,15 @@ pub fn parseFromString(
             var parser = Parser.init(arena, doc_node, page);
             parser.parseXML(html);
 
-            if (parser.err) |pe| {
-                return pe.err;
+            if (parser.err != null or doc_node.firstChild() == null) {
+                // Return a document with a <parsererror> element per spec.
+                const err_doc = try page._factory.document(XMLDocument{ ._proto = undefined });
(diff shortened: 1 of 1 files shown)


sjhddh:

karlseguin:
Thanks. Can you update the test?   This case https://github.com/lightpanda-io/browser/blob/a9b9cf14c375bb0106927cef3580d95de3e1c3a5/src/browser/tests/domparser.html#L13  should be replaced with:

```js
  {
    const parser = new DOMParser();
    let d = parser.parseFromString('', 'text/xml');
    testing.expectEqual('<parsererror>error</parsererror>', new XMLSerializer().serializeToString(d));
  }
```

PR #1828, opened 2026-03-14
fix(net): fetch() should reject with a TypeError on network errors

sjhddh:
### Issue
When `fetch()` encounters a network error (e.g. DNS failure, connection refused, or aborted request), the `httpErrorCallback` in `src/browser/webapi/net/Fetch.zig` was rejecting the promise with the raw Zig `@errorName(err)` string.

This meant the caught exception inside JS was `undefined` or a raw string instead of an actual JS `Error` object, violating the Fetch specification which states that network errors should result in a `TypeError`.

### Fix
Updated `httpErrorCallback` to use `ls.local.isolate.createTypeError("Failed to fetch")`, ensuring that JS `catch (e)` blocks receive a properly structured `TypeError` instance with standard properties (`e.name`, `e.message`).

### DIFF
--- a/src/browser/tests/net/fetch.html
+++ b/src/browser/tests/net/fetch.html
@@ -203,3 +203,15 @@
     testing.expectEqual(true, response.body !== null);
   });
 </script>
+
+<script id=error>
+  testing.async(async (restore) => {
+    try {
+      await fetch('http://198.51.100.1:0');
+      testing.expectTrue(false); // force failure
+    } catch (err) {
+      restore();
+      testing.expectTrue(err instanceof TypeError)
+    }
+  });
+</script>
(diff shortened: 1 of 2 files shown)


sjhddh:

karlseguin:
This has come up a few times before, so I added a helper. If you rebase, you should be able to use:

```zig
ls.toLocal(self._resolver).rejectError("fetch error", .{.type_error = @errorName(err)});
```

and, at the end of `src/browser/tests/net/fetch.html`, you could add this test case:

```js
<script id=error>
  testing.async(async (restore) => {
    try {
      await fetch('http://198.51.100.1:0');
      testing.expectTrue(false); // force failure
    } catch (err) {
      restore();
      testing.expectTrue(err instanceof TypeError)
    }
  });
</script>```

PR #1831, opened 2026-03-14
Update README.md

juztshotz:
Just changed the grammar to be to my liking.


karlseguin:
Hi @juztshotz , the small change is fine, but we ask that you agree the the CLA as noted in the comment above.

PR #1836, opened 2026-03-14
fix: serialize AXValue integer as string per CDP spec

### LINKED ISSUE #1822: AXValue.value serialized as integer instead of string for type 'integer'
## Summary

When calling `Accessibility.getFullAXTree` via CDP, `AXValue.value` is returned as a JSON integer when `type` is `"integer"`, but the CDP spec defines it as always being a string (specifically a serialized remote value).

## Reproduction

1. Start lightpanda: `lightpanda serve --host 127.0.0.1 --port 9222`
2. Connect via WebSocket, create a target for `https://example.com`
3. Attach to the target and call `Accessibility.getFullAXTree`
4. Observe the heading node (nodeId 6):

**Lightpanda returns:**
```json
{"name": "level", "value": {"type": "integer", "value": 1}}
```

**Chrome returns (and CDP spec expects):**
```json
{"name": "level", "value": {"type": "integer", "value": "1"}}
```

## Impact

CDP clients with strict deserialization (e.g. Rust-based clients using serde) fail with:
```
Failed to deserialize CDP response for Accessibility.getFullAXTree: invalid type: integer `2`, expected a string
```

This breaks tools like [agent-browser](https://www.npmjs.com/package/agent-browser) which recently added `--engine lightpanda` support.

## Expected Behavior

`AXValue.value` should be serialized as a string regardless of the `type` field, matching Chrome's behavior per the [CDP Accessibility spec](https://chromedevtools.github.io/devtools-protocol/tot/Accessibility/#type-AXValue).

## Environment

- lightpanda nightly (installed via `pkg.lightpanda.io/install.sh`)
- macOS arm64

mvanhorn:
## Summary

Serialize `AXValue.value` as a JSON string when the type is `"integer"`, matching Chrome's behavior per the [CDP Accessibility spec](https://chromedevtools.github.io/devtools-protocol/tot/Accessibility/#type-AXValue).

## Why this matters

[#1822](https://github.com/lightpanda-io/browser/issues/1822) - CDP clients with strict deserialization (e.g., Rust serde) fail because `AXValue.value` is returned as a JSON integer (`1`) instead of a string (`"1"`) when `type` is `"integer"`.

**Before:** `{"name": "level", "value": {"type": "integer", "value": 1}}`
**After:** `{"name": "level", "value": {"type": "integer", "value": "1"}}`

## Changes

In `writeAXValue` (`src/cdp/AXNode.zig`), the `integer` variant now uses `std.fmt.bufPrint` to convert the `usize` to a string before writing, instead of relying on the `inline else` catch-all which writes it as a JSON number.

Also adds a test case verifying the h1 heading's `level` property is serialized as a string.

## Testing

- Extended the existing "AXNode: writer" test to find the h1 heading node and verify its level property value is the string `"1"` (not the integer `1`)
- `zig fmt --check` passes

Fixes #1822

### DIFF
--- a/src/cdp/AXNode.zig
+++ b/src/cdp/AXNode.zig
@@ -230,6 +230,7 @@ pub const Writer = struct {
         switch (value) {
             .integer => |v| {
                 // CDP spec requires integer values to be serialized as strings.
+                // 20 bytes is enough for the decimal representation of a 64-bit integer.
                 var buf: [20]u8 = undefined;
                 const s = try std.fmt.bufPrint(&buf, "{d}", .{v});
                 try w.write(s);


mvanhorn:

arrufat:
Thank you for the PR :)

src/cdp/AXNode.zig
```diff
@@ -228,6 +228,13 @@ pub const Writer = struct {
 
         try w.objectField("value");
         switch (value) {
+            .integer => |v| {
+                // CDP spec requires integer values to be serialized as strings.
+                // 20 bytes is enough for the decimal representation of a 64-bit integer.
```
arrufat, line 233: I just clarified this
```sh
echo "length(2^64-1)" | bc  # 20
```

PR #1837, opened 2026-03-14
Implement charset detection from first 1024 bytes of HTML

### LINKED ISSUE #531: Implement charset detection from the first 1024 bytes of the HTML
**charset**

In [`browser.zig`](https://github.com/lightpanda-io/browser/blob/main/src/browser/browser.zig#L420), in case of document HTML, we should try to determine the `charset` from a `meta` tag in the first 1024 bytes of the document.

> The meta element can be used, and the charset attribute is preferred [html5:0]. If there is no HTTP declaration or BOM, a meta element must be used [html5:14]. Any meta declaration must use an ascii-compatible encoding [html5:14] [html5:16]. The implication of this is that UTF-16 encoded pages must not use a meta declaration. Any meta declaration must fit in the first 1024 bytes of page [html5:12] [html5:23].
https://www.w3.org/International/articles/spec-summaries/encoding
https://www.w3.org/International/questions/qa-html-encoding-declarations

If we found no charset, we should `mime.charset` and finally `utf-8` by default.

mvanhorn:
## Summary

Implements charset detection by prescanning the first 1024 bytes of HTML documents for `<meta charset>` declarations, as specified by the [W3C encoding guidelines](https://www.w3.org/International/questions/qa-html-encoding-declarations).

## Why this matters

[#531](https://github.com/lightpanda-io/browser/issues/531) - When the HTTP Content-Type header doesn't specify a charset, the browser should fall back to detecting it from `<meta>` tags in the first 1024 bytes of the document. Without this, non-UTF-8 pages served without an explicit HTTP charset header may be decoded incorrectly.

## Changes

**`src/browser/Mime.zig`:**
- Added `prescanCharset()` - scans first 1024 bytes of HTML for:
  - `<meta charset="X">`
  - `<meta http-equiv="Content-Type" content="...;charset=X">`
- Case-insensitive tag and attribute matching
- Helper functions: `findAttrValue`, `extractCharsetFromContentType`, `asciiEqlIgnoreCase`
- Unit tests covering both meta formats, case insensitivity, missing charset, and the 1024-byte limit

**`src/browser/Page.zig`:**
- In `pageDataCallback`, when the HTTP header doesn't specify a charset and the content is HTML, runs the prescan on the first data chunk and updates `mime.charset` with the detected value

## Testing

- Added "Mime: prescanCharset" test with 10 assertions
- `zig fmt --check` passes

Fixes #531

### DIFF
--- a/src/browser/Mime.zig
+++ b/src/browser/Mime.zig
@@ -25,6 +25,7 @@ params: []const u8 = "",
 // We keep 41 for null-termination since HTML parser expects in this format.
 charset: [41]u8 = default_charset,
 charset_len: usize = default_charset_len,
+is_default_charset: bool = true,
 
 /// String "UTF-8" continued by null characters.
 const default_charset = .{ 'U', 'T', 'F', '-', '8' } ++ .{0} ** 36;
@@ -130,6 +131,7 @@ pub fn parse(input: []u8) !Mime {
 
     var charset: [41]u8 = default_charset;
     var charset_len: usize = default_charset_len;
+    var has_explicit_charset = false;
 
     var it = std.mem.splitScalar(u8, params, ';');
     while (it.next()) |attr| {
@@ -156,6 +158,7 @@ pub fn parse(input: []u8) !Mime {
                 // Null-terminate right after attribute value.
                 charset[attribute_value.len] = 0;
                 charset_len = attribute_value.len;
+                has_explicit_charset = true;
             },
         }
     }
@@ -165,6 +168,7 @@ pub fn parse(input: []u8) !Mime {
         .charset = charset,
         .charset_len = charset_len,
         .content_type = content_type,
+        .is_default_charset = !has_explicit_charset,
     };
 }
 
@@ -212,7 +216,7 @@ pub fn prescanCharset(html: []const u8) ?[]const u8 {
 
         // Look for http-equiv="content-type" with content="...;charset=X"
         if (findAttrValue(attrs, "http-equiv")) |he| {
-            if (asciiEqlIgnoreCase(he, "content-type")) {
+            if (std.ascii.eqlIgnoreCase(he, "content-type")) {
                 if (findAttrValue(attrs, "content")) |content| {
                     if (extractCharsetFromContentType(content)) |charset| {
                         return charset;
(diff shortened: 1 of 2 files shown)


mvanhorn:

src/browser/Mime.zig
```diff
@@ -168,6 +168,137 @@ pub fn parse(input: []u8) !Mime {
...
+    while (it.next()) |part| {
+        const trimmed = std.mem.trimLeft(u8, part, &.{ ' ', '\t' });
+        if (trimmed.len > 8 and asciiEqlIgnoreCase(trimmed[0..8], "charset=")) {
+            const val = std.mem.trim(u8, trimmed[8..], &.{ ' ', '\t', '"', '\'' });
+            if (val.len > 0 and val.len <= 40) return val;
+        }
+    }
+    return null;
+}
+
```
karlseguin, line 301: You can use `std.ascii.eqlIgnoreCase`

src/browser/Mime.zig
```diff
@@ -168,6 +168,137 @@ pub fn parse(input: []u8) !Mime {
...
+        // Check for "meta" (case-insensitive)
+        if (pos + 4 >= data.len) return null;
+        var tag_buf: [4]u8 = undefined;
+        _ = std.ascii.lowerString(&tag_buf, data[pos..][0..4]);
+        if (!std.mem.eql(u8, &tag_buf, "meta")) {
+            continue;
+        }
+        pos += 4;
+
+        // Must be followed by whitespace or end of tag
```
karlseguin, line 200: This case causes an endless loop:

```
try testing.expectEqual(null, Mime.prescanCharset("<meta foo=\"bar\"/>"));
```

src/browser/Page.zig
```diff
@@ -848,13 +848,25 @@ fn pageDataCallback(transfer: *HttpClient.Transfer, data: []const u8) !void {
...
+        var mime: Mime = blk: {
             if (transfer.response_header.?.contentType()) |ct| {
                 break :blk try Mime.parse(ct);
             }
             break :blk Mime.sniff(data);
         } orelse .unknown;
 
+        // If the HTTP header didn't specify a charset and this is HTML,
+        // prescan the first 1024 bytes for a <meta charset> declaration.
+        if (mime.content_type == .text_html and std.mem.eql(u8, mime.charsetString(), "UTF-8")) {
```
karlseguin: If the charset is explicitly set by the content type, than this can override it. We'll need a flag on mime, something like: `is_default_charset: bool` and only if this is true should come in here.

mvanhorn:
Addressed all three items in b373fb4:

1. Replaced custom `asciiEqlIgnoreCase` with `std.ascii.eqlIgnoreCase`
2. Fixed the infinite loop - `findAttrValue` now advances past tokens that have no `=` sign (added regression test for `<meta foo="bar"/>`)
3. Added `is_default_charset: bool` to Mime. Set to `false` when `parse()` finds an explicit charset in Content-Type. Page.zig now checks `mime.is_default_charset` instead of comparing the charset string.

All 358 tests pass.

src/browser/Mime.zig
mvanhorn: Done - removed the custom function entirely.

src/browser/Mime.zig
mvanhorn: Fixed. When findAttrValue encounters a token without `=`, it now advances at least one byte before continuing. Added the exact test case you provided.

src/browser/Page.zig
mvanhorn: Added `is_default_charset: bool` field to Mime (defaults to `true`). `parse()` sets it to `false` when it finds an explicit charset attribute. Page.zig now gates the prescan on `mime.is_default_charset`.

karlseguin:
Thank you very much!

PR #1845, opened 2026-03-15
fix(cdp): don't kill WebSocket on unknown domain/method errors

### LINKED ISSUE #1843: Unrecognized CDP method corrupts connection state, crashes subsequent commands
## Summary

Sending a CDP command with a non-existent method name (e.g., `NonExistent.method`) correctly returns a CDP error response. However, it also corrupts the server's internal state — subsequent *valid* CDP commands (like `Network.enable` or `Page.navigate`) crash the WebSocket connection.

## Reproduction Steps

```
1. Target.createTarget → { url: "about:blank" }
2. Target.attachToTarget → { targetId: ..., flatten: true }
3. Page.navigate → "http://localhost:3000/some_page"   // works fine
4. NonExistent.method → {}
   // Returns error response (correct behavior ✓)
5. Page.navigate → "http://localhost:3000/other_page"
   // → WebSocket dies: Connection reset by peer  ✗
```

Step 4 returns the error correctly. But step 5 (or any subsequent command) crashes the connection.

## Expected Behavior

After receiving an error response for an unknown method, the connection should remain healthy. Subsequent valid commands should work normally. The unknown method should be a no-op beyond the error response.

## Actual Behavior

The error response is sent, but internal server state is corrupted. The next CDP command that requires actual processing (navigation, network enable, etc.) causes `ECONNRESET`.

## Impact

This affects **error resilience** of CDP clients. In a robust client, typos or version mismatches in method names should produce an error that the client can handle and recover from — not kill the entire session. Currently, any single bad method call requires a full browser restart.

## Notes

- Simple commands like `Target.getTargets` may still work after the bad method call — the corruption seems to affect page-session-scoped operations specifically
- The issue is consistent across multiple sessions — not a race condition

## Environment

- Lightpanda nightly (macOS aarch64)  
- CDP serve mode via WebSocket
- Discovered while building [capybara-lightpanda](https://github.com/navidemad/capybara-lightpanda) driver

navidemad:
## Summary

- Fixes the WebSocket connection dying when a CDP command with an unrecognized domain (e.g. `NonExistent.method`) is sent
- The error response was sent correctly, but `dispatch()` re-returned the error which propagated up to `handleMessage()` where `catch return false` killed the connection
- Now the error is only propagated if `sendError` itself fails; otherwise the read loop continues

Fixes #1843

## Test plan

- Updated existing `"cdp: invalid json"` test to verify `InvalidMethod`, `UnknownDomain`, and `UnknownMethod` all return error responses without propagating fatal errors

### DIFF
--- a/src/cdp/domains/dom.zig
+++ b/src/cdp/domains/dom.zig
@@ -550,11 +550,12 @@ test "cdp.dom: getSearchResults unknown search id" {
     var ctx = testing.context();
     defer ctx.deinit();
 
-    try testing.expectError(error.BrowserContextNotLoaded, ctx.processMessage(.{
+    try ctx.processMessage(.{
         .id = 8,
         .method = "DOM.getSearchResults",
         .params = .{ .searchId = "Nope", .fromIndex = 0, .toIndex = 10 },
-    }));
+    });
+    try ctx.expectSentError(-31998, "BrowserContextNotLoaded", .{ .id = 8 });
 }
 
 test "cdp.dom: search flow" {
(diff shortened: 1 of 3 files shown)


navidemad:

karlseguin:
I think this is a good idea, @krichprollsch ?

krichprollsch:
I agree with the idea 👍

karlseguin:
Thanks!

PR #1850, opened 2026-03-15
Fix CDP WebSocket connection dying during complex page navigation

### LINKED ISSUE #1849: CDP serve mode: WebSocket connection dies during navigation of complex pages with JS bundles
## Summary

When using Lightpanda in CDP serve mode to navigate to a page served by an in-process HTTP server (e.g., Puma started within the same Ruby process — the standard Capybara test setup), navigation hangs indefinitely if the HTML page references external sub-resources (`<script>`, `<link>`, etc.) served by the same origin.

The browser emits `Page.frameStartedNavigating` and `Page.frameStartedLoading` events but never progresses to `Page.frameNavigated`, `Page.domContentEventFired`, or `Page.loadEventFired`. The CDP WebSocket connection eventually closes or the client times out.

## Environment

- **Lightpanda version**: `be8ba532` (nightly, downloaded 2026-03-14)
- **Platform**: macOS arm64 (Darwin 25.3.0)
- **HTTP server**: Puma 7.2.0 (in-process, 4 threads)
- **Use case**: Capybara system tests with `capybara-lightpanda` gem

### What works

| Scenario | Result |
|----------|--------|
| Navigate to external site (example.com, httpbin.org) | ✅ |
| Navigate to in-process Puma, page has **no** sub-resources | ✅ |
| Navigate to in-process Puma, page has small/large `<script src>` | ✅ |
| Navigate to in-process Puma, page has `<script type="module">` | ✅ |
| Navigate to in-process Puma, page has multiple sub-resources (CSS + JS) | ✅ |
| Navigate to in-process Puma, page with gzip-compressed HTML/JS | ✅ |
| `lightpanda fetch` mode to in-process Puma (full Rails page, 345KB) | ✅ |

### What fails

| Scenario | Result |
|----------|--------|
| CDP serve → in-process Puma serving a **full Rails application page** (~345KB HTML with Turbo/Stimulus JS bundles ~1.1MB) | ❌ Navigation never completes |

## Detailed timeline (from CDP debug logs)

```
t=0.000s  Page.navigate sent to http://127.0.0.1:55918/scan-to-pay/...
t=0.000s  ◀ Page.frameStartedNavigating
t=0.000s  ◀ Page.frameStartedLoading
          ... 30 second gap, no CDP events ...
t=30.0s   Client times out, restarts browser
```

## Lightpanda stderr during the hang

```
$scope=page $level=info $msg=navigate url=http://127.0.0.1:PORT/... method=GET reason=address_bar body=false req_id=1 type=root
$scope=browser $level=info $msg="executing script" src=http://127.0.0.1:PORT/ kind=javascript cacheable=false
$scope=browser $level=info $msg="executing script" src=http://127.0.0.1:PORT/ kind=javascript cacheable=false
$scope=browser $level=info $msg="executing script" src=http://127.0.0.1:PORT/assets/public-HASH.js kind=module cacheable=true
```

After the `"executing script"` log for the module JS bundle, **no further output**. Lightpanda appears to block while loading or executing the module script.

## Key observations

1. **The HTTP server receives and completes the HTML request** — confirmed via both request logging in Puma and independent `curl`/`Net::HTTP` requests to the same Capybara server during the hang.

2. **The JS bundle request may or may not reach Puma** — Lightpanda starts processing inline scripts from the HTML, then attempts to load the external module script. The Lightpanda log confirms it reaches the `executing script` phase for the module URL.

3. **The same page loads fine in `fetch` mode** — `lightpanda fetch --dump html http://127.0.0.1:PORT/...` returns the full 345KB HTML successfully.

4. **Simple reproduction pages work** — We tested with synthetic pages (up to 1.1MB JS bundles, multiple sub-resources, gzip compression, chunked transfer, module scripts) and all passed. The issue appears specific to the combination of a complex real-world page.

5. **The page characteristics**: ~345KB HTML, 2 inline `<script>` tags, 1 `<script type="module" src="/assets/public-HASH.js">` (1.1MB esbuild bundle), 1 `<link rel="stylesheet">`, various `<meta>` tags including Turbo/Stimulus markers, `<turbo-cable-stream-source>` custom elements.

## Suspected cause

The failure seems related to the interaction between Lightpanda's async I/O event loop and the processing of complex pages with large module scripts. When the page HTML is parsed and triggers a sub-resource fetch for a large JS module from the same origin, something in Lightpanda's internal scheduling may deadlock or starve:

- The CDP WebSocket connection and the HTTP client share the same event loop
- While processing a complex page (executing inline scripts, fetching sub-resources), the event loop may not service the CDP WebSocket
- This prevents the client from sending any commands (including `Runtime.evaluate` for readyState polling)

The fact that `fetch` mode works but `serve` mode doesn't supports this theory — `fetch` mode doesn't need to simultaneously handle a CDP WebSocket.

## Reproduction

We have a Ruby reproduction script that works with simple pages. The full reproduction requires a Rails application with:
- Propshaft asset pipeline
- A ~1MB esbuild JS bundle loaded as `<script type="module">`
- Puma running in-process (as Capybara does)

If helpful, we can provide the exact HTML page content that triggers the issue.

## Workaround

None known. The `capybara-lightpanda` gem already sends `Page.navigate` asynchronously and polls `document.readyState` as a fallback, but both approaches fail because Lightpanda stops processing CDP commands during the hang.

## Related

This may be related to how Lightpanda handles concurrent I/O between the CDP WebSocket server and outgoing HTTP requests in serve mode. The Zig async I/O model may need to yield/service the CDP connection while waiting for HTTP sub-resource responses.

navidemad:
## Summary

- Fix unit mismatch in CDP timeout handler: `timestamp(.monotonic)` returns **seconds** but `ms_remaining` is in **milliseconds**, causing incorrect timeout arithmetic
- Fix double-counting: `last_message` was never updated in the `.done` branch, so `elapsed` grew as absolute time rather than incremental, draining `ms_remaining` quadratically
- During complex page loads (large HTML + JS bundles), `Session._wait()` returns `.done` rapidly, triggering the bug and killing the WebSocket in ~2 seconds instead of the configured 10s timeout

Fixes #1849

## Test plan

- [ ] `Page.navigate` to complex pages with sub-resource fetches should not kill the CDP WebSocket
- [ ] Full navigation lifecycle completes: `frameStartedNavigating` → `frameNavigated` → `domContentEventFired` → `loadEventFired`
- [ ] Simple pages continue to work as before
- [ ] `lightpanda fetch <url>` mode unaffected (code path not exercised)

### DIFF
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -295,7 +295,7 @@ pub const Client = struct {
         }
 
         var cdp = &self.mode.cdp;
-        var last_message = timestamp(.monotonic);
+        var last_message = milliTimestamp(.monotonic);
         var ms_remaining = self.ws.timeout_ms;
 
         while (true) {
@@ -304,7 +304,7 @@ pub const Client = struct {
                     if (self.readSocket() == false) {
                         return;
                     }
-                    last_message = timestamp(.monotonic);
+                    last_message = milliTimestamp(.monotonic);
                     ms_remaining = self.ws.timeout_ms;
                 },
                 .no_page => {
@@ -319,16 +319,18 @@ pub const Client = struct {
                     if (self.readSocket() == false) {
                         return;
                     }
-                    last_message = timestamp(.monotonic);
+                    last_message = milliTimestamp(.monotonic);
                     ms_remaining = self.ws.timeout_ms;
                 },
                 .done => {
-                    const elapsed = timestamp(.monotonic) - last_message;
-                    if (elapsed > ms_remaining) {
+                    const now = milliTimestamp(.monotonic);
+                    const elapsed = now - last_message;
+                    if (elapsed >= ms_remaining) {
                         log.info(.app, "CDP timeout", .{});
                         return;
                     }
                     ms_remaining -= @intCast(elapsed);
+                    last_message = now;
                 },
             }
         }
@@ -501,6 +503,7 @@ fn buildJSONVersionResponse(
 }
 
 pub const timestamp = @import("datetime.zig").timestamp;
+pub const milliTimestamp = @import("datetime.zig").milliTimestamp;
 
 const testing = std.testing;
 test "server: buildJSONVersionResponse" {


karlseguin:
Thanks

PR #1860, opened 2026-03-16
fix(cdp): stub Emulation.setUserAgentOverride to unblock Playwright

### LINKED ISSUE #1436: Support for modifying User-Agent?
Hi,

I’m currently testing Lightpanda as a headless browser via CDP.

I would like to ask whether Lightpanda supports modifying the User-Agent string, either:

- via a startup flag (for example: --user-agent), or  
- through CDP, such as using Network.setUserAgentOverride.

It appears that Lightpanda does not currently support the `Emulation.setUserAgentOverride` CDP method, which Playwright uses internally when a custom User-Agent is configured.
```
 playwright: Protocol error (Emulation.setUserAgentOverride): UnknownMethod
```
Could you please confirm whether:

- `Emulation.setUserAgentOverride` is intentionally unsupported?
- or if User-Agent customization is only possible via startup flags?

Thank you for your work on Lightpand

jnMetaCode:
Fixes #1436

## Problem

Playwright calls `Emulation.setUserAgentOverride` when creating a browser context with device emulation or a custom user agent. Lightpanda returns `UnknownMethod`, which crashes the Playwright driver:

```
playwright: Protocol error (Emulation.setUserAgentOverride): UnknownMethod
```

## Solution

Add a noop handler for `setUserAgentOverride`, matching the existing pattern for all other Emulation methods (`setDeviceMetricsOverride`, `setEmulatedMedia`, `setFocusEmulationEnabled`, `setTouchEmulationEnabled`) which are all stubs marked `// TODO: noop method`.

This doesn't implement actual user agent override — it just prevents the CDP handshake from failing so Playwright sessions can proceed.

### Note

Actual user agent override support is a separate discussion (see #1436 thread). This PR only prevents the crash so Playwright can be used with Lightpanda without error.

### DIFF
--- a/src/cdp/domains/emulation.zig
+++ b/src/cdp/domains/emulation.zig
@@ -24,6 +24,7 @@ pub fn processMessage(cmd: anytype) !void {
         setFocusEmulationEnabled,
         setDeviceMetricsOverride,
         setTouchEmulationEnabled,
+        setUserAgentOverride,
     }, cmd.input.action) orelse return error.UnknownMethod;
 
     switch (action) {
@@ -31,6 +32,7 @@ pub fn processMessage(cmd: anytype) !void {
         .setFocusEmulationEnabled => return setFocusEmulationEnabled(cmd),
         .setDeviceMetricsOverride => return setDeviceMetricsOverride(cmd),
         .setTouchEmulationEnabled => return setTouchEmulationEnabled(cmd),
+        .setUserAgentOverride => return setUserAgentOverride(cmd),
     }
 }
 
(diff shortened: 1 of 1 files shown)


jnMetaCode:

karlseguin:
@krichprollsch Do you see any problems with this? It isn't uncommon for us to no-op a message, but could this one lead to surprises? People set a user-agent, we say "ok., then ignore it?

Maybe we should log

```zig
const log = @import("../../log.zig");

...

log.warn(.not_implemented, "setUserAgentOverride", .{});
```

krichprollsch:
The advantage of having a noop is to be compatible w/ more existing script.
I think you are right, a log message is important here :+1: 
But maybe with an `.ignore` more than an `.not_implemented` b/c we explicitely don't want to implement it...

PR #1862, opened 2026-03-16
fix(dom): add default messages for all DOMException error codes

### LINKED ISSUE #82: Add message for all DOMExceptions
Messages are not standardized between existing browsers.

Even for one browser message can be different regarding the context of the error.

jnMetaCode:
Fixes #82

## Problem

`DOMException.getMessage()` only had messages for 6 of 18 error codes. The remaining 12 fell through to the `else` branch which returned `@tagName(self._code)`, producing underscore_case strings like `"wrong_document_error"` instead of human-readable messages.

## Solution

Fill in all 18 error codes with descriptive messages based on the [WebIDL spec error name descriptions](https://webidl.spec.whatwg.org/#idl-DOMException-error-names). The `else => @tagName(self._code)` fallback is no longer needed since all enum variants are now covered.

### Before
```
wrong_document_error
invalid_state_error
```

### After
```
The object is in the wrong document
The object is in an invalid state
```

### DIFF
--- a/src/browser/webapi/DOMException.zig
+++ b/src/browser/webapi/DOMException.zig
@@ -104,13 +104,27 @@ pub fn getMessage(self: *const DOMException) []const u8 {
     }
     return switch (self._code) {
         .none => "",
-        .invalid_character_error => "Invalid Character",
         .index_size_error => "Index or size is negative or greater than the allowed amount",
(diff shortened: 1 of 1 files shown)


jnMetaCode:

PR #1863, opened 2026-03-16
fix(cdp): add missing disable methods to Performance, Inspector, Security domains

jnMetaCode:
## Problem

The `Performance`, `Inspector`, and `Security` CDP domains only implement `enable` but not `disable`. CDP clients (Playwright, Puppeteer) call `*.disable` during session cleanup, which returns `UnknownMethod` and can cause the connection to fail.

## Solution

Add `disable` as a noop handler to all three domains, matching the existing `enable` pattern. This is consistent with how other CDP browsers handle these methods — `disable` simply acknowledges the request without side effects.

### Changes

- **performance.zig**: add `disable` to enum and switch
- **inspector.zig**: add `disable` to enum and switch
- **security.zig**: add `disable` to enum and switch

### DIFF
--- a/src/cdp/domains/security.zig
+++ b/src/cdp/domains/security.zig
@@ -21,11 +21,13 @@ const std = @import("std");
 pub fn processMessage(cmd: anytype) !void {
     const action = std.meta.stringToEnum(enum {
         enable,
+        disable,
         setIgnoreCertificateErrors,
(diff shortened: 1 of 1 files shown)


jnMetaCode:

karlseguin:
Thanks ! :)

PR #1865, opened 2026-03-16
fix: version command should print version number

### LINKED ISSUE #1835: version command should return the version number on stable release

JasonOA888:
## Problem
`lightpanda version` only prints the git commit hash:
```
fe3faa0a
```

Instead of the version number on stable releases.

## Solution
Print both version number and git commit:
```
v0.2.6 (fe3faa0a)
```

## Changes
- Modified `src/main.zig` to print `version` field from `build_config`
- Kept git commit for debugging purposes

Fixes #1835

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -59,7 +59,7 @@ fn run(allocator: Allocator, main_arena: Allocator) !void {
             return std.process.cleanExit();
         },
         .version => {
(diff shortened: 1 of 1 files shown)


mrdimidium:
Duplicate https://github.com/lightpanda-io/browser/pull/1867 from same author.

PR #1867, opened 2026-03-16
fix: use git tag version for stable releases

### LINKED ISSUE #1835: version command should return the version number on stable release

JasonOA888:
## Problem
Fixes #1835 - `lightpanda version` returns git commit hash (e.g., `_fe3faa0a_`) instead of version number (e.g., `v0.2.6`) on stable releases.

## Root Cause
The version is read from `build.zig.zon` which always contains `"0.0.0"`. This is not updated during release builds.

## Solution
Modified `build.zig` to detect git tags at build time:

```zig
// Try to get version from git tag (e.g., v0.2.6), fallback to manifest.version
const git_version: ?[]const u8 = blk: {
    if (b.enable_cache) break :blk null;
    const git_describe = b.addSystemCommand(&.{ "git", "describe", "--tags", "--exact-match" });
    git_describe.captureStdOut();
    git_describe.expect_exit_code = 1; // Don't fail if not on a tag
    break :blk git_describe.captureStdOutOrNull();
};

const version = git_version orelse manifest.version;
```

## Behavior

| Build Type | `lightpanda version` Output |
|------------|----------------------------|
| Stable release (tagged) | `v0.2.6 (fe3faa0a)` |
| Dev build (no tag) | `0.0.0 (dev)` |
| CI build with commit | `0.0.0 (abc1234)` |

## Testing

1. Checkout a tagged commit: `git checkout v0.2.6`
2. Build: `zig build`
3. Run: `./zig-out/bin/lightpanda version`
4. Expected: `v0.2.6 (<commit>)`

## Files Changed
- `build.zig` (+11 lines)

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -30,8 +30,19 @@ pub fn build(b: *Build) !void {
     const prebuilt_v8_path = b.option([]const u8, "prebuilt_v8_path", "Path to prebuilt libc_v8.a");
     const snapshot_path = b.option([]const u8, "snapshot_path", "Path to v8 snapshot");
 
+    // Try to get version from git tag (e.g., v0.2.6), fallback to manifest.version
+    const git_version: ?[]const u8 = blk: {
+        if (b.enable_cache) break :blk null;
+        const git_describe = b.addSystemCommand(&.{ "git", "describe", "--tags", "--exact-match" });
+        git_describe.captureStdOut();
+        git_describe.expect_exit_code = 1; // Don't fail if not on a tag
+        break :blk git_describe.captureStdOutOrNull();
+    };
+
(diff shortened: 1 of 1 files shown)


JasonOA888:

mrdimidium:
Thanks for your contributing. This will work, but running external commands during the build process has some reproducible issues.

It would be better to add the `git_version` option, similar to [`git_commit`](https://github.com/lightpanda-io/browser/blob/main/build.zig#L29), and explicitly pass it to CI, as we do here: https://github.com/lightpanda-io/browser/blob/main/.github/workflows/build.yml#L48

PR #1872, opened 2026-03-16
Gracefull close ws socket

mrdimidium:
Fixes two bugs in WS connection handling:
- if a connection fails to be accepted, we should continue trying new connections; this is normal.
- when closing a WS socket at the browser's initiative, send a CLOSE message.

### DIFF
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -242,7 +242,10 @@ pub const Client = struct {
     fn stop(self: *Client) void {
         switch (self.mode) {
             .http => {},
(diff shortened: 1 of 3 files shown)


src/Server.zig
```diff
@@ -244,6 +244,7 @@ pub const Client = struct {
             .http => {},
             .cdp => |*cdp| cdp.browser.env.terminate(),
         }
+        self.ws.sendClose();
```
karlseguin: This should only be done with `mode == .cdp`?

PR #1874, opened 2026-03-16
feat: add git_version build option for release version detection

JasonOA888:
Fixes #1867

Addresses feedback from @mrdimidium:
- Added `git_version` build option (similar to existing `git_commit`)
- CI can pass `-Dgit_version=$(git describe --tags --exact-match)` for releases
- Falls back to `git_commit` when not on a tagged release

This avoids running external commands during build and improves reproducibility.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -27,12 +27,14 @@ pub fn build(b: *Build) !void {
     const manifest = Manifest.init(b);
 
     const git_commit = b.option([]const u8, "git_commit", "Current git commit");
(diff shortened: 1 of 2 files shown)


JasonOA888:
Updated approach based on @mrdimidium feedback:

1. Added `git_version` as a build option (similar to `git_commit`)
2. CI can pass the version explicitly: `-Dgit_version=$(git describe --tags --exact-match 2>/dev/null || echo "dev")`
3. `version` command outputs `git_version` if available, falls back to `git_commit`

This avoids running external commands during the build process and improves reproducibility.

PR #1878, opened 2026-03-17
Implement window.event property

### LINKED ISSUE #1770: Add support for `window.event`
https://developer.mozilla.org/en-US/docs/Web/API/Window/event

mvanhorn:
## Summary

Implements the `window.event` property per the [WHATWG spec](https://html.spec.whatwg.org/multipage/window-object.html#dom-window-event). Returns the Event currently being handled by an event handler, or `undefined` when no event is being dispatched.

## Why this matters

Many legacy scripts and libraries (jQuery, analytics snippets) depend on `window.event` to access the current event object. Without it, these scripts fail silently or throw. [MDN reference](https://developer.mozilla.org/en-US/docs/Web/API/Window/event).

## Changes

- **Window.zig**: Add `_current_event` field and `getEvent` accessor exposed via JsApi
- **EventManager.zig**: Save/restore `window._current_event` around handler invocation in both `dispatchDirect` and `dispatchNode`, supporting nested event dispatch
- **window_event.html**: Tests for undefined-outside-handler, set-during-handler, and restored-after-handler

## Testing

- `zig build test` - 360/360 tests pass
- New test covers: property is undefined outside handlers, equals the dispatched Event inside handlers, restored to undefined after handler completes

Fixes #1770

### DIFF
--- a/src/browser/EventManager.zig
+++ b/src/browser/EventManager.zig
@@ -233,6 +233,12 @@ const DispatchDirectOptions = struct {
 pub fn dispatchDirect(self: *EventManager, target: *EventTarget, event: *Event, handler: anytype, comptime opts: DispatchDirectOptions) !void {
     const page = self.page;
 
+    // Set window.event to the currently dispatching event (WHATWG spec)
+    const window = page.window;
+    const prev_event = window._current_event;
+    window._current_event = event;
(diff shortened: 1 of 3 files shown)


karlseguin:
Thanks. This is nice addition.

PR #1879, opened 2026-03-17
ScriptManager: fix memory leak and resource handover

arrufat:
Release the arena when an inline script is empty and ensure the handover flag is set correctly for all script execution modes.

### DIFF
--- a/src/browser/ScriptManager.zig
+++ b/src/browser/ScriptManager.zig
@@ -261,7 +261,6 @@ pub fn addFromElement(self: *ScriptManager, comptime from_parser: bool, script_e
             break :blk .normal;
         },
     };
-    handover = true;
 
     const is_blocking = script.mode == .normal;
     if (is_blocking == false) {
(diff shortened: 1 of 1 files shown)


src/browser/ScriptManager.zig
```diff
@@ -260,6 +261,7 @@ pub fn addFromElement(self: *ScriptManager, comptime from_parser: bool, script_e
             break :blk .normal;
         },
     };
+    handover = true;
```
karlseguin: handover specifically means that `self.client.request` was called.  If you set it before `client.request` is called, and `client.request` fails, then you won't cleanup properly. Im not sure what case this is handling, but if it's necessary, then I think we need to rollback on error:

```zig
self.client.request(.{...}) catch |err| {
  handover = false;
   return err;
}
```

arrufat: Instead of setting `handover = true` early and then manually rolling it back in a `catch` block if the request fails, how about just delay setting `handover = true` until after `client.request` successfully completes?

karlseguin: Yes, that's the original code?

arrufat: Yes, but the original code had a hidden double free on the error path (script.deinit()), right?

karlseguin: yes, removing the nested errdefer and keeping handover where it is, seems right.

PR #1882, opened 2026-03-17
fix: special-case Window#onerror per WHATWG spec (5-arg signature)

### LINKED ISSUE #1773: Special-case `Window#onerror`
This seem to take multiple arguments; I'm not sure if its preferred much, yet its better to note it down here.

https://developer.mozilla.org/en-US/docs/Web/API/Window/error_event

Tenith01:
Fixes #1773

### Description
This PR fixes the behavior of `window.onerror` to align with the WHATWG HTML specification for runtime script errors. 

Previously, `onerror` was being invoked with standard event arguments or firing incorrectly. Per [WHATWG HTML §8.1.7.6 (event handlers)](https://html.spec.whatwg.org/multipage/webappapis.html#handler-window-onerror), `window.onerror` is a special case that must be called with exactly five positional arguments: `message`, `source`, `lineno`, `colno`, and `error`. 

This update modifies the dispatch call in `src/browser/webapi/Window.zig` to intercept error events targeting the window and properly extract and pass these 5 arguments to the callback, suppressing the default browser error output if the handler returns `true`.

### Checklist
- [x] Update `Window.zig` to use the 5-argument signature for `onerror`
- [x] Ensure `return true` suppresses default event handling
- [x] Run `zig fmt`
- [x] Passed `zig build test -- --filter 'WebApi: Window.onerror'`

### DIFF
--- a/src/browser/tests/event/report_error.html
+++ b/src/browser/tests/event/report_error.html
@@ -1,0 +1,38 @@
+<!DOCTYPE html>
+<script src="../testing.js"></script>
+
+<script id=onerrorFiveArguments>
+  let called = false;
+  let argCount = 0;
+  window.onerror = function() {
+    called = true;
+    argCount = arguments.length;
+    return true; // suppress default
+  };
+  try { undefinedVariable; } catch(e) { window.reportError(e); }
+  testing.expectEqual(true, called);
+  testing.expectEqual(5, argCount);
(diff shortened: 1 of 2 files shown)


Tenith01:

karlseguin:
Thanks, this is a nice fix!

PR #1883, opened 2026-03-17
fix: show actionable error when server port is already in use

### LINKED ISSUE #1830: Lightpanda doesn't start because port is already used
Image @krichprollsch

Tenith01:
Fixes #1830

### Description
This PR addresses the issue where running `lightpanda serve` on an already occupied port results in a generic `server run error err=AddressInUse` exit. 

Instead of failing silently with an unhelpful system error, the server now specifically catches the `error.AddressInUse` condition during the bind phase in `src/network/Runtime.zig`. It outputs an actionable, user-friendly CLI message indicating the exact address that is blocked and provides a hint to free the port or change the configuration.

### Checklist
- [x] Catch `AddressInUse` error during POSIX bind
- [x] Provide actionable CLI output
- [x] Run `zig fmt`
- [x] Added/passed `zig build test -- --filter 'Runtime: bind fails when port is taken'`

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -94,7 +94,16 @@ fn run(allocator: Allocator, main_arena: Allocator) !void {
             };
 
             var server = lp.Server.init(app, address) catch |err| {
-                log.fatal(.app, "server run error", .{ .err = err });
+                if (err == error.AddressInUse) {
+                    log.fatal(.app, "address already in use", .{
+                        .host = opts.host,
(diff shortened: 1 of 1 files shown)


Tenith01:

karlseguin:
This looks good to me. Tested and works. Not sure it "fixes" #1830 but I'm not sure what else we can do except have a better message, which this does.

Issue #1869, opened 2026-03-16
Cloudflare Turnstile widget fails with error 300030 ("Widget hung")


kjozsa:
When navigating to a page that uses Cloudflare Turnstile for bot detection, the Turnstile widget initializes but immediately hangs and reports error 300030, preventing the challenge from completing.

### Steps to reproduce

* Navigate to any page that loads Cloudflare Turnstile (e.g. https://bithumen.be/login.php)
* Trigger the Turnstile widget (e.g. submit a login form that calls verify-login.php, which dynamically injects challenges.cloudflare.com/turnstile/v0/api.js)
* Observe the JS console and network traffic

### Observed behaviour

```
console.log  Turnstile Widget seem to have hung: <widget-id>
console.warn [Cloudflare Turnstile] Error: 300030.
```
Turnstile then loads a crash-retry iframe:

`GET challenges.cloudflare.com/cdn-cgi/challenge-platform/.../crashed_retry/normal`
followed by:

```
console.warn [Cloudflare Turnstile] Cannot find Widget <widget-id>, consider using turnstile.remove() to clean up a widget.
The challenge never completes. No token is produced, so any server-side validation that requires a Turnstile token fails.

```
### Expected behaviour

Turnstile's managed mode ("invisible" challenge) should complete automatically and invoke the success callback with a token.

### Root cause hypothesis

Turnstile fingerprints the browser environment during initialization. Error 300030 indicates widget initialization failure, which typically means a required browser API (likely related to iframe sandbox, navigator properties, or specific Web APIs) is missing or behaving unexpectedly in Lightpanda. The widget never makes any further requests to challenges.cloudflare.com after loading api.js, suggesting it aborts synchronously during environment detection.

### Environment

Lightpanda version: 0.2.6 (latest)
Connected via CDP using Playwright Python 1.58.0
Turnstile sitekey: 0x4AAAAAAAB_PUTgqXjn4qte (managed/automatic mode)

krichprollsch:
@kjozsa did you have chance to re-test w/ the last nightly? does #1884 fixed the issue?

kjozsa:
@krichprollsch Tested with today's nightly (25a7b5b7). The crash-on-load is gone — Lightpanda now loads the page and the Turnstile iframe initialises. However, the challenge JS (`turnstile/v0/api.js`) hits a TypeError: no access on every setInterval tick inside the iframe, so the token is never produced. Progress, but not a full fix yet.

PR #1884, opened 2026-03-17
fix: stub navigator.permissions, storage, deviceMemory to unblock Turnstile

Tenith01:
Relates to #1869

### Description
This PR provides an investigative fix for Cloudflare Turnstile's `300030` ("Widget Hung") error. 

**Important Note:** I am not claiming this is a full Turnstile bot-detection bypass. This fix specifically targets the 300030 hang, which is caused by Turnstile probing for standard browser environment APIs and hanging indefinitely when those expected types or Promise resolutions are completely missing.

By providing basic stubs for the following APIs, we unblock the widget's synchronous/asynchronous probe sequences so it can finish executing instead of timing out:
- `navigator.permissions.query()`
- `navigator.storage.estimate()`
- `navigator.connection` (`NetworkInformation`)
- `navigator.deviceMemory`
- `navigator.getBattery()` (now correctly returns a rejected Promise instead of `undefined`)

### Diagnostic Output
*As requested in the issue, here is the output from the diagnostic script showing the API hang before this fix:*

```text
HUNG: permissions.query(notifications)
HUNG: storage.estimate
```


Tenith01:

PR #1886, opened 2026-03-17
Use `git_version` option for version command

### LINKED ISSUE #1835: version command should return the version number on stable release

mrdimidium:
Uses changes from #1874 to fix #1835. No change to telemetry, only CLI command

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,6 +6,7 @@ env:
   AWS_BUCKET: ${{ vars.NIGHTLY_BUILD_AWS_BUCKET }}
   AWS_REGION: ${{ vars.NIGHTLY_BUILD_AWS_REGION }}
   RELEASE: ${{ github.ref_type == 'tag' && github.ref_name || 'nightly' }}
+  GIT_VERSION_FLAG: ${{ github.ref_type == 'tag' && format('-Dgit_version={0}', github.ref_name) || '' }}
 
 on:
   push:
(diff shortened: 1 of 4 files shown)


krichprollsch:
Actually I think having the `vX.Y.Z` in telemetry would be useful too.
Telemetry accepts this format now.

.github/workflows/build.yml
```diff
@@ -45,7 +46,7 @@ jobs:
         run: zig build -Dprebuilt_v8_path=v8/libc_v8.a -Doptimize=ReleaseFast snapshot_creator -- src/snapshot.bin
 
       - name: zig build
-        run: zig build -Dsnapshot_path=../../snapshot.bin -Dprebuilt_v8_path=v8/libc_v8.a -Doptimize=ReleaseFast -Dcpu=x86_64 -Dgit_commit=$(git rev-parse --short ${{ github.sha }})
+        run: zig build -Dsnapshot_path=../../snapshot.bin -Dprebuilt_v8_path=v8/libc_v8.a -Doptimize=ReleaseFast -Dcpu=x86_64 -Dgit_commit=$(git rev-parse --short ${{ github.sha }}) ${{ env.GIT_VERSION_FLAG }} ${{ env.GIT_VERSION_FLAG }}
```
krichprollsch: Shouldn't be `-Dgit_version=${{ env.GIT_VERSION_FLAG }}`
?

krichprollsch: Ho my bad, actually the option is in `GIT_VERSION_FLAG` 🤦 
The only issue here is the double `${{ env.GIT_VERSION_FLAG }}`

PR #1888, opened 2026-03-17
Cleanup CI a little bit

mrdimidium:
Fixes several minor issues in CI:
- Updated file paths for checking for changes
- Removed remnants of submodules
- Removed copies of LIGHTPANDA_TELEMETRY_DISABLE
- Moved zig fmt from a separate task to zig-test -- they always run together, and there's no need to wait for a separate worker for an instant operation.

You might notice that `zig-test` and `e2e-test` do extra work by building the browser multiple times and always running with the same changes. They should be combined into `test.yml`, but I don't have access to AWS to verify that everything is working correctly.


.github/workflows/e2e-test.yml
```diff
@@ -29,12 +27,10 @@ on:
 
     paths:
```
krichprollsch, line 28: shouldn't be the `paths` similar to the branche's one?

mrdimidium: @krichprollsch, aren't they identical?

krichprollsch: Oh yes my bad :facepalm:

PR #1889, opened 2026-03-17
Rework header/data callbacks in HttpClient

mrdimidium:
Libcurl is designed so that `headerCallback` and `dataCallback` are called on the thread where `perform()` is called. This means that you can't simply move `multi` to the network runtime, as the callbacks call cdp functions, which are inherently not thread-safe.

Instead of handling redirects and authorization in the callback, we disable libcurl's built-in redirects and explicitly follow the redirect chain in `processMessages`. In addition to simplifying the code without manually parsing HTTP, this allows us to execute network functions in main, retrieve the HTTP header, and process it on the cdp thread. libcurl buffers headers and response codes anyway, so we don't lose any memory.

For data, we write a temporary buffer to move the call to cdp callbacks to processMessages


src/browser/HttpClient.zig
```diff
@@ -1015,15 +959,21 @@ pub const Transfer = struct {
...
+    _auth_challenge: ?http.AuthChallenge = null,
 
     // number of times the transfer has been tried.
     // incremented by reset func.
     _tries: u8 = 0,
     _performing: bool = false,
+    _redirect_count: u8 = 0,
+
+    // Buffered response body. Filled by bufferDataCallback, consumed in processMessages.
+    _stream_buffer: std.ArrayListUnmanaged(u8) = .{},
```
karlseguin:
```suggestion
    _stream_buffer: std.ArrayList(u8) = .{},
```

src/browser/HttpClient.zig
```diff
@@ -1037,36 +987,138 @@ pub const Transfer = struct {
...
+            const arena = self.arena.allocator();
+            var aw: std.Io.Writer.Allocating = .init(arena);
+            try jar.forRequest(req.url, &aw.writer, .{
+                .is_http = true,
+                .origin_url = req.url,
+                .is_navigation = req.resource_type == .document,
+            });
+            const written = aw.written();
+            if (written.len > 0) {
+                try aw.writer.writeByte(0);
```
karlseguin: I like that this is handled internally, but not that it isn't exposed, e.g. in cdp we need to re-execute this (it isn't the performance I'm worried about here, just the poor cohesion). Can this value not be stored in req.headers.cookies and the iterator restored?

src/browser/HttpClient.zig
```diff
@@ -1275,171 +1301,40 @@ pub const Transfer = struct {
...
-            return Net.writefunc_error;
+        const conn: *http.Connection = @ptrCast(@alignCast(data));
+        var transfer = fromConnection(conn) catch |err| {
+            lp.log.err(.http, "get private info", .{ .err = err, .source = "body callback" });
+            return http.writefunc_error;
         };
 
-        if (transfer._redirecting or transfer._auth_challenge != null) {
+        // Skip body for responses that will be retried (redirects, auth challenges).
+        const status = conn.getResponseCode() catch return http.writefunc_error;
```
karlseguin: Should we be logging this? Seems like it could save us a bit of debugging if it ever happens. Or put the log in `getResponseCode` directly since this `catch return` happens a few times.

src/browser/HttpClient.zig
```diff
@@ -194,21 +218,19 @@ pub fn abortFrame(self: *Client, frame_id: u32) void {
...
+        var n = self.in_use.first;
         while (n) |node| {
             n = node.next;
-            const conn: *Net.Connection = @fieldParentPtr("node", node);
+            const conn: *http.Connection = @fieldParentPtr("node", node);
             var transfer = Transfer.fromConnection(conn) catch |err| {
                 // Let's cleanup what we can
                 self.removeConn(conn);
-                log.err(.http, "get private info", .{ .err = err, .source = "abort" });
+                lp.log.err(.http, "get private info", .{ .err = err, .source = "abort" });
```
karlseguin: I don't mind the change (I've talked to Pierre before about this, including for String, Page, js, ...). But it does make a complex PR more noisy than it has to be. It also makes it inconsistent with the rest of the codebase. Prefer to (a) discuss this with the team (b) if agreed, do it throughout the codebase (c) in a dedicated PR.

src/browser/HttpClient.zig
```diff
@@ -1275,171 +1301,40 @@ pub const Transfer = struct {
...
+        const conn: *http.Connection = @ptrCast(@alignCast(data));
+        var transfer = fromConnection(conn) catch |err| {
+            lp.log.err(.http, "get private info", .{ .err = err, .source = "body callback" });
+            return http.writefunc_error;
         };
 
-        if (transfer._redirecting or transfer._auth_challenge != null) {
+        // Skip body for responses that will be retried (redirects, auth challenges).
+        const status = conn.getResponseCode() catch return http.writefunc_error;
+        if ((status >= 300 and status <= 399) or status == 401 or status == 407) {
```
karlseguin: You're doing this check on every chunk. I think you should have a `is_first` flag..and only do that check there. PLUS, you can get the contentLength and (a) immediately enforce max length and (b) pre-size your buffer.

(You could also cache the status maybe, since it's called in a few other places I think).

You could reuse the `_header_done_called` flag for this..but...(see next comment)

src/browser/HttpClient.zig
```diff
@@ -1275,171 +1301,40 @@ pub const Transfer = struct {
...
             return @intCast(chunk_len);
         }
 
-        if (!transfer._header_done_called) {
```
karlseguin, line 1422: Aside from my above comment, to do some work on the first chunk, I'm not sure you need this `_header_done_called` callback. It specifically exists because it could be called from the first data chunk OR called in processMessage from a body-less response. I don't think  (???) it's possible for `transfer.headerDoneCallback` to be called twice anymore.

karlseguin:
1 - I'm not sure we have automated tests in the CI that cover all the CDP interception / auth request / proxy / proxy auth. If not, all of those cases need to be manually tested before thi is merged.

2 - This breaks any concept of "streaming". I don't think it's a huge deal, Fetch already broke it. But it _was_ working in XHR before this. We _were_ sending incremental progress events. Now we'll just send 1 at the end. Again, I think it's fine..for now. But can we think of any future cases where that would be a problem?

3 - The biggest issue, I think, is that is 2x the peak memory + includes an extra copy. What do you think about having the caller provide the arena? I'm not sure if that arena would become the transfer.arena, or if we'd keep transfer.arena for internal stuff, and have a body_arena passed in.

karlseguin:
Currently, every thread still has its own multi/libcurl. The goal is to move it to a single one on the main thread, yes? Won't all this burden that main thread even more? Way more allocations (e.g. the response buffer) and cleanup? And what will the transfer to worker thread look like? Like, now we still do the registered callbacks (e.g. `req.done_callback`) in `perform`, but what about when perform is moved to the main thread? (or am I totally confused?)

mrdimidium:
All points make sense.

1. We have automated tests for redirects and setting cookies during redirects in `demo`, authorization had to be tested manually (but I could add a test in the `demo`).
2. body streaming will continue to work as usual; we'll wake up a `cdp` thread for each chunk. This also eliminates the need to allocate a buffer for the entire response (and reallocate it altogether).
3. We should bring back streaming when we move multi and this will solve the problem of increased memory consumption.

karlseguin:
So your saying that this intermediary step _does_ 2x the memory + introduce an extra copy + remove streaming, the end version won't have any of those issues? Ok.  I think that's fine. (but still worth reviewing my inline comments)

src/browser/HttpClient.zig
krichprollsch: I agree, the PR is touchy enough to review w/o this kind of coding changes.

krichprollsch:
>> 1 - I'm not sure we have automated tests in the CI that cover all the CDP interception / auth request / proxy / proxy auth. If not, all of those cases need to be manually tested before thi is merged.
> 1. We have automated tests for redirects and setting cookies during redirects in demo, authorization had to be tested manually (but I could add a test in the demo).

We have `puppeteer/ri_authenticate.js` which test basic auth and `puppeteer/proxy_auth.js` for proxy auth.
We have `puppeteer/request_interception.js` for basic request interception. They are all part of for the e2e runner.
We run all of them in direct but also through a configured proxy.

src/browser/HttpClient.zig
```diff
@@ -1037,36 +987,138 @@ pub const Transfer = struct {
...
+            try wba.signRequest(self.arena.allocator(), &header_list, authority);
+        }
+
+        // Add cookies from cookie jar.
+        if (req.cookie_jar) |jar| {
+            const arena = self.arena.allocator();
+            var aw: std.Io.Writer.Allocating = .init(arena);
+            try jar.forRequest(req.url, &aw.writer, .{
+                .is_http = true,
+                .origin_url = req.url,
```
krichprollsch: Is it this correct in case of redirection? Isn't a risk to send same site cookies across domains? 

I know this potential bug already exists in `main`.

src/browser/HttpClient.zig
krichprollsch: It's also in `main` but the content-type parsing is already done by `transfer.buildResponseHeader`

src/browser/HttpClient.zig
```diff
@@ -848,14 +777,33 @@ fn processMessages(self: *Client) !bool {
...
+                };
+
+                const conn = transfer._conn.?;
+
+                try self.handles.remove(conn);
+                transfer.reset();
+                try transfer.configureConn(conn);
+                try self.handles.add(conn);
+
+                _ = try self.perform(0);
```
krichprollsch: Should we handle error with transfer failure + deinit instead of bubbling the error?

src/browser/HttpClient.zig
mrdimidium: I left it in because I plan to bring back streaming. We'll be hitting processMessages potentially for every chunk, and we'll need to make sure we haven't called it twice.

This is also useful for asserting, to debug the other error we discussed.

src/browser/HttpClient.zig
mrdimidium: There were some issues with `req.headers.cookies` that I didn't immediately remember. I moved the duplication into a function to avoid copying.

src/browser/HttpClient.zig
mrdimidium: You're right, there was also a problem with double `releaseConn`. I moved the loop body into a function to handle the error in one place.

karlseguin:
LGTM

src/browser/HttpClient.zig
```diff
@@ -1107,6 +1174,20 @@ pub const Transfer = struct {
         }
     }
 
+    pub fn getCookieString(self: *Transfer) !?[:0]const u8 {
+        const jar = self.req.cookie_jar orelse return null;
+        var aw: std.Io.Writer.Allocating = .init(self.arena.allocator());
+        try jar.forRequest(self.req.url, &aw.writer, .{
+            .is_http = true,
+            .origin_url = self.req.url,
```
krichprollsch: It's not correct. Before that change, we used `Page.requestCookie` which set the origin w/ its own url.
Here you always set the origin w/ the request target url, so it always be equal with the request's target url, leading to set cookies incorrectly.

Before that PR, the origin was correctly set for all requests except for redirections.
Now, if I'm not wrong, we set a bad origin for all requests, which is a regression.

I think `Request` should probably embed an origin we could use to always set cookies correctly.

PR #1894, opened 2026-03-18
SemanticTree: implement interactiveOnly filter and optimize token usage

arrufat:


src/browser/interactive.zig
```diff
@@ -253,17 +253,29 @@ pub fn classifyInteractivity(
...
-        if (std.ascii.eqlIgnoreCase(role, r)) return true;
-    }
-    return false;
+    return for (interactive_roles) |r| {
+        if (std.ascii.eqlIgnoreCase(role, r)) break true;
+    } else false;
+}
+
+pub fn isContentRole(role: []const u8) bool {
+    const content_roles = [_][]const u8{
```
karlseguin: For this and `isInteractiveRole`, I'd be tempted to lower the role and then use a StaticStringMap`

```zig
const MAX_LEN = "columnheader".len;
if (role.len > MAX_LEN) {
  return false;
}
var buf: [MAX_LEN]u8 = undefined;
const lowered = std.ascii.lowerString(&buf, role);
....
```

If you're willing to burden the caller, you could even push that up to them. In SemanticTree you could lower once for both isInteractiveRole and isContentRole. But then that's something all callers need to do (are there other callers?)

src/SemanticTree.zig
```diff
@@ -389,36 +406,46 @@ const TextVisitor = struct {
...
             const trimmed = std.mem.trim(u8, text_node.getWholeText(), " \t\r\n");
             if (trimmed.len > 0) {
-                try self.writer.writeAll(trimmed);
+                name_to_print = trimmed;
+            }
+        }
+
+        const is_text_only = std.mem.eql(u8, data.role, "StaticText") or std.mem.eql(u8, data.role, "none") or std.mem.eql(u8, data.role, "generic");
+
+        if (is_text_only and name_to_print != null) {
```
karlseguin: Pretty sure this is the same, and easier to follow:

```zig
try self.writer.print("{d}", .{data.id});
if (is_text_only == false) {
    try self.writer.print(" {s}", .{data.role});
}
if (name_to_print) |n| {
    try self.writer.print(" '{s}'", .{n});
}
```

src/browser/interactive.zig
arrufat: I think I prefer the easier API, but I'll take the StaticStringMap.

PR #1896, opened 2026-03-18
feat: fetch add wait_until parameter for page loads options

shaewe180:
**Overview**
Currently, the `fetch` command uses a hardcoded 5000ms timeout and a fixed "wait until idle" strategy. This limits the tool's effectiveness when dealing with slow-loading pages, heavy Single Page Applications (SPAs), or scenarios requiring specific lifecycle events (e.g., only waiting for the DOM to become interactive).

This PR introduces two new command-line arguments to the `fetch` command: `--wait_ms` and `--wait_until`.

**Key Changes**
1. **Configurable Wait Time**: Replaced the hardcoded 5000ms limit with the `--wait_ms` flag (defaults to 5000ms to maintain backward compatibility).
2. **Flexible Wait Strategies**: Introduced the `--wait_until` flag, allowing users to choose the condition for returning the DOM:
    * `load` (default): Waits for the `window.onload` event to fire.
    * `domcontentloaded`: Returns immediately after the HTML is parsed and scripts have finished running, without waiting for subresources like images to load.
    * `networkidle`: Waits until there are no active HTTP requests and the page is in a stable state.
    * `fixed`: Forces the engine to wait for the exact duration specified by `--wait_ms`, regardless of what events the page triggers.
3. **Engine Refactoring**: Updated `src/browser/Session.zig` to support these strategies in the core wait loop, ensuring the engine can intelligently yield or return based on the requested events and pending JavaScript macrotasks.
4. **System-wide Consistency**: Updated all internal call sites (tests, MCP tools, and CDP domains) to align with the new `Session.wait` function signature.

**Benefits:**
* **Support for Slow-loading Websites**: Users can now scrape pages that take longer than 5 seconds to load (when used in conjunction with `--http_timeout`).
* **Performance Optimization**: For simple pages, using `--wait_until domcontentloaded` skips waiting for unnecessary static resources, significantly speeding up the dump process.
* **Predictability**: The `fixed` strategy is highly valuable for debugging pages with complex, non-deterministic background loading logic.

**Usage Examples:**

```bash
# Wait up to 10 seconds until the network is completely idle
./lightpanda fetch --dump html --wait_ms 10000 --wait_until networkidle https://example.com

# Quick dump immediately when the DOM is ready
./lightpanda fetch --dump html --wait_until domcontentloaded https://example.com
```

### DIFF
--- a/src/Config.zig
+++ b/src/Config.zig
@@ -222,8 +222,6 @@ pub const WaitUntil = enum {
     domcontentloaded,
     networkidle,
     fixed,
-
-    pub const js_enum_from_string = true;
 };
 
 pub const Fetch = struct {


shaewe180:

src/Config.zig
```diff
@@ -217,13 +217,24 @@ pub const DumpFormat = enum {
     semantic_tree_text,
 };
 
+pub const WaitUntil = enum {
+    load,
+    domcontentloaded,
+    networkidle,
+    fixed,
+
+    pub const js_enum_from_string = true;
```
```suggestion
```

This is only needed for enums that are serialized to v8/js.

src/browser/Session.zig
```diff
@@ -319,15 +319,15 @@ fn findPageBy(page: *Page, comptime field: []const u8, id: u32) ?*Page {
     return null;
 }
 
-pub fn wait(self: *Session, wait_ms: u32) WaitResult {
+pub fn wait(self: *Session, wait_ms: u32, wait_until: lp.Config.WaitUntil) WaitResult {
     var page = &(self.page orelse return .no_page);
     while (true) {
-        const wait_result = self._wait(page, wait_ms) catch |err| {
+        const wait_result = self._wait(&page, wait_ms, wait_until) catch |err| {
```
karlseguin: There's no reason to pass a pointer to the pointer of the page. 

```suggestion
        const wait_result = self._wait(page, wait_ms, wait_until) catch |err| {
```

src/browser/Session.zig
```diff
@@ -346,7 +346,7 @@ pub fn wait(self: *Session, wait_ms: u32) WaitResult {
     }
 }
 
-fn _wait(self: *Session, page: *Page, wait_ms: u32) !WaitResult {
+fn _wait(self: *Session, page: **Page, wait_ms: u32, wait_until: lp.Config.WaitUntil) !WaitResult {
```
karlseguin:
```suggestion
fn _wait(self: *Session, page: *Page, wait_ms: u32, wait_until: lp.Config.WaitUntil) !WaitResult {
```

related to the above comment. All of the `page.X` that were changed to `page.*.X` will have to be updated.

src/browser/Session.zig
```diff
@@ -423,37 +425,42 @@ fn _wait(self: *Session, page: *Page, wait_ms: u32) !WaitResult {
...
-                        //     // Look, we want to exit ASAP, but we don't want
-                        //     // to exit so fast that we've run none of the
-                        //     // background jobs.
-                        //     break :blk 50;
-                        // }
+                    const is_event_done = switch (wait_until) {
+                        .fixed => false,
+                        .domcontentloaded => (page.*._load_state == .load or page.*._load_state == .complete),
+                        .load => (page.*._load_state == .complete),
+                        .networkidle => (page.*._load_state == .complete and http_active == 0),
```
karlseguin: What about using `page._notified_network_idle == .done` here? It would introduce a 500ms delay to allow pending work (e.g. a setTimeout) to start new connections. On the flip, that could result in some telemetry/beacon keep the page alive longer than desired.

src/Config.zig
```diff
@@ -217,13 +217,24 @@ pub const DumpFormat = enum {
...
+};
+
 pub const Fetch = struct {
     url: [:0]const u8,
     dump_mode: ?DumpFormat = null,
     common: Common = .{},
     with_base: bool = false,
     with_frames: bool = false,
     strip: dump.Opts.Strip = .{},
+    wait_ms: u32 = 5000,
```
karlseguin, line 234: Could you add the help text for these in the `fetch` section of the `printUsageAndExit` function?

shaewe180:
Thank you for your suggestion. I have  made the modifications.

PR #1897, opened 2026-03-18
Introduce StyleManager

karlseguin:
A Page now has a StyleManager. The StyleManager currently answers two questions: 
1 - Is an element hidden
2 - Does an element have pointer-events == none

This is used in calls such as element.checkVisibility which, on some pages, can be called tens of thousands of times (often through other methods, like element.getBoundingClientRect). This _can_ be a bottleneck.

The StyleManager keeps a list of rules. The rules include the selector, specificity, and properties that we care about. Rules in a stylesheet that contain no properties of interest are ignored. This is the first and likely most significant optimization. Presumably, most CSS rules don't have a display/visibility/opacity or pointer-events property.

The list is rules is cached until stylesheets are modified or delete. When this happens, the StyleManager is flagged as "dirty" and rebuilt on-demand in the next query.  This is our second major optimization.

For now, to check if an element is visible, we still need to scan all rules. But having a pre-build subset of all the rules is a first step.

The next step might be to optimize the matching, or possibly optimizing common cases (e.g. id and/or simple class selector)

### DIFF
--- a/src/browser/StyleManager.zig
+++ b/src/browser/StyleManager.zig
@@ -354,7 +354,7 @@ pub fn computeSpecificity(selector: Selector.Selector) u32 {
 
     // Pack into single u32: (ids << 20) | (classes << 10) | elements
     // This gives us 10 bits each, supporting up to 1023 of each type
-    return (@min(ids, 1023) << 20) | (@min(classes, 1023) << 10) | @min(elements, 1023);
+    return (@as(u32, @min(ids, 1023)) << 20) | (@as(u32, @min(classes, 1023)) << 10) | @min(elements, 1023);
 }
 
 fn countCompoundSpecificity(compound: Selector.Compound, ids: *u32, classes: *u32, elements: *u32) void {
@@ -376,7 +376,6 @@ fn countCompoundSpecificity(compound: Selector.Compound, ids: *u32, classes: *u3
                             const spec = computeSpecificity(nested_sel);
                             if (spec > max_nested) max_nested = spec;
                         }
-                    max_nested = @min(max_nested, 1023);
                         // Unpack and add to our counts
                         ids.* += (max_nested >> 20) & 0x3FF;
                         classes.* += (max_nested >> 10) & 0x3FF;


src/browser/StyleManager.zig
```diff
@@ -0,0 +1,572 @@
...
+    countCompoundSpecificity(selector.first, &ids, &classes, &elements);
+
+    // Count specificity for subsequent segments
+    for (selector.segments) |segment| {
+        countCompoundSpecificity(segment.compound, &ids, &classes, &elements);
+    }
+
+    // Pack into single u32: (ids << 20) | (classes << 10) | elements
+    // This gives us 10 bits each, supporting up to 1023 of each type
+    return (ids << 20) | (classes << 10) | elements;
```
arrufat: I think it's really uncommon, but do you think we should protect ourselves against weird CSS selectors?
```suggestion
    return (@min(ids, 1023) << 20) | (@min(classes, 1023) << 10) | @min(elements, 1023);
```

src/browser/StyleManager.zig
```diff
@@ -0,0 +1,572 @@
...
+                switch (pc) {
+                    // :where() has zero specificity
+                    .where => {},
+                    // :not(), :is(), :has() take specificity of their most specific argument
+                    .not, .is, .has => |nested| {
+                        var max_nested: u32 = 0;
+                        for (nested) |nested_sel| {
+                            const spec = computeSpecificity(nested_sel);
+                            if (spec > max_nested) max_nested = spec;
+                        }
```
arrufat, line 378: Maybe I am too paranoid?
```suggestion
                        }
                    max_nested = @min(max_nested, 1023);
```

src/browser/StyleManager.zig
```diff
@@ -0,0 +1,572 @@
...
+fn rebuildIfDirty(self: *StyleManager) !void {
+    if (!self.dirty) {
+        return;
+    }
+
+    self.dirty = false;
+    const item_count = self.rules.items.len;
+
+    const arena: *std.heap.ArenaAllocator = @ptrCast(@alignCast(self.arena.ptr));
+    _ = arena.reset(.retain_capacity);
```
arrufat: Couldn't we just add a `resetArena` to page and just do `self.page.resetArena(.retain_capacity)`?

Or let StyleManager have its own arena.

PR #1900, opened 2026-03-18
distapch InputEvent on input/TextArea changes

krichprollsch:
fix for ddg integration test 1/2


src/browser/webapi/element/html/Input.zig
```diff
@@ -103,6 +104,11 @@ fn dispatchSelectionChangeEvent(self: *Input, page: *Page) !void {
     try page._event_manager.dispatch(self.asElement().asEventTarget(), event);
 }
 
+fn dispatchInputEvent(self: *Input, data: ?[]const u8, input_type: []const u8, page: *Page) !void {
+    const event = try InputEvent.init("input", .{ .data = data, .inputType = input_type }, page);
```
karlseguin: Should these events be trusted?

krichprollsch: I think so, updated.

PR #1909, opened 2026-03-18
fix(cdp): add ignore-level log to setUserAgentOverride stub

jnMetaCode:
## Summary
- Adds an `.ignore`-level log message to the `setUserAgentOverride` noop stub
- Follow-up to #1860 per discussion with @karlseguin and @krichprollsch
- Uses `.ignore` rather than `.not_implemented` since this is an explicit design choice not to implement
- Adds new `.ignore` variant to `log.Scope` enum to distinguish deliberately ignored features

## Test plan
- Verify log output appears when Playwright calls setUserAgentOverride
- Verify the `.ignore` scope is properly filtered/displayed in log output

### DIFF
--- a/src/cdp/domains/emulation.zig
+++ b/src/cdp/domains/emulation.zig
@@ -69,6 +69,6 @@ fn setTouchEmulationEnabled(cmd: anytype) !void {
 }
 
 fn setUserAgentOverride(cmd: anytype) !void {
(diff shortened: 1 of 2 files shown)


karlseguin:
Thanks, but I think we can stick with the use of the `.app` that I added.

PR #1910, opened 2026-03-19
Bucket stylesheet rules

karlseguin:
In the first iteration of this, we kept an ArrayList of all rules with visibility properties. Why bother evaluating if a rule's selector matches an element if that rule doesn't have any meanignful (i.e. visibility) properties?

This commit enhances that approach by bucketing the rules. Given the following selectors:
```
.hidden {....}
.footer > .small {...}
```

We can store the rules based on their right-most selector. So, given an element we can do:

```
if (getId(el)) |id| {
   const rules = id_lookup.get(id) orelse continue;
   // check rules
}

if (getClasses(el)) |classes| {
   for (classes) |c| {
     const rules = class_lookup(c) orelse continue;
     // chck rules
   }
}
...
```

On an amazon product page, the total list of visibility-related rules was ~230. Now, scanning 230 rules for a match isn't _aweful_, but remember that this has to be done up the ancestor tree AND, for Amazon, this is called over 20K times.

This change requires that the StyleManager becomes more matching/parsing-aware but a typical visibility check on that same Amazon product page only has to check 2 rules (down from 230) and often has to check 0 rules.

Also, we now filter out a few more interactive-related pseudo-elements, e.g. :hover. These aren't supported by the browser as a whole (i.e. they can _never_ match), so they can be filtered out early, when building the rules lookup.

### DIFF
--- a/src/browser/StyleManager.zig
+++ b/src/browser/StyleManager.zig
@@ -100,15 +100,25 @@ fn rebuildIfDirty(self: *StyleManager) !void {
 
     self.dirty = false;
     errdefer self.dirty = true;
+    const id_rules_count = self.id_rules.count();
+    const class_rules_count = self.class_rules.count();
+    const tag_rules_count = self.tag_rules.count();
+    const other_rules_count = self.other_rules.items.len;
 
     self.page._session.arena_pool.resetRetain(self.arena);
 
-    // Clear all buckets and reset document order
-    self.id_rules = .empty;
-    self.class_rules = .empty;
-    self.tag_rules = .empty;
-    self.other_rules = .empty;
     self.next_doc_order = 0;
+
+    self.id_rules = .empty;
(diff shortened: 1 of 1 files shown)


src/browser/StyleManager.zig
```diff
@@ -304,13 +428,94 @@ fn addRule(self: *StyleManager, style_rule: *CSSStyleRule) !void {
...
+    id: []const u8,
+    class: []const u8,
+    tag: Tag,
+    other,
+};
+
+/// Returns the bucket key for a compound selector, or null if the selector
+/// ends with a dynamic pseudo-class that we should skip (hover, active, focus, etc.)
+fn getBucketKey(compound: Selector.Compound) ?BucketKey {
+    // Iterate parts in reverse to find the most specific bucketing key
```
arrufat: Are we iterating in reverse here?

karlseguin: Nope, but that's what I started with initially. Updated.

PR #1914, opened 2026-03-19
SemanticTree: add progressive discoverability

arrufat:
Adds `maxDepth` and `backendNodeId` parameters to `LP.getSemanticTree` to enable **progressive discoverability**. 

This allows AI agents to fetch a high-level overview of a page (saving tokens) and then "zoom in" on specific subtrees by ID when needed. We trade a few extra round-trips for significant token savings and better context management.

### DIFF
--- a/src/SemanticTree.zig
+++ b/src/SemanticTree.zig
@@ -73,7 +73,7 @@ const OptionData = struct {
 };
 
 const NodeData = struct {
-    id: u32,
+    id: CDPNode.Id,
     axn: AXNode,
     role: []const u8,
     name: ?[]const u8,


src/mcp/tools.zig
```diff
@@ -196,12 +200,24 @@ const ToolStreamingText = struct {
                 }
             },
             .semantic_tree => {
+                var root_node = self.page.document.asNode();
+                if (self.backendNodeId) |node_id| {
+                    if (self.registry) |registry| {
+                        if (registry.lookup_by_id.get(node_id)) |n| {
+                            root_node = n.dom;
+                        } else {
+                            log.warn(.mcp, "semantic_tree id {} missing", .{node_id});
```
karlseguin:
```suggestion
                            log.warn(.mcp, "semantic_tree id missing", .{.id = node_id});
```

src/SemanticTree.zig
```diff
@@ -83,7 +84,11 @@ const NodeData = struct {
     node_name: []const u8,
 };
 
-fn walk(self: @This(), node: *Node, xpath_buffer: *std.ArrayList(u8), parent_name: ?[]const u8, visitor: anytype, index: usize, listener_targets: interactive.ListenerTargetMap) !void {
+fn walk(self: @This(), node: *Node, xpath_buffer: *std.ArrayList(u8), parent_name: ?[]const u8, visitor: anytype, index: usize, listener_targets: interactive.ListenerTargetMap, current_depth: u32) !void {
+    if (self.max_depth) |max| {
```
karlseguin: You can default this to some absurd value, e.g. `std.math.maxInt(u32) - 1` so that instead of 2 checks, you can reliably do 1: `if (current_depth > max) return;`

src/mcp/tools.zig
```diff
@@ -161,6 +163,8 @@ const ToolStreamingText = struct {
     action: enum { markdown, links, semantic_tree },
     registry: ?*CDPNode.Registry = null,
     arena: ?std.mem.Allocator = null,
+    backendNodeId: ?u32 = null,
```
karlseguin, line 166: Sometimes this file uses `u32`, sometimes `CDPNode.Id`

src/mcp/tools.zig
arrufat: Oh, this is nice, I was just checking the implementation now. Thank you!

PR #1917, opened 2026-03-19
build: automate version resolution in build.zig

arrufat:
Removes manual git flags from CI and build scripts. Versioning is now automatically derived from git and build.zig.zon.

With this PR, we follow https://semver.org/
Logic:

1. Read the version from build.zig.zon
2. If it doesn't have a `.pre` field (i.e. dev/alpha/beta) it will use that
3. Otherwise it will get the info from git: hash and number of commits since last `.0` version
4. Then build the version: `0.3.0-dev.1493+0896edc3`

Note that, since the latest stable version is `0.2.6`. The convention is to use `0.3.0-dev`, as:
- `0.2.6` < `0.3.0.dev` < `0.3.0`

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -1,6 +1,6 @@
 .{
     .name = .browser,
-    .version = "0.3.0-dev",
+    .version = "1.0.0-dev",
     .fingerprint = 0xda130f3af836cea0, // Changing this has security and trust implications.
     .minimum_zig_version = "0.15.2",
     .dependencies = .{


arrufat:
Yes, I get that having to modify `build.zig.zon` is annoying. We could ignore it completely and just get the version from the latest git tag, though. So the current one would be `0.2.7-dev.123+0896edc3` (we would have some logic to increment the `.patch` when we are not sitting on a tag.

Regarding the fact that it breaks the cache, I haven't noticed any regression: when running `zig build` twice, the second time is "instant" (it takes about 1.5 s, same as with the current approach).

All that being said, it was just a suggestion to have automatic semantic versioning instead of dev or the git hash. But feel free to close if you don't like it :)

mrdimidium:
I have no opinion on semantic versioning, other than that it might break telemetry. We need @krichprollsch approve for it.

Regarding technical changes, instead of manually calling `git` in `build.zig`, you could read the env vars (`GITHUB_SHA` and `GITHUB_REF`) instead of the copy-pasted ones you removed from the .yml files. The changes to `build.zig.zon` parsing also look good.

krichprollsch:
Is there a real pro to have the version into `build.zig.zon`? I agree having to do both the `build.zig.zon` change + the git tag is a bit annoying, but I can handle it.
But let's say I want to tag `v2.7.0` what should I do exactly? BTW the `v` should drop correct?

mrdimidium:
I think the benefit is the ability to install `lightpanda` as a dependency and see real version. And maybe greater integration for external tools in future.

arrufat:
If you want to release version 0.2.7, with the proposed approach you should do something like.

1. modify `build.zig.zon` version to `0.2.7`
2. git add -u && git commit -m 'release version 0.2.7'
3. git tag 0.2.7
4. modify `build.zig.zon` version to `0.3.0-dev` or 0.2.8-dev`
5. git add -u && git commit -m 'bump version to dev`

We could ignore the build.zig.zon and just rely on the git tag, though.

Ah, I think the main issue with this approach is that you need commit access to main...

krichprollsch:
> Note that, since the latest stable version is 0.2.6. The convention is to use 0.3.0-dev, as:

We can't really say `0.2.6` is stable :sweat_smile: it's just a way to have a fixed ref...

> Then build the version: 0.3.0-dev.1493+0896edc3

We agree `0896edc3` is the commit's hash?

> modify build.zig.zon version to 0.3.0-dev or 0.2.8-dev`

I suppose ` 0.2.8-dev` would make more sense in the current workflow :thinking: not sure. 
maybe a `1.0.0-dev` does?

> We could ignore the build.zig.zon and just rely on the git tag, though.

As I said, I'm not against changing `build.zig.zon` if it gives advantages in the Zig ecosystem, which seems to be the case.

I think we can maybe do this way:
* tag and push a 0.2.8.rc
* a ci workflow updates the `build.zon` to `0.2.8`, commit and tag, then it updates again `build.zon` to revert to the `1.0.0-dev` version
* the current release ci builds and publishes `0.2.8` (and ignores the `.rc` tags)
We could even create a PR and merge it for the `build.zon` changes to not change `main` directly. but tagging in a branch is not great.

arrufat:
> We agree 0896edc3 is the commit's hash?

yes, that's correct

> I suppose  0.2.8-dev would make more sense in the current workflow 🤔 not sure.
maybe a 1.0.0-dev does?

That would work, too.

> I think we can maybe do this way:
> 
> tag and push a 0.2.8.rc
> a ci workflow updates the build.zon to 0.2.8, commit and tag, then it updates again build.zon to revert to the 1.0.0-dev version
> the current release ci builds and publishes 0.2.8 (and ignores the .rc tags)
> We could even create a PR and merge it for the build.zon changes to not change main directly. but tagging in a branch is not great.


We don't even need the `.rc` I think. I used "stable" meaning that there's no `.pre` field in the version. I think it's widely understood that versions starting with `0.` are not stable.

But yes, automating this in CI would be great :)

arrufat:
Here's two ways of simplifying (not having to update `build.zig.zon`)
- run `zig build -Dversion_string=0.2.7`
  - that will bypass all logic and just use that version
- change the logic so that when we sit on a git tag, we just use that for the build
  - I like this one a bit more maybe... 

Either of those ways give a nice increasing version for untagged builds, e.g. `1.0.0-dev.123+0896edc3`

krichprollsch:
ok, the `-Dversion_string=0.2.7` usage LGTM.
I updated the telemetry to support this version.
One last thing: when building w/ `-Doptimize=ReleaseFast` I still have the `-dev` in the version. Do you know why?
We always have the `-dev` b/c it's a non-tagged version?

krichprollsch:
oh I get it, it's b/c it's part of the build.zig.zon version... my bad

mrdimidium:
A bit of off-topic on the implementation, not the idea itself. I'm almost certain that running external commands will inevitably create problems, and it's more reliable to read the hash and tag from the environment variables set by CI.

krichprollsch:
> A bit of off-topic on the implementation, not the idea itself. I'm almost certain that running external commands will inevitably create problems, and it's more reliable to read the hash and tag from the environment variables set by CI.

Well, the `-Dversion_string` is built using a `std.SemanticVersion.parse` call. So there is no risk to break the Zig semantic versioning format...
I'm thinking of calling it with `-Dversion_string=1.0.0-nightly` for nightly builds

krichprollsch:
@arrufat @mrdimidium I added the `-Dversion_string` option in the ci + the `1.0.0-dev` in the build.zig.zon.

I think It's good to merge if you are agree.

PR #1925, opened 2026-03-19
Return correct errors in promise rejections

nikneym:
Currently, we throw mixed errors; either errors with incorrect type or just error strings. This PR unifies `DOMException`s and ECMAScript errors to return correct error types.

### DIFF
--- a/src/browser/js/PromiseResolver.zig
+++ b/src/browser/js/PromiseResolver.zig
@@ -20,7 +20,6 @@ const js = @import("js.zig");
 const v8 = js.v8;
 
 const log = @import("../../log.zig");
-const DOMException = @import("../webapi/DOMException.zig");
 
 const DOMException = @import("../webapi/DOMException.zig");
 
@@ -67,17 +66,13 @@ pub fn reject(self: PromiseResolver, comptime source: []const u8, value: anytype
     };
 }
 
-pub const RejectError = union(enum(u4)) {
+pub const RejectError = union(enum) {
     /// Not to be confused with `DOMException`; this is bare `Error`.
     generic_error: []const u8,
     range_error: []const u8,
     reference_error: []const u8,
     syntax_error: []const u8,
     type_error: []const u8,
-    wasm_compile_error: void, // TODO.
-    wasm_link_error: void, // TODO.
-    wasm_runtime_error: void, // TODO.
-    wasm_suspend_error: void, // TODO.
     /// DOM exceptions are unknown to V8, belongs to web standards.
     dom_exception: struct { err: anyerror },
 };
@@ -86,9 +81,9 @@ pub const RejectError = union(enum(u4)) {
 pub fn rejectError(
     self: PromiseResolver,
     comptime source: []const u8,
-    comptime kind: RejectError,
+    err: RejectError,
 ) void {
-    const handle = switch (kind) {
+    const handle = switch (err) {
         .generic_error => |msg| self.local.isolate.createError(msg),
         .range_error => |msg| self.local.isolate.createRangeError(msg),
         .reference_error => |msg| self.local.isolate.createReferenceError(msg),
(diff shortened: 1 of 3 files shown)


src/browser/js/PromiseResolver.zig
```diff
@@ -65,20 +66,55 @@ pub fn reject(self: PromiseResolver, comptime source: []const u8, value: anytype
     };
 }
 
-pub const RejectError = union(enum) {
-    generic: []const u8,
-    type_error: []const u8,
-    dom_exception: anyerror,
+pub const RejectError = enum(u4) {
```
karlseguin: What does specifying the `u4` achieve here?

src/browser/js/PromiseResolver.zig
```diff
@@ -65,20 +66,55 @@ pub fn reject(self: PromiseResolver, comptime source: []const u8, value: anytype
...
+    self: PromiseResolver,
+    comptime kind: RejectError,
+    comptime source: []const u8,
+    /// Depending on `kind`, different payloads are required.
+    payload: switch (kind) {
+        .wasm_compile_error,
+        .wasm_link_error,
+        .wasm_runtime_error,
+        .wasm_suspend_error,
+        => unreachable,
```
karlseguin: If we're just going to make these fail at comptime, why define them as a valid kind? A lot more confusing to anyone using it to see the `wasm_compile_error` as an enum value, and then get an unreachable during compilation.

src/browser/js/PromiseResolver.zig
```diff
@@ -65,20 +66,55 @@ pub fn reject(self: PromiseResolver, comptime source: []const u8, value: anytype
...
+
+pub const DOMExceptionPayload = struct { err: anyerror };
+
+/// Rejects the promise w/ an error object.
+pub fn rejectError(
+    self: PromiseResolver,
+    comptime kind: RejectError,
+    comptime source: []const u8,
+    /// Depending on `kind`, different payloads are required.
+    payload: switch (kind) {
```
karlseguin: Why not make `kind` a union? The union captures a tag and a type, isn't that what this is doing, except by just not using an union?

nikneym: Idea was to have `kind` as comptime and payload as runtime value so I've separated the two.

karlseguin: But why? When should we do it this way vs a union?

nikneym: This saves us running `switch` at runtime; I agree that use case is niche but if we're sure that enum will be known at comptime at all cases, its just more verbose yet better to do.

`payload` could be just `anytype` too, I wanted to make function signature self-explanatory.

karlseguin: > I wanted to make function signature self-explanatory.

This is the function signature:
```zig
pub fn rejectError(
    self: PromiseResolver,
    comptime kind: RejectError,
    comptime source: []const u8,
    /// Depending on `kind`, different payloads are required.
    payload: switch (kind) { 
        .wasm_compile_error,
        .wasm_link_error,
        .wasm_runtime_error,
        .wasm_suspend_error,
        => unreachable, 
        .dom_exception => DOMExceptionPayload,
        inline else => []const u8,
    },
) void {
```

Or, you could do:
```zig
pub fn rejectError(self: PromiseResolver, err: RejectError) void {
}
```

In both cases, you need to lookup `RejectError`. Which signature is more clear?

nikneym: Of course the latter is more "clear" to read yet you lose the benefit of comptime evaluation for an arg that will always be static.

```zig
pub fn rejectError(
    self: PromiseResolver,
    comptime kind: RejectError,
    comptime source: []const u8,
    /// Depending on `kind`, different payloads are required.
    payload: switch (kind) { 
        .dom_exception => DOMExceptionPayload,
        inline else => []const u8,
    },
) void {}
```

What this actually resolves to essentially:

```zig
pub fn rejectWithTypeError(
    self: PromiseResolver,
    message: []const u8,
)
```

Look, we're also passing less arguments now.

Honestly I can't say that I'm satisfied from this argument, though for the sake of it, I made a commit that use `union(enum)`.

PR #1927, opened 2026-03-19
Feat/fetch wait options

karlseguin:
Small tweak/fixes to https://github.com/lightpanda-io/browser/pull/1896


krichprollsch:
LGTM, the options are helpful.
I also think we will have soon an option to wait until a given css selector returns something.

PR #1930, opened 2026-03-19
Return real frame ID in STARTUP getFrameTree when page exists

### LINKED ISSUE #1800: Playwright connectOverCDP fails: frame ID mismatch in Page.getFrameTree (startup path)
## Summary

Playwright's `connectOverCDP()` connects successfully but all `page.goto()` calls fail with "Frame has been detached." The root cause is a frame ID mismatch in the `Page.getFrameTree` response during the auto-attach startup path.

## Environment

- Lightpanda: nightly build `1522c902` (aarch64-macos)
- Playwright: 1.58.2
- Node.js: v22.21.1
- Platform: macOS darwin arm64

## Reproduction

```typescript
import { chromium } from "playwright";

const browser = await chromium.connectOverCDP("http://127.0.0.1:9222");
const page = browser.contexts()[0].pages()[0];

// This fails with "Frame has been detached"
await page.goto("https://example.com");
```

Start Lightpanda with `lightpanda serve --port 9222`, then run the above.

## Root Cause

Protocol trace shows a frame ID / target ID mismatch:

**Step 1:** Playwright sends `Target.setAutoAttach`. Lightpanda responds with:

```json
{
  "method": "Target.attachedToTarget",
  "params": {
    "sessionId": "STARTUP",
    "targetInfo": {
      "targetId": "TID-STARTUP-P",
      "type": "page"
    }
  }
}
```

**Step 2:** Playwright sends `Page.getFrameTree` on session `STARTUP`. Lightpanda responds with:

```json
{
  "result": {
    "frameTree": {
      "frame": {
        "id": "TID-STARTUP-B",
        "url": "about:blank"
      }
    }
  }
}
```

**The problem:** `Page.getFrameTree` returns `"TID-STARTUP-B"` (the browser target ID) as the frame ID, but all subsequent frame events (`Page.frameNavigated`, `Page.frameStartedLoading`, etc.) use `"TID-STARTUP-P"` (the page target ID). Playwright maps the frame to `TID-STARTUP-B` and can't find a session for `TID-STARTUP-P` during navigation.

**Expected behavior:** `Page.getFrameTree` should return the page target ID as the frame ID, matching the ID used in frame lifecycle events.

## Key Detail: Only Affects the Auto-Attach Startup Path

When creating a BrowserContext + Target manually via raw CDP, the frame IDs are consistent:

```
Target.createTarget → targetId: "FID-0000000001"
Page.getFrameTree   → frame.id: "FID-0000000001"  ✓ match
Page.frameNavigated → frameId:  "FID-0000000001"  ✓ match
```

But in the auto-attach startup path (used by `connectOverCDP`):

```
Target.attachedToTarget → targetId: "TID-STARTUP-P"
Page.getFrameTree       → frame.id: "TID-STARTUP-B"  ✗ mismatch (browser target ID)
```

The startup target uses different IDs for the page target vs what `getFrameTree` reports as the frame ID.

## Workaround

Raw CDP via WebSocket works correctly when you manually create a BrowserContext and Target:

```typescript
import WebSocket from "ws";

const ws = new WebSocket("ws://127.0.0.1:9222/");
// ... send/receive helpers ...

const ctx = await send("Target.createBrowserContext");
const target = await send("Target.createTarget", {
  url: "about:blank",
  browserContextId: ctx.browserContextId,
});
await send("Target.attachToTarget", {
  targetId: target.targetId,
  flatten: true,
});
await send("Page.enable");
await send("Page.navigate", { url: "https://example.com" });
// Works correctly — title, DOM, JS evaluation all functional
```

## Additional Note

`GET /json/list` returns 404, which prevents Playwright and other CDP clients from discovering targets via the standard endpoint. Only `/json/version` is implemented.

mvanhorn:
## Summary

- `dispatchStartupCommand` hard-codes `"TID-STARTUP"` as the frame ID in `Page.getFrameTree`. When a driver connects via `connectOverCDP` after a real page already exists, lifecycle events (`Page.frameNavigated`, etc.) use the actual page frame ID from `bc.target_id`. The driver's frame tracking was initialized with `"TID-STARTUP"`, so it can't match lifecycle events to the tracked frame - causing navigation to hang.
- PR #1808 partially addressed this but left the issue open because the `connectOverCDP + page.goto()` case still fails.
- This checks for an existing browser context with a `target_id` in `dispatchStartupCommand`. If present, returns the real frame ID and URL. Falls back to `"TID-STARTUP"` only when no page exists yet.

## Changes

- `dispatchStartupCommand`: check `command.cdp.browser_context` for a live `target_id` before returning the hard-coded startup frame
- New test: STARTUP `getFrameTree` without browser context returns `"TID-STARTUP"`
- New test: STARTUP `getFrameTree` with browser context returns real frame ID

## Test plan

- [x] All 367 existing tests pass (`zig build test`)
- [x] New test covers both paths (no page vs. existing page)
- [x] STARTUP session routing unaffected (keyed on `input.sessionId == "STARTUP"`, not frame ID)

Fixes #1800

This contribution was developed with AI assistance (Claude Code + Codex).

### DIFF
--- a/src/cdp/cdp.zig
+++ b/src/cdp/cdp.zig
@@ -191,6 +191,25 @@ pub fn CDPT(comptime TypeProvider: type) type {
             // Stagehand parses the response and error if we don't return a
             // correct one for this call.
             if (std.mem.eql(u8, method, "Page.getFrameTree")) {
+                // If a real page already exists (connectOverCDP flow), return
+                // its actual frame ID so that subsequent lifecycle events
+                // (frameNavigated, etc.) match the driver's frame tracking.
+                if (command.cdp.browser_context) |*bc| {
+                    if (bc.target_id) |*target_id| {
+                        return command.sendResult(.{
+                            .frameTree = .{
+                                .frame = .{
+                                    .id = target_id,
+                                    .loaderId = "LOADERID24DD2FD56CF1EF33C965C79C",
+                                    .securityOrigin = URL_BASE,
+                                    .url = bc.getURL() orelse "about:blank",
+                                    .secureContextType = "Secure",
+                                },
+                            },
+                        }, .{});
+                    }
+                }
+                // No real page yet - return the synthetic STARTUP frame ID.
                 return command.sendResult(.{
                     .frameTree = .{
                         .frame = .{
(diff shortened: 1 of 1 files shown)


src/cdp/cdp.zig
```diff
@@ -191,6 +191,25 @@ pub fn CDPT(comptime TypeProvider: type) type {
             // Stagehand parses the response and error if we don't return a
             // correct one for this call.
             if (std.mem.eql(u8, method, "Page.getFrameTree")) {
+                // If a real page already exists (connectOverCDP flow), return
+                // its actual frame ID so that subsequent lifecycle events
+                // (frameNavigated, etc.) match the driver's frame tracking.
+                if (command.cdp.browser_context) |*bc| {
```
krichprollsch, line 197: What about dispatching the command directly in this case? all the prerequisites of `Page.getFrameTree` are here: a browser context and a target id.

```zig
                if (command.cdp.browser_context) |*bc| {
                    if (bc.target_id != null) {
                        return dispatchCommand(command, method);
                    }
                }
```

Tests should be adjusted.

krichprollsch:
I clean up a bit more directly into #1949

mvanhorn:
Thanks for picking this up and cleaning it further in #1949, closing in favor of that.

PR #1934, opened 2026-03-20
HTTP Caching

mookums:
This adds caching at the HTTP level, allowing the use of a local directory as a RFC 9111 cache.

### DIFF
--- a/src/browser/HttpClient.zig
+++ b/src/browser/HttpClient.zig
@@ -964,7 +964,7 @@ fn processOneMessage(self: *Client, msg: http.Handles.MultiMessage, transfer: *T
 
     if (transfer._pending_cache_metadata) |metadata| {
         const cache = &self.network.cache.?;
-        cache.put(metadata, body) catch |err| {
+        cache.put(metadata.*, body) catch |err| {
             log.warn(.cache, "cache put failed", .{ .err = err });
         };
     }
@@ -1164,7 +1164,7 @@ pub const Transfer = struct {
     // total bytes received in the response, including the response status line,
     // the headers, and the [encoded] body.
     bytes_received: usize = 0,
-    _pending_cache_metadata: ?CacheMetadata = null,
+    _pending_cache_metadata: ?*CacheMetadata = null,
 
     aborted: bool = false,
 
@@ -1587,9 +1587,11 @@ pub const Transfer = struct {
                     }
                 }
 
-                transfer._pending_cache_metadata = cm;
-                transfer._pending_cache_metadata.?.headers = header_list.items[0..end_of_response];
-                transfer._pending_cache_metadata.?.vary_headers = header_list.items[end_of_response..];
+                const metadata = try transfer.arena.allocator().create(CacheMetadata);
+                metadata.* = cm;
+                metadata.headers = header_list.items[0..end_of_response];
+                metadata.vary_headers = header_list.items[end_of_response..];
+                transfer._pending_cache_metadata = metadata;
             }
         }
 


src/network/cache/FsCache.zig
```diff
@@ -0,0 +1,301 @@
...
+
+    const dir = try cwd.openDir(path, .{ .iterate = true });
+    return .{ .dir = dir };
+}
+
+pub fn deinit(self: *FsCache) void {
+    self.dir.close();
+}
+
+pub fn cache(self: *FsCache) Cache {
```
karlseguin: Is this used?

src/network/cache/FsCache.zig
```diff
@@ -0,0 +1,301 @@
...
+    return Cache.init(self);
+}
+
+const HASHED_KEY_LEN = 16;
+const HASHED_PATH_LEN = HASHED_KEY_LEN + 5;
+const HASHED_TMP_PATH_LEN = HASHED_PATH_LEN + 4;
+
+fn hashKey(key: []const u8) [HASHED_KEY_LEN]u8 {
+    const h = std.hash.Wyhash.hash(0, key);
```
karlseguin: Wyhash is a general purpose hashing function, you'll get collisions, which could result in the wrong cached page being returned.

src/network/cache/FsCache.zig
```diff
@@ -0,0 +1,301 @@
...
+    var hex: [HASHED_KEY_LEN]u8 = undefined;
+    _ = std.fmt.bufPrint(&hex, "{x:0>16}", .{h}) catch unreachable;
+    return hex;
+}
+
+fn serializeMeta(writer: *std.Io.Writer, meta: *const CachedMetadata) !void {
```
karlseguin: A lot of writing / parsing code. Did you consider either using zon or, if space is an issue, a json array?

Whatever you do, I'd add a version number as the first field in the header.

src/network/cache/FsCache.zig
```diff
@@ -0,0 +1,301 @@
...
+            self.dir.deleteFile(&meta_tmp_path) catch {};
+        }
+
+        var buf: [512]u8 = undefined;
+        var meta_file_writer = meta_file.writer(&buf);
+        try serializeMeta(&meta_file_writer.interface, &meta);
+        meta_file.close();
+    }
+    errdefer self.dir.deleteFile(&meta_tmp_path) catch {};
+    try self.dir.rename(&meta_tmp_path, &meta_path);
```
karlseguin: I think there are error (and panic)cases where the meta file sticks around. Not sure how problematic it is to have a meta but no body file.

What are the advantages of splitting this up into 2 files? I feel that a single file would have less edge-cases (e.g. successful write to one file, but not the other) and could result in fewer system calls.

src/network/cache/FsCache.zig
```diff
@@ -0,0 +1,301 @@
...
+
+    var body_tmp_path: [HASHED_TMP_PATH_LEN]u8 = undefined;
+    _ = std.fmt.bufPrint(&body_tmp_path, "{s}.body.tmp", .{hashed_key}) catch
+        @panic("FsCache.put body tmp path overflowed");
+
+    {
+        const body_file = try self.dir.createFile(&body_tmp_path, .{});
+        errdefer {
+            body_file.close();
+            self.dir.deleteFile(&body_tmp_path) catch {};
```
karlseguin: Can writeAll result in a partial write? Because, if you do partially write, and then this fails, you now have garbage in the file that you might fetch in `get`

src/network/cache/Cache.zig
```diff
@@ -0,0 +1,213 @@
...
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+const std = @import("std");
+const Http = @import("../http.zig");
+const FsCache = @import("FsCache.zig");
+
+/// A browser-wide cache for resources across the network.
+/// This mostly conforms to RFC9111 with regards to caching behavior.
+pub const Cache = @This();
```
karlseguin, line 26: I think everything in here has to be thread safe?

src/network/cache/Cache.zig
```diff
@@ -0,0 +1,213 @@
...
+        allocator.free(self.headers);
+        if (self.vary) |v| v.deinit(allocator);
+        if (self.etag) |e| allocator.free(e);
+        if (self.last_modified) |lm| allocator.free(lm);
+    }
+};
+
+pub const CachedData = union(enum) {
+    buffer: []const u8,
+    file: std.fs.File,
```
karlseguin: To be more generic, could this be an io.Reader?

src/browser/HttpClient.zig
```diff
@@ -289,7 +293,62 @@ pub fn request(self: *Client, req: Request) !void {
...
+        },
+        .file => |file| {
+            var buf: [1024]u8 = undefined;
+            var file_reader = file.reader(&buf);
+
+            const reader = &file_reader.interface;
+            var read_buf: [1024]u8 = undefined;
+
+            while (true) {
+                const curr = try reader.readSliceShort(&read_buf);
```
karlseguin: Should errors reading the file call the `error_callback`?

src/browser/HttpClient.zig
```diff
@@ -992,16 +1094,91 @@ pub const Request = struct {
...
+
+    pub fn terminate(self: Response) void {
+        switch (self.inner) {
+            .live => |live| live.terminate(),
+            .cached => {},
+        }
+    }
+};
+
+pub const LiveTransfer = struct {
```
karlseguin: I think we can keep the name `Transfer`.

src/network/cache/Cache.zig
```diff
@@ -0,0 +1,213 @@
...
+
+/// A browser-wide cache for resources across the network.
+/// This mostly conforms to RFC9111 with regards to caching behavior.
+pub const Cache = @This();
+
+kind: union(enum) {
+    fs: FsCache,
+},
+
+pub fn get(self: *Cache, allocator: std.mem.Allocator, key: []const u8) ?CachedResponse {
```
karlseguin: Does anything enforce expiration?

src/browser/HttpClient.zig
```diff
@@ -289,7 +293,62 @@ pub fn request(self: *Client, req: Request) !void {
...
+        if (req.method == .GET) {
+            if (cache.get(self.allocator, req.url)) |cached| {
+                log.debug(.browser, "http.cache.get", .{
+                    .url = req.url,
+                    .found = true,
+                    .metadata = cached.metadata,
+                });
+
+                defer req.headers.deinit();
+                return serveFromCache(self.allocator, req, &cached);
```
karlseguin: Would be nice to come up with an allocation strategy that doesn't rely on this allocator.  Every cacheable GET ends up creating / freeing a lot of small allocations on the main allocation, with no re-use.

src/browser/HttpClient.zig
```diff
@@ -289,7 +293,62 @@ pub fn request(self: *Client, req: Request) !void {
...
+    defer cached.metadata.deinit(allocator);
+
+    if (req.start_callback) |cb| {
+        try cb(response);
+    }
+
+    const proceed = try req.header_callback(response);
+    if (!proceed) {
+        req.error_callback(req.ctx, error.Abort);
+        return;
```
karlseguin, line 331: Isn't there potentially a file or buffer that's still opened?

src/browser/HttpClient.zig
```diff
@@ -848,6 +908,23 @@ fn processMessages(self: *Client) !bool {
             }
         }
 
+        const allocator = transfer.arena.allocator();
```
karlseguin: I think this is only needed if the response will be cached? No need to create it before we know we'll cache it?

src/browser/HttpClient.zig
```diff
@@ -875,23 +952,48 @@ fn processMessages(self: *Client) !bool {
...
-            processed = true;
+            cache: {
+                if (self.network.cache) |*cache| {
+                    const headers = &transfer.response_header.?;
+
+                    const metadata = try CacheMetadata.fromHeaders(
+                        transfer.req.url,
+                        headers.status,
+                        std.time.timestamp(),
+                        header_list.items,
```
karlseguin: Could pass in the iterator rather than collecting it + iterating the collection?

src/browser/HttpClient.zig
```diff
@@ -875,23 +952,48 @@ fn processMessages(self: *Client) !bool {
...
                 requestFailed(transfer, err, true);
                 continue;
             };
 
-            transfer.req.notification.dispatch(.http_request_done, &.{
-                .transfer = transfer,
-            });
-            processed = true;
+            cache: {
+                if (self.network.cache) |*cache| {
```
karlseguin: Should check the method?

src/network/cache/Cache.zig
```diff
@@ -0,0 +1,213 @@
...
+    return switch (self.kind) {
+        inline else => |*c| c.get(allocator, key),
+    };
+}
+
+pub fn put(self: *Cache, key: []const u8, metadata: CachedMetadata, body: []const u8) !void {
```
karlseguin: Cache key should be derived from metadata?

src/network/cache/Cache.zig
karlseguin: instead of `key: []const u8`, this could probably take a struct, `req: CacheRequest`

```zig
const struct CacheRequest = struct {
  url: []const u8,
}
```

Because I feel like you'll want more things in the future, e.g. method...and the caller shouldn't know about "cache key".

src/browser/HttpClient.zig
```diff
@@ -875,23 +952,48 @@ fn processMessages(self: *Client) !bool {
...
+                        transfer.req.url,
+                        headers.status,
+                        std.time.timestamp(),
+                        header_list.items,
+                    ) orelse break :cache;
+
+                    // TODO: Support Vary Keying
+                    const cache_key = transfer.req.url;
+
+                    log.err(.browser, "http cache", .{ .key = cache_key, .metadata = metadata });
```
karlseguin:
```suggestion
                    log.debug(.browser, "http cache", .{ .key = cache_key, .metadata = metadata });
```

src/browser/HttpClient.zig
```diff
@@ -992,16 +1094,91 @@ pub const Request = struct {
...
     arena: ArenaAllocator,
     id: u32 = 0,
     req: Request,
     url: [:0]const u8,
-    ctx: *anyopaque, // copied from req.ctx to make it easier for callback handlers
     client: *Client,
     // total bytes received in the response, including the response status line,
     // the headers, and the [encoded] body.
     bytes_received: usize = 0,
+    body: std.ArrayListUnmanaged(u8) = .empty,
```
karlseguin:
```suggestion
    body: std.ArrayList(u8) = .empty,
```

src/browser/HttpClient.zig
```diff
@@ -1428,7 +1605,12 @@ pub const Transfer = struct {
         }
 
         const chunk = buffer[0..chunk_len];
-        transfer.req.data_callback(transfer, chunk) catch |err| {
+        transfer.body.appendSlice(transfer.arena.allocator(), chunk) catch |err| {
```
karlseguin: Not necessary if we aren't going to be caching this response, which we should know whether or not we will at this point.

src/network/cache/Cache.zig
```diff
@@ -0,0 +1,213 @@
...
+                content_type = hdr.value;
+            } else if (std.ascii.eqlIgnoreCase(hdr.name, "set-cookie")) {
+                has_set_cookie = true;
+            } else if (std.ascii.eqlIgnoreCase(hdr.name, "authorization")) {
+                has_authorization = true;
+            }
+        }
+
+        // no-store: must not be stored
+        if (cc.no_store) return null;
```
karlseguin: we could exit the loop immediately if that's the case? Don't even need to store this in CacheControl:

```zig
cc = CacheControl.parse(hdr.value) orelse return null;
```

That's true for all exit paths.

src/network/cache/Cache.zig
```diff
@@ -0,0 +1,213 @@
...
+        if (cc.max_age == null) return null;
+
+        // Set-Cookie without explicit public
+        if (has_set_cookie and !cc.is_public) return null;
+
+        // Authorization header without explicit public or s-maxage
+        if (has_authorization and !cc.is_public and cc.s_maxage == null) return null;
+
+        // Only cache 200 for now. Technically, we can cache others.
+        switch (status) {
```
karlseguin: Entire body was buffered and headers parsed..this should be done much earlier.

src/network/cache/Cache.zig
```diff
@@ -0,0 +1,213 @@
...
+            .etag = etag,
+            .last_modified = last_modified,
+            .cache_control = cc,
+            .vary = vary,
+            .headers = headers,
+        };
+    }
+
+    pub fn deinit(self: CachedMetadata, allocator: std.mem.Allocator) void {
+        allocator.free(self.url);
```
karlseguin: Arena, ideally from the arenapool.

src/network/cache/Cache.zig
```diff
@@ -0,0 +1,213 @@
...
+        while (iter.next()) |part| {
+            const directive = std.mem.trim(u8, part, &std.ascii.whitespace);
+            if (std.ascii.eqlIgnoreCase(directive, "no-store")) {
+                cc.no_store = true;
+            } else if (std.ascii.eqlIgnoreCase(directive, "no-cache")) {
+                cc.no_cache = true;
+            } else if (std.ascii.eqlIgnoreCase(directive, "must-revalidate")) {
+                cc.must_revalidate = true;
+            } else if (std.ascii.eqlIgnoreCase(directive, "immutable")) {
+                cc.immutable = true;
```
karlseguin: We don't use a a lot of these. Should we? If not, should we bother parsing them and storing them in the meta?

src/network/cache/Cache.zig
```diff
@@ -0,0 +1,213 @@
...
+        if (cc.no_store) return null;
+
+        // Vary: * means the response cannot be cached
+        if (vary) |v| if (v == .wildcard) return null;
+
+        // must have an explicit max-age to be cacheable
+        if (cc.max_age == null) return null;
+
+        // Set-Cookie without explicit public
+        if (has_set_cookie and !cc.is_public) return null;
```
karlseguin: We should never cache anything that isn't public.  While we're the user-agent, the cache doesn't have any isolation and acts much more like a reverse proxy's cache.

src/network/cache/Cache.zig
```diff
@@ -0,0 +1,213 @@
...
+            } else if (std.ascii.eqlIgnoreCase(directive, "immutable")) {
+                cc.immutable = true;
+            } else if (std.ascii.eqlIgnoreCase(directive, "public")) {
+                cc.is_public = true;
+            } else if (std.ascii.startsWithIgnoreCase(directive, "max-age=")) {
+                cc.max_age = std.fmt.parseInt(u64, directive[8..], 10) catch null;
+            } else if (std.ascii.startsWithIgnoreCase(directive, "s-maxage=")) {
+                cc.s_maxage = std.fmt.parseInt(u64, directive[9..], 10) catch null;
+                // s-maxage takes precedence over max-age
+                cc.max_age = cc.s_maxage orelse cc.max_age;
```
karlseguin: we only serialize `max_age`, so if the header is defined as `s-maxage=1000,max-age=0`  it will do the opposite and `max-age` will take precedence over `s-maxage`.

src/network/cache/FsCache.zig
```diff
@@ -0,0 +1,301 @@
...
+        meta.stored_at,
+        meta.age_at_store,
+    });
+    try writer.print("{s}\n", .{meta.etag orelse "null"});
+    try writer.print("{s}\n", .{meta.last_modified orelse "null"});
+
+    // cache-control
+    try writer.print("{d}\n", .{meta.cache_control.max_age orelse 0});
+    try writer.print("{}\n{}\n{}\n{}\n", .{
+        meta.cache_control.max_age != null,
```
karlseguin: do we ever cache something with no max-age? What does a max-age of 0 mean?

src/Config.zig
```diff
@@ -156,6 +156,13 @@ pub fn userAgentSuffix(self: *const Config) ?[]const u8 {
     };
 }
 
+pub fn cacheDir(self: *const Config) ?[]const u8 {
```
mrdimidium: This should have a default folder, something like `$XDG_CACHE_DIR/lightpanda`

src/browser/ScriptManager.zig
```diff
@@ -694,85 +694,85 @@ pub const Script = struct {
...
+        //     self.debug_transfer_id = transfer.id;
+        //     self.debug_transfer_tries = transfer._tries;
+        //     self.debug_transfer_aborted = transfer.aborted;
+        //     self.debug_transfer_bytes_received = transfer.bytes_received;
+        //     self.debug_transfer_notified_fail = transfer._notified_fail;
+        //     self.debug_transfer_redirecting = transfer._redirecting;
+        //     self.debug_transfer_intercept_state = @intFromEnum(transfer._intercept_state);
+        //     self.debug_transfer_auth_challenge = transfer._auth_challenge != null;
+        //     self.debug_transfer_easy_id = if (transfer._conn) |c| @intFromPtr(c.easy) else 0;
+        // }
```
mrdimidium: Why are you removing this log?

src/browser/HttpClient.zig
mrdimidium: At least to reduce the number of conflicts

src/network/cache/FsCache.zig
mookums: Did seperate so you could check the metadata independently without loading the whole body and it allowed us to stream the body instead of loading it entirely at once.

karlseguin: You woulnd't need to load the whole body. Prefix the header with a length, read a 4096 chunk, slice to the length, and then seek to the end of the header. Admittedly, that's a bit of code, but I'm pretty sure it'll be less code than then 2 file dance, with fewer cleanup edge cases and better performance.

src/network/cache/FsCache.zig
mookums: Using JSON for the metadata

src/network/cache/Cache.zig
mookums: Originally tried this but storing the file separately felt gross. I think we would also either need to also provide and store the buffer being used by the Reader.

src/network/cache/Cache.zig
mookums: Removed most of them. Will probably remove must_revalidate and immutable too.

src/network/cache/FsCache.zig
mookums: I believe you can store something with no max-age (and then you depend on the `Expires` header but max-age takes precedence which this doesn't support rn). `max-age` of 0 means that the response is immediately stale so must be refetched.

src/Config.zig
mookums: Kept it as optional as a way to conditionally use the FsCache or not. The plan was to have different Cache backends in the future so shouldn't automatically subscribe to one was the idea.

src/browser/ScriptManager.zig
karlseguin: ☝️

src/network/Network.zig
```diff
@@ -244,8 +253,10 @@ pub fn init(allocator: Allocator, config: *const Config) !Runtime {
         .available = available,
         .connections = connections,
 
+        .app = app,
         .robot_store = RobotStore.init(allocator),
         .web_bot_auth = web_bot_auth,
+        .cache = cache,
```
karlseguin, line 273: This isn't deinit'd.

src/network/cache/FsCache.zig
```diff
@@ -0,0 +1,293 @@
...
+dir: std.fs.Dir,
+locks: [LOCK_STRIPES]std.Thread.Mutex = .{std.Thread.Mutex{}} ** LOCK_STRIPES,
+
+const CacheMetadataFile = struct {
+    version: usize,
+    metadata: CachedMetadata,
+};
+
+fn getLockPtr(self: *FsCache, key: *const [HASHED_KEY_LEN]u8) *std.Thread.Mutex {
+    const lock_idx: usize = @truncate(std.hash.Wyhash.hash(0, key) % LOCK_STRIPES);
```
karlseguin: I don't think you need the `@truncate`.

Fun fact, `std.hash.Wyhash.hash(0, key) & (LOCK_STRIPES - 1)` will perform faster (`&` is much cheaper than `%`, but will only work when `LOCK_STRIPES` is a power of 2 (which it is, but that isn't enforced).

src/network/cache/FsCache.zig
```diff
@@ -0,0 +1,293 @@
...
+    const body_p = bodyPath(&hashed_key);
+
+    const lock = self.getLockPtr(&hashed_key);
+    lock.lock();
+    defer lock.unlock();
+
+    const meta_file = self.dir.openFile(&meta_p, .{ .mode = .read_only }) catch return null;
+    defer meta_file.close();
+
+    const contents = meta_file.readToEndAlloc(arena, 1 * 1024 * 1024) catch return null;
```
karlseguin: So we'll cache a file that's > 1MB, but we wont' serve it from the cache? Seems like it makes more sense to have the max length check on the put.

src/network/cache/FsCache.zig
```diff
@@ -0,0 +1,293 @@
...
+    if (age < 0 or @as(u64, @intCast(age)) >= metadata.cache_control.max_age) {
+        self.dir.deleteFile(&meta_p) catch {};
+        self.dir.deleteFile(&body_p) catch {};
+        return null;
+    }
+
+    const body_file = self.dir.openFile(
+        &body_p,
+        .{ .mode = .read_only },
+    ) catch return null;
```
karlseguin: I feel like almost all of these `catch return null` and `catch {}` should log. Possibly a new log scope, `cache`. I just see trying to troubleshoot issues as being made complicated by all these silent failures.

src/network/cache/FsCache.zig
```diff
@@ -0,0 +1,293 @@
...
+    self.dir.close();
+}
+
+pub fn get(self: *FsCache, arena: std.mem.Allocator, req: CacheRequest) ?Cache.CachedResponse {
+    const hashed_key = hashKey(req.url);
+    const meta_p = metaPath(&hashed_key);
+    const body_p = bodyPath(&hashed_key);
+
+    const lock = self.getLockPtr(&hashed_key);
+    lock.lock();
```
karlseguin, line 94: Any thoughts on a rwlock?

src/network/cache/FsCache.zig
```diff
@@ -0,0 +1,293 @@
...
+        meta_file.close();
+    }
+    errdefer self.dir.deleteFile(&meta_tmp_p) catch {};
+    try self.dir.rename(&meta_tmp_p, &meta_p);
+
+    {
+        const body_file = try self.dir.createFile(&body_tmp_p, .{});
+        errdefer {
+            body_file.close();
+            self.dir.deleteFile(&body_tmp_p) catch {};
```
karlseguin: This is a good example where failing to delete this file, which might have partial data, could result in very difficult to diagnose issues and should 100% be logged.

src/browser/webapi/net/Fetch.zig
```diff
@@ -145,25 +144,24 @@ fn httpHeaderDoneCallback(transfer: *HttpClient.Transfer) !bool {
...
     }
 
-    res._status = header.status;
-    res._status_text = std.http.Status.phrase(@enumFromInt(header.status)) orelse "";
-    res._url = try arena.dupeZ(u8, std.mem.span(header.url));
-    res._is_redirected = header.redirect_count > 0;
+    res._status = response.status().?;
+    res._status_text = std.http.Status.phrase(@enumFromInt(response.status().?)) orelse "";
+    res._url = try arena.dupeZ(u8, response.url());
+    res._is_redirected = response.redirectCount().? > 0;
```
karlseguin, line 165: Under what condition does `redirectCount()` return null, and why is it always safe to unwrap here?

src/browser/webapi/net/Fetch.zig
```diff
@@ -145,25 +144,24 @@ fn httpHeaderDoneCallback(transfer: *HttpClient.Transfer) !bool {
...
-            .status = header.status,
+            .status = response.status(),
         });
     }
 
-    res._status = header.status;
-    res._status_text = std.http.Status.phrase(@enumFromInt(header.status)) orelse "";
-    res._url = try arena.dupeZ(u8, std.mem.span(header.url));
-    res._is_redirected = header.redirect_count > 0;
+    res._status = response.status().?;
```
karlseguin, line 162: Under what condition does `status()` always return null and why is it safe to unwrap here?

src/browser/HttpClient.zig
```diff
@@ -311,7 +293,69 @@ pub fn request(self: *Client, req: Request) !void {
...
+    }
+
+    switch (cached.data) {
+        .buffer => |data| {
+            if (data.len > 0) {
+                try req.data_callback(response, data);
+            }
+        },
+        .file => |file| {
+            defer file.close();
```
karlseguin: File is never closed on any error that happens before we reach this point. Where serveIsCached is called, can we do:

```zig
if (cache.get(arena, .{ .url = req.url, .timestamp = std.time.timestamp() })) |cached| {
    // existing log message...
    defer cache.deinit()   // <-- ADD
    defer req.headers.deinit();
    return serveFromCache(req, &cached);
}
```

to make sure we close the file in all cases?

src/browser/HttpClient.zig
```diff
@@ -311,7 +293,69 @@ pub fn request(self: *Client, req: Request) !void {
...
+            var buf: [1024]u8 = undefined;
```
karlseguin, line 343: Some JS files are in the MB, so this seems like a small buffer.

src/browser/webapi/net/Fetch.zig
mookums: Pretty sure this (and all of the other ones in here) are directly adapted from the existing Fetch.zig code that does this unwrapped (it just unwraps at the response_header layer while with the new `Response`, you have to unwrap at the field). These are the httpHeaderDoneCallback so all of these should reasonably exist?

src/network/cache/FsCache.zig
mookums: Not opposed but I think we can make some changes to avoid having a lock at all? On Linux (and MacOS I think), rename is atomic (and the open files just read from the old inode). We already use a tmp file when writing so if we just incorporate an atomic id (or thread id) to the tmp file name, this would be thread-safe from my understanding.

karlseguin: According to `https://danluu.com/file-consistency/` that's only true in normal cases, but not during crashes.

src/network/cache/FsCache.zig
```diff
@@ -0,0 +1,580 @@
...
+    lock.lock();
+    defer lock.unlock();
+
+    const file = self.dir.openFile(&cache_p, .{ .mode = .read_only }) catch |e| {
+        switch (e) {
+            std.fs.File.OpenError.FileNotFound => {
+                log.debug(.cache, "miss", .{ .url = req.url, .hash = &hashed_key });
+            },
+            else => |err| {
+                log.warn(.cache, "open file err", .{ .url = req.url, .err = err });
```
karlseguin, line 103: 👍

src/browser/HttpClient.zig
```diff
@@ -1101,7 +1296,7 @@ pub const Transfer = struct {
         });
 
         if (execute_callback) {
-            self.req.error_callback(self.ctx, err);
+            self.req.error_callback(self.req.ctx, err);
```
karlseguin, line 1292: 2 lines under this, it's still `self.ctx`

krichprollsch:
@mookums can you update the help command to list `--cache-dir` option.
BTW we moved from `_` to `-` separator for command line options, so rename `--cache_dir` into `--cache-dir`. No need to support `--cache_dir` for this new option.

krichprollsch:
I've got a crash during `intgegration` test suite:
```
$ go run integration/main.go
=== OK  6.363657507s    node integration/duckduckgo.js
=== OK  3.130629926s    node integration/algolia.js
=== OK  6.311556639s    node integration/github.js
=== ERR node integration/xange.js
=== ERR node integration/readme_amiibo.js
=== ERR node integration/quickstart_wikipedia.js
=== ERR node integration/quickstart_hn.js
=== ERR node integration/mastodon.js
=== ERR node integration/old-reddit.js
=== ERR node integration/reddit.js
=== ERR node integration/google_news.js
=== ERR node integration/google_news_redirection.js
=== ERR node integration/bing.js
=== ERR node integration/anthropic_docs.js
11 failures
exit status 1
```

```
Segmentation fault at address 0x7d7b48e8203e
/usr/local/zig-0.15.2/lib/std/mem.zig:1815:5: 0x37eb5b6 in round (lightpanda)
    const value: T = @bitCast(buffer.*);
    ^
/usr/local/zig-0.15.2/lib/std/crypto/sha2.zig:128:24: 0x37050a8 in update (lightpanda)
                d.round(b[off..][0..64]);
                       ^
/usr/local/zig-0.15.2/lib/std/crypto/sha2.zig:110:21: 0x35c6b66 in hash (lightpanda)
            d.update(b);
                    ^
/home/pierre/wrk/browser/src/network/cache/FsCache.zig:55:37: 0x34e89b9 in hashKey (lightpanda)
    std.crypto.hash.sha2.Sha256.hash(key, &digest, .{});
                                    ^
/home/pierre/wrk/browser/src/network/cache/FsCache.zig:201:31: 0x34e82ba in put (lightpanda)
    const hashed_key = hashKey(meta.url);
                              ^
/home/pierre/wrk/browser/src/network/cache/Cache.zig:45:34: 0x34e8fcc in put (lightpanda)
        inline else => |*c| c.put(metadata, body),
                                 ^
/home/pierre/wrk/browser/src/browser/HttpClient.zig:986:18: 0x34e9f9e in processOneMessage (lightpanda)
        cache.put(metadata, body) catch |err| {
                 ^
/home/pierre/wrk/browser/src/browser/HttpClient.zig:1003:44: 0x34eadec in processMessages (lightpanda)
        const done = self.processOneMessage(msg, transfer) catch |err| blk: {
                                           ^
/home/pierre/wrk/browser/src/browser/HttpClient.zig:819:33: 0x2e7c8a0 in perform (lightpanda)
    if (try self.processMessages()) {
                                ^
/home/pierre/wrk/browser/src/browser/HttpClient.zig:284:24: 0x2e7c345 in tick (lightpanda)
    return self.perform(@intCast(timeout_ms));
                       ^
/home/pierre/wrk/browser/src/browser/Runner.zig:230:53: 0x38a0302 in _tick__anon_425706 (lightpanda)
            const http_result = try http_client.tick(@intCast(@min(opts.ms, ms_to_wait)));
                                                    ^
/home/pierre/wrk/browser/src/browser/Runner.zig:78:39: 0x38a0557 in _wait__anon_425702 (lightpanda)
        const tick_result = self._tick(is_cdp, tick_opts) catch |err| {
                                      ^
/home/pierre/wrk/browser/src/browser/Runner.zig:66:22: 0x38a0885 in waitCDP (lightpanda)
    return self._wait(true, opts);
                     ^
/home/pierre/wrk/browser/src/cdp/CDP.zig:133:26: 0x38a0a46 in pageWait (lightpanda)
    return runner.waitCDP(.{ .ms = ms });
```

src/browser/HttpClient.zig
```diff
@@ -311,7 +314,81 @@ pub fn request(self: *Client, req: Request) !void {
...
+
+            if (cache.get(arena, .{
+                .url = req.url,
+                .timestamp = std.time.timestamp(),
+                .request_headers = req_header_list.items,
+            })) |cached| {
+                log.debug(.browser, "http.cache.get", .{
+                    .url = req.url,
+                    .found = true,
+                    .metadata = cached.metadata,
```
krichprollsch: Logging meta data this way is ueless:
```
      metadata = .{ .url = { 104, 116, 116, 112, 115, 58, 47, 47, 119, 119, 119, 46, 103, 115, 116, 97, 116, 105, 99, 46, 99, 111, 109, 47, 95, 47, 109, 115, 115, 47, 98, 111, 113, 45, 100, 111, 116, 115, 47, 95, 47, 106, 115, 47, 107, 61, 98, 111, 113, 4
5, 100, 111, 116, 115, 46, 68, 111, 116, 115, 83, 112, 108, 97, 115, 104, 85, 105, 95, 100, 101, 115, 107, 116, 111, 112, 95, 109, 115, 46, 101, 110, 95, 85, 83, 46, 78, 49, 113, 95, 122, 45, 45, 89, 85, 115, 73, 46, 50, 48, 49, 56, 46, 79, 47, 99, 107, 6
1, 98, 111, 113, 45, 100, 111, 116, 115, 46, 68, 111, 116, 115, 83, 112, 108, 97, 115, 104, 85, 105, 95, 100, 101, 115, 107, 116, 111, 112, 95, 109, 115, 46, 118, 84, 100, 53, 106, 50, 98, 104, 90, 121, 99, 46, 76, 46, 88, 46, 79, 47, 97, 109, 61, 71, 65,
 73, 70, 65, 65, 65, 65, 65, 74, 103, 70, 66, 104, 103, 47, 100, 61, 49, 47, 101, 120, 109, 61, 65, 55, 102, 67, 85, 44, 66, 66, 73, 55, 52, 44, 66, 86, 103, 113, 117, 102, 44, 66, 101, 74, 89, 116, 102, 44, 66, 111, 70, 107, 79, 44, 67, 76, 68, 76, 97, 1
00, 44, 68, 75, 90, 118, 69, 44, 69, 66, 87, 48, 117, 99, 44, 69, 70, 81, 55, 56, 99, 44, 70, 98, 99, 120, 82, 99, 44, 70, 121, 108, 121, 87, 101, 44, 72, 52, 82, 81, 115, 100, 44, 73, 90, 84, 54, 51, 44, 73, 97, 107, 101, 48, 44, 74, 78, 111, 120, 105, 4
4, 74, 88, 78, 56, 55, 44, 74, 102, 53, 119, 75, 102, 44, 74, 110, 116, 122, 65, 101, 44, 75, 57, 57, 113, 89, 44, 75, 85, 77, 55, 90, 44, 76, 69, 105, 107, 90, 101, 44, 76, 118, 71, 104, 114, 102, 44, 77, 73, 54, 107, 55, 99, 44, 77, 100, 85, 122, 85, 10
1, 44, 77, 112, 74, 119, 90, 99, 44, 77, 113, 57, 110, 48, 99, 44, 77, 120, 86, 122, 118, 100, 44, 78, 119, 72, 48, 72, 44, 79, 49, 71, 106, 122, 101, 44, 79, 54, 121, 56, 101, 100, 44, 79, 84, 65, 51, 65, 101, 44, 79, 109, 99, 99, 69, 98, 44, 80, 54, 115
, 81, 79, 99, 44, 80, 72, 85, 73, 121, 98, 44, 80, 107, 51, 74, 111, 44, 80, 114, 80, 89, 82, 100, 44, 81, 73, 104, 70, 114, 44, 82, 76, 119, 49, 57, 44, 82, 77, 104, 66, 102, 101, 44, 82, 106, 83, 87, 80, 99, 44, 82, 113, 106, 85, 76, 100, 44, 82, 121, 1
18, 97, 85, 98, 44, 83, 100, 99, 119, 72, 98, 44, 83, 104, 65, 106, 97, 102, 44, 83, 112, 115, 102, 83, 98, 44, 85, 52, 72, 112, 48, 100, 44, 85, 85, 74, 113, 86, 101, 44, 85, 88, 74, 104, 85, 100, 44, 85, 97, 115, 57, 72, 100, 44, 85, 108, 109, 109, 114,
 100, 44, 86, 51, 100, 68, 79, 98, 44, 86, 107, 106, 100, 72, 100, 44, 87, 79, 57, 101, 101, 44, 87, 104, 112, 122, 112, 101, 44, 87, 120, 107, 89, 57, 44, 88, 86, 77, 78, 118, 100, 44, 89, 81, 77, 114, 111, 99, 44, 89, 83, 48, 70, 70, 100, 44, 89, 89, 10
9, 72, 122, 98, 44, 90, 53, 117, 76, 108, 101, 44, 90, 68, 90, 99, 114, 101, 44, 90, 110, 76, 100, 104, 99, 44, 90, 119, 68, 107, 57, 100, 44, 95, 98, 44, 95, 116, 112, 44, 97, 54, 112, 113, 49, 44, 97, 87, 51, 112, 89, 44, 97, 88, 121, 85, 101, 99, 44, 9
8, 121, 102, 84, 79, 98, 44, 100, 117, 70, 81, 70, 99, 44, 101, 53, 113, 70, 76, 99, 44, 101, 89, 110, 121, 72, 44, 101, 98, 90, 51, 109, 98, 44, 101, 100, 101, 82, 103, 44, 103, 50, 74, 68, 75, 102, 44, 103, 107, 105, 122, 76, 101, 44, 103, 121, 99, 104,
 103, 44, 104, 75, 83, 107, 51, 101, 44, 104, 99, 54, 85, 98, 100, 44, 107, 51, 88, 110, 84, 101, 44, 107, 106, 75, 100, 88, 101, 44, 108, 97, 122, 71, 55, 98, 44, 108, 115, 106, 86, 109, 99, 44, 108, 119, 79, 106, 83, 98, 44, 108, 119, 100, 100, 107, 102
, 44, 109, 57, 111, 86, 44, 109, 73, 51, 76, 70, 98, 44, 109, 100, 82, 55, 113, 44, 109, 105, 118, 72, 111, 44, 110, 55, 51, 113, 119, 102, 44, 110, 75, 117, 70, 112, 98, 44, 111, 111, 118, 105, 51, 100, 44, 111, 116, 121, 76, 81, 100, 44, 112, 51, 104, 1
09, 82, 99, 44, 112, 77, 88, 68, 122, 100, 44, 112, 106, 73, 67, 68, 101, 44, 112, 119, 55, 48, 71, 99, 44, 113, 48, 120, 84, 105, 102, 44, 114, 121, 78, 66, 112, 44, 115, 79, 88, 70, 106, 44, 115, 111, 72, 120, 102, 44, 117, 56, 81, 102, 98, 100, 44, 119
, 57, 104, 68, 118, 44, 119, 70, 69, 117, 87, 100, 44, 119, 103, 49, 80, 54, 98, 44, 119, 115, 57, 84, 108, 99, 44, 120, 81, 116, 90, 98, 44, 120, 85, 100, 105, 112, 102, 44, 120, 104, 103, 75, 72, 44, 120, 115, 70, 113, 99, 102, 44, 121, 119, 79, 82, 53,
 99, 44, 122, 83, 75, 67, 112, 100, 44, 122, 98, 77, 76, 51, 99, 44, 122, 114, 49, 106, 114, 98, 44, 122, 116, 104, 77, 54, 47, 101, 120, 99, 109, 61, 95, 98, 44, 95, 116, 112, 44, 104, 111, 109, 101, 118, 105, 101, 119, 47, 101, 100, 61, 49, 47, 119, 116
, 61, 50, 47, 117, 106, 103, 61, 49, 47, 114, 115, 61, 65, 76, 115, 48, 110, 50, 78, 45, 120, 117, 105, 74, 65, 97, 114, 112, 55, 50, 74, 109, 45, 99, 76, 121, 52, 85, 79, 51, 107, 69, 120, 118, 67, 119, 47, 101, 101, 61, 69, 86, 78, 104, 106, 102, 58, 11
2, 119, 55, 48, 71, 99, 59, 69, 109, 90, 50, 66, 102, 58, 122, 114, 49, 106, 114, 98, 59, 74, 115, 98, 78, 104, 99, 58, 88, 100, 56, 105, 85, 100, 59, 75, 53, 110, 89, 84, 100, 58, 90, 68, 90, 99, 114, 101, 59, 75, 70, 106, 116, 117, 98, 58, 122, 116, 104
, 77, 54, 59, 76, 66, 103, 82, 76, 99, 58, 83, 100, 99, 119, 72, 98, 59, 77, 101, 51, 50, 100, 100, 58, 77, 69, 101, 89, 103, 99, 59, 78, 74, 49, 114, 102, 101, 58, 113, 84, 110, 111, 66, 102, 59, 78, 80, 75, 97, 75, 58, 83, 100, 99, 119, 72, 98, 59, 78,
83, 69, 111, 88, 58, 101, 89, 110, 121, 72, 59, 80, 106, 112, 108, 117, 100, 58, 69, 69, 68, 79, 82, 98, 59, 81, 71, 82, 48, 103, 100, 58, 77, 108, 104, 109, 121, 59, 83, 78, 85, 110, 51, 58, 90, 119, 68, 107, 57, 100, 59, 83, 99, 73, 51, 89, 99, 58, 101,
 55, 72, 122, 103, 98, 59, 85, 118, 99, 56, 111, 58, 86, 68, 111, 118, 78, 99, 59, 89, 73, 90, 109, 82, 100, 58, 65, 49, 121, 110, 53, 100, 59, 90, 119, 73, 103, 71, 99, 58, 108, 119, 79, 106, 83, 98, 59, 97, 53, 54, 112, 78, 101, 58, 74, 69, 102, 67, 119
, 98, 59, 99, 69, 116, 57, 48, 98, 58, 119, 115, 57, 84, 108, 99, 59, 100, 73, 111, 83, 66, 98, 58, 83, 112, 115, 102, 83, 98, 59, 100, 111, 119, 73, 71, 98, 58, 101, 98, 90, 51, 109, 98, 59, 101, 66, 65, 101, 83, 98, 58, 122, 98, 77, 76, 51, 99, 59, 105,
 70, 81, 121, 75, 102, 58, 81, 73, 104, 70, 114, 59, 108, 79, 79, 48, 86, 100, 58, 79, 84, 65, 51, 65, 101, 59, 111, 71, 116, 65, 117, 99, 58, 115, 79, 88, 70, 106, 59, 112, 88, 100, 82, 89, 98, 58, 77, 100, 85, 122, 85, 101, 59, 113, 81, 69, 111, 79, 99,
 58, 75, 85, 77, 55, 90, 59, 113, 97, 102, 66, 80, 100, 58, 121, 68, 86, 86, 107, 98, 59, 113, 100, 100, 103, 75, 101, 58, 120, 81, 116, 90, 98, 59, 117, 112, 52, 90, 121, 98, 58, 113, 89, 49, 88, 101, 102, 59, 119, 82, 53, 70, 82, 98, 58, 79, 49, 71, 106
, 122, 101, 59, 120, 113, 90, 105, 113, 102, 58, 66, 66, 73, 55, 52, 59, 121, 120, 84, 99, 104, 102, 58, 75, 85, 77, 55, 90, 59, 122, 120, 110, 80, 115, 101, 58, 100, 117, 70, 81, 70, 99, 47, 100, 116, 105, 61, 49, 47, 109, 61, 105, 56, 111, 78, 90, 98, 4
4, 71, 80, 72, 89, 74, 100, 44, 82, 105, 73, 78, 87, 101, 44, 98, 90, 48, 109, 111, 100, 44, 117, 111, 69, 117, 48, 99, 44, 76, 66, 97, 74, 120, 98 }, .content_type = { 116, 101, 120, 116, 47, 106, 97, 118, 97, 115, 99, 114, 105, 112, 116, 59, 32, 99, 104
, 97, 114, 115, 101, 116, 61, 85, 84, 70, 45, 56 }, .status = 200, .stored_at = 1775051360, .age_at_store = 0, .cache_control = .{ .max_age = 31536000 }, .headers = { .{ .name = { ... }, .value = { ... } }, .{ .name = { ... }, .value = { ... } }, .{ .name
 = { ... }, .value = { ... } }, .{ .name = { ... }, .value = { ... } }, .{ .name = { ... }, .value = { ... } }, .{ .name = { ... }, .value = { ... } }, .{ .name = { ... }, .value = { ... } }, .{ .name = { ... }, .value = { ... } }, .{ .name = { ... }, .va
lue = { ... } }, .{ .name = { ... }, .value = { ... } }, .{ .name = { ... }, .value = { ... } }, .{ .name = { ... }, .value = { ... } }, .{ .name = { ... }, .value = { ... } }, .{ .name = { ... }, .value = { ... } }, .{ .name = { ... }, .value = { ... } }
, .{ .name = { ... }, .value = { ... } }, .{ .name = { ... }, .value = { ... } } }, .vary_headers = {  } }
```

krichprollsch:
I got another crash, but not sure it relates with this PR...

```
Segmentation fault at address 0x75f804c607c8
/home/pierre/wrk/browser/src/browser/webapi/net/Fetch.zig:238:13: 0x339a4f4 in httpErrorCallback (lightpanda)
        self._owns_response = false;
            ^
/home/pierre/wrk/browser/src/browser/HttpClient.zig:1296:36: 0x34df91b in requestFailed__anon_218145 (lightpanda)
            self.req.error_callback(self.req.ctx, err);
                                   ^
/home/pierre/wrk/browser/src/browser/HttpClient.zig:951:35: 0x34e9c5a in processOneMessage (lightpanda)
            transfer.requestFailed(error.Abort, true);
                                  ^
/home/pierre/wrk/browser/src/browser/HttpClient.zig:1000:44: 0x34eaf0c in processMessages (lightpanda)
        const done = self.processOneMessage(msg, transfer) catch |err| blk: {
                                           ^
/home/pierre/wrk/browser/src/browser/HttpClient.zig:819:33: 0x2e7c8a0 in perform (lightpanda)
    if (try self.processMessages()) {
                                ^
/home/pierre/wrk/browser/src/browser/HttpClient.zig:791:25: 0x2e7c6a4 in makeRequest (lightpanda)
    _ = try self.perform(0);
                        ^
/home/pierre/wrk/browser/src/browser/HttpClient.zig:643:36: 0x2e78e8b in process (lightpanda)
            return self.makeRequest(conn, transfer);
                                   ^
/home/pierre/wrk/browser/src/browser/HttpClient.zig:403:28: 0x2e784d8 in processRequest (lightpanda)
        return self.process(transfer);
                           ^
/home/pierre/wrk/browser/src/browser/HttpClient.zig:289:35: 0x2e7a98d in request (lightpanda)
        return self.processRequest(req);
                                  ^
/home/pierre/wrk/browser/src/browser/webapi/net/Fetch.zig:89:28: 0x33992dc in init (lightpanda)
    try http_client.request(.{
                           ^
/home/pierre/wrk/browser/src/browser/webapi/Window.zig:249:22: 0x339a612 in fetch (lightpanda)
    return Fetch.init(input, options, page);
                     ^
/home/pierre/wrk/browser/src/browser/js/Caller.zig:589:9: 0x339ab8f in _call__anon_174283 (lightpanda)
        const res = @call(.auto, func, args);
        ^
/home/pierre/wrk/browser/src/browser/js/Caller.zig:563:31: 0x2de47f3 in call__anon_96972 (lightpanda)
        const js_value = _call(T, &caller.local, info, func, opts) catch |err| {
                              ^
/home/pierre/wrk/browser/src/browser/js/bridge.zig:144:41: 0x2de4584 in wrap (lightpanda)
                    Caller.Function.call(T, handle.?, func, opts);
                                        ^
???:?:?: 0x76620c6 in ??? (???)
Unwind error at address `:0x76620c6` (error.MissingFDE), trace may be incomplete
```

krichprollsch:
Another crash:
```
Segmentation fault at address 0x7ba150bd6436
/usr/local/zig-0.15.2/lib/compiler_rt/memcpy.zig:170:17: 0x8ae44aa in memcpyFast (compiler_rt)
        d[i] = s[i];
                ^
/usr/local/zig-0.15.2/lib/std/Io/Writer.zig:521:34: 0x2ad9b8d in write (lightpanda)
        @memcpy(w.buffer[w.end..][0..bytes.len], bytes);
                                 ^
/usr/local/zig-0.15.2/lib/std/Io/Writer.zig:532:51: 0x2ad7131 in writeAll (lightpanda)
    while (index < bytes.len) index += try w.write(bytes[index..]);
                                                  ^
/home/pierre/wrk/browser/src/log.zig:272:31: 0x2add820 in writeString__anon_7026 (lightpanda)
        return writer.writeAll(value);
                              ^
/home/pierre/wrk/browser/src/log.zig:251:41: 0x2e2e840 in writeValue__anon_103660 (lightpanda)
                u8 => return writeString(format, value, writer),
                                        ^
/home/pierre/wrk/browser/src/log.zig:189:23: 0x35c2253 in logPretty__anon_255543 (lightpanda)
        try writeValue(.pretty, @field(data, f.name), writer);
                      ^
/home/pierre/wrk/browser/src/log.zig:143:33: 0x35c24ab in logTo__anon_255440 (lightpanda)
        .pretty => try logPretty(scope, level, msg, data, out),
                                ^
/home/pierre/wrk/browser/src/log.zig:124:10: 0x34e3b14 in log__anon_218445 (lightpanda)
    logTo(scope, level, msg, data, &writer.interface) catch |log_err| {
         ^
/home/pierre/wrk/browser/src/log.zig:93:8: 0x2e7f75d in debug__anon_109630 (lightpanda)
    log(scope, .debug, msg, data);
       ^
/home/pierre/wrk/browser/src/browser/ScriptManager.zig:300:22: 0x2e7e0ee in addFromElement__anon_108731 (lightpanda)
            log.debug(.http, "script queue", .{
                     ^
/home/pierre/wrk/browser/src/browser/Page.zig:1085:40: 0x2e801f1 in scriptAddedCallback__anon_108712 (lightpanda)
    self._script_manager.addFromElement(from_parser, script, "parsing") catch |err| {
                                       ^
/home/pierre/wrk/browser/src/browser/Page.zig:3096:33: 0x2e9eb59 in nodeIsReady__anon_108699 (lightpanda)
        self.scriptAddedCallback(from_parser, script) catch |err| {
                                ^
/home/pierre/wrk/browser/src/browser/Page.zig:2748:33: 0x2ea100c in _insertNodeRelative__anon_108625 (lightpanda)
            try self.nodeIsReady(false, child);
                                ^
/home/pierre/wrk/browser/src/browser/Page.zig:2620:36: 0x2ea28de in appendNode (lightpanda)
    return self._insertNodeRelative(false, parent, child, .append, opts);
    ```

**Comment by krichprollsch** [comment]
I think it would be helpful to have a bit more about cache miss:
```
$ node integration/xange.js
```
```
DEBUG cache : miss . . . . . . . . . . . . . . . . . . . . .  [+2307ms]
      url = https://www.xange.vc
      hash = 8cdca7d6871192ff492446af82d136338310fc26af094b2ada4520b5d6056870
```
```
DEBUG cache : miss . . . . . . . . . . . . . . . . . . . . .  [+3975ms]
      url = https://cdn.matomo.cloud/xange.matomo.cloud/matomo.js
      hash = 3dc4ff37d7fb29636b1272797565beaf1f3b9dd8f7787854d6f75d9345a4547f
```

:point_up: I have a warmup cache, so I guess these files are either: not-cachable or invalide already?
If not cachable, it would be great to have a log and a reason. Same about the reason we miss the cache: missing or invalid?

File: src/browser/ScriptManager.zig [hunk 42]
```diff
@@ -694,82 +701,86 @@ pub const Script = struct {
         self.manager.page.releaseArena(self.arena);
     }
 
-    fn startCallback(transfer: *HttpClient.Transfer) !void {
-        log.debug(.http, "script fetch start", .{ .req = transfer });
+    fn startCallback(response: HttpClient.Response) !void {
+        log.debug(.http, "script fetch start", .{ .req = response });
```
karlseguin, line 705: Transfer implements `format` so it can be used meaningfully in logs. `Response` doesn't. This is done in a few different places.

src/network/cache/FsCache.zig
```diff
@@ -0,0 +1,580 @@
...
+pub fn put(self: *FsCache, meta: CachedMetadata, body: []const u8) !void {
+    const hashed_key = hashKey(meta.url);
+    const cache_p = cachePath(&hashed_key);
+    const cache_tmp_p = cacheTmpPath(&hashed_key);
+
+    const lock = self.getLockPtr(&hashed_key);
+    lock.lock();
+    defer lock.unlock();
+
+    const file = try self.dir.createFile(&cache_tmp_p, .{ .truncate = true });
```
karlseguin: Are we concerned about too many files in the same directory?

krichprollsch: good point, I guess we could reach the limit. But an improvements could be done in a followup PR I think.

src/network/cache/Cache.zig
```diff
@@ -0,0 +1,176 @@
...
+    cache_control: ?[]const u8,
+    vary: ?[]const u8,
+    age: ?[]const u8,
+    has_set_cookie: bool,
+    has_authorization: bool,
+) !?CachedMetadata {
+    if (status != 200) return null;
+    if (has_set_cookie) return null;
+    if (has_authorization) return null;
+    if (vary) |v| if (std.mem.eql(u8, v, "*")) return null;
```
krichprollsch: You partially implemented `vary`. You don't parse headers here to set `vary_headers` correctly.

src/network/cache/Cache.zig
```diff
@@ -0,0 +1,176 @@
...
+    status: u16,
+    content_type: ?[]const u8,
+    cache_control: ?[]const u8,
+    vary: ?[]const u8,
+    age: ?[]const u8,
+    has_set_cookie: bool,
+    has_authorization: bool,
+) !?CachedMetadata {
+    if (status != 200) return null;
+    if (has_set_cookie) return null;
```
krichprollsch: It appears we should handle that w/ vary and cache-control. I think we should remove this check.

src/network/cache/Cache.zig
```diff
@@ -0,0 +1,176 @@
...
+                    cc.max_age = max_age;
+                    max_age_set = true;
+                    max_s_age_set = true;
+                }
+            }
+        }
+
+        if (!max_age_set) return null;
+        if (!is_public) return null;
+        if (cc.max_age == 0) return null;
```
krichprollsch, line 87: So we cache only if CC has bot public and a max-age? 
I think `Cache-Control: max-age=604800` should be cached, even is public is missing.
and `Cache-Control: public` should be cached too, but I think we should set a max-age ourselves in this case.

src/network/cache/Cache.zig
```diff
@@ -0,0 +1,176 @@
...
+        var max_age_set = false;
+        var max_s_age_set = false;
+        var is_public = false;
+
+        var iter = std.mem.splitScalar(u8, value, ',');
+        while (iter.next()) |part| {
+            const directive = std.mem.trim(u8, part, &std.ascii.whitespace);
+            if (std.ascii.eqlIgnoreCase(directive, "no-store")) {
+                return null;
+            } else if (std.ascii.eqlIgnoreCase(directive, "no-cache")) {
```
krichprollsch, line 65: We can cache but we need re-validation. To implement in a follow up

> The no-cache response directive indicates that the response can be stored in caches, but the response must be validated with the origin server before each reuse, even when the cache is disconnected from the origin server.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cache-Control

krichprollsch: We should handle `must-revalidate` in the follow up to.

src/network/cache/Cache.zig
```diff
@@ -0,0 +1,176 @@
...
+    vary: ?[]const u8,
+    age: ?[]const u8,
+    has_set_cookie: bool,
+    has_authorization: bool,
+) !?CachedMetadata {
+    if (status != 200) return null;
+    if (has_set_cookie) return null;
+    if (has_authorization) return null;
+    if (vary) |v| if (std.mem.eql(u8, v, "*")) return null;
+    const cc = CacheControl.parse(cache_control orelse return null) orelse return null;
```
krichprollsch: if no `cache_control` is set, we should still cache the response according to the heuristic caching

> HTTP is designed to cache as much as possible, so even if no Cache-Control is given, responses will get stored and reused if certain conditions are met. This is called heuristic caching.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Caching#heuristic_caching

krichprollsch:
We hit the memory limit regression :/

`Peak resident set size: 27216`
in main https://github.com/lightpanda-io/browser/actions/runs/23876527633/job/69621059714

`Peak resident set size: 28864`
here

src/browser/HttpClient.zig
```diff
@@ -311,7 +314,81 @@ pub fn request(self: *Client, req: Request) !void {
...
+
+            var iter = req.headers.iterator();
+            const req_header_list = try iter.collect(arena);
+
+            if (cache.get(arena, .{
+                .url = req.url,
+                .timestamp = std.time.timestamp(),
+                .request_headers = req_header_list.items,
+            })) |cached| {
+                log.debug(.browser, "http.cache.get", .{
```
krichprollsch: why you use `.browser` instead of `.cache` here? what about using the `hit` (as you use `miss`)  to indicate you get the cache?

mookums: These logs were before I introduced the `.cache` scope. Will update.

src/network/cache/Cache.zig
mookums: We build the vary_headers in the `HttpClient`, the same way we build the headers and add it to the `CacheMetadata`. This is just to ensure that our Vary isn't wildcard (which means not to cache).

src/network/cache/Cache.zig
mookums: The status check or the has_set_cookie check (or both)? We really should never cache anything that has set cookie. This cache right now isn't private so you could get the same cookie set on two different sessions if we serve from the cache.

src/network/cache/Cache.zig
mookums: The cache doesn't support revalidation right now so if we heuristically cache something (and our guess is wrong), we will serve a stale file from the cache. I think its better to take the network hit and ensure we serve the proper resource.

PR #1940, opened 2026-03-21
mcp: initialize server in mcpThread to avoid V8 isolate crashes

### LINKED ISSUE #1938: Native MCP segmentation fault when calling the goto tool on macOS (x86_64)
## Introduction

Hi again! First, thank you so much for the super quick fix on the JSON-RPC formatting issue (#1928).

However, while testing the MCP integration with the Cline extension in VS Code, I ran into a `segmentation fault` specifically when the agent tries to use the `goto` tool.

## Steps to Reproduce

To isolate the issue, I ran the macOS `x86_64` nightly binary directly in the terminal and fed it the JSON-RPC commands manually via `stdin`. 

Here is the step-by-step reproduction:

```sh
# Starting the native MCP server
./lightpanda mcp

# Test 1: tools/list method (Works perfectly)
{"method":"tools/list","jsonrpc":"2.0","id":1}

# Response 1
{"jsonrpc":"2.0","id":1,"result":{"tools":[{"name":"goto","description":"Navigate to a specified URL..."}]}}

# Test 2: calling the goto tool (Crashes)
{"method":"tools/call","params":{"name":"goto","arguments":{"url":"https://www.google.com/"}},"jsonrpc":"2.0","id":5}

# Response 2 (Error)
zsh: segmentation fault  ./lightpanda mcp
```

## Important Context (Core engine works fine):

I wanted to check if the crash was related to the core engine rendering the page or just the MCP wrapper. If I run the standard CLI `fetch` command on the exact same URL, it works perfectly without crashing:

```sh
./lightpanda fetch --obey_robots --log_format pretty  --log_level info https://www.google.com/

INFO  telemetry : telemetry status . . . . . . . . . . . . .  [+0ms]
      disabled = false

INFO  page : navigate . . . . . . . . . . . . . . . . . . . . [+5ms]
      url = https://www.google.com/
      method = GET
      reason = address_bar
      body = false
      req_id = 1
      type = root

...

INFO  http : request complete . . . . . . . . . . . . . . . . [+2837ms]
      source = xhr
      url = https://play.google.com/log?format=json&hasfast=true
      status = 200
      len = 131
```

Since the `fetch` command succeeds but the `goto` tool over MCP segfaults, it seems the crash might be related to how the native MCP server handles the page state in memory or processes the response specifically on the `x86_64` macOS build.

Let me know if you need me to test any specific patches or provide crash logs! Thanks again for the amazing work on this project.

arrufat:
When running mcp server, it initialized lp.mcp.Server in the main thread which also implicitly created the V8 isolate in the main thread. When processing requests (like calling the goto tool) inside mcpThread, V8 would assert that the isolate doesn't match the current thread.

Fixes #1938

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -168,7 +168,7 @@ fn fetchThread(app: *App, url: [:0]const u8, fetch_opts: lp.FetchOpts) void {
 
 fn mcpThread(allocator: std.mem.Allocator, app: *App) void {
     defer app.network.stop();
-    
+
     var stdout = std.fs.File.stdout().writer(&.{});
     var mcp_server: *lp.mcp.Server = lp.mcp.Server.init(allocator, app, &stdout.interface) catch |err| {
         log.fatal(.mcp, "mcp init error", .{ .err = err });


karlseguin:
I know we don't do this for any other command, but maybe it's worth adding a CI task that runs

```
echo '{"method": "tools/call", "params": {"name": "links", "arguments": {"url": "http://127.0.0.1:1234"}}, "id": 0}' | zig build run -- mcp --log_level warn
```

and checking the output? cc @krichprollsch

PR #1941, opened 2026-03-21
Add waitForSelector MCP tool

mvanhorn:
## Summary

Adds a `waitForSelector` MCP tool that polls for a CSS selector match with a configurable timeout. Returns the `backendNodeId` of the matched element for use with `click` and `fill` tools.

## Why this matters

The MCP server has interaction tools (`click`, `fill`, `scroll`) but no way to wait for dynamic content before acting on it. AI agents navigating SPAs or JS-heavy pages need to wait for elements to render before interacting. This is a standard browser automation primitive -- Playwright and Puppeteer both have `waitForSelector`.

@krichprollsch noted in [PR #1927](https://github.com/lightpanda-io/browser/pull/1927#issuecomment-4099130030): "I also think we will have soon an option to wait until a given css selector returns something."

## Changes

Single file change to `src/mcp/tools.zig`:
- Tool definition with `selector` (required) and `timeout` (optional, default 5000ms) parameters
- Enum/map entries following existing pattern
- Handler that polls `Selector.querySelector` while running the session event loop via `session.wait` between checks (100ms intervals)
- Returns the `backendNodeId` of the matched element so it can be passed directly to `click` or `fill`
- Returns an error on invalid selectors or timeout

## Testing

- Builds with `zig build`
- Follows the exact pattern of existing tools (`handleClick`, `handleFill`, `handleScroll`)
- Uses `Selector.querySelector` (same as `interactiveElements` and `links` tools), `server.node_registry.register` (same as CDP `lp` domain), and `session.wait` (same as `performGoto`)

### DIFF
--- a/src/browser/tests/mcp_wait_for_selector.html
+++ b/src/browser/tests/mcp_wait_for_selector.html
@@ -1,0 +1,14 @@
+<!DOCTYPE html>
+<html>
+<body>
+    <div id="existing">Already here</div>
+    <script>
+        setTimeout(function() {
+            var el = document.createElement("div");
+            el.id = "delayed";
+            el.textContent = "Appeared after delay";
+            document.body.appendChild(el);
+        }, 200);
+    </script>
+</body>
+</html>
--- a/src/mcp/tools.zig
+++ b/src/mcp/tools.zig
@@ -723,3 +723,104 @@ test "MCP - Actions: click, fill, scroll" {
 
     try testing.expect(result.isTrue());
 }
+
+test "MCP - waitForSelector: existing element" {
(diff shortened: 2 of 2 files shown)


mvanhorn:
Added tests in e9bed18c. Three test cases covering the key behaviors:
1. Immediate match on an existing element (`#existing`)
2. Polling match on a delayed element (`#delayed` - added after 200ms via `setTimeout`)
3. Timeout error on a non-existent element with a short timeout

Also added `mcp_wait_for_selector.html` test fixture following the same pattern as `mcp_actions.html`.

PR #1946, opened 2026-03-21
Encode non-utf8 Network.getResponseBody in base64 

krichprollsch:
Follow up for #1921

* add a struct for captured response to handle encoded and raw response
* Avoid an additional allocation by encoding directly the stream into the captured response 
* keep raw responses depending the resp content-type.

### DIFF
--- a/src/cdp/cdp.zig
+++ b/src/cdp/cdp.zig
@@ -326,7 +326,7 @@ pub fn BrowserContext(comptime CDP_T: type) type {
     const AXNode = @import("AXNode.zig");
 
     const CapturedResponse = struct {
-        encode: enum { none, base64 },
+        must_encode: bool,
         data: std.ArrayList(u8),
     };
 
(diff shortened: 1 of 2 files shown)


src/cdp/cdp.zig
```diff
@@ -650,11 +681,16 @@ pub fn BrowserContext(comptime CDP_T: type) type {
...
+
+            if (resp.encode == .none) {
+                return resp.data.appendSlice(arena, msg.data);
             }
-            try gop.value_ptr.appendSlice(arena, try arena.dupe(u8, msg.data));
+
+            const encoded_len = std.base64.standard.Encoder.calcSize(msg.data.len);
+            const start = resp.data.items.len;
+            try resp.data.resize(arena, start + encoded_len);
+            _ = std.base64.standard.Encoder.encode(resp.data.items[start..], msg.data);
```
karlseguin: I don't think you can glue base64 fragments together. Every chunk will [potentially] get padding, which will result in a whole that isn't valid.

Might need to collect everything in `data` with a "processed: bool", and then ,the first time the capture is requested, if `processed == false`, encode it?

karlseguin: This can be done. But you process the chunk up to an amount that's divisible by 3, store the leftover (0, 1 or 2 bytes), prepend that to the next chunk, and repeat. Tedious...and still requires a lazy-last step when the data is requested OR a new http_done event.

krichprollsch: So you think it's safer to encode when returning the cdp response instead of trying to stream encoding here?

karlseguin: Certainly easier.

krichprollsch: updated

PR #1958, opened 2026-03-23
Extract Session.wait into a Runner

karlseguin:
This is done for a couple reasons. The first is just to have things a little more self-contained for eventually supporting more advanced "wait" logic, e.g. waiting for a selector.

The other is to provide callers with more fine-grained controlled. Specifically the ability to manually "tick", so that they can [presumably] do something after every tick. This is needed by the test runner to support more advanced cases (cases that need to test beyond 'load') and it also improves (and fixes potential use-after-free, the lp.waitForSelector)

### DIFF
--- a/src/browser/tests/mcp_wait_for_selector.html
+++ b/src/browser/tests/mcp_wait_for_selector.html
@@ -8,7 +8,7 @@
             el.id = "delayed";
             el.textContent = "Appeared after delay";
             document.body.appendChild(el);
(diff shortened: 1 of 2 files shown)


karlseguin:
@arrufat tweaks the new `waitForSelector`. It wasn't safe to hold a reference to a page over a call to `session.wait`, something could cause the page to navigate, which would deinit the page and create a new one.

PR #1972, opened 2026-03-23
Fix Expo Web crash by gracefully handling at-rules in CSSStyleSheet.insertRule

### LINKED ISSUE #1970: Expo Web (React Native for Web) SPA fails to render: CSSStyleSheet.insertRule throws 'invalid argument'
## Summary

Expo Web apps fail to render when accessed via Lightpanda (both \`fetch\` command and CDP/Playwright mode). The root \`<div id="root">\` remains empty and the app never hydrates.

## Reproduction

Any app built with **Expo Web / React Native for Web + react-native-reanimated**.

To reproduce locally:
1. Create a new Expo Router project: \`npx create-expo-app my-app\`
2. Add \`react-native-reanimated\` (already included by default in many templates)
3. Build for web: \`npx expo export --platform web\`
4. Serve the dist folder and fetch with Lightpanda

## Environment

- Lightpanda: nightly (aarch64-macos, 2026-03-23)
- Mode tested: \`lightpanda fetch\` and \`lightpanda serve\` + Playwright CDP

## Root cause identified

\`CSSStyleSheet.insertRule()\` throws \`TypeError: invalid argument\` inside Expo's Reanimated animation stylesheet initialization:

```
TypeError: invalid argument
    at n.onload (expo-bundle.js:1271:310)
```

The code at that location:
```javascript
document.createElement('style');
n.id = o;
n.onload = () => {
  if (n.sheet)
    for (const t in i.Animations)
      n.sheet.insertRule(i.Animations[t].style);  // ← throws here
  else
    t.logger.error('Failed to create layout animations stylesheet...')
}
```

Expo's \`react-native-reanimated\` creates a \`<style>\` element and calls \`sheet.insertRule()\` with CSS animation keyframe strings. Lightpanda throws \`invalid argument\` on these rules.

## Workaround attempted

Wrapping \`CSSStyleSheet.prototype.insertRule\` to swallow errors prevents the initial crash:

```javascript
const orig = CSSStyleSheet.prototype.insertRule;
CSSStyleSheet.prototype.insertRule = function(rule, index) {
  try { return orig.call(this, rule, index); } catch(e) { return 0; }
};
```

With this workaround, \`page.evaluate()\` starts working (no more "Execution context was destroyed"), React Navigation globals initialize, but the root div remains empty — suggesting additional CDP stability issues remain.

## Steps to reproduce

```bash
# Build a minimal Expo Web app and serve it, then:
lightpanda fetch --dump markdown http://localhost:PORT

# Expected: page content
# Actual: JsException error, empty output
```

## Expected behavior

Expo Web apps should render their content, similar to how Chrome/Firefox headless handles them.

## Additional context

- This appears to affect all Expo Web / React Native for Web apps
- The \`insertRule\` issue is triggered by \`react-native-reanimated\`'s animation stylesheet initialization
- Other screen API polyfills (\`window.screen\`, \`window.innerWidth\` etc.) are also needed but secondary to this issue

arrufat:
Closes #1970

### DIFF
--- a/src/browser/webapi/css/CSSStyleSheet.zig
+++ b/src/browser/webapi/css/CSSStyleSheet.zig
@@ -88,6 +88,8 @@ pub fn insertRule(self: *CSSStyleSheet, rule: []const u8, maybe_index: ?u32, pag
         return error.SyntaxError;
     };
 
+    if (it.next() != null) return error.SyntaxError;
+
     const style_rule = try CSSStyleRule.init(page);
     try style_rule.setSelectorText(parsed_rule.selector, page);
 
--- a/src/browser/tests/css/stylesheet.html
+++ b/src/browser/tests/css/stylesheet.html
@@ -480,6 +480,24 @@
 }
 </script>
 
+<script id="CSSStyleSheet_insertRule_multiple_rules">
+{
+    const style = document.createElement('style');
+    document.head.appendChild(style);
+    const sheet = style.sheet;
+
+    let caught = false;
+    try {
+        sheet.insertRule('a { color: red; } b { color: blue; }');
+    } catch (e) {
+        caught = true;
+        testing.expectEqual('SyntaxError', e.name);
+    }
+    testing.expectTrue(caught);
+    testing.expectEqual(0, sheet.cssRules.length);
+}
+</script>
+
 <script id="CSSStyleSheet_replaceSync">
 {
     const sheet = new CSSStyleSheet();


karlseguin:
Should this be in the RulesIterator of the `parseStylesheet` so that it skips for all users?

arrufat:
> Should this be in the RulesIterator of the `parseStylesheet` so that it skips for all users?

Yes, good idea! :)

src/browser/webapi/css/CSSStyleSheet.zig
```diff
@@ -75,7 +75,16 @@ pub fn getOwnerRule(self: *const CSSStyleSheet) ?*CSSRule {
 pub fn insertRule(self: *CSSStyleSheet, rule: []const u8, maybe_index: ?u32, page: *Page) !u32 {
     const index = maybe_index orelse 0;
     var it = Parser.parseStylesheet(rule);
-    const parsed_rule = it.next() orelse return error.SyntaxError;
+    const parsed_rule = it.next() orelse {
+        if (it.has_skipped_at_rule) {
+            // Lightpanda currently skips at-rules (e.g., @keyframes, @media) in its
+            // CSS parser. To prevent JS apps (like Expo/Reanimated) from crashing
+            // during initialization, we simulate a successful insertion by returning
+            // the requested index.
```
krichprollsch, line 85: Shouldn't we add some not_implemented debug log in this case?

PR #1976, opened 2026-03-23
Add `lightpanda` and `default` package to flake.nix

lukasl-dev:
This PR adds the `lightpanda` and `default` packages to flake.nix, such that lightpanda can be build via.

```bash
nix build . # or `nix build github:lightpanda-io/browser`
```

This allows Nixians to include lightpanda directly into their system config as a flake input:

```nix
# flake.nix
inputs.lightpanda.url = "github:lightpanda-io/browser";

# lightpanda.nix
{ inputs, pkgs, ... }:
{
  environment.systemPackages = [
    inputs.lightpanda.packages.${pkgs.stdenv.hostPlatform.system}.default
  ];
}
```

It can also be run via `nix run`:

```bash
nix run github:lightpanda-io/browser # or `nix run .`
```

The package build reuses the existing FHS environment already defined in flake.nix, and adds the necessary Zig and Cargo dependency handling so the build works in the Nix sandbox.


lukasl-dev:

mookums:
Hi! Tried this today and it fails for me.

```bash
error: Cannot build '/nix/store/hvma90wfwyxf8wqkybb079l8nn6aiv50-source.drv'.
       Reason: builder failed with exit code 1.
       Output paths:
         /nix/store/xy9bhpxljg97p4qhhhs65gmrws16sy2y-source
       Last 7 log lines:
       >
       > trying https://codeload.github.com/lightpanda-io/zig-v8-fork/tar.gz/refs/tags/v0.3.4
       >   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
       >                                  Dload  Upload   Total   Spent    Left  Speed
       > 100 57993    0 57993    0     0   159k      0 --:--:-- --:--:-- --:--:--  159k
       > unpacking source archive /build/v0.3.4
       > do not know how to unpack source archive /build/v0.3.4
       For full logs, run:
         nix log /nix/store/hvma90wfwyxf8wqkybb079l8nn6aiv50-source.drv
error: Cannot build '/nix/store/jm19kbiqd865a789ynnr2j3cf6yiigyn-zig-packages.drv'.
       Reason: 1 dependency failed.
       Output paths:
         /nix/store/9xv22gf5cql3hy9mka23hv6x1h4az1dp-zig-packages
```

lukasl-dev:
That's interesting as I'm already using it in [my own config](https://github.com/lukasl-dev/nixos/blob/56bad16c52dab732abf51832b14f96b5ca10f376/flake.nix#L49). 

It seems like Nix couldn't figure out how to extract the archive for codeload urls. I think I found a fix, however, could you share your `nix --version` and your system (x86_64-linux, aarch64-darwin, etc.)? Maybe it's a Nix version thing (which would be funny).

lukasl-dev:
My VPS and my Raspberry PI had a similar issue. Not fully identical, but similar:

```
error: Cannot build '/nix/store/6k0j6p8s28pf2vjj6rqd84iyd1g2y9w3-source.drv'.
       Reason: builder failed with exit code 1.
       Output paths:
         /nix/store/d6k8my29bvn3q1gsghz4kb01wf0hxsad-source
       Last 8 log lines:
       >
       > trying https://chromium.googlesource.com/chromium/tools/depot_tools.git/+archive/4ce8ba39a3488397a2d1494f167020f21de502f3.tar.gz
       >   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
       >                                  Dload  Upload   Total   Spent    Left  Speed
       > 100 1251k    0 1251k    0     0  2199k      0 --:--:-- --:--:-- --:--:-- 2203k
       > unpacking source archive /build/4ce8ba39a3488397a2d1494f167020f21de502f3.tar.gz
       > error: zip file must contain a single file or directory.
       > hint: Pass stripRoot=false; to fetchzip to assume flat list of files.
       For full logs, run:
         nix log /nix/store/6k0j6p8s28pf2vjj6rqd84iyd1g2y9w3-source.drv
```

Apparently, zon2nix struggles with redirected urls and with archives that have no root directory. I created a `zon2nix.sh` to generate the required Nix file using [zon2nix](https://github.com/nix-community/zon2nix) and patch it afterwards (canonical urls, root stripping). The script is available inside the devShell via `zon2nix` (without any arguments) and via `nix run .#zon2nix`. 

And I renamed `deps.nix` to `build.zig.zon.nix` to make intent more clear, given there are cargo dependencies too.

I've tested it on all three of my Nix devices and two of them build successfuly now.

There's one issue left w.r.t. FHS for packaging.

lukasl-dev:
FHS-related issues are now also fixed. 3/3 Nix devices build successfully now.

mookums:
Hello,

This does build for me now and it seems to work fine so I appreciate all of the fixes. I'm not entirely sure of the maintenance burden here considering most of the team doesn't use Nix and this seems to require regeneration of the `build.zig.zon.nix` whenever dependencies are bumped. I think a solution like this might be viable later down the line, whenever we hit a stable release cadence and generation of this files can be a part of preparation for a release.

lukasl-dev:
Hi! I understand your concern, but we as Nix users are pretty much used to pulling foreign Nix flakes that break every now and then (opencode, fff, etc.). That's why I included the `zon2nix.sh` script so that other Nixians can open PRs as soon as something breaks or dependencies change unexpectedly. Ideally, unless something changes drastically, a simple `./zon2nix.sh` should suffice to update the `build.zig.zon.nix` file. So I don't believe the burden of maintaining the flake package is necessarily on your side, and Nix users are always* happy to see a project with a flake package :>

flake.nix
```diff
@@ -17,6 +17,8 @@
     };
 
     flake-utils.url = "github:numtide/flake-utils";
+
+    zon2nix.url = "github:nix-community/zon2nix";
```
mookums, line 21:
```suggestion
    zon2nix.url = "github:nix-community/zon2nix";
    zon2nix.inputs.nixpkgs.follows = "nixpkgs";
```

flake.nix
```diff
@@ -44,6 +52,21 @@
...
+          runtimeInputs =
+            with pkgs;
+            [
+              gawk
+              ripgrep
+            ]
+            ++ [ zon2nix.packages.${system}.default ];
+          text = # bash
+            ''
+              exec bash "$PWD/zon2nix.sh"
```
mookums, line 66: Tried this today and i don't think you even need this wrapper script.

You can just do `zon2nix > build.zig.zon.nix` and it seems to generate a valid file that still builds.

flake.nix
```diff
@@ -55,30 +78,143 @@
...
+            }
+          else
+            null;
+
+        prebuiltV8Arg = pkgs.lib.optionalString (prebuiltV8 != null) "-Dprebuilt_v8_path=${prebuiltV8}";
       in
       {
+        packages = rec {
+          default = lightpanda;
+          lightpanda = pkgs.stdenv.mkDerivation {
```
mookums, line 138: Would prefer if this was up in the let and then you just had something like:

```nix
...
lightpanda = pkgs.stdenv.mkDerivation {
...
}
in
{
packages =  {
     default = lightpanda;
      lightpanda = lightpanda;
    }
}
```

just makes the final part less noisy.

zon2nix.sh
```diff
@@ -0,0 +1,132 @@
+#!/usr/bin/env bash
```
mookums, line 1: We don't need this file at all. See above comment about `zon2nix`

mookums (changes requested):
Fair enough with regards to maintenance. I'm willing to get this pushed through if you can just make some of these changes + make sure the flake and `build.zig.zon.nix` is up to date.

flake.nix
mookums: Something like this seems to work fine:

```nix
zon2nixScript = pkgs.writeShellApplication {
  name = "zon2nix";
  runtimeInputs = [ zon2nix.packages.${system}.default ];
  text = ''
    zon2nix > build.zig.zon.nix
  '';
};
```

mookums:
This PR also does not build for me if we do not provide a prebuilt v8.

mookums:
Going to say this is not planned for now. I'm willing to revisit adding Nix build support to Lightpanda at a later date (especially in the future when we have stable builds).

PR #1990, opened 2026-03-25
Remove cdp generic

karlseguin:
CDPT used to be a generic so that we could inject Browser, Session, Page and
Client. At some point, it [thankfully] became a generic only to inject Client.

This commit removes the generic and bakes the *Server.Client instance in CDP.
It uses a socketpair for testing.

BrowserContext is still generic, but that's generic for a very different reason
and, while I'd like to remove that generic too, it belongs in a different PR.

### DIFF
--- a/src/lightpanda.zig
+++ b/src/lightpanda.zig
@@ -192,9 +192,9 @@ fn dumpWPT(page: *Page, writer: *std.Io.Writer) !void {
 
 pub inline fn assert(ok: bool, comptime ctx: []const u8, args: anytype) void {
     if (!ok) {
-        // if (comptime IS_DEBUG) {
-        //     unreachable;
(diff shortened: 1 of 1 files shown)


src/lightpanda.zig
```diff
@@ -193,9 +192,9 @@ fn dumpWPT(page: *Page, writer: *std.Io.Writer) !void {
 
 pub inline fn assert(ok: bool, comptime ctx: []const u8, args: anytype) void {
     if (!ok) {
-        if (comptime IS_DEBUG) {
-            unreachable;
-        }
+        // if (comptime IS_DEBUG) {
+        //     unreachable;
+        // }
```
krichprollsch: Do you keep that change?

PR #1992, opened 2026-03-25
CDP: implement Page.reload

navidemad:
## Summary

- Add `Page.reload` to the CDP Page domain dispatch enum
- Reuse the existing `page.navigate()` path with `NavigationKind.reload`, matching what `Location.reload` already does for the JS `location.reload()` API
- Accept standard CDP params (`ignoreCache`, `scriptToEvaluateOnLoad`) per the [Chrome DevTools Protocol spec](https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-reload)
- Copy the current page URL to the stack before `replacePage()` to avoid a use-after-free when the old page's arena is freed

## Motivation

CDP clients (Puppeteer, capybara-lightpanda, Selenium, etc.) that call `Page.reload` currently get `UnknownMethod`. This forces workarounds like navigating to `location.href` (which creates a new history entry instead of a proper reload, and doesn't send `Cache-Control` headers).

The implementation is minimal (~30 lines) since all the infrastructure already exists: `NavigationKind.reload` is defined, `Location.reload` uses it via `scheduleNavigation`, and the `navigate` CDP handler provides the exact pattern to follow.

## Test plan

- [x] `Page.reload` dispatches correctly (enum + switch added)
- [ ] Existing CDP tests still pass (CI)
- [ ] Manual test: `Page.navigate` to a URL, then `Page.reload` triggers a reload navigation with the same URL

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/cdp/domains/page.zig
+++ b/src/cdp/domains/page.zig
@@ -269,12 +269,8 @@ fn doReload(cmd: anytype) !void {
     const session = bc.session;
     var page = session.currentPage() orelse return error.PageNotLoaded;
 
-    // Copy URL to stack before replacePage() frees the old page's arena.
-    var url_buf: [8192:0]u8 = undefined;
-    const len = @min(page.url.len, url_buf.len);
-    @memcpy(url_buf[0..len], page.url[0..len]);
-    url_buf[len] = 0;
-    const reload_url: [:0]const u8 = url_buf[0..len :0];
+    // Dupe URL before replacePage() frees the old page's arena.
+    const reload_url = try cmd.arena.dupeZ(u8, page.url);
 
     if (page._load_state != .waiting) {
         page = try session.replacePage();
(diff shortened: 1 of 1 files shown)


src/cdp/domains/page.zig
```diff
@@ -252,6 +254,40 @@ fn navigate(cmd: anytype) !void {
...
+
+    if (bc.session_id == null) {
+        return error.SessionIdNotLoaded;
+    }
+
+    const session = bc.session;
+    var page = session.currentPage() orelse return error.PageNotLoaded;
+
+    // Copy URL to stack before replacePage() frees the old page's arena.
+    var url_buf: [8192:0]u8 = undefined;
```
karlseguin:
```zig
const reload_url = try cmd.arena.dupeZ(u8, page.url);
```

src/cdp/domains/page.zig
```diff
@@ -784,3 +820,27 @@ test "cdp.page: getLayoutMetrics" {
         },
     }, .{ .id = 12 });
 }
+
+test "cdp.page: reload" {
+    var ctx = testing.context();
```
karlseguin: If you rebase, you'll need a `try` here, `try testing.context();`

karlseguin:
Thank you

PR #1997, opened 2026-03-26
build: bump zig-v8 to v0.3.7

arrufat:


arrufat:
Ah, we need to wait until the 0.3.7 artifacts are created... I will re-run then

PR #2006, opened 2026-03-26
URL: bound '@' search to authority section

krichprollsch:
The '@' delimiter for userinfo was searched across the full URL string, including path, query, and fragment. A URL like `http://attacker.com/@bank.com/` would incorrectly parse `bank.com` as the host, enabling a Same-Origin Policy bypass.

Restrict the '@' search to the authority section (before any `/?#`) in getOrigin, getUserInfo, and getHost.


krichprollsch:
closed in favor of #1998

PR #2021, opened 2026-03-28
fix: navigator.languages should include base language per spec

evan108108:
## Summary

`navigator.languages` currently returns `["en-US"]` (1 element). Per the [HTML spec](https://html.spec.whatwg.org/multipage/system-state.html#dom-navigator-languages) and real browser behavior (Chrome, Firefox, Safari), the preferred languages list should include the base language tag alongside the regional variant — i.e. `["en-US", "en"]`.

This improves compatibility with sites that perform language negotiation and expect the base language to be present.

## Changes

- `Navigator.getLanguages()` — return `["en-US", "en"]` instead of `["en-US"]`
- Updated navigator test to expect 2 languages

## Test plan

- [x] `zig fmt --check` passes
- [x] Navigator test updated to match new behavior
- [ ] `zig build test` (CI will verify)

### DIFF
--- a/src/browser/tests/navigator/navigator.html
+++ b/src/browser/tests/navigator/navigator.html
@@ -15,8 +15,9 @@
   testing.expectEqual(true, validPlatforms.includes(navigator.platform));
   testing.expectEqual('en-US', navigator.language);
   testing.expectEqual(true, Array.isArray(navigator.languages));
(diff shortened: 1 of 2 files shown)


evan108108:

evan108108:
recheck

evan108108:
recheck

karlseguin:
Thanks :)

PR #2022, opened 2026-03-28
fix: cache canvas 2D context and lock context type per spec

evan108108:
## Summary

`HTMLCanvasElement.getContext()` currently creates a new `CanvasRenderingContext2D` on every call. Per the [HTML spec](https://html.spec.whatwg.org/multipage/canvas.html#dom-canvas-getcontext):

1. **Repeated calls with the same type should return the same object** — code like `ctx === canvas.getContext("2d")` should be `true`
2. **Once a context type is established, requesting a different type should return `null`** — e.g. after `getContext("2d")`, calling `getContext("webgl")` returns `null`

## Changes

- Add `_context_type` and `_ctx_2d` cache fields to `Canvas`
- Cache the 2D context on first creation, return it on subsequent calls
- Track the first-requested context type and return `null` for incompatible types

## Test plan

- [ ] `zig build test` (CI will verify)
- [x] `zig fmt --check` passes
- Manual: `canvas.getContext("2d") === canvas.getContext("2d")` now returns `true`
- Manual: `canvas.getContext("2d"); canvas.getContext("webgl")` now returns `null` for the second call

### DIFF
--- a/src/browser/webapi/element/html/Canvas.zig
+++ b/src/browser/webapi/element/html/Canvas.zig
@@ -30,6 +30,13 @@ const OffscreenCanvas = @import("../../canvas/OffscreenCanvas.zig");
 const Canvas = @This();
 _proto: *HtmlElement,
 
+/// Cached context type. Once set, requesting a different type returns null (per spec).
+_context_type: ContextType = .none,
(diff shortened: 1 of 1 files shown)


evan108108:
I have read the CLA Document and I hereby sign the CLA.

evan108108:
recheck

PR #2024, opened 2026-03-28
Rework finalizers

karlseguin:
This commit involves a number of changes to finalizers, all aimed towards better consistency and reliability.

A big part of this has to do with v8::Inspector's ability to move objects across IsolatedWorlds. There has been a few previous efforts on this, the most significant being https://github.com/lightpanda-io/browser/pull/1901. To recap, a Zig instance can map to 0-N v8::Objects. Where N is the total number of IsolatedWorlds. Generally, IsolatedWorlds between origins are...isolated...but the v8::Inspector isn't bound by this. So a Zig instance cannot be tied to a Context/Identity/IsolatedWorld...it has to live until all references, possibly from different IsolatedWorlds, are released (or the page is reset).

Finalizers could previously be managed via reference counting or explicitly toggling the instance as weak/strong. Now, only reference counting is supported. weak/strong can essentially be seen as an acquireRef (rc += 1) and releaseRef (rc -= 1). Explicit setting did make some things easier, like not having to worry so much about double-releasing (e.g. XHR abort being called multiple times), but it was only used in a few places AND it simply doesn't work with objects shared between IsolatedWorlds. It is never a boolean now, as 3 different IsolatedWorlds can each hold a reference.

Temps and Globals are tracked on the Session. Previously, they were tracked on the Identity, but that makes no sense. If a Zig instance can outlive an Identity, then any of its Temp references can too. This hasn't been a problem because we've only seen MutationObserver and IntersectionObserver be used cross-origin, but the right CDP script can make this crash with a use-after-free (e.g. `MessageEvent.data` is released when the Identity is done, but `MessageEvent` is still referenced by a different IsolateWorld).

Rather than deinit with a `comptime shutdown: bool`, there is now an explicit `releaseRef` and `deinit`.

Bridge registration has been streamlined. Previously, types had to register their finalizer AND acquireRef/releaseRef/deinit had to be declared on the entire prototype chain, even if these methods just delegated to their proto. Finalizers are now automatically enabled if a type has a `acquireRef` function. If a type has an `acquireRef`, then it must have a `releaseRef` and a `deinit`. So if there's custom cleanup to do in `deinit`, then you also have to define `acquireRef` and `releaseRef` which will just delegate to the _proto.

Furthermore these finalizer methods can be defined anywhere on the chain.

Previously:

```zig
const KeywboardEvent = struct {
  _proto: *Event,
  ...

  pub fn deinit(self: *KeyboardEvent, session: *Session) void {
    self._proto.deinit(session);
  }

  pub fn releaseRef(self: *KeyboardEvent, session: *Session) void {
    self._proto.releaseRef(session);
  }
}
```

```zig
const KeyboardEvent = struct {
  _proto: *Event,
  ...
  // no deinit, releaseRef, acquireref
}
```

Since the `KeyboardEvent` doesn't participate in finalization directly, it doesn't have to define anything. The bridge will detect the most specific place they are defined and call them there.

### DIFF
--- a/src/browser/webapi/IntersectionObserver.zig
+++ b/src/browser/webapi/IntersectionObserver.zig
@@ -190,11 +190,11 @@ pub fn disconnect(self: *IntersectionObserver, page: *Page) void {
     const observing_count = self._observing.items.len;
     self._observing.clearRetainingCapacity();
 
+    page.unregisterIntersectionObserver(self);
+
     if (observing_count > 0) {
         _ = self.releaseRef(page._session);
     }
-
-    page.unregisterIntersectionObserver(self);
 }
 
 pub fn takeRecords(self: *IntersectionObserver, page: *Page) ![]*IntersectionObserverEntry {


src/browser/webapi/event/ErrorEvent.zig
```diff
@@ -80,11 +80,19 @@ fn initWithTrusted(arena: Allocator, typ: String, opts_: ?Options, trusted: bool
...
-    self._proto.deinit(shutdown, session);
+    self._proto.deinit(session);
+}
+
+pub fn acquireRef(self: *ErrorEvent) void {
+    self._proto.acquireRef();
+}
+
+pub fn releaseRef(self: *ErrorEvent, session: *Session) void {
+    self._proto._rc.release(self, session);
```
krichprollsch, line 95: I guess you need t re-declare proto's function here to change the `releaseRef` behavior, but why do you need you to change it? 

It looks like it's the same than the _proto's releaseRef 🤔

src/browser/webapi/IntersectionObserver.zig
```diff
@@ -163,21 +175,26 @@ pub fn unobserve(self: *IntersectionObserver, target: *Element, page: *Page) voi
...
+    self._previous_states.clearRetainingCapacity();
+
+    const observing_count = self._observing.items.len;
+    self._observing.clearRetainingCapacity();
+
+    if (observing_count > 0) {
+        _ = self.releaseRef(page._session);
+    }
+
+    page.unregisterIntersectionObserver(self);
```
krichprollsch: Claude warns about the risk of calling this after the releaseRef call:
>   2. IntersectionObserver.disconnect calls releaseRef which may call deinit and free self, then calls page.unregisterIntersectionObserver(self) (IntersectionObserver.zig:1089-1091). If releaseRef drops the last ref, self is freed and the subsequent
  unregisterIntersectionObserver(self) is a use-after-free. The _ = self.releaseRef(...) discards the return but releaseRef is void — the issue is the ordering. Same pattern in MutationObserver.disconnect at line 1229.

karlseguin: I switched the order, so that releaseRef is called last. But, right now, v8 would always be holding onto a 1 reference when this function is called, so releaseRef could never cause deinit. But, internally, someone could start to use the IO outside of v8, in which case it would be an issue. So fixed.

src/browser/webapi/event/ErrorEvent.zig
karlseguin: Because we want to pass `self: *ErrorEvent`, not `self: *Event` to do the RC(T).release, so that, if `_refs == 0`, it'll call `ErrorEvent.deinit`, not `Event.deinit`. That will be easy to overlook.

PR #2027, opened 2026-03-29
mcp: allow configuring protocol version

### LINKED ISSUE #2023: Support for MCP version 2024-11-05
LM Studio (and I believe ollama) currently only support version 2024-11-05.

Lightpanda is using MCP version 2025-11-25 which fails to start when using LM Studio.

It would be great if, for compatibility, lightpanda supported MCP 2024-11-05, perhaps with an env var

```
export LIGHTPANDA_MCP_VERSION="2024-11-05"
lightpanda mcp
```

arrufat:
Closes #2023

### DIFF
--- a/src/Config.zig
+++ b/src/Config.zig
@@ -189,7 +189,6 @@ pub fn webBotAuth(self: *const Config) ?WebBotAuthConfig {
     };
 }
 
-
 pub fn maxConnections(self: *const Config) u16 {
     return switch (self.mode) {
         .serve => |opts| opts.cdp_max_connections,


src/Config.zig
```diff
@@ -640,18 +656,41 @@ fn parseMcpArgs(
     allocator: Allocator,
     args: *std.process.ArgIterator,
 ) !Mcp {
-    var mcp: Mcp = .{};
+    var result: Mcp = .{};
+    var env_ver: ?[]const u8 = null;
+    var arg_ver: ?[]const u8 = null;
+
+    if (std.posix.getenv("LIGHTPANDA_MCP_VERSION")) |ver| {
```
karlseguin: Is there a reason to support this for this one parameter and none other?

src/Config.zig
```diff
@@ -188,6 +189,13 @@ pub fn webBotAuth(self: *const Config) ?WebBotAuthConfig {
     };
 }
 
+pub fn mcpVersion(self: *const Config) []const u8 {
```
karlseguin: I think exposing the enum is generally more useful, e.g. it can be switched on.

src/Config.zig
arrufat: Right, it's the only one. The person who opened the issue suggested it, but I will remove it.

Issue #1994, opened 2026-03-25
[BUG] Incorrect window.location.href assignment parsing: "http:/test" resolves to wrong URL compared to Chrome


cnjn:
## Summary

When assigning a value like `window.location.href = location.protocol + '/test'` on a page such as `http://localhost/index`, Lightpanda incorrectly resolves the resulting navigation URL as `http://localhost/index/http:/test`. In Chrome and other browsers, the navigation correctly resolves to `http://localhost/test`.

## Steps to Reproduce
1. Start Lightpanda and open a page at `http://localhost/index`.
2. Run the following JavaScript in the console:
   ```js
   window.location.href = location.protocol + '/test';
   ```
3. Observe that Lightpanda navigates to:
   ```
   http://localhost/index/http:/test
   ```
   while Chrome navigates to:
   ```
   http://localhost/test
   ```

## Expected Behavior
Lightpanda should resolve `window.location.href = location.protocol + '/test'` as `http://localhost/test`, just like Chrome and the URL Standard specify.

## Actual Behavior
Lightpanda resolves `http:/test` as a path relative to the current directory, resulting in `http://localhost/index/http:/test`, which is incorrect and leads to a 404 error if `/index/http:/test` does not exist.

## Additional Notes
- This breaks compatibility with code that programmatically sets `window.location.href` using `location.protocol`.
- Spec: [WHATWG URL Standard (Special Relative URLs)](https://url.spec.whatwg.org/#relative-state)
- This bug can lead to subtle navigation errors and is hard to debug if not aware of the parsing discrepancy.

## Suggested Fix
The URL parsing and resolution logic in Lightpanda should be updated to match the WHATWG URL Standard and Chrome's behavior, especially when the input begins with a scheme but only a single slash (e.g. `http:/test`).

**Please let me know if further information or reproduction cases are needed!**

Dinisimys2018:
Note https://url.spec.whatwg.org/#scheme-state, step 6.
The scheme path must match the base scheme, otherwise it will not be in the relative state. Also, pay attention to the special behavior of the file state.
Some examples:
Base = https://lightpanda.io/
Path                    →  Resolved url
"https:blog"         → "https://lightpanda.io/blog"    // same schemes
"http:blog"           → "http://blog/"                         // different schemes → blog = hostname  
"file:blog"            → "file:///blog"                           // special file → blog = path
[...]

PR #2030, opened 2026-03-29
fix: handle scheme in path for resolving url with specific cases

Dinisimys2018:
Fix related to issue: https://github.com/lightpanda-io/browser/issues/1994
Changes based on [https://url.spec.whatwg.org/](Spec) .
**Relevant sections:**
- https://url.spec.whatwg.org/#special-scheme
- https://url.spec.whatwg.org/#scheme-state (section 5 and 6)
- https://url.spec.whatwg.org/#special-relative-or-authority-state
See examples in test cases.
**Important**:
The current algorithm working for special schemes and only specific non-special schemes ( blob: and data:). At the moment, I haven’t found any other commonly used **custom** schemes.

### DIFF
--- a/src/browser/URL.zig
+++ b/src/browser/URL.zig
@@ -24,22 +24,6 @@ const ResolveOpts = struct {
     always_dupe: bool = false,
 };
 
-const scheme_full_separator = "://";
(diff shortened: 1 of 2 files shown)


Dinisimys2018:

PR #2035, opened 2026-03-30
feat: add --user-agent flag for full User-Agent override

### LINKED ISSUE #2029: Modify User Agent
I am aware of the ability to add a suffix to the user agent:

```
--user-agent-suffix
                Suffix to append to the Lightpanda/X.Y User-Agent
```

However, for anyone who's moving real existing systems from Chrome headless or other, they probably already have custom user agents set up. In our case, lots of our customers have already whitelisted our own user agent in their bot detection systems so we can serve them.

At the moment, my only option is to inform all the customers that the user agent will be changed, and give them a cut off time (in months) otherwise their systems will not be working as expected anymore - which is really a big challenge for both sides.

Any chance we can get the ability to modify the user agent completely?

tmchow:
Adds a `--user-agent` CLI flag that replaces the entire User-Agent header. When set, the provided string is used as-is instead of the default `Lightpanda/1.0` prefix.

This addresses users migrating from Chrome headless who need full UA control. The existing `--user-agent-suffix` behavior is unchanged.

The change is in `Config.zig`:
- New `user_agent` field in `Common` options
- New `userAgent()` accessor (same pattern as `userAgentSuffix()`)
- `HttpHeaders.init` checks `userAgent()` first, falls back to suffix, then default
- Arg parser handles `--user-agent` / `--user_agent`

The CDP `Emulation.setUserAgentOverride` command is still stubbed. Wiring a runtime UA override through the CDP session to the HTTP layer is a larger change that could follow separately.

Fixes #2029

### DIFF
--- a/src/Config.zig
+++ b/src/Config.zig
@@ -156,6 +156,13 @@ pub fn userAgentSuffix(self: *const Config) ?[]const u8 {
     };
 }
 
+pub fn userAgent(self: *const Config) ?[]const u8 {
+    return switch (self.mode) {
+        inline .serve, .fetch, .mcp => |opts| opts.common.user_agent,
+        .help, .version => null,
+    };
+}
+
 pub fn httpCacheDir(self: *const Config) ?[]const u8 {
     return switch (self.mode) {
         inline .serve, .fetch, .mcp => |opts| opts.common.http_cache_dir,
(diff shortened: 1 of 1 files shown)


tmchow:

tmchow:
recheck

krichprollsch:
Merged via #2100 
Thanks for your contribution.

Issue #1962, opened 2026-03-23
Stagehand: `Target.createTarget` fails with `-31998 TargetAlreadyLoaded` on second call


spolu:
# Bug: `Target.createTarget` fails with `-31998 TargetAlreadyLoaded` on second call

## Summary

When using Lightpanda via CDP, creating a second page target fails with error `-31998 TargetAlreadyLoaded`. The first `Target.createTarget` succeeds, but any subsequent call returns this error.

## Steps to Reproduce

Connect to Lightpanda via CDP websocket and execute the following sequence:

```
1. Target.setAutoAttach { autoAttach: true, flatten: true, waitForDebuggerOnStart: true }
2. Target.setDiscoverTargets { discover: true }
3. Target.createTarget { url: "about:blank" }   // ✅ succeeds, returns targetId
4. Target.createTarget { url: "about:blank" }   // ❌ fails: -31998 TargetAlreadyLoaded
```

## Context
This occurs when using [Stagehand](https://github.com/browserbase/stagehand) (v3) which manages pages via its own CDP-based `V3Context`. Stagehand's `context.newPage()` method calls `Target.createTarget` to open new tabs. The relevant code path in Stagehand:

```javascript
// stagehand/dist/esm/lib/v3/understudy/context.js — newPage()
const { targetId } = await this.conn.send("Target.createTarget", { url: "about:blank" });
```

The bootstrap sequence prior to this uses:
```javascript
// stagehand/dist/esm/lib/v3/understudy/cdp.js — enableAutoAttach()
await this.send("Target.setAutoAttach", {
    autoAttach: true,
    flatten: true,
    waitForDebuggerOnStart: true,
});
await this.send("Target.setDiscoverTargets", { discover: true });
```

## Expected Behavior
Multiple `Target.createTarget` calls should each create a new independent page target, as they do in Chrome.=

## Actual Behavior
The second `Target.createTarget` call returns error `-31998 TargetAlreadyLoaded`.

## Environment
- Lightpanda connected via CDP websocket (port 9222)
- Stagehand v3 (`@browserbasehq/stagehand`)
- Node.js / TypeScript

krichprollsch:
Hello @spolu, thanks for the issue.
Lightpanda doesn't support multiple targets with the same CDP connection for now.
We support multiple connections, each as its own browser, but you open one page par connection at a time.
It's a feature we plan to implement in the future, but it's a big task to do and it will take time to release...

[...]

PR #2039, opened 2026-03-30
fix(cdp): auto-close existing target on createTarget

arrufat:
When `Target.createTarget` is called and a target already exists in the browser context, the existing target is now automatically closed and detached instead of returning a `TargetAlreadyLoaded` error. This unblocks automation frameworks (like Stagehand) that call `createTarget` multiple times sequentially.

Note: this does not add concurrent multi-tab support; Lightpanda still supports one page per connection. The existing target is replaced, not kept alongside the new one.

Addresses #1962

### DIFF
--- a/src/cdp/domains/target.zig
+++ b/src/cdp/domains/target.zig
@@ -471,6 +471,10 @@ fn doCloseTarget(cmd: anytype, bc: anytype) !void {
         bc.session_id = null;
     }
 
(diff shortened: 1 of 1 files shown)


krichprollsch:
As we discussed, creating new target is on our roadmap.
I'm afraid faking the creation by closing/creating new target will drive client in a wrong state since it expects with the feature the previous target is still accessible.

PR #2040, opened 2026-03-30
Url/resolve scheme in path

karlseguin:
Takes inspiration from https://github.com/lightpanda-io/browser/pull/2030 and fixes https://github.com/lightpanda-io/browser/issues/1994

A url like http:/test gets special treatment. If the scheme, `http:` matches the base scheme, then it's treated as relative to the base. If it doesn't match the base scheme, then it's normalized to http://test, e.g. the path becomes the host.

### DIFF
--- a/src/browser/URL.zig
+++ b/src/browser/URL.zig
@@ -24,125 +24,101 @@ const ResolveOpts = struct {
     always_dupe: bool = false,
 };
 
-const scheme_full_separator = "://";
-const special_schemes = [_][]const u8{ "https", "http", "ws", "wss", "file", "ftp" };
-
-fn isSpecialScheme(scheme: []const u8) bool {
-    if (scheme.len == 0 or scheme.len > 5) {
-        return false;
-    }
-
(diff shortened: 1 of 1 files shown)


Dinisimys2018:
@karlseguin Hi. I not understand, why not using ```errdefer allocator.free(path);``` ?

karlseguin:
> @karlseguin Hi. I not understand, why not using `errdefer allocator.free(path);` ?

The allocator passed to resolve is always an ArenaAllocator. It's an internal convention. Admittedly, there's a risk that one day someone wants to use a non-ArenaAllocator, and then we'd have a leak.

Dinisimys2018:
@karlseguin  Thank you !
If you don’t mind, I’ve opened a new [PR](https://github.com/lightpanda-io/browser/pull/2046) incorporating your suggested adjustments. However, I still believe that isCompleteHttpUrl is not appropriate in this case, as there are many edge cases.
I’ve verified all these cases using the Chrome Dev Console.

PR #2046, opened 2026-03-30
fix: url resolve path scheme

Dinisimys2018:
Fix related to issue: https://github.com/lightpanda-io/browser/issues/1994
Path url with a scheme can be resolve as absolute URLs or as relative paths, depending on the context.
General example:
1. base = https://host/a | path = https:/b | url = https://host/b (same schemes and resolve path as relative)
2. base = https://host/a | path = http:/b | url = http://b (different schemes and resolve path as absolute)

See more examples in test cases.

Changes based on [https://url.spec.whatwg.org/](Spec) .
**Relevant sections:**
- https://url.spec.whatwg.org/#special-scheme
- https://url.spec.whatwg.org/#scheme-state (section 5 and 6)
- https://url.spec.whatwg.org/#special-relative-or-authority-state

### DIFF
--- a/src/browser/URL.zig
+++ b/src/browser/URL.zig
@@ -58,12 +58,12 @@ pub fn resolve(allocator: Allocator, base: [:0]const u8, source_path: anytype, c
                                 while (rest_start < path.len and (path[rest_start] == '/' or path[rest_start] == '\\')) {
                                     rest_start += 1;
                                 }
-                                //special scheme need to any symbols after "://"
-                                if (rest_start >= path.len) {
-                                    return error.InvalidURL;
+                                // A special scheme (exclude "file") must contain at least anu chars after "://"
+                                if (rest_start == path.len and !std.ascii.eqlIgnoreCase(scheme_path, "file")) {
+                                    return error.TypeError;
                                 }
                                 //File scheme allow empty host
-                                const separator: []const u8 = if (std.ascii.eqlIgnoreCase(scheme_path, "file") and !has_double_slashas) ":///" else "://";
+                                const separator: []const u8 = if (!has_double_slashas and std.ascii.eqlIgnoreCase(scheme_path, "file")) ":///" else "://";
 
                                 path = try std.mem.joinZ(allocator, "", &.{ scheme_path, separator, path[rest_start..] });
                                 return processResolved(allocator, path, opts);
@@ -1694,6 +1694,12 @@ test "URL: resolve path scheme" {
             .path = "file:/path/to/file",
             .expected = "file:///path/to/file",
         },
+        //different schemes and path as absolute (path scheme=file, host is empty)
+        .{
+            .base = "https://www.example.com/example",
+            .path = "file:/",
+            .expected = "file:///",
+        },
         //different schemes without :// and normalize "file" scheme, absolute path
         .{
             .base = "https://www.example.com/example",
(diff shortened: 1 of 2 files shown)


src/browser/URL.zig
```diff
@@ -25,28 +25,54 @@ const ResolveOpts = struct {
...
-            const duped = try allocator.dupeZ(u8, base);
-            return processResolved(allocator, duped, opts);
-        }
-        if (comptime opts.encode) {
-            return processResolved(allocator, base, opts);
+pub fn resolve(allocator: Allocator, base: [:0]const u8, source_path: anytype, comptime opts: ResolveOpts) ![:0]const u8 {
+    const PT = @TypeOf(source_path);
+
+    if (source_path.len == 0) {
+        return processResolved(allocator, base, opts);
```
karlseguin: `base` should be dupe'd if `opts.always_dupe` is true.

src/browser/URL.zig
```diff
@@ -25,28 +25,54 @@ const ResolveOpts = struct {
...
+    }
+
+    // Minimum is "x:" and skip relative path (very common case)
+    if (path.len >= 2 and path[0] != '/') {
+        if (std.mem.indexOfScalar(u8, path[0..], ':')) |scheme_path_end| {
+            scheme_check: {
+                //from "ws" to "https"
+                if (scheme_path_end >= 2 and scheme_path_end <= 5) {
+                    const scheme_path = path[0..scheme_path_end];
+                    const has_double_sleshes: bool = path[scheme_path_end + 1] == '/' and path[scheme_path_end + 2] == '/';
```
karlseguin: I think this can out of bounds if the path is just something like `http:`

karlseguin:
```suggestion
                    const has_double_slashes: bool = path[scheme_path_end + 1] == '/' and path[scheme_path_end + 2] == '/';
```

Dinisimys2018:
> In addition to a few inline comments...
> 
> `isCompleteHTTPUrl` is used elsewhere. So if it's incorrect, that could be a problem there also. Also, isn't any url with a `://` unambiguously a complete URL that can be returned as is (or duped if necessary)? Seems like a common case and it would be nice to keep that early exit/check.

Interestingly, a path like _http:path.some_ is already considered **absolute**. However, we still need to normalize it — we cannot just return it as-is.
A path like _http:/path.some_ is also considered **absolute** if the path scheme differs from the base scheme.
At the same time, _http:/path.some_ is considered **relative** if the path scheme is the same as the base scheme.
This implies that the _isCompleteHTTPUrl_ check depends on the relationship between base and path.
Most likely, we can release the current changes and then further investigate how _isCompleteHTTPUrl_ is used outside of URL.resolve().

karlseguin:
Ok. But still need the possible overflow and missing dupe fixed (and the typo ideally).

src/browser/URL.zig
Dinisimys2018: Yes, I missed this point, fixed.
Note: I moved this check to occur after path normalization. This does not degrade the algorithm, because we check that path.len >= 2 (so exclude path.len == 0) before normalization, and after normalization the path may become empty. For example, the path "http:" is normalized to an empty path (path = ""). I did it to avoid duplicating the path.len == 0 check before and after path normalization .

src/browser/URL.zig
Dinisimys2018: Yes, added check path len , "http://".len = 4+3 = 7, so check `scheme_path_end + 3 <= path.len`

Dinisimys2018: Yes, rename tp `has_double_slashes` ))

Dinisimys2018:
> Ok. But still need the possible overflow and missing dupe fixed (and the typo ideally).

@karlseguin  Fixed your comments and check working with cases: `http:` `file:` etc. (file scheme is specifically).
**Note: I added error.TypeError** 
This is a very tricky area — introducing a new error where previously there was only an OutOfMemory error.
I checked other places and, as far as I understand, this error is valid here.
For example:
path = https://
base = http://site/b/mmm
So, 1 - other scheme, 2 - not contains any symbols after "://"
image
Specific `file` scheme:
image

PR #2047, opened 2026-03-31
Add --wait-selector, --wait-script and --wait-script-file options to …

karlseguin:
…fetch

These new optional parameter run AFTER --wait-until, allowing the (imo) useful combination of `--wait-until load --wait-script "report.complete === true"`. However, if `--wait-until` IS NOT specified but `--wait-selector/script` IS, then there is no default wait and it'll just check the selector/script. If neither `--wait-selector` or `--wait-script/--wait-script-file` are specified
 then  `--wait-until` continues to default to `done`.

These waiters were added to the Runner, and the existing Action.waitForSelector now uses the runner's version. Selector querying has been split into distinct parse and query functions, so that we can parse once, and query on every tick.

We could potentially optimize --wait-script to compile the script once and call it on each tick, but we'd have to detect page navigation to recompile the script in the new context. Something I'd rather optimize separately.

### DIFF
--- a/src/browser/actions.zig
+++ b/src/browser/actions.zig
@@ -114,6 +114,6 @@ pub fn waitForSelector(selector: [:0]const u8, timeout_ms: u32, session: *Sessio
     const remaining = timeout_ms -| elapsed;
     if (remaining == 0) return error.Timeout;
 
-    const el = try runner.waitForSelector(selector, timeout_ms);
+    const el = try runner.waitForSelector(selector, remaining);
     return el.asNode();
 }
--- a/src/lightpanda.zig
+++ b/src/lightpanda.zig
@@ -129,14 +129,14 @@ pub fn fetch(app: *App, url: [:0]const u8, opts: FetchOpts) !void {
         const elapsed: u32 = @intCast(timer.read() / std.time.ns_per_ms);
         const remaining = opts.wait_ms -| elapsed;
         if (remaining == 0) return error.Timeout;
-        _ = try runner.waitForSelector(selector, opts.wait_ms);
+        _ = try runner.waitForSelector(selector, remaining);
     }
 
     if (opts.wait_script) |script| {
         const elapsed: u32 = @intCast(timer.read() / std.time.ns_per_ms);
         const remaining = opts.wait_ms -| elapsed;
         if (remaining == 0) return error.Timeout;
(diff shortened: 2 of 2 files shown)


src/browser/actions.zig
```diff
@@ -110,28 +110,10 @@ pub fn waitForSelector(selector: [:0]const u8, timeout_ms: u32, session: *Sessio
...
-            .done => return error.Timeout,
-            .ok => |recommended_sleep_ms| {
-                if (recommended_sleep_ms > 0) {
-                    // guanrateed to be <= 20ms
-                    std.Thread.sleep(std.time.ns_per_ms * recommended_sleep_ms);
-                }
-            },
-        }
-    }
+    const el = try runner.waitForSelector(selector, timeout_ms);
```
arrufat:
```suggestion
    const el = try runner.waitForSelector(selector, remaining);
```

src/lightpanda.zig
```diff
@@ -111,7 +113,31 @@ pub fn fetch(app: *App, url: [:0]const u8, opts: FetchOpts) !void {
...
+        // This allows the caller to ONLY --wait-selector or ONLY --wait-script
+        // or combine --wait-until WITH --wait-selector/script
+        try runner.wait(.{ .ms = opts.wait_ms, .until = .done });
+    }
+
+    if (opts.wait_selector) |selector| {
+        const elapsed: u32 = @intCast(timer.read() / std.time.ns_per_ms);
+        const remaining = opts.wait_ms -| elapsed;
+        if (remaining == 0) return error.Timeout;
+        _ = try runner.waitForSelector(selector, opts.wait_ms);
```
arrufat:
```suggestion
        _ = try runner.waitForSelector(selector, remaining);
```

src/lightpanda.zig
```diff
@@ -111,7 +113,31 @@ pub fn fetch(app: *App, url: [:0]const u8, opts: FetchOpts) !void {
...
+    }
+
+    if (opts.wait_script) |script| {
+        const elapsed: u32 = @intCast(timer.read() / std.time.ns_per_ms);
+        const remaining = opts.wait_ms -| elapsed;
+        if (remaining == 0) return error.Timeout;
+        try runner.waitForScript(script, opts.wait_ms);
```
arrufat:
```suggestion
        try runner.waitForScript(script, remaining);
```

Issue #1738, opened 2026-03-08
SIGSEGV when fetching https://www.nist.gov/


Zabrane:
## Description

Lightpanda crashes with a segmentation fault (SIGSEGV, exit code 139) when fetching `https://www.nist.gov/`. The crash is reproducible on every attempt.

## Reproduction

```bash
lightpanda fetch --dump html https://www.nist.gov/
# macOS: exit code 139 (128 + 11 = SIGSEGV)
# Linux: "Segmentation fault (core dumped)"
```

## Environment

Reproduced on both:

- **macOS Tahoe** (Darwin 26.3), arm64 (Apple Silicon)
- **Linux** (x86_64): Ubuntu 22.04LTS / 24.04LTS
- **Lightpanda version**: `21313adf`

## Core dump

A core dump (689 MB) from the Linux crash is available on request.

karlseguin:
Hi. Sorry about that. The issue is fixed in the latest nightly.

PR #2050, opened 2026-03-31
fix: handle null pointer from GetAlignedPointerFromInternalField to prevent SIGSEGV

soilSpoon:
`v8__Object__GetAlignedPointerFromInternalField` can return null even when `InternalFieldCount > 0` — for example when a JS library passes objects whose internal fields are allocated but not yet initialized.

The previous code used `.?` (unsafe unwrap) on the return value, which panics on null and causes SIGSEGV. This replaces `.?` with `orelse` to gracefully handle null:

- **TaggedOpaque.zig**: returns `error.InvalidArgument` (same pattern as the `internal_field_count == 0` check above)
- **Inspector.zig**: returns `null` (matches the function's `?*TaggedOpaque` return type and existing null returns)

## Use case

I'm building a browser automation tool for [wishket.com](https://www.wishket.com/) (a Korean freelancing platform) using Lightpanda's CDP server with Playwright. Lightpanda crashes with SIGSEGV every time it loads the site because wishket.com includes Sentry's `bundle.tracing.min.js`, which extensively probes DOM APIs and passes objects that trigger the null pointer path.

## Minimal reproduction

```javascript
// test-sigsegv.mjs — requires: playwright-core
import { spawn } from 'child_process';
import { chromium } from 'playwright-core';

const proc = spawn('lightpanda', [
  'serve', '--host', '127.0.0.1', '--port', '9222',
  '--log-level', 'warn', '--timeout', '30',
], { stdio: ['ignore', 'pipe', 'pipe'] });

proc.stderr.on('data', d => process.stderr.write(d));
proc.on('exit', (code, signal) => console.log(`EXIT code=${code} signal=${signal}`));
await new Promise(r => setTimeout(r, 2000));

const browser = await chromium.connectOverCDP('ws://127.0.0.1:9222');
const ctx = await browser.newContext({});

// Crashes with SIGSEGV
await page.goto('https://www.wishket.com/', { timeout: 15000, waitUntil: 'domcontentloaded' });

await page.close();
await ctx.close();
await browser.close();
proc.kill();
```

## Workaround

Blocking tracking scripts via `page.route()` prevents the crash:

```javascript
const BLOCK = ['sentry-cdn.com', 'clarity.ms', 'googletagmanager.com',
  'googleoptimize.com', 'doubleclick.net', 'channel.io'];
await page.route(u => BLOCK.some(d => u.toString().includes(d)), r => r.abort());
await page.goto('https://www.wishket.com/', { timeout: 15000, waitUntil: 'domcontentloaded' });
// Works — title: "대한민국 대표 IT 프로젝트 플랫폼 · 위시켓(Wishket)", body: 11032 chars
```

## Debug log (last lines before crash)

```
INFO  browser : executing script  src = https://www.wishket.com/static/renewal/js/moment.min.js
DEBUG http : script fetch start   req = GET https://browser.sentry-cdn.com/7.8.0/bundle.tracing.min.js
[LP EXIT] code=null signal=SIGSEGV
```

## Tested sites

| Site | Result | Tracking JS |
|------|--------|------------|
| wishket.com | SIGSEGV | Sentry + Clarity + GTM |
| naver.com | SIGSEGV | Naver analytics + CDN |
| wikipedia.org | OK | No Sentry |
| github.com | OK | Lightweight JS |

## Environment

- Version: `1.0.0-nightly.5301+a6a87527`
- OS: macOS Darwin 25.4.0 arm64 (Apple Silicon)

Related: #1738

### DIFF
--- a/src/browser/js/Inspector.zig
+++ b/src/browser/js/Inspector.zig
@@ -368,7 +368,9 @@ pub fn getTaggedOpaque(value: *const v8.Value) ?*TaggedOpaque {
         return null;
     }
 
-    const tao_ptr = v8.v8__Object__GetAlignedPointerFromInternalField(value, 0).?;
+    const tao_ptr = v8.v8__Object__GetAlignedPointerFromInternalField(value, 0) orelse {
+        return null;
+    };
     return @ptrCast(@alignCast(tao_ptr));
 }
 
--- a/src/browser/js/TaggedOpaque.zig
+++ b/src/browser/js/TaggedOpaque.zig
@@ -106,7 +106,9 @@ pub fn fromJS(comptime R: type, js_obj_handle: *const v8.Object) !R {
         @compileError("unknown Zig type: " ++ @typeName(R));
     }
 
-    const tao_ptr = v8.v8__Object__GetAlignedPointerFromInternalField(js_obj_handle, 0).?;
+    const tao_ptr = v8.v8__Object__GetAlignedPointerFromInternalField(js_obj_handle, 0) orelse {
+        return error.InvalidArgument;
+    };
     const tao: *TaggedOpaque = @ptrCast(@alignCast(tao_ptr));
     const expected_type_index = bridge.JsApiLookup.getId(JsApi);
 


soilSpoon:

karlseguin:
Hi. Sorry for the delay. I'm skeptical about merging this because I feel that it will just hide real problems. We have seen this crash before, but it was always something that should have been fixed elsewise. The most recent fix for this was https://github.com/lightpanda-io/browser/pull/1991

I haven't been able to reproduce the crash on wishket.com. Can you confirm that you're still seeing this with the lastest nightly?

soilSpoon:
> Hi. Sorry for the delay. I'm skeptical about merging this because I feel that it will just hide real problems. We have seen this crash before, but it was always something that should have been fixed elsewise. The most recent fix for this was #1991
> 
> I haven't been able to reproduce the crash on wishket.com. Can you confirm that you're still seeing this with the lastest nightly?

Investigation complete. Closing this PR as the root cause has been fixed differently in main.

## Root Cause (verified)

The crash is **use-after-free**, not a null pointer:

1. DOM objects created in context A have their TAO allocated in `context_arena_A`, with the raw pointer stored in V8's internal field via `SetAlignedPointerInInternalField`
2. When navigating cross-origin, context A is destroyed → `context_arena_A` is freed → **TAO memory released**
3. The V8 object survives in `identity_map` (session-level), still holding the now-dangling pointer
4. When V8 Inspector initializes the new context, it traverses the heap and calls `valueSubtype` on the old object → `GetAlignedPointerFromInternalField` returns the dangling pointer → accessing `tao.subtype` → **SIGSEGV**

The `.?` force-unwrap doesn't actually trigger the crash (the dangling pointer is non-null). The crash happens on the subsequent field access.

## Fix already in main

`76db8170` ("Store TAO in IdentityMap", Apr 3 2026) resolved this by moving TAO allocation from `context_arena` to `ctx.identity_arena` (session-level):

```diff
- const tao = try context_arena.create(TaggedOpaque);
+ const tao = try ctx.identity_arena.create(TaggedOpaque);
```

## Reproduction verified

- `a6a87527` (pre-fix): navigate wishket.com → naver.com with CDP connected → **SIGSEGV** (process dies)  
- `b6020e47` (current main, post-fix): same sequence → **survives**

Closing in favor of `76db8170`.

PR #2052, opened 2026-03-31
Expand the lifetime of the XHR reference

karlseguin:
We need to take the self-reference to the XHR object as soon as the request is made. Previously, we were waiting until we got the start callback, but v8 could (and does) drop the reference before that happens. Unfortunately, that means we can no longer use _transfer == null to tell if we own a reference or not, so a new boolean was added.

### DIFF
--- a/src/browser/webapi/net/XMLHttpRequest.zig
+++ b/src/browser/webapi/net/XMLHttpRequest.zig
@@ -502,7 +502,7 @@ fn httpDoneCallback(ctx: *anyopaque) !void {
         .loaded = loaded,
     }, page);
 
-    self.releaseRef(page._session);
+    self.releaseSelfRef();
 }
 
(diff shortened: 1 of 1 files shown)


src/browser/webapi/net/XMLHttpRequest.zig
```diff
@@ -252,6 +261,8 @@ pub fn send(self: *XMLHttpRequest, body_: ?[]const u8) !void {
         .error_callback = httpErrorCallback,
         .shutdown_callback = httpShutdownCallback,
     });
+    self.acquireRef();
+    self._has_ref = true;
```
krichprollsch: shouldn't we move this line into `acquireRef`?

karlseguin: No. `acquireRef` is called by v8.  Maybe `_has_ref` should be renamed to `_zig_has_ref` or `_active_request` ?

Issue #1952, opened 2026-03-21
Attempt to load a page that uses WebSocket resulted in "WebSocket is not Defined"


bosunski:
Hello there,

Thank you for this awesome piece of software.

I'm trying to replace my current headless chrome setup with Lightpanda but when I load a page that uses WebSocket , I found out that the page is always broken (not rendered). I try to investigate this by running `lightpanda fetch https://my-page-with-websocket` and got this warning:

```
$time=1774114854971 $scope=js $level=warn $msg=console.error param.0=[vite] failed to connect to websocket (ReferenceError: WebSocket is not defined).
```

The same page does work with headless chrome which make me wonder if Lightpanda supports WebSocket or not.

Happy to provide more information as needed.

Thank you!

karlseguin:
Hi @bosunski it doesn't, though we've talked about it. I do think it's a pretty simple feature to add, and it would help compatibility. There's some work being done on the network layer right now, but I think we can get to it after the current work stabilizes.

bosunski:
@karlseguin thanks for the feedback.

Is there a public compatibility note somewhere for reference? I was trying to find this at the time of the issue but couldn't. Will be good to know how Lightpanda compares with other browsers.

krichprollsch:
@bosunski you can check a (very) detailed result of web platform test at https://perf.lightpanda.io/wpt

I will work on a test result  summary.

krichprollsch:
@bosunski thanks to #2063 a support of websockets is available. Would you like to make a try?

ShalokShalom:
Can I see documentation for this one?

karlseguin:
@ShalokShalom what exactly are you looking for? The implementation was relatively straightforward, as libcurl supports websockets. It fit pretty smoothly into the existing network code / loop. As for the API itself: https://developer.mozilla.org/en-US/docs/Web/API/WebSocket

PR #2063, opened 2026-04-01
WebSocket WebAPI

karlseguin:
Uses libcurl's websocket capabilities to add support for WebSocket.

Depends on https://github.com/lightpanda-io/zig-v8-fork/pull/167 Issue: https://github.com/lightpanda-io/browser/issues/1952

This is a WIP because it currently uses the same connection pool used for all HTTP requests. It would be pretty easy for a page to starve the pool and block any progress.

We previously stored the *Transfer inside of the easy's private data. We now store the *Connection, and a Connection now has a `transport` field which is a union for `http: *Transfer` or `websocket: *Websocket`.

### DIFF
--- a/src/browser/HttpClient.zig
+++ b/src/browser/HttpClient.zig
@@ -270,7 +270,11 @@ fn _abort(self: *Client, comptime abort_all: bool, frame_id: u32) void {
         var leftover: usize = 0;
         while (it) |node| : (it = node.next) {
             const conn: *http.Connection = @fieldParentPtr("node", node);
-            std.debug.assert((Transfer.fromConnection(conn) catch unreachable).aborted);
+            switch (conn.transport) {
+                .http => |transfer| std.debug.assert(transfer.aborted),
+                .websocket => {},
+                .none => {},
+            }
             leftover += 1;
         }
         std.debug.assert(self.active == leftover);
@@ -1015,7 +1019,7 @@ fn processMessages(self: *Client) !bool {
                     ws.disconnected(null);
                 }
 
-                return true;
+                processed = true;
             },
             .none => unreachable,
         }
--- a/src/browser/tests/net/websocket.html
+++ b/src/browser/tests/net/websocket.html
@@ -304,35 +304,6 @@
 }
 </script>
 
-<script id=binary_int32array22 type=module>
-{
-  const state = await testing.async();
-  let received = [];
-
-  let ws = new WebSocket('ws://127.0.0.1:9584/');
-  ws.binaryType = 'arraybuffer';
-
-  ws.addEventListener('open', () => {
-    const arr = new Int32Array([0x01020304, 0x05060708]);
-    ws.send(arr);
-  });
-
-  ws.addEventListener('message', (e) => {
-    received.push(e.data.byteLength);
-    ws.close();
-  });
-
-  ws.addEventListener('close', () => {
-    state.resolve();
-  });
-
-  await state.done(() => {
-    // 1 marker byte + 8 bytes (2 x 4-byte int32)
-    testing.expectEqual([9], received);
-  });
-}
-</script>
-
 <script id=binary_blob type=module>
 {
   const state = await testing.async();
@@ -412,8 +383,8 @@
   });
 
   await state.done(() => {
-    // Connection was not cleanly closed
-    testing.expectEqual(['closed', false], received);
+    // Connection was not cleanly closed - error fires before close
+    testing.expectEqual(['error', 'closed', false], received);
   });
 }
 </script>
@@ -579,6 +550,14 @@
 
   await state.done(() => {
     testing.expectEqual([true, 6, 0xEE, 1], received);
+  });
+}
+</script>
+
+<script id=url_with_fragment_rejected>
+{
+  testing.expectError('SyntaxError', () => {
+    new WebSocket('ws://127.0.0.1:9584/#fragment');
   });
 }
 </script>
--- a/src/browser/webapi/net/WebSocket.zig
+++ b/src/browser/webapi/net/WebSocket.zig
@@ -99,6 +99,10 @@ pub fn init(url: []const u8, protocols_: ?[]const u8, page: *Page) !*WebSocket {
         if (!std.mem.startsWith(u8, normalized_start, "ws://") and !std.mem.startsWith(u8, normalized_start, "wss://")) {
             return error.SyntaxError;
         }
+        // Fragments are not allowed in WebSocket URLs
+        if (std.mem.indexOfScalar(u8, url, '#') != null) {
+            return error.SyntaxError;
+        }
     }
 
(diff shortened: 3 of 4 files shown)


src/browser/webapi/net/WebSocket.zig
```diff
@@ -0,0 +1,697 @@
...
+    }
+    return buffered;
+}
+
+pub fn getProtocol(self: *const WebSocket) []const u8 {
+    return self._protocol;
+}
+
+pub fn getExtensions(self: *const WebSocket) []const u8 {
+    return self._extensions;
```
krichprollsch: `_extensions` doesn't exists and `getExtensions` is not called. Maybe we should remove it. Same for `getProtocol`.

src/browser/HttpClient.zig
```diff
@@ -871,30 +872,63 @@ fn processOneMessage(self: *Client, msg: http.Handles.MultiMessage, transfer: *T
...
+            .websocket => |ws| {
+                if (msg.err) |err| switch (err) {
+                    error.GotNothing => ws.disconnected(null),
+                    else => ws.disconnected(err),
+                } else {
+                    // Clean close - no error
+                    ws.disconnected(null);
+                }
+
+                return true;
```
krichprollsch: Should we return here? shouldn't continue the loop instead?

src/network/http.zig
```diff
@@ -299,20 +305,62 @@ pub const Connection = struct {
...
+        green: bool = false,
+        blue: bool = false,
+        alpha: bool = false,
+        // Optional padding to match a specific size, e.g., a u32
+        _padding: u28 = 0,
+    };
+
+    pub fn pause(
+        self: *Connection,
+        flags: libcurl.CurlPauseFlags,
```
krichprollsch, line 354: You declare `PauseFlags` but you use ` libcurl.CurlPauseFlags` here. Is `PauseFlags` useful?

src/browser/tests/net/websocket.html
```diff
@@ -0,0 +1,584 @@
...
+  });
+
+  await state.done(() => {
+    // 1 marker byte + 8 bytes (2 x 4-byte int32)
+    testing.expectEqual([9], received);
+  });
+}
+</script>
+
+<script id=binary_int32array22 type=module>
```
krichprollsch: `binary_int32array22` seems to be the exact same test than `binary_int32array`

src/browser/webapi/net/WebSocket.zig
```diff
@@ -0,0 +1,697 @@
...
+            self._ready_state = .closing;
+            self.disconnected(null);
+        },
+        .ping, .pong, .cont => {},
+    }
+}
+
+// libcurl has no mechanism to signal that the connection is established. The
+// best option I could come up with was looking for an upgrade header response.
+fn receivedHeaderCalllback(buffer: [*]const u8, header_count: usize, buf_len: usize, data: *anyopaque) usize {
```
krichprollsch:
```suggestion
fn receivedHeaderCallback(buffer: [*]const u8, header_count: usize, buf_len: usize, data: *anyopaque) usize {
```

src/browser/webapi/net/WebSocket.zig
```diff
@@ -0,0 +1,697 @@
...
+
+    if (err_) |err| {
+        log.warn(.websocket, "disconnected", .{ .err = err, .url = self._url });
+    } else {
+        log.info(.websocket, "disconnected", .{ .url = self._url, .reason = "closed" });
+    }
+
+    self.cleanup();
+
+    // Use 1006 (abnormal closure) if connection wasn't cleanly closed
```
krichprollsch, line 190: Claude complains about calling _on_error on abnormal closure

> No error event dispatch — The spec requires an error event before close on abnormal closures. The _on_error handler is stored but never invoked.

src/browser/webapi/net/WebSocket.zig
```diff
@@ -0,0 +1,697 @@
...
+            // Parse close frame: 2-byte code (big-endian) + optional reason
+            self._close_code = if (message.len >= 2)
+                @as(u16, message[0]) << 8 | message[1]
+            else
+                1005; // No status code received
+            if (message.len > 2) {
+                self._close_reason = try self._arena.dupe(u8, message[2..]);
+            }
+            self._ready_state = .closing;
+            self.disconnected(null);
```
krichprollsch: Claude complains about close frame back.

> No reciprocal close frame (WebSocket.zig:573-587) — On server-initiated close, the client should send a close frame back per RFC 6455 §5.5.1. Currently it just disconnects.

src/browser/webapi/net/WebSocket.zig
```diff
@@ -0,0 +1,697 @@
...
+                try self.queueMessage(.{ .binary = .{
+                    .arena = arena,
+                    .data = try arena.dupe(u8, binary.asBuffer()),
+                } });
+            }
+        },
+    }
+}
+
+pub fn close(self: *WebSocket, code_: ?u16, reason_: ?[]const u8) !void {
```
krichprollsch, line 303: it seems we should validate the code?
> close(code) doesn't validate the code — Per spec, should throw InvalidAccessError if code is not 1000 and not in range 3000-4999.

src/browser/webapi/net/WebSocket.zig
```diff
@@ -0,0 +1,697 @@
...
+    closing = 2,
+    closed = 3,
+};
+
+pub const BinaryType = enum {
+    blob,
+    arraybuffer,
+};
+
+pub fn init(url: []const u8, protocols_: ?[]const u8, page: *Page) !*WebSocket {
```
krichprollsch, line 87: Claude says we should reject URL including fragments
>  URL validation is incomplete (WebSocket.zig:94-102) — Doesn't reject fragments (spec requires SyntaxError).

src/browser/webapi/net/WebSocket.zig
```diff
@@ -0,0 +1,697 @@
...
+    // Use 1006 (abnormal closure) if connection wasn't cleanly closed
+    const code = if (was_clean) self._close_code else 1006;
+    const reason = if (was_clean) self._close_reason else "";
+
+    self.dispatchCloseEvent(code, reason, was_clean) catch |err| {
+        log.err(.websocket, "close event dispatch failed", .{ .err = err });
+    };
+}
+
+fn cleanup(self: *WebSocket) void {
```
krichprollsch, line 206: should we clean send_queue also?

src/browser/HttpClient.zig
karlseguin: Yes, you're right. I was copying what http does..but http doesn't return true, it set processed to true and that eventually returns.

src/browser/webapi/net/WebSocket.zig
karlseguin: It shouldn't matter, but better safe than sorry.

PR #2066, opened 2026-04-01
mcp: improve navigation reliability and add CDP support

arrufat:


src/Server.zig
```diff
@@ -343,6 +348,24 @@ pub const Client = struct {
         }
     }
 
+    fn isBusy(self: *const Client) bool {
```
karlseguin: This implies that the .done signal from Session.wait is wrong. I'd rather fix Session.wait. I don't think our websocket server should have all this knowledge.

src/browser/Runner.zig
```diff
@@ -94,7 +94,18 @@ fn _wait(self: *Runner, comptime is_cdp: bool, opts: WaitOpts) !CDPWaitResult {
 
         const ms_elapsed = timer.lap() / 1_000_000;
         if (ms_elapsed >= ms_remaining) {
-            return .done;
+            // Don't timeout if there's still active work (HTTP requests,
+            // intercepted requests, background JS tasks, or pending macrotasks).
+            if (self.http_client.active > 0 or self.http_client.intercepted > 0) {
```
karlseguin: What's the point of the timeout then? How do you stop a page from loading that uses a setInterval and never cancels it?

arrufat: You're right, this disables the timeout for pages with recurring activity like `setInterval`. Fixed it to use an absolute timeout, which now always fires after the specified duration regardless of active work.

src/browser/Runner.zig
```diff
@@ -92,11 +91,10 @@ fn _wait(self: *Runner, comptime is_cdp: bool, opts: WaitOpts) !CDPWaitResult {
             .cdp_socket => if (comptime is_cdp) return .cdp_socket else unreachable,
         };
 
-        const ms_elapsed = timer.lap() / 1_000_000;
-        if (ms_elapsed >= ms_remaining) {
-            return .done;
+        const ms_elapsed: u32 = @intCast(timer.read() / std.time.ns_per_ms);
+        if (ms_elapsed >= opts.ms) {
+            return error.Timeout;
```
karlseguin: If `fetch` hits this, it exits with a:

```
FATAL app : fetch error . . . . . . . . . . . . . . . . . . . [+2157ms]
      err = Timeout
      url = http://github.com/lightpanda-io/browser
```

Which makes it seem like a failure.  For a lot of "modern" sites, this is normal / expected / unavoidable. On the one hand, I agree that it's better to surface this to the caller (But it needs to be handled better). On the other, I almost just want to assume it'll _always_ return this error (I know it won't, but I think that's a reasonable mental model, because there is no difference if it returns error.Timeout vs .done).

src/Server.zig
```diff
@@ -326,19 +327,8 @@ pub const Client = struct {
...
-                    const now = milliTimestamp(.monotonic);
-                    const elapsed = now - last_message;
-                    if (elapsed >= ms_remaining) {
-                        log.info(.app, "CDP timeout", .{});
-                        return;
-                    }
-                    ms_remaining -= @intCast(elapsed);
-                    last_message = now;
                 },
+                .done => unreachable,
```
karlseguin: If this is true, doesn't it mean CDPWaitResult should really just be a single enum `cdp_socket` and if that's the case, why no just return `!void` ?

PR #2067, opened 2026-04-01
percent encode version query string for crash report

krichprollsch:


krichprollsch:
~~ok this is a bad way, we can create the percent encoded at build time~~ fixed

PR #2076, opened 2026-04-03
build with sqlite3, just to see how it behaves in the CI

karlseguin:


karlseguin:
Do no merge Just seeing how it behaves through the CI...too lazy to start jobs individually, opening a PR is easier.

Issue #1096, opened 2025-09-28
Multi-context & multi-page environment


mrdimidium:
First of all, thank's for the amazing project!

I would be interested about multiple pages and page persistence support (cookies, localStorage, etc.).

The simplest example that comes to mind: we want to test that collaborative editing works in Google Docs. We open two tabs (they share a single authorization cookie), change the text in the first, and check that the changes have been applied in the second.

In the current version, we could spin up two instances, but then we can't verify that local synchronization is working (like Logux, which uses a localStorage subscription).

I guess your cloud runs multiple instances, and the load balancer allows to connect to a free instance. But it seems that this approach requires building your own balancer, which in the case of Chromium is the browser process.

It also seems that without support for multiple JS environments, you can't support service workers, web workers, and Chrome-style extensions.

I understand that you work with a small team with limited resources, I'm just interested to know your long-term plans.

P.S. Related to https://github.com/lightpanda-io/browser/issues/882

krichprollsch:
Hello @mrdimidium,

Multi-context and Multi-page is in one of our goal for the near future (~6 months), but it requires important changes in the browser.
The main one is the way we are using libdom, for now we can't have two DOM tree open at the same time.

mrdimidium:
Sounds great! Am I correct in understanding the problem that libdom has some global objects that prevent multiple instances from running in a single process?

Do you think the lightpanda implementation will be multi-process (like blink/gecko/servo/webkit2) or will all pages run in a shared process? Will pages have their own sandboxes, or is isolation implied at a higher level (for example, in a container)?

mquandalle:
I'm just curious, does the work on replacing [LibDOM with your own Zig implementation](https://lightpanda.io/blog/posts/migrating-our-dom-to-zig) helps with this issue?

mrdimidium:
Hello @mquandalle. I'm happy to announce that I joined lightpanda to develop this :)

Overall, yes, switching to zigdom was important for multit-context because it allowed us to get rid of the global state that was limiting the creation of multiple instances.

knroy:
Any update/progress on this?

mrdimidium:
@knroy, multiple connection support is available in stable releases, multiple context support is in progress.

PR #2077, opened 2026-04-03
fix: Target.attachToTarget returns unique session id per call

soilSpoon:
## Problem

`Target.attachToTarget` returns the existing `bc.session_id` when one is already set, instead of generating a new unique session ID. The [CDP spec](https://chromedevtools.github.io/devtools-protocol/tot/Target/#method-attachToTarget) requires a fresh, unique `sessionId` per call — a single target can have multiple sessions attached simultaneously.

### How it breaks

When a CDP client (e.g. Playwright) calls `Target.attachToTarget` from an existing page session (`SID-1`), lightpanda returns the same `SID-1`. The client then creates a child session keyed by `SID-1`, overwriting the original page session's entry. All pending callbacks on the original session become orphaned — responses arrive but no callback exists to handle them.

Captured via CDP proxy:
```
C→S  id=32  Target.attachToTarget  sessionId=SID-1
S→C  id=32  result={"sessionId":"SID-1"}   ← same id returned
```

### Root cause

In `doAttachtoTarget`:
```zig
// Before:
const session_id = bc.session_id orelse cmd.cdp.session_id_gen.next();
// → when bc.session_id is already set, next() is never called
```

### Chrome reference

Chrome's [`TargetHandler::AttachToTarget()`](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/devtools/protocol/target_handler.cc) always creates a new session with `base::UnguessableToken::Create()` and stores it in an unbounded `std::map<string, unique_ptr<Session>>`. Multiple sessions per target is explicitly supported.

### 1. Always generate a new session ID (`target.zig`)

```zig
// After:
const session_id = cmd.cdp.session_id_gen.next();  // always new
```

- First attach → stored as `bc.session_id` (primary)
- Subsequent attach → stored as `bc.alt_session_id`, leaving primary intact

### 2. Copy IDs into fixed buffers (`target.zig`)

`session_id_gen` uses a zero-allocation `Incrementing` generator that reuses its internal buffer — the slice returned by `next()` is invalidated by the next call. Both `session_id` and `alt_session_id` are now copied into 14-byte fixed arrays (`session_id_buf` / `alt_session_id_buf`).

### 3. Accept alt session in validation (`CDP.zig`)

`isValidSessionId` now checks both `session_id` and `alt_session_id`.

### 4. Route inspector responses to the requesting session (`CDP.zig`, `runtime.zig`)

V8 inspector has one session per `BrowserContext`. When a command arrives via `alt_session_id`, the response must be tagged with that session, not the primary. `callInspector` stores the requesting session in `inspector_reply_session`; `onInspectorResponse` reads it back. This is safe because `callInspector` → `runMicrotasks` is synchronous.

## Limitations

- This adds a single `alt_session_id` slot (primary + 1 alt). Chrome supports unlimited concurrent sessions via `std::map`. For lightpanda's current single-BrowserContext / single-page architecture, one alt slot is sufficient. A future multi-session refactor would replace this with a session list.
- End-to-end Playwright verification was not possible because `connectOverCDP` depends on other unimplemented features (e.g. `Target.createBrowserContext`). The fix is verified via unit tests and raw CDP client testing.

## Verification

```
417 of 417 tests passed
```

New test: `"cdp.target: attachToTarget returns unique sessionId per call"`
- Two successive `attachToTarget` calls return different session IDs
- Primary `session_id` remains unchanged after second attach
- Events and results carry correct session IDs

### DIFF
--- a/src/cdp/CDP.zig
+++ b/src/cdp/CDP.zig
@@ -263,8 +263,13 @@ fn dispatchCommand(command: *Command, method: []const u8) !void {
 
 fn isValidSessionId(self: *const CDP, input_session_id: []const u8) bool {
     const browser_context = &(self.browser_context orelse return false);
-    const session_id = browser_context.session_id orelse return false;
-    return std.mem.eql(u8, session_id, input_session_id);
+    if (browser_context.session_id) |sid| {
+        if (std.mem.eql(u8, sid, input_session_id)) return true;
+    }
+    if (browser_context.alt_session_id) |alt| {
+        if (std.mem.eql(u8, alt, input_session_id)) return true;
+    }
+    return false;
 }
 
 pub fn createBrowserContext(self: *CDP) ![]const u8 {
@@ -351,7 +356,19 @@ pub const BrowserContext = struct {
     // is all pretty straightforward, but it still needs to be enforced, i.e.
     // if we get a request with a sessionId that doesn't match the current one
     // we should reject it.
-    session_id: ?[]const u8,
+    // session_id_gen.next() reuses its internal buffer, so slices returned
+    // by it are only valid until the next call.  We copy into fixed buffers
+    // so that session IDs remain stable across multiple next() calls.
+    session_id: ?[]const u8 = null,
+    session_id_buf: [14]u8 = undefined,
+
+    // Additional session ID from explicit Target.attachToTarget calls.
+    alt_session_id: ?[]const u8 = null,
+    alt_session_id_buf: [14]u8 = undefined,
+
+    // Tracks which session to reply to for in-flight inspector commands.
+    // Set before callInspector, cleared after runMicrotasks.
+    inspector_reply_session: ?[]const u8 = null,
 
     security_origin: []const u8,
     page_life_cycle_events: bool,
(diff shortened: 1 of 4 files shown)


soilSpoon:
While implementing this fix I noticed that `BrowserContext` currently has only one extra session slot (`alt_session_id`) beyond the primary. This is sufficient for Playwright's `newCDPSession` use case, but won't cover scenarios where `Target.attachToTarget` is called more than twice on the same target.

I saw in #1096 that multi-context support is actively in progress. Could you share where things stand with that work? Specifically:

- Is there a working branch or design doc we could look at?
- Would `alt_session_id` need to be replaced as part of that work (e.g. with a session list)?

Happy to update this PR to avoid conflicts if needed. Thanks!

krichprollsch:
Hey @soilSpoon,

You are right about your analysis. Returning the same session ID every time has worked well for us so far, even though we do not follow the specs.

Multi-context support is being added step by step in `main`. The next step is to share clients into the main thread. There is still a long way to go before multiple contexts are fully supported.

I am not in favor of adding a second alternative session ID. As you said, it only fixes the issue for a 2nd call, but will break if the client asks for more.

Question: did returning the same session ID actually break your use case?

soilSpoon:
> Hey @soilSpoon,
> 
> You are right about your analysis. Returning the same session ID every time has worked well for us so far, even though we do not follow the specs.
> 
> Multi-context support is being added step by step in `main`. The next step is to share clients into the main thread. There is still a long way to go before multiple contexts are fully supported.
> 
> I am not in favor of adding a second alternative session ID. As you said, it only fixes the issue for a 2nd call, but will break if the client asks for more.
> 
> Question: did returning the same session ID actually break your use case?

@krichprollsch

Yes — here's the concrete scenario:

```js
await page.goto('https://example.com');
const session = await context.newCDPSession(page);  // Target.attachToTarget → returns SID-1 (same as page session)
await page.goto('https://example.com/other');       // assert(!object.id) crash
```

Captured via CDP proxy:
```
C→S  Target.attachToTarget  (from SID-1)
S→C  result: { sessionId: "SID-1" }     ← same ID returned

C→S  Page.navigate  id=38  sessionId=SID-1
S→C  id=38  result={...}  → no callback found → assert
```

Playwright's `createChildSession("SID-1")` overwrites `_sessions["SID-1"]` with a new empty object, orphaning the page session's pending callbacks.

---

You're right that `alt_session_id` is a half-fix. Given that multi-context is coming, what approach would you prefer here? We could:

1. Use `ArrayList` for sessions (proper fix, but may conflict with the multi-context redesign)
2. Keep `alt_session_id` as a temporary patch with a TODO pointing to the multi-context work
3. Drop the PR and wait for multi-context if you expect it soon

Happy to go whichever direction makes the most sense for the roadmap.

karlseguin:
I think https://github.com/lightpanda-io/browser/pull/2976 fixed this (with more or less the same approach).

PR #2079, opened 2026-04-03
Layering HTTP Client

mookums:
This allows us to write HTTP Client layers to add functionality without messing with the core HTTPClient too much. This is just an experiment right now so don't merge but would love to hear thoughts.

This ideally should make cache revalidation easier to implement as it would be entirely local to the Cache Layer and transparent to the other layers.


karlseguin:
Something like this is needed. A bit of rambling thoughts...

### layers
We need singleflight protection (both module import and robots implement their own mechanism for this) and request interception handling (and we should probably make sure we handle robots.txt and caches correct with request interception).

### ergonomics
I don't know it this would be implemented as a "layer", but I'd like synchronous request handling to better. Right now ScriptManager handles it, but it should be pushed down to the http client library, since we need it in a few other places:

```zig
const res = try http.requestSync(.{.arena = arena, ...});
// res complete
```

And probably improve the ergonomics for async/callback too so that they can just register 1 `doneCallback` and get the whole response (only XHR (and in theory fetch) care about streaming, and even they just want the whole body collected at the end).

### script caching
The concern with this type of layering is that it makes some things a little harder. I was thinking of how Script caching would work with layers. I think you might end up losing a bit of efficiency (and maybe even consistency), because script caching might need to happen separately from http caching (e.g. in the ScriptManager). Maybe not. But, it did make me realize that the new caching code could itself be split into Storage and HTTP caching.

### script manager
Tha Robots and Cache layers are nice since they cover recent additions, I think going back to the original users of HttClient (ScriptManager, XHR and fetch) and see how they can be improved is worthwhile. That's where I think of things like singleflight and synchronous requests and there are probably other things that I'm missing from their code that would be nice to streamline if possible.

krichprollsch:
I tried to run `go run integrations/main.go` with this branch, and I get a leak reports after stopping the server:
```
error(gpa): memory address 0x742b70605640 leaked:
/home/pierre/wrk/browser/src/network/Network.zig:150:56: 0x372cc12 in _allocBlock (lightpanda)
        const slice = instance.?.allocator.alignedAlloc(u8, .fromByteUnits(alignment), Block.fullsize(size)) catch return null;
                                                       ^
/home/pierre/wrk/browser/src/network/Network.zig:199:35: 0x364c38a in strdup (lightpanda)
        const header = _allocBlock(len + 1) orelse return null;
                                  ^
/home/pierre/wrk/browser/src/sys/libcurl.zig:544:41: 0x358378e in cb (lightpanda)
            return @ptrCast(alloc.strdup(s));
                                        ^
/home/pierre/.cache/zig/p/N-V-__8AALp9QAGn6CCHZ6fK_FfMyGtG824LSHYHHasM3w-y/lib/slist.c:87:19: 0x840fa1f in curl_slist_append (/home/pierre/.cache/zig/p/N-V-__8AALp9QAGn6CCHZ6fK_FfMyGtG824LSHYHHasM3w-y/lib/slist.c)
  char *dupdata = curlx_strdup(data);
                  ^
/home/pierre/wrk/browser/src/sys/libcurl.zig:899:31: 0x2efde6f in curl_slist_append (lightpanda)
    return c.curl_slist_append(list, header);
                              ^
/home/pierre/wrk/browser/src/network/http.zig:97:58: 0x2efe00b in add (lightpanda)
        const updated_headers = libcurl.curl_slist_append(self.headers, header);
                                                         ^
/home/pierre/wrk/browser/src/browser/Page.zig:432:24: 0x2efe29a in headersForRequest (lightpanda)
        try headers.add(referer);
                       ^
/home/pierre/wrk/browser/src/browser/ScriptManager.zig:146:36: 0x2efe5f0 in getHeaders (lightpanda)
    try self.page.headersForRequest(&headers);
                                   ^
/home/pierre/wrk/browser/src/browser/ScriptManager.zig:316:51: 0x2f00b81 in addFromElement__anon_110945 (lightpanda)
                    .headers = try self.getHeaders(),
                                                  ^
/home/pierre/wrk/browser/src/browser/Page.zig:1144:40: 0x2f03111 in scriptAddedCallback__anon_110924 (lightpanda)
    self._script_manager.addFromElement(from_parser, script, "parsing") catch |err| {
```

src/browser/HttpClient.zig
```diff
@@ -44,1065 +44,881 @@ pub const Headers = http.Headers;
...
-
-pub fn init(allocator: Allocator, network: *Network) !*Client {
-    var transfer_pool = std.heap.MemoryPool(Transfer).init(allocator);
-    errdefer transfer_pool.deinit();
-
-    const client = try allocator.create(Client);
-    errdefer allocator.destroy(client);
+pub const PerformStatus = enum { cdp_socket, normal };
+
+pub const Transport = struct {
```
karlseguin, line 49: Don't think this is ever used externally?  (Same with `Context`..I think). Though, I'm not opposed to breaking this out in to _many_ files.

src/browser/HttpClient.zig
```diff
@@ -44,1065 +44,881 @@ pub const Headers = http.Headers;
...
+pub const RobotsLayer = @import("../network/layer/RobotsLayer.zig");
+pub const WebBotAuthLayer = @import("../network/layer/WebBotAuthLayer.zig");
+pub const SingleFlightLayer = @import("../network/layer/SingleFlightLayer.zig");
+pub const CacheLayer = @import("../network/layer/CacheLayer.zig");
 
-    transfer.req.notification.dispatch(.http_request_fail, &.{
-        .transfer = transfer,
-        .err = err,
-    });
+pub const Layers = LayerStack(.{ RobotsLayer, WebBotAuthLayer, SingleFlightLayer, CacheLayer });
```
karlseguin, line 712: Cache should be first, no? Why do all that work if it's cached?

src/browser/HttpClient.zig
```diff
@@ -44,1065 +44,881 @@ pub const Headers = http.Headers;
...
-        cb(transfer.req.ctx);
-    }
-}
+const Client = @This();
 
-// Same restriction as changeProxy. Should be ok since this is only called on
-// BrowserContext deinit.
-pub fn restoreOriginalProxy(self: *Client) !void {
-    try self.ensureNoActiveConnection();
+transport: *Transport,
```
karlseguin, line 716: What's the benefit of having all these previously Clients field moved to Transport (and having a Transport at all)

And, if there is a good reason to split it out, why have it a pointer her? Client owns the Transport? I looked for who else creates a transport and nothing else does. `transport: Transport` would just document this better. (And if the issue is you need a stable pointer *Client is already on the heap, so `&client.transport` is a stable pointer)

src/browser/HttpClient.zig
```diff
@@ -44,1065 +44,881 @@ pub const Headers = http.Headers;
...
-            return self.makeRequest(conn, transfer);
-        }
+    fn ensureNoActiveConnection(self: *const Transport) !void {
+        if (self.active > 0) return error.InflightConnection;
     }
+};
 
-    self.queue.append(&transfer._node);
-}
+pub const Context = struct {
```
karlseguin, line 638: Not sure I see the point of this. We could inject `*Client` into every layer? Or even store *Client in `req`.

src/browser/HttpClient.zig
```diff
@@ -44,1065 +44,881 @@ pub const Headers = http.Headers;
...
+};
 
-// For an intercepted request
-pub fn fulfillTransfer(self: *Client, transfer: *Transfer, status: u16, headers: []const http.Header, body: ?[]const u8) !void {
-    if (comptime IS_DEBUG) {
-        std.debug.assert(transfer._intercept_state != .not_intercepted);
-        log.debug(.http, "filfull transfer", .{ .intercepted = self.intercepted });
-    }
-    self.intercepted -= 1;
+pub fn LayerStack(comptime layer_types: anytype) type {
```
karlseguin, line 659: This is fancy, but we're not using that fanciness. It could all be just hard-coded/inlined in Client.init:

```
// client fields
cache: CacheLayer,
robots: RobotsLayer,
webbotauth: WebBotAuthLayer,
singleflight: SingleFlightLayer,

pub fn init(...) !*Client {
     const client = try allocator.create(Client);
     errdefer allocator.destroy(client);

     client.* = .{
         // .. existing fields ..
         .cache = .{},
         .robots = .{
             .obey_robots = net.config.obeyRobots(),
             .allocator = allocator,
         },
         // .. other layers ..
     };

     client.cache.next = client.robots.layer(); //vtable
     client.robots.next = client.webbotauth.layer();
     // .. other layers ..
     
     return client;
}

pub fn request(self: *Client, req: Request) !void {
  return self.cache.layer().request(req);
}
```

(this also lets you easily remove layers when they aren't needed, e.g. don't wire up the RobotsLayer when obey-robots is false, don't wire the CacheLayer when cache isn't enabled, ... Although, you could probably figure out how to do that with your current approach..maybe).

src/browser/HttpClient.zig
```diff
@@ -44,1065 +44,881 @@ pub const Headers = http.Headers;
...
+                ptrs[i] = ptr;
+                layers[i] = ptr.layer();
+            }
 
-fn makeTransfer(self: *Client, req: Request) !*Transfer {
-    errdefer req.headers.deinit();
+            // Wire inner: each layer's `next` points to the next layer.
+            // Done after all layers are created so pointers are stable.
+            inline for (layer_types, 0..) |T, i| {
+                if (@hasField(T, "next")) {
```
karlseguin, line 680: What if it doesn't?

src/browser/HttpClient.zig
```diff
@@ -44,1065 +44,881 @@ pub const Headers = http.Headers;
...
+        const transfer = try self.makeTransfer(req);
 
-    var it = self.in_use.first;
-    while (it) |node| : (it = node.next) {
-        const conn: *http.Connection = @fieldParentPtr("node", node);
-        try conn.setTlsVerify(verify, self.use_proxy);
-    }
-    self.tls_verify = verify;
-}
+        transfer.req.params.notification.dispatch(.http_request_start, &.{ .transfer = transfer });
```
karlseguin, line 223: Not sent if the request was serviced by the cache layer? I don't think that's correct.

src/network/layer/RobotsLayer.zig
```diff
@@ -0,0 +1,213 @@
...
+            else => {
+                log.debug(.http, "unexpected status on robots", .{
+                    .url = self.robots_url,
+                    .status = self.status,
+                });
+                try network.robot_store.putAbsent(self.robots_url);
+            },
+        }
+
+        try l.next.request(ctx, req);
```
karlseguin, line 187: So if this throws an error, transport will end up calling `errorCallback` which itself calls `l.next.request` (the failing code) again. I think.

Oh, and if that's right, you `defer self.deinit()` at the top, so the errorCallback will be a use-after-free.

src/network/layer/RobotsLayer.zig
```diff
@@ -0,0 +1,213 @@
...
+                }
+            },
+            .absent => {},
+        }
+
+        return self.next.request(ctx, req);
+    }
+
+    // robots not cached yet, fetch it first
+    const robots_ctx = try arena.create(RobotsContext);
```
karlseguin, line 80: I guess the singleflight has been removed from here because of the SingleFlight layer that will dedupe these requests. But we are creating a lot of extra data now and each one will need to parse the robots.txt file and will each try to store it.

karlseguin (commented):
I need to review more of it, I'll go through the individual layers more thoroughly tomorrow.

src/network/layer/CacheLayer.zig
```diff
@@ -0,0 +1,240 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+const std = @import("std");
+const log = @import("../../log.zig");
+
+const http = @import("../http.zig");
+const Transfer = @import("../../browser/HttpClient.zig").Transfer;
```
karlseguin, line 23: Should move `HttpClient` to `src/network/` IMO.

src/network/layer/SingleFlightLayer.zig
```diff
@@ -0,0 +1,206 @@
...
+
+    // only single flight idempotent reqursts.
+    if (!req.params.method.idempotent()) {
+        return self.next.request(ctx, req);
+    }
+
+    const arena = try ctx.network.app.arena_pool.acquire(.small, "SingleFlightLayer");
+    errdefer ctx.network.app.arena_pool.release(arena);
+
+    const key = try arena.dupeZ(u8, req.params.url);
```
karlseguin, line 69: only has to be duped when `found_existing == false`

src/network/layer/CacheLayer.zig
```diff
@@ -0,0 +1,240 @@
...
+
+    var iter = req.params.headers.iterator();
+    const req_header_list = try iter.collect(arena);
+
+    if (network.cache.?.get(arena, .{
+        .url = req.params.url,
+        .timestamp = std.time.timestamp(),
+        .request_headers = req_header_list.items,
+    })) |cached| {
+        defer req.params.headers.deinit();
```
karlseguin, line 66: I think we should add a `req.deinit()` to internalize how this is cleared and so that, if we ever have more to do, we don't have to update 10 different places.

src/network/layer/SingleFlightLayer.zig
```diff
@@ -0,0 +1,206 @@
...
+
+const URL = @import("../../browser/URL.zig");
+const Robots = @import("../Robots.zig");
+const Context = @import("../../browser/HttpClient.zig").Context;
+const Request = @import("../../browser/HttpClient.zig").Request;
+const Response = @import("../../browser/HttpClient.zig").Response;
+const Layer = @import("../../browser/HttpClient.zig").Layer;
+const Forward = @import("Forward.zig");
+
+const SingleFlightLayer = @This();
```
karlseguin, line 30: SingleFlight is necessary for module import and nice to have for robots.txt. I think for everything else, it's more trouble than it's worth. It breaks RI expectations (which is a more general problem) and it seems to have a number of edge cases. Of note, it's purely URL-based, so 2 request with the same URL but (e.g.) different cookies, would get coalesced.  I was hoping we could unify it, but now I think the original approach of Robots and ScriptManager doing their own thing is better.

I'll still add comments for the struct, incase you think it can still add value.

src/network/layer/SingleFlightLayer.zig
```diff
@@ -0,0 +1,206 @@
...
+    forward: Forward,
+
+    pub fn headerCallback(resp: Response) !bool {
+        const self: *FlightContext = @ptrCast(@alignCast(resp.ctx));
+        const flight: *Flight = self.layer.flights.getPtr(self.url).?;
+
+        for (flight.waiters.items) |waiter| {
+            var sub_resp = resp;
+            sub_resp.ctx = waiter.ctx;
+            const proceed = try waiter.header_callback(sub_resp);
```
karlseguin, line 116: What happens to the other waiting waiters?

src/network/layer/SingleFlightLayer.zig
```diff
@@ -0,0 +1,206 @@
...
+
+    const key = try arena.dupeZ(u8, req.params.url);
+    var gop = try self.flights.getOrPut(self.allocator, key);
+
+    // if we already have this flight, just add it to the list.
+    if (gop.found_existing) {
+        log.debug(.browser, "single flight join", .{ .url = key });
+        try gop.value_ptr.append(req);
+        ctx.network.app.arena_pool.release(arena);
+        return;
```
karlseguin, line 77: If there's a http_client.abort(), these aren't `in_use` so never get canceled.

src/network/layer/SingleFlightLayer.zig
```diff
@@ -0,0 +1,206 @@
...
+
+        log.debug(.browser, "single flight data", .{
+            .url = self.url,
+            .count = flight.waiters.items.len,
+        });
+
+        for (flight.waiters.items) |waiter| {
+            var sub_resp = resp;
+            sub_resp.ctx = waiter.ctx;
+            try waiter.data_callback(sub_resp, chunk);
```
karlseguin, line 135: What happens to other waiters on error?

src/network/layer/SingleFlightLayer.zig
```diff
@@ -0,0 +1,206 @@
...
+            self.ctx.network.app.arena_pool.release(arena);
+        }
+
+        log.debug(.browser, "single flight done", .{
+            .url = self.url,
+            .count = flight.waiters.items.len,
+        });
+
+        for (flight.waiters.items) |waiter| {
+            try waiter.done_callback(waiter.ctx);
```
karlseguin, line 157: When does the waiting request get deinit'd ? (I think this is the leak @krichprollsch reported?)

src/network/layer/CacheLayer.zig
```diff
@@ -0,0 +1,240 @@
...
+                            });
+                        }
+                    }
+                }
+
+                const metadata = try allocator.create(CachedMetadata);
+                metadata.* = cm;
+                metadata.headers = header_list.items[0..end_of_response];
+                metadata.vary_headers = header_list.items[end_of_response..];
+                self.pending_metadata = metadata;
```
karlseguin, line 203: So we only cache if there's a vary header? Is that intentional?

src/network/layer/SingleFlightLayer.zig
karlseguin, line 69: Key would need to include `method` at the very least, and probably the same headers that make a request unique from the point of view of caching

src/browser/HttpClient.zig
```diff
@@ -44,1065 +44,881 @@ pub const Headers = http.Headers;
...
+        const self: *SyncContext = @ptrCast(@alignCast(ctx));
+        self.completion = .shutdown;
+    }
+};
 
-    // make sure the transfer can't be immediately aborted from a callback
-    // since we still need it here.
-    transfer._performing = true;
-    defer transfer._performing = false;
+pub fn syncRequest(self: *Client, allocator: std.mem.Allocator, params: RequestParams) !SyncResponse {
```
karlseguin, line 833:
```suggestion
pub fn syncRequest(self: *Client, allocator: Allocator, params: RequestParams) !SyncResponse {
```

src/browser/HttpClient.zig
```diff
@@ -44,1065 +44,881 @@ pub const Headers = http.Headers;
...
+    var sync_ctx = SyncContext{ .allocator = allocator, .body = .empty };
+
+    try self.request(.{
```
karlseguin, line 836: I think it's possible for this call to populate `sync_ctx.body`, so the errdefer needs to moved up before this line.

src/network/layer/Forward.zig
```diff
@@ -0,0 +1,135 @@
...
+    done: ?Request.DoneCallback = null,
+    err: ?Request.ErrorCallback = null,
+    shutdown: ?Request.ShutdownCallback = null,
+};
+
+pub fn wrapRequest(
+    self: *Forward,
+    req: Request,
+    new_ctx: anytype,
+    comptime field: []const u8,
```
karlseguin, line 57: Is this ever not "foward" ?

karlseguin (commented):
I think it's worth removing `SingleFlightLayer`.

I believe there's a lot of footguns around error propagation in the doneCallbacks firing the errorCallback (free after uses, and questionable retries in the error path). I'm not sure what the fix is (or even if there is a fix). The problem might exist in main, not sure. 

Interaction with request interaction needs to be flushed out. Does chrome allow RI for cached items? Can it be a layer, and if so, it should probably the first one (with cache being the second one).

There are no new tests. I'm not sure what can and cannot be truly unit tested, but we do launch an HTTP server in the unit tests, and we do have the demo project for integration test (https://github.com/lightpanda-io/demo/pull/164 would enable setting up more complex cases).

mookums:
> I think it's worth removing `SingleFlightLayer`.

Agreed. Been trying to fight it lately and it seems like whenever an issue is solved that another one pops up. It makes some of the lifetimes really nasty and duplicates a lot of behavior that existed in the `HttpClient` to handle the fanout.

src/browser/HttpClient.zig
mookums: Yeah so then this would require the `InterceptionLayer` right? I can't see any other way you'd handle all of these (including the fact that some notifications would need to now take a `Response` instead of a `Transfer` to ensure we can still intercept on cached responses?)

mookums: I have some local changes trying this out with the `InterceptionLayer` but I'm running into some trouble. What's the best way to handle this because the RI stuff feels so tightly coupled with the `*Transfer` type (in both the HttpClient but also in all of the CDP paths). I'm able to reasonably hoist the interception part but the auth part is giving me challenges.

karlseguin: Do we know how RI and caching interact on Chrome?

mookums: You're able to intercept the cached responses on Chrome I'm pretty sure. The `Response` type has a couple cache fields like `fromDiskCache`, `fromPrefetchCache` and `fromEarlyHints` which are all just signifying that it comes from one of the caches. (https://chromedevtools.github.io/devtools-protocol/tot/Network/#type-Response)

mookums: I believe you are also able to disable the cache from CDP if you want to just get fresh responses with `Network.setCacheDisabled`.

karlseguin: https://github.com/lightpanda-io/demo/pull/165  passes in chrome. So in this case, it doesn't.

src/network/layer/RobotsLayer.zig
```diff
@@ -0,0 +1,269 @@
...
+                .resource_type = .fetch,
+            },
+            .header_callback = RobotsContext.headerCallback,
+            .data_callback = RobotsContext.dataCallback,
+            .done_callback = RobotsContext.doneCallback,
+            .error_callback = RobotsContext.errorCallback,
+            .shutdown_callback = RobotsContext.shutdownCallback,
+        });
+    } else {
+        client.network.app.arena_pool.release(arena);
```
krichprollsch: If I'm not wrong here, the arena is `req.params.arena`.
Is it really the responsibility of a layer to release the request's params arena?
What would happen if another layer use it?

src/network/layer/RobotsLayer.zig
```diff
@@ -0,0 +1,269 @@
...
+}
+
+fn fetchRobotsThenRequest(
+    self: *RobotsLayer,
+    client: *Client,
+    arena: std.mem.Allocator,
+    robots_url: [:0]const u8,
+    req: Request,
+) !void {
+    errdefer client.network.app.arena_pool.release(arena);
```
krichprollsch: same here about the arena

src/cdp/CDP.zig
```diff
@@ -450,8 +450,16 @@ pub const BrowserContext = struct {
...
         // since some of these might callback into the page/scriptmanager
-        for (self.intercept_state.pendingTransfers()) |transfer| {
-            transfer.abort(error.ClientDisconnect);
+        for (self.intercept_state.pendingIntercepts()) |intercept| {
+            switch (intercept) {
+                .transfer => |t| {
+                    t.abort(error.ClientDisconnect);
+                },
+                .request => |r| {
+                    defer r.deinit();
```
krichprollsch: I'm not sure, but don't you let the interception layer in invalid state? Shouldn't you decrement its intercepted counter? What about releasing the request's pool arena also, like in `HttpClient.deinitRequest`?

src/cdp/CDP.zig
```diff
@@ -450,8 +450,16 @@ pub const BrowserContext = struct {
 
         // abort all intercepted requests before closing the session/page
         // since some of these might callback into the page/scriptmanager
-        for (self.intercept_state.pendingTransfers()) |transfer| {
-            transfer.abort(error.ClientDisconnect);
+        for (self.intercept_state.pendingIntercepts()) |intercept| {
+            switch (intercept) {
+                .transfer => |t| {
+                    t.abort(error.ClientDisconnect);
```
krichprollsch, line 465: Same about intercepted counter

src/network/layer/InterceptionLayer.zig
```diff
@@ -0,0 +1,263 @@
...
+pub fn continueRequest(self: *InterceptionLayer, client: *Client, req: Request) anyerror!void {
+    if (comptime IS_DEBUG) {
+        log.debug(.http, "continue transfer", .{ .intercepted = self.intercepted });
+    }
+
+    self.intercepted -= 1;
+    self.next.request(client, req) catch |err| {
+        const ctx: *InterceptContext = @ptrCast(@alignCast(req.ctx));
+        ctx.client.deinitRequest(req);
+        return err;
```
krichprollsch, line 204: I'm not sure I get who is responsible to call `req.err_callback` in case of error.
Should we here?

src/network/layer/InterceptionLayer.zig
```diff
@@ -0,0 +1,263 @@
...
+};
+
+// CDP Callbacks
+
+pub fn continueRequest(self: *InterceptionLayer, client: *Client, req: Request) anyerror!void {
+    if (comptime IS_DEBUG) {
+        log.debug(.http, "continue transfer", .{ .intercepted = self.intercepted });
+    }
+
+    self.intercepted -= 1;
```
krichprollsch, line 199: what about adding a `lp.assert(self.intercepted > 0)` before?

src/network/layer/InterceptionLayer.zig
```diff
@@ -0,0 +1,263 @@
...
+        ctx.client.deinitRequest(req);
+        return err;
+    };
+}
+
+pub fn abortRequest(self: *InterceptionLayer, client: *Client, req: Request) void {
+    if (comptime IS_DEBUG) {
+        log.debug(.http, "abort transfer", .{ .intercepted = self.intercepted });
+    }
+    self.intercepted -= 1;
```
krichprollsch, line 213: what about adding a `lp.assert(self.intercepted > 0)` before?

src/network/layer/InterceptionLayer.zig
```diff
@@ -0,0 +1,263 @@
...
+    req: Request,
+    status: u16,
+    headers: []const http.Header,
+    body: ?[]const u8,
+) !void {
+    if (comptime IS_DEBUG) {
+        log.debug(.http, "fulfill transfer", .{ .intercepted = self.intercepted });
+    }
+
+    self.intercepted -= 1;
```
krichprollsch, line 263: what about adding a `lp.assert(self.intercepted > 0)` before?

src/browser/HttpClient.zig
```diff
@@ -349,345 +390,99 @@ pub fn tick(self: *Client, timeout_ms: u32) !PerformStatus {
...
     status: u16 = 0,
+    body: std.ArrayList(u8),
 
-    pub fn deinit(self: *RobotsRequestContext) void {
-        self.client.allocator.free(self.robots_url);
-        self.buffer.deinit(self.client.allocator);
-        self.client.allocator.destroy(self);
+    fn headerCallback(response: Response) anyerror!bool {
+        const self: *SyncContext = @ptrCast(@alignCast(response.ctx));
+        self.status = response.status().?;
```
krichprollsch, line 429: Worth a `lp.assert` not null  before maybe?

src/browser/HttpClient.zig
krichprollsch: deadcode? (and then `_fulfill` too)

krichprollsch: Claude thinks this is not possible anymore?

> - `Transfer.fulfill` / `Transfer._fulfill` (HttpClient.zig:1474-1520), `ResponseHead._injected_headers`, and the fallback branch of `Transfer.responseHeaderIterator` (1470-1471) have no remaining callers — `InterceptionLayer.fulfillRequest` does its own fulfilling via `FulfilledResponse`. Dead code, worth removing.

src/network/layer/CacheLayer.zig
```diff
@@ -0,0 +1,233 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+const std = @import("std");
+const log = @import("../../log.zig");
```
krichprollsch: Align w/ the rest of the code:
```zig
const log = lp.log;
```

src/network/layer/InterceptionLayer.zig
```diff
@@ -0,0 +1,263 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+const std = @import("std");
+const builtin = @import("builtin");
+const log = @import("../../log.zig");
```
krichprollsch: Align w/ the rest of the code:
```zig
const log = lp.log;
```

src/network/layer/RobotsLayer.zig
```diff
@@ -0,0 +1,269 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+const std = @import("std");
+const log = @import("../../log.zig");
```
krichprollsch: Align w/ the rest of the code:
```zig
const log = lp.log;
```

src/network/layer/WebBotAuthLayer.zig
```diff
@@ -0,0 +1,50 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+const std = @import("std");
+const log = @import("../../log.zig");
```
krichprollsch: Align w/ the rest of the code:
```zig
const log = lp.log;
```

krichprollsch (commented):
Here are 2 more comments from Claude:

> - `r.deinit()` exists publicly on both `Request` and `RequestParams` and only frees headers. Easy to mistake for "free everything" (CDP.zig:459 already did). Consider renaming to `deinitHeaders` or removing the public surface in favor of always going through `Client.deinitRequest`. 

> - `Client.request`'s errdefer only releases the arena — it doesn't `req.deinit()` the headers. Callers compensate (`addFromElement` line 286), but `ScriptManager.preloadImport` (419) and `getAsyncImport` (527), and `webapi/Fetch.zig` / `XMLHttpRequest.zig` / `Worker.zig` / `Frame.zig` do **not** have `errdefer headers.deinit()` — if `client.request` fails, the libcurl slist leaks. (This is preexisting — not introduced by this refactor — but the contract is fragile and worth pinning down. Easiest fix: have `Client.request` deinit headers in its errdefer.)

src/network/layer/InterceptionLayer.zig
mookums: We now pass the errors up the Layer chain. Ideally, you can just return an error and expect the `error_callback` to be called and the `Request` to be de-initalized. On cases where we manage our own `Request` like in `RobotsLayer` OR the CDP callbacks in `InterceptionLayer`, you have to capture and call the callbacks + deinit yourself.

src/network/layer/RobotsLayer.zig
mookums: It now creates a new arena for the subrequest that is used by the RobotsContext. This is cleaned up later by `Transfer.deinit()` or by any deinit of the Request.

krichprollsch (approved):

Issue #2043, opened 2026-03-30
Question: Roadmap for CDP automation features (setFileInputFiles, Input events, dialog, history, window.open)?


noricha-vr:
## Context

Hi! I've been testing Lightpanda v0.2.7 as a headless browser backend for AI agent automation using Puppeteer CDP. Overall very impressed with the progress — DOM extraction, multi-page navigation, fetch API, Cookie/localStorage, and even GitHub/Wikipedia work great.

I ran a comprehensive compatibility test (29 items) and found several CDP/Web API gaps that affect browser automation use cases. Before filing individual feature requests, I wanted to ask: **are any of these on the roadmap, or intentionally out of scope (like visual rendering)?**

### High impact (blocks common automation patterns)

| Feature | Status | Detail |
|---------|--------|--------|
| `DOM.setFileInputFiles` | CDP `UnknownMethod` | Can't simulate file selection via Puppeteer `uploadFile()` |
| Keyboard event propagation | Events don't reach JS listeners | `keyboard.press()` → `keydown` listener not triggered |
| `waitForNavigation` | Times out | Related: #2032 may fix this |

### Medium impact

| Feature | Status | Detail |
|---------|--------|--------|
| Mouse hover events | `mouseenter` not fired | `element.hover()` doesn't trigger listeners |
| `window.open()` | `not a function` | Popup/new tab creation |
| `history.pushState` | Function exists but `pathname` doesn't change | Affects SPA routing |
| `DataTransfer` / `DataTransferItem` | `undefined` | Drag & drop file upload |
| `Emulation.setDeviceMetricsOverride` (viewport) | No effect (fixed 1920x1080) | - |

### Low impact

| Feature | Status | Detail |
|---------|--------|--------|
| `Page.handleJavaScriptDialog` | Dialog events not forwarded | `alert()` / `confirm()` handling |

## Already tracked

I see these are already filed separately — just noting for completeness:
- WebSocket: #1952
- Web Workers: #2017  
- Console interception: #1953
- User-Agent override: #2029 / #2035
- Playwright CDP: #1838 / #1839
- Navigation events: #2032

## Question

Which of the untracked items above (if any) are you planning to support? Happy to file individual issues for anything that's in scope. Understanding the boundary between "not yet implemented" and "intentionally out of scope" would help us plan accordingly.

Thanks for the amazing project!

krichprollsch:
Hey @noricha-vr,

[...]

PR #2084, opened 2026-04-03
fix: update page URL and location on pushState/replaceState

### LINKED ISSUE #2081: history.pushState does not update location.pathname
Per the CDP automation compatibility audit in #2043 (by @noricha-vr), calling `history.pushState()` completes without error but `location.pathname` still reflects the old value. This breaks SPA routing detection in automation scripts that check pathname after navigation.

The `pushState` function exists in the WebAPI layer but the URL state doesn't appear to propagate to the location object.

Ref: #2043 (history.pushState row)

tmchow:
`history.pushState()` and `replaceState()` added entries to the navigation history but didn't update `page.url` or reinitialize `window._location`. This meant `location.pathname` still returned the old path after pushState, breaking SPA routing detection in automation scripts.

The fix adds two lines to both `pushState` and `replaceState` in `History.zig`:
1. `page.url = url` to update the page's active URL
2. `page.window._location = try Location.init(url, page)` to reinitialize the Location object

This follows the same pattern used during actual navigation (Page.zig lines 477, 560, 895).

`replaceState` gets the same fix since it had the same gap.

Fixes #2081
Ref #2043

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)

### DIFF
--- a/src/browser/webapi/History.zig
+++ b/src/browser/webapi/History.zig
@@ -20,6 +20,7 @@ const std = @import("std");
 const js = @import("../js/js.zig");
 
 const Page = @import("../Page.zig");
+const Location = @import("Location.zig");
 const PopStateEvent = @import("event/PopStateEvent.zig");
 
 const History = @This();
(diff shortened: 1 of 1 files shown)


mookums:
Thank you for the PR! Adapted to https://github.com/lightpanda-io/browser/pull/2131 with some modifications.

PR #2085, opened 2026-04-03
feat: emit Page.javascriptDialogOpening CDP events for JS dialogs

### LINKED ISSUE #2082: Page.handleJavaScriptDialog CDP method not implemented
Per the CDP automation compatibility audit in #2043 by @noricha-vr, JavaScript dialog events (alert, confirm, prompt) are not forwarded via CDP, and `Page.handleJavaScriptDialog` is not implemented. This means Puppeteer's `page.on('dialog')` never fires and dialogs can't be accepted or dismissed programmatically.

The Dialog HTML element already exists in the WebAPI layer. Wiring a CDP handler for `Page.handleJavaScriptDialog` and emitting `Page.javascriptDialogOpening` events would close this gap.

Ref: #2043 (Page.handleJavaScriptDialog row)

tmchow:
`window.alert()`, `confirm()`, and `prompt()` previously ran as silent stubs with no CDP visibility. Puppeteer's `page.on('dialog')` never fired because no `Page.javascriptDialogOpening` event was emitted.

This wires the dialog functions through the Notification system to the CDP layer:

1. **Window.zig**: `alert`, `confirm`, and `prompt` now dispatch a `javascript_dialog_opening` notification with the page URL, message text, and dialog type. Return values are unchanged (alert is void, confirm returns false, prompt returns null) since dialogs auto-dismiss in headless mode.

2. **Notification.zig**: New `JavascriptDialogOpening` event type with `url`, `message`, and `dialog_type` fields.

3. **CDP.zig**: Registers a listener for `javascript_dialog_opening` and forwards to the page domain handler.

4. **page.zig**: Emits `Page.javascriptDialogOpening` CDP event with `url`, `message`, `type`, `hasBrowserHandler`, and `defaultPrompt` fields per the [CDP spec](https://chromedevtools.github.io/devtools-protocol/tot/Page/#event-javascriptDialogOpening). Also adds `Page.handleJavaScriptDialog` as a method that accepts `accept` and optional `promptText` params.

The notification dispatch follows the same pattern used by `page_dom_content_loaded`, `page_loaded`, etc. (`Page -> Notification -> CDP.BrowserContext -> page.zig -> sendEvent`).

Dialogs still auto-dismiss rather than blocking JS execution. Supporting interactive accept/dismiss (where `confirm()` returns the CDP client's choice) would require suspending the JS runtime, which is a larger change that could follow separately.

All 422 existing tests pass.

Fixes #2082
Ref #2043

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)

### DIFF
--- a/src/cdp/domains/page.zig
+++ b/src/cdp/domains/page.zig
@@ -646,14 +646,14 @@ fn sendPageLifecycle(bc: *CDP.BrowserContext, name: []const u8, timestamp: u64, 
 
 // https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-handleJavaScriptDialog
 fn handleJavaScriptDialog(cmd: *CDP.Command) !void {
-    // Dialogs auto-dismiss in headless mode, so this is an acknowledgement.
-    // accept and promptText params are parsed but not used since the dialog
-    // already returned by the time the CDP client sends this.
+    // Dialogs auto-dismiss in headless mode. By the time the CDP client
+    // sends this command, the dialog has already returned and there is
+    // no pending dialog to accept or dismiss.
     _ = try cmd.params(struct {
         accept: bool,
         promptText: ?[]const u8 = null,
     });
-    try cmd.sendResult(null, .{});
+    return cmd.sendError(-32000, "No dialog is showing", .{});
 }
 
 // https://chromedevtools.github.io/devtools-protocol/tot/Page/#event-javascriptDialogOpening


karlseguin:
@krichprollsch what do you think about this? The point is that we emit a CDP message on dialogs (e.g. alert). The CDP client is then supposed to be able to accept/reject the the dialog (i.e. by responding to the notification with its own message). This PR adds the browser -> CDP notification, but no-ops the CDP -> Browser accept/reject. 

If I was to implement this, i'd probably do the same thing, but does it do more harm than good? We'd be signaling the dialog to the CDP client, and would give the illusion that it can be cancelled, but we'd already have executed right past it...

krichprollsch:
What about returning an explicit error to the client trying to act on the dialog?

Idk if it's useful, I agree I don't want to stop js execution except if we are e relevant use case.

@tmchow you have a case where you need the event but you don't care about the CDP client's response?

tmchow:
> What about returning an explicit error to the client trying to act on the dialog?
> 
> Idk if it's useful, I agree I don't want to stop js execution except if we are e relevant use case.
> 
> @tmchow you have a case where you need the event but you don't care about the CDP client's response?

@krichprollsch Yes, that's the use case. Browser automation frameworks like Playwright detect `alert()` / `confirm()` / `prompt()` via the `Page.javascriptDialogOpening` CDP event. Without it, the dialog is invisible to the automation client. The event alone is enough for the common case: detecting that a dialog fired, logging its message, and continuing.

Good call on the explicit error. Updated `handleJavaScriptDialog` to return `-32000 "No dialog is showing"` instead of the silent success. That way clients that try to accept/dismiss get a clear signal that there's nothing to act on.

karlseguin:
Thank you.

PR #2086, opened 2026-04-05
Move the NodeList created from a ChildNode to the ChildNode's arena

karlseguin:


src/browser/webapi/collections/ChildNodes.zig
```diff
@@ -126,8 +126,12 @@ fn versionCheck(self: *ChildNodes, page: *Page) bool {
 }
 
 const NodeList = @import("NodeList.zig");
-pub fn runtimeGenericWrap(self: *ChildNodes, page: *Page) !*NodeList {
-    return page._factory.create(NodeList{ ._data = .{ .child_nodes = self } });
```
krichprollsch, line 130: No interest to keep the factory?

karlseguin: Not when an well-scoped arena is already available. With the arena, when the ChildNode is freed, so too will any NodeList that was created.

PR #2098, opened 2026-04-07
Comptime CLI builder and parser

nikneym:
This PR adds a comptime CLI builder that generates a tagged union parser from a
declarative command recipe. Each command becomes a union variant whose
payload is a struct with one field per option.

## Command descriptor fields

  - `name: []const u8` — canonical command name on the command line.
  - `options: tuple` — tuple of option descriptors (see below). Use `.{}` for none.
  - `aliases: tuple` (optional) — alternative names for the command.
  - `shared_options: tuple` (optional) — extra options merged into this
    command. Useful for common flags shared across commands.
  - `positional: struct` (optional) — a single positional argument with
    `.name` and `.type`. Type must be an optional pointer-to-u8 slice
    (e.g. `?[:0]const u8`). Positionals can appear anywhere in argv.

## Option descriptor fields

  - `name: []const u8` — snake_case field name. Both `--snake_case` and
    `--kebab-case` are accepted on the command line.
  - `type` — the Zig type of the parsed value (see supported types below).
  - `default` (optional) — compile-time default when the flag is absent.
    Rules vary by type; see the defaults section below.
  - `shortcuts: tuple` (optional) — single-character short flags. Each
    shortcut is matched as `-X` on the command line.
  - `multiple: bool` (optional) — when `true`, the field becomes a
    `std.ArrayList(type)` and each occurrence appends.
  - `validator: fn` (optional) — custom parse function that replaces the
    built-in type switch. See the validator section below.

## Supported types and their defaults

  - `bool` — presence sets `true`; always defaults to `false`.
    Specifying `default` is a compile error. `?bool` is not allowed.
  - Integers (`u8`, `u16`, `u31`, `usize`, etc.) — parsed with
    `std.fmt.parseInt`. Requires `default` unless wrapped in `?`.
  - `[]const u8`, `[:0]const u8` (and mutable variants) — string slices,
    duped from argv. Sentinel is preserved. Requires `default` unless `?`.
  - Enums — parsed via `std.meta.stringToEnum`. Requires `default` unless `?`.
  - Packed structs of `bool` fields — parsed from a comma-separated list
    (e.g. `--strip js,css`). The literal `"all"` sets every field.
    Requires `default`.
  - Optional types default to `null` when `default` is omitted.

## Validators

A `validator` is a custom parse function that takes over argument
consumption for an option. The expected signature depends on whether
`multiple` is set:

  - Single: `fn (Allocator, *ArgIterator) !T` — returns the parsed value.
  - Multiple: `fn (Allocator, *ArgIterator, *std.ArrayList(T)) !void` —
    appends directly into the list.

When a validator is present, the built-in type switch is skipped entirely.

## Example

```zig
const StripMode = packed struct(u2) {
    js: bool = false,
    css: bool = false,
};

const WaitUntil = enum { load, domcontentloaded, networkidle };

const CommonOptions = .{
    .{ .name = "verbose", .type = bool },
    .{ .name = "log_level", .type = ?log.Level },
    .{ .name = "timeout", .type = u31, .default = 30 },
};

const Cli = cli.Builder(.{
    .{
        .name = "serve",
        .aliases = .{"s"},
        .options = .{
            .{ .name = "host", .shortcuts = .{"h"}, .type = []const u8, .default = "127.0.0.1" },
            .{ .name = "port", .shortcuts = .{"p"}, .type = u16, .default = 9222 },
        },
        .shared_options = CommonOptions,
    },
    .{
        .name = "fetch",
        .positional = .{ .name = "url", .type = ?[:0]const u8 },
        .options = .{
            .{ .name = "dump", .type = ?DumpFormat, .validator = dumpValidator },
            .{ .name = "strip", .type = StripMode, .default = .{} },
            .{ .name = "wait_until", .type = ?WaitUntil },
            .{ .name = "extra_header", .type = []const u8, .multiple = true },
        },
        .shared_options = CommonOptions,
    },
    .{ .name = "version", .options = .{} },
    .{ .name = "help", .aliases = .{ "h", "?" }, .options = .{} },
});

const _, const cmd = try Cli.parse(arena);
switch (cmd) {
    .serve => |opts| listen(opts.host, opts.port),
    .fetch => |opts| fetch(opts.url.?, opts.dump),
    .version => printVersion(),
    .help => printHelp(),
}
```

### DIFF
--- a/src/Config.zig
+++ b/src/Config.zig
@@ -39,25 +39,41 @@ const Config = @This();
 
 /// Common CLI args.
 const CommonOptions = .{
-    .{ .name = "obey_robots", .type = bool, .default = false },
-    .{ .name = "proxy_bearer_token", .type = ?[:0]const u8, .default = null },
-    .{ .name = "http_proxy", .type = ?[:0]const u8, .default = null },
-    .{ .name = "http_max_concurrent", .type = ?u8, .default = null },
-    .{ .name = "http_max_host_open", .type = ?u8, .default = null },
-    .{ .name = "http_timeout", .type = ?u31, .default = null },
-    .{ .name = "http_connect_timeout", .type = ?u31, .default = null },
-    .{ .name = "http_max_response_size", .type = ?usize, .default = null },
-    .{ .name = "ws_max_concurrent", .type = ?u8, .default = null },
-    .{ .name = "insecure_disable_tls_host_verification", .type = bool, .default = false },
-    .{ .name = "log_level", .type = ?log.Level, .default = null },
-    .{ .name = "log_format", .type = ?log.Format, .default = null },
+    .{ .name = "obey_robots", .type = bool },
+    .{ .name = "proxy_bearer_token", .type = ?[:0]const u8 },
+    .{ .name = "http_proxy", .type = ?[:0]const u8 },
+    .{ .name = "http_max_concurrent", .type = ?u8 },
+    .{ .name = "http_max_host_open", .type = ?u8 },
+    .{ .name = "http_timeout", .type = ?u31 },
+    .{ .name = "http_connect_timeout", .type = ?u31 },
+    .{ .name = "http_max_response_size", .type = ?usize },
+    .{ .name = "ws_max_concurrent", .type = ?u8 },
+    .{ .name = "insecure_disable_tls_host_verification", .type = bool },
+    .{ .name = "log_level", .type = ?log.Level },
+    .{ .name = "log_format", .type = ?log.Format },
     // TODO: log_filter_scopes (?[]log.Scope) — parser doesn't yet support `multiple` for enums.
-    .{ .name = "user_agent_suffix", .type = ?[]const u8, .default = null },
-    .{ .name = "http_cache_dir", .type = ?[]const u8, .default = null },
-    .{ .name = "web_bot_auth_key_file", .type = ?[]const u8, .default = null },
-    .{ .name = "web_bot_auth_keyid", .type = ?[]const u8, .default = null },
-    .{ .name = "web_bot_auth_domain", .type = ?[]const u8, .default = null },
+    .{ .name = "user_agent_suffix", .type = ?[]const u8 },
+    .{ .name = "http_cache_dir", .type = ?[]const u8 },
+    .{ .name = "web_bot_auth_key_file", .type = ?[]const u8 },
+    .{ .name = "web_bot_auth_keyid", .type = ?[]const u8 },
+    .{ .name = "web_bot_auth_domain", .type = ?[]const u8 },
 };
+
+fn dumpValidator(_: Allocator, args: *std.process.ArgIterator) !?DumpFormat {
+    // Peek next argument.
+    var peek_args = args.*;
+    if (peek_args.next()) |next_arg| {
+        // Skip the argument we peek if successful.
+        defer _ = args.next();
+        return std.meta.stringToEnum(DumpFormat, next_arg) orelse {
+            return error.UnknownDumpOption;
+        };
+    }
+
+    // Means we couldn't get something like `--dump html` but we do have
+    // `--dump`; which should fall to `html` by default.
+    return .html;
+}
 
 /// Definition for all the commands and its arguments. See @cli.zig for further.
 const Commands = cli.Builder(.{
@@ -66,7 +82,7 @@ const Commands = cli.Builder(.{
         .options = .{
             .{ .name = "host", .type = []const u8, .default = "127.0.0.1" },
             .{ .name = "port", .type = u16, .default = 9222 },
-            .{ .name = "advertise_host", .type = ?[]const u8, .default = null },
+            .{ .name = "advertise_host", .type = ?[]const u8 },
             .{ .name = "timeout", .type = u31, .default = 10 },
             .{ .name = "cdp_max_connections", .type = u16, .default = 16 },
             .{ .name = "cdp_max_pending_connections", .type = u16, .default = 128 },
@@ -76,24 +92,30 @@ const Commands = cli.Builder(.{
     .{
         .name = "fetch",
         .aliases = .{ "f", "get" },
-        // Fetch only; this argument can be given anywhere.
+        // This argument can be given out of order.
         .positional = .{ .name = "url", .type = ?[:0]const u8 },
         .options = .{
-            .{ .name = "dump", .type = ?DumpFormat, .default = null },
-            .{ .name = "with_base", .type = bool, .default = false },
-            .{ .name = "with_frames", .type = bool, .default = false },
+            .{
+                .name = "dump",
+                // Prefixed by `-` (single dash).
+                .shortcuts = .{"d"},
+                .type = ?DumpFormat,
+                .validator = dumpValidator,
+            },
+            .{ .name = "with_base", .type = bool },
+            .{ .name = "with_frames", .type = bool },
             .{ .name = "strip", .type = dump.Opts.Strip, .default = dump.Opts.Strip{} },
             .{ .name = "wait_ms", .type = u32, .default = 5_000 },
-            .{ .name = "wait_until", .type = ?WaitUntil, .default = null },
-            .{ .name = "wait_script", .type = ?[:0]const u8, .default = null },
-            .{ .name = "wait_selector", .type = ?[:0]const u8, .default = null },
+            .{ .name = "wait_until", .type = ?WaitUntil },
+            .{ .name = "wait_script", .type = ?[:0]const u8 },
+            .{ .name = "wait_selector", .type = ?[:0]const u8 },
         },
         .shared_options = CommonOptions,
     },
     .{
         .name = "mcp",
         .options = .{
-            .{ .name = "cdp_port", .type = ?u16, .default = null },
+            .{ .name = "cdp_port", .type = ?u16 },
         },
         .shared_options = CommonOptions,
     },
--- a/src/Config.zig
+++ b/src/Config.zig
@@ -653,3 +653,15 @@ pub fn parseArgs(allocator: Allocator) !Config {
     const exec_name, const command = try Commands.parse(allocator);
     return .init(allocator, exec_name, command);
 }
+
+pub fn validateUserAgent(ua: []const u8) !void {
+    for (ua) |c| {
+        if (!std.ascii.isPrint(c)) {
+            return error.NonPrintable;
+        }
+    }
+
+    if (std.ascii.indexOfIgnoreCase(ua, "mozilla") != null) {
+        return error.Reserved;
+    }
+}
--- a/src/Config.zig
+++ b/src/Config.zig
@@ -74,6 +74,8 @@ const CommonOptions = .{
     .{ .name = "user_agent", .type = ?[]const u8 },
     .{ .name = "block_private_networks", .type = bool },
     .{ .name = "block_cidrs", .type = ?[]const u8 },
+    .{ .name = "cookie", .type = ?[]const u8 },
+    .{ .name = "cookie_jar", .type = ?[]const u8 },
 };
 
 fn dumpValidator(_: Allocator, args: *std.process.ArgIterator) !?DumpFormat {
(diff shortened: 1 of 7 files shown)


nikneym:
https://github.com/lightpanda-io/browser/actions/runs/24248843738/job/70804474493?pr=2098

Command sniffing is missing; will make a commit reintroducing that.

src/Config.zig
```diff
@@ -242,28 +337,20 @@ pub fn maxPendingConnections(self: *const Config) u31 {
...
 }
 
-pub const Mode = union(RunMode) {
-    help: bool, // false when being printed because of an error
-    fetch: Fetch,
-    serve: Serve,
-    version: void,
-    mcp: Mcp,
-};
+//pub const Serve = struct {
```
karlseguin, line 340: Any reason to keep this as a comment?

src/Config.zig
```diff
@@ -21,32 +21,127 @@ const builtin = @import("builtin");
...
+        },
+        .shared_options = CommonOptions,
+    },
+    .{
+        .name = "fetch",
+        .aliases = .{ "f", "get" },
+        // This argument can be given out of order.
+        .positional = .{ .name = "url", .type = ?[:0]const u8 },
+        .options = .{
+            .{ .name = "dump", .shortcuts = .{ "d", "D" }, .type = ?DumpFormat, .validator = dumpValidator },
```
karlseguin, line 116: how do you decide what does and doesn't get a shortcut?

src/cli.zig
```diff
@@ -0,0 +1,575 @@
...
+                }
+            }
+
+            unreachable;
+        }
+
+        /// Try to sniff the command out of given option.
+        /// Only exists for legacy reasons; hence hardcoded.
+        fn sniffCommand(cmd_str: []const u8) error{UnknownCommand}!Enum {
+            if (std.mem.startsWith(u8, cmd_str, "--") == false) {
```
karlseguin, line 304: This check isn't compatible with the single dash options....e.g. a few lines down you check for `-P`, but that will never be possible.

src/Config.zig
```diff
@@ -21,32 +21,127 @@ const builtin = @import("builtin");
...
+    // `--dump`; which should fall to `html` by default.
+    return .html;
+}
+
+/// Definition for all the commands and its arguments. See @cli.zig for further.
+const Commands = cli.Builder(.{
+    .{
+        .name = "serve",
+        .options = .{
+            .{ .name = "host", .shortcuts = .{ "h", "H" }, .type = []const u8, .default = "127.0.0.1" },
```
karlseguin: I personally don't love having multiple ways to set the same option.  (and I _think_ case insensitive single-char options is uncommon?)

src/Config.zig
```diff
@@ -482,12 +530,12 @@ pub fn printUsageAndExit(self: *const Config, success: bool) void {
         \\                Argument must be 'html', 'markdown', 'semantic_tree', or 'semantic_tree_text'.
         \\                Defaults to no dump.
         \\
-        \\--strip-mode    Comma separated list of tag groups to remove from dump
-        \\                the dump. e.g. --strip-mode js,css
+        \\--strip         Comma separated list of tag groups to remove from dump
```
karlseguin, line 533: We ok with breaking backwards compatibility? (both the name and changing all -> full) ?

src/cli.zig
```diff
@@ -0,0 +1,575 @@
...
+/// const CommonOptions = .{
+///     .{ .name = "verbose", .type = bool },
+///     .{ .name = "log_level", .type = ?log.Level },
+///     .{ .name = "timeout", .type = u31, .default = 30 },
+/// };
+///
+/// const Cli = cli.Builder(.{
+///     .{
+///         .name = "serve",
+///         .aliases = .{"s"},
```
karlseguin, line 96: Like the argument "shortcuts", don't like having multiple ways to do this.

src/cli.zig
```diff
@@ -0,0 +1,575 @@
...
+                            // If type is an optional type without default value, prefer null.
+                            if (!has_default) {
+                                break :blk @as(*const anyopaque, @ptrCast(&@as(T, null)));
+                            }
+                            // We have default value for an optional.
+                            break :blk @as(*const anyopaque, @ptrCast(&@as(T, option.default)));
+                        },
+                        .bool => {
+                            if (has_default) {
+                                @compileError("booleans are always `false` by default");
```
karlseguin, line 184: Why? that means we can't make --obey-robots default to true in the future?

src/cli.zig
```diff
@@ -0,0 +1,575 @@
...
+                            std.mem.replaceScalar(u8, &output, '_', '-');
+                            break :casing output;
+                        };
+
+                        const match =
+                            std.mem.eql(u8, option_name, "--" ++ option.name) or
+                            std.mem.eql(u8, option_name, "--" ++ kebab_cased);
+
+                        // Name not matched; try shortcuts if provided.
+                        if (!match) {
```
karlseguin, line 374:
```zig
if (match) {
  break :blk true;
}
```

Look at how it's written now

```
match = X

{
  block of code that doesn't care about match
}

return match;
```


Vs:

```
if (match) {
  return match;
}

{
  block of code that doesn't care about match
}
```

In the revised version, the 640KB of memory my brain is working on doesn't have to hold onto `match` across a scope that doesn't use it.

(of course, we could just remove shortcuts).

src/Config.zig
```diff
@@ -21,32 +21,127 @@ const builtin = @import("builtin");
...
 pub const CDP_MAX_HTTP_REQUEST_SIZE = 4096;
 
 // max message size
 // +14 for max websocket payload overhead
 // +140 for the max control packet that might be interleaved in a message
 pub const CDP_MAX_MESSAGE_SIZE = 512 * 1024 + 14 + 140;
 
+const Config = @This();
+
+fn logFilterScopesValidator(allocator: Allocator, args: *std.process.ArgIterator, list: *std.ArrayList(log.Scope)) !void {
```
karlseguin, line 39: Is there something special here that requires validation? Given the builder a type of `[]LogScope`, can't it figure out what is and isn't valid on its own?

src/cli.zig
```diff
@@ -0,0 +1,575 @@
...
+                                        // DupeZ branch.
+                                        if (comptime pointer.sentinel()) |sentinel| {
+                                            const buf = try allocator.alignedAlloc(u8, .fromByteUnits(pointer.alignment), str.len + 1);
+                                            @memcpy(buf[0..str.len], str);
+                                            buf[str.len] = sentinel;
+                                            break :blk buf[0..str.len :sentinel];
+                                        }
+
+                                        // Dupe branch.
+                                        const buf = try allocator.alignedAlloc(u8, .fromByteUnits(pointer.alignment), str.len);
```
karlseguin, line 441: We can't use dupe here?

src/cli.zig
```diff
@@ -0,0 +1,575 @@
...
+                                    if (not_u8_slice) {
+                                        @compileError("Only []u8, []const u8, [:sentinel]u8 and [:sentinel]const u8 pointers are supported");
+                                    }
+
+                                    const v = blk: {
+                                        const str = args.next() orelse return error.MissingArgument;
+
+                                        // DupeZ branch.
+                                        if (comptime pointer.sentinel()) |sentinel| {
+                                            const buf = try allocator.alignedAlloc(u8, .fromByteUnits(pointer.alignment), str.len + 1);
```
karlseguin, line 434: we can't use dupeZ here?

src/cli.zig
```diff
@@ -0,0 +1,575 @@
...
+
+                                    const str = args.next() orelse return error.MissingArgument;
+
+                                    if (std.mem.eql(u8, str, "all")) {
+                                        // "all" sets all the fields of packed struct.
+                                        const Int = _struct.backing_integer.?;
+                                        @field(c, option.name) = @bitCast(@as(Int, std.math.maxInt(Int)));
+                                    } else {
+                                        // Parse given args.
+                                        var it = std.mem.splitScalar(u8, str, ',');
```
karlseguin, line 471: `tokenizeScalar` might be better here

karlseguin (commented):
--
`./lightpanda --help`

Current:  outputs the help (sure, it was more accidental than intentional, but I think it was pretty useful)

Branch: output is unhelpful
FATAL app : exit . . . . . . . . . . . . . . . . . . . . . .  [+0ms]
      err = UnknownCommand

--
` ./lightpanda serve serve`

Current outputs the help (which might not be very helpful), but also prints:
FATAL app : unknown argument . . . . . . . . . . . . . . . .  [+0ms]
      mode = serve
      arg = serve

Branch: output is unhelpful:
FATAL app : exit . . . . . . . . . . . . . . . . . . . . . .  [+0ms]
      err = UnknownOption

--
./lightpanda fetch --strip-mode spice http://127.0.0.1:1234/

Current: logs that `spice` is invalid
Branch: silently ignores invalid value (for current, use `--strip` since `strip-mode` was renamed to `strip`)

karlseguin:
The code is more or less the same number of lines, and it's a lot more complicated. It does add some features, but I think aliases and shortcut both hurt usability.

It does introduce a few breaking changes. I'm fine with that, but dunno if we should keep supporting the existing stuff during a transition with a deprecation warning?

I am curious why every type, except a boolean, is optional. _every_ option is optional and the system doesn't seem to enforce non-optional options anyways (except for the positional, which in theory could be optional, but in our case isn't). So there appears to be a mix of making this generic and making this specific to our use-case. I guess, my question is: why make options optional? Seems simpler to remove that.

src/Config.zig
nikneym: Agreed. I needed a generic name that'll be used to include everything so went with `all`. For `strip-mode` to `strip` I don't have any execuses though 🙈; I think its better to think about those once we get this PR merged, so I reverted both.

src/Config.zig
nikneym: Yes, `cli.Builder` assumes multiple options are provided in such fashion:
```
--header "My-Token: ..." --header "User-Agent: Lightpanda"
```

Though this option was implemented to receive multiple values like so:
```
--log-filter-scopes="http,event"
```

src/cli.zig
nikneym: We can but I wanted it to be consistent with alignment of the given type; so went manual.

src/cli.zig
nikneym: We can but I wanted it to be consistent with alignment and sentinel of the given type; so went manual.

nikneym:
> I am curious why every type, except a boolean, is optional. every option is optional and the system doesn't seem to enforce non-optional options anyways (except for the positional, which in theory could be optional, but in our case isn't). So there appears to be a mix of making this generic and making this specific to our use-case. I guess, my question is: why make options optional? Seems simpler to remove that.

I'm not sure what approach should be taken on this, we can either accept such:
```zig
.{ .name = "advertise_host", .type = []const u8 }
```

And create a nullable (`?[]const u8`) on the other side though that may look confusing. Or, we can prefer `[]const u8` directly and either set it to `undefined` or zero-init though I think that'd be horrendous. I feel first one is more your taste, but as I said, I couldn't fully understand it.

src/cli.zig
nikneym: I've changed the behavior of this; it accepts `default` value and setting the option flips the `default`.

One thing we may also add is accepting `--no-*` prefix to disable the option.

src/cli.zig
karlseguin: So dupe isn't consistent? We use dupe a lot in the codebase, should we switch? Can you help me understand when one is preferred / correct vs the other?

src/Config.zig
karlseguin: So it naturally works with --log-filter-scopes dom  --log-filter-scopes mcp    but this validation lets it work with a comma-separated list? Nice.

karlseguin:
Similar examples as above between `master` and `branch`


### 1
```
master fetch --strip-mode spice http://127.0.0.1:1234/
FATAL app : invalid option choice . . . . . . . . . . . . . . [+0ms]
      arg = --strip-mode
      value = spice
```
vs

```
branch fetch --strip-mode spice http://127.0.0.1:1234/
FATAL app : exit . . . . . . . . . . . . . . . . . . . . . .  [+0ms]
      err = UnknownArgument

```

### 2
```
master fetch --asd --port 9223
FATAL app : unknown argument . . . . . . . . . . . . . . . .  [+0ms]
      mode = fetch
      arg = --asd
```
vs
```
branch fetch --asd --port 9223
FATAL app : exit . . . . . . . . . . . . . . . . . . . . . .  [+0ms]
      err = TooManyPositionalArguments
```

### 3
```
master serve --port 33 --x 3
FATAL app : unknown argument . . . . . . . . . . . . . . . .  [+0ms]
      mode = serve
      arg = --x
```
vs
```
branch serve --port 33 --x 3
FATAL app : exit . . . . . . . . . . . . . . . . . . . . . .  [+0ms]
err = UnknownOption
```
But if we can improve this (and double check those non-dupe dupes), it'll be good to go.

src/cli.zig
nikneym: `dupe` assumes the slice is natural alignment always; this should only be preferred if slice is over-aligned (something like `[]align(8) const u8`), I agree that such scenario is niche (especially for CLI strings). We can give a `@compileError` if slice is over-aligned if we go with `dupe`.

karlseguin (approved):

PR #2100, opened 2026-04-07
Allow user agent override with restrictions

### LINKED ISSUE #2029: Modify User Agent

krichprollsch:
Add `--user-agent` option to override the user-agent.

Rules:
* we don't want to allow browser impersonation, so any user-agent ~starting with `Mozilla/5.0`~ containing `mozilla` is rejected.
* `--user-agent` and `--user-agent-suffix` are exclusive
* The browser will always send the [`Sec-Ch-Ua` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Sec-CH-UA) with `"Lightpanda";v="1"` hint.

Follow-up for #2035 
Fix #2029

### DIFF
--- a/src/Config.zig
+++ b/src/Config.zig
@@ -413,7 +413,7 @@ pub fn printUsageAndExit(self: *const Config, success: bool) void {
         \\
         \\--user-agent    Override the User-Agent header entirely
         \\                User-Agent mustn't impersonate other browser.
-        \\                Starting the value with "Mozilla/5.0" is forbidden.
+        \\                Any value containing "Mozilla" is forbidden.
         \\                The browser will continue to send Sec-Ch-Ua header.
         \\                Incompatible with --user-agent-suffix
         \\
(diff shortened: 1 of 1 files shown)


src/Config.zig
```diff
@@ -1043,11 +1060,45 @@ fn parseCommonArg(
...
+        };
+
+        for (str) |c| {
+            if (!std.ascii.isPrint(c)) {
+                log.fatal(.app, "not printable character", .{ .arg = opt });
+                return error.InvalidArgument;
+            }
+        }
+
+        if (std.mem.startsWith(u8, str, "Mozilla/5.0")) {
```
karlseguin: What about expanding this? 
1 - case insensitive
2 - anywhere in the string, not just the start
3 - just "mozilla"

It isn't impossible some scripts out there are doing something like `if (user_agent.includes("mozilla")) {`

PR #2122, opened 2026-04-10
feat: enrich CDP /json/version and add /json/list endpoint

mvanhorn:
## Summary

Enriches the `/json/version` CDP endpoint with `Browser`, `Protocol-Version`, and `User-Agent` fields. Adds `/json/list` (and `/json`) endpoint returning an empty JSON array.

## Why this matters

Every CDP-compatible browser returns rich metadata from `/json/version`. Chrome returns 7+ fields. Lightpanda currently returns only `webSocketDebuggerUrl`. Automation tools use these fields for capability detection:

- Puppeteer reads `Browser` and `Protocol-Version`
- Chromedp reads `webSocketDebuggerUrl` (already works) and checks response structure
- Stagehand checks protocol version for feature support (#1962 may be related)

The `/json/list` endpoint is also expected by most CDP clients for target discovery.

## Demo

Demo

Before: `{"webSocketDebuggerUrl": "ws://127.0.0.1:9222/"}`

After: `{"Browser": "Lightpanda/X.Y", "Protocol-Version": "1.3", "User-Agent": "Lightpanda/X.Y", "webSocketDebuggerUrl": "ws://127.0.0.1:9222/"}`

## Changes

- `buildJSONVersionResponse` in `src/Server.zig`: added `Browser`, `Protocol-Version`, `User-Agent` fields using `lp.build_config.version`
- `handleHTTPRequest`: added handler for `/json/list`, `/json/list/`, `/json`, `/json/` returning empty JSON array `[]`
- Added `empty_json_list_response` constant for the static response
- Updated test to verify all new fields are present in the response

## Testing

Updated the existing `buildJSONVersionResponse` test to verify:
- HTTP 200 status and correct content type
- All four JSON fields present: `Browser`, `Protocol-Version`, `User-Agent`, `webSocketDebuggerUrl`
- Version string from `build_config` embedded in Browser and User-Agent fields

### DIFF
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -755,20 +755,16 @@ test "server: 404" {
 }
 
 test "server: get /json/version" {
-    const expected_response =
-        "HTTP/1.1 200 OK\r\n" ++
-        "Content-Length: 48\r\n" ++
-        "Connection: Close\r\n" ++
-        "Content-Type: application/json; charset=UTF-8\r\n\r\n" ++
-        "{\"webSocketDebuggerUrl\": \"ws://127.0.0.1:9222/\"}";
-
     {
         // twice on the same connection
         var c = try createTestClient();
         defer c.deinit();
 
         const res1 = try c.httpRequest("GET /json/version HTTP/1.1\r\n\r\n");
-        try testing.expectEqual(expected_response, res1);
+        try testing.expect(std.mem.startsWith(u8, res1, "HTTP/1.1 200 OK\r\n"));
+        try testing.expect(std.mem.indexOf(u8, res1, "\"Browser\": \"Lightpanda/") != null);
+        try testing.expect(std.mem.indexOf(u8, res1, "\"Protocol-Version\": \"1.3\"") != null);
+        try testing.expect(std.mem.indexOf(u8, res1, "\"webSocketDebuggerUrl\": \"ws://127.0.0.1:9222/\"") != null);
     }
 
     {
@@ -777,7 +773,8 @@ test "server: get /json/version" {
         defer c.deinit();
 
         const res1 = try c.httpRequest("GET /json/version HTTP/1.1\r\n\r\n");
-        try testing.expectEqual(expected_response, res1);
+        try testing.expect(std.mem.startsWith(u8, res1, "HTTP/1.1 200 OK\r\n"));
+        try testing.expect(std.mem.indexOf(u8, res1, "\"Browser\": \"Lightpanda/") != null);
     }
 }
 
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -444,6 +444,14 @@ pub const Client = struct {
             return false;
         }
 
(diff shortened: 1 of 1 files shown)


src/Server.zig
```diff
@@ -486,8 +494,15 @@ fn buildJSONVersionResponse(
...
         });
     }
-    const body_format = "{{\"webSocketDebuggerUrl\": \"ws://{s}:{d}/\"}}";
-    const body_len = std.fmt.count(body_format, .{ host, port });
+    const version = lp.build_config.version;
+    const body_format =
+        "{{" ++
+        "\"Browser\": \"Lightpanda/{s}\", " ++
+        "\"Protocol-Version\": \"1.3\", " ++
+        "\"User-Agent\": \"Lightpanda/{s}\", " ++
```
krichprollsch: I would change the User-Agent to return a fixed `Lightpanda/1.0` here.
ie. the default User-Agent.

Idk if it's uselful to return the resolved UA according to --user-agent-suffix` and `--user-agent` options, but I think  `Lightpanda/1.0`  is better than ``Lightpanda/1.0.0-dev.5492+a127284b`

krichprollsch (commented):
I'm wondering how clients takes care of this `/json/version` vs CDP [`Browser.getVersion`](https://chromedevtools.github.io/devtools-protocol/tot/Browser/#method-getVersion)

We are currently faking the the CDP version returned because we saw clients refusing to connect with real versions.

* https://github.com/lightpanda-io/browser/blob/main/src/cdp/domains/browser.zig#L33
* https://github.com/lightpanda-io/browser/blob/main/src/cdp/domains/browser.zig#L61-L70

mvanhorn:
Good catch on the User-Agent — changed it to fixed `Lightpanda/1.0` in 2629d21. The dev version string isn't useful for CDP clients.

Re: `/json/version` vs CDP `Browser.getVersion` — they serve different purposes:

- `/json/version` is the HTTP discovery endpoint (hit before WebSocket connection). Puppeteer and chromedp use it to find `webSocketDebuggerUrl`. Returning `Lightpanda/1.0` here is honest identification.
- `Browser.getVersion` is a CDP command sent after connection. It returns the Chrome identity (`Chrome/124.0.6367.29`, full Chrome UA) because clients like Puppeteer check it for compatibility and would refuse to work with an unknown browser.

The inconsistency is intentional — discovery says "I'm Lightpanda" while the CDP protocol says "I'm Chrome-compatible" so clients don't bail. This matches how the codebase already handles it (the `CDP_USER_AGENT` const in browser.zig is explicitly documented as "Many clients check the CDP server is a Chrome browser").

If you'd prefer the `/json/version` endpoint to also fake Chrome identity for maximum compatibility, I can change that — but I think honest identification at the HTTP layer is better since automation tools that specifically target Lightpanda (like the MCP server) would want to confirm they're connected to the right thing.

krichprollsch:
rebased and merged via #2133 

PR #2125, opened 2026-04-10
feat: add --cookies-file flag for session persistence

### LINKED ISSUE #335: Pass in cookies / dump out cookies?
how does lightpanda handle cookies for a given request? ie let's say i make a request to `google.com` which 301 redirects to `www.google.com` but sets a cookie, in the subsequent request to `www.google.com` is the cookie sent/persisted?

can i use lightpanda to pass in cookies?

alternatively, can i use lightpanda cli to dump cookies at the end of a request?

the interface could be something like (to pass in cookies) (interface is up to you, but this is just an idea):
```
./lightpanda --dump --cookies 'cookie-str [or cookie-json]' 'https://news.google.com'
```

and to get cookies out:
```
./lightpanda --dump --dump-cookies --cookies 'cookie-str [or cookie-json]' 'https://news.google.com'

=> 
{
  dom: '<html> ... </html>',
  cookies: 'cookie-json'
}
```

mvanhorn:
## Summary

Adds a `--cookies-file` CLI option that loads cookies from a JSON file at startup and saves them back on exit. Works with `fetch`, `serve`, and `mcp` commands.

## Why this matters

AI agents that need to maintain login sessions across multiple Lightpanda invocations are currently blocked. Issue #335 (7 comments) asks for cookie pass/dump. Steel Browser and Browserbase both offer session persistence, which is table-stakes for browser automation.

The cookie JSON format matches CDP `Network.Cookie`, so cookies exported from Puppeteer's `page.cookies()` or Playwright's `context.cookies()` work directly.

## Demo

Demo

Usage:
```
lightpanda fetch --cookies-file session.json --dump markdown https://example.com
```

First run saves cookies to `session.json`. Subsequent runs load them, sending cookies with requests. Updated cookies are saved back on exit.

## Changes

**`src/Config.zig`**: Added `cookies_file` field to `Common` options, accessor `cookiesFile()`, CLI parsing for `--cookies-file`, and help text.

**`src/cookies.zig`** (new): Cookie file I/O module with `loadFromFile` and `saveToFile`. Parses/writes JSON arrays of `{name, value, domain, path, expires, secure, httpOnly}` objects. Uses `std.json` for parsing and manual JSON writing for output. Handles missing files gracefully (first run creates the file on exit).

**`src/lightpanda.zig`**: Calls `cookies.loadFromFile` after session creation and `defer` calls `cookies.saveToFile` before session cleanup.

## Cookie file format

```json
[
  {"name": "session_id", "value": "abc123", "domain": ".example.com",
   "path": "/", "expires": 1744329600, "secure": true, "httpOnly": true}
]
```

Fixes #335

### DIFF
--- a/src/Config.zig
+++ b/src/Config.zig
@@ -177,9 +177,16 @@ pub fn httpCacheDir(self: *const Config) ?[]const u8 {
     };
 }
 
-pub fn cookiesFile(self: *const Config) ?[]const u8 {
+pub fn cookieFile(self: *const Config) ?[]const u8 {
     return switch (self.mode) {
-        inline .serve, .fetch, .mcp => |opts| opts.common.cookies_file,
+        inline .serve, .fetch, .mcp => |opts| opts.common.cookie,
+        else => null,
+    };
+}
+
+pub fn cookieJarFile(self: *const Config) ?[]const u8 {
+    return switch (self.mode) {
+        inline .fetch, .mcp => |opts| opts.common.cookie_jar,
         else => null,
     };
 }
@@ -303,7 +310,8 @@ pub const Common = struct {
     user_agent_suffix: ?[]const u8 = null,
     user_agent: ?[]const u8 = null,
     http_cache_dir: ?[]const u8 = null,
-    cookies_file: ?[]const u8 = null,
+    cookie: ?[]const u8 = null,
+    cookie_jar: ?[]const u8 = null,
 
     web_bot_auth_key_file: ?[]const u8 = null,
     web_bot_auth_keyid: ?[]const u8 = null,
@@ -442,11 +450,17 @@ pub fn printUsageAndExit(self: *const Config, success: bool) void {
         \\                Omitting this will result is no caching.
         \\                Defaults to no caching.
         \\
-        \\--cookies-file
-        \\                Path to a JSON file for cookie persistence. Cookies are loaded
-        \\                from this file at startup and saved back on exit.
+        \\--cookie
+        \\                Path to a JSON file to load cookies from at startup
+        \\                (read-only).
         \\                Format: [{{name, value, domain, path, expires, secure, httpOnly}}]
-        \\                Defaults to no persistence.
+        \\                Defaults to no cookie loading.
+        \\
+        \\--cookie-jar
+        \\                Path to a JSON file to save cookies to on exit (write-only).
+        \\                Available for fetch and mcp commands.
+        \\                Format: [{{name, value, domain, path, expires, secure, httpOnly}}]
+        \\                Defaults to no cookie saving.
     ;
 
     //                                                                     MAX_HELP_LEN|
@@ -699,6 +713,14 @@ fn parseServeArgs(
                 return error.InvalidArgument;
             };
             continue;
+        }
+
+        if (std.mem.eql(u8, "--cookie-jar", opt)) {
+            log.fatal(.app, "invalid argument value", .{
+                .arg = opt,
+                .detail = "--cookie-jar is only available for fetch and mcp",
+            });
+            return error.InvalidArgument;
         }
 
         if (try parseCommonArg(allocator, opt, args, &serve.common)) {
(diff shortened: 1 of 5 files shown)


krichprollsch:
Thanks for the contribution.
This is an important feature, but I have some questions and comments.

1. Curl uses 2 distinct options to load and save cookies: `--cookie` is a read-only option to load cookies, and `--cookie-jar` is a write-only option to save cookies. If you want both, you can use the same file for both options. I think it's good to follow the same convention.

2. Curl uses the [Netscape file format](https://curl.se/docs/http-cookies.html) to save cookies. However, this format has been abandoned by browsers and does not support modern fields like `secure` and `http-only`, which are important when running JS. So I don't want to use this format, but I'm wondering if other formats exist that we could be compatible with.
Once introduced, a file format can be hard to change while keeping backward compatibility.
EDIT: [Claude thinks](https://claude.ai/share/8dca22f5-953a-451d-a1bc-847258924def) the cookie JSON format you implemented is compatible enough.

3. My biggest concern is about CDP server and multiple connection support. You can create multiple connections through CDP, and each connection gets its own browser instance. How should we handle the cookie store in this case? Should the last browser instance replace the cookie storage? Should we create one file per session?

4. You say "Works with fetch, serve, and mcp commands." but as far as I can see, the current implementation only loads and saves cookies for the `fetch` command — is that correct?

mvanhorn:
Thanks for the detailed review!

Fixed the build errors in cc6ffb50 — unescaped braces in the help text format string and wrong writer API usage (`fs.File.Writer` vs `std.Io.Writer` interface). CI should go green now.

On your points:

**1. `--cookie` / `--cookie-jar` split:** Agree, following curl's convention makes sense. I'll split it into `--cookie` (load-only) and `--cookie-jar` (save-only), where using the same file for both gives the current behavior.

**2. JSON format:** Good to know you're OK with the CDP-compatible JSON format. I'll keep it as-is.

**3. CDP multi-session:** This is the key design question. Currently each `Session` owns its own `cookie_jar`. For the CDP/serve case with multiple connections, I see a few options:

- **Shared jar, last-write-wins:** All sessions read from the file at start, each session saves on close. Simple but lossy if sessions overlap.
- **Per-session files:** e.g. `cookies-{session_id}.json`. Avoids conflicts but creates file sprawl.
- **Load once, save once:** Load cookies into the `Browser` (or a shared store) when the server starts, save when it shuts down. Sessions share the jar in-memory.

I'd lean toward option 3 for serve/CDP since the cookie jar is really "this browser's state" rather than per-tab state. What do you think?

**4. Scope:** You're right, the current implementation only hooks into `fetch()`. The `cookiesFile()` accessor already handles all three modes in Config, but the actual load/save logic is only in `lightpanda.zig:fetch()`. I'll extend it to `serve` and `mcp` once we align on the multi-session approach (point 3).

krichprollsch:
I suggest to do the following:

* add `--cookie` option for `fetch`, `mcp` and `cdp`, no write problem in this case
* add `--cookie-jar` only for `fetch` and `mcp` for now, the implementation is straightforward

Let's keep `--cookie-jar` implementation for CDP on hold for now.

But I'm thinking of another alternative: add a `LP` custom command on CDP to request a cookie dump into `--cookie-jar`. Then the control would be delegated to the user.

mvanhorn:
Done in 4b78b19 — split into `--cookie` (read-only) and `--cookie-jar` (write-only) following curl's convention:

- `--cookie` available on fetch, mcp, and serve/CDP (load only)
- `--cookie-jar` available on fetch and mcp (save on exit)
- serve mode rejects `--cookie-jar` with a clear error
- Added cookie integration to MCP server init/deinit and CDP session creation

Holding off on the CDP custom command idea for now per your suggestion.

krichprollsch:
I created  #2169 to rebase and fix conflict + add small changes

mvanhorn:
Thanks @krichprollsch! Happy to see this picked up.

krichprollsch:
done in #2169

krichprollsch:

PR #2134, opened 2026-04-10
Cache-Control is public by default

karlseguin:
- If private isn't specified, default to public.
- Add some tests
- Optimize parsing by lower-casing once and switch to std.mem


krichprollsch:
The rules by default were conservative to wait for cache validation mechanism.

mookums (approved):

PR #2141, opened 2026-04-11
feat(console): implement group, groupCollapsed, and groupEnd

luckyyyyy:
Add missing console grouping APIs per the WHATWG Console spec:
- console.group(...data): logs label and increments group depth
- console.groupCollapsed(...data): same as group (headless mode has no visual collapsing, so behavior is identical to group)
- console.groupEnd(): decrements group depth, clamped at 0

Depth is tracked via _group_depth (u32) on the Console struct. Saturation arithmetic (+|=) prevents overflow on runaway group() calls.

Fixes TypeError crashes on sites that use console.group* APIs (e.g. React/Vite dev builds).

### DIFF
--- a/src/browser/webapi/Console.zig
+++ b/src/browser/webapi/Console.zig
@@ -26,7 +26,6 @@ const Console = @This();
 
 _timers: std.StringHashMapUnmanaged(u64) = .{},
 _counts: std.StringHashMapUnmanaged(u64) = .{},
-_group_depth: u32 = 0,
 
 pub const init: Console = .{};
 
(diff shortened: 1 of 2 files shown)


luckyyyyy:

src/browser/webapi/Console.zig
```diff
@@ -26,6 +26,7 @@ const Console = @This();
 
 _timers: std.StringHashMapUnmanaged(u64) = .{},
 _counts: std.StringHashMapUnmanaged(u64) = .{},
+_group_depth: u32 = 0,
```
karlseguin: If we're just going to ignore the group_depth (which I'm ok with), why bother even capturing it? Why not remove this field and  have `groupEnd` do nothing and `group` and `groupCollapse` log, without incrementing `_group_depth`.

karlseguin:
Thanks for this. Your commit was merged as part of https://github.com/lightpanda-io/browser/pull/2144.

PR #2155, opened 2026-04-13
`Fetch`: cookie jar should only be included for `include` and same or…

nikneym:
…igin requests


src/browser/webapi/net/Fetch.zig
```diff
@@ -86,6 +86,24 @@ pub fn init(input: Input, options: ?InitOpts, page: *Page) !js.Promise {
         log.debug(.http, "fetch", .{ .url = request._url });
     }
 
+    const cookie_jar = switch (request._credentials) {
+        .omit => null,
+        .include => &page._session.cookie_jar,
+        .@"same-origin" => blk: {
```
karlseguin: `page.isSameOrigin(request._url),`

nikneym: That's much cleaner, though I think the function should be returning `bool` directly without error union. Made the change for that too.

PR #2182, opened 2026-04-17
Fix `ariaSnapshot` noise vs Chromium

### LINKED ISSUE #1813: ariaSnapshot much noisier than chromium's
Hi! Love this project, have been trying it out for a few hours and it's great.

One thing I noticed trying to build some agent-y automation around it: `await page.locator(':root').ariaSnapshot()` produces something much bigger than chromium does, in some cases.

Repro:

Install Playwright Core:

```bash
npm install playwright-core
```

Start Lightpanda with a non-zero timeout:

```bash
lightpanda serve --host 127.0.0.1 --port 9222 --timeout 600
```

Get a Chromium baseline for ariaSnapshot() on wikipedia.org:

```bash
node --input-type=module <<'EOF' > chromium-aria.txt
import { chromium } from 'playwright-core'
const browser = await chromium.launch({ headless: true })
const page = await browser.newPage()
await page.goto('https://www.wikipedia.org/', { waitUntil: 'load' })
console.log(await page.locator(':root').ariaSnapshot())
await browser.close()
EOF
```

Get the Lightpanda one via CDP (after running `lightpanda serve --host 127.0.0.1 --port 9222 --timeout 600`):

```bash
node --input-type=module <<'EOF' > lightpanda-aria.txt
import { chromium } from 'playwright-core'
const browser = await chromium.connectOverCDP('http://127.0.0.1:9222')
const context = await browser.newContext()
const page = await context.newPage()
await page.goto('https://www.wikipedia.org/', { waitUntil: 'load' })
console.log(await page.locator(':root').ariaSnapshot())
await browser.close()
EOF
```

```bash
diff chromium-aria.txt lightpanda-aria.txt
```

On Lightpanda, relative to Chromium, there's:

- an extra top-level `text: Wikipedia`
- an extra decorative heading subtree for `Wikipedia The Free Encyclopedia`
- an extra dialog subtree
- unnamed `searchbox` / `combobox` where Chromium gives names
- a much larger `navigation "All languages"` subtree
- hundreds of extra `listitem` nodes under the language lists

The main thing here is that "All languages" has 300+ additional items, maybe chromium is skipping that list simply because it's so big it's probably not useful in the snapshot?

arrufat:
Four independent gaps caused Playwright's `page.locator(':root').ariaSnapshot()` to emit ~6.7× more lines than Chromium on `wikipedia.org`. This PR fixes all four.

| | Chromium | Lightpanda main | This PR |
| --- | ---: | ---: | ---: |
| Lines | 185 | 1,231 | **184** |
| Bytes | 7,214 | 40,095 | **7,198** |
| `searchbox` named | ✓ | ✗ | **✓** |
| `combobox` named | ✓ | ✗ | **✓** |

**Root causes and fixes:**

1. **CDP `Accessibility.getFullAXTree` didn't filter hidden subtrees.**
2. **Accessible names from `<label>` weren't wired up.**
3. **`getComputedStyle(el).display` / `.visibility` ignored stylesheet rules**
4. **CSS tokenizer mis-parsed unquoted `url(…)` values.**

Closes #1813

### DIFF
--- a/src/browser/StyleManager.zig
+++ b/src/browser/StyleManager.zig
@@ -231,6 +231,33 @@ pub fn isHidden(self: *StyleManager, el: *Element, cache: ?*VisibilityCache, opt
     return false;
 }
 
+/// Computed display:none for a single element (own property, no ancestor walk).
+/// Also honors the HTML `hidden` attribute, matching the UA stylesheet rule
+/// `[hidden] { display: none }`.
+pub fn hasDisplayNone(self: *StyleManager, el: *Element) bool {
+    self.rebuildIfDirty() catch return false;
+    if (el.hasAttributeSafe(comptime .wrap("hidden"))) return true;
+    return self.isElementHidden(el, .{});
+}
+
+/// Computed visibility:hidden for an element, walking ancestors since `visibility`
+/// inherits by default. Only considers visibility-level hidden (not display:none)
+/// — display:none on an ancestor means the element isn't rendered, but its
+/// computed `visibility` still reflects inherited visibility chain.
+pub fn hasVisibilityHiddenInherited(self: *StyleManager, el: *Element) bool {
+    self.rebuildIfDirty() catch return false;
+    var current: ?*Element = el;
+    while (current) |elem| {
+        const combined = self.isElementHidden(elem, .{ .check_visibility = true });
+        if (combined) {
+            const display_only = self.isElementHidden(elem, .{});
+            if (!display_only) return true;
+        }
+        current = elem.parentElement();
+    }
+    return false;
+}
+
 /// Check if a single element (not ancestors) is hidden.
 fn isElementHidden(self: *StyleManager, el: *Element, options: CheckVisibilityOptions) bool {
     // Track best match per property (value + priority)
--- a/src/browser/StyleManager.zig
+++ b/src/browser/StyleManager.zig
@@ -240,18 +240,16 @@ pub fn hasDisplayNone(self: *StyleManager, el: *Element) bool {
     return self.isElementHidden(el, .{});
 }
 
-/// Computed visibility:hidden for an element, walking ancestors since `visibility`
-/// inherits by default. Only considers visibility-level hidden (not display:none)
-/// — display:none on an ancestor means the element isn't rendered, but its
-/// computed `visibility` still reflects inherited visibility chain.
+/// Computed visibility:hidden for an element, considering only the `visibility`
+/// chain (walks ancestors since `visibility` inherits by default). Ignores
+/// display:none: an ancestor with display:none means the element isn't
+/// rendered, but its computed `visibility` still reflects inherited visibility.
 pub fn hasVisibilityHiddenInherited(self: *StyleManager, el: *Element) bool {
     self.rebuildIfDirty() catch return false;
     var current: ?*Element = el;
     while (current) |elem| {
-        const combined = self.isElementHidden(elem, .{ .check_visibility = true });
-        if (combined) {
-            const display_only = self.isElementHidden(elem, .{});
-            if (!display_only) return true;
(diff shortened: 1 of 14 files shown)


src/cdp/AXNode.zig
```diff
@@ -1068,7 +1155,7 @@ fn isIgnore(self: AXNode, page: *Page) bool {
             var it = node.childrenIterator();
             while (it.next()) |child| {
                 const axn = AXNode.fromNode(child);
-                if (!axn.isIgnore(page)) {
+                if (!axn.isIgnore(page, cache, false)) {
```
karlseguin: is `false` here correct? the `is_area_hidden` isn't preserved for children? Parent can be hidden, children might still be visible?

src/browser/StyleManager.zig
```diff
@@ -231,6 +231,33 @@ pub fn isHidden(self: *StyleManager, el: *Element, cache: ?*VisibilityCache, opt
...
+
+/// Computed visibility:hidden for an element, walking ancestors since `visibility`
+/// inherits by default. Only considers visibility-level hidden (not display:none)
+/// — display:none on an ancestor means the element isn't rendered, but its
+/// computed `visibility` still reflects inherited visibility chain.
+pub fn hasVisibilityHiddenInherited(self: *StyleManager, el: *Element) bool {
+    self.rebuildIfDirty() catch return false;
+    var current: ?*Element = el;
+    while (current) |elem| {
+        const combined = self.isElementHidden(elem, .{ .check_visibility = true });
```
karlseguin: I think adding a `check_display: bool = true` to `CheckVisibilityOptions` would help clean this up a bit.

src/cdp/AXNode.zig
```diff
@@ -948,7 +987,48 @@ fn writeAccessibleNameFallback(node: *DOMNode, writer: *std.Io.Writer, page: *Pa
...
+
+    if (Label.findWrappingLabel(el)) |wrap_label| {
+        if (try writeLabelInnerText(wrap_label, page, w)) return .label_wrap;
+    }
+
+    return null;
+}
+
+fn writeLabelInnerText(label_el: *DOMNode.Element, page: *Page, w: anytype) !bool {
+    var buf: std.Io.Writer.Allocating = .init(page.call_arena);
```
karlseguin, line 1034: In a separate PR, what do you think of adding a `temp_arena` to `Writer`, so that it can be reset between each of these temp allocations:

```zig
defer temp_arena.reset(.{.retain_with_limit = 64 * 1024});
var buf: std.Io.Writer.Allocating = .init(temp_arena.allocator());
```

I can do it, or you can.

src/cdp/AXNode.zig
```diff
@@ -948,7 +987,48 @@ fn writeAccessibleNameFallback(node: *DOMNode, writer: *std.Io.Writer, page: *Pa
...
+        .button, .meter, .output, .progress, .select, .textarea, .input => true,
+        else => false,
+    };
+}
+
+fn writeLabelName(node: *DOMNode, el: *DOMNode.Element, page: *Page, w: anytype) !?AXSource {
+    if (el.getAttributeSafe(comptime .wrap("id"))) |id_value| {
+        if (id_value.len > 0) {
+            if (node.ownerDocument(page)) |doc| {
+                if (Label.findLabelByFor(doc.asNode(), id_value)) |label_el| {
```
karlseguin: This is a full tree walk for every labellable tag. Could lazily populate a lookup, trading memory for speed.

src/cdp/AXNode.zig
arrufat: It is, because there's an early return a few lines above. But I will change it to `in_area_hidden` to be clearer and robust for possible future refactors.

src/cdp/AXNode.zig
arrufat: Sure, I can do it :)

karlseguin (approved):

PR #2184, opened 2026-04-17
ax: route AXNode.Writer scratch allocations through a dedicated arena

arrufat:
Depends on #2182

### DIFF
--- a/src/cdp/AXNode.zig
+++ b/src/cdp/AXNode.zig
@@ -28,6 +28,10 @@ const Node = @import("Node.zig");
 
 const AXNode = @This();
 
+// Max bytes retained in the name-resolution scratch arena across resets.
+// Anything beyond is freed back to the backing allocator.
+const scratch_retain_limit = 64 * 1024;
+
 // Need a custom writer, because we can't just serialize the node as-is.
 // Sometimes we want to serializ the node without children, sometimes with just
 // its direct children, and sometimes the entire tree.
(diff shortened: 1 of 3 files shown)


src/cdp/AXNode.zig
```diff
@@ -1286,12 +1318,15 @@ test "AXNode: writer" {
     // allocator and let the page arena clean them up.
     var visibility_cache: DOMNode.Element.VisibilityCache = .empty;
     var label_index: Label.LabelByForIndex = .{};
+    var temp_arena = std.heap.ArenaAllocator.init(page.call_arena);
```
karlseguin: Nested arenas are not intuitive. Freeing the child area almost never frees that memory back to the parent (because the parent itself is a bump allocator).

In this specific case, it's still a net-gain. Best case: reuse 64K. Worst case: no different than before, page_arena grows.

So, I think this is all fine. And I wasn't going to say anything Except, I saw later on, you do:

```zig 
var temp_arena = std.heap.ArenaAllocator.init(page.call_arena);
defer temp_arena.deinit();
```

And you can kind of solve both issues with:

```zig
const allocator = try page.getArena(.medium, "AXNode");
defer page.releaseArena(allocator);
```

This returns an `Allocator`. Which is generally convenient. Your `scratchAllocator` can just be `temp_arena orelse page.call_arena`, but your `resetAllocator` is going to be a bit uglier: 

```zig
if (temp_arena) |a| {
    const arena: *ArenaAllocator = @ptrCast(@alignCast(a.ptr));
    _ = arena.reset(.{ .retain_with_limit = scratch_retain_limit });
}
```

arrufat: Not that ugly, using:

```zig
defer if (temp_arena) |a| page._session.arena_pool.reset(a, scratch_retain_limit);
```

PR #2196, opened 2026-04-20
Sqlite

karlseguin:
Replaces: https://github.com/lightpanda-io/browser/pull/2076

Adds an app.storage which is a union around configurable storage engine (currently, only sqlite).

It is _not_ being used anywhere right now. The goal is to get feedback on the implementation and then move cache to it.

This doesn't expose a generic query API. The goal is that the storage will expose high level methods, e.g. `cacheGet(req: CacheGetRequest)` and every storage engine will translate the `storage.CacheGetRequest` as needed.

A thin wrapper around the Sqlite C api is included, e.g. exec(SQL, .{args}) a `rows` and `row` fetcher. A connection pool is included. By default, an in-memory DB is currently created. And a `migrations` table with an id of `1` is created/inserted. I don't imagine needing fancy migratations.


krichprollsch:
LGTM

PR #2203, opened 2026-04-21
Common httpclient

mrdimidium:
Finally moves the entire HTTP stack to the network, removes libcurl from HttpClient.

> [!WARNING]
>  There is a slight performance regression (you may see a longer timeout), I'm looking into it.


mrdimidium:
See #2303

PR #2204, opened 2026-04-21
Shrink screenshot.png

karlseguin:
I asked claude to do for the png what it did for the pdf. The savings aren't as big AND, it isn't as nice as the original.

I'm tempted to say we should keep the original, but I don't feel strongly about it and maybe saving a few KB is worth it?


krichprollsch:
Let's save some KB.

But we should already be careful with these changes, [Hermes partially relies on the file size.](https://github.com/NousResearch/hermes-agent/pull/7144/changes#diff-b4397d8cbd4671ffc08da0c5a73f5fd4a3c35cd89922879270dded1d3b311f52R494)
No impact in this case, but somebody could rely more closely (shasum?).

I guess we will have surprises with the definition of public API and BC in the future :sweat:

krichprollsch (approved):

PR #2226, opened 2026-04-23
Setup timeout via tcp keepalive

mrdimidium:


krichprollsch:
I like it. I'm wondering if we really need to keep `--timeout` honestly. but maybe it's still useful?

mrdimidium:
@krichprollsch, I don't think tuning TCP Keepalive is actually useful, but removing the parameter could break existing installations.

Could we write a warn if an explicit --timeout is specified and ignore the value?

karlseguin:
Agree with accepting the parameter but no-oping it with a warning.

Issue #2056, opened 2026-04-01
[FEATURE REQUEST] Add script/eval js in CLI with fetch


kelvinauta:
I think it would be very useful when making fetch to a site, you could load a custom script.
That way you can create quick automations without needing to open a whole process with Puppetter or Playwright.
For example, I would like to add a script simply to scroll so that the content keeps loading...

I can suggest something like
`lightpanda fetch <url> (--script <file> | --eval <javascript/text>)`

kaelkaram1231-bit:
This is a really nice idea.
Being able to inject a small script during fetch would cover a lot of
lightweight automation use cases (like scrolling, triggering lazy loads,
etc.) without the overhead of Puppeteer/Playwright.

nikneym:
This is now supported (#2235), closing in favor.

PR #2235, opened 2026-04-24
`fetch`: add support for `--inject-script` and `--inject-script-file` options

nikneym:
Allows passing inline JS codes or files as an arg to be inserted into `<head>` tag. Addressing #2056.

### DIFF
--- a/src/Config.zig
+++ b/src/Config.zig
@@ -130,6 +130,24 @@ fn waitScriptFileValidator(allocator: Allocator, args: *std.process.ArgIterator)
     };
 }
 
+fn injectScriptFileValidator(
+    allocator: Allocator,
+    args: *std.process.ArgIterator,
+    list: *std.ArrayList([:0]const u8),
+) !void {
+    const path = args.next() orelse {
+        log.fatal(.app, "missing argument value", .{ .arg = "--inject-script-file" });
+        return error.InvalidArgument;
+    };
+
+    const bytes = std.fs.cwd().readFileAllocOptions(allocator, path, 1024 * 1024, null, .of(u8), 0) catch |err| {
+        log.fatal(.app, "failed to read file", .{ .arg = "--inject-script-file", .path = path, .err = err });
+        return error.InvalidArgument;
+    };
+
+    return list.append(allocator, bytes);
+}
+
 /// Definition for all the commands and its arguments. See @cli.zig for further.
 const Commands = cli.Builder(.{
     .{
@@ -163,7 +181,14 @@ const Commands = cli.Builder(.{
                 },
             },
             .{ .name = "wait_selector", .type = ?[:0]const u8 },
-            .{ .name = "script", .type = ?[:0]const u8 },
+            .{
+                .name = "inject_script",
+                .type = [:0]const u8,
+                .multiple = true,
+                .variants = .{
+                    .{ .name = "inject_script_file", .validator = injectScriptFileValidator },
+                },
+            },
             .{ .name = "terminate_ms", .type = ?u32 },
         },
         .shared_options = CommonOptions,
@@ -642,8 +667,15 @@ pub fn printUsageAndExit(self: *const Config, success: bool) void {
         \\--wait-script-file
         \\                Like --wait-script, but reads the script from a file.
         \\
-        \\--script        Path to a JavaScript file to execute in the page context
-        \\                once the --wait- conditions have been met, before the dump.
+        \\--inject-script JavaScript to inject as an inline <script> element in
+        \\                <head> once the --wait-* conditions have been met,
+        \\                before the dump.
+        \\                Can be passed multiple times; scripts run in order.
+        \\
+        \\--inject-script-file
+        \\                Like --inject-script, but reads the script from a file.
+        \\                Can be passed multiple times; can be mixed with
+        \\                --inject-script and runs in CLI order.
         \\
         \\--terminate-ms  Hard deadline in milliseconds. After this time elapses,
         \\                JavaScript execution is forcibly terminated (e.g. for
--- a/src/Config.zig
+++ b/src/Config.zig
@@ -163,6 +163,7 @@ const Commands = cli.Builder(.{
                 },
             },
             .{ .name = "wait_selector", .type = ?[:0]const u8 },
+            .{ .name = "script", .type = ?[:0]const u8 },
             .{ .name = "terminate_ms", .type = ?u32 },
         },
         .shared_options = CommonOptions,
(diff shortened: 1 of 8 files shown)


src/browser/Runner.zig
```diff
@@ -278,6 +278,33 @@ pub fn waitForSelector(self: *Runner, selector: [:0]const u8, timeout_ms: u32) !
     }
 }
 
+pub fn runScriptFile(runner: *Runner, allocator: std.mem.Allocator, file_path: [:0]const u8) !void {
+    const frame = runner.frame;
+
+    const bytes = std.fs.cwd().readFileAlloc(allocator, file_path, 64 * 1024 * 1024) catch |err| {
+        switch (err) {
```
karlseguin: Switch isn't buying you much.

```zig
... catch (err) {
  log.err(.app, "Runner.runScriptFile", .{ .path = file_path, .err = err });
  return err;
};
```

src/browser/Runner.zig
```diff
@@ -278,6 +278,33 @@ pub fn waitForSelector(self: *Runner, selector: [:0]const u8, timeout_ms: u32) !
...
+    frame.js.localScope(&ls);
+    defer ls.deinit();
+
+    var try_catch: js.TryCatch = undefined;
+    try_catch.init(&ls.local);
+    defer try_catch.deinit();
+
+    return ls.local.eval(bytes, "script") catch |err| {
+        const caught = try_catch.caughtOrError(frame.call_arena, err);
+        log.err(.app, "run script file error", .{ .err = caught });
```
karlseguin:
```suggestion
        log.warn(.app, "run script file error", .{ .err = caught, .path = file_path });
```

Nothing higher than `warn` for script error.

src/Config.zig
```diff
@@ -612,6 +613,9 @@ pub fn printUsageAndExit(self: *const Config, success: bool) void {
         \\--wait-script-file
         \\                Like --wait-script, but reads the script from a file.
         \\
+        \\--script        Path to a JavaScript file to execute in the page context
```
karlseguin: Worth sticking with `--script-file` for consistency with `--wait-script-file` ?

Maybe `--inject-script-file` is less ambiguous.  There might be other `--xyz-script-file` in the future

src/browser/Runner.zig
```diff
@@ -278,6 +278,33 @@ pub fn waitForSelector(self: *Runner, selector: [:0]const u8, timeout_ms: u32) !
     }
 }
 
+pub fn runScriptFile(runner: *Runner, allocator: std.mem.Allocator, file_path: [:0]const u8) !void {
+    const frame = runner.frame;
+
+    const bytes = std.fs.cwd().readFileAlloc(allocator, file_path, 64 * 1024 * 1024) catch |err| {
```
karlseguin: `--wait-script-file` used to be read in `Config` (I think that's broken now) and exposed a consistent view with `--wait-script`. I kind of like that model.  So you could add a `--inject-script-file` and `--inject-script` but the config would just expose `--inject-script`

src/lightpanda.zig
```diff
@@ -155,6 +156,10 @@ pub fn fetch(app: *App, url: [:0]const u8, opts: FetchOpts) !void {
         try runner.waitForScript(script, remaining);
     }
 
+    if (opts.script) |script| {
+        try runner.runScriptFile(app.allocator, script);
```
karlseguin: I thought we were going to inject these in the head?

src/Config.zig
nikneym: Yes, was unsure about that. Worth sticking to `--script-file` or `--inject-script-file` for consistency with `-wait-*` options.

src/lightpanda.zig
nikneym: Correct, made a change for that too. Now we're getting correct dump output with injected scripts. One thing we can also do is preferring `src` instead of inline script `<script>...</script>` style for `--inject-script-file`; either works but setting `src` might be what's deserved for `-file`.

src/Config.zig
```diff
@@ -641,6 +667,16 @@ pub fn printUsageAndExit(self: *const Config, success: bool) void {
         \\--wait-script-file
         \\                Like --wait-script, but reads the script from a file.
         \\
+        \\--inject-script JavaScript to inject as an inline <script> element in
+        \\                <head> once the --wait-* conditions have been met,
```
karlseguin: This is no longer true

src/browser/Frame.zig
```diff
@@ -1130,7 +1130,58 @@ fn frameDoneCallback(ctx: *anyopaque) !void {
                     self._parse_state = .complete;
                 }
 
-                const raw_html = html.buffer.items;
+                const raw_html = blk: {
+                    const inject_scripts = self._session.inject_scripts;
+                    if (inject_scripts.len > 0) {
```
karlseguin: I was thinking of something like this too. But after seeing the result....The input might not have a head tag (it's optional in HTML5) , and the parsing would be thrown off by something like `<meta name="i" value"3">`. 

Did you consider trying to inject it in `createElementNS` when `from_parser == true and _parse_mode == .document) and the element is 'head' (I think html5ever will _always_ emit a head, even if the source doesn't have one, but that need to be verified). Then you can execute it directly against the frame.js, without any DOM changes, like `waitForScript` does. (You might need a `_script_injectd: bool` to guard double-injection in this case, because I'm pretty sure for like a DocumetFragment we don't  want to inject.

nikneym: Well, this works but injected scripts won't appear in dumped HTML anymore. Good thing is html5ever is smart enough to figure out to not to make us create `<head>` tags more than once so we don't have to guard against that.

src/browser/Frame.zig
```diff
@@ -2122,12 +2122,36 @@ pub fn createElementNS(self: *Frame, namespace: Element.Namespace, name: []const
...
+
+                            var try_catch: JS.TryCatch = undefined;
+                            try_catch.init(&ls.local);
+                            defer try_catch.deinit();
+
+                            for (inject_scripts) |inject_script| {
+                                ls.local.eval(inject_script, "inject_script") catch |err| {
+                                    const caught = try_catch.caughtOrError(self.call_arena, err);
+                                    log.err(.app, "inject script error", .{ .err = caught });
+                                    return error.InjectScriptError;
```
karlseguin: I think just logging is better. Either break, or continue (but if you `continue`,you'll need to move the try_catch inside the loop).

karlseguin (commented):
this should be testable. You can add an `inject_script: ?[]const u8 = null` to the HtmlRunnerOpts (https://github.com/lightpanda-io/browser/blob/nikneym/cli-script-eval/src/testing.zig#L343) and which can set (and defer unset!) on the `test_session`

PR #2241, opened 2026-04-24
browser: bound per-tick memory growth on JS-heavy pages

arrufat:
Two related changes that stop Runner._wait from freezing inside a single tick on JS-heavy pages (github.com/features/copilot and similar React marketing sites reproduce deterministically), and nudge V8 to release wrappers during long waits.

1. HttpClient.processMessages drains every ready HTTP completion in one call, and each completion synchronously runs the transfer's done_callback — which for a main-page HTML response kicks the parser, allocating DOM nodes / strings and invoking JS. A burst of many simultaneous completions runs parser / JS work for seconds inside a single Runner._tick, allocating GBs before control returns to the wait loop. During that stall the wait loop can't hint GC, check bounds, or yield.

   Cap at 16 completions per tick. Remaining messages stay in curl's queue and are picked up on the next tick.

2. Runner._wait can iterate for the full opts.ms budget (up to 30s in fetch, longer in agent tool-use loops). V8 was only nudged to GC on session/page teardown (Browser.deinit, Page.deinit), so a page that stays alive while running heavy JS accumulates wrappers and external-ref'd Zig allocations V8 has no reason to drop. Fire memoryPressureNotification(.moderate) once per second from the wait loop.

Repro:

    lightpanda fetch --dump html --wait-ms 30000 \
        https://github.com/features/copilot

Before: wait loop iterates ~1s, then stalls inside one http_client.tick for 10+ seconds while RSS grows ~190 MB/s. Process OOMs the host (or cgroup) without returning from that tick.

After: wait loop iterates continuously through the full --wait-ms budget. Growth is still ~100 MB/s (the native-side DOM allocator leak is a separate, architectural fix — Node/Element need to participate in the V8 finalizer system so Frame.removeNode can _factory.destroy the removed node) but the process exits cleanly at the deadline instead of freezing inside curl.

### DIFF
--- a/src/browser/Runner.zig
+++ b/src/browser/Runner.zig
@@ -72,7 +72,21 @@ fn _wait(self: *Runner, comptime is_cdp: bool, opts: WaitOpts) !CDPWaitResult {
         .ms = 200,
         .until = opts.until,
     };
+
+    // Periodic V8 GC hint during long waits. V8 is otherwise only nudged on
+    // session/page teardown (Browser.zig, Page.zig), so a page that stays
+    // alive for seconds while running heavy JS accumulates wrappers and
+    // external-ref'd Zig allocations V8 has no reason to drop. `.moderate`
+    // speeds up incremental GC without stalling the tick.
+    const gc_hint_period_ns: u64 = std.time.ns_per_s;
+    var gc_hint_timer = std.time.Timer.start() catch unreachable;
+
     while (true) {
+        if (gc_hint_timer.read() >= gc_hint_period_ns) {
+            gc_hint_timer.reset();
+            self.session.browser.env.memoryPressureNotification(.moderate);
+        }
+
         const tick_result = self._tick(is_cdp, tick_opts) catch |err| {
             switch (err) {
                 error.JsError => {}, // already logged (with hopefully more context)


arrufat:
According to Claude, I cite verbatim:

> **Follow-up work to fully plug the leak** (out of scope for this change):
> 
> 1. **Enroll `Node`/`Element` in the V8 finalizer system.** `Factory` already allocates DOM nodes through `SlabAllocator` (`src/browser/Factory.zig:49,290,310`) which supports individual `rawFree`, but `Frame.removeNode` (`src/browser/Frame.zig:2658-2763`) never calls `_factory.destroy(child)` on the removed node. Every removed node leaks its slab slot until `Page.deinit()`. Naively destroying in `removeNode` is a use-after-free — JS may hold a wrapper, and `TreeWalker`/`Range`/`NodeIterator` keep raw `*Node` pointers across re-entrant operations. Proper fix: add `releaseRef` for `Node`/`Element` so V8's weak-callback path can `_factory.destroy` the node once the wrapper is GC'd and the node is detached from the tree.
> 
> 2. **Bound the main thread's `curl_multi_perform` / Transfer buffering during dump.** `Network.run()` keeps pumping curl on the main thread while the worker is in `dump.root` or other long Zig-side work. `Transfer.dataCallback` → `_stream_buffer.appendSlice` runs with no consumer, growing buffers unboundedly. Repro: `fetch --dump html --wait-ms 3000 https://github.com/features/copilot` still OOMs because the dump phase grows ~500 MB/s from accumulating response bodies. Options: pause-flag on the main thread during dump, or cap total buffered bytes across all active transfers.
>

karlseguin:
I'm surprised your first change does anything. The default max concurrent request is 10, so I'm not sure how limiting it to 16 messages helps.

For the follow up items.

1. Yes. We're slowly improving this, but have limited ourselves to non-DOM objects. These are normally simpler and can hold on to large allocations (e.g. XHR). But even these have proved difficult to get right (see various `release overflow` errors). DOM objects have the potential to be referenced from a much greater number of places and to be used by the CDP layer in ways that makes tracking more problematic (e.g. via the Inspector that seems to be able to cheat boundaries). Still, worth considering expanding this for more types once what we've done stabilizes.

2. I don't see how that's a problem today, since http_client.tick() runs in the same thread as the JavaScript, so tick data is always processed. But it might become an issue in the http refactor @mrdimidium 

That site, https://github.com/features/copilot, highlights 2 other problems:
1 - There's something in that site that results in an endless loop of the script. 
2 - ctrl-c doesn't terminate such scripts (at least not on mac)

karlseguin:
PR is good, just not sure about that `16`

arrufat:
> I'm surprised your first change does anything. The default max concurrent request is 10, so I'm not sure how limiting it to 16 messages helps.                                                                               
                                                                                                                                                                                                                                 
You're right. I will revert it.

I stopped investigating this, because I felt I couldn't actually tell if Claude was doing the right thing. I need to know the browser internals a bit better.

PR #2242, opened 2026-04-24
Brave's Rust adblock integration + various misc small fixes

codeandsolder:
Disclosure: mostly created by fighly suprevised collection of LLMs, all tests pass, seems to work, reviewed manually by somebody with no prior experience in Zig or Rust, feel free to disregard, use partially, or do anything else.
With regard to the CLA It seems to me like the process used would qualify me as a person entitled to grant licenses to the code under the current legal standards as I understand them.


codeandsolder:

karlseguin:
Thanks. I created https://github.com/lightpanda-io/browser/pull/2247 that extracts the "various misc small fixes". `String.wrap` is a relatively new addition which is why so many places weren't using it.

Will talk to the team about it, but I suspect we'll pass on the adblock integration, although something like this has been talked about before.

codeandsolder:
No problem, if you were to integrate it it would probably make more sense to bundle the FFI here and use the adblock-rust crate from there, I thought the idea of using Brave's was clever without realising it was C++ and cannot (as far as I understand) be easily used from Zig so the current implementation is a bit more clunky than it needs to.
I think it improves the usability a fair bit because it eliminates a lot of noise in the data for free and improves the speed quite a bit, but I understand it's not perfectly aligned with the scope/general ideology of the project, up to you of course, feel free to take the FFI from the other repo if you want, happy to help in any way you prefer.

Issue #2206, opened 2026-04-21
Add --max-timeout flag to CLI fetch command


fuleinist:
When running `lightpanda fetch` on slow or unresponsive sites, the command hangs indefinitely with no way to set a custom timeout.

**Solution:** Add a `--max-timeout <seconds>` flag to `lightpanda fetch` (and optionally `lightpanda serve`) so users can bail out early when a page is stalling, rather than having to kill the process manually.

**Use case:** Running automated scraping jobs where certain pages hang, and wanting graceful timeout handling without relying on external tooling like `timeout(1)`.

krichprollsch:
`--wait-ms` isn't what you want?

```
--wait-ms       Wait time in milliseconds. Supersedes all other --wait
                parameters.
                Defaults to 5000.
```

karlseguin:
@krichprollsch one issue is with scripts that never yield, `for(;;){}`  Even ctrl-c might not kill it.  I think the only reliable way to deal with this is to have a watcher thread which will issue a `TerminateExecution`. For `serve` I think we can come up with a more elegant solution (related to the conversations we've had about web socket ping and write failures). Are we ok with launching a watcher thread on `fetch`  should it be optional based on some new flag `--terminate-ms`  or something?

krichprollsch:
@karlseguin but doing this will stop the fetch process w/o getting any result or I'm wrong?

krichprollsch:
@fuleinist I would be great to have a real case example if you can

fuleinist:
Replying to the thread on krichprollsch's question about whether '--max-timeout' returns no result:

Not necessarily — there's a meaningful distinction between '--max-timeout' and 'no result':

With TerminateExecution, you get a **partial result** — whatever the browser DOM state was at the moment of termination:

| Scenario | Result |
|---|---|
| Page still loading HTML/assets | Partial HTML returned |
| JS modifying the DOM | Partial DOM captured |
| `for(;;){}` infinite loop | Execution killed, DOM state at that moment returned |

The result is *incomplete*, not *empty*. You get whatever Lightpanda's semantic tree captured before the JS context was terminated.

**Two separate concerns:**
1. `--max-timeout` — overall fetch deadline, returns whatever was captured up to that point
2. `--terminate-ms` (separate mechanism Karl proposed) — watcher thread that detects JS that never yields to the event loop and kills it specifically

Short answer: you do get a result, just a partial one.

karlseguin:
[...]

PR #2246, opened 2026-04-25
Adds --terminate-ms command line argument + ctrl-c improvements in fetch

karlseguin:
The main.zig path for `fetch` now captures the *Browser so that browser.env.terminate() can be called. This is a bit more complex than the serve path because the Browser owns the Isolate and can't be moved from one thread to another.

With main having access to the browser, two things are now possible: 
1 - We can support a --terminate-ms flag (https://github.com/lightpanda-io/browser/issues/2206) 
2 - ctrl-c can correctly stop blocked JavaScript processes

1 is implemented via setitimer to set a timer for SIGALRM, avoiding the need to add another "watcher" thread, or putting a timer in Network.run.

@krichprollsch mind helping test this on linux?  I created https://github.com/lightpanda-io/demo/pull/166 so that you can test with: `zig build run -- fetch --dump html --terminate-ms 1000 http://localhost:1234/endless.html`   or without `--terminate-ms` but with ctrl+c.

### DIFF
--- a/src/Config.zig
+++ b/src/Config.zig
@@ -133,6 +133,7 @@ const Commands = cli.Builder(.{
             .{ .name = "wait_until", .type = ?WaitUntil },
             .{ .name = "wait_script", .type = ?[:0]const u8 },
             .{ .name = "wait_selector", .type = ?[:0]const u8 },
+            .{ .name = "terminate_ms", .type = ?u32 },
         },
         .shared_options = CommonOptions,
     },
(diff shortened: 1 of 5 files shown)


krichprollsch:
I confirm it works correctly on linux :+1:

src/Config.zig
```diff
@@ -610,6 +611,12 @@ pub fn printUsageAndExit(self: *const Config, success: bool) void {
         \\--wait-script-file
         \\                Like --wait-script, but reads the script from a file.
         \\
+        \\--terminate-ms  Hard deadline in milliseconds. After this time elapses,
```
krichprollsch, line 614: I'm wondering if we shouldn't merge `--wait-ms` and `--terminate-ms`.
is it useful to have both? 🤔

krichprollsch (approved):

src/Config.zig
karlseguin: Maybe. When `--wait-ms` triggers (and, in some cases it might never, e.g. endless loop), you can still execute JavaScript. So you can combine `--wait-ms` with `--wait-script`, for example.  When `--terminate-ms` triggers, you can no longer run any JavaScript, so it can't be used with `--wait-script`.

To me, the question is, between `--wait-until`, `--wait-script` and `--wait-script-file`  does the current `--wait-ms` behavior add any value.

Should we change the default behavior to: `--wait-until load` and then have `--wait-ms` behave like `--terminate-ms`. I like this because it's more deterministic, but it could break existing cases that just use defaults.

PR #2248, opened 2026-04-26
Add and default to Blackhole storage

karlseguin:
Tweak sqlite build to omit more features, hoping to shrink the size when sqlite is used.


karlseguin:
This helps win back much of the of the recent memory increase.

krichprollsch (approved):

PR #2255, opened 2026-04-27
network: accept empty params on clearBrowserCookies, add getAllCookies

### LINKED ISSUE #2254: network: clearBrowserCookies rejects empty params object; getAllCookies missing
## Summary

Two CDP `Network.*` cookie-management gaps prevent spec-conformant clients from clearing cookies or enumerating them across origins:

1. **`Network.clearBrowserCookies` rejects an empty `params: {}` object** with `-31998 InvalidParams`, even though the CDP spec defines no parameters for the method. Most CDP libraries (chrome-remote-interface, chromedp, custom websocket clients) include `"params":{}` on every command unconditionally.
2. **`Network.getAllCookies` is not implemented** (returns `-31998 UnknownMethod`), so there is no way to enumerate the in-memory cookie jar across origins without iterating each visited URL via `Network.getCookies(urls:[...])`.

Both behaviors diverge from Chrome's CDP server, which (1) silently accepts `params:{}` and (2) implements `Network.getAllCookies` that returns the entire jar.

### A1 — `clearBrowserCookies` rejects an empty params object

`src/cdp/domains/network.zig` `clearBrowserCookies` validates parameters with an inverted guard:

```zig
fn clearBrowserCookies(cmd: *CDP.Command) !void {
    if (try cmd.params(struct {}) != null) return error.InvalidParams;
    ...
}
```

`cmd.params(T)` returns `null` when the message lacks a `params` field entirely, and a parsed value when `params` is present (even when it's the empty object `{}`). The condition therefore fires `InvalidParams` for any spec-conformant client that includes `"params":{}` on every command.

```mermaid
sequenceDiagram
    participant C as CDP Client
    participant LP as Lightpanda
    participant Jar as Cookie jar
    C->>LP: Network.clearBrowserCookies (with empty params object)
    LP->>LP: cmd.params parses params, returns non-null
    Note over LP: inverted guard fires
    LP-->>C: error -31998 InvalidParams
    Note over Jar: jar untouched
```

### B3 — `Network.getAllCookies` is not in the dispatch enum

`Network.getAllCookies` is missing from the action enum in `processMessage`, so `stringToEnum` returns `null` and the method falls through to `error.UnknownMethod`. There is no other CDP-level path that returns every cookie in the jar without specifying a URL or a `browserContextId`:

- `Network.getCookies(no urls)` — origin-scoped to the current frame; raises `InvalidDomain` on `about:blank`.
- `Network.getCookies(urls:[...])` — requires the client to already know every origin it has set cookies for.
- `Storage.getCookies(browserContextId)` — works, but it lives in the `Storage` domain, requires the caller to track the context id, and is the wrong domain for code that otherwise stays in `Network.*`.

```mermaid
sequenceDiagram
    participant C as CDP Client
    participant LP as Lightpanda
    participant Jar as Cookie jar
    C->>LP: Network.getAllCookies
    LP->>LP: stringToEnum returns null (method not in dispatch)
    LP-->>C: error -31998 UnknownMethod
    Note over Jar: jar contains cookies, but no Network.* path returns them all
```

## Expected behavior

Match Chrome's CDP behavior in both cases:

- [`Network.clearBrowserCookies`](https://chromedevtools.github.io/devtools-protocol/tot/Network/#method-clearBrowserCookies) — clears the jar; takes no parameters; ignores an empty params object.
- [`Network.getAllCookies`](https://chromedevtools.github.io/devtools-protocol/tot/Network/#method-getAllCookies) — returns every cookie in the jar regardless of the current frame's URL.

JSON-RPC convention is to silently ignore unrecognized parameters. Lightpanda's `Storage.clearCookies` and `Storage.getCookies` (in `src/cdp/domains/storage.zig`) already follow this convention; `Network.clearBrowserCookies` is the outlier.

```mermaid
sequenceDiagram
    participant C as CDP Client
    participant LP as Lightpanda
    participant Jar as Cookie jar
    C->>LP: Network.clearBrowserCookies (with empty params object)
    LP->>Jar: clearRetainingCapacity
    Jar-->>LP: ok
    LP-->>C: result is empty object
    C->>LP: Network.getAllCookies
    LP->>Jar: removeExpired then iterate jar.cookies
    Jar-->>LP: every cookie in jar
    LP-->>C: result.cookies = full list
```

## Reproducer

Self-contained Node + raw WebSocket. No browser-driver SDK, no framework. One `npm install --no-save ws` on first run.

`repro.sh`:

```bash
#!/usr/bin/env bash
set -euo pipefail
cd "$(cd "$(dirname "$0")" && pwd)"

PORT="${LP_PORT:-9222}"

if [ ! -d node_modules ]; then
  npm install --silent --no-save ws >/dev/null
fi

"$LIGHTPANDA_BIN" serve --host 127.0.0.1 --port "$PORT" >/tmp/lp-repro.log 2>&1 &
LP_PID=$!
trap 'kill -TERM "$LP_PID" 2>/dev/null || true; wait "$LP_PID" 2>/dev/null || true' EXIT

for _ in $(seq 1 50); do
  if grep -q "server running" /tmp/lp-repro.log 2>/dev/null; then
    break
  fi
  sleep 0.1
done

LP_WS_URL="ws://127.0.0.1:${PORT}" node repro.js
```

`repro.js`:

```js
"use strict";
const WebSocket = require("ws");
const WS_URL = process.env.LP_WS_URL || "ws://127.0.0.1:9222";

function connect(url) {
    ws.once("open", () => resolve(ws));
    ws.once("error", reject);
  });
}

async function main() {
  const ws = await connect(WS_URL);
  let id = 0;
  const pending = new Map();

  ws.on("message", (data) => {
    const msg = JSON.parse(data.toString());
    if (msg.id != null && pending.has(msg.id)) {
      const { resolve, reject } = pending.get(msg.id);
      pending.delete(msg.id);
      if (msg.error) reject(msg.error); else resolve(msg.result);
    }
  });

  function send(method, params = {}, sessionId) {
      const reqId = ++id;
      pending.set(reqId, { resolve, reject });
      const req = { id: reqId, method, params };
      if (sessionId) req.sessionId = sessionId;
      ws.send(JSON.stringify(req));
    });
  }

  const target = await send("Target.createTarget", { url: "about:blank" });
  const attached = await send("Target.attachToTarget", { targetId: target.targetId, flatten: true });
  const sessionId = attached.sessionId;

  let bugs = 0;

  await send("Network.setCookie", { name: "foo", value: "bar", url: "https://example.com/" }, sessionId);
  console.log("[A1] Network.clearBrowserCookies with params:{}");
  try {
    await send("Network.clearBrowserCookies", {}, sessionId);
    console.log("     PASS");
  } catch (e) {
    console.log(`     BUG  -> code=${e.code} message=${JSON.stringify(e.message)}`);
    bugs++;
  }

  await send("Network.setCookie", { name: "a", value: "1", url: "https://example.com/" }, sessionId);
  await send("Network.setCookie", { name: "b", value: "2", url: "https://other.test/" }, sessionId);
  console.log("[B3] Network.getAllCookies");
  try {
    const result = await send("Network.getAllCookies", {}, sessionId);
    const names = (result.cookies || []).map((c) => `${c.name}@${c.domain}`).join(", ");
    console.log(`     PASS -> ${result.cookies.length} cookie(s): ${names}`);
  } catch (e) {
    console.log(`     BUG  -> code=${e.code} message=${JSON.stringify(e.message)}`);
    bugs++;
  }

  ws.close();
  process.exit(bugs > 0 ? 1 : 0);
}

main().catch((err) => { console.error("Unexpected error:", err); process.exit(2); });
```

**Run**: `bash repro.sh`

- **Today** (`1.0.0-nightly.5812+b3257754` = main `b3257754`):
  ```
  [A1] Network.clearBrowserCookies with params:{}
       BUG  -> code=-31998 message="InvalidParams"
  [B3] Network.getAllCookies
       BUG  -> code=-31998 message="UnknownMethod"
  exit 1
  ```
- **Expected after fix**:
  ```
  [A1] Network.clearBrowserCookies with params:{}
       PASS
  [B3] Network.getAllCookies
       PASS -> 2 cookie(s): a@example.com, b@other.test
  exit 0
  ```

## Likely fix location

`src/cdp/domains/network.zig`:

- `clearBrowserCookies` — drop the inverted-logic param guard so the function follows the same "ignore unknown params" convention used by every other handler in this file.
- `processMessage` — add `getAllCookies` to the dispatch enum + a small handler that returns the jar via `CdpStorage.CookieWriter` (mirrors `Storage.getCookies` minus the `browserContextId` filter, since `Network.*` is already scoped to the current browser context).

## Environment

- Lightpanda: `1.0.0-nightly.5812+b3257754` (= `main` HEAD `b3257754` at filing time)
- OS: macOS aarch64
- CDP client: raw `ws` / Node 20

navidemad:
## What

Two small `Network.*` cookie-management fixes:

1. `Network.clearBrowserCookies` no longer rejects an empty `params: {}` object — most CDP clients (chrome-remote-interface, chromedp, raw websocket clients) include `"params":{}` on every command and Chrome silently accepts it.
2. `Network.getAllCookies` is now implemented. Previously it returned `UnknownMethod`, leaving clients no `Network.*` path to enumerate the in-memory jar across origins.

Closes #2254.

## Root cause

`src/cdp/domains/network.zig` `clearBrowserCookies` had an inverted-logic guard:

```zig
if (try cmd.params(struct {}) != null) return error.InvalidParams;
```

`cmd.params(T)` returns `null` only when the JSON message has no `params` field at all. With `"params":{}` present it parses successfully and returns `Some({})`, so the `!= null` branch fires and the jar is left untouched. Spec-conformant clients hit this on every call.

`getAllCookies` was simply absent from the `Network` dispatch enum, so `stringToEnum` returned `null` and the method dispatched to `error.UnknownMethod`.

```mermaid
flowchart LR
    A["Network.clearBrowserCookies<br/>(empty params object)"] --> B["guard: params != null fires"]
    B --> C["error -31998 InvalidParams"]
    D["Network.getAllCookies"] --> E["stringToEnum returns null"]
    E --> F["error -31998 UnknownMethod"]
    style B fill:#fdd
    style C fill:#fdd
    style E fill:#fdd
    style F fill:#fdd
```

## Fix

- `src/cdp/domains/network.zig` `clearBrowserCookies`: drop the inverted guard. Replaced with a comment explaining why an empty params object is accepted (CDP / JSON-RPC convention; matches sibling `Storage.clearCookies`).
- `src/cdp/domains/network.zig` `processMessage`: add `getAllCookies` to the action enum and switch.
- `src/cdp/domains/network.zig` new function `getAllCookies`: returns every cookie in the jar via the existing `CdpStorage.CookieWriter`, mirroring `Storage.getCookies` minus the `browserContextId` filter. No URL list, no origin scoping.

```mermaid
flowchart LR
    A["Network.clearBrowserCookies<br/>(empty params object)"] --> B["clear jar"]
    B --> C["empty result"]
    D["Network.getAllCookies"] --> E["CookieWriter, no urls filter"]
    E --> F["result.cookies = full list"]
    style B fill:#dfd
    style C fill:#dfd
    style E fill:#dfd
    style F fill:#dfd
```

## Test

Two new test blocks in `src/cdp/domains/network.zig`:

- `test "cdp.Network: clearBrowserCookies accepts empty params object"` — sends `params: .{}` (the spec-conformant call that fails today), asserts success and an empty jar afterward.
- `test "cdp.Network: getAllCookies returns whole jar regardless of current origin"` — sets cookies on two distinct origins (`example.com`, `other.test`), then asserts both shapes — with `params: .{}` and without any `params` field — return both cookies.

Both tests are added alongside the existing `test "cdp.Network: cookies"` block (kept untouched). Verified with the reproducer from the linked issue: exits 1 on `main`, exits 0 with this commit.

## Notes

`Network.clearBrowserCookies` previously also crashed the WebSocket on pre-`v0.2.6` (#1821 stopped the crash by adding the missing `clearRetainingCapacity()` call). This patch fixes a separate regression — the surviving handler then refused any spec-conformant call. The `Storage.clearCookies` and `Storage.getCookies` handlers already accept an empty params object via `(try cmd.params(T)) orelse T{}`; this brings `Network.clearBrowserCookies` in line.

### DIFF
--- a/src/cdp/domains/network.zig
+++ b/src/cdp/domains/network.zig
@@ -600,12 +600,11 @@ test "cdp.Network: clearBrowserCookies accepts empty params object" {
 
     // Most CDP clients (chrome-remote-interface, chromedp, etc.) always include
     // a `params` field on every command, even for methods that take none.
-    // Chrome ignores the empty object; we should too.
-    try ctx.processMessage(.{
-        .id = 2,
-        .method = "Network.clearBrowserCookies",
-        .params = .{},
-    });
+    // Chrome ignores the empty object; we should too. Sent as raw JSON because
+    // an empty Zig anonymous struct serializes as `[]`, not `{}`.
+    try ctx.processMessage(
+        \\{"id":2,"method":"Network.clearBrowserCookies","params":{}}
+    );
     try ctx.expectSentResult(null, .{ .id = 2 });
 
     try ctx.processMessage(.{
@@ -639,11 +638,11 @@ test "cdp.Network: getAllCookies returns whole jar regardless of current origin"
     });
     try ctx.expectSentResult(null, .{ .id = 1 });
 
-    try ctx.processMessage(.{
-        .id = 2,
-        .method = "Network.getAllCookies",
-        .params = .{},
-    });
+    // Empty params object — sent as raw JSON because an empty Zig anonymous
+    // struct serializes as `[]`, not `{}`.
+    try ctx.processMessage(
+        \\{"id":2,"method":"Network.getAllCookies","params":{}}
+    );
     try ctx.expectSentResult(.{
         .cookies = &[_]ResCookie{
             .{ .name = "a", .value = "1", .domain = "example.com", .path = "/", .size = 2, .secure = true },
--- a/src/cdp/domains/network.zig
+++ b/src/cdp/domains/network.zig
@@ -45,6 +45,7 @@ pub fn processMessage(cmd: *CDP.Command) !void {
         setCookie,
         setCookies,
         getCookies,
+        getAllCookies,
         getResponseBody,
     }, cmd.input.action) orelse return error.UnknownMethod;
 
@@ -59,6 +60,7 @@ pub fn processMessage(cmd: *CDP.Command) !void {
         .setCookie => return setCookie(cmd),
         .setCookies => return setCookies(cmd),
         .getCookies => return getCookies(cmd),
+        .getAllCookies => return getAllCookies(cmd),
         .getResponseBody => return getResponseBody(cmd),
     }
 }
@@ -149,7 +151,10 @@ fn deleteCookies(cmd: *CDP.Command) !void {
 }
 
 fn clearBrowserCookies(cmd: *CDP.Command) !void {
-    if (try cmd.params(struct {}) != null) return error.InvalidParams;
+    // Network.clearBrowserCookies takes no parameters per the CDP spec, but most
+    // CDP clients (chrome-remote-interface, chromedp, custom websocket clients)
+    // include an empty `"params":{}` object on every command for ergonomics.
+    // Chrome accepts that and clears the jar; reject only on truly malformed JSON.
     const bc = cmd.browser_context orelse return error.BrowserContextNotLoaded;
     bc.session.cookie_jar.clearRetainingCapacity();
     return cmd.sendResult(null, .{});
@@ -202,6 +207,18 @@ fn getCookies(cmd: *CDP.Command) !void {
     var jar = &bc.session.cookie_jar;
     jar.removeExpired(null);
     const writer = CdpStorage.CookieWriter{ .cookies = jar.cookies.items, .urls = urls.items };
+    try cmd.sendResult(.{ .cookies = writer }, .{});
+}
+
+fn getAllCookies(cmd: *CDP.Command) !void {
+    // Returns every cookie in the jar regardless of the current frame's origin.
+    // Mirrors Chrome's Network.getAllCookies and Storage.getCookies (without
+    // the latter's browserContextId filter, since Network commands are scoped
+    // to the current browser context already).
+    const bc = cmd.browser_context orelse return error.BrowserContextNotLoaded;
+    var jar = &bc.session.cookie_jar;
+    jar.removeExpired(null);
+    const writer = CdpStorage.CookieWriter{ .cookies = jar.cookies.items };
     try cmd.sendResult(.{ .cookies = writer }, .{});
 }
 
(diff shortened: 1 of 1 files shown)


karlseguin:
Looks good, thanks.

PR #2259, opened 2026-04-27
page: replay POST method/body/header on Page.reload

### LINKED ISSUE #2258: page: Page.reload regresses POST navigations to GET
## Summary

CDP `Page.reload` always issues a GET to the current URL, even when the prior navigation was POST. Chrome replays the original request — same method, same body, same `Content-Type` — so a form-submitted page stays form-submitted across reload. Lightpanda silently regresses to GET, so any handler that mutates state on POST never re-runs and the second request looks nothing like the first.

## Today's behavior

`src/cdp/domains/page.zig` `doReload` builds a `NavigateOpts` with only `url` and `kind = .reload`; `method`, `body`, and `header` default to `.GET`/`null`/`null`. The frame's `_navigated_options` records the prior navigation's `method` but never carries the body or content-type header, and `doReload` never reads back any of those fields — so the reload always becomes a GET regardless of how the page was loaded.

```mermaid
sequenceDiagram
    participant C as CDP Client
    participant LP as Lightpanda
    participant S as HTTP server
    C->>LP: form submit (POST /echo, body="key=value")
    LP->>S: POST /echo<br/>Content-Type: x-www-form-urlencoded<br/>key=value
    S-->>LP: 200 "method=POST"
    LP-->>C: page rendered with "method=POST"
    C->>LP: Page.reload
    LP->>LP: doReload reads frame.url only
    LP->>S: GET /echo
    Note over LP,S: prior method/body/header dropped
    S-->>LP: 200 "method=GET"
    LP-->>C: page now reads "method=GET" (regression)
```

## Expected behavior

The HTML Living Standard's [reload navigation](https://html.spec.whatwg.org/multipage/browsing-the-web.html#reload) re-fetches the document that produced the current entry. In Chrome, both `Page.reload` and the user-driven F5/Ctrl-R replay the same HTTP request that produced the current document — including method, body, and the relevant request headers. POST-loaded pages stay POST-loaded after reload.

```mermaid
sequenceDiagram
    participant C as CDP Client
    participant LP as Lightpanda
    participant S as HTTP server
    C->>LP: form submit (POST /echo, body="key=value")
    LP->>S: POST /echo<br/>Content-Type: x-www-form-urlencoded<br/>key=value
    S-->>LP: 200 "method=POST"
    LP-->>C: page rendered with "method=POST"
    C->>LP: Page.reload
    LP->>LP: doReload replays prior method/body/header
    LP->>S: POST /echo<br/>Content-Type: x-www-form-urlencoded<br/>key=value
    S-->>LP: 200 "method=POST"
    LP-->>C: page still reads "method=POST" (correct)
```

## Reproducer

Self-contained: a tiny Python echo server, an HTML form fixture, and a CDP driver via `chrome-remote-interface`. Drop the four files into a directory and run `bash repro.sh`. Verified to exit `1` against `1.0.0-nightly.5812+b3257754` (homebrew, 2026-04-26) and current `main` HEAD `a578f4d6`.

`repro.html`:

```html
<!doctype html>
<html>
<head><meta charset="utf-8"><title>reload-replay-POST repro</title></head>
<body>
<form id="f" method="post" action="/echo">
  <input type="text" name="key" value="value">
  <button id="b" type="submit">submit</button>
</form>
</body>
</html>
```

`repro_server.py`:

```python
#!/usr/bin/env python3
"""Tiny HTTP echo server: serves repro.html on GET and echoes the request
method back as HTML for /echo."""

import http.server
import socketserver
import os
import sys

PORT = int(os.environ.get("HTTP_PORT", "9580"))
ROOT = os.path.dirname(os.path.abspath(__file__))


    def _serve_html(self, body):
        self.send_response(200)
        self.send_header("Content-Type", "text/html; charset=utf-8")
        self.end_headers()
        self.wfile.write(body)

    def do_GET(self):
        if self.path in ("/", "/repro.html"):
            with open(os.path.join(ROOT, "repro.html"), "rb") as f:
                self._serve_html(f.read())
        elif self.path == "/echo":
            self._serve_html(b"<html><body>method=GET</body></html>")
        else:
            self.send_error(404)

    def do_POST(self):
        if self.path == "/echo":
            length = int(self.headers.get("Content-Length", "0"))
            _ = self.rfile.read(length)  # discard body, echo method only
            self._serve_html(b"<html><body>method=POST</body></html>")
        else:
            self.send_error(404)

    def log_message(self, fmt, *args):
        sys.stderr.write("[server] " + (fmt % args) + "\n")


if __name__ == "__main__":
        httpd.allow_reuse_address = True
        sys.stderr.write(f"[server] listening on 127.0.0.1:{PORT}\n")
        httpd.serve_forever()
```

`repro.js`:

```js
// Reproducer: Page.reload should replay POST, not regress to GET.
const CDP = require("chrome-remote-interface");

const HTTP_PORT = process.env.HTTP_PORT || "9580";
const CDP_PORT = parseInt(process.env.CDP_PORT || "9222", 10);

async function readBody(client, sessionId) {
    "Runtime.evaluate",
    { expression: "document.body.innerText", returnByValue: true },
    sessionId,
  );
  return result.value || "";
}

const sleep = (ms) => new Promise((r) => setTimeout(r, ms));

(async () => {
  let client;
  try {
    // target + local are both required against Lightpanda's CDP endpoint:
    //   /json/list returns [] (skip target discovery via target:)
    //   /json/protocol returns 404 (use bundled descriptor via local:)

    const { targetId } = await client.send("Target.createTarget", { url: "about:blank" });
    const { sessionId } = await client.send("Target.attachToTarget", { targetId, flatten: true });

    await client.send("Page.enable", {}, sessionId);
    await client.send("Runtime.enable", {}, sessionId);

    await client.send(
      "Page.navigate",
      { url: `http://127.0.0.1:${HTTP_PORT}/repro.html` },
      sessionId,
    );
    await sleep(500);

    await client.send(
      "Runtime.evaluate",
      { expression: "document.getElementById('f').submit()" },
      sessionId,
    );
    await sleep(800);

    const afterPost = await readBody(client, sessionId);
    console.log(`after POST submit: ${JSON.stringify(afterPost)}`);
    if (!afterPost.includes("method=POST")) {
      console.error("FAIL: form submission did not produce a POST navigation.");
      await client.close();
      process.exit(2);
    }

    await client.send("Page.reload", {}, sessionId);
    await sleep(800);

    const afterReload = await readBody(client, sessionId);
    console.log(`after Page.reload:  ${JSON.stringify(afterReload)}`);

    await client.close();

    if (afterReload.includes("method=POST")) {
      console.log("PASS: Page.reload replayed POST.");
      process.exit(0);
    }
    if (afterReload.includes("method=GET")) {
      console.error("FAIL: Page.reload regressed to GET.");
      process.exit(1);
    }
    console.error(`FAIL: unexpected body: ${JSON.stringify(afterReload)}`);
    process.exit(1);
  } catch (e) {
    console.error("ERROR:", e && e.stack ? e.stack : e);
    process.exit(2);
  }
})();
```

`repro.sh`:

```bash
#!/usr/bin/env bash
set -euo pipefail

CDP_PORT="${CDP_PORT:-9222}"

# Kill any prior listeners — a stale lightpanda silently keeps answering CDP.

PIDS=()
trap cleanup EXIT

cd "$REPRO_DIR"

HTTP_PORT="$HTTP_PORT" python3 repro_server.py >server.log 2>&1 &
PIDS+=("$!")

PIDS+=("$!")

# Readiness probe: assert the listener is Lightpanda, not stale Chrome.
for _ in $(seq 1 50); do
  sleep 0.1
done

fi

exec node repro.js
```

**Run**: `bash repro.sh`
- Today: prints `after Page.reload:  "method=GET"` and exits 1.
- Expected: prints `after Page.reload:  "method=POST"` and exits 0.

## Likely fix location

`src/cdp/domains/page.zig` `doReload`. The frame already records the prior navigation's `method` in `_navigated_options`; extending that struct to also retain `body` and (the request) `header` would let `doReload` replay them. The captures need to happen in `cmd.arena` before `session.replacePage()` frees the old frame's arena.

## Environment

- Lightpanda: `1.0.0-nightly.5812+b3257754` (homebrew, 2026-04-26) and `main` HEAD `a578f4d6` (`1.0.0-dev.5789+a578f4d6`)
- OS: darwin / aarch64
- CDP client: `chrome-remote-interface` via Node 20

navidemad:
## What

Make CDP `Page.reload` replay the original request — same method, same body, same `Content-Type` header — instead of silently downgrading every reload to GET. Matches Chrome's behavior and the HTML "reload navigation" spec.

Closes #2258.

## Root cause

`src/cdp/domains/page.zig` `doReload` builds a `NavigateOpts` with only `url` and `kind = .reload`; `method`, `body`, and `header` default to `.GET` / `null` / `null`. The frame's `_navigated_options` only carried the prior `method`, never the body or content-type, so even reading those fields back wouldn't have helped — the reload had nothing to replay.

```mermaid
flowchart LR
    A[Page.reload] --> B[doReload reads frame.url only]
    B --> C[frame.navigate with default GET, body=null, header=null]
    C --> D[POST body and Content-Type lost; second request is a GET]
    style B fill:#fdd
    style C fill:#fdd
    style D fill:#fdd
```

## Fix

- `src/browser/Frame.zig` — `NavigatedOpts`: add `body: ?[]const u8` and `header: ?[:0]const u8`. The frame retains them across the request lifetime.
- `src/browser/Frame.zig` — `Frame.navigate`: dupe the incoming `opts.body` / `opts.header` into `self.arena` when assigning `_navigated_options`, so they outlive the transient request arena.
- `src/cdp/domains/page.zig` — `doReload`: dupe the prior `body` / `header` from `_navigated_options` into `cmd.arena` *before* `session.replacePage()` frees the old frame, then pass them — and the prior `method` — into the new frame's `navigate` call.
- `src/cdp/domains/page.zig` — new test `cdp.frame: reload replays POST navigation`: drives a POST navigation to `/echo_method`, asserts the page body says `method=POST`, calls `Page.reload`, asserts the page body still says `method=POST`.
- `src/testing.zig` — new `/echo_method` route in `testHTTPHandler` that echoes the request method back as HTML; used by the new test.

```mermaid
flowchart LR
    A[Page.reload] --> B[doReload captures url + method/body/header from _navigated_options into cmd.arena]
    B --> C[frame.navigate with replayed method, body, header]
    C --> D[Second request matches the first: POST stays POST, GET stays GET]
    style B fill:#dfd
    style C fill:#dfd
    style D fill:#dfd
```

## Test

- **Unit test**: `cdp.frame: reload replays POST navigation` (`src/cdp/domains/page.zig`). Fails on `main` — the post-reload body reads `method=GET`. Passes with this commit.
- **Reproducer** from #2258: exits `1` on `main` (`after Page.reload:  "method=GET"`) and exits `0` with this commit (`after Page.reload:  "method=POST"`). Same assertion at the CDP boundary, no Zig in the loop.
- **Full suite**: `mise exec -- zig build test $V8` — 486/486 pass.

## Notes

`Location.reload()` (the JS `window.location.reload()` API in `src/browser/webapi/Location.zig`) has the same gap and is **not** addressed by this PR; it routes through `scheduleNavigation`, not through `doReload`. A follow-up can read the same `_navigated_options` and pass `method`/`body`/`header` through `scheduleNavigationWithArena`. Keeping scope tight here so this stays a small, focused review.


src/cdp/domains/page.zig
```diff
@@ -333,8 +333,20 @@ fn doReload(cmd: *CDP.Command) !void {
...
     var frame = session.currentFrame() orelse return error.FrameNotLoaded;
 
-    // Dupe URL before replacePage() frees the old frame's arena.
+    // Capture URL plus the prior navigation's method/body/header before
+    // replacePage() frees the old frame's arena. Replaying the same HTTP
+    // method on reload matches Chrome's F5 behavior — POST navigations
+    // re-submit, GET navigations re-fetch.
     const reload_url = try cmd.arena.dupeZ(u8, frame.url);
+    const prev_nav = frame._navigated_options;
+    const prev_body: ?[]const u8 = if (prev_nav) |p|
```
karlseguin:
```zig
const prev_body: ?[]const u8, const prev_header: ?[:0]const u8 = blk: {
   const p = prev_nav orelse break :blk .{null, null};
   break blk: {....}
};
```

karlseguin (commented):
I think in `frameHeaderDoneCallback` we should do:
```zig
  if (response.redirectCount() > 0) {
      self._navigated_options.method = .GET;
      self._navigated_options.body = null;
      self._navigated_options.header = null;
}
```

This also doesn't cover all cases correctly, but it's a safer default. (We could implement the entire thing correctly by looking at the methods/redirect status codes..but that seems like a lot of work for rarer cases). Important thing is: safer default to NEVER submit form data after a redirect than to ALWAYS submit it.

src/cdp/domains/page.zig
navidemad: Adopted in `1b9e8ad4`. The single-`blk:` form is cleaner — only one `prev_nav` check, and it makes the body/header pairing obvious.

navidemad:
Done in `1b9e8ad4`. Reset `_navigated_options.{method,body,header}` in `frameHeaderDoneCallback` whenever `response.redirectCount() > 0` — kept the conservative "never replay POST after a redirect" framing rather than the spec-correct 301/302/303-vs-307/308 split. New test `cdp.frame: reload after POST→redirect drops the POST` drives `POST /redirect_to_echo` → 302 → `GET /echo_method`, calls `Page.reload`, asserts the second request is GET. The existing `cdp.frame: reload replays POST navigation` (no redirect involved) still replays POST as before.

Also merged `origin/main` into the branch (5b1452f1) to resolve the conflict with #2265's `cdp.frame: navigate inherits original fragment across redirect` — both tests now sit side-by-side in `page.zig`.

PR #2261, opened 2026-04-27
lp: add LP.handleJavaScriptDialog to drive confirm/prompt return values

### LINKED ISSUE #2260: cdp: no opt-in for clients to drive confirm()/prompt() return values
## Summary

There is no CDP method that lets a client influence the JS-side return value of `window.confirm()` or `window.prompt()`. The headless auto-dismiss values (`alert`→void, `confirm`→`false`, `prompt`→`null`) from PR #2085 are the only outcome regardless of what the client sends.

`Page.handleJavaScriptDialog` is wired into the dispatch enum but its handler unconditionally responds with `-32000 "No dialog is showing"` (commit `7208934b`). PR #2085 (closing #2082) wired up the `Page.javascriptDialogOpening` event but explicitly deferred the return-value override: *"Supporting interactive accept/dismiss (where confirm() returns the CDP client's choice) would require suspending the JS runtime, which is a larger change that could follow separately."* This issue tracks that follow-up.

## Today's behavior

`src/cdp/domains/page.zig` `handleJavaScriptDialog` parses params then calls `cmd.sendError(-32000, "No dialog is showing", .{})`. `Window.alert` / `Window.confirm` / `Window.prompt` in `src/browser/webapi/Window.zig` dispatch a `javascript_dialog_opening` notification and immediately return their hardcoded auto-dismiss value. Nothing connects any CDP method to the JS return path.

```mermaid
sequenceDiagram
    participant CDP as Lightpanda CDP
    participant V8 as V8 / Window.zig
    Client->>CDP: Page.handleJavaScriptDialog<br/>{accept: true, promptText: "secret"}
    CDP-->>Client: {error: -32000<br/>"No dialog is showing"}
    Note over Client,CDP: handler always errors —<br/>nothing is ever stored
    Client->>V8: Runtime.evaluate<br/>"confirm('proceed?')"
    V8->>CDP: dispatch(javascript_dialog_opening)
    CDP-->>Client: Page.javascriptDialogOpening event
    V8-->>Client: result.value = false<br/>(headless auto-dismiss)
```

## Expected behavior

A CDP client should be able to drive `confirm()` / `prompt()` return values. In Chrome the dialog actually blocks JS and `Page.handleJavaScriptDialog` completes it; in a headless engine without runtime suspension the equivalent semantics are a *pre-armed response* — the client sets `{accept, promptText}` *before* triggering the JS that opens the dialog, and the engine returns that value when the dialog fires.

When no pre-armed response is queued, the headless auto-dismiss values from PR #2085 must be preserved (`confirm`→`false`, `prompt`→`null`, `alert`→void), so existing callers that don't issue the new method see no behavior change.

## Resolution: `LP.handleJavaScriptDialog` (PR #2261)

Per [maintainer feedback on PR #2261](https://github.com/lightpanda-io/browser/pull/2261#issuecomment-4339852197), the proactive pre-arm flow is exposed under the Lightpanda namespace as **`LP.handleJavaScriptDialog`** rather than repurposing `Page.handleJavaScriptDialog`. This keeps reactive Chrome-style drivers undisturbed (the standard `Page.*` method continues to return `-32000`) and matches the precedent of `LP.getInteractiveElements`, `LP.waitForSelector`, etc.

```mermaid
sequenceDiagram
    participant CDP as Lightpanda CDP
    participant V8 as V8 / Window.zig
    Client->>CDP: LP.handleJavaScriptDialog<br/>{accept: true, promptText: "secret"}
    Note over CDP: store {accept, promptText}<br/>on BrowserContext
    CDP-->>Client: {result: {}}
    Client->>V8: Runtime.evaluate<br/>"prompt('name?')"
    V8->>CDP: dispatch(javascript_dialog_opening)<br/>(passes &response output param)
    Note over CDP: pop pending →<br/>response.{accept,prompt_text}
    CDP-->>Client: Page.javascriptDialogOpening event
    V8-->>Client: result.value = "secret"<br/>(driven by pre-arm)
```

`Page.handleJavaScriptDialog` remains the `-32000` stub. If V8 suspension lands later (the larger change deferred in #2082 / PR #2085), the same `BrowserContext` slot can back a Chrome-faithful reactive `Page.handleJavaScriptDialog` without conflicting with the `LP.*` opt-in.

## Reproducer

Self-contained: one HTML fixture, one shell script, one Node CDP driver via `chrome-remote-interface`. No browser-automation framework involved.

`repro.html`:

```html
<!doctype html>
<html>
  <head><title>handleJavaScriptDialog repro</title></head>
  <body>
    <p>handleJavaScriptDialog return-value override repro.</p>
  </body>
</html>
```

`repro.sh`:

```bash
#!/usr/bin/env bash
# Demonstrates that no CDP method can influence the JS return value of
# confirm() / prompt() in Lightpanda. Page.handleJavaScriptDialog
# responds -32000 "No dialog is showing" regardless of whether a dialog
# is open, so headless auto-dismiss values are the only outcome.
#
# Run:           ./repro.sh
# Expected today (bug):
#   FAIL stage 1: Page.handleJavaScriptDialog responded with an error:
#     -32000 No dialog is showing
#   exit 1
# Expected after PR #2261 (using LP.handleJavaScriptDialog):
#   OK    stage 1: LP.handleJavaScriptDialog responded with success.
#   OK    stage 2a: confirm() returned true after accept=true pre-arm.
#   OK    stage 2b: prompt() returned "hello" after promptText pre-arm.
#   OK    stage 3a: confirm() with no pre-arm returned false (default preserved).
#   OK    stage 3b: prompt() with no pre-arm returned null (default preserved).
#   PASS: handleJavaScriptDialog drives confirm/prompt return values; defaults preserved.
#   exit 0

set -euo pipefail

CDP_PORT="${CDP_PORT:-9222}"

PIDS=()
cleanup() {
}
trap cleanup EXIT

cd "$REPRO_DIR"
PIDS+=("$!")

PIDS+=("$!")

# Readiness probe: assert the listener is actually Lightpanda.
for _ in $(seq 1 50); do
  sleep 0.1
done

fi

exec node repro.js
```

`repro.js`:

```js

const HTTP_PORT = process.env.HTTP_PORT || '9580';
const CDP_PORT = parseInt(process.env.CDP_PORT || '9222', 10);

(async () => {
  let client;
  try {
    // target + local are both required against Lightpanda — /json/list returns
    // [] and /json/protocol returns 404 against the default discovery flow.

    // Stage 1: can any CDP method pre-arm a dialog response?
    // After PR #2261, swap the method name to 'LP.handleJavaScriptDialog'.
    try {
      await client.send('Page.handleJavaScriptDialog', { accept: true, promptText: 'secret' }, sessionId);
    } catch (e) {
      console.log('FAIL stage 1: Page.handleJavaScriptDialog responded with an error:');
      console.log('  ' + (e.response ? `${e.response.code} ${e.response.message}` : e.message));
      console.log('  → CDP clients cannot pre-arm dialog responses; confirm()/prompt() always auto-dismiss.');
      process.exit(1);
    }
    console.log('OK    stage 1: handler responded with success.');

    // Stage 2: does the pre-arm flow through to confirm()/prompt()?
    await client.send('Page.handleJavaScriptDialog', { accept: true }, sessionId);
    const c = await client.send('Runtime.evaluate', { expression: 'confirm("proceed?")', returnByValue: true }, sessionId);
    if (c.result && c.result.value === true) {
      console.log('OK    stage 2a: confirm() returned true after accept=true pre-arm.');
    } else {
      console.log(`FAIL stage 2a: confirm() returned ${JSON.stringify(c.result && c.result.value)} (expected true).`);
      process.exit(1);
    }

    await client.send('Page.handleJavaScriptDialog', { accept: true, promptText: 'hello' }, sessionId);
    const p = await client.send('Runtime.evaluate', { expression: 'prompt("name?")', returnByValue: true }, sessionId);
    if (p.result && p.result.value === 'hello') {
      console.log('OK    stage 2b: prompt() returned "hello" after promptText pre-arm.');
    } else {
      console.log(`FAIL stage 2b: prompt() returned ${JSON.stringify(p.result && p.result.value)} (expected "hello").`);
      process.exit(1);
    }

    // Stage 3: defaults preserved when no pre-arm is queued.
    const cd = await client.send('Runtime.evaluate', { expression: 'confirm("default?")', returnByValue: true }, sessionId);
    if (cd.result && cd.result.value === false) {
      console.log('OK    stage 3a: confirm() with no pre-arm returned false (default preserved).');
    } else {
      console.log(`FAIL stage 3a: confirm() default returned ${JSON.stringify(cd.result && cd.result.value)} (expected false).`);
      process.exit(1);
    }

    const pd = await client.send('Runtime.evaluate', { expression: 'prompt("default?")', returnByValue: true }, sessionId);
    if (pd.result && pd.result.value === null) {
      console.log('OK    stage 3b: prompt() with no pre-arm returned null (default preserved).');
    } else {
      console.log(`FAIL stage 3b: prompt() default returned ${JSON.stringify(pd.result && pd.result.value)} (expected null).`);
      process.exit(1);
    }

    console.log('PASS: handleJavaScriptDialog drives confirm/prompt return values; defaults preserved.');
    await client.close();
    process.exit(0);
  } catch (e) {
    process.exit(2);
  }
})();
```

**Run**: `bash repro.sh`

- Today (nightly.5812 / `main` HEAD `9fbade45`): prints `FAIL stage 1: Page.handleJavaScriptDialog responded with an error: -32000 No dialog is showing` and exits **1**.
- After PR #2261 (with the method name swapped to `LP.handleJavaScriptDialog`): prints all `OK` lines and `PASS:` and exits **0**.

## Environment

- Lightpanda: `1.0.0-nightly.5812+b3257754` (also reproducible on `main` at `9fbade45`)
- OS: darwin (aarch64)
- CDP client: `chrome-remote-interface` 0.33.x via Node 22

navidemad:
## What

Adds `LP.handleJavaScriptDialog`, a Lightpanda-namespaced opt-in that lets CDP clients influence the JS-side return value of `window.confirm()` / `window.prompt()` via a *pre-arm* model that fits Lightpanda's auto-dismiss architecture without runtime suspension.

`Page.handleJavaScriptDialog` is intentionally **unchanged** — it continues to return `-32000 "No dialog is showing"` from PR #2085, so reactive Chrome-style drivers see the same surface they do today.

Closes #2260.

## Why an `LP.*` method instead of repurposing `Page.handleJavaScriptDialog`?

Standard CDP drivers send `Page.handleJavaScriptDialog` *reactively* in response to a `Page.javascriptDialogOpening` event — Chrome blocks JS at the dialog, the client picks accept/dismiss, the runtime resumes. Lightpanda's dialogs auto-dismiss in headless mode (`window.alert/confirm/prompt` return immediately rather than blocking V8), so by the time the event reaches the client, JS has already returned. A reactive `Page.handleJavaScriptDialog` call has nothing left to influence — full Chrome-faithful behavior would require the V8 suspension that #2082 / PR #2085 explicitly deferred.

The earlier revision of this PR repurposed `Page.handleJavaScriptDialog` for proactive pre-arm. Per [maintainer feedback](https://github.com/lightpanda-io/browser/pull/2261#issuecomment-4331894497), that risks surprising reactive drivers — the next [unknown] dialog they trigger after a stale pre-arm would be blindly accepted. Moving the pre-arm to a Lightpanda-namespaced method keeps the standard `Page.*` surface conservative and makes the proactive flow an explicit opt-in — matching the precedent of `LP.getInteractiveElements`, `LP.waitForSelector`, `LP.clickNode`, etc.

## Behavior

- **`LP.handleJavaScriptDialog({accept, promptText})`** — stashes `{accept, promptText}` on the `BrowserContext` and returns `{}`. The next `window.alert/confirm/prompt` consumes the stash exactly once.
- **`Page.javascriptDialogOpening`** — still emitted on every dialog. The event listener pops the pre-armed response from the `BrowserContext` and writes it into the dispatch's output param so `confirm()` returns `accept` (`true` / `false`) and `prompt()` returns `promptText` (or `null` when `accept=false`).
- **`Page.handleJavaScriptDialog`** — returns `-32000 "No dialog is showing"` unchanged from PR #2085.
- **No pre-arm queued** — `confirm` → `false`, `prompt` → `null`, `alert` → void (PR #2085 defaults preserved).

## Flow

```mermaid
flowchart LR
    A["LP.handleJavaScriptDialog<br/>(accept, promptText)"] --> B[stash on BrowserContext]
    B --> C[result success]
    D["Window.confirm / prompt"] --> E[dispatch with &response]
    E --> F[event listener pops pending → response]
    F --> G[Page.javascriptDialogOpening event]
    D --> H[return response.accept / response.prompt_text]
    I["Page.handleJavaScriptDialog"] --> J["sendError -32000<br/>(unchanged from #2085)"]
    style B fill:#dfd
    style C fill:#dfd
    style F fill:#dfd
    style H fill:#dfd
    style J fill:#fff3cd
```

## Files

- **`src/cdp/domains/lp.zig`** — adds `handleJavaScriptDialog` to the LP action enum + dispatch + handler. Stashes `{accept, promptText}` on `BrowserContext` (string dup'd into `bc.arena` so it outlives the CDP message arena).
- **`src/cdp/domains/page.zig`** — `Page.handleJavaScriptDialog` reverted to the upstream `-32000` stub; `javascriptDialogOpening` event listener pops `bc.pending_dialog_response` and fills `event.response.*` so `Window.zig` returns the pre-armed value.
- **`src/cdp/CDP.zig`** — `BrowserContext` gains `pending_dialog_response: ?Notification.DialogResponse = null` slot.
- **`src/Notification.zig`** — `JavascriptDialogOpening` gains a `response: *DialogResponse` output param. New `DialogResponse` struct holds `{accept, prompt_text}`.
- **`src/browser/webapi/Window.zig`** — `alert` / `confirm` / `prompt` allocate a stack-local `DialogResponse`, pass `&response` into the dispatch, then return `response.accept` (confirm) / `response.prompt_text` (prompt). Default-zero `DialogResponse` preserves PR #2085's auto-dismiss values.

## Test

- **Unit tests** in `src/cdp/domains/lp.zig`:
  - `cdp.lp: handleJavaScriptDialog accepts/dismisses without an open dialog` — handler responds success, errors `NoBrowserContext` when no context, accepts `accept` / `promptText` / empty params.
  - `cdp.lp: handleJavaScriptDialog controls confirm/prompt/alert return values` — pre-armed `accept=true` makes `confirm()` return `true`; `accept=true` + `promptText="hello"` makes `prompt()` return `"hello"`; `accept=true` with no `promptText` returns `""` per CDP spec; `accept=false` makes `prompt()` return `null`; no pre-arm preserves `confirm→false` / `prompt→null` defaults; `alert()` consumes pending so a leaked pre-arm does not leak across dialogs. All `Page.javascriptDialogOpening` events still fire unchanged.
- Reverting the production handler back to a no-op fails both tests; restoring makes them pass.
- Full unit-test suite (497 tests) is green.

## Notes

- **Single-slot pending, not a queue.** A second `LP.handleJavaScriptDialog` overwrites the first if no dialog has fired in between. Sufficient for the common pattern (pre-arm immediately before the action that triggers exactly one dialog) and avoids the lifetime / drain-on-page-reset complexity of a queue.
- **Pre-arm vs Chrome's suspend-and-resume.** Chrome blocks JS at `confirm()`, awaits `Page.handleJavaScriptDialog`, then resumes; that pattern requires the V8 suspension that PR #2085 deferred. Pre-arm achieves the same observable outcome ("CDP client controls the return value") without the suspension machinery, and is the natural fit for Lightpanda's auto-dismiss architecture. If suspend-and-resume lands later, this state can be repurposed to default the response when the client doesn't intervene before a timeout — and `Page.handleJavaScriptDialog` could be wired to it without conflicting with `LP.handleJavaScriptDialog`.
- The default-zero `DialogResponse` (`accept: false`, `prompt_text: null`) intentionally matches PR #2085's auto-dismiss values, so callers that never issue `LP.handleJavaScriptDialog` see no behavior change.


navidemad:
The `demo-scripts` failure on the previous CI run ([job 73133487623](https://github.com/lightpanda-io/browser/actions/runs/24977602123/job/73133487623)) is a pre-existing flake in `puppeteer/frame.js`, not caused by this PR:

- Same `ReferenceError: frameLoad is not defined` at `HTMLIFrameElement.<anonymous>` hit unrelated branches recently — e.g. [run 24901977590](https://github.com/lightpanda-io/browser/actions/runs/24901977590) on `fix/bound-memory-growth-long-waits` (2026-04-24, no overlap with dialog code).
- Root cause is a parse-order race in `lightpanda-io/demo`'s `public/frames/index.html`: the `<iframe onload="frameLoad()" src="frame1.html">` is parsed before the `<script>` that defines `frameLoad`. When `frame1.html` loads faster than the parser advances to the script tag, the inline handler runs in a window scope where `frameLoad` is still undefined.
- Filed as lightpanda-io/demo#167 with a one-line fix (move the `<script>` block above the `<iframe>`).
- All other checks on this PR are green: `zig fmt`, `zig test`, `zig build release`, `cdp-and-hyperfine-bench`, `wba-demo-scripts`, `browser fetch`.

Pushed an empty commit to retrigger CI.

karlseguin:
Do you have a use-case where this helps? I assume typical CDP drivers only send `Page.handleJavaScriptDialog` in response to an `Page.javascriptDialogOpening` event. Doesn't this PR result in the next [unknown] dialog being blindly accepted?

navidemad:
The use-case is proactive ("set up the answer, then perform the action"), not reactive: test frameworks that wrap dialog interaction in a synchronous block, e.g. `accept_modal(:confirm) { click_button("Delete") }`. The framework registers the answer before triggering the action that calls `confirm()`. Without this PR there is no way for a CDP client to make `confirm()` return `true` or `prompt()` return a specific string — the auto-dismiss values from #2082 are the only outcome.

You are right about the typical reactive flow, and you are right that this PR's stash would pre-arm the next dialog if a standard reactive driver sent it. The reason I went proactive is structural: in Lightpanda there is no V8 suspension, so by the time `Page.javascriptDialogOpening` reaches the client the JS has already returned. A reactive `handleJavaScriptDialog` call after-the-fact has nothing to influence — Chrome's behavior needs the runtime suspension that #2085 deferred as a "larger change that could follow separately."

To address your concern about standard CDP drivers being surprised: would you prefer I move the pre-arm to `LP.handleJavaScriptDialog` (or another Lightpanda-namespaced method)? That keeps `Page.handleJavaScriptDialog` returning `-32000 No dialog is showing` so reactive drivers see no semantic change — only Lightpanda-aware callers opt in. There is existing precedent (`LP.getInteractiveElements`, `LP.waitForSelector`, `LP.clickNode`, etc.).

The other option is closing this PR and waiting on the V8-suspension work behind #2082 — that's the Chrome-faithful answer and I'm happy to go either way. Let me know which lane you'd prefer.

karlseguin:
Yes, moving it to the LP domain seems like a good middle ground.

navidemad:
Adopted in `8cc82d1d`. Moved the pre-arm to `LP.handleJavaScriptDialog`; `Page.handleJavaScriptDialog` reverts to the upstream `-32000 "No dialog is showing"` stub from #2085, so reactive Chrome-style drivers see the same surface they do today.

The `Page.javascriptDialogOpening` event listener still pops `BrowserContext.pending_dialog_response` and fills `event.response.*` — only the *setter* moved to LP. `Window.alert/confirm/prompt`, `BrowserContext.pending_dialog_response`, and `Notification.DialogResponse` are unchanged from the previous revision.

Merged `origin/main` into the branch in `fd2f26a0` (HTMLFrameSetElement work, no conflicts). Updated PR title + body and issue #2260 to reflect the LP-namespace resolution. The reproducer in #2260 still demonstrates the gap on `main` (exits 1 against `Page.handleJavaScriptDialog`); after this PR, swapping the method name to `LP.handleJavaScriptDialog` makes it exit 0.

Full unit-test suite (497 tests) green locally.

karlseguin:
Can you zig fmt src/cdp/domains/page.zig plz

PR #2264, opened 2026-04-27
forms: skip FormData entry for <select> with no selectedness candidate

### LINKED ISSUE #2262: forms: FormData appends a phantom entry for a <select> with zero options
## Summary

When constructing a `FormData` from a form that contains a `<select>` with zero `<option>` children (e.g. a placeholder dropdown that is populated lazily by JS), Lightpanda appends a single empty-string entry under the `<select>`'s `name`. Per the HTML Living Standard, a `<select>` should only contribute entries for options whose selectedness is true — with no options, that set is empty, so no entry should be appended.

## Today's behavior

Given:

```html
<form id="f">
  <select name="empty"></select>
  <input name="other" value="kept">
</form>
```

`new FormData(document.getElementById('f'))` yields:

| Call | Observed |
| --- | --- |
| `fd.has("empty")` | `true` |
| `fd.get("empty")` | `""` |
| `fd.getAll("empty")` | `[""]` |
| `[...fd.entries()]` | `[["empty",""],["other","kept"]]` |

The phantom `["empty",""]` entry comes from the singular-select fast path in `collectForm` (`src/browser/webapi/net/FormData.zig`), which always appends `select.getValue(frame)` for non-multiple selects — even when the `<select>` has no options to derive selectedness from.

```mermaid
sequenceDiagram
    participant Client as JS / CDP Client
    participant FD as FormData.collectForm
    participant Sel as Form.Select
    Client->>FD: new FormData(form)
    FD->>Sel: getMultiple()
    Sel-->>FD: false
    FD->>Sel: getValue(frame)
    Note over Sel: zero options -> returns ""
    Sel-->>FD: ""
    FD-->>Client: appends ("empty", "") unconditionally
```

## Expected behavior

[HTML Living Standard, "constructing the form data set", step 5.5](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#constructing-form-data-set):

> If the field is a `select` element, then for each `option` element in the `select` element's list of options whose selectedness is true and that is not disabled, append an entry to entry list with name and the value of the `option` element.

A non-multiple `<select>` derives its selectedness from an explicitly-selected option, falling back to [the first non-disabled option in tree order](https://html.spec.whatwg.org/multipage/form-elements.html#selectedness-setting-algorithm). With zero options (or only disabled options), no option is selected, so no entry should be appended. Chrome and Firefox both omit the entry in this case.

```mermaid
sequenceDiagram
    participant Client as JS / CDP Client
    participant FD as FormData.collectForm
    participant Sel as Form.Select
    Client->>FD: new FormData(form)
    FD->>Sel: any non-disabled option?
    Sel-->>FD: no (zero options)
    Note over FD: skip — no selectedness candidate
    FD-->>Client: appends only ("other", "kept")
```

## Reproducer

Self-contained: one HTML fixture + one shell script + a small Node CDP driver. Runs against a fresh `lightpanda serve` with a single `npm install --no-save chrome-remote-interface`.

`repro.html`:
```html
<!DOCTYPE html>
<html>
<head><title>formdata-empty-select-pending</title></head>
<body>
<form id="f">
  <select name="empty"></select>
  <input name="other" value="kept">
</form>
<script>
  const form = document.getElementById('f');
  const fd = new FormData(form);
  const result = {
    has_empty: fd.has('empty'),
    get_empty: fd.get('empty'),
    getAll_empty: fd.getAll('empty'),
    entries: Array.from(fd.entries()),
  };
  document.title = JSON.stringify(result);
</script>
</body>
</html>
```

`repro.js` (Node CDP driver):
```js

const HTTP_PORT = process.env.HTTP_PORT || '9580';
const CDP_PORT = parseInt(process.env.CDP_PORT || '9222', 10);

(async () => {
  let client;
  try {

      'Runtime.evaluate',
      { expression: 'document.title', returnByValue: true },
      sessionId,
    );
    const observed = JSON.parse(result.value);

    console.log('Observed FormData state:');
    console.log('  has("empty")    =', JSON.stringify(observed.has_empty));
    console.log('  get("empty")    =', JSON.stringify(observed.get_empty));
    console.log('  getAll("empty") =', JSON.stringify(observed.getAll_empty));
    console.log('  entries         =', JSON.stringify(observed.entries));

    const bug =
      observed.has_empty === true ||
      observed.get_empty !== null ||
      (Array.isArray(observed.getAll_empty) && observed.getAll_empty.length > 0) ||
      observed.entries.some(([k]) => k === 'empty');

    if (bug) {
      console.log('FAIL: a <select> with zero options leaked an entry into FormData.');
      await client.close();
      process.exit(1);
    }

    console.log('OK: empty <select> contributed no entries (matches HTML spec).');
    await client.close();
    process.exit(0);
  } catch (e) {
    process.exit(2);
  }
})();
```

`repro.sh`:
```bash
#!/usr/bin/env bash
set -euo pipefail

CDP_PORT="${CDP_PORT:-9222}"

PIDS=()
trap cleanup EXIT

cd "$REPRO_DIR"
PIDS+=("$!")

PIDS+=("$!")

for _ in $(seq 1 50); do
  sleep 0.1
done

fi

exec node repro.js
```

**Run**: `bash repro.sh`
- Today: prints `FAIL: a <select> with zero options leaked an entry into FormData.` and exits 1.
- Expected: prints `OK: empty <select> contributed no entries (matches HTML spec).` and exits 0.

## Likely fix location

`src/browser/webapi/net/FormData.zig` — `collectForm`. The non-multiple branch unconditionally falls through to `try list.append(arena, name, value)`. Guard it on whether the `<select>` has at least one non-disabled `<option>` before emitting; otherwise skip per the HTML spec.

## Environment

- Lightpanda: `1.0.0-nightly.5812+b3257754` (also reproduces on `main` HEAD `9fbade45`)
- OS: macOS aarch64 (Darwin 25.4.0)
- CDP client: `chrome-remote-interface` v0.33.x

navidemad:
## What

A `FormData` constructed from a form containing a `<select>` with zero `<option>` children leaks a phantom `(name, "")` entry. Per the HTML Living Standard ["constructing the form data set" step 5.5](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#constructing-form-data-set), a `<select>` should only contribute entries for options whose selectedness is true; with no options, that set is empty and nothing should be appended. This patch guards the singular-select fast path in `collectForm` on the presence of at least one non-disabled `<option>` and skips the entry otherwise.

Closes #2262.

## Root cause

In `src/browser/webapi/net/FormData.zig#collectForm`, the non-multiple branch unconditionally falls through to `try list.append(arena, name, value)` with `value = select.getValue(frame)`. `Form.Select.getValue` returns the empty string for a `<select>` with no options, so the entry is appended as `(name, "")` even though the HTML spec says no entry should be appended at all. The same shortcut also misfires when every option is `disabled`: the auto-selectedness rule in [HTML §form-elements§selectedness-setting-algorithm](https://html.spec.whatwg.org/multipage/form-elements.html#selectedness-setting-algorithm) only auto-selects the first **non-disabled** option, so an all-disabled `<select>` likewise has no selectedness candidate.

```mermaid
flowchart LR
    A[FormData ctor with form arg] --> B["collectForm: select.getMultiple() == false branch"]
    B --> C["append name, select.getValue() unconditionally"]
    C --> D["Phantom (name, &quot;&quot;) entry"]
    style B fill:#fdd
    style C fill:#fdd
    style D fill:#fdd
```

## Fix

- `src/browser/webapi/net/FormData.zig` — `collectForm`: in the non-multiple `<select>` branch, walk the children once before emitting; if there is no non-disabled `<option>` child, `continue` past the trailing `list.append` call so no entry is appended. The existing `getValue(frame)` fallback is preserved for the common case where at least one candidate exists.

```mermaid
flowchart LR
    A[FormData ctor with form arg] --> B["collectForm: select.getMultiple() == false branch"]
    B --> C{"any non-disabled &lt;option&gt; child?"}
    C -->|yes| D["append name, select.getValue()"]
    C -->|no| E[skip — no selectedness candidate]
    style B fill:#dfd
    style C fill:#dfd
    style D fill:#dfd
    style E fill:#dfd
```

## Test

- **Unit test**: `src/browser/tests/net/form_data.html` — three new fixture blocks (`selectWithoutOptions`, `selectMultipleWithoutOptions`, `selectAllOptionsDisabled`) covering single-select with zero options, multi-select with zero options, and single-select where every option is disabled. The existing `WebApi: FormData` runner picks them up automatically. All three fixtures assert `fd.has(name) === false`, `fd.get(name) === null`, `fd.getAll(name)` is empty, and the entry list contains no row for the `<select>`'s name.
- **Reproducer**: a self-contained `repro.html` + `repro.sh` + `repro.js` (Node CDP driver) drives `Page.navigate` against a `<select name="empty"></select>` fixture, reads back the FormData state, and asserts no `empty` key. Today it prints `FAIL: a <select> with zero options leaked an entry into FormData.` and exits 1; with this commit it prints `OK: empty <select> contributed no entries (matches HTML spec).` and exits 0. See the issue for the full scripts.
- Verified locally with `mise exec -- zig build test -Dprebuilt_v8_path=...` (487 of 487 tests pass) and end-to-end via the reproducer against `./zig-out/bin/lightpanda` built from this branch.

## Notes

The behavior matches Chrome and Firefox in interactive testing: both omit the entry for a zero-option or all-disabled `<select>`. No public API surface changes; `Form.Select.getValue` is left untouched (it still returns `""` for callers like `select.value` that document a string return type even when no option is selected).

### DIFF
--- a/src/browser/webapi/net/FormData.zig
+++ b/src/browser/webapi/net/FormData.zig
@@ -196,6 +196,17 @@ fn collectForm(arena: Allocator, form_: ?*Form, submitter_: ?*Element, frame: *F
 
             if (element.is(Form.Select)) |select| {
                 if (select.getMultiple() == false) {
+                    // Per the HTML spec, a single-select's selectedness comes
+                    // from an explicitly-selected option, falling back to the
+                    // first non-disabled option in tree order. With no options
+                    // (or only disabled ones), nothing is selected and no
+                    // entry is appended.
+                    var children = select.asNode().childrenIterator();
+                    const has_candidate = while (children.next()) |child| {
+                        const option = child.is(Form.Select.Option) orelse continue;
+                        if (!option.getDisabled()) break true;
+                    } else false;
+                    if (!has_candidate) continue;
                     break :blk select.getValue(frame);
                 }
 
--- a/src/browser/webapi/net/FormData.zig
+++ b/src/browser/webapi/net/FormData.zig
@@ -196,18 +196,12 @@ fn collectForm(arena: Allocator, form_: ?*Form, submitter_: ?*Element, frame: *F
 
             if (element.is(Form.Select)) |select| {
                 if (select.getMultiple() == false) {
-                    // Per the HTML spec, a single-select's selectedness comes
-                    // from an explicitly-selected option, falling back to the
-                    // first non-disabled option in tree order. With no options
-                    // (or only disabled ones), nothing is selected and no
-                    // entry is appended.
-                    var children = select.asNode().childrenIterator();
-                    const has_candidate = while (children.next()) |child| {
-                        const option = child.is(Form.Select.Option) orelse continue;
-                        if (!option.getDisabled()) break true;
-                    } else false;
-                    if (!has_candidate) continue;
-                    break :blk select.getValue(frame);
+                    // Per the HTML spec, a single-select with no selectedness
+                    // candidate (zero options or every option disabled)
+                    // contributes no entry. Otherwise emit the candidate's
+                    // value.
+                    const opt = select.effectiveOption() orelse continue;
+                    break :blk opt.getValue(frame);
                 }
 
                 var options = try select.getSelectedOptions(frame);
--- a/src/browser/tests/net/form_data.html
+++ b/src/browser/tests/net/form_data.html
@@ -813,6 +813,72 @@
 }
 </script>
 
+<script id=selectWithoutOptions>
+{
+  // Per HTML spec (constructing form data set, step 5.5), a <select> only
+  // appends entries for options whose selectedness is true. With zero options,
+  // nothing is selected and no entry is appended.
+  const form = document.createElement('form');
+
+  const sel = document.createElement('select');
+  sel.name = 'choice';
+  form.appendChild(sel);
+
+  const sibling = document.createElement('input');
+  sibling.name = 'sibling';
+  sibling.value = 'kept';
+  form.appendChild(sibling);
+
+  const fd = new FormData(form);
+  testing.expectEqual(false, fd.has('choice'));
+  testing.expectEqual(null, fd.get('choice'));
+  testing.expectEqual([], fd.getAll('choice'));
+  testing.expectEqual('kept', fd.get('sibling'));
+  testing.expectEqual([['sibling', 'kept']], Array.from(fd.entries()));
+}
+</script>
+
+<script id=selectMultipleWithoutOptions>
+{
+  // Same rule for <select multiple> — with zero options, no entry.
+  const form = document.createElement('form');
+
+  const sel = document.createElement('select');
+  sel.name = 'tags';
+  sel.multiple = true;
+  form.appendChild(sel);
+
+  const fd = new FormData(form);
+  testing.expectEqual(false, fd.has('tags'));
+  testing.expectEqual([], fd.getAll('tags'));
+  testing.expectEqual([], Array.from(fd.entries()));
+}
+</script>
+
+<script id=selectAllOptionsDisabled>
+{
+  // A single-select with only disabled options has no auto-selected option,
+  // so no entry is appended.
+  const form = document.createElement('form');
+
+  const sel = document.createElement('select');
+  sel.name = 'mode';
+
+  const opt = document.createElement('option');
+  opt.value = 'off';
+  opt.disabled = true;
+  opt.textContent = 'off';
+  sel.appendChild(opt);
+
+  form.appendChild(sel);
+
+  const fd = new FormData(form);
+  testing.expectEqual(false, fd.has('mode'));
+  testing.expectEqual(null, fd.get('mode'));
+  testing.expectEqual([], fd.getAll('mode'));
+}
+</script>
+
 <script id=formDataEventModifyFormData>
 {
   // Listeners can modify formData during the event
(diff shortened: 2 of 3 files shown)


src/browser/webapi/net/FormData.zig
```diff
@@ -196,6 +196,17 @@ fn collectForm(arena: Allocator, form_: ?*Form, submitter_: ?*Element, frame: *F
 
             if (element.is(Form.Select)) |select| {
                 if (select.getMultiple() == false) {
+                    // Per the HTML spec, a single-select's selectedness comes
+                    // from an explicitly-selected option, falling back to the
+                    // first non-disabled option in tree order. With no options
+                    // (or only disabled ones), nothing is selected and no
+                    // entry is appended.
+                    var children = select.asNode().childrenIterator();
```
karlseguin: What do you think of having a `effectiveOption` (or something, not sure about the name) to `Select` which returns a `?*Option`. Then both this code and `Select.getValue` can re-use the same logic, only differing in how they handle the null. Bonus, single iteration through the list.

navidemad: Adopted in `6e654877`. Extracted `Select.effectiveOption() ?*Option` and routed `Select.getValue` and `FormData.collectForm` through it — single iteration in each call site, and the selectedness-candidate algorithm now lives in one place.

PR #2267, opened 2026-04-27
forms: clamp <input type=range> value to min/max

### LINKED ISSUE #2266: forms: <input type=range> does not clamp value to min/max
## Summary

`<input type="range">` does not perform value sanitization against the `min` / `max` attributes. Assigning an out-of-range numeric value through the JS `value` setter leaves the assigned string in place rather than clamping it to the boundary — `el.value === "999"` instead of `"100"` for `<input type="range" min="0" max="100">`.

## Today's behavior

When a CDP client (or page-side JS) assigns a value to an `<input type="range">` element, `Input.zig`'s `sanitizeValue` only validates the input as a "valid floating-point number" string and otherwise passes it through verbatim. Both the min/max clamp **and** the spec-mandated default value (`min + (max - min) / 2`) are missing. A valid float that is far outside `[min, max]` survives unchanged.

```mermaid
sequenceDiagram
    participant Client as CDP / Page JS
    participant LP as Lightpanda
    participant Input as Input.zig
    Client->>LP: el.value = "999" (min="0" max="100")
    LP->>Input: sanitizeValue(value="999")
    Note over Input: .range branch only checks isValidFloatingPoint
    Input-->>LP: "999" (passed through)
    LP-->>Client: el.value === "999" (out of range, not clamped)
```

## Expected behavior

Per WHATWG HTML "input type=range" value sanitization algorithm (https://html.spec.whatwg.org/multipage/input.html#range-state-(type=range)):

> 1. If the value of the element is not a valid floating-point number, then set it to the best representation, as a floating-point number, of the default value (= the minimum plus half the difference between the minimum and the maximum).
> 2. If the value is less than the minimum, then set it to a string representing the minimum.
> 3. If the value is greater than the maximum, then set it to a string representing the maximum.

Defaults per the HTML spec: `min` defaults to `0`, `max` defaults to `100`. Step matching (snap to nearest step) is a separate sanitization rule and out of scope here.

Chrome behavior matches the spec: `<input type="range" min="0" max="100">; el.value = "999"` ⇒ `el.value === "100"`.

```mermaid
sequenceDiagram
    participant Client as CDP / Page JS
    participant LP as Lightpanda
    participant Input as Input.zig
    Client->>LP: el.value = "999" (min="0" max="100")
    LP->>Input: sanitizeValue(value="999")
    Note over Input: validate float, parse min/max, clamp
    Input-->>LP: "100" (clamped to max)
    LP-->>Client: el.value === "100"
```

## Reproducer

Self-contained: one HTML fixture + one shell script + one Node CDP driver. Runs against a fresh `lightpanda serve` with a single `npm install --no-save chrome-remote-interface`.

`repro.html`:
```html
<!DOCTYPE html>
<html>
<head><title>range-clamp repro</title></head>
<body>
<input id="r" type="range" min="0" max="100">
<script>
  const el = document.getElementById('r');
  el.value = "999";
  document.title = el.value;
</script>
</body>
</html>
```

`repro.sh`:
```bash
#!/usr/bin/env bash
set -euo pipefail

HTTP_PORT="${HTTP_PORT:-9582}"
CDP_PORT="${CDP_PORT:-9224}"

PIDS=()
trap cleanup EXIT

cd "$REPRO_DIR"
PIDS+=("$!")

PIDS+=("$!")

for _ in $(seq 1 50); do
  sleep 0.1
done

fi

exec node repro.js
```

`repro.js`:
```js

const HTTP_PORT = process.env.HTTP_PORT || '9582';
const CDP_PORT = parseInt(process.env.CDP_PORT || '9224', 10);

(async () => {
  let client;
  try {

      'Runtime.evaluate',
      { expression: 'document.title', returnByValue: true },
      sessionId,
    );

    const observed = result.value;
    const expected = '100';
    console.log(`assigned el.value = "999" with min="0" max="100"`);
    console.log(`document.title (= el.value after sanitization) = "${observed}"`);
    console.log(`expected per WHATWG spec = "${expected}"`);
    await client.close();
    process.exit(observed === expected ? 0 : 1);
  } catch (e) {
    process.exit(2);
  }
})();
```

**Run**: `bash repro.sh`

- Today: prints `document.title (= el.value after sanitization) = "999"` and exits 1.
- Expected: prints `document.title (= el.value after sanitization) = "100"` and exits 0.

## Likely fix location

`src/browser/webapi/element/html/Input.zig` — `sanitizeValue`, `.range` branch (around line 567). The branch currently runs `isValidFloatingPoint(value)` and falls through to `"50"` when invalid, but never reads the `min`/`max` attributes or compares the value against them. `getMin()` / `getMax()` already exist on `Input` and return the raw attribute or `""`. The fix can route `.range` through a dedicated helper that parses min/max (with spec defaults of 0 and 100), validates the value as a float, and applies the three clamp rules.

## Notes

Step matching (snap to nearest step) is a separate sanitization rule per the spec and is intentionally out of scope for this issue/PR — it would meaningfully expand surface area. Filing as a follow-up once min/max clamping lands.

## Environment

- Lightpanda: `1.0.0-nightly.5812+b3257754` (also confirmed against `main` HEAD `9fbade45`)
- OS: darwin (macOS aarch64)
- CDP client: `chrome-remote-interface` (latest)

navidemad:
## What

`<input type="range">` did not perform value sanitization against the `min` / `max` attributes. Setting `el.value = "999"` on `<input type="range" min="0" max="100">` left `el.value === "999"` instead of clamping to `"100"`. The previous `.range` branch in `sanitizeValue` only validated the value as a "valid floating-point number" and otherwise returned `"50"` (a hardcoded default that's only correct when min=0 / max=100).

This change implements the WHATWG HTML "input type=range" value sanitization algorithm: clamp valid floats to `[min, max]` (defaults 0 and 100) and fall back to `min + (max - min) / 2` when the assigned value isn't a valid floating-point number.

Closes #2266.

## Root cause

`sanitizeValue` in `src/browser/webapi/element/html/Input.zig` had a single-line `.range` branch that only checked `isValidFloatingPoint(value)`. It never read `min` / `max` and never applied the three clamp rules from https://html.spec.whatwg.org/multipage/input.html#range-state-(type=range). The fall-through default was hardcoded to `"50"` rather than the spec's `min + (max - min) / 2` midpoint.

```mermaid
flowchart LR
    A[value setter] --> B[.range: isValidFloatingPoint?]
    B -->|valid| C[return value as-is, no clamp]
    B -->|invalid| D[return hardcoded '50']
    style C fill:#fdd
    style D fill:#fdd
```

## Fix

Route the `.range` branch through a new `sanitizeRange` helper that parses `min` / `max` (with spec defaults of 0 and 100, applied when the attribute is absent or not a valid floating-point number), validates the value, and applies the three sanitization rules in order. A small `formatFloat` helper handles "best representation" formatting via `std.fmt.bufPrint("{d}", .{value})`, which already produces canonical decimal output (`100.0 -> "100"`, `0.5 -> "0.5"`, `0.0 -> "0"`).

- `src/browser/webapi/element/html/Input.zig` — `sanitizeValue` `.range` branch: delegates to `sanitizeRange`.
- `src/browser/webapi/element/html/Input.zig` — `sanitizeRange`: new helper implementing the WHATWG algorithm.
- `src/browser/webapi/element/html/Input.zig` — `formatFloat`: arena-allocated `f64`-to-string helper.
- `src/browser/tests/element/html/input-attrs.html` — new `range-clamp` script block with seven sub-cases (above max, below min, in range, negative bounds, fractional bounds, default min/max, non-numeric default midpoint).

```mermaid
flowchart LR
    A[value setter] --> B[.range: sanitizeRange]
    B --> C[parse min/max, default 0/100]
    C --> D[validate float]
    D -->|invalid| E[return min+&#40;max-min&#41;/2]
    D -->|valid| F[clamp to &#91;min, max&#93;]
    style B fill:#dfd
    style E fill:#dfd
    style F fill:#dfd
```

## Test

- **Unit test**: `WebApi: HTML.Input` (the `range-clamp` block in `src/browser/tests/element/html/input-attrs.html`). Fails on `main` with `expected '100' got '999'`; passes with this commit. Full suite: `487 of 487 tests passed`.
- **Reproducer**: the `repro.sh` from #2266 exits 1 against the current nightly (`1.0.0-nightly.5812+b3257754`), exits 0 against this branch's local debug build. The script asserts that assigning `el.value = "999"` to `<input type="range" min="0" max="100">` ends up reading `"100"` (clamped) rather than `"999"` (passed through).

## Notes

Step matching (snap to nearest valid step) is a separate sanitization rule from the same spec section and is intentionally out of scope here — it would meaningfully expand surface area. Filing as a follow-up once min/max clamping lands.

### DIFF
--- a/src/browser/tests/element/html/input-attrs.html
+++ b/src/browser/tests/element/html/input-attrs.html
@@ -129,8 +129,10 @@
     testing.expectEqual('1.5', r3.value);
     r3.value = '0';
     testing.expectEqual('0.5', r3.value);
-    r3.value = '1';
-    testing.expectEqual('1', r3.value);
+    // Note: in-range pass-through under clamping alone is exercised by r1/r4.
+    // An assertion like `r3.value = '1' -> '1'` would conflict with WHATWG step
+    // matching (default step=1, base=min=0.5 -> nearest valid is '1.5'), which
+    // is intentionally out of scope for this PR; tracking separately.
 
     // Default min/max (0..100) when attributes absent
     const r4 = document.createElement('input');
--- a/src/browser/tests/element/html/input-attrs.html
+++ b/src/browser/tests/element/html/input-attrs.html
@@ -85,3 +85,75 @@
     testing.expectEqual('', i3.autocomplete);
   }
 </script>
+
+<script id="range-clamp">
+  {
+    // Per WHATWG HTML spec, value sanitization for `type=range` clamps to
+    // [min, max] with defaults of 0 and 100.
+    // https://html.spec.whatwg.org/multipage/input.html#range-state-(type=range)
+    const r = document.createElement('input');
+    r.type = 'range';
+    r.min = '0';
+    r.max = '100';
+
+    // Above max -> max
+    r.value = '999';
+    testing.expectEqual('100', r.value);
+
+    // Below min -> min
+    r.value = '-10';
+    testing.expectEqual('0', r.value);
+
+    // In range -> unchanged
+    r.value = '50';
+    testing.expectEqual('50', r.value);
+
+    // Negative range
+    const r2 = document.createElement('input');
+    r2.type = 'range';
+    r2.min = '-50';
+    r2.max = '50';
+    r2.value = '0';
+    testing.expectEqual('0', r2.value);
+    r2.value = '100';
+    testing.expectEqual('50', r2.value);
+    r2.value = '-100';
+    testing.expectEqual('-50', r2.value);
+
+    // Fractional bounds
+    const r3 = document.createElement('input');
+    r3.type = 'range';
+    r3.min = '0.5';
+    r3.max = '1.5';
+    r3.value = '2';
+    testing.expectEqual('1.5', r3.value);
+    r3.value = '0';
+    testing.expectEqual('0.5', r3.value);
+    r3.value = '1';
+    testing.expectEqual('1', r3.value);
+
+    // Default min/max (0..100) when attributes absent
+    const r4 = document.createElement('input');
+    r4.type = 'range';
+    r4.value = '999';
+    testing.expectEqual('100', r4.value);
+    r4.value = '-1';
+    testing.expectEqual('0', r4.value);
+
+    // Non-numeric value falls back to spec default `min + (max - min) / 2`
+    const r5 = document.createElement('input');
+    r5.type = 'range';
+    r5.min = '0';
+    r5.max = '100';
+    r5.value = 'garbage';
+    testing.expectEqual('50', r5.value);
+
+    // Spec default is the midpoint, not a hardcoded "50"
+    const r6 = document.createElement('input');
+    r6.type = 'range';
+    r6.min = '20';
+    r6.max = '40';
+    r6.value = 'garbage';
+    testing.expectEqual('30', r6.value);
+  }
+</script>
--- a/src/browser/webapi/element/html/Input.zig
+++ b/src/browser/webapi/element/html/Input.zig
@@ -564,7 +564,7 @@ fn sanitizeValue(self: *Input, comptime dupe: bool, value: []const u8, frame: *F
         .time => return if (isValidTime(value)) if (comptime dupe) try frame.dupeString(value) else value else "",
         .@"datetime-local" => return try sanitizeDatetimeLocal(dupe, value, frame.arena),
         .number => return if (isValidFloatingPoint(value)) if (comptime dupe) try frame.dupeString(value) else value else "",
-        .range => return if (isValidFloatingPoint(value)) if (comptime dupe) try frame.dupeString(value) else value else "50",
+        .range => return try sanitizeRange(dupe, value, self.getMin(), self.getMax(), frame),
         .color => {
             if (value.len == 7 and value[0] == '#') {
                 var needs_lower = false;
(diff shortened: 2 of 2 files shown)


src/browser/tests/element/html/input-attrs.html
```diff
@@ -85,3 +85,75 @@
...
+    const r3 = document.createElement('input');
+    r3.type = 'range';
+    r3.min = '0.5';
+    r3.max = '1.5';
+    r3.value = '2';
+    testing.expectEqual('1.5', r3.value);
+    r3.value = '0';
+    testing.expectEqual('0.5', r3.value);
+    r3.value = '1';
+    testing.expectEqual('1', r3.value);
```
karlseguin: In Firefox, this is `1.5`

src/browser/webapi/element/html/Input.zig
```diff
@@ -786,6 +786,48 @@ fn sanitizeDatetimeLocal(comptime dupe: bool, value: []const u8, arena: std.mem.
...
+    const v = std.fmt.parseFloat(f64, value) catch unreachable; // grammar already validated
+    if (v < min) return try formatFloat(frame.arena, min);
+    if (v > max) return try formatFloat(frame.arena, max);
+    return if (comptime dupe) try frame.dupeString(value) else value;
+}
+
+/// Format an f64 to its shortest decimal representation, arena-allocated.
+fn formatFloat(arena: std.mem.Allocator, value: f64) ![]const u8 {
+    var buf: [64]u8 = undefined;
+    const formatted = std.fmt.bufPrint(&buf, "{d}", .{value}) catch unreachable;
```
karlseguin: Is there a reason to prefer this over:

```zig
return std.fmt.allocPrint(arena, "{d}', .{value});
```

src/browser/tests/element/html/input-attrs.html
navidemad: Good catch — that's the missing step-matching rule from the same spec section (default `step=1`, base = `min` = `0.5`, so the valid ladder is `0.5, 1.5`; `'1'` is exactly between and ties round up to `'1.5'`). Step matching was deliberately scoped out of this PR — tracking it as a follow-up at #2277.

Dropped the `r3.value = '1' -> '1'` assertion in `fecab22e` so the test doesn't pin a value that'll need to change once #2277 lands. The other fractional-bounds cases (`'2' -> '1.5'`, `'0' -> '0.5'`) still cover the clamping rule.

src/browser/webapi/element/html/Input.zig
navidemad: Adopted in `fecab22e`. No good reason — `allocPrint` is the cleaner one-liner and removes the `catch unreachable` foot-gun from `bufPrint`.

PR #2269, opened 2026-04-27
selector: decode CSS escape sequences inside quoted attribute values

### LINKED ISSUE #2268: selector: decode CSS escape sequences inside quoted attribute values
## Summary

`document.querySelector` and `querySelectorAll` do not decode CSS escape sequences inside quoted attribute selector values. A selector like `[data-x="abc\\def"]` is supposed to match an element whose attribute value is the 7-byte string `abc\def` (one literal backslash), but Lightpanda keeps the escape un-decoded and matches against the literal 8-byte string `abc\\def` instead — so the call returns `null` even though the element is present in the tree.

The escape decoder for `#id` / `.class` selectors landed in #1350; `[attr="..."]` was missed.

## Today's behavior

`Parser.attributeValue` in `src/browser/webapi/selector/Parser.zig` reads a quoted attribute value with `std.mem.indexOfScalarPos` to find the closing quote and returns the raw byte slice between the quotes. Backslash escapes are neither honored as escape sequences nor decoded, so:

1. `[data-x="foo\"bar"]` is read as the value `foo\` (the parser stops at the first `"` it sees, ignoring the `\` in front of it).
2. `[data-x="abc\\def"]` returns the 8-byte slice `abc\\def` to the matcher; the element's actual attribute value is the 7-byte string `abc\def`, so they never compare equal.
3. Hex escapes like `\41` and Unicode codepoint escapes never decode.

```mermaid
sequenceDiagram
    participant Client as CDP / JS Client
    participant Parser as Parser.attributeValue
    participant Matcher as Attribute matcher
    Client->>Parser: querySelector('p[data-x="abc\\\\def"]')
    Parser->>Parser: indexOfScalarPos(input, 1, '"')<br/>finds closing quote, ignores backslashes
    Parser-->>Matcher: value = "abc\\def" (8 bytes, two literal backslashes)
    Matcher-->>Client: null (no element has attribute value "abc\\def")
    Note over Parser: src/browser/webapi/selector/Parser.zig<br/>fn attributeValue
```

## Expected behavior

Per [CSS Syntax Module Level 3 §4.3.5 "Consume a string token"](https://drafts.csswg.org/css-syntax/#consume-string-token), the tokenizer must walk the quoted string and decode escape sequences:

- `\\` -> `\`
- `\"` -> `"` (when in a `"..."` string)
- `\'` -> `'` (when in a `'...'` string)
- `\<hex digits>` -> Unicode code point (1–6 hex digits, optional whitespace terminator)
- `\<newline>` -> empty (line continuation)
- a bare newline inside the string is a parse error

After decoding, the matcher receives the literal value the author intended, so `[data-x="abc\\def"]` correctly matches the element with `data-x="abc\def"`.

```mermaid
sequenceDiagram
    participant Client as CDP / JS Client
    participant Parser as Parser.attributeValue
    participant Matcher as Attribute matcher
    Client->>Parser: querySelector('p[data-x="abc\\\\def"]')
    Parser->>Parser: walk string, decode \\ -> \, \" -> ", \<hex> -> codepoint
    Parser-->>Matcher: value = "abc\def" (7 bytes, one literal backslash)
    Matcher-->>Client: <p data-x="abc\def">hit</p>
```

## Reproducer

Self-contained: one HTML fixture + one shell script. Drives `lightpanda serve` over CDP using `chrome-remote-interface`. No other dependencies beyond `node` and `python3`.

`repro.html`:
```html
<!doctype html>
<meta charset="utf-8">
<title>attribute-value escape decode repro</title>
<p data-random="abc\def">hit</p>
```

`repro.js`:
```js

const HTTP_PORT = process.env.HTTP_PORT || '9584';
const CDP_PORT = parseInt(process.env.CDP_PORT || '9226', 10);

(async () => {
  let client;
  try {
    // target + local are both required against Lightpanda: /json/list returns []
    // and /json/protocol is not served, so chrome-remote-interface's defaults fail.

    await new Promise((r) => setTimeout(r, 600));

    const expression = "document.querySelector('p[data-random=\"abc\\\\\\\\def\"]')?.textContent ?? null";
      'Runtime.evaluate',
      { expression, returnByValue: true },
      sessionId,
    );

    if (result.value === 'hit') {
      console.log('PASS: querySelector matched the element with the literal backslash value.');
      await client.close();
      process.exit(0);
    }
    console.log(`FAIL: expected "hit", got ${JSON.stringify(result.value)}.`);
    await client.close();
    process.exit(1);
  } catch (e) {
    process.exit(2);
  }
})();
```

`repro.sh`:
```bash
#!/usr/bin/env bash
set -euo pipefail

HTTP_PORT="${HTTP_PORT:-9584}"
CDP_PORT="${CDP_PORT:-9226}"

PIDS=()
trap cleanup EXIT

cd "$REPRO_DIR"
PIDS+=("$!")

PIDS+=("$!")

for _ in $(seq 1 50); do
  [[ "$body" == *'"Browser": "Lightpanda'* ]] && break
  sleep 0.1
done

fi

exec node repro.js
```

**Run**: `bash repro.sh`
- Today: prints `FAIL: expected "hit", got null.` and exits 1.
- Expected after fix: prints `PASS: querySelector matched the element with the literal backslash value.` and exits 0.

## Likely fix location

`src/browser/webapi/selector/Parser.zig` — `fn attributeValue`. Walk the quoted string respecting backslash escapes (handle `\\`, `\"`/`\'`, `\<hex digits>`, and `\<newline>` line continuations), reusing the existing `parseEscape` helper that already powers `parseIdentifier` for `#id` / `.class` selectors.

## Environment

- Lightpanda: `1.0.0-nightly.5812+b3257754` (also reproduces on `main` HEAD `9fbade45`)
- OS: macOS aarch64 (also reproduces on Linux x86_64 nightly)
- CDP client: `chrome-remote-interface` 0.33.x

navidemad:
## What

`document.querySelector('[data-x="abc\\\\def"]')` returned `null` against an element whose attribute value is the 7-byte string `abc\def`, because the selector parser did not decode CSS escape sequences inside quoted attribute values. This decodes them per [CSS Syntax Module Level 3 §4.3.5 "Consume a string token"](https://drafts.csswg.org/css-syntax/#consume-string-token), reusing the existing `parseEscape` helper that powers `parseIdentifier` for `#id` / `.class` selectors.

Closes #2268.

## Root cause

`Parser.attributeValue` walked the quoted value with `std.mem.indexOfScalarPos` to find the closing quote and returned `input[1..end]` raw — backslashes were neither honored as escape boundaries nor decoded. So `[data-x="abc\\def"]` returned the 8-byte slice `abc\\def` to the matcher, and `[data-x="foo\"bar"]` truncated at the first inner `"` it saw (returning `foo\`).

```mermaid
flowchart LR
    A[selector with backslash escape] --> B[indexOfScalarPos to first quote]
    B --> C[raw 8-byte slice]
    C --> D[matcher returns null]
    style B fill:#fdd
    style C fill:#fdd
    style D fill:#fdd
```

## Fix

- `src/browser/webapi/selector/Parser.zig` — `attributeValue`: walk the quoted string char-by-char, calling `parseEscape` on each `\<x>` sequence and stopping at the matching unescaped closing quote. Decode `\\`, `\"`/`\'`, and `\<hex digits>`; drop `\<newline>` line continuations; surface bare newlines and trailing backslashes as `InvalidAttributeSelector`.
- `src/browser/webapi/selector/Parser.zig` — `attribute`: thread the existing arena through to `attributeValue` so escape decoding can allocate when needed.

```mermaid
flowchart LR
    A[selector with backslash escape] --> B[walk string and call parseEscape]
    B --> C[decoded 7-byte value]
    C --> D[matcher finds the element]
    style B fill:#dfd
    style C fill:#dfd
    style D fill:#dfd
```

## Test

- **Unit test**: `src/browser/webapi/selector/Parser.zig:test "Selector: Parser.attributeValue"` — covers unquoted values, plain double/single-quoted values, escaped backslash, escaped quote (both flavors), hex escape with explicit space terminator, hex escape without delimiter (consumes up to 6 hex digits), multi-byte UTF-8 hex escape, line-continuation escaped newline, and three error cases (missing closing quote, bare newline inside string, trailing backslash before EOF). Fails on `main`, passes with this commit.
- **HTML fixture**: `src/browser/tests/element/attribute_value_escapes.html` — runs through the `WebApi: Element` directory test runner (`htmlRunner("element", .{})`) and exercises `querySelector` + `querySelectorAll` against an element whose attribute contains a literal backslash, an embedded quote, and the equivalent hex-escape-encoded selector form.
- **Reproducer**: the `repro.sh` from #2268 exits 1 against current nightly / `main`, exits 0 with this commit. The script drives `lightpanda serve` over CDP and asserts that `document.querySelector('p[data-random="abc\\\\def"]')?.textContent === 'hit'`.

### DIFF
--- a/src/browser/webapi/selector/Parser.zig
+++ b/src/browser/webapi/selector/Parser.zig
@@ -921,8 +921,7 @@ fn attribute(self: *Parser, arena: Allocator) !Selector.Attribute {
     const matcher_type = try self.attributeMatcher();
     _ = self.skipSpaces();
 
-    const value_raw = try self.attributeValue(arena);
-    const value = try arena.dupe(u8, value_raw);
+    const value = try self.attributeValue(arena);
     _ = self.skipSpaces();
 
     // Parse optional case-sensitivity flag
@@ -1014,7 +1013,7 @@ fn attributeValue(self: *Parser, arena: Allocator) ![]const u8 {
         // Decode \\ \" \' and \<hex digits>, treat \<newline> as a line continuation,
         // and stop at the matching unescaped closing quote.
         // https://drafts.csswg.org/css-syntax/#consume-string-token
-        var result: std.ArrayList(u8) = .empty;
+        var result = try std.ArrayList(u8).initCapacity(arena, input.len);
         var i: usize = 1;
         while (i < input.len and input[i] != quote) {
             const b = input[i];
@@ -1063,7 +1062,7 @@ fn attributeValue(self: *Parser, arena: Allocator) ![]const u8 {
 
     const value = input[0..i];
     self.input = input[i..];
-    return value;
+    return arena.dupe(u8, value);
 }
 
 fn asUint(comptime string: anytype) std.meta.Int(
--- a/src/browser/webapi/selector/Parser.zig
+++ b/src/browser/webapi/selector/Parser.zig
@@ -921,7 +921,7 @@ fn attribute(self: *Parser, arena: Allocator) !Selector.Attribute {
     const matcher_type = try self.attributeMatcher();
     _ = self.skipSpaces();
 
-    const value_raw = try self.attributeValue();
+    const value_raw = try self.attributeValue(arena);
     const value = try arena.dupe(u8, value_raw);
     _ = self.skipSpaces();
 
@@ -1002,7 +1002,7 @@ fn attributeMatcher(self: *Parser) !std.meta.FieldEnum(Selector.AttributeMatcher
     };
 }
 
-fn attributeValue(self: *Parser) ![]const u8 {
+fn attributeValue(self: *Parser, arena: Allocator) ![]const u8 {
     const input = self.input;
     if (input.len == 0) {
         return error.InvalidAttributeSelector;
@@ -1010,10 +1010,41 @@ fn attributeValue(self: *Parser) ![]const u8 {
 
     const quote = input[0];
     if (quote == '"' or quote == '\'') {
-        const end = std.mem.indexOfScalarPos(u8, input, 1, quote) orelse return error.InvalidAttributeSelector;
-        const value = input[1..end];
-        self.input = input[end + 1 ..];
-        return value;
+        // Walk the string respecting backslash escapes per CSS Syntax Level 3 §4.3.5.
+        // Decode \\ \" \' and \<hex digits>, treat \<newline> as a line continuation,
+        // and stop at the matching unescaped closing quote.
+        // https://drafts.csswg.org/css-syntax/#consume-string-token
+        var result: std.ArrayList(u8) = .empty;
+        var i: usize = 1;
+        while (i < input.len and input[i] != quote) {
+            const b = input[i];
+            if (b == '\\') {
+                if (i + 1 >= input.len) {
+                    // Backslash at EOF inside a string is a parse error per spec;
+                    // surface it as a missing closing quote.
+                    return error.InvalidAttributeSelector;
+                }
+                const after = input[i + 1];
+                if (after == '\n') {
+                    // Escaped newline inside a string is a line continuation: drop both.
+                    i += 2;
+                    continue;
+                }
+                const escape = try parseEscape(input[i + 1 ..], arena);
+                try result.appendSlice(arena, escape.bytes);
+                i += 1 + escape.consumed;
+                continue;
+            }
+            if (b == '\n') {
+                // Bare newline terminates a string token (parse error).
+                return error.InvalidAttributeSelector;
+            }
+            try result.append(arena, b);
+            i += 1;
+        }
+        if (i >= input.len) return error.InvalidAttributeSelector;
+        self.input = input[i + 1 ..];
+        return result.items;
     }
 
     var i: usize = 0;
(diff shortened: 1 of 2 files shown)


src/browser/webapi/selector/Parser.zig
```diff
@@ -921,7 +921,7 @@ fn attribute(self: *Parser, arena: Allocator) !Selector.Attribute {
     const matcher_type = try self.attributeMatcher();
     _ = self.skipSpaces();
 
-    const value_raw = try self.attributeValue();
+    const value_raw = try self.attributeValue(arena);
```
karlseguin: Since we're passing in the arena and allocating inside the escape-path, we end up double duping (the line after this dupes). I think we should remove line 925 and then let `attributeValue` be responsible for ensure the value is owned, i.e by duping in the non-escape path.

src/browser/webapi/selector/Parser.zig
```diff
@@ -1002,18 +1002,49 @@ fn attributeMatcher(self: *Parser) !std.meta.FieldEnum(Selector.AttributeMatcher
...
     if (quote == '"' or quote == '\'') {
-        const end = std.mem.indexOfScalarPos(u8, input, 1, quote) orelse return error.InvalidAttributeSelector;
-        const value = input[1..end];
-        self.input = input[end + 1 ..];
-        return value;
+        // Walk the string respecting backslash escapes per CSS Syntax Level 3 §4.3.5.
+        // Decode \\ \" \' and \<hex digits>, treat \<newline> as a line continuation,
+        // and stop at the matching unescaped closing quote.
+        // https://drafts.csswg.org/css-syntax/#consume-string-token
+        var result: std.ArrayList(u8) = .empty;
```
karlseguin:
```suggestion
        var result = try std.ArrayList(u8).initCapacity(arena, input.len);
```

src/browser/webapi/selector/Parser.zig
navidemad: Done in a5e5639a. `attributeValue` now owns the result allocation in both paths — the quoted path builds into the arena via the `ArrayList`, and the unquoted path does `arena.dupe(u8, value)` itself — so the call site in `fn attribute` drops the redundant `arena.dupe` and is just `const value = try self.attributeValue(arena);`.

src/browser/webapi/selector/Parser.zig
navidemad: Adopted in a5e5639a. Pre-sizing to `input.len` is a clean upper bound since CSS escape sequences only ever shrink (`\\` -> `\`, `\41 ` -> `A`), so the `ArrayList` never has to grow past its initial capacity.

PR #2271, opened 2026-04-27
return null on getContext('webgl') and getContext('experimental-webgl')

karlseguin:
Currently, we return a dummy `WebGLRenderingContext`. But this type implements virtually nothing. From MDN: "If the context identifier is not supported null is returned."

Now, normally we like to move things along as much as we can. Returning `WebGLRenderingContext` lets us know the next thing the code needs. But in this case, it seems to be problematic. At least some code is defensive around a null value from getContext('webgl'), but once we return an instance, it expects it to work correctly. This is what causes https://github.com/features/copilot to enter an endless JS loop.


karlseguin:
This is a specific fix for https://github.com/lightpanda-io/browser/pull/2241

krichprollsch (approved):

PR #2280, opened 2026-04-28
forms: round <input type=range> value to nearest step on the step ladder

### LINKED ISSUE #2277: forms: <input type=range> does not perform step matching
## Summary

`<input type="range">` does not perform step matching as specified in the WHATWG HTML "input type=range" value sanitization algorithm. After clamping to `[min, max]` (which #2266 / #2267 implements), the value should be rounded to the nearest valid step on the step ladder; this final rounding is currently skipped.

This is an explicit follow-up to #2266 / PR #2267, which intentionally scoped clamping only.

## Spec

https://html.spec.whatwg.org/multipage/input.html#range-state-(type=range) — "Suffering from a step mismatch" subsection. Algorithm:

1. Let `step base` be the result of step base for the element (defaults to `min`).
2. Let `step value` be the value of the `step` attribute, or `1` if absent / not a valid floating-point number.
3. If the value is not on the step ladder defined by `step base + step value * n` (for integer `n`), round to the nearest valid value, with ties rounded up. The rounded value must remain in `[min, max]`; if the rounded-up tie would exceed `max`, use the rounded-down value instead.

## Observed (current behavior)

```html
<input type="range" min="0.5" max="1.5">
```

```js
el.value = '1';
console.log(el.value);  // "1" (Lightpanda)
```

## Expected (Firefox / Chrome)

```js
el.value = '1';
console.log(el.value);  // "1.5"
// step base = min = 0.5, step = 1 (default), valid ladder: 0.5, 1.5
// '1' is exactly between; spec rounds ties up -> '1.5'
```

## Reproducer

Pure DOM, no CDP needed:

```html
<!DOCTYPE html>
<input id="r" type="range" min="0.5" max="1.5">
<script>
  const r = document.getElementById('r');
  r.value = '1';
  document.title = r.value;  // Firefox: "1.5", Lightpanda: "1"
</script>
```

```mermaid
sequenceDiagram
    participant Caller
    participant Input as HTMLInputElement
    participant Sanitize as sanitizeRange
    Caller->>Input: el.value = "1"
    Input->>Sanitize: sanitize("1", min=0.5, max=1.5)
    Note over Sanitize: parse min/max, default 0/100
    alt After PR #2267 (current main)
        Sanitize->>Sanitize: clamp to [0.5, 1.5]
        Note right of Sanitize: 1 is in range, return "1"
        Sanitize-->>Input: "1"
    else Spec / Firefox / Chrome
        Sanitize->>Sanitize: clamp to [0.5, 1.5]
        Sanitize->>Sanitize: snap to step ladder<br/>(base=0.5, step=1 -> 0.5, 1.5)
        Note right of Sanitize: 1 -> tie, round up -> 1.5
        Sanitize-->>Input: "1.5"
    end
    Input-->>Caller: el.value === "1" (bug) or "1.5" (spec)
```

## Scope

- Add a `step` getter respecting the `step` attribute (defaulting to `1`).
- Add `step base` resolution (defaults to `min`, but per spec falls back further when `min` is absent).
- Implement the step-ladder rounding inside `sanitizeRange` after the existing min/max clamp (`src/browser/webapi/element/html/Input.zig`).
- Extend the `range-clamp` block in `src/browser/tests/element/html/input-attrs.html` with step-matching cases.

The same algorithm covers `step="any"` (no rounding) and `step="<float>"` non-default values.

## Related

- #2266 / PR #2267 — clamping only (this issue's prerequisite, currently in review).

## Discovered via

Review feedback on PR #2267 (`https://github.com/lightpanda-io/browser/pull/2267#discussion_r3146448733`).

navidemad:
## What

`<input type=range>` value sanitization clamps to `[min, max]` but skips the WHATWG step-matching step that should round to the nearest `step base + step * n`. As a result, `el.value = '1'` on `<input type=range min=0.5 max=1.5>` returns `'1'` instead of the spec-required `'1.5'` (the only valid rung in `[0.5, 1.5]` other than `0.5`). This PR adds the missing step-snap, mirroring Firefox / Chrome behavior.

Closes #2277.

## Root cause

`sanitizeRange` in `src/browser/webapi/element/html/Input.zig` implemented steps 1–3 of the [value sanitization algorithm](https://html.spec.whatwg.org/multipage/input.html#range-state-(type=range)) (default-on-invalid, clamp to min, clamp to max) but never read the `step` attribute. The "Suffering from a step mismatch" rounding step was explicitly deferred (see code comment around `r3.value = '1'` in `input-attrs.html`).

```mermaid
flowchart LR
    A[el.value = X] --> B[parse min/max attrs]
    B --> C[clamp X to min,max]
    C --> D[return clamped value]
    style D fill:#fdd
```

## Fix

- `src/browser/webapi/element/html/Input.zig` — `sanitizeRange`: thread `getStep()` and the `value` content attribute through and call the new `snapToStep` helper after the clamp; preserve the original string when the input is already on the ladder so `'1.0'` does not get rewritten to `'1'`.
- `src/browser/webapi/element/html/Input.zig` — new `snapToStep`: parse `step` (defaults to 1, treats `<= 0` as 1, treats `step="any"` case-insensitively as a bypass), compute `n = round((value - step_base) / step)` with ties rounding toward positive infinity, fall back to `n - 1` when the rounded-up neighbor exceeds max, and leave the value untouched when no rung lands in `[min, max]`.
- Step base resolution follows the [spec's fallback chain](https://html.spec.whatwg.org/multipage/input.html#concept-input-min): `min` content attr → `value` content attr → `0`. So `<input type=range value="3.5" max="10">` then `el.value = '5.3'` snaps to `5.5` (ladder `[3.5, 4.5, 5.5, …]`) rather than `5` (ladder `[0, 1, 2, …]`).

```mermaid
flowchart LR
    A[el.value = X] --> B[parse min/max/step attrs]
    B --> C[clamp X to min,max]
    C --> D[snapToStep: round to nearest rung]
    D --> E[return snapped value]
    style D fill:#dfd
    style E fill:#dfd
```

## Test

- **Unit test**: `WebApi: HTML.Input` (`src/browser/tests/element/html/input-attrs.html`, new `range-step-matching` block) — fails on `main`, passes with this commit. Covers default `step=1` with fractional min, integer steps with ties, `step="any"` (and `"ANY"`), clamp-then-snap, tie-exceeds-max fallback, on-ladder string preservation, invalid step (`0`, `-2`) defaulting to 1, negative ranges, the non-numeric-falls-back-to-midpoint path, the step-base fallback to `value` content attr (s10), the unparseable-`value`-attr-falls-through-to-0 path (s11), and `min` content attr taking precedence over `value` content attr (s12).
- **Pre-existing tests**: the `range-clamp` block continues to pass; the dropped comment about `r3.value = '1'` being out of scope is no longer applicable.

## Notes

- Floating-point precision is left to f64. Cases like `step=0.1` may exhibit tiny rounding error on pathological inputs; matching Chrome's `Decimal`-based precision is out of scope. The new test cases use steps that round exactly in f64.

### DIFF
--- a/src/browser/webapi/element/html/Input.zig
+++ b/src/browser/webapi/element/html/Input.zig
@@ -564,7 +564,10 @@ fn sanitizeValue(self: *Input, comptime dupe: bool, value: []const u8, frame: *F
         .time => return if (isValidTime(value)) if (comptime dupe) try frame.dupeString(value) else value else "",
         .@"datetime-local" => return try sanitizeDatetimeLocal(dupe, value, frame.arena),
         .number => return if (isValidFloatingPoint(value)) if (comptime dupe) try frame.dupeString(value) else value else "",
-        .range => return try sanitizeRange(dupe, value, self.getMin(), self.getMax(), self.getStep(), frame),
+        .range => {
+            const value_attr = self.asConstElement().getAttributeSafe(comptime .wrap("value")) orelse "";
+            return try sanitizeRange(dupe, value, self.getMin(), self.getMax(), self.getStep(), value_attr, frame);
+        },
         .color => {
             if (value.len == 7 and value[0] == '#') {
                 var needs_lower = false;
@@ -798,13 +801,16 @@ fn sanitizeDatetimeLocal(comptime dupe: bool, value: []const u8, arena: std.mem.
 ///      neighbor instead.
 /// `min`/`max` default to 0 and 100 respectively when the attribute is missing
 /// or fails to parse as a valid floating-point number. `step` defaults to 1;
-/// `step="any"` (case-insensitive) disables step matching.
+/// `step="any"` (case-insensitive) disables step matching. The step base
+/// (https://html.spec.whatwg.org/multipage/input.html#concept-input-min) falls
+/// back through `min` content attr → `value` content attr → 0.
 fn sanitizeRange(
     comptime dupe: bool,
     value: []const u8,
     min_attr: []const u8,
     max_attr: []const u8,
     step_attr: []const u8,
+    value_attr: []const u8,
     frame: *Frame,
 ) ![]const u8 {
     const min: f64 = if (isValidFloatingPoint(min_attr))
@@ -815,16 +821,22 @@ fn sanitizeRange(
         std.fmt.parseFloat(f64, max_attr) catch 100
     else
         100;
+    const step_base: f64 = if (isValidFloatingPoint(min_attr))
+        std.fmt.parseFloat(f64, min_attr) catch 0
+    else if (isValidFloatingPoint(value_attr))
+        std.fmt.parseFloat(f64, value_attr) catch 0
+    else
+        0;
 
     if (!isValidFloatingPoint(value)) {
-        return try formatFloat(frame.arena, snapToStep(min + (max - min) / 2, min, max, min, step_attr));
+        return try formatFloat(frame.arena, snapToStep(min + (max - min) / 2, min, max, step_base, step_attr));
     }
 
     const v0 = std.fmt.parseFloat(f64, value) catch unreachable; // grammar already validated
     var v = v0;
     if (v < min) v = min;
     if (v > max) v = max;
-    const snapped = snapToStep(v, min, max, min, step_attr);
+    const snapped = snapToStep(v, min, max, step_base, step_attr);
     if (v == v0 and snapped == v) {
         // Already valid and on the ladder — preserve the original string so
         // assignments like `el.value = "1.0"` round-trip without canonicalizing.
@@ -839,7 +851,7 @@ fn sanitizeRange(
 /// unchanged for `step="any"` (case-insensitive) or when no ladder rung lands
 /// in `[min, max]`.
 fn snapToStep(value: f64, min: f64, max: f64, step_base: f64, step_attr: []const u8) f64 {
-    if (step_attr.len == 3 and std.ascii.eqlIgnoreCase(step_attr, "any")) return value;
+    if (std.ascii.eqlIgnoreCase(step_attr, "any")) return value;
 
     const step: f64 = blk: {
         if (isValidFloatingPoint(step_attr)) {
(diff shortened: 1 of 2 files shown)


src/browser/webapi/element/html/Input.zig
```diff
@@ -812,13 +817,53 @@ fn sanitizeRange(
...
+    return try formatFloat(frame.arena, snapped);
+}
+
+/// Snap `value` (already clamped to `[min, max]`) to the nearest value on the
+/// step ladder `step_base + step * n`. Ties round up; if the rounded-up
+/// neighbor exceeds `max`, use the rounded-down neighbor. Returns `value`
+/// unchanged for `step="any"` (case-insensitive) or when no ladder rung lands
+/// in `[min, max]`.
+fn snapToStep(value: f64, min: f64, max: f64, step_base: f64, step_attr: []const u8) f64 {
+    if (step_attr.len == 3 and std.ascii.eqlIgnoreCase(step_attr, "any")) return value;
```
karlseguin: `eqlIgnoreCase` does its own length check, so it's redundant (and wasteful) to do it first.

karlseguin (commented):
> el.value = X from JS does not touch the content attribute, so the fallback is unreachable from the value-setter path.

But it is reached via input.setAttribute('value', 'x');

Maybe that isn't worth worry about, but worth pointing out maybe.

src/browser/webapi/element/html/Input.zig
navidemad: Adopted in 577a405f. Confirmed `std.ascii.eqlIgnoreCase` already short-circuits on length mismatch (`ascii.zig:329`), so the outer guard was strictly redundant.

navidemad:
Good catch — implemented the full fallback chain in 577a405f. Step base now resolves through `min` content attr → `value` content attr → 0 per [the spec](https://html.spec.whatwg.org/multipage/input.html#concept-input-min), with new test cases s10–s12 in `input-attrs.html` covering the fallback, the unparseable-`value`-attr path, and `min` taking precedence.

PR #2286, opened 2026-04-28
forms: implement HTML constraint validation API

### LINKED ISSUE #2284: forms: HTML constraint validation API not implemented (validity, checkValidity, setCustomValidity)
## Summary

Lightpanda doesn't implement the [HTML constraint validation API](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#the-constraint-validation-api). On a freshly-loaded page, `HTMLInputElement.prototype.validity`, `validationMessage`, `checkValidity`, `reportValidity`, and `setCustomValidity` are all `undefined`, and `ValidityState` is not exposed as a global. The same surface is missing from `HTMLSelectElement`, `HTMLTextAreaElement`, `HTMLButtonElement`, and `HTMLFormElement`. CDP clients that drive form interactions through JS (Puppeteer's `$eval`, raw `Runtime.evaluate`) cannot detect required-but-empty fields, type-mismatched email/URL inputs, range-bound numbers, or set/clear custom errors.

## Today's behavior

`Runtime.evaluate` returning `'validity' in HTMLInputElement.prototype` produces `false`. Calls like `document.querySelector('input').checkValidity()` throw `TypeError: undefined is not a function`. There is no `ValidityState` constructor on the global.

```mermaid
sequenceDiagram
    participant LP as Lightpanda
    participant V8 as V8 / Bridge
    Client->>LP: Runtime.evaluate `inp.validity.valueMissing`
    LP->>V8: lookup `validity` on HTMLInputElement.prototype
    V8-->>LP: undefined (no accessor registered)
    LP-->>Client: TypeError: Cannot read properties of undefined
    Note over LP,V8: src/browser/webapi/element/html/Input.zig only exposes `willValidate`
```

## Expected behavior

Per [HTML Living Standard §4.10.18.4–§4.10.18.6](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#the-constraint-validation-api), every form-associated submittable element exposes:

- `validity: ValidityState` — readonly, with 11 boolean flags (`valueMissing`, `typeMismatch`, `patternMismatch`, `tooLong`, `tooShort`, `rangeUnderflow`, `rangeOverflow`, `stepMismatch`, `badInput`, `customError`, `valid`)
- `validationMessage: string` — UA-defined error message, empty when valid
- `willValidate: boolean` — whether the element participates in constraint validation
- `checkValidity(): boolean` — runs the constraint validation steps; fires the `invalid` event (cancelable) on each failing element
- `reportValidity(): boolean` — same, plus UI feedback (in headless: identical to `checkValidity`)
- `setCustomValidity(message: string)` — sets `ValidityState.customError` (clears it if `message` is empty)

`HTMLFormElement` exposes `checkValidity()` / `reportValidity()` that iterate the form's listed elements.

```mermaid
sequenceDiagram
    participant LP as Lightpanda
    participant V8 as V8 / Bridge
    Client->>LP: Runtime.evaluate `inp.validity.valueMissing`
    LP->>V8: lookup `validity` on HTMLInputElement.prototype
    V8->>LP: Input.getValidity(self) → *ValidityState
    LP->>V8: lookup `valueMissing` on ValidityState.prototype
    V8->>LP: ValidityState.getValueMissing(self) → bool
    LP-->>Client: true / false
```

## Reproducer

Self-contained: one HTML fixture + one shell script + one Node CDP driver. Requires `node` and `python3`; one-time `npm install --no-save ws` happens in the script.

`repro.html`:

```html
<!DOCTYPE html>
<title>constraint validation API repro</title>
<form id="f">
  <input id="empty-required" type="text" required>
  <input id="filled-required" type="text" required value="x">
  <input id="bad-email" type="email" value="not-an-email">
  <input id="too-low" type="number" min="10" value="3">
</form>
```

`repro.sh`:

```bash
#!/usr/bin/env bash
set -euo pipefail
HERE="$(cd "$(dirname "$0")" && pwd)"
BIN="${LIGHTPANDA_BIN:-lightpanda}"

lsof -ti tcp:9222 -sTCP:LISTEN | xargs -r kill -9 2>/dev/null || true
lsof -ti tcp:8765 -sTCP:LISTEN | xargs -r kill -9 2>/dev/null || true

(cd "$HERE" && python3 -m http.server 8765 >/tmp/repro-http.log 2>&1) &
HTTP_PID=$!
"$BIN" serve --host 127.0.0.1 --port 9222 >/tmp/repro-lp.log 2>&1 &
LP_PID=$!
trap 'kill $HTTP_PID $LP_PID 2>/dev/null; lsof -ti tcp:9222 -sTCP:LISTEN | xargs -r kill -9 2>/dev/null; lsof -ti tcp:8765 -sTCP:LISTEN | xargs -r kill -9 2>/dev/null' EXIT

for i in $(seq 1 30); do
  if curl -sf http://127.0.0.1:8765/repro.html >/dev/null && \
     curl -sf http://127.0.0.1:9222/json/version >/dev/null; then
    break
  fi
  sleep 0.2
done

if [ ! -d "$HERE/node_modules/ws" ]; then
  (cd "$HERE" && npm install --no-save --silent ws >/dev/null 2>&1)
fi

cd "$HERE"
set +e
node repro.js
exit $?
```

`repro.js`:

```js

function makeClient(url) {
  let nextId = 1;
  const pending = new Map();
  const handlers = new Map();
  ws.on('message', (raw) => {
      pending.delete(m.id);
      else resolve(m.result);
    } else if (m.method) {
    }
  });
  return {
      const id = nextId++;
      });
    },
    close() { ws.close(); },
  };
}

(async () => {
  const client = makeClient('ws://127.0.0.1:9222/');
  await client.ready();
  await client.send('Page.navigate', { url: 'http://127.0.0.1:8765/repro.html' }, sessionId);
  await new Promise(r => setTimeout(r, 500));

  const { result } = await client.send('Runtime.evaluate', {
    expression: `(() => {
      const out = {};
      const inp = document.getElementById('filled-required');
      out.has_validity = 'validity' in inp;
      out.has_checkValidity = typeof inp.checkValidity === 'function';
      out.has_setCustomValidity = typeof inp.setCustomValidity === 'function';
      out.has_ValidityState_global = typeof ValidityState !== 'undefined';
      if (out.has_validity) {
        out.empty_valueMissing = document.getElementById('empty-required').validity.valueMissing;
        out.bad_email_typeMismatch = document.getElementById('bad-email').validity.typeMismatch;
        out.too_low_rangeUnderflow = document.getElementById('too-low').validity.rangeUnderflow;
      }
      return out;
    })()`,
    returnByValue: true,
  }, sessionId);

  console.log(JSON.stringify(result.value, null, 2));
  client.close();

  const v = result.value;
  const failed =
    !v.has_validity || !v.has_checkValidity || !v.has_setCustomValidity ||
    !v.has_ValidityState_global || v.empty_valueMissing !== true ||
    v.bad_email_typeMismatch !== true || v.too_low_rangeUnderflow !== true;
  process.exit(failed ? 1 : 0);
```

- **Today** (current public nightly `1.0.0-nightly.5816+a578f4d6`): prints
  ```
  {
    "has_validity": false,
    "has_checkValidity": false,
    "has_setCustomValidity": false,
    "has_ValidityState_global": false
  }
  ```
  and exits **1**.
- **Expected**: prints all four `true` plus `empty_valueMissing: true`, `bad_email_typeMismatch: true`, `too_low_rangeUnderflow: true`, and exits **0**.

## Likely fix location

- New file: `src/browser/webapi/element/html/ValidityState.zig` — the interface itself, registered in `src/browser/js/bridge.zig` `PageJsApis`.
- `src/browser/webapi/element/html/Input.zig` — add `getValidity` / `getValidationMessage` / `checkValidity` / `reportValidity` / `setCustomValidity` and the per-flag predicate functions (`suffersValueMissing`, `suffersTypeMismatch`, `suffersTooLong`, `suffersRangeUnderflow` / `Overflow`, etc.). Wire them into the existing `JsApi` block alongside `willValidate`.
- `src/browser/webapi/element/html/Select.zig`, `TextArea.zig`, `Button.zig` — same surface; constraint set is narrower per element type (Select: only `valueMissing` + `customError`; TextArea: + `tooLong` / `tooShort`; Button: only `customError` and only when `type=submit`).
- `src/browser/webapi/element/html/Form.zig` — `checkValidity()` / `reportValidity()` iterate `Form.iterator()` and dispatch the `invalid` event on each failing element.

## Environment

- Lightpanda: `1.0.0-nightly.5816+a578f4d6` (public nightly)
- OS: darwin (aarch64)
- CDP client: raw WebSocket (`ws` npm package); protocol unchanged from Chrome's CDP for `Runtime.evaluate` / `Target.attachToTarget`

navidemad:
## What

Lightpanda did not implement the [HTML constraint validation API](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#the-constraint-validation-api): `element.validity`, `validationMessage`, `checkValidity`, `reportValidity`, `setCustomValidity` were absent on every form-associated element type, and `ValidityState` was not registered as a global. CDP clients driving form interactions through `Runtime.evaluate` could not detect required-but-empty fields, type-mismatched email/URL inputs, range-bound numbers, or set/clear custom errors.

This change adds the full surface — a new `ValidityState` interface with all 11 readonly boolean flags, plus the validation methods and predicates on `HTMLInputElement`, `HTMLSelectElement`, `HTMLTextAreaElement`, `HTMLButtonElement`, and `HTMLFormElement`. `checkValidity` fires a cancelable `invalid` event on each element whose `ValidityState.valid` is false.

Closes #2284.

## Root cause

`src/browser/webapi/element/html/Input.zig` exposed only `willValidate`. There was no `ValidityState.zig`, no `checkValidity` / `reportValidity` / `setCustomValidity` functions, and no per-flag predicates (`suffersValueMissing`, `suffersTypeMismatch`, etc.). The same was true for `Select.zig`, `TextArea.zig`, `Button.zig`, and `Form.zig`. Lookups like `inp.validity.valueMissing` returned `undefined` because the `validity` accessor was never registered on the JS prototype.

```mermaid
flowchart LR
    A[CDP Client: inp.validity.valueMissing] --> B[V8 prototype lookup of validity]
    B --> C[undefined accessor]
    C --> D[TypeError: undefined is not a function]
    style B fill:#fdd
    style C fill:#fdd
    style D fill:#fdd
```

## Fix

```mermaid
flowchart LR
    A[CDP Client: inp.validity.valueMissing] --> B[Input.getValidity → *ValidityState]
    B --> C[ValidityState.getValueMissing dispatches on owner element type]
    C --> D[Input.suffersValueMissing → bool]
    D --> E[true / false]
    style B fill:#dfd
    style C fill:#dfd
    style D fill:#dfd
    style E fill:#dfd
```

- `src/browser/webapi/element/html/ValidityState.zig` (new): the interface itself. Stores a single `_owner: *Element` and dispatches each flag getter on the concrete element type, since the per-flag rules are defined per element in the HTML spec.
- `src/browser/js/bridge.zig`: registers `ValidityState` in `PageJsApis` so V8 can construct the prototype.
- `src/browser/webapi/element/html/Input.zig`: adds `_custom_validity` field, `getValidity` / `getValidationMessage` / `checkValidity` / `reportValidity` / `setCustomValidity`, and per-flag predicates (`suffersValueMissing`, `suffersTypeMismatch` for `email` / `url`, `suffersTooLong` / `suffersTooShort` via codepoint count, `suffersRangeUnderflow` / `Overflow` for numeric input types). Honors the existing `getWillValidate` gating (hidden / button / reset / disabled / datalist-ancestor inputs are barred).
- `src/browser/webapi/element/html/Select.zig`: adds the same surface; `suffersValueMissing` triggers when the effective option's `value` attribute is empty (matching the spec's "placeholder label option" semantics).
- `src/browser/webapi/element/html/TextArea.zig`: adds the same surface plus `tooLong` / `tooShort` against `maxlength` / `minlength`.
- `src/browser/webapi/element/html/Button.zig`: adds the surface but limited to `customError` per spec — buttons with `type="submit"` participate in constraint validation only via `setCustomValidity`; `type="reset"` and `type="button"` are barred entirely.
- `src/browser/webapi/element/html/Form.zig`: `checkValidity()` and `reportValidity()` iterate `Form.iterator()` and dispatch the `invalid` event on each failing element.

Three flags are deliberately not implemented in this PR and return `false`:
- `patternMismatch` — needs JS RegExp evaluation from Zig (`new RegExp("^(?:" + pattern + ")$").test(value)`); no clean Zig-side path yet. TODO comment in `Input.suffersPatternMismatch`.
- `stepMismatch` — overlaps with PR #2280 (range step rounding); deferring to keep that PR's diff clean.
- `badInput` — set when the UA cannot convert raw user input (e.g. typing letters in `<input type=number>`). Headless Lightpanda receives values via attributes / JS only, so there is no user input to be "bad".

## Test

- **Unit tests**: 4 new HTML fixtures under `src/browser/tests/element/html/` cover Input (12 scripts: surface, valueMissing, checkbox/radio, typeMismatch for email and url, range, tooLong, willValidate, customError, checkValidity, invalid event, reportValidity, validationMessage, ValidityState identity), Select (6 scripts), TextArea (6 scripts), Form (5 scripts). Run via the existing `htmlRunner` harness (`WebApi: HTML.Input` etc.).
- **Toggle-off check**: deleting the validity-related lines from `Input.JsApi` makes `input-validity.html` fail at the surface assertion (10ms execution, 0 of 1 passed). Restoring brings it back to all green.
- **Reproducer**: against the locally-built binary from this branch, the repro from #2284 prints all four `has_*: true` plus the three behavior assertions and exits 0. Against the older nightly binary, it prints all four `has_*: false` and exits 1.
- **Full unit suite** (`mise exec -- zig build test $V8`): 490 of 490 tests pass; no regressions.

## Notes

- `ValidityState` instances are created fresh per `getValidity()` call rather than cached on the element. The objects hold only an `*Element` pointer, so identity is cheap and the cost of allocation is small. If JS code captures `el.validity` and then mutates the element, the next read still reflects current state.
- `validationMessage` returns short generic English strings (`"Please fill out this field."`, etc.) rather than fully localized strings — matching the rough shape of Chrome's defaults but skipping the locale machinery.
- `Input.radioGroupHasChecked` does a tree walk to find sibling radios with the same `name`. This mirrors the existing `uncheckRadioGroup` walker pattern in the same file.


krichprollsch:
@navidemad Thanks for the contribution.
Do you have a use case for the change? I'm wondering if it's really useful to implement that since the server should already enforce the constraints.

navidemad:
Hi Karl, thanks for taking a look.

The use case is **headless test automation**, not scraping. CDP clients driving browser-level tests need to assert client-side form state without ever hitting the server — patterns like `expect(field).to be_valid` (Capybara), `await page.$eval('input', el => el.validity.valueMissing)` (Puppeteer), or `inp.checkValidity()` from a custom `Runtime.evaluate`. The constraints are an HTML-authoring decision (`required`, `type=email`, `min`/`max`, `setCustomValidity` from app JS), and tests verify the *page's* validation logic, not the server's.

Two things make this lower-cost than it looks:

1. The constraint logic is already in Lightpanda. `willValidate` is exposed; `required` / `min` / `max` / `type=email` already parse and gate submission. The PR adds the *query surface* (`validity.valueMissing`, `checkValidity`, `setCustomValidity`) on top of state Lightpanda already computes — not new enforcement.

2. Real-world SPA submit handlers call `form.checkValidity()` directly:
   ```js
   form.addEventListener('submit', e => {
     if (!form.checkValidity()) { e.preventDefault(); showCustomError(); }
   });
   ```
   With `checkValidity` undefined this throws `TypeError`, and depending on whether the site catches it, either the form submits unguarded or the page visibly breaks. Either way it surfaces as "this site doesn't work in Lightpanda" rather than as a missing-API issue.

Concrete downstream impact: in our Capybara CDP driver, two `#has_field with valid` matchers are skip-listed purely for this reason. After this lands they pass without any gem-side polyfill.

Happy to scope it down if you'd prefer a smaller starting cut — e.g. land just `validity` + `checkValidity` + `setCustomValidity` on `HTMLInputElement`, defer Select / TextArea / Button / Form / form-iteration to follow-ups. Let me know what shape works.

(Also just merged `main` into the branch — `3b3e4e41` — so the PR is mergeable on top of current upstream.)

src/browser/webapi/element/html/Button.zig
```diff
@@ -114,6 +119,52 @@ pub fn getLabels(self: *Button, frame: *Frame) !js.Array {
...
+pub fn getValidationMessage(self: *const Button) []const u8 {
+    if (!self.getWillValidate()) return "";
+    return self._custom_validity orelse "";
+}
+
+pub fn checkValidity(self: *Button, frame: *Frame) !bool {
+    if (!self.getWillValidate()) return true;
+    if (self._custom_validity == null) return true;
+
+    const event = try Event.init("invalid", .{ .cancelable = true }, frame._page);
```
krichprollsch: Since the event is generated by the browser I think it's better to use `Event.initTrusted` here.

src/browser/webapi/element/html/Input.zig
```diff
@@ -211,6 +213,209 @@ fn hasDatalistAncestor(self: *const Input) bool {
...
+    if (self.suffersRangeOverflow()) return "Value is too large.";
+    return "";
+}
+
+pub fn checkValidity(self: *Input, frame: *Frame) !bool {
+    if (!self.getWillValidate()) return true;
+    const v = ValidityState{ ._owner = self.asElement() };
+    if (v.getValid()) return true;
+
+    const event = try Event.init("invalid", .{ .cancelable = true }, frame._page);
```
krichprollsch: same about `initTrusted`

src/browser/webapi/element/html/Select.zig
```diff
@@ -250,6 +253,64 @@ pub fn getLabels(self: *Select, frame: *Frame) !js.Array {
...
+    if (self.suffersValueMissing()) return "Please select an item in the list.";
+    return "";
+}
+
+pub fn checkValidity(self: *Select, frame: *Frame) !bool {
+    if (!self.getWillValidate()) return true;
+    const v = ValidityState{ ._owner = self.asElement() };
+    if (v.getValid()) return true;
+
+    const event = try Event.init("invalid", .{ .cancelable = true }, frame._page);
```
krichprollsch: same about `initTrusted`

src/browser/webapi/element/html/TextArea.zig
```diff
@@ -284,6 +286,77 @@ pub fn getLabels(self: *TextArea, frame: *Frame) !js.Array {
...
+    if (self.suffersTooShort()) return "Please lengthen this text.";
+    return "";
+}
+
+pub fn checkValidity(self: *TextArea, frame: *Frame) !bool {
+    if (!self.getWillValidate()) return true;
+    const v = ValidityState{ ._owner = self.asElement() };
+    if (v.getValid()) return true;
+
+    const event = try Event.init("invalid", .{ .cancelable = true }, frame._page);
```
krichprollsch: same about `initTrusted`

src/browser/webapi/element/html/Input.zig
```diff
@@ -211,6 +213,209 @@ fn hasDatalistAncestor(self: *const Input) bool {
...
 }
 
+// Constraint validation API
+// https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#the-constraint-validation-api
+
+pub fn getValidity(self: *Input, frame: *Frame) !*ValidityState {
+    return frame._factory.create(ValidityState{ ._owner = self.asElement() });
+}
+
+pub fn getValidationMessage(self: *Input) []const u8 {
```
krichprollsch: I think it would be better to use `*const Input` and cascade the change on the sub-functions here.

src/browser/webapi/element/html/Select.zig
```diff
@@ -250,6 +253,64 @@ pub fn getLabels(self: *Select, frame: *Frame) !js.Array {
...
+
+pub fn getWillValidate(self: *const Select) bool {
+    return !self.getDisabled();
+}
+
+pub fn getValidity(self: *Select, frame: *Frame) !*ValidityState {
+    return frame._factory.create(ValidityState{ ._owner = self.asElement() });
+}
+
+pub fn getValidationMessage(self: *Select) []const u8 {
```
krichprollsch: I think it would be better to use `*const Select` and cascade the change on the sub-functions here.

src/browser/webapi/element/html/Button.zig
```diff
@@ -114,6 +119,52 @@ pub fn getLabels(self: *Button, frame: *Frame) !js.Array {
...
+
+pub fn reportValidity(self: *Button, frame: *Frame) !bool {
+    return self.checkValidity(frame);
+}
+
+pub fn setCustomValidity(self: *Button, message: []const u8, frame: *Frame) !void {
+    if (message.len == 0) {
+        self._custom_validity = null;
+    } else {
+        self._custom_validity = try frame.arena.dupe(u8, message);
```
krichprollsch: You should use `frame.dupeString(message)` instead.

src/browser/webapi/element/html/Input.zig
```diff
@@ -211,6 +213,209 @@ fn hasDatalistAncestor(self: *const Input) bool {
...
+pub fn reportValidity(self: *Input, frame: *Frame) !bool {
+    // Headless: no UI to draw, so reportValidity matches checkValidity exactly.
+    return self.checkValidity(frame);
+}
+
+pub fn setCustomValidity(self: *Input, message: []const u8, frame: *Frame) !void {
+    if (message.len == 0) {
+        self._custom_validity = null;
+    } else {
+        self._custom_validity = try frame.arena.dupe(u8, message);
```
krichprollsch: You should use `frame.dupeString(message)` instead.

src/browser/webapi/element/html/Select.zig
```diff
@@ -250,6 +253,64 @@ pub fn getLabels(self: *Select, frame: *Frame) !js.Array {
...
+
+pub fn reportValidity(self: *Select, frame: *Frame) !bool {
+    return self.checkValidity(frame);
+}
+
+pub fn setCustomValidity(self: *Select, message: []const u8, frame: *Frame) !void {
+    if (message.len == 0) {
+        self._custom_validity = null;
+    } else {
+        self._custom_validity = try frame.arena.dupe(u8, message);
```
krichprollsch: You should use `frame.dupeString(message)` instead.

src/browser/webapi/element/html/TextArea.zig
```diff
@@ -284,6 +286,77 @@ pub fn getLabels(self: *TextArea, frame: *Frame) !js.Array {
...
+
+pub fn reportValidity(self: *TextArea, frame: *Frame) !bool {
+    return self.checkValidity(frame);
+}
+
+pub fn setCustomValidity(self: *TextArea, message: []const u8, frame: *Frame) !void {
+    if (message.len == 0) {
+        self._custom_validity = null;
+    } else {
+        self._custom_validity = try frame.arena.dupe(u8, message);
```
krichprollsch: You should use `frame.dupeString(message)` instead.

src/browser/webapi/element/html/Input.zig
```diff
@@ -211,6 +213,209 @@ fn hasDatalistAncestor(self: *const Input) bool {
...
+}
+
+pub fn suffersRangeOverflow(self: *const Input) bool {
+    return numericRangeBreach(self, .overflow);
+}
+
+fn numericRangeBreach(self: *const Input, comptime kind: enum { underflow, overflow }) bool {
+    // Only types whose value can be interpreted as a number have range constraints.
+    switch (self._input_type) {
+        .number, .range, .date, .time, .@"datetime-local", .month, .week => {},
```
krichprollsch: Should we accept dates/times type here?
The following `isValidFloatingPoint` call won't success isn't it?

src/browser/webapi/element/html/Input.zig
```diff
@@ -211,6 +213,209 @@ fn hasDatalistAncestor(self: *const Input) bool {
...
+
+fn radioGroupHasChecked(self: *Input) bool {
+    const element = self.asElement();
+    const name = element.getAttributeSafe(comptime .wrap("name")) orelse return self._checked;
+    if (name.len == 0) return self._checked;
+
+    const root = element.asNode().getRootNode(null);
+    const TreeWalker = @import("../../TreeWalker.zig");
+    var walker = TreeWalker.Full.init(root, .{});
+    while (walker.next()) |node| {
```
krichprollsch: Same behavior it more correctly implemented into `uncheckRadioGroup`. You must also check if both radio are in the same form element. maybe we can factorize someway?

src/browser/webapi/element/html/Input.zig
```diff
@@ -211,6 +213,209 @@ fn hasDatalistAncestor(self: *const Input) bool {
...
+    return self._custom_validity != null;
+}
+
+pub fn suffersValueMissing(self: *Input) bool {
+    if (!self.getWillValidate()) return false;
+    if (!self.getRequired()) return false;
+    return switch (self._input_type) {
+        .checkbox => !self._checked,
+        .radio => !self.radioGroupHasChecked(),
+        // file inputs aren't supported yet (#2175); treat as always-empty when required.
```
krichprollsch: Maybe add a `TODO` to prefix the comment.

src/browser/webapi/element/html/TextArea.zig
```diff
@@ -284,6 +286,77 @@ pub fn getLabels(self: *TextArea, frame: *Frame) !js.Array {
...
+pub fn suffersValueMissing(self: *const TextArea) bool {
+    if (!self.getWillValidate()) return false;
+    if (!self.getRequired()) return false;
+    return self.getValue().len == 0;
+}
+
+pub fn suffersTooLong(self: *const TextArea) bool {
+    const value = self._value orelse return false;
+    const max_attr = self.asConstElement().getAttributeSafe(comptime .wrap("maxlength")) orelse return false;
+    const max = std.fmt.parseInt(i32, max_attr, 10) catch return false;
```
krichprollsch: Would be great to implement `getMaxLenght` (similar as `Input`) and expose the getter.
https://developer.mozilla.org/en-US/docs/Web/API/HTMLTextAreaElement/maxLength

src/browser/webapi/element/html/TextArea.zig
```diff
@@ -284,6 +286,77 @@ pub fn getLabels(self: *TextArea, frame: *Frame) !js.Array {
...
+    if (max < 0) return false;
+    const count = std.unicode.utf8CountCodepoints(value) catch value.len;
+    return count > @as(usize, @intCast(max));
+}
+
+pub fn suffersTooShort(self: *const TextArea) bool {
+    const value = self._value orelse return false;
+    if (value.len == 0) return false;
+    const min_attr = self.asConstElement().getAttributeSafe(comptime .wrap("minlength")) orelse return false;
+    const min = std.fmt.parseInt(i32, min_attr, 10) catch return false;
```
krichprollsch: Would be great to extract and expose minlength getter
https://developer.mozilla.org/en-US/docs/Web/API/HTMLTextAreaElement/minLength

src/browser/webapi/element/html/Input.zig
```diff
@@ -211,6 +213,209 @@ fn hasDatalistAncestor(self: *const Input) bool {
...
+    const max = self.getMaxLength();
+    if (max < 0) return false;
+    return codepointCount(value) > @as(usize, @intCast(max));
+}
+
+pub fn suffersTooShort(self: *const Input) bool {
+    const value = self._value orelse return false;
+    if (value.len == 0) return false;
+    const min_attr = self.asConstElement().getAttributeSafe(comptime .wrap("minlength")) orelse return false;
+    const min = std.fmt.parseInt(i32, min_attr, 10) catch return false;
```
krichprollsch: Would be great to extract and expose minlength getter
https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/minLength

src/browser/webapi/element/html/Input.zig
```diff
@@ -211,6 +213,209 @@ fn hasDatalistAncestor(self: *const Input) bool {
     return false;
 }
 
+// Constraint validation API
+// https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#the-constraint-validation-api
+
+pub fn getValidity(self: *Input, frame: *Frame) !*ValidityState {
```
krichprollsch, line 220: We should cache the `ValidityState` with the element instead of creating a new one at each call.
For example, if the `getValidity is called per keystroke, it can allocate many structs...

Moreover it could fix the equality issue.

navidemad:
Done in `604f8128`.

- `_validity: ?*ValidityState` cache on Input/Select/TextArea/Button — `el.validity === el.validity` now holds, and per-keystroke `getValidity` calls no longer reallocate.
- Four `Event.initTrusted` + four `frame.dupeString(message)` switches.
- `*const` cascade through `getValidationMessage` → `suffersValueMissing` for Input and Select; `Select.effectiveOption` rewritten to walk children via `firstChild`/`nextSibling`.
- Factored a shared `radioGroupIterator` + a frame-less `sameFormOwner` helper. `radioGroupHasChecked` now enforces the same-form check too (was only in `uncheckRadioGroup`). The helper string-compares the `form` attribute when set, otherwise compares ancestor `<form>` pointers — keeps the const validation path working without plumbing `*Frame` through `ValidityState`.
- `numericRangeBreach` narrowed to `.number, .range`. The date/time/month/week/datetime-local arms were silently returning false because `isValidFloatingPoint` rejects those values; left a TODO for type-specific conversion (days-since-epoch, ms-since-midnight, etc.) since that's a larger change.
- `getMinLength`/`setMinLength` added to Input and TextArea; `getMaxLength`/`setMaxLength` added to TextArea (Input already had them). Wired through `suffersTooLong`/`suffersTooShort` and exposed via IDL.
- TODO prefix on the file-input value-missing comment.

Tests extended in `input-validity.html` / `select-validity.html` / `textarea-validity.html` / `button.html` for the new IDL surface, the validity-cache identity, and `isTrusted: true` on the dispatched `invalid` event.

krichprollsch (approved):

PR #2289, opened 2026-04-28
page: implement Page.getNavigationHistory and Page.navigateToHistoryEntry

### LINKED ISSUE #2288: page: Page.getNavigationHistory and Page.navigateToHistoryEntry not implemented
## Summary

The CDP `Page` domain does not implement `Page.getNavigationHistory` or `Page.navigateToHistoryEntry`. Both calls return `-31998 UnknownMethod`, even though the underlying browsing-context history is fully tracked internally (`window.history.length`, `history.back()`, `history.forward()`, the `Navigation` API in `webapi/navigation/Navigation.zig` all work). CDP clients that expose programmatic back/forward (the standard "go back N entries" automation surface) must therefore fall back to evaluating `history.back()` / `history.forward()` via `Runtime.evaluate`, which is async, opaque to the protocol, and can't address arbitrary entries by id.

## Today's behavior

Both methods are absent from the `Page` dispatch enum in `src/cdp/domains/page.zig`, so the generic `UnknownMethod` path runs:

```mermaid
sequenceDiagram
    participant Page as cdp/domains/page.zig
    participant Nav as webapi/navigation/Navigation.zig
    Client->>Page: Page.getNavigationHistory
    Page-->>Client: error -31998 UnknownMethod
    Note over Page,Nav: Nav._entries (with _index, _url, _id) exists<br/>but no dispatch handler reads it
    Client->>Page: Page.navigateToHistoryEntry { entryId: 0 }
    Page-->>Client: error -31998 UnknownMethod
    Note over Page,Nav: Nav.navigateInner(_, .{traverse=N}, _) exists<br/>but no dispatch handler invokes it
```

## Expected behavior

Per the CDP spec:

- [`Page.getNavigationHistory`](https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-getNavigationHistory) returns `{ currentIndex: integer, entries: NavigationEntry[] }` where each `NavigationEntry` has `id`, `url`, `userTypedURL`, `title`, `transitionType`.
- [`Page.navigateToHistoryEntry`](https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-navigateToHistoryEntry) takes `{ entryId: integer }` and traverses the session history to that entry.

```mermaid
sequenceDiagram
    participant Page as cdp/domains/page.zig
    participant Nav as webapi/navigation/Navigation.zig
    Client->>Page: Page.getNavigationHistory
    Page->>Nav: read _entries + _index
    Nav-->>Page: entries, currentIndex
    Page-->>Client: { currentIndex, entries: [...] }
    Client->>Page: Page.navigateToHistoryEntry { entryId }
    Page->>Nav: find entry by id, get index
    Page->>Nav: frame.navigate(url, .{ kind = .{ .traverse = idx } })
    Nav-->>Page: ok (Page.loadEventFired follows)
    Page-->>Client: {}
```

## Reproducer

Self-contained: one shell script + one Node CDP driver + one tiny static server. No Ruby/Capybara/Rails — just `lightpanda serve` plus `node` and `python3`. The CDP driver is ~80 lines using raw `ws` (no `chrome-remote-interface` because Lightpanda's `/json/list` returns `[]` and `/json/protocol` 404s).

`repro.sh` (orchestration):

```bash
#!/usr/bin/env bash
set -euo pipefail
cd "$(dirname "$0")"

lsof -ti :9222 | xargs -r kill -9 2>/dev/null || true
lsof -ti :8765 | xargs -r kill -9 2>/dev/null || true

python3 -c "
import http.server, socketserver
class H(http.server.BaseHTTPRequestHandler):
    def do_GET(self):
        path = self.path.lstrip('/')
        body = f'<!doctype html><html><head><title>{path.upper()}</title></head><body>page-{path}</body></html>'
        self.send_response(200)
        self.send_header('Content-Type', 'text/html')
        self.send_header('Content-Length', str(len(body)))
        self.end_headers()
        self.wfile.write(body.encode())
    def log_message(self, *a): pass
with socketserver.TCPServer(('127.0.0.1', 8765), H) as httpd:
    httpd.serve_forever()
" &
HTTP_PID=$!

"$LIGHTPANDA_BIN" serve --host 127.0.0.1 --port 9222 >/tmp/repro-lp.log 2>&1 &
LP_PID=$!

cleanup() { kill -9 $LP_PID $HTTP_PID 2>/dev/null || true; wait 2>/dev/null || true; }
trap cleanup EXIT

for i in {1..30}; do curl -sf http://127.0.0.1:8765/a >/dev/null 2>&1 && break; sleep 0.1; done
for i in {1..30}; do curl -sf http://127.0.0.1:9222/json/version >/dev/null 2>&1 && break; sleep 0.1; done

[ -d node_modules/ws ] || npm install --no-save --silent ws

set +e
node repro.js
RC=$?
set -e
echo "repro exit=$RC"
exit $RC
```

`repro.js` (CDP probe):

```js

(async () => {
  const c = await connect();

  await c.navigate('http://127.0.0.1:8765/a');
  await c.navigate('http://127.0.0.1:8765/b');
  await c.navigate('http://127.0.0.1:8765/c');

  let history;
  try {
    history = await c.send('Page.getNavigationHistory');
  } catch (e) {
    console.log('FAIL: Page.getNavigationHistory -', e.message);
    c.close(); process.exit(1);
  }

  const entries = history.entries || [];
  if (entries.length !== 3 || history.currentIndex !== 2 || !entries[2].url.endsWith('/c')) {
    console.log('FAIL: unexpected history shape:', JSON.stringify(history));
    c.close(); process.exit(1);
  }

  try {
    await c.send('Page.navigateToHistoryEntry', { entryId: entries[0].id });
  } catch (e) {
    console.log('FAIL: Page.navigateToHistoryEntry -', e.message);
    c.close(); process.exit(1);
  }

  const ok = await c.waitFor(`window.location.pathname === '/a'`);
  if (!ok) {
    console.log('FAIL: did not traverse back to /a');
    c.close(); process.exit(1);
  }

  const after = await c.send('Page.getNavigationHistory');
  if (after.currentIndex !== 0 || (after.entries || []).length !== 3) {
    console.log('FAIL: post-traversal currentIndex wrong:', JSON.stringify(after));
    c.close(); process.exit(1);
  }

  console.log('PASS: getNavigationHistory + navigateToHistoryEntry behave per CDP spec');
  c.close(); process.exit(0);
})().catch(e => { console.error('UNEXPECTED:', e); process.exit(2); });
```

`cdp.js` is the same WebSocket + `Target.createTarget`/`attachToTarget`/`Page.enable`/`Runtime.enable` scaffolding used in our other reproducers. Happy to inline if useful.

- Today (`main` HEAD `716b6f33`): `FAIL: Page.getNavigationHistory - -31998: UnknownMethod` — exit 1.
- After fix: `PASS: getNavigationHistory + navigateToHistoryEntry behave per CDP spec` — exit 0.

## Likely fix location

`src/cdp/domains/page.zig` — `processMessage`. Add `getNavigationHistory` and `navigateToHistoryEntry` to the dispatch enum + switch, and implement two thin handlers that read `bc.session.navigation._entries` / `_index` for the get, and call `frame.navigate(url, .{ .reason = .history, .kind = .{ .traverse = idx } })` for the navigate. Internal infrastructure already exists in `webapi/navigation/Navigation.zig` (the `back`/`forward`/`navigateInner` paths use the same `_entries` array) — this is purely a CDP plumbing gap.

## Context

Filing per #2043 — this was top-ranked in that thread's prioritization comment. The downstream-consumer payoff is replacing async `Runtime.evaluate("history.back()")` + navigation polling with one synchronous CDP round-trip targeting a specific entry id.

## Environment

- Lightpanda: `1.0.0-dev.5835+716b6f33` (`main` HEAD)
- OS: macOS aarch64
- CDP client: raw `ws` v8 (Node 22)

navidemad:
## What

Implement `Page.getNavigationHistory` and `Page.navigateToHistoryEntry` in the CDP `Page` domain. Both methods returned `-31998 UnknownMethod` even though the underlying browsing-context history is fully tracked in `webapi/navigation/Navigation.zig` (`back`/`forward`/`navigateInner` already use it). This is purely a CDP plumbing gap.

Closes #2288.

## Root cause

`processMessage` in `src/cdp/domains/page.zig` did not include `getNavigationHistory` or `navigateToHistoryEntry` in its dispatch enum or switch, so any client call hit the generic `UnknownMethod` arm.

```mermaid
flowchart LR
    A[CDP: Page.getNavigationHistory or navigateToHistoryEntry] --> B[processMessage stringToEnum]
    B --> C[no match → -31998 UnknownMethod]
    style B fill:#fdd
    style C fill:#fdd
```

## Fix

- `src/cdp/domains/page.zig` — `processMessage`: add `getNavigationHistory` and `navigateToHistoryEntry` to the dispatch enum + switch.
- `src/cdp/domains/page.zig` — new `getNavigationHistory(cmd)`: walks `bc.session.navigation._entries`, returns `{ currentIndex, entries: [{id, url, userTypedURL, title, transitionType}] }`. Internal `_id` is a decimal-formatted counter; parsed to `i64` for the response shape (CDP spec uses integer ids).
- `src/cdp/domains/page.zig` — new `navigateToHistoryEntry(cmd)`: looks up the entry by `entryId`, mirrors the `navigate`/`doReload` pattern (reset isolated worlds + `replacePage` if the frame isn't waiting), then calls `frame.navigate(url, .{ .reason = .history, .kind = .{ .traverse = idx } })`. Re-uses the existing traverse path in `Navigation.commitNavigation`, which only updates `_index` (no new entry pushed).
- The URL is duplicated into the new frame's arena before navigation because `replacePage` releases the previous arena that backed the entry's `_url` slice.

```mermaid
flowchart LR
    A[CDP: Page.getNavigationHistory] --> B[read Navigation._entries + _index]
    B --> C[response: currentIndex + entries]
    D[CDP: Page.navigateToHistoryEntry] --> E[find entry by id]
    E --> F[frame.navigate kind=.traverse]
    F --> G[Page.loadEventFired follows]
    style B fill:#dfd
    style C fill:#dfd
    style E fill:#dfd
    style F fill:#dfd
    style G fill:#dfd
```

## Test

- **Unit test**: `src/cdp/domains/page.zig` — `cdp.frame: getNavigationHistory + navigateToHistoryEntry`. Builds a 3-entry history, asserts the get response shape (entries + currentIndex), traverses back to entry 0, traverses forward to entry 1, asserts unknown `entryId` returns `InvalidParams`, and asserts both methods return `BrowserContextNotLoaded` when no session is attached. Fails on `main` (UnknownMethod), passes with this commit.
- **Reproducer**: the `repro.sh` from #2288 exits 1 against `main` with `FAIL: Page.getNavigationHistory - -31998: UnknownMethod` and exits 0 with this commit (`PASS: ...`). The script drives a Lightpanda + raw-`ws` CDP probe through three navigations, then exercises both methods and asserts the URL after a traversal.
- **Full suite**: `mise exec -- zig build test $V8` — 491/491 pass.

## Notes

- `title` is returned as `""` for now — Lightpanda's `NavigationHistoryEntry` doesn't track per-entry titles. Could be wired up later by capturing `document.title` at `commitNavigation` time and storing alongside `_state`.
- `transitionType` is hardcoded to `"other"`. Mapping from internal `NavigateReason` (`anchor`/`address_bar`/`form`/`script`/`history`) to the CDP enum (`link`/`typed`/`form_submit`/`reload`/...) is straightforward but felt out of scope for this PR — happy to follow up.
- `userTypedURL` mirrors `url` since Lightpanda doesn't separately track address-bar input vs. resolved URL. Matches what other engines do for non-typed entries.


src/cdp/domains/page.zig
```diff
@@ -368,6 +372,92 @@ fn doReload(cmd: *CDP.Command) !void {
...
+    }
+
+    const nav = &bc.session.navigation;
+    const entries_in = nav._entries.items;
+
+    const entries_out = try cmd.arena.alloc(NavigationEntry, entries_in.len);
+    for (entries_in, 0..) |entry, i| {
+        // Internal _id is a decimal-formatted monotonically-increasing counter
+        // (see Navigation.pushEntry); fall back to index if parse ever fails.
+        const eid = std.fmt.parseInt(i64, entry._id, 10) catch @as(i64, @intCast(i));
```
mookums: Since this is contractually guaranteed by the `Navigation.pushEntry` for now, might be worth having this be:

```zig
const eid = std.fmt.parseInt(i64, entry._id, 10) catch @panic("Navigation Id was not a integer");
```

Or something similar. I'm not sure if it's always guaranteed that the index is the same as the entry id.

mookums (changes requested):

src/cdp/domains/page.zig
navidemad: Done in `ff93f4fc`. Adopted the panic approach in both `getNavigationHistory` and (for consistency) `navigateToHistoryEntry`, where the previous `catch continue` would silently skip the entry and surface a misleading `InvalidParams`. `Navigation.pushEntry` always `allocPrint`s `_id` from a `usize` counter via `{d}`, so parse failure is an internal invariant violation either way.

mookums (approved):

PR #2292, opened 2026-04-28
events: implement KeyboardEvent.keyCode and charCode legacy attributes

### LINKED ISSUE #2291: events: KeyboardEvent.keyCode and charCode hardcoded to 0
## Summary

`KeyboardEvent.keyCode` is hardcoded to return `0` for every key. Per [W3C UI Events § Annex C: Legacy Key Attributes](https://www.w3.org/TR/uievents/#legacy-key-attributes), `keyCode` should report the legacy fixed virtual key code for the key being pressed (e.g. `84` for `'t'`/`'T'`, `16` for `Shift`, `13` for `Enter`). `KeyboardEvent.charCode` is also hardcoded to `0`; per the same spec it should be the Unicode code point of the character produced on `keypress` events.

These are deprecated attributes but they remain widely used by JS event handlers and test harnesses driving the browser via CDP `Input.dispatchKeyEvent`.

## Today's behavior

Both getters in `src/browser/webapi/event/KeyboardEvent.zig` are stubs that always return `0`, regardless of which key the event represents:

```zig
// src/browser/webapi/event/KeyboardEvent.zig:273-282
pub fn getCharCode(self: *const KeyboardEvent) u32 {
    _ = self;
    return 0;
}

pub fn getKeyCode(self: *const KeyboardEvent) u32 {
    _ = self;
    return 0;
}
```

Effect: a CDP client that drives the browser through `Input.dispatchKeyEvent` and reads `e.keyCode` from a JS `keydown` listener observes `0` for every key.

```mermaid
sequenceDiagram
    participant LP as Lightpanda
    participant JS as Page JS
    Client->>LP: Input.dispatchKeyEvent type=keyDown key=T modifiers=8
    LP->>JS: dispatch keydown event
    JS->>LP: read e.keyCode
    LP-->>JS: 0
    Note over LP: getKeyCode returns 0 unconditionally
    JS-->>Client: keydown:0
```

## Expected behavior

Per [W3C UI Events § Annex C](https://www.w3.org/TR/uievents/#legacy-key-attributes):

- For modifier and navigation keys, return the fixed virtual key code from the spec table (e.g. `Shift`=16, `Enter`=13, `ArrowLeft`=37, `F1`=112).
- For printable characters, return the uppercase ASCII of the unmodified key value (`'a'`/`'A'` → 65, `'0'` → 48, `' '` → 32).
- `charCode` is meaningful only on `keypress` events; for `keydown`/`keyup` it returns `0`. On `keypress` it returns the Unicode code point of the character produced.

```mermaid
sequenceDiagram
    participant LP as Lightpanda
    participant JS as Page JS
    Client->>LP: Input.dispatchKeyEvent type=keyDown key=T modifiers=8
    LP->>JS: dispatch keydown event
    JS->>LP: read e.keyCode
    LP-->>JS: 84
    Note over LP: getKeyCode reads _key, returns spec value
    JS-->>Client: keydown:84
```

## Reproducer

Self-contained: one HTML fixture + one Node CDP driver + a shell harness. No frameworks.

`repro.html`:
```html
<!DOCTYPE html>
<html>
<head><meta charset="utf-8"><title>KeyboardEvent.keyCode reproducer</title></head>
<body>
<input id="i" type="text" />
<script>
  const input = document.getElementById('i');
  input.focus();
  window._observed = [];
  input.addEventListener('keydown', function(e) {
    window._observed.push({ key: e.key, code: e.code, keyCode: e.keyCode, charCode: e.charCode });
  });
</script>
</body>
</html>
```

`repro.js` (CDP driver — uses raw `ws`, no `chrome-remote-interface` since Lightpanda's `/json/list` returns `[]` and needs `Target.createTarget` + `Target.attachToTarget` for every session):

```js

function client(url) {
  let id = 1; const pend = new Map();
  ws.on('message', (raw) => {
    if (m.id != null && pend.has(m.id)) {
      const { resolve, reject } = pend.get(m.id);
      pend.delete(m.id);
      m.error ? reject(new Error(m.error.message)) : resolve(m.result);
    }
  });
  return {
    ready: () => new Promise(r => ws.on('open', r)),
    send: (method, params = {}, sessionId) => new Promise((resolve, reject) => {
      const myId = id++;
      pend.set(myId, { resolve, reject });
      ws.send(JSON.stringify({ id: myId, method, params, sessionId }));
    }),
    close: () => ws.close(),
  };
}

(async () => {
  const c = client('ws://127.0.0.1:9222/');
  await c.ready();
  const { targetId } = await c.send('Target.createTarget', { url: 'about:blank' });
  const { sessionId } = await c.send('Target.attachToTarget', { targetId, flatten: true });
  await c.send('Page.enable', {}, sessionId);
  await c.send('Runtime.enable', {}, sessionId);
  await c.send('Page.navigate', { url: 'http://127.0.0.1:8125/repro.html' }, sessionId);
  // settle
    const r = await c.send('Runtime.evaluate', {
    }, sessionId).catch(() => null);
    if (r && r.result && r.result.value === 'complete') break;
  }

  const cases = [
    { key: 'Shift',     code: 'ShiftLeft', want: 16 },
    { key: 'T',         code: 'KeyT',      want: 84, text: 'T', unmodifiedText: 't', mods: 8 },
    { key: 'Enter',     code: 'Enter',     want: 13 },
    { key: 'ArrowLeft', code: 'ArrowLeft', want: 37 },
    { key: 'a',         code: 'KeyA',      want: 65, text: 'a' },
    { key: '5',         code: 'Digit5',    want: 53, text: '5' },
    { key: 'F1',        code: 'F1',        want: 112 },
  ];
  for (const e of cases) {
    const p = { type: 'keyDown', key: e.key, code: e.code, modifiers: e.mods || 0 };
    if (e.text) p.text = e.text;
    if (e.unmodifiedText) p.unmodifiedText = e.unmodifiedText;
    await c.send('Input.dispatchKeyEvent', p, sessionId);
    await c.send('Input.dispatchKeyEvent', { type: 'keyUp', key: e.key, code: e.code }, sessionId);
  }

  const r = await c.send('Runtime.evaluate', {
    expression: 'JSON.stringify(window._observed)', returnByValue: true,
  }, sessionId);
  const got = JSON.parse(r.result.value);

  let pass = true;
  for (let i = 0; i < cases.length; i++) {
    const ok = got[i] && got[i].keyCode === cases[i].want;
    console.log(
      `${ok ? 'OK  ' : 'FAIL'}  key=${JSON.stringify(cases[i].key).padEnd(12)} ` +
      `expected keyCode=${String(cases[i].want).padEnd(4)} got=${got[i] ? got[i].keyCode : '<missing>'}`
    );
    if (!ok) pass = false;
  }
  c.close();
  process.exit(pass ? 0 : 1);
```

`repro.sh`:
```bash
#!/usr/bin/env bash
set -euo pipefail
cd "$(dirname "$0")"
PORT=9222
FIXTURE_PORT=8125
lsof -ti tcp:$PORT 2>/dev/null | xargs -r kill -9 2>/dev/null || true
lsof -ti tcp:$FIXTURE_PORT 2>/dev/null | xargs -r kill -9 2>/dev/null || true
python3 -m http.server "$FIXTURE_PORT" --bind 127.0.0.1 >/tmp/keycode.fixture.log 2>&1 &
FIXTURE_PID=$!
"$LIGHTPANDA_BIN" serve --host 127.0.0.1 --port "$PORT" >/tmp/keycode.lp.log 2>&1 &
LP_PID=$!
trap 'kill "$FIXTURE_PID" 2>/dev/null || true; kill "$LP_PID" 2>/dev/null || true' EXIT
for i in $(seq 1 30); do
  curl -sf "http://127.0.0.1:$FIXTURE_PORT/repro.html" >/dev/null 2>&1 && \
    curl -sf "http://127.0.0.1:$PORT/json/version" >/dev/null 2>&1 && break
  sleep 0.1
done
[ -d node_modules/ws ] || npm i --no-save ws >/dev/null 2>&1
node repro.js
```

Today (against `main` HEAD `2bbf23b3`):
```
FAIL  key="Shift"      expected keyCode=16   got=0
FAIL  key="T"          expected keyCode=84   got=0
FAIL  key="Enter"      expected keyCode=13   got=0
FAIL  key="ArrowLeft"  expected keyCode=37   got=0
FAIL  key="a"          expected keyCode=65   got=0
FAIL  key="5"          expected keyCode=53   got=0
FAIL  key="F1"         expected keyCode=112  got=0
exit=1
```

Expected (after the fix):
```
OK    key="Shift"      expected keyCode=16   got=16
OK    key="T"          expected keyCode=84   got=84
OK    key="Enter"      expected keyCode=13   got=13
OK    key="ArrowLeft"  expected keyCode=37   got=37
OK    key="a"          expected keyCode=65   got=65
OK    key="5"          expected keyCode=53   got=53
OK    key="F1"         expected keyCode=112  got=112
exit=0
```

## Likely fix location

`src/browser/webapi/event/KeyboardEvent.zig` — `getKeyCode` and `getCharCode`. Add a `keyCode()` method on the `Key` union that returns the W3C UI Events § Annex C fixed virtual key code for each named key, and computes the uppercase-ASCII fallback for the `.standard` printable variant. Have `getKeyCode` delegate to it. Have `getCharCode` check the event type (returns 0 unless `keypress`) and read the first byte of `_key.standard`.

## Caveat (worth flagging on the PR)

For shift-modified symbol keys (e.g. shift+1='!'), spec-strict `keyCode` is the unmodified key's value (49 for digit '1'). Today, `Input.dispatchKeyEvent` only stores the resolved `key` field on `KeyboardEvent`, not the unmodified equivalent. Returning ASCII of the modified char is what most JS engines without OS-level keyboard layout access do, and it's enough for letters and digits where uppercase is identity. Spec-strict behavior for symbols would need plumbing `unmodifiedText` from CDP through to `KeyboardEventOptions` — separate, larger change.

## Environment

- Lightpanda: `main` HEAD `2bbf23b3`
- OS: darwin (macOS aarch64)
- CDP client: raw `ws` over Node 25

navidemad:
## What

`KeyboardEvent.keyCode` and `KeyboardEvent.charCode` are getters that always return `0`. Per [W3C UI Events § Annex C: Legacy Key Attributes](https://www.w3.org/TR/uievents/#legacy-key-attributes), they should report the legacy fixed virtual key code and the Unicode code point of the produced character respectively. This PR implements both per spec.

Closes #2291.

## Root cause

`getKeyCode` and `getCharCode` in `src/browser/webapi/event/KeyboardEvent.zig` were stubs:

```zig
pub fn getCharCode(self: *const KeyboardEvent) u32 {
    _ = self;
    return 0;
}

pub fn getKeyCode(self: *const KeyboardEvent) u32 {
    _ = self;
    return 0;
}
```

A CDP client driving the browser through `Input.dispatchKeyEvent` and reading `e.keyCode` from a JS `keydown` listener observed `0` for every key, regardless of which key was dispatched.

```mermaid
flowchart LR
    A[Input.dispatchKeyEvent key=T] --> B[KeyboardEvent created with _key=T]
    B --> C[getKeyCode reads _key]
    C --> D[stub returns 0]
    style C fill:#fdd
    style D fill:#fdd
```

## Fix

- `src/browser/webapi/event/KeyboardEvent.zig` — `Key.keyCode()`: new method on the `Key` union mapping each named variant to its W3C-spec fixed virtual key code (`Shift`=16, `Enter`=13, `ArrowLeft`=37, function keys 112-123, etc.). For the `.standard` printable variant, returns uppercase ASCII for letters, ASCII for digits, and best-effort ASCII of the first byte for other characters.
- `src/browser/webapi/event/KeyboardEvent.zig` — `getKeyCode`: delegates to `_key.keyCode()`.
- `src/browser/webapi/event/KeyboardEvent.zig` — `getCharCode`: returns 0 for non-`keypress` events; on `keypress`, returns the first byte of `_key.standard` (0 if the key isn't a printable character).

```mermaid
flowchart LR
    A[Input.dispatchKeyEvent key=T] --> B[KeyboardEvent created with _key=T]
    B --> C[getKeyCode reads _key]
    C --> D[Key.keyCode returns 84]
    style C fill:#dfd
    style D fill:#dfd
```

## Test

- **Unit test**: `WebApi: KeyboardEvent` (HTML fixture at `src/browser/tests/event/keyboard.html`) — five new `<script>` blocks (`keyCodeLetter`, `keyCodeDigit`, `keyCodeSpace`, `keyCodeNamed`, `keyCodeUnknown`, `charCode`) cover letters (case-insensitive uppercasing), digits, named keys (modifiers, navigation, editing, function), unknown-key fallback to 0, and the keypress-only semantics of `charCode`. The pre-existing `charCodeKeyCode` block (which asserted `0`/`0`) was replaced. Toggle-off verified: reverting `KeyboardEvent.zig` while keeping the new fixture causes the test to fail (`0 of 1 tests passed`).
- **Reproducer**: the `repro.sh` from #2291 exits 1 on `main` (every key reports `keyCode=0`) and exits 0 with this commit (all seven test cases match the expected spec values).
- **Full suite**: `mise exec -- zig build test $V8` — 490/490 tests pass.

## Notes

- For shift-modified symbol keys (e.g. shift+1='!'), spec-strict `keyCode` is the unmodified key's value (49 for digit '1'). This PR returns the modified char's ASCII because `KeyboardEvent` doesn't currently store the unmodified key — the CDP `Input.dispatchKeyEvent` `unmodifiedText` parameter isn't plumbed through `KeyboardEventOptions`. Letters and digits aren't affected because uppercase('t') == uppercase('T') and digits are unmodified by shift in their key value. Plumbing `unmodifiedText` is a separate, larger change worth doing if a reviewer wants spec-strict behavior across all symbol keys.
- `charCode` is reported per-byte; non-ASCII characters (e.g. 'é') would need full UTF-8 decoding to surface as the correct Unicode code point. No test exercises that path today.

### DIFF
--- a/src/browser/webapi/event/KeyboardEvent.zig
+++ b/src/browser/webapi/event/KeyboardEvent.zig
@@ -160,6 +160,94 @@ pub const Key = union(enum) {
             else => |k| @tagName(k),
         };
     }
+
+    /// Legacy `KeyboardEvent.keyCode` value per UI Events spec § Annex C
+    /// (https://www.w3.org/TR/uievents/#legacy-key-attributes). Returns 0 for
+    /// keys without a defined fixed virtual key code.
+    ///
+    /// Spec note: for printable characters, `keyCode` is calculated from the
+    /// **unmodified** key value's uppercase ASCII. We don't track the
+    /// unmodified key, so we uppercase `_key` instead. This is exact for
+    /// letters (uppercase('t') == uppercase('T')) and digits, but for
+    /// shift-modified symbols (e.g., shift+1='!') it returns the modified
+    /// char's ASCII rather than the digit's keyCode. Callers needing
+    /// spec-strict behavior should pass `unmodifiedText` through CDP
+    /// `Input.dispatchKeyEvent` and use that instead.
+    pub fn keyCode(self: Key) u32 {
+        return switch (self) {
+            // Modifier keys
+            .Alt, .AltGraph => 18,
+            .CapsLock => 20,
+            .Control => 17,
+            .Meta, .Hyper, .Super => 91,
+            .NumLock => 144,
+            .ScrollLock => 145,
+            .Shift => 16,
+
+            // Whitespace keys (Space hits the .standard path below)
+            .Enter => 13,
+            .Tab => 9,
+
+            // Navigation keys
+            .ArrowDown => 40,
+            .ArrowLeft => 37,
+            .ArrowRight => 39,
+            .ArrowUp => 38,
+            .End => 35,
+            .Home => 36,
+            .PageDown => 34,
+            .PageUp => 33,
+
+            // Editing keys
+            .Backspace => 8,
+            .Clear => 12,
+            .Delete => 46,
+            .Insert => 45,
+
+            // UI keys
+            .Cancel => 3,
+            .ContextMenu => 93,
+            .Escape => 27,
+            .Execute => 43,
+            .Help => 47,
+            .Pause => 19,
+            .Select => 41,
+
+            // Function keys
+            .F1 => 112,
+            .F2 => 113,
+            .F3 => 114,
+            .F4 => 115,
+            .F5 => 116,
+            .F6 => 117,
+            .F7 => 118,
+            .F8 => 119,
+            .F9 => 120,
+            .F10 => 121,
+            .F11 => 122,
+            .F12 => 123,
+
+            .standard => |s| {
+                if (s.len == 0) return 0;
+                const c = s[0];
+                // Letters: uppercase ASCII
+                if (c >= 'a' and c <= 'z') return c - 'a' + 'A';
+                if (c >= 'A' and c <= 'Z') return c;
+                // Digits: ASCII
+                if (c >= '0' and c <= '9') return c;
+                // Space: 32 — also returned via the ASCII fall-through below,
+                // but called out for clarity since it's the most common case.
+                if (c == ' ') return 32;
+                // Other ASCII chars (best-effort: legacy keyCode for symbols
+                // is platform-specific and depends on the unmodified key,
+                // which we don't track).
+                return c;
+            },
+
+            // Keys without a defined legacy keyCode
+            else => 0,
+        };
+    }
 };
 
 pub const Location = enum(i32) {
@@ -270,15 +358,22 @@ pub fn getShiftKey(self: *const KeyboardEvent) bool {
     return self._shift_key;
 }
 
-// Deprecated: tracked as 0 since we don't synthesise legacy character codes.
+// https://www.w3.org/TR/uievents/#dom-keyboardevent-charcode
+// charCode is the Unicode code point of the character produced by the key,
+// and is only meaningful on `keypress` events. For `keydown` and `keyup` it
+// is 0. (Deprecated, but read by legacy event handlers.)
 pub fn getCharCode(self: *const KeyboardEvent) u32 {
-    _ = self;
-    return 0;
+    const event = self._proto._proto;
+    if (!std.mem.eql(u8, event._type_string.str(), "keypress")) return 0;
+    return switch (self._key) {
+        .standard => |s| if (s.len > 0) s[0] else 0,
+        else => 0,
+    };
 }
 
+// https://www.w3.org/TR/uievents/#dom-keyboardevent-keycode
 pub fn getKeyCode(self: *const KeyboardEvent) u32 {
-    _ = self;
-    return 0;
+    return self._key.keyCode();
 }
 
 pub fn initKeyboardEvent(
--- a/src/browser/webapi/event/KeyboardEvent.zig
+++ b/src/browser/webapi/event/KeyboardEvent.zig
@@ -248,6 +248,18 @@ pub const Key = union(enum) {
             else => 0,
         };
     }
+
+    /// Legacy `KeyboardEvent.charCode` value per UI Events spec § Annex C
+    /// (https://www.w3.org/TR/uievents/#legacy-key-attributes). Returns the
+    /// Unicode code point of the character produced by the key. Only
+    /// meaningful inside a `keypress` event — callers must gate accordingly.
+    pub fn charCode(self: Key) u32 {
+        return switch (self) {
+            .Enter => 13,
+            .standard => |s| if (s.len > 0) s[0] else 0,
+            else => 0,
+        };
+    }
 };
 
 pub const Location = enum(i32) {
@@ -362,17 +374,22 @@ pub fn getShiftKey(self: *const KeyboardEvent) bool {
 // charCode is the Unicode code point of the character produced by the key,
 // and is only meaningful on `keypress` events. For `keydown` and `keyup` it
 // is 0. (Deprecated, but read by legacy event handlers.)
+//
+// Chrome returns 0 for synthetic events (those created via
+// `new KeyboardEvent(...)` rather than dispatched by the user agent), so we
+// gate on `_is_trusted` to match.
 pub fn getCharCode(self: *const KeyboardEvent) u32 {
     const event = self._proto._proto;
-    if (!std.mem.eql(u8, event._type_string.str(), "keypress")) return 0;
-    return switch (self._key) {
-        .standard => |s| if (s.len > 0) s[0] else 0,
-        else => 0,
-    };
+    if (event._is_trusted == false) return 0;
+    if (event._type_string.eql(comptime .wrap("keypress")) == false) return 0;
+    return self._key.charCode();
 }
 
 // https://www.w3.org/TR/uievents/#dom-keyboardevent-keycode
+//
+// As with `charCode`, Chrome returns 0 for synthetic events.
 pub fn getKeyCode(self: *const KeyboardEvent) u32 {
+    if (self._proto._proto._is_trusted == false) return 0;
     return self._key.keyCode();
 }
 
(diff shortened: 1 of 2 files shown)


src/browser/webapi/event/KeyboardEvent.zig
```diff
@@ -270,15 +358,22 @@ pub fn getShiftKey(self: *const KeyboardEvent) bool {
...
-// Deprecated: tracked as 0 since we don't synthesise legacy character codes.
+// https://www.w3.org/TR/uievents/#dom-keyboardevent-charcode
+// charCode is the Unicode code point of the character produced by the key,
+// and is only meaningful on `keypress` events. For `keydown` and `keyup` it
+// is 0. (Deprecated, but read by legacy event handlers.)
 pub fn getCharCode(self: *const KeyboardEvent) u32 {
-    _ = self;
-    return 0;
+    const event = self._proto._proto;
+    if (!std.mem.eql(u8, event._type_string.str(), "keypress")) return 0;
```
karlseguin:
```suggestion
    if (event._type_string.eql(comptime .wrap("keypress")) == false) return 0;
```

src/browser/webapi/event/KeyboardEvent.zig
```diff
@@ -270,15 +358,22 @@ pub fn getShiftKey(self: *const KeyboardEvent) bool {
...
+    return switch (self._key) {
+        .standard => |s| if (s.len > 0) s[0] else 0,
+        else => 0,
```
karlseguin:
```
.Enter => 13,
```
??

karlseguin (commented):
A lot of these tests don't pass in Chrome. I believe `getKeyCode` and `getCharCode` should continue to return 0 when `self._proto._proto._is_trusted == false`.

navidemad:
Addressed in 73a007f8. Both `getKeyCode` and `getCharCode` now early-return `0` when `self._proto._proto._is_trusted == false`, matching Chrome's behavior for synthetic events. The existing JS assertions in `keyboard.html` are rewritten to expect `0`, and the full per-key mapping table is covered by two new pure-function Zig unit tests (`KeyboardEvent: Key.keyCode mapping` and `KeyboardEvent: Key.charCode mapping`) that exercise `Key.keyCode()` / `Key.charCode()` directly without needing a `Frame`.

src/browser/webapi/event/KeyboardEvent.zig
navidemad: Adopted in 73a007f8.

src/browser/webapi/event/KeyboardEvent.zig
navidemad: Done in 73a007f8 — extracted the charCode mapping into a `Key.charCode()` helper that mirrors `Key.keyCode()`, with a `.Enter => 13` arm so a trusted `keypress` for Enter produces `\r` (U+000D) per the [UI Events legacy attributes](https://www.w3.org/TR/uievents/#legacy-key-attributes) spec. Covered by the `KeyboardEvent: Key.charCode mapping` Zig unit test.

PR #2294, opened 2026-04-28
css: apply UA stylesheet display:none defaults for unrendered elements

### LINKED ISSUE #2293: css: UA stylesheet missing display:none for unrendered elements (head/script/style/template/<input type=hidden>/closed <details> children)
## Summary

Lightpanda's user-agent stylesheet only emits `display: none` for elements carrying the `[hidden]` attribute. The other UA-stylesheet `display: none` rules from the [HTML Rendering spec §15.3.1 "Hidden elements"](https://html.spec.whatwg.org/multipage/rendering.html#hidden-elements) — the unrendered tag-name list, `input[type="hidden" i]`, and `details:not([open]) > *:not(summary)` — are missing.

Consequence: `getComputedStyle(el).display` reports `"block"` for `<head>`, `<script>`, `<style>`, `<link>`, `<meta>`, `<title>`, `<noscript>`, `<template>`, `<param>`, `<source>`, `<track>`, `<area>`, `<datalist>`, `<input type="hidden">`, and the non-`<summary>` direct children of a closed `<details>`. `el.checkVisibility()` returns `true` for the same set. CDP clients that hit-test elements via `getComputedStyle` or `checkVisibility` see "visible" elements that should be skipped per spec.

## Today's behavior

`StyleManager.hasDisplayNone` ([`src/browser/StyleManager.zig:239-243`](https://github.com/lightpanda-io/browser/blob/main/src/browser/StyleManager.zig#L239-L243)) is the centralized truth for `getComputedStyle(el).display === "none"`. It honors only the `[hidden]` UA rule and otherwise defers to `isElementHidden`, which only matches user/author CSS rules — no UA tag-name defaults, no `input[type="hidden"]` rule, no `details:not([open])` parent rule.

`StyleManager.isHidden` (the ancestor-walking variant used by `el.checkVisibility()`, `getBoundingClientRect()`, click hit-testing) calls the same `isElementHidden`, so the gap surfaces identically through both surfaces. Even the existing `[hidden]` short-circuit in `hasDisplayNone` is bypassed when called via `isHidden` — `el.checkVisibility()` returns `true` for an element with `[hidden]` (verified empirically; see Reproducer below).

```mermaid
sequenceDiagram
    participant LP as Lightpanda
    Client->>LP: Runtime.evaluate getComputedStyle(scriptEl).display
    LP->>SM: hasDisplayNone(scriptEl)
    SM->>SM: hasAttributeSafe("hidden") false
    SM->>SM: isElementHidden — only walks user CSS rules
    SM-->>LP: false (no UA tag-name rule applied)
    LP-->>Client: "block"
    Note over SM: spec rule script display:none never consulted
```

## Expected behavior

Per the [HTML Rendering spec §15.3.1 "Hidden elements"](https://html.spec.whatwg.org/multipage/rendering.html#hidden-elements), the UA stylesheet contains:

```css
[hidden]:not([hidden=until-found i]),
area, base, basefont, datalist, head, link, meta, noembed, noframes,
param, rp, script, source, style, template, track, title {
  display: none;
}
input[type="hidden" i] { display: none !important; }
@media (scripting) {
  noscript { display: none !important; }
}
```

Additionally, [§4.11.1 "The `details` element" rendering](https://html.spec.whatwg.org/multipage/rendering.html#the-details-and-summary-elements) yields:

```css
details:not([open]) > *:not(summary) { display: none; }
```

`getComputedStyle(el).display` should return `"none"` for any element matching one of these rules; `el.checkVisibility()` should return `false`. Because Lightpanda always has scripting enabled, the `@media (scripting)` block applies unconditionally to `<noscript>`.

```mermaid
sequenceDiagram
    participant LP as Lightpanda
    Client->>LP: Runtime.evaluate getComputedStyle(scriptEl).display
    LP->>SM: hasDisplayNone(scriptEl)
    SM->>SM: matches UA tag-name rule
    SM-->>LP: true
    LP-->>Client: "none"
```

## Reproducer

Self-contained: pure CDP + a Node `ws` driver. No Ruby, no framework, no `chrome-remote-interface`.

`repro.html`:

```html
<!doctype html>
<html>
<head>
<title>UA stylesheet probe</title>
<meta charset="utf-8">
<style id="s1">.x { color: red }</style>
<script id="sc1">/* head script */</script>
<link id="lk1" rel="canonical" href="/">
<noscript id="ns1">no-script-1</noscript>
</head>
<body>
<p id="p1">visible paragraph</p>
<input type="hidden" id="ih1" name="csrf" value="abc">
<input type="text" id="it1" name="username">
<template id="tp1"><div>template content</div></template>
<details id="d1">
  <summary id="sm1">summary closed</summary>
  <p id="dc1">collapsed-when-closed</p>
</details>
<details id="d2" open>
  <summary id="sm2">summary open</summary>
  <p id="dc2">visible-when-open</p>
</details>
<script id="sc2">/* body script */</script>
</body>
</html>
```

`probe.js` (uses `ws` directly because `chrome-remote-interface` doesn't fit Lightpanda's `/json/list` shape):

```js
const http = require('http');
const fs = require('fs');
const path = require('path');

function makeClient(url) {
  let nextId = 1;
  const pending = new Map();
  ws.on('message', (raw) => {
      pending.delete(m.id);
      else resolve(m.result);
    }
  });
  return {
    send(method, params, sessionId) {
      const id = nextId++;
        ws.send(JSON.stringify({ id, method, params: params || {}, sessionId }));
      });
    },
    close() { ws.close(); },
  };
}

(async () => {
  const html = fs.readFileSync(path.join(__dirname, 'repro.html'));
  const srv = http.createServer((_, res) => { res.writeHead(200, {'Content-Type':'text/html'}); res.end(html); });
  await new Promise(r => srv.listen(8124, r));

  const c = makeClient('ws://127.0.0.1:9222/');
  await c.ready();
  const { targetId } = await c.send('Target.createTarget', { url: 'http://127.0.0.1:8124/' });
  const { sessionId } = await c.send('Target.attachToTarget', { targetId, flatten: true });
  await c.send('Page.enable', {}, sessionId);
  await c.send('Runtime.enable', {}, sessionId);

  // Settle: poll readyState === 'complete'
    const r = await c.send('Runtime.evaluate', { expression: 'document.readyState', returnByValue: true }, sessionId);
  }

  const r = await c.send('Runtime.evaluate', {
    expression: `(() => {
      const ids = ['sc1','sc2','s1','lk1','ns1','ih1','tp1','dc1'];
      const out = {};
      for (const id of ids) {
        const el = document.getElementById(id);
        const cs = window.getComputedStyle(el);
        out[id] = { tag: el.tagName, display: cs.display, checkVisibility: el.checkVisibility() };
      }
      out.head = { display: window.getComputedStyle(document.head).display, checkVisibility: document.head.checkVisibility() };
      return out;
    })()`,
    returnByValue: true,
  }, sessionId);

  console.log(JSON.stringify(r.result.value, null, 2));

  let failures = 0;
  for (const [id, v] of Object.entries(r.result.value)) {
    if (v.display !== 'none' || v.checkVisibility !== false) {
      console.log(`FAIL ${id} (${v.tag || 'HEAD'}): display=${v.display}, checkVisibility=${v.checkVisibility}`);
      failures++;
    }
  }
  c.close();
  srv.close();
  process.exit(failures > 0 ? 1 : 0);
```

`repro.sh`:

```bash
#!/bin/bash
set -u
DIR="$(cd "$(dirname "$0")" && pwd)"
trap '[[ -n "${LP_PID:-}" ]] && kill "$LP_PID" 2>/dev/null; wait 2>/dev/null' EXIT
lsof -ti :9222 2>/dev/null | xargs -r kill -9 2>/dev/null || true
sleep 0.2
"$LIGHTPANDA_BIN" serve --host 127.0.0.1 --port 9222 >/tmp/lp.log 2>&1 &
LP_PID=$!
for _ in $(seq 1 50); do
  curl -fs http://127.0.0.1:9222/json/version 2>/dev/null | grep -qi 'lightpanda' && break
  sleep 0.1
done
cd "$DIR"
[[ -d node_modules/ws ]] || npm install --no-save ws >/dev/null 2>&1
node "$DIR/probe.js"
```

**Run**: `bash repro.sh`

- Today (against `1.0.0-nightly.5816+a578f4d6` / build `2026.04.27.033406`): prints `FAIL` for 9 distinct cases (`<head>`, `<script>` x2, `<style>`, `<link>`, `<noscript>`, `<input type="hidden">`, `<template>`, closed-`<details>` non-summary child) and exits `1`.
- Expected: prints the JSON snapshot with `display: "none"` and `checkVisibility: false` for every hidden case, and exits `0`.

## Likely fix location

`src/browser/StyleManager.zig` — `hasDisplayNone` / `isElementHidden`. The cleanest place is to add a `matchesUaDisplayNoneRule(el)` helper that checks the `[hidden]` attribute, the unrendered tag-name list (via a new `Tag.isHiddenByUaStylesheet()` predicate on `Element.Tag`), `input[type="hidden" i]`, and the closed-`<details>` parent rule, then short-circuit `isElementHidden` on it before the user-CSS rule walker. Routing through `isElementHidden` rather than only `hasDisplayNone` gets `el.checkVisibility()` and the ancestor walk in `isHidden` to honor the same UA truth.

The spec's `basefont`, `noembed`, `noframes`, `rp` tags aren't represented in `Element.Tag` (obsolete) and would need separate handling if desired; in practice those tags are vanishingly rare.

## Environment

- Lightpanda: `1.0.0-nightly.5816+a578f4d6` (build `2026.04.27.033406`), `main` HEAD `827626db` at filing time.
- OS: macOS aarch64 (the bug is platform-independent — pure JS-API behavior).
- CDP client: raw `ws` (Node 18+).

navidemad:
## What

`StyleManager.hasDisplayNone` (and through it `el.checkVisibility()`, `getBoundingClientRect()`, click hit-testing) honored only the `[hidden]` UA-stylesheet rule. The rest of [HTML Rendering §15.3.1 "Hidden elements"](https://html.spec.whatwg.org/multipage/rendering.html#hidden-elements) — the unrendered tag-name list, `input[type="hidden" i]`, and `details:not([open]) > *:not(summary)` — was unimplemented, so `getComputedStyle(scriptEl).display` reported `"block"` and `checkVisibility()` returned `true` for elements that should never render.

Closes #2293.

## Root cause

`hasDisplayNone` short-circuited on `[hidden]` but otherwise deferred to `isElementHidden`, which only matched user/author CSS rules — no UA tag-name defaults, no input-type rule, no closed-`<details>` parent rule. The same `isElementHidden` is used by the ancestor-walking `isHidden` (via `el.checkVisibility()`), so the gap surfaced identically through both `getComputedStyle()` and `checkVisibility()`. The existing `[hidden]` short-circuit lived in `hasDisplayNone` only, so `el.checkVisibility()` even bypassed that — a `<div hidden>` reported as visible.

```mermaid
flowchart LR
    A[getComputedStyle el display] --> B[hasDisplayNone]
    B --> C[isElementHidden]
    C --> D[walk only user CSS rules]
    D --> E[returns block]
    style D fill:#fdd
    style E fill:#fdd
```

## Fix

Centralize the UA-stylesheet truth in a new `matchesUaDisplayNoneRule` helper, consulted at the top of `isElementHidden` so both `hasDisplayNone` and `isHidden` paths honor it. The helper handles the `[hidden]` attribute (moving the existing short-circuit out of `hasDisplayNone`), the unrendered tag-name list via a new `Tag.isHiddenByUaStylesheet()` predicate on `Element.Tag`, the case-insensitive `input[type="hidden" i]` rule, and the `details:not([open]) > *:not(summary)` parent-relationship rule. Inline `display` still wins per CSS cascade — the UA check is skipped when `display_priority == INLINE_PRIORITY`.

- `src/browser/webapi/Element.zig` — `Tag.isHiddenByUaStylesheet`: new predicate over `area, base, datalist, head, link, meta, noscript, param, script, source, style, template, title, track`. The spec also lists `basefont, noembed, noframes, rp` but those tags aren't represented in the enum (obsolete) and fall through to `.unknown`/`.custom`; documented in the comment.
- `src/browser/StyleManager.zig` — `matchesUaDisplayNoneRule`: per-element UA-rule check. Called from `isElementHidden` after the inline-style block; `hasDisplayNone` no longer needs its own `[hidden]` short-circuit.
- `src/cdp/domains/dom.zig` — adjust the `getBoxModel` test fixture to query `<p>` instead of `<head>` (head is now correctly `display: none` per spec, so its box model is all zeros).

```mermaid
flowchart LR
    A[getComputedStyle el display] --> B[hasDisplayNone]
    B --> C[isElementHidden]
    C --> D[matchesUaDisplayNoneRule]
    D --> E[returns none]
    style D fill:#dfd
    style E fill:#dfd
```

## Test

- **Unit tests** — `src/browser/tests/element/check_visibility.html` gains five new fixture blocks: `ua_unrendered_tags_are_display_none` (tag list), `ua_input_type_hidden_is_display_none` (case-insensitive type matching), `ua_closed_details_hides_non_summary_children` (closed-`<details>` rule plus the open-then-closed-again toggle and grandchild ancestor walk), `hidden_attribute_propagates_through_check_visibility` (the missing ancestor walk for `[hidden]`), and `inline_display_overrides_ua_default` (cascade ordering). All five fail on `main`, pass with this commit.
- **Reproducer** — pure CDP + `ws` + a single HTML fixture demonstrates 9 distinct violations (full script in #2293). Exits `1` on `1.0.0-nightly.5816+a578f4d6` / build `2026.04.27.033406`, exits `0` with this commit applied locally.
- **Full suite** — `mise exec -- zig build test $V8`: 490/490 passing.

## Notes

- Inline `display` still wins (cascade-correct). UA defaults short-circuit otherwise, mirroring the existing `[hidden]` behavior. A future change could thread UA rules through the rule-priority machinery instead, so author CSS like `script { display: block }` could override the UA rule, but that's exceptionally rare in practice and orthogonal to this fix.
- `Tag.isHiddenByUaStylesheet` deliberately omits the obsolete `basefont`, `noembed`, `noframes`, `rp` tags — they're not in `Element.Tag` and would need separate handling. Documented in the predicate's comment.
- The `getBoxModel` test was the only fallout in the existing suite. The expected coordinate change reflects deterministic-layout coordinates for the `<p>` node rather than `<head>` — the test's intent (non-zero box model for an element with content) remains intact.


src/browser/StyleManager.zig
```diff
@@ -234,14 +234,45 @@ pub fn isHidden(self: *StyleManager, el: *Element, cache: ?*VisibilityCache, opt
...
+        if (el.parentElement()) |parent| {
+            if (parent.getTag() == .details and !parent.hasAttributeSafe(comptime .wrap("open"))) {
+                return true;
+            }
+        }
+    }
+
+    return false;
+}
+
```
arrufat: Maybe we can tighten the comment a bit here, and also reorder stuff to improve perf. Something like this:
```suggestion
/// Centralizes UA-stylesheet display:none truth so `getComputedStyle().display`
/// (via `hasDisplayNone`) and `el.checkVisibility()` (via `isHidden`) agree.
/// Spec: HTML Rendering §15.3.1 "Hidden elements".
fn matchesUaDisplayNoneRule(el: *Element) bool {
    // Tag check first: O(1) switch, exits for the ~95% of elements with
    // ordinary tags before we touch the attribute list.
    const tag = el.getTag();
    if (tag.isHiddenByUaStylesheet()) return true;

    if (el.hasAttributeSafe(comptime .wrap("hidden"))) return true;

    // input[type="hidden" i] { display: none !important }
    // _input_type is parsed case-insensitively at attribute-set time.
    if (tag == .input) {
        if (el.is(Input)) |input| {
            if (input._input_type == .hidden) return true;
        }
    }

    // details:not([open]) > *:not(summary) { display: none }
    if (tag != .summary) {
        if (el.parentElement()) |parent| {
            if (parent.getTag() == .details and !parent.hasAttributeSafe(comptime .wrap("open"))) {
                return true;
            }
        }
    }

    return false;
}
```

src/browser/StyleManager.zig
```diff
@@ -313,6 +344,19 @@ fn isElementHidden(self: *StyleManager, el: *Element, options: CheckVisibilityOp
...
 
+    // UA stylesheet display:none rules (HTML Rendering §15.3.1 "Hidden elements").
+    // Inline `display` already wins above (sets display_priority == INLINE_PRIORITY);
+    // skip the UA check in that case so author inline overrides keep working.
+    // Otherwise short-circuit return true: the spec lists `input[type=hidden]`
+    // and `noscript` as `!important` and the rest are normal-origin UA rules
+    // that author CSS could in principle override, but author overrides of
+    // `<script>`/`<head>`/closed-`<details>` children are exceptionally rare
+    // and matching the existing `[hidden]` short-circuit keeps the behavior
+    // consistent across all UA rules.
```
arrufat: Same here, maybe we can reduce the verbosity of the comment a bit. What do you think?


```suggestion
    // UA stylesheet display:none rules (HTML Rendering §15.3.1 "Hidden elements").
    // Skipped when an inline `display` is set so author overrides win per cascade.
    // All UA rules are short-circuited here; the spec marks some as `!important`
    // (`input[type=hidden]`, `noscript`) and others as normal-origin, but author
    // CSS overriding `<script>`/`<head>`/closed-`<details>` children is rare
    // enough that uniform treatment is acceptable.
```

src/browser/StyleManager.zig
navidemad: Adopted in `dfa0ba77`. Tag-check first matches the perf intent and `input._input_type == .hidden` is consistent with the `el.is(Input)` pattern already used in `EventManager`, `Form`, and `RadioNodeList`. Behavior-preserving — `_input_type` is parsed case-insensitively at attribute-set time, so it agrees with the previous `std.ascii.eqlIgnoreCase` path. Added `const Input = Element.Html.Input;` to the type aliases.

src/browser/StyleManager.zig
navidemad: Adopted in `dfa0ba77`. Tighter phrasing keeps the cascade rationale and the `!important` distinction without the verbosity.

navidemad:
Thanks for the review @arrufat! Both suggestions adopted in `dfa0ba77`:

- **`matchesUaDisplayNoneRule`** — tighter docstring, tag-check first (O(1) switch, exits before the attribute lookup for ordinary tags), and `el.is(Input) + input._input_type == .hidden` instead of the case-insensitive string compare on the raw `type` attribute. Matches the pattern already used in `EventManager`, `Form`, and `RadioNodeList`; behavior-preserving since `_input_type` is parsed case-insensitively at attribute-set time.
- **Comment block above the call site** — compressed while keeping the cascade rationale and the `!important` distinction.

Also merged `main` into the branch to pick up the recent http-client-layering changes (no conflicts).

Full unit suite (494 tests) passes locally.

arrufat (approved):

PR #2297, opened 2026-04-28
keep the existing page active until the pending one is loaded

### LINKED ISSUE #2187: Runtime.evaluate after a click-driven navigation fails with "Cannot find default execution context"
A `Runtime.evaluate` issued (without `contextId`) right after a click that triggers a same-tab navigation fails with `Cannot find default execution context`. Chrome handles the same
   script. Affects Puppeteer / chromedp / playwright code that does `page.evaluate(...)` immediately after `page.click(<link>)`.

 ### Reproduction

  ```js
  await page.goto('https://demo-browser.lightpanda.io/', { waitUntil: 'networkidle0' });
  await page.click("a[href='campfire-commerce/']");
  const res = await client.send('Runtime.evaluate', { expression: '1+1' });
```

  - Chrome → {"result":{"type":"number","value":2}}
  - Lightpanda → ProtocolError: Protocol error (Runtime.evaluate): Cannot find default execution context

  ### CDP trace around the click

  > Input.dispatchMouseEvent (mouseReleased)
  < Runtime.executionContextsCleared          ← old default destroyed
  < Page.frameStartedNavigating
  < Page.frameStartedLoading
  < Network.requestWillBeSent
  > Runtime.evaluate {"expression":"1+1"}
  < error: "Cannot find default execution context"

  No Page.frameNavigated / Runtime.executionContextCreated between the cleared event and the evaluate.

### PR DESCRIPTION
**Pull request by krichprollsch**
During a root navigation, we keep the existing page active until we get
the headers callback from the pending page. Then
Session.commitPendingPage makes the switch.

It delays the deinit of CPD execution context to handle JS execution in
the meantime.

Now session has an array of two pages, _active_idx points to the main
page.

Both active and pending pages share the same frame_id, it must remains
stable. So this PR adds a Request.protect_from_abort to avoid removing
the request form the pending page when deinit the previous active page.

fix #2187

### DIFF
--- a/src/browser/HttpClient.zig
+++ b/src/browser/HttpClient.zig
@@ -340,13 +340,14 @@ fn _abort(self: *Client, comptime abort_all: bool, frame_id: u32) void {
 
     if (comptime IS_DEBUG and abort_all) {
         // Even after an abort_all, we could still have transfers, but, at the
-        // very least, they should all be flagged as aborted.
+        // very least, they should all be flagged as aborted (or be a transfer
+        // we explicitly protected with protect_from_abort).
         var it = self.in_use.first;
         var leftover: usize = 0;
         while (it) |node| : (it = node.next) {
             const conn: *http.Connection = @fieldParentPtr("node", node);
             switch (conn.transport) {
-                .http => |transfer| std.debug.assert(transfer.aborted),
+                .http => |transfer| std.debug.assert(transfer.aborted or transfer.req.params.protect_from_abort),
                 .websocket => {},
                 .none => {},
             }
@@ -363,7 +364,8 @@ fn abortConnections(list: std.DoublyLinkedList, comptime abort_all: bool, frame_
         const conn: *http.Connection = @fieldParentPtr("node", node);
         switch (conn.transport) {
             .http => |transfer| {
-                if ((comptime abort_all) or transfer.req.params.frame_id == frame_id) {
+                const matches = (comptime abort_all) or transfer.req.params.frame_id == frame_id;
+                if (matches and !transfer.req.params.protect_from_abort) {
                     transfer.kill();
                 }
             },
@@ -877,6 +879,13 @@ pub const RequestParams = struct {
     credentials: ?[:0]const u8 = null,
     notification: *Notification,
     timeout_ms: u32 = 0,
+
+    // Set on an in-flight root-navigation transfer that was issued against a
+    // pending Page. The old Page's frame.deinit (called from Session.commit
+    // PendingPage when response headers arrive) calls http_client.abort() —
+    // that abort_all path skips transfers with this flag so the callback
+    // chain we are sitting inside isn't killed mid-flight.
+    protect_from_abort: bool = false,
 
     const ResourceType = enum {
         document,
--- a/src/browser/HttpClient.zig
+++ b/src/browser/HttpClient.zig
@@ -324,9 +324,10 @@ fn _abort(self: *Client, comptime abort_all: bool, frame_id: u32) void {
         while (n) |node| {
             n = node.next;
             const transfer: *Transfer = @fieldParentPtr("_node", node);
+            const params = transfer.req.params;
             if (comptime abort_all) {
                 transfer.kill();
-            } else if (transfer.req.params.frame_id == frame_id) {
+            } else if (params.frame_id == frame_id and !params.protect_from_abort) {
                 q.remove(node);
                 transfer.kill();
             }
@@ -339,15 +340,12 @@ fn _abort(self: *Client, comptime abort_all: bool, frame_id: u32) void {
     }
 
     if (comptime IS_DEBUG and abort_all) {
-        // Even after an abort_all, we could still have transfers, but, at the
-        // very least, they should all be flagged as aborted (or be a transfer
-        // we explicitly protected with protect_from_abort).
         var it = self.in_use.first;
         var leftover: usize = 0;
         while (it) |node| : (it = node.next) {
             const conn: *http.Connection = @fieldParentPtr("node", node);
             switch (conn.transport) {
-                .http => |transfer| std.debug.assert(transfer.aborted or transfer.req.params.protect_from_abort),
+                .http => |transfer| std.debug.assert(transfer.aborted),
                 .websocket => {},
                 .none => {},
             }
@@ -364,8 +362,9 @@ fn abortConnections(list: std.DoublyLinkedList, comptime abort_all: bool, frame_
         const conn: *http.Connection = @fieldParentPtr("node", node);
         switch (conn.transport) {
             .http => |transfer| {
-                const matches = (comptime abort_all) or transfer.req.params.frame_id == frame_id;
-                if (matches and !transfer.req.params.protect_from_abort) {
+                const params = transfer.req.params;
+                const matches = (comptime abort_all) or (params.frame_id == frame_id and !params.protect_from_abort);
+                if (matches) {
                     transfer.kill();
                 }
             },
--- a/src/browser/HttpClient.zig
+++ b/src/browser/HttpClient.zig
@@ -54,9 +54,9 @@ pub const InterceptionLayer = @import("../network/layer/InterceptionLayer.zig");
 //
 // The app has other secondary http needs, like telemetry. While we want to
 // share some things (namely the ca blob, and maybe some configuration
-// (TODO: ??? should proxy settings be global ???)), we're able to do call
-// client.abort() to abort the transfers being made by a frame, without impacting
-// those other http requests.
+// (TODO: ??? should proxy settings be global ???)), we're able to call
+// client.abortFrame() to abort the transfers being made by a frame, without
+// impacting those other http requests.
 pub const Client = @This();
 
 // Count of active ws requests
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

**Comment by krichprollsch** [comment]
I have to add a new e2e test
EDIT: https://github.com/lightpanda-io/demo/pull/169

**Comment by krichprollsch** [comment]
The branch works on my own test case, but fails on https://github.com/lightpanda-io/browser/issues/2187#issuecomment-4322739061 script.

EDIT: it works now :heavy_check_mark:

File: src/browser/Runner.zig [hunk 1]
```diff
@@ -141,6 +141,10 @@ pub fn tickCDP(self: *Runner, opts: TickOpts) !CDPTickResult {
 }
 
 fn _tick(self: *Runner, comptime is_cdp: bool, opts: TickOpts) !CDPTickResult {
+    // Refresh self.frame from session. In case of pending page, we want to
+    // take its state while loading. If we use only the current frame, we will
+    // return a .done result immediately.
+    self.frame = self.session.currentOrPendingFrame() orelse return .done;
```
krichprollsch: I'm not sure about that... But I don't really have a better solution :/

src/browser/Session.zig
```diff
@@ -81,7 +103,9 @@ pub fn init(self: *Session, browser: *Browser, notification: *Notification) !voi
     errdefer arena_pool.release(arena);
 
     self.* = .{
-        .page = null,
+        ._pages = .{ null, null },
```
karlseguin: I'm guilty of doing this too, but you both set default value sin the fields, then set those same defaults on init...

src/browser/Session.zig
```diff
@@ -65,9 +65,31 @@ arena_pool: *ArenaPool,
...
+// names the active vs pending slot.
+//
+// Why two slots: at any given moment we may need to hold one active Page
+// (the user-visible page) AND one pending Page (the in-flight destination
+// of a root navigation, kept alive across the HTTP round-trip per Chrome's
+// behavior). After commit, the OLD active slot is freed and becomes
+// available for the next pending allocation.
+//
+// Convention: a slot is "occupied" iff its `?Page` is non-null.
+_pages: [2]?Page = .{ null, null },
```
karlseguin: did you consider going with two fields and allocating page?

```zig
active: ?*Page,
pending: ?*Page,
```

src/browser/Session.zig
```diff
@@ -166,42 +263,29 @@ pub fn releaseArena(self: *Session, allocator: Allocator) void {
...
+// Runtime callers MUST NOT use this; it is only for the navigation
+// machinery (Frame.navigate / commitPendingPage).
+pub fn pendingPage(self: *Session) ?*Page {
+    const idx = self._pending_idx orelse return null;
+    return &self._pages[idx].?;
 }
 
-pub fn currentPage(self: *Session) ?*Page {
-    return &(self.page orelse return null);
+pub fn currentOrPendingFrame(self: *Session) ?*Frame {
```
karlseguin:
```suggestion
pub fn pendingOrCurrentFrame(self: *Session) ?*Frame {
```

src/browser/Session.zig
```diff
@@ -384,49 +468,182 @@ fn processPopupNavigation(self: *Session, frame: *Frame, qn: *QueuedNavigation)
...
 
-    // create a copy before the frame is cleared
+    // Detach the QueuedNavigation. Whether we keep it on the active frame
+    // (synthetic path) or transfer it to the pending frame (HTTP path), the
+    // current frame must no longer claim it.
     const qn = current_frame._queued_navigation.?;
     current_frame._queued_navigation = null;
 
+    // Synthetic navigations (about:blank, blob:) commit instantly — no HTTP,
+    // so there is no in-flight window to worry about. Use the legacy
```
karlseguin: The fact that it's "legacy" isn't relevant. We normally say something is legacy to explain why a behavior is the way it is. But here, `replaceRootImmediate` doesn't exist for some "legacy" reason..it's just a more optimized / simpler (maybe more correct?) path to take for these cases. Or am I misunderstanding and `replaceRootImmediate` exists only to keep an pre-existing behavior?

src/browser/Session.zig
```diff
@@ -384,49 +468,182 @@ fn processPopupNavigation(self: *Session, frame: *Frame, qn: *QueuedNavigation)
...
+    // immediate-swap path for them.
+    const is_synthetic = std.mem.eql(u8, qn.url, "about:blank") or
```
karlseguin: `qn.is_about_blank`  

We need to check this when `QueuedNavigation` is created (in `Frame.scheduleNavigationWithArena`) and I know it was checked again during actual navigation, so I figured might as well save the boolean.

src/browser/Session.zig
```diff
@@ -384,49 +468,182 @@ fn processPopupNavigation(self: *Session, frame: *Frame, qn: *QueuedNavigation)
...
+//   5. OLD Page.deinit + free LAST. Its frame.deinit calls
+//      http_client.abortFrame(frame_id) on the frame_id that the OLD
+//      page shares with the now-active pending page; the in-flight
+//      navigation transfer (whose callback we are inside) is shielded
+//      by protect_from_abort, which abortFrame's default .normal scope
+//      honors. The caller clears the flag AFTER we return.
+pub fn commitPendingPage(self: *Session) !void {
+    const pending_idx = self._pending_idx orelse {
+        lp.assert(false, "Session.commitPendingPage - no pending page", .{});
+        return error.NoPendingPage;
```
karlseguin: unreachable. 

The intent here is unclear. Do you want the assert only in debug mode?   (this pattern happens 3 or so times in this file, I think they should all be cleaned up).

src/browser/Session.zig
```diff
@@ -384,49 +468,182 @@ fn processPopupNavigation(self: *Session, frame: *Frame, qn: *QueuedNavigation)
...
+    };
+    const old_idx = self._active_idx orelse {
```
karlseguin: old_active_idx?   just a big more clear later on when you do:

```zig
    self._pages[old_active_idx].?.deinit();
    self.freeSlot(old_active_idx);
```

(pretty minor, I'd probably have named it `old_idx` too!)

src/browser/HttpClient.zig
```diff
@@ -878,6 +890,15 @@ pub const RequestParams = struct {
...
     timeout_ms: u32 = 0,
 
+    // Set on an in-flight root-navigation transfer that was issued against a
+    // pending Page. The old Page's frame.deinit (called from Session.commit
+    // PendingPage when response headers arrive) calls abortFrame() on the
+    // shared frame_id; abortFrame's default .normal scope skips transfers
+    // with this flag so the callback chain we are sitting inside isn't killed
+    // mid-flight. Session.discardPendingPage uses .full scope to override
+    // the flag in failure paths.
+    protect_from_abort: bool = false,
```
karlseguin, line 900: This is clever, and it makes sense. BUT, I think if we do give HttpClient a major cleanup (e.g. merging Request and Transfer), I'd try to come up with a better way to track ownership of frame <-> request.

krichprollsch: Yeah, this trick is only because  I must share the same frame_id between actually 2 different frames...
And I dislike it...

src/browser/Session.zig
krichprollsch: good catch, I did move that after a rebase... It was closer to the deinit at first.

src/browser/Session.zig
krichprollsch: good, point, Claude did see that too and I didn't fix it :facepalm:

src/browser/Session.zig
krichprollsch: you are correct.

src/browser/Session.zig
krichprollsch: I removed from init, but I'm not sure it it's better this way or to remove default from fields?

src/browser/Session.zig
krichprollsch: I did, and I disliked it :thinking:  But I didn't remember exactly why. I'm trying again.
You think it's better to allocate to avoid these empty structs in the stack?

karlseguin: I think it's minor.  A few places that are currently 2 lines ((1) get the index (2) get the page) become 1. That's pretty much it.

```zig
pub fn currentPage(self: *Session) ?*Page {
    const idx = self._active_idx orelse return null;
    return &self._pages[idx].?;
}
```

becomes 

```zig
pub fn currentPage(self: *Session) ?*Page {
    return self.active;
}
```

Or, if you look at the start of`initRootNavigation`, it becomes:

```
    self.discardPendingPage(); // can do const page = self.pending orelse return;

    const page = try self.pageInit(frame_id);
    errdefer page.deinit();

    page._state = .pending;
    self.pending = page;
    errdefer self.pending = null
    ..
```


(again, quite minor)

src/browser/Session.zig
karlseguin: I think either is better than both :)

field defaults have the advantage of letting you create an empty struct without an init. But, when that's not an option (like here), the practical different is small. I do think field defaults might better improve documentation / discoverability. If I want to understand a field, I'll go scroll up to where it's defined, hope for a comment, and the default might provide more insight too.

src/browser/Session.zig
krichprollsch: ok, it's simpler after all. But where should I allocate the pages? Claude suggests to use app.allocator directly, but not sure about that. WDYT?

karlseguin: Ya, or a dedicated `std.heap.MemoryPool(Page)` on the session if you think we'll create a lot. Or on the Browser if you think we'll create a lot of Sessions...

karlseguin (approved):

PR #2299, opened 2026-04-28
Initial idn support

karlseguin:
Links to libidn2 and builds libcurl with it. This makes libcurl work, and by extension browser, work on international domain names, e.g.

zig build run -- fetch "https://räksmörgås.se/"

With it available, we can use it in our WebAPIs which should also support these domains, e.g:
  testing.expectEqual('xn--rksmrgs-5wao1o.se', new URL('https://räksmörgås.se').hostname);

There is more integration to be done here, but this is a first step.

claude wrote all of the build.zig code.

I don't have a strong opinion about this feature, I just dislike that our WPT /url/* tests are at 1704 / 9095 and, this is the biggest chunk (although, this specific commit just does the basic integration and probably won't fix too many WPT cases directly).

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -34,6 +34,10 @@
             .url = "https://github.com/allyourcodebase/sqlite3/archive/8f840560eae88ab66668c6827c64ffbd0d74ef37.tar.gz",
             .hash = "sqlite3-3.51.0-DMxLWssOAABZ8cAvU_LfBIbp0kZjm824PU8sSLXpEDdr",
         },
+        .libidn2 = .{
+            .url = "https://ftp.gnu.org/gnu/libidn/libidn2-2.3.8.tar.gz",
+            .hash = "N-V-__8AABGOuAC_dhAN07kfoP4dycCFi8Bka4O-tuhriNH8",
+        },
     },
     .paths = .{""},
 }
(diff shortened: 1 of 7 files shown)


build.zig.zon
```diff
@@ -34,6 +34,10 @@
             .url = "https://github.com/allyourcodebase/sqlite3/archive/8f840560eae88ab66668c6827c64ffbd0d74ef37.tar.gz",
             .hash = "sqlite3-3.51.0-DMxLWssOAABZ8cAvU_LfBIbp0kZjm824PU8sSLXpEDdr",
         },
+        .libidn2 = .{
+            .url = "https://ftp.gnu.org/gnu/libidn/libidn2-2.3.8.tar.gz",
```
krichprollsch, line 38: Should we consider to use GH instead?
https://github.com/libidn/libidn2/releases/tag/v2.3.8

krichprollsch (approved):

build.zig.zon
karlseguin: They are different. The gnu package is self contained and includes everything we need to configure/build (or, at least, enough for our needs). The libidn2 is purely the libidn source tree. So, for example, if we did this from github, we'd also have to fetch and manage a libunistring dependency.

PR #2300, opened 2026-04-28
fix: accept CDP SameSite cookie casing

guyua9:
## Summary
- Load cookie JSON `sameSite` values case-insensitively, so CDP-style `Lax` / `Strict` / `None` values are accepted.
- Keep saved cookie output on the existing enum-backed path.
- Add a regression test for capitalized `SameSite` input.

## Why
The cookie loader says it matches the CDP `Network.Cookie` format used by Puppeteer and Playwright. Those tools commonly serialize `sameSite` as capitalized values such as `Lax`. Parsing directly into the lowercase Zig enum rejects that JSON and prevents the cookie file from loading.

## Tests
- `zig fmt --check src/cookies.zig`
- `git diff --check`
- Standalone Zig regression for the JSON parsing path passed locally.
- `zig build test --summary all` could not complete locally because Zig package fetching returned `HttpConnectionClosing` for upstream dependencies.

### DIFF
--- a/src/cookies.zig
+++ b/src/cookies.zig
@@ -43,7 +43,7 @@ fn _loadFromFile(session: *Session, path: []const u8) !void {
         return;
     };
 
-    const json_cookies = std.json.parseFromSliceLeaky([]const JsonCookie, arena, content, .{
+    const json_cookies = std.json.parseFromSliceLeaky([]const LoadJsonCookie, arena, content, .{
         .ignore_unknown_fields = true,
     }) catch |err| {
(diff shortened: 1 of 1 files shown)


guyua9:

guyua9:
recheck

karlseguin:
Hi. This is nice. But I think you could avoid the 2 structures, use ` sameSite: ?[]const u8 = null,` and, on "save", use:

```zig
.sameSite = @tagName(c.same_site),
```

PR #2301, opened 2026-04-28
HTTP Cache Testing + CDP

mookums:
This does a few things. It mostly adds a few CDP Events that allow us to better test caching behavior. It is correlated to the `demo` branch of the same name (`http-cache-cdp`).

This adds:
- `clear` fn to Cache to clear out all of the entries
- `Network.requestServedFromCache` CDP Event that gets emitted by `InterceptionLayer` when request is `.cached`.
- `fromDiskCache` field on `Network.Response` CDP type that signifies that the Response is from Cache.
- `Network.setCacheDisabled` CDP method that allows you to disable putting and getting from the Cache. This uses a simple flag in `CacheLayer`.
- Fulfilled Requests from CDP do not cache (or more generically, only responses from the Network (`.transfer`) are cached).
- `InterceptionLayer` is only enabled on `.serve` mode.

### DIFF
--- a/src/network/cache/FsCache.zig
+++ b/src/network/cache/FsCache.zig
@@ -265,6 +265,9 @@ pub fn put(self: *FsCache, meta: CachedMetadata, body: []const u8) !void {
 }
 
 pub fn clear(self: *FsCache) !void {
+    for (&self.locks) |*lock| lock.lock();
+    defer for (&self.locks) |*lock| lock.unlock();
+
     var iter = self.dir.iterate();
     while (try iter.next()) |entry| {
         if (entry.kind != .file) continue;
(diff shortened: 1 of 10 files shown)


src/network/cache/FsCache.zig
```diff
@@ -264,6 +264,19 @@ pub fn put(self: *FsCache, meta: CachedMetadata, body: []const u8) !void {
     log.debug(.cache, "put", .{ .url = meta.url, .hash = &hashed_key, .body_len = body.len });
 }
 
+pub fn clear(self: *FsCache) !void {
+    var iter = self.dir.iterate();
```
karlseguin, line 271: No locks

karlseguin (commented):
clear needs locking.

These changes are worth adding as actual features. But, for testing via the demo repo, I think you get a lot more flexibility by simply controlling the cache-control header of the response. It gives you per-request control and lets you inject invalid/weird/edge values.

mookums:
Depends on https://github.com/lightpanda-io/demo/pull/168 to be merged first.

mookums:
Going to split this PR into smaller ones.

PR #2303, opened 2026-04-28
WIP: common libcurl

mrdimidium:
This PR finally moves network requests to a shared event pool.

In serve mode, Lightpanda creates a Network object on the main thread that owns the ws listener and libcurl multi handler. A separate thread with separate CDP and Browser objects is created for each CDP connection. A Network.Handle object is created for each thread, which services the current thread's requests, passes them to the Network, and manages synchronization.

HttpClient acts as a small wrapper around Network.Handle, adding browser-specific logic (caches, robots, etc.). XMLHttpRequest/Fetch/WebSocket, in turn, wrap HttpClient in a JavaScript API.

Future changes: Currently, CDP and pages are tightly coupled to a single thread. To create multiple pages in a single connection and maintain a detachable browse context, we'll have to move CDP to the main thread (even though it's not part of Network), and leave only Browser on worker threads.

It would also be helpful to clean up the Network<->HttpClient<->Connection relationship. Currently, request state is spread across Transfer, Connection, and queues in Handles, while WebSocket depends on HttpClient (even though it only needs a Handle pointer). We need to move the libcurl logic entirely to Handles, leaving the operations in http.Connection (since not all libcurl calls are thread-safe), and move some network logic (such as redirects and buffer ownership) to Network.

```mermaid
flowchart
  Server == "spawn(handleConnection)" ==> WorkerT

  Handle == "submit*<br/>(pending_add/remove/ops)" ==> Network
  Network == "on_complete<br/>(pushCompletion)" ==> Handle

  subgraph MainT["Main thread — Network.run()"]
    Network["Network<br/>(libcurl multi, accept)"]
    Server["Server<br/>(onAccept)"]
    Network -- "accept → onAccept" --> Server
  end

  subgraph WorkerT["CDP worker thread × N"]
    CDP --> Browser
    CDP --> Handle
    Browser --> HttpClient
    HttpClient --> Handle
  end
```

### DIFF
--- a/src/browser/Browser.zig
+++ b/src/browser/Browser.zig
@@ -70,6 +70,10 @@ pub fn nextFrameId(self: *Browser) u32 {
     return id;
 }
 
+pub fn httpClient(self: *Browser) *HttpClient {
+    return &self.http_client;
+}
+
 pub fn init(self: *Browser, app: *App, opts: InitOpts, cdp: ?*CDP) !void {
     const allocator = app.allocator;
 
--- a/src/browser/HttpClient.zig
+++ b/src/browser/HttpClient.zig
@@ -809,7 +809,7 @@ fn processOneMessage(self: *Client, msg: Network.Handle.Completion, transfer: *T
     if (body.len > 0) {
         try transfer.req.data_callback(Response.fromTransfer(transfer), body);
 
-        if (transfer.state == .aborted) {
+        if (transfer.isAborted()) {
             transfer.requestFailed(error.Abort, true);
             return true;
         }
@@ -1095,6 +1095,7 @@ pub const Transfer = struct {
     client: *Client,
 
     start_time: u64,
+    aborted: std.atomic.Value(bool) = .init(false),
 
     _notified_fail: bool = false,
 
@@ -1199,6 +1200,15 @@ pub const Transfer = struct {
         arena_pool.release(arena);
     }
 
+    pub fn isAborted(self: *const Transfer) bool {
+        return self.aborted.load(.acquire);
+    }
+
+    fn setAborted(self: *Transfer) void {
+        self.aborted.store(true, .release);
+        self.state = .aborted;
+    }
+
     // Cancel this transfer with `err`. Fires error_callback once (latched
     // via _notified_fail), then either deinits synchronously or, if we're
     // mid-perform with a libcurl handle still in the multi, detaches and
@@ -1237,8 +1247,8 @@ pub const Transfer = struct {
     // nothing left referencing this transfer and we can safely deinit
     // inline.
     fn detachOrDeinit(self: *Transfer) void {
+        if (self.isAborted()) return;
         switch (self.state) {
-            .aborted => return,
             .completing => self.detachInPerform(),
             .inflight => {
                 if (self._conn) |conn| {
@@ -1266,9 +1276,7 @@ pub const Transfer = struct {
     //     `owner == null` and skips the list-remove that would otherwise
     //     UAF against a freed list.
     fn detachInPerform(self: *Transfer) void {
-        // `_conn` (if any) rides through .aborted untouched; deinit
-        // releases it once libcurl is done.
-        self.state = .aborted;
+        self.setAborted();
         self.req.start_callback = null;
         self.req.shutdown_callback = null;
         self.req.header_callback = Noop.headerCallback;
@@ -1546,7 +1554,7 @@ pub const Transfer = struct {
             return err;
         };
 
-        return proceed and transfer.state != .aborted;
+        return proceed and !transfer.isAborted();
     }
 
     fn dataCallback(buffer: [*]const u8, chunk_count: usize, chunk_len: usize, data: *anyopaque) usize {
@@ -1596,7 +1604,7 @@ pub const Transfer = struct {
             return http.writefunc_error;
         };
 
-        if (transfer.state == .aborted) {
+        if (transfer.isAborted()) {
             return http.writefunc_error;
         }
 
--- a/src/browser/HttpClient.zig
+++ b/src/browser/HttpClient.zig
@@ -65,8 +65,7 @@ pub const HeaderIterator = http.HeaderIterator;
 // impacting those other http requests.
 pub const Client = @This();
 
-// Our curl multi handle. Owns in_use/ready_queue/http_active/ws_active/
-// performing — see Network.Handle.
+// Our curl multi handle. Owns in_use/http_active/ws_active — see Network.Handle.
 handle: Network.Handle,
 
 // WebSockets with queued events to be drained from the worker thread.
@@ -214,6 +213,30 @@ pub fn init(self: *Client, allocator: Allocator, network: *Network, cdp: ?*CDP) 
 
 pub fn deinit(self: *Client) void {
     self.abort();
+
+    // Cancellations submitted by abort flow through the network thread
+    // and come back as canceled completions. If the network thread has
+    // already stopped, drive its queues ourselves; otherwise spin until
+    // they all arrive and we drained them.
+    var spins: usize = 0;
+    while (self.handle.in_use.first != null and spins < 1000) : (spins += 1) {
+        if (self.network.shutdown.load(.acquire)) {
+            self.network.drainPendingForShutdown();
+        }
+        _ = self.processMessages() catch {};
+        self.drainReadyWs();
+        if (self.handle.in_use.first == null) break;
+        std.Thread.sleep(std.time.ns_per_ms);
+    }
+    if (self.handle.in_use.first != null) {
+        log.warn(.http, "deinit with active conns", .{
+            .count = self.handle.http_active + self.handle.ws_active,
+        });
+    } else if (comptime IS_DEBUG) {
+        std.debug.assert(self.transfers.size == 0);
+        std.debug.assert(self.queue.first == null);
+    }
+
     self.handle.deinit();
 
     self.ws_ready.deinit(self.allocator);
@@ -269,14 +292,8 @@ pub fn setTlsVerify(self: *Client, verify: bool) !void {
 
     var it = self.handle.in_use.first;
     while (it) |node| : (it = node.next) {
-        const conn: *http.Connection = @fieldParentPtr("node", node);
-        try self.handle.submitTlsVerify(conn, verify, self.use_proxy);
-    }
-
-    it = self.handle.ready_queue.first;
-    while (it) |node| : (it = node.next) {
-        const conn: *http.Connection = @fieldParentPtr("node", node);
-        try self.handle.submitTlsVerify(conn, verify, self.use_proxy);
+        const conn: *http.Connection = @fieldParentPtr("_worker_node", node);
+        self.handle.submitTlsVerify(conn, verify, self.use_proxy);
     }
 
     self.tls_verify = verify;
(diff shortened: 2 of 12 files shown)


karlseguin:
Gonna try to get through it tomorrow, but as a quick test:

```
(wp/mrdimidium/common-libcurl✔) make test | grep 'WARN' | wc -l
       6
vs

(main✔) make test | grep 'WARN' | wc -l
       0
```

These aren't test failures, but they're warnings that I feel we shouldn't suddenly be getting. All 6 are either "submitRemove bad state" or a couple different "curl transfer error"

karlseguin:
Do we have benchmarks / memory for the cases where it's supposed to be the most benefit (e.g. concurrent connections)?

karlseguin:
Did you consider moving the CDP socket into the network thread and treating it as just another completion?

src/network/WsConnection.zig
```diff
@@ -305,301 +306,429 @@ pub fn Reader(comptime EXPECT_MASK: bool) type {
...
 
-pub const WsConnection = struct {
-    // CLOSE, 2 length, code
-    const CLOSE_NORMAL = [_]u8{ 136, 2, 3, 232 }; // code: 1000
-    const CLOSE_GOING_AWAY = [_]u8{ 136, 2, 3, 233 }; // code: 1001
-    const CLOSE_TOO_BIG = [_]u8{ 136, 2, 3, 241 }; // 1009
-    const CLOSE_PROTOCOL_ERROR = [_]u8{ 136, 2, 3, 234 }; //code: 1002
-    // "private-use" close codes must be from 4000-49999
-    const CLOSE_TIMEOUT = [_]u8{ 136, 2, 15, 160 }; // code: 4000
+pub const WsConnection = @This();
```
karlseguin: This isn't necessarily a WebSocket connection (it starts off as an HTTP connection, answers HTTP requests and might never upgrade to a WebSocket), and it has enough CDP-specific knowledge that it belongs in cdp/Connection.zig  (or cdp/Client.zig or something).

The Reader could be extracted into network/WsReader.zig, since that's generic websocket stuff.

src/network/WsConnection.zig
```diff
@@ -305,301 +306,429 @@ pub fn Reader(comptime EXPECT_MASK: bool) type {
...
+const CLOSE_TIMEOUT = [_]u8{ 136, 2, 15, 160 }; // code: 4000
+
+socket: posix.socket_t,
+socket_flags: usize,
+reader: Reader(true),
+send_arena: ArenaAllocator,
+json_version_response: []const u8,
+
+pub fn init(
+    self: *WsConnection,
```
karlseguin: From where this is used, not sure why this takes a `*WsConnection`  versus returning a `WsConnection`, nothing here needs a stable pointer.

src/network/WsConnection.zig
```diff
@@ -305,301 +306,429 @@ pub fn Reader(comptime EXPECT_MASK: bool) type {
...
-            pos += written;
+pub const HttpResult = enum { more, upgraded, close };
+
+pub fn handshake(self: *WsConnection) !bool {
+    // Liveness is enforced by TCP keepalive configured in
+    // Server.setTcpKeepalive; a dead peer surfaces as a poll error or
+    // EOF from read(). The poll blocks for ~24 days rather than tracking
+    // an app-level timeout. Capped at i32-max because posix.poll narrows
+    // to c_int.
+    const wait_ms: i32 = std.math.maxInt(i32);
```
karlseguin: We should be using a sane timeout for handshake. As-is, a silent client consumes a slot + resources.

src/network/WsConnection.zig
```diff
@@ -305,301 +306,429 @@ pub fn Reader(comptime EXPECT_MASK: bool) type {
...
+    while (true) {
+        var pfds = [_]posix.pollfd{.{
+            .fd = self.socket,
+            .events = posix.POLL.IN,
+            .revents = 0,
+        }};
+        const n = try posix.poll(&pfds, wait_ms);
```
karlseguin: Could this be done in the network loop? I'm surprised to find any network I/O happening in the worker thread.

src/network/WsConnection.zig
```diff
@@ -305,301 +306,429 @@ pub fn Reader(comptime EXPECT_MASK: bool) type {
...
-    ) !void {
-        // Dangerous API!. We assume the caller has reserved the first 10
-        // bytes in `buf`.
-        const framed = fillWebsocketHeader(buf);
-        return self.send(framed);
+    if (std.mem.eql(u8, url, "/json/version") or std.mem.eql(u8, url, "/json/version/")) {
+        try self.send(self.json_version_response);
+        // Chromedp (a Go driver) does an http request to /json/version
+        // then to / (websocket upgrade) using a different connection.
+        // Since we only allow 1 connection at a time, the 2nd one (the
```
karlseguin: This should be revisited, since we no longer only allow 1 connection at a time. Removing this would speed up every other client.

src/cdp/CDP.zig
```diff
@@ -78,25 +81,36 @@ frame_arena: std.heap.ArenaAllocator,
...
         .frame_arena = std.heap.ArenaAllocator.init(allocator),
         .message_arena = std.heap.ArenaAllocator.init(allocator),
         .notification_arena = std.heap.ArenaAllocator.init(allocator),
         .browser_context_arena = std.heap.ArenaAllocator.init(allocator),
     };
+
+    try self.ws.init(socket, self.app.allocator, json_version_response);
+    errdefer self.ws.deinit();
+
+    try self.browser.init(app, .{ .env = .{ .with_inspector = true } }, .{
```
karlseguin: If the notes about ChromeDB are right, and it uses 2 connections, 1 to get the json version, then 1 to upgrade to websocket, than this is pretty wasteful for that specific case. But it does simplify things.

src/cdp/domains/page.zig
```diff
@@ -144,7 +144,7 @@ fn setLifecycleEventsEnabled(cmd: *CDP.Command) !void {
         try sendPageLifecycle(bc, "load", now, frame_id, loader_id);
 
         const http_client = frame._session.browser.http_client;
-        const http_active = http_client.http_active;
+        const http_active = http_client.handle.http_active;
```
karlseguin, line 151: line 146, making a copy of `http_client`

Since this is an easy mistake to make and http_client is loaded in _many_ places, what do you think about a

```zig
pub fn httpClient(self: *Browser) *HttpClient {
   return &self._http_client;
}
```

src/browser/Browser.zig
```diff
@@ -39,32 +39,33 @@ app: *App,
...
-http_client: *HttpClient,
+http_client: HttpClient,
 
 const InitOpts = struct {
     env: js.Env.InitOpts = .{},
-    http_client: *HttpClient,
 };
 
-pub fn init(app: *App, opts: InitOpts) !Browser {
+pub fn init(self: *Browser, app: *App, opts: InitOpts, cdp_client: ?HttpClient.CDPClient) !void {
```
karlseguin: `cdp_client` is optional, so it feels like it would fit nicely in `InitOpts`.

src/Server.zig
```diff
@@ -33,29 +33,29 @@ const Allocator = std.mem.Allocator;
...
 app: *App,
-allocator: Allocator,
 json_version_response: []const u8,
 
 // Thread management
 active_threads: std.atomic.Value(u32) = .init(0),
-clients: std.ArrayList(*Client) = .{},
-client_mutex: std.Thread.Mutex = .{},
-clients_pool: std.heap.MemoryPool(Client),
+pending: std.ArrayList(*CDP) = .{},
```
karlseguin: `pending` and `conns` are separated for the shutdown path? To know whether or not to send the close packet?  If WsConnection knew what phase it was in (which I think is reasonable), you could then have a single list..always calling `ws.close()` which would do nothing when `self.mode == .http`. Not a huge difference either way, was just trying to understand why the shuffling between the two lists.

karlseguin (commented):
I haven't gone through the heavy changes in `WebSocket`, `Network` and `HttpClient`. Sorry.

src/network/WsConnection.zig
mrdimidium: A failed client will be dropped via TCP keepalive

karlseguin: A failed client will, but not a silent one.  Start the serve, and then connect to it `socat - TCP:127.0.0.1:9222`

src/network/Network.zig
```diff
@@ -417,37 +455,69 @@ pub fn run(self: *Network) void {
...
 
-        if (self.multi) |multi| {
-            // Kickstart newly added handles (DNS/connect) so that
-            // curl registers its sockets before we poll.
-            libcurl.curl_multi_perform(multi, &running_handles) catch |err| {
-                lp.log.err(.app, "curl perform", .{ .err = err });
-            };
+        self.drainQueue();
 
-            self.preparePollFds(multi);
```
karlseguin, line 739: I don't think `preparePollFds` is being used anymore, but it's still in the code.

src/network/Network.zig
```diff
@@ -417,37 +455,69 @@ pub fn run(self: *Network) void {
...
+                extra_fds[extra_len] = .{
+                    .fd = listen_fd.fd,
+                    .events = .{ .pollin = true },
+                    .revents = .{},
+                };
+                extra_len += 1;
+                break :blk idx;
+            } else null;
+
+            libcurl.curl_multi_poll(self.multi, extra_fds[0..extra_len], timeout, null) catch |err| {
```
karlseguin: Can `curl_multi_wakeup` instead of the wakeup pipe?

src/network/Network.zig
```diff
@@ -496,46 +566,229 @@ pub fn run(self: *Network) void {
...
-        self.multi = m;
-        break :blk m;
+            conn._submission = .idle;
+            self.removeFromOpsLocked(conn);
+            to_remove.append(self.allocator, conn) catch @panic("OOM");
+        }
+    }
+    for (to_remove.items) |conn| self.handleRemove(conn);
+
+    var to_add: std.ArrayList(*http.Connection) = .empty;
```
karlseguin: `to_remove` and `to_add` exist to reduce the `submission_mutex` lock?  2 possible optimizations:

1 - Store them as fields on Network so that, instead of `deinit`, we can clearRetainingCapacity
2 - Use 2 lists `pending_add_1` and `pending_add_2` and have `pending_add = &pending_add_1`. Then you need a very short lock to swap them.

karlseguin:
1 - If I remove all but the first test from websocket.html, the test hangs. This isn't the case in `main`
2 - In `main`, we silence error.GotNothing for websockets: https://github.com/lightpanda-io/browser/blob/d360fcc07b13667cc49d013806601be7b3641239/src/browser/HttpClient.zig#L788

src/browser/webapi/net/WebSocket.zig
```diff
@@ -612,46 +771,54 @@ fn _receivedDataCallback(conn: *http.Connection, data: []const u8) !void {
...
 
             if (self._ready_state == .closing) {
-                // Client-initiated close: this is the server's response.
-                // Close handshake complete - disconnect.
-                self.disconnected(null);
+                // Client-initiated close — server's response. Don't
+                // disconnect inline: we're inside a libcurl callback
+                // and tearing the conn down here would UAF the easy
+                // handle. Curl will deliver normal completion when the
+                // server closes the socket per RFC 6455 §5.5.1.
```
karlseguin: I don't think we should rely on servers following specs.

src/browser/HttpClient.zig
```diff
@@ -216,8 +200,30 @@ pub fn init(self: *Client, allocator: Allocator, network: *Network, cdp_client:
...
+        if (self.network.shutdown.load(.acquire)) {
+            self.network.drainPendingForShutdown();
+        }
+        _ = self.processMessages() catch {};
+        self.drainReadyWs();
+        if (self.handle.in_use.first == null) break;
+        std.Thread.sleep(std.time.ns_per_ms);
+    }
+    if (self.handle.in_use.first != null) {
+        log.warn(.http, "deinit with active conns", .{
```
karlseguin: This seems like a more serious issue. At this point, we'll crash N seconds later in a seemingly not-related part. I'd rather just crash here.

Also, can the loop not replace the `spins` and `sleep` with a call to `handle.poll` with a timeout?

Finally,  is there any danger in calling `processMessage` here? Can we have a tailor-made cleanup?

karlseguin:
Still plan on doing this, but both the Zig 0.16 and HttpClient have caused too much divergence. think the new transfer event queue, which is also used by websockets (one of the difficult things this PR had to deal with) potentially changes (and hopefully simplifies) how we should approach this.

PR #2305, opened 2026-04-28
xpath: implement XPath 1.0 (Document.evaluate, XPathResult, DOM.performSearch)

navidemad:
## Summary

Ports the capybara-lightpanda XPath 1.0 polyfill into Lightpanda's Zig codebase. Exposes the WHATWG `Document.evaluate` / `XPathResult` / `XPathEvaluator` / `XPathExpression` surface and routes CDP `DOM.performSearch` XPath queries through the new evaluator. Motivation: the capybara-lightpanda gem currently injects a ~700-line JS polyfill on every page load to make `document.evaluate` work; this PR moves that into the engine so the gem can drop the polyfill in its next release.

Full behavior spec + the 91-case acceptance battery this port implements: [capybara-lightpanda/XPATH_COMPLIANCE.md](https://github.com/navidemad/capybara-lightpanda/blob/main/XPATH_COMPLIANCE.md).

> [!NOTE]
> Downstream coordination: capybara-lightpanda will drop its ~700-line JS polyfill in the same release that bumps `MINIMUM_NIGHTLY_BUILD` past the merge build of this PR. No action needed from this repo's reviewers — flagged for visibility.

## Scope

- New `src/browser/xpath/` module: Tokenizer, Parser, Ast, Evaluator, Functions, Result (~3,000 LOC).
- New `src/browser/webapi/{XPathResult,XPathExpression,XPathEvaluator}.zig` WHATWG types (~470 LOC).
- `Document.evaluate`, `Document.createExpression`, `Document.createNSResolver` wired via the JS bridge.
- `DOM.performSearch` heuristic-based XPath/CSS branching — gives Playwright/Puppeteer/Capybara XPath-via-CDP for free.
- Behavior fixtures: 91-case conformance battery, result-type surface, evaluator-API surface, CDP perform-search.

## `Document.evaluate` flow

```mermaid
sequenceDiagram
    participant JS
    participant Document
    participant XPathResult
    participant Parser
    participant Evaluator

    JS->>Document: evaluate(expr, ctx, resolver, type, result)
    Document->>XPathResult: fromExpression(expr, ctx, type, frame)
    XPathResult->>Parser: parse(arena, expr)
    Parser-->>XPathResult: *Ast.Expr
    XPathResult->>Evaluator: evaluate(arena, frame, ast, ctx)
    Evaluator-->>XPathResult: Result.Result
    XPathResult-->>JS: XPathResult { resultType, ... }
```

## `DOM.performSearch` query-type detection

```mermaid
flowchart LR
    A[CDP query] --> B{isXPathQuery?}
    B -->|path operator or :: axis| C[xpath.searchAll]
    B -->|otherwise| D[Selector.querySelectorAll]
    C --> E[finishSearch]
    D --> E
    E --> F[CDP result: searchId, resultCount]
```

The full heuristic: query starts with `/`, `.//`, `(/`, `(./`, or contains `::` substring (axis specifier) → XPath; otherwise CSS.

## Coverage

Full table in [XPATH_COMPLIANCE.md](https://github.com/navidemad/capybara-lightpanda/blob/main/XPATH_COMPLIANCE.md).

| Area | Implemented | Stubs |
|------|-------------|-------|
| Path expressions | 11/11 | — |
| Axes | 11/12 | `namespace::` (returns `[]`) |
| Node tests | 4 type tests + name + wildcard | `processing-instruction('name')` consumes target literal but matches any PI |
| Operators | union, `or`/`and`, `=`/`!=`, `<`/`<=`/`>`/`>=`, `+`/`-`, `*`/`div`/`mod`, unary `-` | — |
| Node-set fns | `last`, `position`, `count`, `id`, `local-name`, `name` | `namespace-uri()` always `""` |
| String fns | `string`, `concat`, `starts-with`, `contains`, `substring-before`, `substring-after`, `substring`, `string-length`, `normalize-space`, `translate` | — |
| Boolean fns | `boolean`, `not`, `true`, `false` | `lang()` always `false` |
| Number fns | `number`, `sum`, `floor`, `ceiling`, `round` | — |
| Variable refs | parsed | `$name` always `""` |

<details>
<summary>Intentional stubs (HTML pragmatism — polyfill parity)</summary>

These match the prior capybara-lightpanda polyfill, which is the original motivation for this PR (HTML pragmatism over strict XPath 1.0):

- `lang(string)` → always `false`
- `namespace::` axis → always `[]`
- `namespace-uri()` → always `""`
- `processing-instruction('target')` → target literal consumed but matches any PI
- `name()` / `local-name()` → lowercased (HTML-friendly, not source-case as the spec demands)
- Variable references `$name` → always `""`
- Tokenizer skips unknown chars silently
- `DOM.performSearch` top-level scalar XPath → empty (matches the polyfill's `xpathFind` semantics — the API is for finding nodes)

</details>

## Test plan

- [x] `make test` — 601/601 pass
- [x] 30 Tokenizer + 40 Parser (parametric over the 91-case battery) + 11 Evaluator + 15 Functions + 9 Result Zig unit tests
- [x] 4 HTML behavior fixtures: `document_evaluate.html`, `xpath_conformance.html` (91 cases), `xpath_result.html`, `xpath_evaluator.html`
- [x] CDP fixture `perform_search_xpath.html` + Zig test driving 5 query shapes
- [x] Custom test runner detects leaks under debug build; clean run confirms no leaks


src/browser/webapi/Document.zig
```diff
@@ -412,6 +414,40 @@ pub fn createNodeIterator(_: *const Document, root: *Node, what_to_show: ?js.Val
     return DOMNodeIterator.init(root, try whatToShow(what_to_show), filter, frame);
 }
 
+pub fn evaluate(
```
karlseguin, line 417: Is this signature correct? From playing with Firefox, it seems like `context_node` is not optional, but everything else is. (so `context_node: *Node` and  `result_type: ?u16`).

src/browser/webapi/XPathResult.zig
```diff
@@ -0,0 +1,282 @@
...
+    return false;
+}
+
+// ----- methods -----
+
+pub fn iterateNext(self: *XPathResult) !?*Node {
+    if (self._type != UNORDERED_NODE_ITERATOR_TYPE and self._type != ORDERED_NODE_ITERATOR_TYPE) {
+        return error.InvalidStateError;
+    }
+    if (self._iter_pos >= self._value.nodes.len) return null;
```
karlseguin:
```zig
const pos = self._iter_pos;
if (pos >= self._value.nodes.len) return null;
const node = self._value.nodes[pos];
self._iter_pos = pos + 1;
```

My take is that this is (a) easier to read and (b) possibly faster. So a win-win, but minor.

src/browser/webapi/XPathExpression.zig
```diff
@@ -0,0 +1,105 @@
...
+    const expr = try xpath.Parser.parse(arena, owned);
+    const xe = try arena.create(XPathExpression);
+    xe.* = .{ ._arena = arena, ._expr = expr };
+    return xe;
+}
+
+pub fn evaluate(
+    self: *XPathExpression,
+    context_node: *Node,
+    requested_type: u16,
```
karlseguin: looks like this is optional an default to 0 /  `ANY_TYPE`

src/browser/xpath/Evaluator.zig
```diff
@@ -0,0 +1,732 @@
...
+//! slices, formatted numbers, materialized attribute nodes) into the
+//! caller's arena. The context `Frame` is needed for `getElementById`
+//! and to materialize attributes (the attribute axis returns full
+//! `Attribute` nodes so the result is `*Node`-uniform).
+//!
+//! Document-order sort happens once at the public boundary
+//! (`evaluate()`); intermediate step results stay in axis order so
+//! reverse-axis positional predicates evaluate against proximity.
+
+const std = @import("std");
```
karlseguin, line 30: Sorry to be _that guy_, but we generally have 3 top-level groups

1. importing libraries (std, lightpanda, buitin are the most common)
2. importing relative files (I personally like to go from "further" to "nearest", but that's just me)
3. constants

So this should look something like:


```zig
const std = @import("std");

const Node = @import("../webapi/Node.zig");

const Ast = @import("Ast.zig");
const Parser = @import("Parser.zig");
const Result = @import("Result.zig");
const Functions = @import("Functions.zig");

const Frame = lp.Frame;
const Element = Node.Element;
const Document = Node.Document;
const Allocator = std.mem.Allocator;
```

The webapi stuff looks ok, but there are a few places in this xpath code that could be tidied up.

src/browser/xpath/Evaluator.zig
```diff
@@ -0,0 +1,732 @@
...
+    UnknownFunction,
+    UnionRequiresNodeSets,
+};
+
+arena: Allocator,
+frame: *Frame,
+
+/// Public entry. Returns the AST's value; node-sets are sorted into
+/// document order before return per XPath spec §3.3.
+pub fn evaluate(arena: Allocator, frame: *Frame, expr: *const Ast.Expr, context_node: *Node) Error!Result.Result {
```
karlseguin: `frame` is generally the last parameter. There's no strict reason for this, but since the js bridge enforces this for webapis, it's something we tend to follow throughout.

src/browser/xpath/result.zig
```diff
@@ -0,0 +1,200 @@
...
+//! `toNumber`, `toBoolean`) follows XPath 1.0 spec §3, with HTML-pragmatic
+//! shortcuts inherited from the polyfill (decision #2).
+
+const std = @import("std");
+const Allocator = std.mem.Allocator;
+
+const Node = @import("../webapi/Node.zig");
+const CData = Node.CData;
+
+pub const Result = union(enum) {
```
karlseguin, line 33: The filename should be lowercase:

```zig
File names fall into two categories: types and namespaces. If the file (implicitly a struct) has top level fields, it should be named like any other struct with fields using TitleCase. Otherwise, it should use snake_case. Directory names should be snake_case
```

I wish this was handled better in Zig, e.g. I wish you could make a file be an enum or union somehow.

https://ziglang.org/documentation/master/#toc-Names

src/browser/xpath/Ast.zig
```diff
@@ -0,0 +1,134 @@
...
+    /// this variant; the evaluator returns an empty node-set.
+    unknown,
+};
+
+pub const NodeTest = union(enum) {
+    /// Element / attribute name. Special values:
+    ///   - "*"            → wildcard
+    ///   - "prefix:*"     → namespace wildcard
+    ///   - "prefix:local" → namespace-prefixed name
+    /// The evaluator splits these.
```
karlseguin: Does it? I feel like `matchNameTest` doesn't. We should remove the comment. and maybe put a TODO somewhere.

src/browser/xpath/Parser.zig
```diff
@@ -0,0 +1,923 @@
...
+pub const Error = error{
+    OutOfMemory,
+    UnexpectedToken,
+    ExpectedNodeTest,
+    ExpectedPrimaryExpr,
+};
+
+arena: Allocator,
+tokens: []const Token,
+pos: usize = 0,
```
karlseguin, line 51: I think we should have a depth counter, and limit it to something reasonable 32? 64?

karlseguin (commented):
My biggest concern is about the performance of common queries, especially with respect to memory usage. I'll be honest, I only have a high level understanding of xpath and of your implementation. In our CSS selector, we apply some optimizations, including mostly doing right-to-left processing, to minimize the in-memory size. Claude informs me that probably doesn't work so well with xpath since they simply allow for a much broader set of instructions.

I also don't want to make the PR or the code too complicated. So tell me what you think of these possible optimization. I don't think any of these are mutually exclusive, but possibly some of them don't make sense or are impractical.

1 - Optimizing for common shapes that have easy optimizations. e.g. //foo, //foo[@id='x'] and //*[@id='x']

2 - right-to-left processing where where predicates can be evaluated independently per node, without knowing their position, e.g. //foo[bar]

3 - Introducing a re-usable arena so that scratch allocations made within each step, or ideally in the per-ctx inner loopof step, doesn't go into the main arena. The arena can be reset between each iteration. 

4 - Using an iterator for the axis (rather than collecting them in memory).
    - Allows for further optimizations (early exist)
    - last() adds complexity to this, but it can be detected at parse time, and that could trigger materialization while still exposing an a consistent iterator API to everything.

5 - Whether it's an iterator or we continue to collect the full axis, there are common cases where the dedupe list is already in document order. This can be detected (not by iterating the list, but based on what the xpath itself) and, when true, dedupe.keys() does not need to be sorted

src/browser/webapi/XPathResult.zig
navidemad: Applied — uses the local `pos` form.

src/browser/xpath/Evaluator.zig
navidemad: Done — restructured imports across `Evaluator.zig`, `Parser.zig`, `functions.zig`, and `result.zig` to follow lib (std/lp) → relative (far → near) → aliases. The webapi `XPath*.zig` files were already in that shape, as you noted.

src/browser/xpath/Evaluator.zig
navidemad: Done — moved `frame` to last on `Evaluator.evaluate`, `Evaluator.searchAll`, `Functions.call`, and `Functions.idFn`. Call sites updated in `webapi/XPath{Result,Expression}.zig` and `cdp/domains/dom.zig`.

src/browser/xpath/result.zig
navidemad: Done — renamed `Result.zig` / `Ast.zig` / `Functions.zig` to snake_case. Kept `Tokenizer.zig` / `Parser.zig` / `Evaluator.zig` as TitleCase since those have top-level fields.

src/browser/webapi/Document.zig
navidemad: Half-applied. `result_type: ?u16` defaulting to `ANY_TYPE` — done across `Document.evaluate`, `XPathEvaluator.evaluate`, and `XPathExpression.evaluate`.

Kept `context_node: ?*Node` with the document fallback, though. Reasons:
- The fixture `document_evaluate.html` (`default_context` block) explicitly tests that `document.evaluate("//p", null, …)` and `document.evaluate("//p", undefined, …)` both resolve to the document — that mirrors the polyfill (decision n2, exact parity).
- Firefox throws `TypeError` on a *missing* arg, but the JS bridge can't distinguish "missing" from "explicit `null`" without changes upstream — both surface as `null` to Zig.
- MDN says the spec's `Node? contextNode` is nullable; the throw is only for missing.

So polyfill parity wins the ambiguous case here. Happy to flip to `*Node` if you'd rather match Firefox's stricter behavior — would mean dropping the `default_context` fixture block.

src/browser/webapi/XPathExpression.zig
navidemad: Done — `requested_type` is now `?u16` defaulting to `ANY_TYPE` here too (and on `XPathEvaluator.evaluate`).

src/browser/xpath/Ast.zig
navidemad: Good catch — fixed. The doc comment now says namespaced forms (`prefix:*`, `prefix:local`) are stored verbatim and fall through to a literal `mem.eql` against the node name, consistent with the `namespace::` axis stub (decision n3). TODO marker added for if/when the polyfill drops the namespace stub.

src/browser/xpath/Parser.zig
navidemad: Done — added a depth counter capped at 64 with a new `error.MaxDepthExceeded`. Increment/decrement happens in `parseExpr` (covers parens, predicates, function args) and `parseUnaryExpr` (covers chained `-`). Two regression tests added: deep `((((…1…))))` and deep `-----1` both reject past the limit.

navidemad:
@karlseguin Thanks — spent some time on measurement and Point 1 (the cheapest, highest-value lever from your list). Recap below, then numbers.

## Benchmark fixture

Added `src/browser/tests/xpath/xpath_perf.html`: deterministic 500-node DOM (mix of `div`/`span`/`p` × `alpha`/`beta`/`gamma`, decorrelated periods so `//div[@class='alpha']` isn't a degenerate restatement of `//div`), 50 iterations per query after a 3-iter warmup, asserts result counts so a regression in correctness can't hide behind a timing line.

Run: `TEST_VERBOSE=false TEST_FILTER="XPath perf" zig build test`.

## Point 1 — id-lookup fast path

`tryIdLookupFastPath` at the top of `Evaluator.evalPath`. Recognizes:

```
//tag[@id='x']    →  ds::node() / child::tag[@id='literal']
.//tag[@id='x']   →  self::node() / ds::node() / child::tag[@id='literal']
```

…and serves them via `frame.getElementByIdFromNode` (case-insensitive tag check, containment check for relative paths, accepts the literal on either side of `=`). Falls through to the general path on any deviation: extra step, extra predicate, non-eq, non-literal RHS, or unresolvable search root.

Inherits the same compromise `webapi/selector/List.zig:optimizeSelector` already ships for `querySelector(All)` — the id-map only stores the first element per ID in document order, so a page with duplicate IDs returns one match where a strict tree walk would find all. Capybara/Selenium hot paths assume unique IDs and the same compromise has been deployed in CSS for years.

## Results — debug build, 500-node tree, 50 iters/query

| label | baseline µs | after µs | speedup |
|---|---:|---:|---:|
| `//*[@id='target']` | 3231 | 22.6 | **143×** |
| `//span[@id='target']` (hit) | 2931 | 22.4 | **131×** |
| `//div[@id='target']` (miss, tag mismatch) | 2089 | 21.7 | **96×** |
| `//div` | 2964 | 1133 | flat* |
| `//span` | 2659 | 3354 | flat* |
| `//*` | 8986 | 6595 | flat* |
| `//*[@class='alpha']` | 7210 | 5751 | flat* |
| `//div[@class='alpha']` | 3128 | 3272 | flat |
| `(//div)[1]` | 2631 | 3518 | flat* |
| `(//div)[last()]` | 2331 | 2561 | flat |
| `//div[contains(@class,'alpha')]` | 4005 | 4390 | flat |
| `//div[starts-with(@id,'n')]` | 4637 | 4724 | flat |
| `count(//div)` | 2880 | 1962 | flat* |

\* ±30% across runs (debug build, 50 iters). Fast path falls through immediately for these shapes — any movement is system noise. The 22µs floor on ID lookups is parser + bridge round-trip; at that point we're measuring V8 ↔ Zig overhead, not XPath work.

All 79 XPath tests still pass, including cases the fast path correctly *rejects* at the predicate-shape check (e.g. `//*[@id='heading' and @class='primary']`).

## Next

Pushed Point 1 to this PR in `8733e33b`. My read on the rest:

- **Point 4** (iterator axes with parse-time `needs_size` detection) — folds in Point 2's logic, biggest payoff for `(//x)[1]` / `count(//x)` / `//foo[bar]`. Worth it if the noise-level numbers above bother you in real Capybara runs. The biggest remaining cost is `//*[@class='x']` (~7ms) and `//*` (~9ms), both of which touch the full tree.
- **Point 3** (per-step scratch arena) — defer until profilers point at peak alloc.
- **Point 5** (skip dedupe sort when already in doc order) — skip. Detection is fragile (single ctx + all forward axes + no overlapping subtrees) and the savings are bounded by sort cost, which is sub-ms in practice.

navidemad:
Update — pushed `f61be05f` extending the Point 1 fast path to non-positional descendant queries.

`tryFusedDescendantFastPath` accepts the same AST shape (`//<test>` / `.//<test>` lowering) but allows any boolean/node-set predicate that doesn't depend on outer position. Walks the search root's descendants once in document order, applies the node test + predicates inline. No per-step axis materialization, no dedup hash map (single-context forward walk preserves doc order).

Safety gate rejects predicates whose top-level expression is numeric (number literal, neg, arithmetic binop, numeric-returning fn-call) and any predicate containing `position()` or `last()` anywhere. Conservative — a nested sub-path's local positional predicate (`[bar[1]]`) is rejected even though `[1]` is scoped to bar's axis. Easy to relax later if it shows up in real usage.

`tryIdLookupFastPath` refactored to share `matchDescendantPathShape` with the new fast path; behavior unchanged.

### Numbers (debug build, 500-node tree, 50 iters/query)

| label | baseline | after Point 1 | after fused | speedup |
|---|---:|---:|---:|---:|
| `//*[@id='target']` | 3231 | 22.6 | 22.6 | **143×** |
| `//span[@id='target']` (hit) | 2931 | 22.4 | 22.8 | **129×** |
| `//div[@id='target']` (tag miss) | 2089 | 21.7 | 21.7 | **96×** |
| `//div` | 2964 | 1133 | 386 | **8×** |
| `//span` | 2659 | 3354 | 384 | **7×** |
| `//*` | 8986 | 6595 | 1018 | **9×** |
| `//*[@class='alpha']` | 7210 | 5751 | 1498 | **5×** |
| `//div[@class='alpha']` | 3128 | 3272 | 625 | **5×** |
| `(//div)[1]` | 2631 | 3518 | 87 | **30×** |
| `(//div)[last()]` | 2331 | 2561 | 94 | **25×** |
| `//div[contains(@class,'alpha')]` | 4005 | 4390 | 704 | **6×** |
| `//div[starts-with(@id,'n')]` | 4637 | 4724 | 794 | **6×** |
| `count(//div)` | 2880 | 1962 | 90 | **32×** |

`(//div)[1]`, `(//div)[last()]`, and `count(//div)` pick up the speedup transitively — `//div` is now a fused-walk shape, so the inner expression is fast even though the outer filter/count is unchanged.

All 79 XPath tests still pass. Conformance cases that correctly *fall through* to the general path because their predicates are positional or reference `position()`/`last()`: `//li[1]`, `//li[last()]`, `//li[position() > 2]`, `//li[position() mod 2 = 1]`, `(//section)[2]`.

navidemad:
Two commits in, the fast paths land 5–143× wins across the benchmark — see the previous comment for the full table. The fused-walk extension absorbed most of what I'd originally pegged for Point 4, since `(//x)[1]` / `count(//x)` benefit transitively from the inner path being fast. Points 2, 4, and 5 from your list are effectively done or skipped.

One remaining lever I'd consider: **Point 3** (per-predicate-iteration scratch arena). Worth more than I implied earlier — the XPath result arena (`XPathResult._arena`) lives until JS releases the result via `XPathResult.deinit`, not just for the duration of the call, so peak per-call allocation persists across calls if a script holds results. Point 3 would target that. Structurally more invasive than the surgical shape-detection though: need to thread a separate scratch arena through `evalStep` / predicate evaluation.

@karlseguin How do you want to proceed?

- **Land as-is**, file Point 3 as a separate follow-up.
- **Push deeper on this branch** — Point 3 before merge, bigger diff but more complete perf coverage.
- Re-run the benchmark in release mode first to see if the gains are still material outside debug builds.

Lightly leaning toward the first option — your original framing was "common shapes that have easy optimizations" and we're past that. But happy to keep going if you'd rather not bifurcate the work.

karlseguin:
We can merge this as-is. I do think a follow up to leverage a scratch arena is worth it, those temp allocations can be quite large (entire DOM tree) and they're going to live until v8 frees them - which might be only on page end. While the new fast paths should help, I imagine some scripts do a lot of these xpath queries to fetch various fields and it could add up.

@krichprollsch still want to review it more?

src/cdp/domains/dom.zig
```diff
@@ -616,6 +649,70 @@ test "cdp.dom: search flow" {
...
+test "cdp.dom: isXPathQuery heuristic" {
+    // XPath-shaped queries — each line covers a distinct heuristic branch.
+    try std.testing.expect(isXPathQuery("/html"));
+    try std.testing.expect(isXPathQuery("//p"));
+    try std.testing.expect(isXPathQuery(".//foo"));
+    try std.testing.expect(isXPathQuery("(//foo)[1]"));
+    try std.testing.expect(isXPathQuery("(./bar)[2]"));
+    try std.testing.expect(isXPathQuery("descendant::p"));
+    try std.testing.expect(isXPathQuery("ancestor-or-self::*"));
+    try std.testing.expect(isXPathQuery("//*[@id='x']"));
```
karlseguin, line 729: I think if you add cases like `a::before` it'll fail. You should use a `StaticStringMap` and match against the specific axis names.

navidemad: Good catch — `a::before` did indeed pass (`a` matches `[a-zA-Z-]`). Fixed in d4de5e6: walks back the identifier run before `::` and looks it up in a `StaticStringMap` of the 13 XPath 1.0 axis names. Test now also covers `a::before`, `div::after`, `p::first-line`, `input::placeholder`, and `[data-x="x::y"]`.

krichprollsch (approved):

krichprollsch:
Thanks @navidemad for this hard work, it's a nice and useful addition to the browser :pray: 
Thanks @karlseguin for the review :pray:

PR #2310, opened 2026-04-29
dom: implement HTMLElement.isContentEditable IDL attribute

### LINKED ISSUE #2309: dom: HTMLElement.isContentEditable IDL attribute not implemented
## Summary

`HTMLElement.isContentEditable` is not exposed as an IDL attribute. Reading `el.isContentEditable` from JavaScript returns `undefined` for every element, regardless of whether the element (or any ancestor) carries a `contenteditable` attribute.

CDP clients that probe the editability of focused/clicked elements before issuing `Input.dispatchKeyEvent` can't tell whether a node is part of an editable region, and JS-driven helpers that branch on `isContentEditable` to choose between value-setter and `innerHTML` paths fall through their first guard.

## Today's behavior

Reading the IDL attribute returns `undefined` (`typeof === "undefined"`) for every element. The DOM source has no `isContentEditable` accessor on `HTMLElement` — `src/browser/webapi/element/Html.zig` declares getters/setters for `dir`, `hidden`, `lang`, `tabIndex`, `title`, but no `isContentEditable`. The only `contenteditable` reference in the engine sits at `src/browser/interactive.zig:258`, and it is consumed by the LP semantic-tree categorization, not exposed to V8.

```mermaid
sequenceDiagram
    participant LP as Lightpanda
    participant V8 as V8 / DOM
    Client->>LP: Runtime.evaluate "el.isContentEditable"
    LP->>V8: read isContentEditable on HTMLElement
    V8-->>LP: property missing on prototype
    LP-->>Client: result.value = undefined
    Note over LP,V8: src/browser/webapi/element/Html.zig: no IDL accessor for isContentEditable
```

## Expected behavior

Per [HTML §7.7.5.2 — The isContentEditable IDL attribute](https://html.spec.whatwg.org/multipage/interaction.html#dom-iscontenteditable):

> The `isContentEditable` IDL attribute, on getting, must return true if the element is either an editing host or editable, and false otherwise.

Per [HTML §7.7.5 — Making document regions editable](https://html.spec.whatwg.org/multipage/interaction.html#making-document-regions-editable-the-contenteditable-content-attribute), `contenteditable` is an enumerated attribute with keywords `""` (empty), `"true"`, `"false"`, and `"plaintext-only"`. The `false` keyword maps to the false state; everything else maps to an editable state. The effective state is determined by walking up to the nearest ancestor with the attribute set.

```mermaid
sequenceDiagram
    participant LP as Lightpanda
    participant V8 as V8 / DOM
    Client->>LP: Runtime.evaluate "el.isContentEditable"
    LP->>V8: getIsContentEditable(el)
    V8->>V8: walk ancestors for contenteditable
    V8-->>LP: true / false based on nearest non-"false" ancestor
    LP-->>Client: result.value = boolean
```

## Reproducer

Self-contained: one HTML fixture + one shell script + one Node CDP probe (raw `ws`). Asserts the spec-defined boolean for the 9 fixture cases; exits 1 if any return value diverges, exits 0 if all match.

`repro.html`:
```html
<!doctype html>
<html><body>
<div id="own" contenteditable="true">own</div>
<div id="inherit-host" contenteditable="true"><span id="inherit-child">child</span></div>
<div id="explicit-false" contenteditable="false"><span id="explicit-false-child">x</span></div>
<div id="empty-string" contenteditable=""><span id="empty-string-child">x</span></div>
<div id="plain"><span id="plain-child">x</span></div>
</body></html>
```

`probe.js` (uses raw `ws`; full helper at the gist below):
```js

(async () => {
  const c = await connect();
  await c.navigate('http://127.0.0.1:9580/repro.html');

  const probe = await c.eval(`
    (() => {
      const ids = ['own','inherit-host','inherit-child','explicit-false','explicit-false-child',
                   'empty-string','empty-string-child','plain','plain-child'];
      const out = {};
      for (const id of ids) {
        const el = document.getElementById(id);
        out[id] = { isContentEditable: el ? el.isContentEditable : 'no-element',
                    typeofVal:        el ? typeof el.isContentEditable : 'no-element' };
      }
      return out;
    })()
  `);

  const expected = {
    'own': true, 'inherit-host': true, 'inherit-child': true,
    'explicit-false': false, 'explicit-false-child': false,
    'empty-string': true, 'empty-string-child': true,
    'plain': false, 'plain-child': false,
  };

  let pass = true;
  for (const [id, want] of Object.entries(expected)) {
    if (probe[id].isContentEditable !== want) {
      console.log(`MISMATCH ${id}: got=${JSON.stringify(probe[id].isContentEditable)} (typeof=${probe[id].typeofVal}) want=${want}`);
      pass = false;
    }
  }

  c.close();
  process.exit(pass ? 0 : 1);
```

`repro.sh`:
```bash
#!/usr/bin/env bash
set -euo pipefail
HTTP_PORT=9580; CDP_PORT=9222

lsof -ti:"$CDP_PORT" 2>/dev/null  | xargs -r kill 2>/dev/null || true

PIDS=()
trap 'for p in "${PIDS[@]:-}"; do kill "$p" 2>/dev/null || true; done' EXIT

cd "$REPRO_DIR"
python3 -m http.server "$HTTP_PORT" >http.log 2>&1 & PIDS+=("$!")
for _ in $(seq 1 30); do curl -fsS "http://127.0.0.1:${HTTP_PORT}/repro.html" >/dev/null 2>&1 && break; sleep 0.1; done

  "$LIGHTPANDA_BIN" serve --host 127.0.0.1 --port "$CDP_PORT" >lightpanda.log 2>&1 & PIDS+=("$!")
for _ in $(seq 1 50); do
  [[ "$body" == *'"Browser": "Lightpanda'* ]] && break
  sleep 0.1
done

[ -d node_modules/ws ] || npm install --no-save --silent ws
exec node probe.js
```

`cdp.js` is a small raw-`ws` helper that wraps `Target.createTarget` + `Target.attachToTarget` + sessionId-tracked `Runtime.evaluate`. Drop in [this gist](https://gist.github.com/) (or use any equivalent CDP client capable of driving Lightpanda's `flatten: true` session flow).

**Run**: `bash repro.sh`

- **Today** on `1.0.0-dev.5891+e133b26e`: every case prints `typeofVal: "undefined"` and the script exits 1.

  ```
  MISMATCH own: got=undefined (typeof=undefined) want=true
  MISMATCH inherit-host: got=undefined (typeof=undefined) want=true
  MISMATCH inherit-child: got=undefined (typeof=undefined) want=true
  MISMATCH explicit-false: got=undefined (typeof=undefined) want=false
  MISMATCH explicit-false-child: got=undefined (typeof=undefined) want=false
  MISMATCH empty-string: got=undefined (typeof=undefined) want=true
  MISMATCH empty-string-child: got=undefined (typeof=undefined) want=true
  MISMATCH plain: got=undefined (typeof=undefined) want=false
  MISMATCH plain-child: got=undefined (typeof=undefined) want=false
  FAIL: isContentEditable returns wrong values (see mismatches above)
  ```

- **Expected after fix**: every case returns a boolean and the script exits 0.

  ```
  OK: isContentEditable matches spec for all 9 cases
  ```

## Likely fix location

`src/browser/webapi/element/Html.zig` — add a `getIsContentEditable(self: *HtmlElement) bool` getter that walks up via `Element.parentElement()`, finds the nearest ancestor (or self) with a `contenteditable` attribute, and returns `true` unless the value (case-insensitive) is `"false"`. Bind it in `JsApi` next to the existing `hidden` / `tabIndex` accessors. The `contenteditable` attribute name is 15 bytes — past the comptime SSO limit on `String.wrap` — so the wrap has to run at runtime, mirroring the pattern at `src/browser/interactive.zig:258`.

`SVGElement` (`src/browser/webapi/element/Svg.zig`) carries the same IDL attribute via the `ElementContentEditable` mixin in the spec; out of scope here because the SVG `JsApi` currently has no IDL accessors at all and would need its own design pass.

## Environment

- Lightpanda: `1.0.0-dev.5891+e133b26e` (current `main` HEAD)
- OS: macOS 14 (darwin aarch64)
- CDP client: raw WebSocket via Node's `ws` package (no `chrome-remote-interface` — the default `/json/list` flow doesn't work against Lightpanda)

navidemad:
## What

`HTMLElement.isContentEditable` was missing as an IDL attribute — reading `el.isContentEditable` from JS returned `undefined` for every element. This wires up the spec-defined boolean per [HTML §7.7.5.2](https://html.spec.whatwg.org/multipage/interaction.html#dom-iscontenteditable).

Closes #2309.

## Root cause

`src/browser/webapi/element/Html.zig` declared IDL accessors for `dir`, `hidden`, `lang`, `tabIndex`, and `title` on `HTMLElement`, but had no entry for `isContentEditable`. With the property absent from the prototype, `Runtime.evaluate "el.isContentEditable"` resolved to `undefined` regardless of whether the element (or any ancestor) carried a `contenteditable` attribute. The only `contenteditable` reference in the engine was at `src/browser/interactive.zig:258`, consumed by the LP semantic-tree categorization rather than exposed to V8.

```mermaid
flowchart LR
    A[el.isContentEditable] --> B[no IDL accessor on HTMLElement]
    B --> C[returns undefined]
    style B fill:#fdd
    style C fill:#fdd
```

## Fix

- `src/browser/webapi/element/Html.zig` — add `getIsContentEditable(self: *HtmlElement) bool`. Walks up via `Element.parentElement()` from self to root; returns `true` at the first ancestor whose `contenteditable` attribute is non-`false` (case-insensitive), `false` if it is exactly `false`, and `false` if no ancestor carries the attribute. Bound as a read-only IDL accessor in `JsApi` next to `hidden` / `tabIndex`.
- `src/browser/tests/element/html/contenteditable.html` — new fixture covering own / inherited / `false` / empty-string / `plaintext-only` / case-insensitive / dynamic-attribute cases.

```mermaid
flowchart LR
    A[el.isContentEditable] --> B[walk ancestors for contenteditable]
    B --> C[true unless nearest is false]
    style B fill:#dfd
    style C fill:#dfd
```

The `contenteditable` attribute name is 15 bytes — past the comptime SSO limit on `String.wrap` — so the wrap runs at runtime, mirroring the pattern already used at `src/browser/interactive.zig:258`.

## Test

- **Unit test**: `WebApi: HTMLElement.contenteditable` (5 sub-tests in `src/browser/tests/element/html/contenteditable.html`) — fails on `main` because every `el.isContentEditable` read is `undefined`; passes with this commit.
- **Reproducer**: the `repro.sh` from #2309 exits 1 on `main` (every fixture case prints `typeof === "undefined"`) and exits 0 with this commit (every case returns the correct boolean). Asserts the spec-defined value across own / inherited / `false` / empty-string / plain cases.

## Notes

`SVGElement` (`src/browser/webapi/element/Svg.zig`) carries the same IDL attribute via the `ElementContentEditable` mixin in the spec. Out of scope here because the SVG `JsApi` currently exposes no IDL accessors at all and would need a broader design pass; happy to follow up in a separate PR if useful.

### DIFF
--- a/src/browser/webapi/element/Html.zig
+++ b/src/browser/webapi/element/Html.zig
@@ -382,6 +382,24 @@ pub fn setTitle(self: *HtmlElement, value: []const u8, frame: *Frame) !void {
     try self.asElement().setAttributeSafe(comptime .wrap("title"), .wrap(value), frame);
 }
 
+// HTML §7.7.5.2: returns true iff the element's effective content editable
+// state is "true" or "plaintext-only". Walk up to the nearest ancestor (or
+// self) with a `contenteditable` attribute; the keyword "false" maps to the
+// false state, every other value (including the empty string, "true", and
+// "plaintext-only") maps to an editable state. With no ancestor carrying the
+// attribute the document's designMode default ("off") yields false.
+//
+// "contenteditable" is 15 bytes — past the comptime SSO limit — so the
+// String wrap runs at runtime, mirroring the pattern in interactive.zig.
+pub fn getIsContentEditable(self: *HtmlElement) bool {
+    var current: ?*Element = self.asElement();
+    while (current) |el| : (current = el.parentElement()) {
+        const raw = el.getAttributeSafe(.wrap("contenteditable")) orelse continue;
+        return !std.ascii.eqlIgnoreCase(raw, "false");
+    }
+    return false;
+}
+
 pub fn getAttributeFunction(
     self: *HtmlElement,
     listener_type: GlobalEventHandler,
@@ -1220,6 +1238,7 @@ pub const JsApi = struct {
 
     pub const dir = bridge.accessor(HtmlElement.getDir, HtmlElement.setDir, .{});
     pub const hidden = bridge.accessor(HtmlElement.getHidden, HtmlElement.setHidden, .{});
+    pub const isContentEditable = bridge.accessor(HtmlElement.getIsContentEditable, null, .{});
     pub const lang = bridge.accessor(HtmlElement.getLang, HtmlElement.setLang, .{});
     pub const tabIndex = bridge.accessor(HtmlElement.getTabIndex, HtmlElement.setTabIndex, .{});
     pub const title = bridge.accessor(HtmlElement.getTitle, HtmlElement.setTitle, .{});
@@ -1356,4 +1375,7 @@ test "WebApi: HTML.event_listeners" {
 }
 test "WebApi: HTMLElement.props" {
     try testing.htmlRunner("element/html/htmlelement-props.html", .{});
+}
+test "WebApi: HTMLElement.contenteditable" {
+    try testing.htmlRunner("element/html/contenteditable.html", .{});
 }
--- a/src/browser/webapi/element/Html.zig
+++ b/src/browser/webapi/element/Html.zig
@@ -382,12 +382,16 @@ pub fn setTitle(self: *HtmlElement, value: []const u8, frame: *Frame) !void {
     try self.asElement().setAttributeSafe(comptime .wrap("title"), .wrap(value), frame);
 }
 
-// HTML §7.7.5.2: returns true iff the element's effective content editable
-// state is "true" or "plaintext-only". Walk up to the nearest ancestor (or
-// self) with a `contenteditable` attribute; the keyword "false" maps to the
-// false state, every other value (including the empty string, "true", and
-// "plaintext-only") maps to an editable state. With no ancestor carrying the
-// attribute the document's designMode default ("off") yields false.
+// HTML §7.7.5.2 specifies the IDL attribute as true iff the element's effective
+// content editable state is "true" or "plaintext-only". Lightpanda has no
+// caret/keyboard editing pipeline, so a true answer cannot be honored
+// end-to-end — downstream CDP tools (notably Puppeteer's dispatchKeyEvent
+// path) would route into an input pipeline that silently no-ops. Always
+// return false, and log .not_implemented when the spec would have said true
+// so usage surfaces in telemetry rather than silently depending on an
+// unsupported value. Spec walk per HTML §7.7.5.2 still applies — the nearest
+// ancestor with `contenteditable` wins; "false" disables. See PR #2310 for
+// the routing-vs-fail-loud discussion.
 //
 // "contenteditable" is 15 bytes — past the comptime SSO limit — so the
 // String wrap runs at runtime, mirroring the pattern in interactive.zig.
@@ -395,7 +399,10 @@ pub fn getIsContentEditable(self: *HtmlElement) bool {
     var current: ?*Element = self.asElement();
     while (current) |el| : (current = el.parentElement()) {
         const raw = el.getAttributeSafe(.wrap("contenteditable")) orelse continue;
-        return !std.ascii.eqlIgnoreCase(raw, "false");
+        if (!std.ascii.eqlIgnoreCase(raw, "false")) {
+            log.info(.not_implemented, "IsContentEditable", .{});
+        }
+        break;
     }
     return false;
 }
--- a/src/browser/tests/element/html/contenteditable.html
+++ b/src/browser/tests/element/html/contenteditable.html
@@ -1,0 +1,80 @@
+<!DOCTYPE html>
+<script src="../../testing.js"></script>
+
+<div id="own-true" contenteditable="true">own true</div>
+<div id="own-empty" contenteditable="">own empty</div>
+<div id="own-false" contenteditable="false">own false</div>
+<div id="own-plaintext" contenteditable="plaintext-only">own plaintext-only</div>
+<div id="own-uppercase" contenteditable="TRUE">own UPPERCASE</div>
+<div id="own-falseupper" contenteditable="FALSE">own FALSEUPPER</div>
+
+<div id="ancestor-true" contenteditable="true">
+  <span id="child-of-true">child</span>
+  <div><span id="grandchild-of-true">grandchild</span></div>
+</div>
+
+<div id="ancestor-false" contenteditable="false">
+  <span id="child-of-false">child</span>
+  <div contenteditable="true">
+    <span id="grandchild-rehosted">re-hosted</span>
+  </div>
+</div>
+
+<div id="plain"><span id="child-of-plain">no editing context</span></div>
+
+<script id="own-states">
+  {
+    testing.expectEqual(true,  document.getElementById('own-true').isContentEditable);
+    testing.expectEqual(true,  document.getElementById('own-empty').isContentEditable);
+    testing.expectEqual(false, document.getElementById('own-false').isContentEditable);
+    testing.expectEqual(true,  document.getElementById('own-plaintext').isContentEditable);
+    testing.expectEqual(true,  document.getElementById('own-uppercase').isContentEditable);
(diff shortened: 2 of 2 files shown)


src/browser/webapi/element/Html.zig
```diff
@@ -382,6 +382,24 @@ pub fn setTitle(self: *HtmlElement, value: []const u8, frame: *Frame) !void {
...
+// HTML §7.7.5.2: returns true iff the element's effective content editable
+// state is "true" or "plaintext-only". Walk up to the nearest ancestor (or
+// self) with a `contenteditable` attribute; the keyword "false" maps to the
+// false state, every other value (including the empty string, "true", and
+// "plaintext-only") maps to an editable state. With no ancestor carrying the
+// attribute the document's designMode default ("off") yields false.
+//
+// "contenteditable" is 15 bytes — past the comptime SSO limit — so the
+// String wrap runs at runtime, mirroring the pattern in interactive.zig.
+pub fn getIsContentEditable(self: *HtmlElement) bool {
```
karlseguin, line 398: This one is tough. We don't support content editing. What do you think about always returning false, but logging a `log.info(.not_implemented, "IsContentEditable", .{});`  when it _should_ return `true`?

navidemad: Before flipping it: worth flagging that downstream CDP automation tools (Puppeteer, Playwright, my capybara adapter gem too) use `isContentEditable` as a routing signal — `true` → mutate `innerHTML`/`textContent` directly; `false` → go through `.value` (inputs only) or skip the element. Lightpanda supports the programmatic-mutation path (innerHTML works on contenteditable hosts), so always-`false` would route those tools *away from* a code path the engine actually handles. The piece that genuinely doesn't work is the user-facing caret/keyboard side.

If the goal is observability rather than the return value itself, would spec-correct + `log.info(.not_implemented, "IsContentEditable", .{})` on the `true` branch work for you? You'd get usage telemetry, and downstream tools still route to a working path. Happy to ship the always-`false` + log version if you'd still prefer it — just wanted to surface the routing case first.

karlseguin: So I looked into this more and tried to understand what both Playwright and Puppeteer are doing (😭😭😭). With claude's help, I think Playwright is relying on `Input.insertText` and Puppeteer is relying on a combination of setting `innerText`d irectly and using both `Input.dispatchKeyEvent` and `Input.insertText` (seems to depend on the length of the input...why, I don't know). So Puppeteer is a real concern..plus the fact that we have 3 drivers (yours, puppeteer and playwright) doing this 3 different ways. I'm still leaning towards a "fail loud/early" answer.

We should probably make Frame.insertText contentEditable aware, because it's easy. But that doesn't help with Puppeteer. We're return `true` and then silently ignore the input. Far from ideal.

For the time being, can you handle this in your driver? Checking the `contenteditable` attribute on the element, rather than the isContentEditable getter?

navidemad: Done in 2af95af6. `getIsContentEditable` now always returns `false`; the §7.7.5.2 ancestor walk runs only to gate `log.info(.not_implemented, "IsContentEditable", .{})` when the spec would have said true, so misuse surfaces in telemetry. Test fixture trimmed to assert always-false across the same shape of inputs. On the gem side: my polyfill already walks `contenteditable` attributes on ancestors, so the always-false getter just falls through to the working path — no driver-side change needed. Happy to also pick up `Frame.insertText` contentEditable-aware as a follow-up if you want it tracked separately.

PR #2315, opened 2026-04-29
selector: walk fieldset/optgroup ancestors when matching :disabled

### LINKED ISSUE #2314: selector: `:disabled` does not inherit through ancestor `<fieldset disabled>` / `<optgroup disabled>`
## Summary

`Element.matches(':disabled')` reads only the element's own `disabled` content attribute. It does not honor the HTML spec's "actually disabled" inheritance: a form control inside a `<fieldset disabled>` should match `:disabled` (with the first-`<legend>` exception), and an `<option>` inside `<optgroup disabled>` should match `:disabled`.

The same evaluator backs `:disabled` / `:enabled` for `Element.matches`, `Element.querySelector`, `Element.querySelectorAll`, and CDP `DOM.performSearch` — every consumer sees the wrong answer.

## Today's behavior

`matchesPseudoClass` in `src/browser/webapi/selector/List.zig:537-541` returns `el.getAttributeSafe("disabled") != null` for `:disabled` (and the inverse for `:enabled`). No ancestor walk; no per-element-type rules.

`Element.isDisabled` (`src/browser/webapi/Element.zig:603-627`) DOES walk the ancestor `<fieldset disabled>` chain (with the first-`<legend>` exception), but the `:disabled` selector matcher doesn't call it. `isDisabled` also doesn't yet recognize the `<option>` + `<optgroup disabled>` case.

```mermaid
sequenceDiagram
    participant Client as CDP / JS Caller
    participant Match as matchesPseudoClass (List.zig)
    participant Attr as Element.attributes
    Client->>Match: el.matches(':disabled')
    Match->>Attr: getAttributeSafe("disabled")
    Attr-->>Match: null (own attr absent)
    Match-->>Client: false
    Note over Match,Attr: ancestor disabled state never consulted<br/>even though <fieldset disabled> / <optgroup disabled> exist
```

## Expected behavior

Per the [HTML pseudo-class definition](https://html.spec.whatwg.org/multipage/semantics-other.html#selector-disabled), `:disabled` matches:

- `button` / `input` / `select` / `textarea` / form-associated custom element that is *disabled* — own `disabled` attribute set, **or** a descendant of a `<fieldset disabled>` and not a descendant of that fieldset's first `<legend>` child ([HTML §4.10.18.3 — "Enabling and disabling form controls"](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#enabling-and-disabling-form-controls))
- `optgroup` whose `disabled` attribute is set
- `option` that is *disabled* — own `disabled` attribute set, **or** parent is an `<optgroup disabled>` ([HTML §4.10.10 — `<option>` "concept-option-disabled"](https://html.spec.whatwg.org/multipage/form-elements.html#concept-option-disabled))

Note `<option>` inside `<select disabled>` does NOT match `:disabled` — only an `<optgroup disabled>` parent contributes. Verified empirically against Chrome 130.

```mermaid
sequenceDiagram
    participant Client as CDP / JS Caller
    participant Match as matchesPseudoClass (List.zig)
    participant IsDis as Element.isDisabled
    Client->>Match: el.matches(':disabled')
    Match->>IsDis: el.isDisabled()
    IsDis->>IsDis: own attr? -> if set, return true
    IsDis->>IsDis: <option>? -> check optgroup parent
    IsDis->>IsDis: walk ancestors for <fieldset disabled><br/>(skip elements inside the first <legend>)
    IsDis-->>Match: true / false
    Match-->>Client: true / false
```

## Reproducer

Self-contained: one HTML fixture + one Node CDP driver + one shell harness. No Ruby or framework dependencies; only `node` (with `ws` auto-installed on first run), `python3`, and `lightpanda` on `PATH`.

`repro.html`:
```html
<!doctype html>
<meta charset="utf-8">
<body>
<form>
  <fieldset id="fs" disabled>
    <legend><input id="legend_input" type="text"></legend>
    <input id="fs_input" type="text">
    <button id="fs_button" type="button">btn</button>
  </fieldset>

  <select id="sel_disabled" disabled>
    <option id="opt_in_dis_select" value="a">a</option>
  </select>

  <select id="sel_enabled">
    <optgroup id="og" disabled label="g">
      <option id="opt_in_dis_optgroup" value="b">b</option>
    </optgroup>
    <option id="opt_loose" value="c">c</option>
  </select>

  <input id="own_disabled" type="text" disabled>
  <input id="own_enabled" type="text">
</form>
</body>
```

`probe.js` (drives the assertions over CDP via raw `ws`):
```js
const { connect } = require('./cdp.js'); // probe-lib helper, vendored alongside

(async () => {
  const c = await connect();
  await c.navigate('http://127.0.0.1:8000/repro.html');

  const cases = [
    ['fs_input',           true,  '<input> inside <fieldset disabled>'],
    ['fs_button',          true,  '<button> inside <fieldset disabled>'],
    ['legend_input',       false, '<input> inside first <legend> of <fieldset disabled>'],
    ['opt_in_dis_select',  false, '<option> inside <select disabled> (no optgroup) — NOT disabled'],
    ['opt_in_dis_optgroup',true,  '<option> inside <optgroup disabled>'],
    ['opt_loose',          false, '<option> outside any disabled ancestor'],
    ['own_disabled',       true,  '<input disabled> (own attribute)'],
    ['own_enabled',        false, '<input> with no disabled'],
  ];

  let bugs = 0;
  for (const [id, expected, label] of cases) {
    const got = await c.eval(`document.getElementById(${JSON.stringify(id)}).matches(':disabled')`);
    const ok = got === expected;
    console.log(`  ${ok ? 'OK' : 'BUG'} #${id}.matches(':disabled') = ${got} (expected ${expected}) — ${label}`);
    if (!ok) bugs++;
  }

  c.close();
  process.exit(bugs > 0 ? 1 : 0);
```

`repro.sh`:
```bash
#!/usr/bin/env bash
set -euo pipefail
HERE="$(cd "$(dirname "$0")" && pwd)"; cd "$HERE"
LP_BIN="${LIGHTPANDA_BIN:-lightpanda}"
LP_PORT=9222; HTTP_PORT=8000

cleanup() { [[ -n "${LP_PID:-}" ]] && kill "$LP_PID" 2>/dev/null || true
            [[ -n "${HTTP_PID:-}" ]] && kill "$HTTP_PID" 2>/dev/null || true; }
trap cleanup EXIT
lsof -ti tcp:$LP_PORT 2>/dev/null | xargs -r kill -9 2>/dev/null || true
lsof -ti tcp:$HTTP_PORT 2>/dev/null | xargs -r kill -9 2>/dev/null || true

python3 -m http.server $HTTP_PORT --bind 127.0.0.1 >/dev/null 2>&1 & HTTP_PID=$!
"$LP_BIN" serve --host 127.0.0.1 --port $LP_PORT >/tmp/lp-disabled-probe.log 2>&1 & LP_PID=$!

for i in {1..40}; do
  curl -sf "http://127.0.0.1:$LP_PORT/json/version" >/dev/null 2>&1 && break
  sleep 0.1
done

if ! node -e "require('ws')" 2>/dev/null; then npm install --no-save --silent ws >/dev/null; fi
node probe.js
```

`cdp.js` is the standard probe-lib helper (raw `ws` wrapper with `Target.createTarget` + `Target.attachToTarget` flow). Available alongside the rest of the reproducer in the linked PR's `repro/` directory.

**Run**: `bash repro.sh`

- Today (`1.0.0-dev.5839+2bbf23b3`): exits **1** with output:

  ```
    BUG #fs_input.matches(':disabled') = false (expected true) — <input> inside <fieldset disabled>
    BUG #fs_button.matches(':disabled') = false (expected true) — <button> inside <fieldset disabled>
    OK  #legend_input.matches(':disabled') = false (expected false) — ...
    OK  #opt_in_dis_select.matches(':disabled') = false (expected false) — ...
    BUG #opt_in_dis_optgroup.matches(':disabled') = false (expected true) — <option> inside <optgroup disabled>
    OK  #opt_loose.matches(':disabled') = false (expected false) — ...
    OK  #own_disabled.matches(':disabled') = true (expected true) — ...
    OK  #own_enabled.matches(':disabled') = false (expected false) — ...
  ```

- Expected after the fix: exits **0** with all 8 cases reported `OK`.

## Likely fix location

- `src/browser/webapi/selector/List.zig` — `matchesPseudoClass` arms `.disabled` and `.enabled`: route through `Element.isDisabled` instead of reading `disabled` directly.
- `src/browser/webapi/Element.zig` — `isDisabled`: extend to recognize the `<option>` + `<optgroup disabled>` parent case (currently only walks the fieldset chain).

## Environment

- Lightpanda: `1.0.0-dev.5839+2bbf23b3` (public nightly, asset published 2026-04-28 03:33 UTC)
- OS: macOS aarch64 (Darwin 25.4.0)
- Reference: Chrome 130 confirms the spec-correct values above

navidemad:
## What

`Element.matches(':disabled')` ignored the HTML "actually disabled" inheritance: a form control inside `<fieldset disabled>` reported `false`, and an `<option>` inside `<optgroup disabled>` reported `false`. The same evaluator backs `:disabled`/`:enabled` for `Element.matches`, `querySelector{,All}`, and CDP `DOM.performSearch`, so every consumer saw the wrong answer.

This PR routes the `:disabled`/`:enabled` pseudo-class matchers through `Element.isDisabled` (which already walks the ancestor `<fieldset disabled>` chain with the first-`<legend>` exception) and extends `isDisabled` to recognize the `<option>` + `<optgroup disabled>` case. It also gates the disabled concept on the form-control allowlist from HTML "concept-fe-disabled" so non-form-controls (e.g. `<div>`) never match `:disabled` or `:enabled`.

Closes #2314.

## Root cause

The pseudo-class matchers checked only the element's own `disabled` content attribute:

```zig
// before — src/browser/webapi/selector/List.zig:537-541
.disabled => return el.getAttributeSafe(comptime .wrap("disabled")) != null,
.enabled  => return el.getAttributeSafe(comptime .wrap("disabled")) == null,
```

`Element.isDisabled` already implemented the spec-correct fieldset walk for callers that needed it (FormData, accessibility tree, semantic tree, MCP `detectForms`), but the selector path never invoked it. `isDisabled` also didn't yet cover the `<option>` + `<optgroup disabled>` rule from [HTML §4.10.10 "concept-option-disabled"](https://html.spec.whatwg.org/multipage/form-elements.html#concept-option-disabled), and there was no gate restricting the disabled concept to the form-control allowlist from [HTML §4.10.18.5 "concept-fe-disabled"](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#concept-fe-disabled), so `<div disabled>.matches(':disabled')` returned `true` and `<div>.matches(':enabled')` returned `true`.

```mermaid
flowchart LR
    A[el.matches :disabled] --> B[selector List.zig - own attr only]
    B --> C[false for fieldset/optgroup descendant]
    style B fill:#fdd
    style C fill:#fdd
```

## Fix

- `src/browser/webapi/Element.zig` — new `hasDisabledConcept()` returns `true` only for `button`, `input`, `select`, `textarea`, `optgroup`, `option`, `fieldset` (per HTML "concept-fe-disabled"). `isDisabled` returns `false` early when the element doesn't participate in the disabled concept, then keeps the existing logic: own `[disabled]` attribute, `<option>` + `<optgroup disabled>` parent rule, ancestor `<fieldset disabled>` walk with the first-`<legend>` exception.
- `src/browser/webapi/selector/List.zig` — `matchesPseudoClass`: `:disabled` arm delegates to `el.isDisabled()`. `:enabled` arm gates on `el.hasDisabledConcept() and !el.isDisabled()` so non-form-controls no longer match `:enabled` (matches Chrome).

```mermaid
flowchart LR
    A[el.matches :disabled or :enabled] --> B{hasDisabledConcept?}
    B -- no  --> Z[both return false]
    B -- yes --> C[Element.isDisabled]
    C --> D{tag is option?}
    D -- yes --> E[check optgroup parent]
    D -- no  --> F[walk ancestors for fieldset disabled<br/>skip first legend descendants]
    E --> G[true / false]
    F --> G
    style B fill:#dfd
    style G fill:#dfd
    style Z fill:#dfd
```

## Test

- **HTML fixture**: `src/browser/tests/element/disabled_inheritance.html` — six `<script>` blocks covering fieldset inheritance + first-`<legend>` exception, `<optgroup disabled>` propagation to `<option>`, intentional non-propagation from `<select disabled>` to `<option>`, own-attribute baseline, `:enabled` complement, the disabled-concept gate (`<div>` / `<span>` / `<div disabled>` never match either pseudo), and a `querySelectorAll(':disabled')` end-to-end check. Picked up by the existing `webapi/Element.zig`'s `htmlRunner("element", .{})`.
- **Pre-fix**: `TEST_FILTER='WebApi: Element#disabled_inheritance.html' mise exec -- zig build test $V8` reports `0 of 1 test passed` (production lines reverted to `origin/main` via `git checkout origin/main -- <files>` for the toggle-off check).
- **Post-fix**: same command reports `1 of 1 test passed`. Full unit-test suite (`mise exec -- zig build test $V8`) is `497 of 497 tests passed`.
- **End-to-end reproducer**: the script attached to #2314 exits 1 against `1.0.0-dev.5839+2bbf23b3` (3 of 8 cases mismatch) and exits 0 against this branch's locally-built `zig-out/bin/lightpanda` (all 8 OK).

## Notes

- `Element.isDisabled` is also called from `webapi/net/FormData.zig`, `cdp/AXNode.zig`, `SemanticTree.zig`, `browser/forms.zig`, and `browser/interactive.zig`. The new `<option>` branch makes those callers report `<option>` + `<optgroup disabled>` as disabled too — that's consistent with the spec and is the expected behavior for AX trees and semantic categorization. The new `hasDisabledConcept` gate is also a no-op for those callers since they iterate form-associated elements only.

### DIFF
--- a/src/browser/webapi/selector/List.zig
+++ b/src/browser/webapi/selector/List.zig
@@ -538,7 +538,7 @@ fn matchesPseudoClass(el: *Node.Element, pseudo: Selector.PseudoClass, scope: *N
             return el.isDisabled();
         },
         .enabled => {
-            return !el.isDisabled();
+            return el.hasDisabledConcept() and !el.isDisabled();
         },
         .indeterminate => {
             const input = el.is(Node.Element.Html.Input) orelse return false;
--- a/src/browser/webapi/selector/List.zig
+++ b/src/browser/webapi/selector/List.zig
@@ -535,10 +535,10 @@ fn matchesPseudoClass(el: *Node.Element, pseudo: Selector.PseudoClass, scope: *N
             return input.getChecked();
         },
         .disabled => {
-            return el.getAttributeSafe(comptime .wrap("disabled")) != null;
+            return el.isDisabled();
         },
         .enabled => {
-            return el.getAttributeSafe(comptime .wrap("disabled")) == null;
+            return !el.isDisabled();
         },
         .indeterminate => {
             const input = el.is(Node.Element.Html.Input) orelse return false;
--- a/src/browser/tests/element/disabled_inheritance.html
+++ b/src/browser/tests/element/disabled_inheritance.html
@@ -1,0 +1,111 @@
+<!DOCTYPE html>
+<script src="../testing.js"></script>
+
+<form>
+  <fieldset id="fs" disabled>
+    <legend><input id="legend_input" type="text"></legend>
+    <input id="fs_input" type="text">
+    <button id="fs_button" type="button">btn</button>
+    <select id="fs_select"></select>
+    <textarea id="fs_textarea"></textarea>
+  </fieldset>
+
+  <fieldset id="fs_enabled">
+    <legend><input id="enabled_legend_input"></legend>
+    <input id="fs_enabled_input">
+  </fieldset>
+
+  <select id="sel_disabled" disabled>
+    <option id="opt_in_disabled_select">a</option>
+  </select>
+
+  <select id="sel_enabled">
+    <optgroup id="og_disabled" disabled label="g">
+      <option id="opt_in_disabled_optgroup">b</option>
+    </optgroup>
+    <optgroup id="og_enabled" label="h">
+      <option id="opt_in_enabled_optgroup">c</option>
+    </optgroup>
+    <option id="opt_loose">d</option>
+  </select>
+
+  <input id="own_disabled" type="text" disabled>
+  <input id="own_enabled" type="text">
+</form>
+
+<script id="fieldset_inheritance">
+{
+  // Form controls inside <fieldset disabled> match :disabled.
+  testing.expectTrue($('#fs_input').matches(':disabled'));
+  testing.expectTrue($('#fs_button').matches(':disabled'));
+  testing.expectTrue($('#fs_select').matches(':disabled'));
+  testing.expectTrue($('#fs_textarea').matches(':disabled'));
+
+  // The fieldset's first <legend> exempts its descendants.
+  testing.expectFalse($('#legend_input').matches(':disabled'));
+
+  // Enabled fieldset doesn't propagate.
+  testing.expectFalse($('#fs_enabled_input').matches(':disabled'));
+  testing.expectFalse($('#enabled_legend_input').matches(':disabled'));
+}
+</script>
+
+<script id="optgroup_inheritance">
+{
+  // <option> inside <optgroup disabled> matches :disabled.
+  testing.expectTrue($('#opt_in_disabled_optgroup').matches(':disabled'));
+
+  // <option> outside any disabled ancestor does not.
+  testing.expectFalse($('#opt_in_enabled_optgroup').matches(':disabled'));
+  testing.expectFalse($('#opt_loose').matches(':disabled'));
+
+  // <option> inside <select disabled> (no optgroup) does NOT match :disabled
+  // per HTML "concept-option-disabled" — only own attr or <optgroup disabled>
+  // parent contributes. <select disabled> itself matches, but does not
+  // propagate to its <option> children for selector purposes.
+  testing.expectFalse($('#opt_in_disabled_select').matches(':disabled'));
+}
+</script>
+
+<script id="own_attribute">
+{
+  testing.expectTrue($('#own_disabled').matches(':disabled'));
+  testing.expectFalse($('#own_enabled').matches(':disabled'));
+
+  // The disabled <fieldset>/<select>/<optgroup> elements themselves match.
+  testing.expectTrue($('#fs').matches(':disabled'));
+  testing.expectTrue($('#sel_disabled').matches(':disabled'));
+  testing.expectTrue($('#og_disabled').matches(':disabled'));
+}
+</script>
+
+<script id="enabled_complement">
+{
+  // :enabled is the negation of :disabled for elements that have a
+  // disabled concept. An <input> inside <fieldset disabled> is :disabled,
+  // so it must NOT also be :enabled.
+  testing.expectFalse($('#fs_input').matches(':enabled'));
+  testing.expectFalse($('#opt_in_disabled_optgroup').matches(':enabled'));
+  testing.expectTrue($('#fs_enabled_input').matches(':enabled'));
+  testing.expectTrue($('#opt_loose').matches(':enabled'));
+  // Legend descendants are enabled despite the disabled fieldset.
+  testing.expectTrue($('#legend_input').matches(':enabled'));
+}
+</script>
+
+<script id="querySelectorAll">
+{
+  // querySelectorAll(':disabled') should find every disabled element.
+  const ids = Array.from(document.querySelectorAll(':disabled')).map(e => e.id).sort();
+  // Expected (sorted):
+  //   fs, fs_button, fs_input, fs_select, fs_textarea,
+  //   og_disabled, opt_in_disabled_optgroup,
+  //   own_disabled, sel_disabled
+  const expected = [
+    'fs', 'fs_button', 'fs_input', 'fs_select', 'fs_textarea',
+    'og_disabled', 'opt_in_disabled_optgroup',
+    'own_disabled', 'sel_disabled',
+  ].sort();
+  testing.expectEqual(JSON.stringify(expected), JSON.stringify(ids));
+}
+</script>
--- a/src/browser/tests/element/disabled_inheritance.html
+++ b/src/browser/tests/element/disabled_inheritance.html
@@ -33,6 +33,10 @@
   <input id="own_enabled" type="text">
 </form>
 
+<div id="div_plain">x</div>
+<div id="div_with_disabled_attr" disabled>y</div>
+<span id="span_plain">z</span>
+
 <script id="fieldset_inheritance">
 {
   // Form controls inside <fieldset disabled> match :disabled.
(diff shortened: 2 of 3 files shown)


src/browser/webapi/selector/List.zig
```diff
@@ -535,10 +535,10 @@ fn matchesPseudoClass(el: *Node.Element, pseudo: Selector.PseudoClass, scope: *N
             return input.getChecked();
         },
         .disabled => {
-            return el.getAttributeSafe(comptime .wrap("disabled")) != null;
+            return el.isDisabled();
         },
         .enabled => {
-            return el.getAttributeSafe(comptime .wrap("disabled")) == null;
+            return !el.isDisabled();
```
karlseguin: It seems like the original code was wrong for another reason: some controls can never be disabled. I think we should fix that too.  I think we need a something like:

```zig
  pub fn hasDisabledConcept(self: *Element) bool {
      return switch (self.getTag()) {
          .button, .input, .select, .textarea, .optgroup, .option, .fieldset => true,
          else => false,
      };
  }
```

And `isDisabled` should check this first and always return `false` when `self.hasDisabledConcept() == false`. On top of that, `:enabled` should also never match when `el.hasDisabledConcept() == false`, so it would need to be:

```
.disabled => el.isDisabled(), 
.enabled => el.hasDisabledConcept() and !el.isDisabled(),

...


pub fn isDisabled(self: *Element) bool 
  if (self.hasDisabledConcept() == false) {
    return false;
  }
  ...
}
```

Don't love the double `hasDisabledConcept` check in `hasDisabledConcept`, but I can live with it.

navidemad: Adopted in `aca65188`. Added `Element.hasDisabledConcept()` returning `true` for `button, input, select, textarea, optgroup, option, fieldset`, gated `isDisabled` on it, and updated the `:enabled` arm to `el.hasDisabledConcept() and !el.isDisabled()`. Fixture extended with `<div>`, `<span>`, and `<div disabled>` cases asserting both pseudo-classes return `false`. Toggle-off + `497 of 497 tests passed`.

karlseguin:
@arrufat heads up, this could impact SemanticTree and friends, as it changes what we consider "disabled". Generally, fewer things will now be disabled, e.g. a div cannot be disabled. Probably very little practical impact, but just flagging.

PR #2329, opened 2026-04-29
Deferred Layer

mookums:
This solves the issue of scripts running while blocking scripts are running by introducing a `DeferringLayer`. This layer defers the completion of requests while a blocking request is running, clearing them later on `staticScriptsDone`.

This tracks the currently active request by its ID ensuring that each frame can only have one blocking request at a time and deferring the scripts.

### DIFF
--- a/src/browser/ScriptManagerBase.zig
+++ b/src/browser/ScriptManagerBase.zig
@@ -390,17 +390,6 @@ pub fn staticScriptsDone(self: *ScriptManagerBase) void {
     lp.assert(self.static_scripts_done == false, "ScriptManagerBase.staticScriptsDone", .{});
     self.static_scripts_done = true;
 
-    const frame_id = self.owner.frameId();
-    if (comptime IS_DEBUG) {
-        const blocking_request_id = self.client.blocking_requests.get(frame_id);
-        lp.assert(
-            blocking_request_id == null,
-            "there should be no blocking request on this frame",
-            .{ .value = blocking_request_id },
-        );
-    }
-    self.client.deferring_layer.flushFrame(frame_id);
-
     self.evaluate();
 }
 
--- a/src/browser/ScriptManagerBase.zig
+++ b/src/browser/ScriptManagerBase.zig
@@ -389,6 +389,18 @@ pub fn getAsyncImport(self: *ScriptManagerBase, url: [:0]const u8, cb: ImportAsy
 pub fn staticScriptsDone(self: *ScriptManagerBase) void {
     lp.assert(self.static_scripts_done == false, "ScriptManagerBase.staticScriptsDone", .{});
     self.static_scripts_done = true;
+
+    const frame_id = self.owner.frameId();
+    if (comptime IS_DEBUG) {
+        const blocking_request_id = self.client.blocking_requests.get(frame_id);
+        lp.assert(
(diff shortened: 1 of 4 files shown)


karlseguin:
I think there's a serious UAF here. The transfer of an deferred request will be deinit'd while the request still lives in the deferred list. I tested this out by writing a log when transfer.deinit is called, and that always happens before the request is "fired".

Another issue is that this defers the done/error/shutdown  but the header/data are just as important to block from firing.

karlseguin:
A while ago, I looked at something that did something similar - delaying new requests until blocking requests were complete. I don't remember exactly what I measured, but it was that it would introduce significant latency. You can imagine:

```js
const xhr = new XMLHttpRequest();
...
xhr.send();
// do something that take 500ms
```

If the DeferredLayer defers this request, then it adds 500ms before the request _starts_. You want to start the HTTP request ASAP, just not process the callbacks.


I also think the code current leaks requests, e.g. what happens if `deinit` is called while there are still deferred requests?

mookums:
Reworked it a bit around your new changes with Transfer and all.

src/browser/ScriptManagerBase.zig
```diff
@@ -395,6 +395,18 @@ pub fn getAsyncImport(self: *ScriptManagerBase, url: [:0]const u8, cb: ImportAsy
...
+    const frame_id = self.owner.frameId();
+    if (comptime IS_DEBUG) {
+        const blocking_request_id = self.client.blocking_requests.get(frame_id);
+        lp.assert(
+            blocking_request_id == null,
+            "there should be no blocking request on this frame",
+            .{ .value = blocking_request_id },
+        );
+    }
+    self.client.deferring_layer.flushFrame(frame_id);
```
karlseguin: So every callsite that wants to make a blocking request (and there are more), needs to do this?

src/network/layer/DeferringLayer.zig
```diff
@@ -0,0 +1,343 @@
...
+        self.terminal = true;
+        self.buffered.append(self.arena, .shutdown) catch {};
+    }
+
+    // Replay all buffered events in order, then clean up.
+    fn fire(self: *DeferredContext) void {
+        defer self.deinit();
+
+        const req = self.transfer.req;
+        const stable_response = self.stable_resp.?;
```
karlseguin: `firePartial` does: `const stable_response = self.stable_resp orelse return;`  which makes me think this isn't always not-null?

src/browser/ScriptManagerBase.zig
mookums: I have it flush the frame after the evaluation of the `syncRequest` script completes now which seems to work fine. I think generally yes, we would need to evaluate whatever script we got synchronously (that potentially blocked others) and then flush any other requests after.

src/network/layer/DeferringLayer.zig
mookums: Yeah, fixed this by having it set the stable response whenever it can and also ensuring that it is present for the events that require it.

mookums:
Fixed some correctness issues as well regarding the `DeferredContext` arena and double-frees on shutdown during `forwardDone`.

mookums:
Memory regression test failure seems to be fixed by a rebase.

PR #2331, opened 2026-04-30
browser: change node filter from lower to tag_name

sunguru98:
This circumvents the tag name lower logic, as the same matching happens in node_live.zig based on namespace. 
WPT tests after.
```json
== WPT Results==
{
  "url": "http://127.0.0.1:8000/dom/nodes/Element-getElementsByTagName.html",
  "status": "",
  "summary": {
    "total": 19,
    "passed": 17,
    "failed": 2,
    "timeout": 0,
    "notrun": 0,
    "unsupported": 0
  },
  "not_passed": [
    {
      "name": "Shouldn't be able to set unsigned properties on a HTMLCollection (strict mode)",
      "status": "Fail",
      "message": "assert_throws_js: function \"function() {       \"use strict\";       l[5] = \"foopy\"     }\" did not throw"
    },
    {
      "name": "hasOwnProperty, getOwnPropertyDescriptor, getOwnPropertyNames",
      "status": "Fail",
      "message": "assert_equals: expected (object) Element node <pre id=\"x\"></pre> but got (undefined) undefined"
    }
  ]
}
```

### DIFF
--- a/src/browser/webapi/Node.zig
+++ b/src/browser/webapi/Node.zig
@@ -989,7 +989,7 @@ pub fn getElementsByTagName(self: *Node, tag_name: []const u8, frame: *Frame) !G
     }
 
     const arena = frame.arena;
-    const filter = try String.init(arena, lower, .{});
+    const filter = try String.init(arena, tag_name, .{});
     return .{ .tag_name = collections.NodeLive(.tag_name).init(self, filter, frame) };
 }
 


karlseguin:
Just to add a note, I thought this change would break the [much more] common case of where the HTML tag is lower-case, but the input to  gtElementsByTagName is uppercase. However, the node_live tag_name branch already handles case-insensitive matching for HTML elements. So the forced lowercase we were doing before was only interfering with the correct namespace/case-sensitive handling

Issue #1986, opened 2026-03-24
npm package naming: @lightpanda/browser vs the actual browser


illutr0n:
The npm package `@lightpanda/browser` is versioned independently from the browser binary, which causes confusion — someone comparing npm (1.2.0) to GitHub releases (0.2.6) has no obvious way to know they're unrelated.

Something like `@lightpanda/node` or `@lightpanda/sdk` would make the relationship clearer without implying it's the browser itself.

PR #2332, opened 2026-04-30
docs(readme): clarify browser vs npm package versioning

Gujiassh:
## Summary
- add a short package naming note to the quick start section
- clarify that the browser binary in this repository is versioned independently from npm packages under the `@lightpanda/*` scope
- explain that the npm packages are SDK/integration packages rather than the browser runtime itself

## Why
Issue #1986 reports confusion when users compare GitHub browser releases with npm package versions and assume they should match. This note makes the separation explicit in the main install path without renaming packages.

## Testing
- `git diff --check origin/main...HEAD`

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,14 @@ Due to the nature of Playwright, a script that works with the current version of
 
 ## Quick start
 
+### Package naming note
+
+The browser binary in this repository is versioned independently from npm packages under the
+`@lightpanda/*` scope.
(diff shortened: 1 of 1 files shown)


krichprollsch:
Thanks for the contribution, but I think this precision must land direclty into the package's README: https://github.com/lightpanda-io/node-packages/tree/main/packages/browser

PR #2333, opened 2026-04-30
Display actual port when binding --port 0

webcoyote:
## Summary
- When `--port 0` is used, the OS assigns an ephemeral port. After binding, call `getsockname` to read back the actual port and update the address so log output shows the real port instead of `0`.
- Changes `Network.bind` to take `*net.Address` (pointer) so the bound address can be written back to the caller.

## Test plan
- [x] Run `zig-out/bin/browser --port 0` and verify the log line shows the actual assigned port (e.g. `:54321`) instead of `:0`
- [x] Run with an explicit port (`--port 8080`) and verify behavior is unchanged

### DIFF
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -57,8 +57,9 @@ pub fn init(app: *App, address: net.Address) !*Server {
         .clients_pool = std.heap.MemoryPool(Client).init(allocator),
     };
 
-    try self.app.network.bind(address, self, onAccept);
-    log.info(.app, "server running", .{ .address = address });
+    var bound_address = address;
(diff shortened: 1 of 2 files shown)


webcoyote:

recheck

webcoyote:
I guess I should have mentioned: this is useful because scripts can launch the browser and read back the actual port-in-use, simplifying automation. Right now I have a python script that binds an ephemeral port and then relaunches browser with that port, which is less-than-ideal (TOCTOU problem).

karlseguin:
Thanks for this :)

PR #2338, opened 2026-04-30
Add interactive agent mode with LLM-powered web browsing

arrufat:
Adds `lightpanda agent`: an interactive agent built on top of the same headless browser, plus the shared browser-tool layer that backs both the agent and the MCP server.

### What it does

`lightpanda agent` runs in four modes that all share the same browser tools (`goto`, `click`, `fill`, `tree`, `markdown`, `search`, `extract`, …):

- **LLM REPL** — an agent that drives the browser via tool calls (Anthropic, OpenAI, Gemini, Ollama), with auto-detection of an API key from the environment.
- **One-shot `--task`** — runs a single task and prints the answer to stdout.
- **Basic REPL (`--no-llm`)** — hand-driven slash commands, no LLM or API key required.
- **Scripted runner** — records a session and replays it deterministically as a JavaScript `.js` script.

The same tool surface is exposed over MCP via `lightpanda mcp`, so an external LLM agent (Claude Code, any MCP client) can drive Lightpanda without Lightpanda needing its own API key.

### Highlights

- **Shared tool layer** (`src/browser/tools.zig`) used by both the agent and the MCP server — page reading (`tree`, `markdown`, `html`, `links`, `nodeDetails`, `findElement`, `interactiveElements`), interaction (`click`, `fill`, `hover`, `press`, `scroll`, `selectOption`, `setChecked`), navigation/eval (`goto`, `eval`, `waitForSelector`, `waitForScript`), and structured extraction (`extract` with a JSON-schema-driven follow option).
- **JavaScript scripting** — recorded sessions are emitted and replayed as JavaScript (replacing the earlier PandaScript approach). `ScriptRuntime.zig` marshals tool calls; recorded tools without marshalling are now a compile-time error.
- **Interactive REPL** built on a vendored/patched isocline: slash commands, syntax highlighting, completions, ghost-text hints, JS evaluation mode, persistent history, kitty keyboard protocol, and a status spinner.
- **`/save`** synthesizes a clean replayable script from a recorded session using the LLM.
- Cooperative cancellation (Ctrl-C), model/provider picker, env-var (`LP_*`) substitution and redaction, console-log capture, and safety handling for model refusals.


docs/agent.md
```diff
@@ -0,0 +1,226 @@
...
+command. Comments start with `#`. Strings are quoted with `'`, `"`, or `'''…'''`
+for values that mix both quote styles. Quoting rules are content-aware so that
+recorded scripts round-trip through the parser.
+
+| Command          | Form                                  | Notes                                                |
+|------------------|---------------------------------------|------------------------------------------------------|
+| `GOTO`           | `GOTO <url>`                          | Navigate. URL is unquoted.                           |
+| `CLICK`          | `CLICK '<selector>'`                  | CSS selector.                                        |
+| `TYPE`           | `TYPE '<selector>' '<value>'`         | Fills an input. `$LP_*` env refs auto-resolve.       |
+| `WAIT`           | `WAIT '<selector>'`                   | Wait for selector.                                   |
```
karlseguin: "Wait for selector ..." to be present or visible?

src/browser/Frame.zig
```diff
@@ -256,6 +256,7 @@ _parent_notified: bool = false,
 
 _type: enum { root, frame }, // only used for logs right now
 _req_id: u32 = 0,
+_console_messages: std.ArrayListUnmanaged(ConsoleMessage) = .{},
```
karlseguin:
```suggestion
_console_messages: std.ArrayList(ConsoleMessage) = .{},
```

src/browser/Frame.zig
```diff
@@ -2785,6 +2786,26 @@ fn isXmlNameChar(c: u21) bool {
         (c >= 0x203F and c <= 0x2040);
 }
 
+const max_console_messages = 1000;
+
+pub fn appendConsoleMessage(self: *Frame, level: ConsoleMessage.Level, values: []JS.Value) void {
```
karlseguin: There are two things here that could lead to a bit of memory wastage:
1 - This limits to 1000 messages, but has no limit on message size
2 - While `_console_messages` is cleared + retained, the underlying string is not

Ideally, this hooks into a notification: https://github.com/lightpanda-io/browser/pull/2339

Whether or not that happens, instead of staging the ConsoleMessage in `_console_messages`, I'd make `console_messages an `Io.Writer.Allocating` and format the final output directly into it, with a trailing `\n`.

So it goes from:
1 - dupe the string  // retained until frame end, most  memory is here
2 - Append into the arraylist
3 - Reuse the arraylist

To:
1 - Write into the Writer
2 - Reuse the Writer  

Allows you to apply a more meaningful max length (total message size) and better memory re-use.

src/browser/tools.zig
```diff
@@ -0,0 +1,1054 @@
...
+};
+
+const google_block_url_marker = "/sorry/";
+const google_block_text_marker = "detected unusual traffic";
+
+fn execSearch(session: *lp.Session, arena: std.mem.Allocator, registry: *CDPNode.Registry, arguments: ?std.json.Value) ToolError![]const u8 {
+    const args = try parseArgsOrErr(SearchParams, arena, arguments) orelse return ToolError.InvalidParams;
+    if (args.query.len == 0) return ToolError.InvalidParams;
+
+    const encoded = percentEncodeQuery(arena, args.query) catch return ToolError.OutOfMemory;
```
karlseguin: There is `try URL.percentEncodeSegment(allocator, args.query, .query)`  if you want to make it public

src/browser/tools.zig
```diff
@@ -0,0 +1,1054 @@
...
+    getCookies,
+};
+
+pub fn call(
+    session: *lp.Session,
+    arena: std.mem.Allocator,
+    registry: *CDPNode.Registry,
+    tool_name: []const u8,
+    arguments: ?std.json.Value,
+) ToolError![]const u8 {
```
karlseguin: You could pass an `*std.Io.Writer` and write directly to it, rather than using a `Writer.Allocating` in each call.  The initial implementation could simply use a `Writer.Allocating` itself, but then at least it's a detail of the caller. 

In the end, the benefit is that you could probably go directly from result -> stderr, without the intermediary buffering. BUT, I say that and I'm 100% sure I'd do it exactly like you did because you'd really have to write your own Writer that wraps another Writer (stderr) to implement the max-length and stuff..and Zig's IO is just a pain.

Issue #1953, opened 2026-03-22
[Bug]: Missing console API coverage breaks browser automation scripts that rely on console.log interception


ai-nurmamat:
### Summary
Lightpanda lacks implementation of the `console` Web API, which is critical for browser automation scripts that intercept `console.log`, `console.error`, `console.warn`, etc. Many popular automation frameworks (Playwright, Puppeteer) rely on console event listeners for debugging, error tracking, and page state verification.

### Problem
When running browser automation scripts that use `page.on('console', ...)` or similar console interception APIs, Lightpanda fails to capture or forward console messages from the page's JavaScript execution. This breaks a fundamental pattern in browser automation.

### Root Cause Analysis

**Code Location**: `src/console/` directory (missing)

Looking at the implemented APIs in the README:
- HTTP loader (Libcurl)
- HTML parser (html5ever)
- DOM tree
- Javascript support (v8)
- DOM APIs
- Ajax (XHR, Fetch)
- CDP/websockets server
- Click, Input form
- Cookies, Custom HTTP headers
- Proxy support, Network interception
- robots.txt

The `console` API is not listed, confirming it is not implemented.

The `console` object is a fundamental part of the Web API that JavaScript code running in pages expects to exist. When pages call `console.log()`, `console.error()`, etc., these calls need to be handled by the browser engine (v8 in Lightpanda's case) and potentially forwarded via CDP events.

### Steps to Reproduce

```javascript
// Using Puppeteer with Lightpanda CDP
const browser = await puppeteer.connect({
  browserWSEndpoint: 'ws://127.0.0.1:9222',
});

const page = await browser.newPage();

// This listener never fires in Lightpanda
page.on('console', msg => {
  console.log('PAGE LOG:', msg.text());
});

// This JavaScript executes but console calls are silent
await page.evaluate(() => {
  console.log('Hello from page');
  console.error('This is an error');
});
```

Expected: The console listener receives the messages.
Actual: No console events are emitted.

### Expected Behavior
- `console.log`, `console.warn`, `console.error`, `console.info` should forward messages via CDP `Runtime.consoleAPICalled` event
- `console.debug` should also be forwarded
- `console.exception` (deprecated alias for `console.error`) should work
- `console.group`, `console.groupEnd`, `console.groupCollapsed` should track console message groups

### Actual Behavior
Console calls from page JavaScript are silently ignored, breaking automation scripts.

### Environment
- OS: Ubuntu 22.04
[...]

PR #2339, opened 2026-04-30
cdp: implement Console

krichprollsch:
Relates with #1953

* [x] add `Console` CDP domain
* [x] support `Console.messageAdded` event
* [x] implement the `Console.messageAdded` event for all `Console` API (and compare behavior w/ Chrome)
* [ ] ~~implement `Log` CDP domain for `Console` API (Puppeteer uses `Log` instead of `Console`)~~
* [x] implement `Runtime.consoleAPICalled` events (when `Runtime.enable` is `on`) (Puppeteer uses this)
* [x] Add e2e tests https://github.com/lightpanda-io/demo/pull/170

### DIFF
--- a/src/cdp/CDP.zig
+++ b/src/cdp/CDP.zig
@@ -641,6 +641,14 @@ pub const BrowserContext = struct {
         self.notification.unregister(.console_message, self);
     }
 
+    pub fn runtimeEnable(self: *BrowserContext) !void {
+        try self.notification.register(.runtime_console_message, self, onRuntimeConsoleMessage);
+    }
+
+    pub fn runtimeDisable(self: *BrowserContext) void {
+        self.notification.unregister(.runtime_console_message, self);
+    }
+
     pub fn onFrameRemove(ctx: *anyopaque, _: Notification.FrameRemove) !void {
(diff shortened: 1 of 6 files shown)


src/cdp/CDP.zig
```diff
@@ -913,7 +847,21 @@ pub const BrowserContext = struct {
...
 
-        try cdp.ws.sendJSONRaw(buf);
+        try cdp.client.sendJSONRaw(buf);
+    }
+
+    pub fn onConsoleMessage(ctx: *anyopaque, msg: *const Notification.ConsoleMessage) !void {
+        const self: *BrowserContext = @ptrCast(@alignCast(ctx));
+        defer self.resetNotificationArena();
+
+        return @import("domains/console.zig").consoleMessage(self, msg);
```
karlseguin: The pattern in the other callsites is that we pass the arena as the first parameter, e.g.:

```zig
return @import("domains/console.zig").consoleMessage(self.notification_arena, self, msg);
```

So `consoleMessage` just gets an `arena: Allocator` as their first parameter.

karlseguin (approved):

krichprollsch:
Thanks for the review and the fix :pray:

PR #2366, opened 2026-05-05
Protect DOM mutations against custom element callbacks.

karlseguin:
Same issue as https://github.com/lightpanda-io/browser/pull/2313

DOM mutations have certain expectations that custom element callbacks can violate.

### DIFF
--- a/src/browser/tests/node/append_child.html
+++ b/src/browser/tests/node/append_child.html
@@ -65,3 +65,32 @@
   assertChildren(['a'], d3);
   testing.expectEqual(null, b.parentNode);
 </script>
+
+<div id=d4></div>
+<div id=d4_stash></div>
+<script id=appendChild_disconnect_callback_reparents>
+  // Moving a connected child into a disconnected target makes
+  // will_be_reconnected=false, so disconnectedCallback fires synchronously
+  // inside removeNode. The callback re-parents the child; appendChild
+  // respects that placement instead of overriding it.
+  const d4 = $('#d4');
+  const stash = $('#d4_stash');
+
+  class ReparentOnDisconnect extends HTMLElement {
+    disconnectedCallback() {
+      if (this.parentNode === null) {
+        stash.appendChild(this);
+      }
+    }
+  }
+  customElements.define('reparent-on-disconnect', ReparentOnDisconnect);
+
+  const rpd = document.createElement('reparent-on-disconnect');
+  rpd.id = 'rpd';
+  d4.appendChild(rpd);
+
+  const detached = document.createElement('div');
+  detached.appendChild(rpd);
+
+  testing.expectEqual(stash, rpd.parentNode);
+</script>
--- a/src/browser/tests/node/insert_before.html
+++ b/src/browser/tests/node/insert_before.html
@@ -39,3 +39,34 @@
   assertChildren([], d1);
   assertChildren([c1, c2], d2);
 </script>
+
+<div id=d3></div>
+<div id=d3_stash></div>
+<script id=insertBefore_disconnect_callback_reparents>
+  // Same disconnectedCallback re-parenting pattern as in append_child.html,
+  // exercised through insertBefore. insertBefore respects the callback's
+  // placement instead of overriding it.
+  const d3 = $('#d3');
+  const stash = $('#d3_stash');
+
+  class IBReparentOnDisconnect extends HTMLElement {
+    disconnectedCallback() {
+      if (this.parentNode === null) {
+        stash.appendChild(this);
+      }
+    }
+  }
+  customElements.define('ib-reparent-on-disconnect', IBReparentOnDisconnect);
+
+  const moving = document.createElement('ib-reparent-on-disconnect');
+  moving.id = 'ib_moving';
+  d3.appendChild(moving);
+
+  const detached = document.createElement('div');
+  const ref = document.createElement('span');
+  detached.appendChild(ref);
+
+  detached.insertBefore(moving, ref);
+
+  testing.expectEqual(stash, moving.parentNode);
+</script>
--- a/src/browser/webapi/Node.zig
+++ b/src/browser/webapi/Node.zig
@@ -253,6 +253,11 @@ pub fn appendChild(self: *Node, child: *Node, frame: *Frame) !*Node {
         try frame.adoptNodeTree(child, child_owner.?, parent_owner);
     }
 
+    // A custom element callback can re-parent the node. If it does, we're done
+    if (child._parent != null) {
+        return child;
+    }
+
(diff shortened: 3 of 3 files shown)


karlseguin:
@krichprollsch I think I figured out the root of this issue. Currently, we trigger custom element callbacks in the middle of a DOM mutation. So if a node is moved from one document to another, we'll do something like:

1. remove the node from documentA
2. trigger disconnectedCallback
3. move the node to documentB
4. trigger connectedCallback

But in reality, it should be:
1. remove the node from documentA
2. move the node to documentB    
3. trigger disconnectedCallback
4. trigger connectedCallback

(They are still fired synchronously, but in a different order). Not only is this the "correct" order, but it eliminates all of these footguns. In step 1 and 2, no user-code (no callback) has executed, so the state of the DOM is known.  For example, in our current (incorrect) order, in step 3, we can't be sure the node has no parent  - maybe step 2 attached it. But in the correct order, we can be 100% sure the node has no parent (because we just removed it).

I'm not sure how easy the fix is (I think, not that much, since this can go on recursively - callbacks doing mutations, triggering more callbacks). And I'm not sure how important it is.

krichprollsch:
I like the idea, I agree having the confidence nodes won't be changed by callback in a middle of a modification is easy. 
I guess that means to stack the events until the changes are done, then treating the dispatching stack?
But as you mentioned, what about callback re-changing the DOM? should we dispatch the new events once the change is done by the given CBK, or should we wait until all CBK are treated? it's easier to treat them during the CBK isn't it?

navidemad:
Reviewed the patch. The two early-returns are correct for the cases they cover, and the tests do exercise the right path (`disconnectedCallback` firing inside `removeNode` while the outer `appendChild` / `insertBefore` is mid-flight).

A few observations though, mostly about sibling call sites in `Node.zig` that have the same reentrancy shape and aren't covered here. (Line links below are pinned to commit `5112adc6`.)

### 1. `replaceChildren` — same crash class, not patched

[`Node.zig#L1083-L1090`](https://github.com/lightpanda-io/browser/blob/5112adc69fcbe4b47574c9d6295659eec07ed5ae/src/browser/webapi/Node.zig#L1083-L1090):

```zig
const parent_is_connected = self.isConnected();
for (children_to_add.items) |child| {
    var child_connected = false;
    if (child._parent) |previous_parent| {
        child_connected = child.isConnected();
        frame.removeNode(previous_parent, child, .{ .will_be_reconnected = parent_is_connected });
    }
    try frame.appendNode(self, child, .{ .child_already_connected = child_connected });
}
```

A `disconnectedCallback` fired by `removeNode` can re-parent `child`. The follow-up `appendNode` will then trip the precondition in [`Frame._insertNodeRelative`](https://github.com/lightpanda-io/browser/blob/5112adc69fcbe4b47574c9d6295659eec07ed5ae/src/browser/Frame.zig#L3037-L3041):

```zig
lp.assert(child._parent == null, "Frame.insertNodeRelative parent", .{});
```

Same failure mode this PR is protecting against, reached through a different public API.

### 2. `replaceChildren`'s removal loop uses a stale-prone iterator

[`Node.zig#L1075-L1079`](https://github.com/lightpanda-io/browser/blob/5112adc69fcbe4b47574c9d6295659eec07ed5ae/src/browser/webapi/Node.zig#L1075-L1079):

```zig
// Remove all existing children
var it = self.childrenIterator();
while (it.next()) |child| {
    frame.removeNode(self, child, .{ .will_be_reconnected = false });
}
```

[`NodeIterator.next`](https://github.com/lightpanda-io/browser/blob/5112adc69fcbe4b47574c9d6295659eec07ed5ae/src/browser/webapi/Node.zig#L1101-L1107) caches `node._child_link.next` *before* the callback runs, so a `disconnectedCallback` that detaches or reorders later siblings will stale the iterator.

Compare with the `firstChild()` drain pattern already used elsewhere — [`Frame.appendAllChildren`](https://github.com/lightpanda-io/browser/blob/5112adc69fcbe4b47574c9d6295659eec07ed5ae/src/browser/Frame.zig#L2982-L2990), [`Frame.insertAllChildrenBefore`](https://github.com/lightpanda-io/browser/blob/5112adc69fcbe4b47574c9d6295659eec07ed5ae/src/browser/Frame.zig#L2997-L3005), and [`Element.setInnerHTML`](https://github.com/lightpanda-io/browser/blob/5112adc69fcbe4b47574c9d6295659eec07ed5ae/src/browser/webapi/Element.zig#L492-L499). `replaceChildren` should switch to the same pattern.

### 3. `replaceChild` inherits the new early-return awkwardly

[`Node.zig#L642-L660`](https://github.com/lightpanda-io/browser/blob/5112adc69fcbe4b47574c9d6295659eec07ed5ae/src/browser/webapi/Node.zig#L642-L660):

```zig
pub fn replaceChild(self: *Node, new_child: *Node, old_child: *Node, frame: *Frame) !*Node {
    ...
    _ = try self.insertBefore(new_child, old_child, frame);

    // Special case: if we replace a node by itself, we don't remove it.
    // insertBefore is an noop in this case.
    // Re-check parent after insertBefore since callbacks (e.g. connectedCallback)
    // could have already removed old_child from self.
    if (new_child != old_child and old_child._parent == self) {
        frame.removeNode(self, old_child, .{ .will_be_reconnected = false });
    }
    return old_child;
}
```

If `insertBefore` early-returns because the callback re-parented `new_child` outside `self`, the second branch still removes `old_child`. End state: `old_child` is gone, `new_child` is wherever the callback wanted — semantically a removal, not a replacement. The existing comment only covers the `old_child`-was-removed-by-callback half. Worth either tightening the post-condition or explicitly documenting that callback-driven detours win.

### On the bigger refactor floated [above](https://github.com/lightpanda-io/browser/pull/2366#issuecomment-4378800251)

Queueing all callbacks until the outer mutation finishes is the right long-term invariant — these defensive patches are basically per-site rediscoveries of the same problem. But the early-returns are still worth landing now: they prevent immediate parent-link corruption at the highest-traffic public APIs without waiting for the queue refactor. I'd just suggest extending them to `replaceChildren` (and fixing the iterator there) in this same PR so the protection set is coherent — otherwise the description slightly oversells where the protection ends.

Tests pass locally — `TEST_FILTER="WebApi: Integration" make test` (which exercises both new `append_child.html` and `insert_before.html` cases) gives 515/515.

krichprollsch (approved):

karlseguin:
@navidemad Yes, I (or rather claude) also found a number of other cases where this could happen. But after reviewing everything, and the ugliness of adding these guards everywhere, I opted to address only the issues we were actually seeing in the wild. A bit short-sighted, I know, but I am hoping to address the core issue sooner rather than later

PR #2376, opened 2026-05-06
Fix /json/version advertising ws://host:0/ when --port 0

webcoyote:
## Summary
Builds on `Display actual port when binding --port 0` PR (#2333), which exposed the OS-assigned port via `getsockname` but didn't propagate it into the CDP handshake response.

## Test plan
- [x] `zig build check` passes
- [x] `zig build test` — 518/518
- [x] Manual: `lightpanda serve --port 0 --log-level info`, then `curl /json/version`
shows `ws://127.0.0.1:<bound port>/`

### DIFF
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -44,23 +44,22 @@ conns_mutex: std.Thread.Mutex = .{},
 conns_pool: std.heap.MemoryPool(CDP),
 
 pub fn init(app: *App, address: net.Address) !*Server {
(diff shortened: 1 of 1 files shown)


webcoyote:
From my analysis (and I'm not an expert on lightpanda) these failed tests are unrelated to my changes, but are an artifact of 100ms timeouts on `waitForFunction` in `playwright/cdp.js`.

PR #2387, opened 2026-05-07
Fix zig fmt step in CI

mookums:
This fixes the `zig fmt` step in the CI ensuring that it fails when formatting is broken.


mookums:
This just ensures that the failed formatting checks actually fail the CI instead of passing like before. This is evident [here](https://github.com/lightpanda-io/browser/actions/runs/25491799779/job/74801423237) where it failed but passed in the zig fmt step.

krichprollsch (approved):

PR #2396, opened 2026-05-08
Add AGENTS.md and CLAUDE.md

navidemad:
Adds a minimal `AGENTS.md` covering the project-specific bits that aren't already in the README or derivable from reading the code:

- Test filter / env var flags (`F=`, `TEST_FILTER`, `TEST_VERBOSE`, `TEST_FAIL_FIRST`, `METRICS`).
- The custom test runner's leak-detection invariant — a test that allocates without freeing fails outright.
- The exact `zig fmt --check` command CI runs.
- A short pointer at the existing `@import` alias-case and struct-init-inference patterns to mirror.

`CLAUDE.md` is a one-line `@AGENTS.md` import so Claude Code (which doesn't read AGENTS.md natively) picks up the same content without any duplication.

Deliberately out of scope:
- No `docs/` directory or architecture map — drifts easily, and humans/agents can read the code.
- No `.gitignore` changes for agent tooling — kept locally.
- No tool-specific rule files (`.cursorrules`, `.codex/...`) — AGENTS.md is the single source of truth.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,29 @@
+# AGENTS.md
+
+## Tests
+
+```bash
+make test                                       # Run all tests
+make test F="server"                            # Filter by substring
+TEST_FILTER="WebApi: #selector_all" make test   # Filter main + subtest (separator: #)
+TEST_VERBOSE=true make test
+TEST_FAIL_FIRST=true make test
+METRICS=true make test                          # Capture allocation/duration metrics as JSON
+```
+
+The custom test runner (`src/test_runner.zig`) detects memory leaks in debug builds. **A test that allocates without freeing fails** — not just lints.
+
+## Formatting
+
+```bash
+zig fmt --check ./*.zig ./**/*.zig    # Exact command CI runs
+```
+
+`zig build` depends on the fmt step, so a local build catches drift too.
+
+## Conventions
+
+Mirror the patterns in neighboring files. In particular:
+
+- `@import` alias case follows the imported file's basename (`const Frame = @import("Frame.zig")`, `const ast = @import("ast.zig")`).
+- Prefer struct-init type inference (`.{ ... }`) where the expected type is known from the function signature or variable annotation.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,5 +1,7 @@
 # AGENTS.md
 
+See [CONTRIBUTING.md](CONTRIBUTING.md) for how to open a pull request (CLA, dev setup, pre-PR checks).
+
 ## Tests
 
 ```bash
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,1 @@
+@AGENTS.md
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,6 +2,21 @@
 
 Lightpanda accepts pull requests through GitHub.
 
+## Development
+
+- Run the tests: `make test`
+- Check formatting: `zig fmt --check ./*.zig ./**/*.zig`
+
+See [AGENTS.md](AGENTS.md) for the full set of test, formatting, and code conventions (test filters, the leak-detection invariant, `@import` alias case, struct-init inference).
+
+## Before opening a PR
+
+- [ ] Tests pass (`make test`).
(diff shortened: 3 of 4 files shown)


karlseguin:
I'm ok with this, @krichprollsch ?

krichprollsch:
Hello @navidemad thanks for the PR.

Do you use these files currently for your contributions?

I feel like this content should be put directly into CONTRIBUTING.md file which is pointed from the README. 
But idk if agents will benefit of it. 

WDYT?

navidemad:
> Hello @navidemad thanks for the PR.
> 
> Do you use these files currently for your contributions?
> 
> I feel like this content should be put directly into CONTRIBUTING.md file which is pointed from the README. But idk if agents will benefit of it.
> 
> WDYT?

Not anymore. My Claude Code has auto-memory turned on, so it has learned not to repeat the same mistakes.

I’m also using GitNexus locally, which probably helps as well.

On top of that, I have a global CLAUDE.md that includes this: https://github.com/forrestchang/andrej-karpathy-skills but it's more a personal preference.

That said, I think adding them to CONTRIBUTING.md is a good idea.

krichprollsch:
Thanks for sharing your tools :pray: 
So if you want to move the changes into `CONTRIBUTING.md`, I would be pleased to merge it :)

navidemad:
Pushed 827e696a after considering whether to fold this file into `CONTRIBUTING.md` instead.

### What I considered

**Option A — keep three files separate** (current state): `AGENTS.md` is the cross-vendor agent doc (auto-loaded by Codex/Cursor/etc.), `CLAUDE.md` is a one-line `@AGENTS.md` shim, `CONTRIBUTING.md` is human-facing.

**Option B — fold operational content into `CONTRIBUTING.md`**, leaving `AGENTS.md`/`CLAUDE.md` as pointers. Tempting because `CONTRIBUTING.md` is GitHub's canonical surface and the single-source-of-truth feel is appealing.

### Why I landed on A

Cross-referenced GitNexus, which has matured this exact problem (`CONTRIBUTING.md` ~210 lines, `AGENTS.md` ~215 lines, `CLAUDE.md` ~55 lines as a delta). They keep the three separate for two reasons that apply here too:

1. **Audience mismatch.** `CONTRIBUTING.md` is "how to legally and procedurally submit a PR" (CLA, PR conventions). `AGENTS.md` is "how to operate the repo as an agent" (test filters, leak-detection invariant, `@import` alias case). Mixing them dilutes both.
2. **Auto-discovery.** `agents.md` is the convention non-Claude agents (Codex, Cursor) read automatically from the repo root. A pointer from `AGENTS.md` → `CONTRIBUTING.md` doesn't trigger that auto-load.

### What this PR does instead

Three small additive edits — no restructuring:

- `AGENTS.md`: one-line pointer to `CONTRIBUTING.md` at the top.
- `CONTRIBUTING.md`: new **Development** section (test + fmt commands) with a pointer back to `AGENTS.md` for full conventions.
- `CONTRIBUTING.md`: new **Before opening a PR** checklist (tests, fmt, CLA).
- CLA paragraph preserved verbatim, moved into its own section.

Net diff is +17 lines, additive only.

### What I deliberately did **not** do

- **No `TESTING.md`.** The Tests section in `AGENTS.md` is 9 lines — splitting it now is file proliferation. Trigger to revisit: when we start documenting *how to write* tests (allocator patterns, fuzz conventions, `METRICS=true` schema, test-runner internals), that's the moment to extract.
- **No version headers / changelog tables** on `AGENTS.md` / `CLAUDE.md`. GitNexus does this; their files are ~200 lines each and earn it. At 31 lines, overhead doesn't pay rent. Revisit if the file passes ~100 lines.
- **No `GUARDRAILS.md`** as a separate hard-rules doc. The existing `.claude/rules/*.md` already covers that role for Claude.
- **No scope table** (read/write/execute/off-limits). Useful for multi-package monorepos; this is one Zig project.

krichprollsch (approved):

krichprollsch:
> A pointer from AGENTS.md → CONTRIBUTING.md doesn't trigger that auto-load

So that work for `CLAUDE.md` but not for `AGENTS.md`? what a shame :/

krichprollsch:
Thanks for the PR :pray:

Issue #2397, opened 2026-05-08
High memory usage when fetching some webpages


blieusong:
Hi 

When I fetch the following page

> https://www.apple.com/us/shop/buy-iphone/iphone-16/

The memory usage seems to spike at around 3+Gb

But fetching the equivalent page in, say, UK or Japan only requires about 300Mb 

> https://www.apple.com/uk/shop/buy-iphone/iphone-16/

The following fails with exit code 137:

```
docker run --rm -m 1g lightpanda/browser:latest lightpanda fetch --dump html --obey-robots https://www.apple.com/us/shop/buy-iphone/iphone-16/
```

It's only at 4Gb allocated memory that it works:

```
docker run --rm -m 4g lightpanda/browser:latest lightpanda fetch --dump html --obey-robots https://www.apple.com/us/shop/buy-iphone/iphone-16/
```

All the US iPhone pages seem to fail. It works on the Mac pages I tried it with. Eg.

```
docker run --rm -m 1g lightpanda/browser:latest lightpanda fetch --dump html --obey-robots https://www.apple.com/us/shop/buy-mac/macbook-neo/
```

The dumped HTML for https://www.apple.com/us/shop/buy-iphone/iphone-16/ is bigger than https://www.apple.com/uk/shop/buy-iphone/iphone-16/ (seems to contains trade in data and others that UK does not have) but it seems strange that this would yield a 10x memory usage.

Could it be an issue in Lighpanda's parsing / JS engine? Or is it just that some webpage require much more memory?

Thank you.

Benh

navidemad:
Reproduced and root-caused. Posting findings; fix is up at #2406.

## Reproduced

Local build, peak RSS via `/usr/bin/time -l`:

| Page | Output HTML | Peak RSS |
|---|---|---|
| `apple.com/uk/.../iphone-16/` | 692 KB | 394 MB |
| `apple.com/us/.../iphone-16/` | 1.29 MB | **3.50 GB** |

~9× memory for ~2× HTML — matches the report exactly.

## Root cause

`Frame.appendNew`'s `.text` branch (`src/browser/Frame.zig:1799-1822`) merged each incoming text chunk into the existing text node by re-allocating on the page-lifetime arena:

```zig
cdata._data = try String.concat(self.arena, &.{ existing, txt });
```

[...]

PR #2406, opened 2026-05-09
parser: defer raw-text merge to bound memory growth

navidemad:
## What this fixes

Issue #2397: `lightpanda fetch --dump html` on apple.com US iPhone pages peaked at ~3.5 GB RSS (vs ~390 MB on the equivalent UK page). The 1 GB Docker container OOMs (`exit 137`). Reproducible with:

```sh
docker run --rm -m 1g lightpanda/browser:latest \
  lightpanda fetch --dump html --obey-robots \
  https://www.apple.com/us/shop/buy-iphone/iphone-16/
```

## Root cause

`Frame.appendNew`'s `.text` branch in `src/browser/Frame.zig:1799-1822` did:

```zig
cdata._data = try String.concat(self.arena, &.{ existing, txt });
```

every time html5ever delivered a chunk via the `AppendText` callback. The page-lifetime arena (`frame.arena`) doesn't free individual allocations, so each merge left the prior buffer as dead memory. Total bytes written ≈ `N²/(2·chunk_size)`.

The trigger is html5ever's tokenizer: in `script-data` (and `rawtext` / `rcdata`) state, encountering `<` flushes the pending character buffer and re-enters via a fresh `AppendText`. The US iPhone page has a 347 KB inline `window.PRODUCT_SELECTION_BOOTSTRAP = {…}` literal stuffed with embedded HTML strings like `"header":"<span>Model</span>"` — every `<` becomes a chunk boundary.

Empirically validated with synthetic 100 KB scripts on `main`:

| Script body | Peak RSS |
|---|---|
| 100 KB of `'a'` (no `<`) | 40 MB |
| 100 KB with `<` every 100 chars | 190 MB |
| 100 KB with `<` every 10 chars | **1.5 GB** |

PR #2241 capped per-tick growth and added GC hints in the wait loop. That fix is post-parse; this pathology happens during initial HTML parse — RSS hits 3.2 GB by `--wait-ms 1000`, before the wait loop runs.

## What this PR changes

Move the merge into the parser. Same-parent text chunks accumulate in a `std.ArrayListUnmanaged(u8)` on the per-parse arena; one `String.dupe` commits the final value to the frame arena.

```zig
// src/browser/parser/Parser.zig
const PendingText = struct {
    parent: *Node,
    text_node: *CData,
    buf: std.ArrayListUnmanaged(u8) = .empty,
};
pending_text: ?PendingText,
```

Flush points are the natural ends of a text run:
- A non-text child appended adjacent (`_appendCallback` `.node` branch)
- A foster / before-sibling insertion (`_appendBeforeSiblingCallback`)
- Parent element popped (`_popCallback`, **unconditional** — covers custom-element constructors that may read sibling `textContent` mid-parse)
- Parse call returning (`parse`, `parseWithEncoding`, `parseXML`, `parseFragment`, `Streaming.done`)

`Frame.appendNew` is simplified to take `*Node` directly — it had no non-parser callers (verified by grep + GitNexus impact). `Streaming.done` returns `!void` to propagate the final flush; the only call site (`Document.close`) was already in `!void` context.

Total peak allocation per text run is now ~2N (one ArrayList capacity + one final String.dupe) instead of ~N²/(2·c).

## Results

| Reproducer | Before | After |
|---|---|---|
| apple.com US iPhone fetch | **3.50 GB** | 125 MB |
| Standalone bootstrap script (347 KB) | 2.54 GB | 45 MB |
| 100 KB synthetic, `<` every 10 chars | 1.50 GB | 41 MB |
| apple.com UK iPhone (already fine) | 394 MB | 106 MB |

Dumped HTML on the US page is **byte-identical** to pre-patch (1,285,747 B). UK has only React `useId` differences across runs (same byte size) — pre-existing non-determinism.

## Where to focus review

- `Parser.zig` `flushPendingText` / `appendTextChunk` (new) — the invariant is "`pending_text.text_node === pending_text.parent.lastChild()` while pending is non-null"; flush is called any time that invariant could be violated.
- `Parser.zig` `_popCallback` flushes **unconditionally**, not only when `pt.parent == node`. A custom-element constructor reaching back to a sibling's `textContent` during pop must observe finalized data.
- `Frame.appendNew` signature change to `(parent: *Node, child: *Node)` — the only caller change is `Parser._appendDoctypeToDocument` which now passes the node directly.
- Allocator lifetime: pending buffer lives on `parser.arena` (per-parse); committed `String` lives on `frame.arena` (page-lifetime), matching the prior `String.concat(self.arena, …)` site exactly. Verified for `Frame.parseBlob`, `Frame.parse`, `parseHtmlAsChildren` (innerHTML), and `Document.write`'s foster path.

## Test plan

- [x] New regression test `src/browser/tests/cdata/raw_text_chunked.html` asserts `<script>`/`<style>`/`<textarea>`/`<title>` with embedded `<` parse to a single text node with verbatim content.
- [x] Full suite: `mise exec -- zig build test $V8` — 521/521 pass.
- [x] Live: `lightpanda fetch --dump html --wait-ms 8000` on apple.com US iPhone page — 125 MB peak RSS (was 3.5 GB), output byte-identical.
- [x] `gitnexus_detect_changes` — 1 affected process (`insertAdjacentHTML → parseFragment`), covered by the new flush at the end of `parseFragment`.

Refs #2397

### DIFF
--- a/src/browser/Frame.zig
+++ b/src/browser/Frame.zig
@@ -1813,26 +1813,12 @@ pub fn notifyNetworkAlmostIdle(self: *Frame) void {
     });
 }
 
-// called from the parser
-pub fn appendNew(self: *Frame, parent: *Node, child: Node.NodeOrText) !void {
-    const node = switch (child) {
-        .node => |n| n,
-        .text => |txt| blk: {
-            // If we're appending this adjacently to a text node, we should merge
-            if (parent.lastChild()) |sibling| {
-                if (sibling.is(CData.Text)) |tn| {
-                    const cdata = tn._proto;
-                    const existing = cdata.getData().str();
-                    cdata._data = try String.concat(self.arena, &.{ existing, txt });
-                    return;
-                }
-            }
-            break :blk try self.createTextNode(txt);
-        },
-    };
-
-    lp.assert(node._parent == null, "Frame.appendNew", .{});
-    try self._insertNodeRelative(true, parent, node, .append, .{
+// called from the parser. Text-node merging is the parser's responsibility
+// (see Parser.appendTextChunk in src/browser/parser/Parser.zig); this is the
+// "insert this fully-formed node as a new last child of parent" entry point.
+pub fn appendNew(self: *Frame, parent: *Node, child: *Node) !void {
+    lp.assert(child._parent == null, "Frame.appendNew", .{});
+    try self._insertNodeRelative(true, parent, child, .append, .{
         // this opts has no meaning since we're passing `true` as the first
         // parameter, which indicates this comes from the parser, and has its
         // own special processing. Still, set it to be clear.
--- a/src/browser/parser/Parser.zig
+++ b/src/browser/parser/Parser.zig
@@ -45,12 +45,11 @@ pub const ParsedNode = struct {
 // AppendText callback per chunk. Merging via String.concat in the previous
 // implementation was O(N^2/chunk_size) on the page-lifetime arena, blowing
 // memory on inline JS that contains embedded HTML strings (issue #2397).
-// Instead, we accumulate same-parent chunks in this struct and commit a
-// single allocation on flush.
+// Instead, we keep a single Parser-level buf and accumulate same-parent
+// chunks into it, committing once on flush.
 const PendingText = struct {
     parent: *Node,
     text_node: *CData,
-    buf: std.ArrayListUnmanaged(u8) = .empty,
 };
 
 const Parser = @This();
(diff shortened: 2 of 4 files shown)


karlseguin:
This is great. I wonder if

1 - We can lift the buffer into the Parser and then re-use it, e.g. replacing `defer pt.buf.deinit(self.arena);` with `defer self.buf.clearRetainingCapacity();`. 

2 - Can we lazily buffer? in `main`, we currently do 1 copy.  In this PR, we do 3 copies, even when it isn't needed. Don't get me wrong, overall, I still think it's an improvement. But I think if you lazily build up the buffer, in the normal case, you'll go back to the 1 copy. 

3 - I was wondering if there are ever cases where pending texts aren't flushed? I'm not sure, but should `flushPendingText` be called at the end of a successful Streaming.read?

navidemad:
Thanks for the review.

**1 + 2 — agreed on both.** With `parser.arena == frame.arena` in the streaming case, `pt.buf.deinit` is a no-op (pooled arena, individual frees don't reclaim), so each text run leaves its buf bytes on the page-lifetime arena. Lifting `buf` onto the Parser and clearing via `clearRetainingCapacity` bounds total dead memory to one peak-run-sized allocation. And you're right that the common single-chunk run regressed from 1 copy to 3 (`createTextNode` → seed buf → flush dupe). Leaving the first chunk in `_data` only, then seeding `buf` from `_data.str()` when a second chunk arrives, returns the common case to 1 copy. Pushing shortly.

**3 — would lean against.** Audited the flush points: `_popCallback`, `_appendCallback` (.node branch), `_appendBeforeSiblingCallback`, `_removeFromParentCallback`, `_reparentChildrenCallback` all flush; `Streaming.done()` flushes; the panic path in `Document.write` explicitly flushes after dropping the handle. The cost of an unconditional flush per `read()` is real: if a text run spans two `document.write()` calls, the second `read()` enters `appendTextChunk` with `pending_text == null` and `lastChild` a text node, taking the recovery branch that re-buffers the just-committed `_data` for every continuing chunk — quadratic again. Happy to revisit if you can point at a specific path I missed.

PR #2408, opened 2026-05-10
fix: require fetch URL before runtime

guyua9:
## Summary
- Treat a missing positional argument as `error.MissingArgument` even though the parser stores positional fields as optional during initialization.
- Prevent `lightpanda fetch` from reaching runtime with a null URL and crashing on the later force unwrap.

## Tests
- `git diff --check upstream/main...HEAD`
- Not run: `zig` is not installed in this environment.

### DIFF
--- a/src/cli.zig
+++ b/src/cli.zig
@@ -641,10 +641,12 @@ pub fn Builder(comptime commands: anytype) type {
                 }
             }
 
-            // A non-optional positional that is still null after parsing is missing.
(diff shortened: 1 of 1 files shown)


guyua9:
Closing this because, after re-evaluating it from the maintainer/reviewer perspective, the current change is too small and under-evidenced to justify a standalone PR: it has no linked issue and no regression test proving the missing-URL behavior. Sorry for the noise.

PR #2417, opened 2026-05-11
`Performance`: add dummy `setResourceTimingBufferSize`

nikneym:


src/browser/webapi/Performance.zig
```diff
@@ -281,6 +286,7 @@ pub const JsApi = struct {
     pub const measure = bridge.function(Performance.measure, .{ .dom_exception = true });
     pub const clearMarks = bridge.function(Performance.clearMarks, .{});
     pub const clearMeasures = bridge.function(Performance.clearMeasures, .{});
+    pub const setResourceTimingBufferSize = bridge.function(Performance.setResourceTimingBufferSize, .{});
```
karlseguin:
```suggestion
    pub const setResourceTimingBufferSize = bridge.function(Performance.setResourceTimingBufferSize, .{.noop = true});
```

Saves a bit of memory and faster when called..it won't even bother calling `Performance.setResourceTimingBufferSize`.

PR #2420, opened 2026-05-11
HttpClient Improvements

karlseguin:
1 - Track owner of a request (for simpler / more accurate abort (TBD))

2 - Create Transfer upfront, make everything work on Transfer (not Request). This helps remove ambiguity about cleanup and simplifies layers. For example Robots request is just another normal request, not a special case. Also gives everything a stable address (the `*Transfer` which can be looked up by id)

3 - Track the transfers linked to a Frame (or WorkerGlobalState). This significantly simplifies the abortFrame logic (now abortList) and removes ambiguity when 2 frames share the same frame_id

4 - Queue page destruction for the next Runner tick, avoiding frames from being freed during a callback [which relies on the frame]

### DIFF
--- a/src/Notification.zig
+++ b/src/Notification.zig
@@ -21,9 +21,7 @@ const lp = @import("lightpanda");
 
 const Frame = @import("browser/Frame.zig");
 const Transfer = @import("browser/HttpClient.zig").Transfer;
-const Request = @import("browser/HttpClient.zig").Request;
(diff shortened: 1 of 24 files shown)


krichprollsch (approved):
I like the ownership (and the removal of the ugly protect from abort) + the transfer_id into cdp interceptions.

PR #2424, opened 2026-05-11
Ability to run wpt with Nix/NixOS

mookums:
This allows you to run the WPT tests inside of Nix (using Docker/Podman).

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -83,7 +83,7 @@
             demoRepo = pkgs.fetchFromGitHub {
               owner = "lightpanda-io";
               repo = "demo";
-              rev = "main";
+              rev = "6120b426dd3bb5d0b06512efe735f2e2778caa15";
               sha256 = "sha256-TaLi1uIpM+fUk8LFVnPOZM6rKbdHKmB2FcofvGHtcpM=";
             };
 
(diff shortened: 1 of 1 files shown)


karlseguin:
I can't offer any insight on nix, I can just run claude through it. The only issue it suggested revisiting was:

```
 ---
  🔴 Reproducibility — rev = "main" / rev = "fork" with a fixed hash

  demoRepo = pkgs.fetchFromGitHub { ... rev = "main"; sha256 = "..."; };
  wptRepo  = pkgs.fetchFromGitHub { ... rev = "fork"; sha256 = "..."; };

  This is the Nix antipattern of "pin to a moving branch with a hash." The behavior is:
  - Upstream branch moves → the next nix build fails with a hash mismatch.
  - The author then bumps the hash → flake works again, but at a brand new (unaudited) snapshot.
  - There is no way to reproduce "yesterday's WPT run" without trawling git history of the flake itself.

  Pin to commit SHAs (rev = "<full-sha>") or — better — promote them to inputs in the flake header so flake.lock does the pinning
  automatically. The CI workflow at least pins ref: fork deliberately as "always latest of this branch"; for a flake you want the opposite.
  ```
  But if that isn't really an issue, than feel free to merge.

PR #2429, opened 2026-05-12
Custom element reactions v2

karlseguin:
Re-applies https://github.com/lightpanda-io/browser/pull/2394 with fixes which was reverted in  https://github.com/lightpanda-io/browser/pull/2428

Should not be merged until WPT tests can reliably be run against the /custom-elements/ category.


krichprollsch:
@karlseguin I think this one is mergeable isn't it?
I ran WPT tests locally, I get good results.

PR #2431, opened 2026-05-12
fix(cdp): remove duplicate Page.frameNavigated and fix context regist…

krichprollsch:
…ration for iframes

The frameNavigated CDP handler sent Page.frameNavigated twice per navigation and always used the root frame's V8 context for inspector.contextCreated, even during iframe navigations. This caused the root frame's inspector context id to be silently re-registered (and the previous id invalidated) whenever an iframe navigated.

The duplicate Page.frameNavigated masked this bug — Puppeteer happened to pick up the re-registered context id. Removing the duplicate exposed the underlying issue: callFunctionOn with the original id failed with "Cannot find context with specified id".

Fix the test issue with [magic8ball test](https://github.com/lightpanda-io/demo/pull/171)

### DIFF
--- a/src/cdp/domains/page.zig
+++ b/src/cdp/domains/page.zig
@@ -671,7 +671,7 @@ pub fn frameNavigated(arena: Allocator, bc: *CDP.BrowserContext, event: *const N
             "",
             frame.origin orelse "",
             aux_data,
-            true,
+            is_root_frame,
         );
     }
     // Isolated worlds are session-wide (single V8 context shared across
--- a/src/cdp/domains/page.zig
+++ b/src/cdp/domains/page.zig
@@ -645,7 +645,7 @@ pub fn frameNavigated(arena: Allocator, bc: *CDP.BrowserContext, event: *const N
     // frame's JS context for inspector.contextCreated prevents re-registering
     // the root context under a new id (which silently invalidates the
     // previous id on the V8 side).
-    const frame = bc.session.findFrameByFrameId(event.frame_id) orelse root_frame;
+    const frame = bc.session.findFrameByFrameId(event.frame_id) orelse return error.FrameNotFound;
 
     // frameNavigated event
     try cdp.sendEvent("Page.frameNavigated", .{
--- a/src/cdp/domains/page.zig
+++ b/src/cdp/domains/page.zig
@@ -624,7 +624,8 @@ pub fn frameNavigated(arena: Allocator, bc: *CDP.BrowserContext, event: *const N
         }, .{ .session_id = session_id });
     }
 
-    const frame = bc.session.currentFrame() orelse return error.FrameNotLoaded;
+    const root_frame = bc.session.currentFrame() orelse return error.FrameNotLoaded;
+    const is_root_frame = event.frame_id == root_frame._frame_id;
 
     // When we actually recreated the context we should have the inspector send
     // this event, see: resetContextGroup. Sending this event will tell the
@@ -635,9 +636,16 @@ pub fn frameNavigated(arena: Allocator, bc: *CDP.BrowserContext, event: *const N
     // frames (iframes), clearing all contexts would destroy the main frame's
     // context, causing Puppeteer's frame.evaluate()/frame.content() to hang
     // forever.
-    if (event.frame_id == frame._frame_id) {
+    if (is_root_frame) {
         try cdp.sendEvent("Runtime.executionContextsCleared", null, .{ .session_id = session_id });
     }
+
+    // Look up the actual navigated frame. For main frame navigations this is
+    // the root frame; for iframes it is the child frame. Using the correct
+    // frame's JS context for inspector.contextCreated prevents re-registering
+    // the root context under a new id (which silently invalidates the
+    // previous id on the V8 side).
+    const frame = bc.session.findFrameByFrameId(event.frame_id) orelse root_frame;
 
     // frameNavigated event
     try cdp.sendEvent("Page.frameNavigated", .{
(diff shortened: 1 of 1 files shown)


src/cdp/domains/page.zig
```diff
@@ -635,10 +636,17 @@ pub fn frameNavigated(arena: Allocator, bc: *CDP.BrowserContext, event: *const N
...
+    if (is_root_frame) {
         try cdp.sendEvent("Runtime.executionContextsCleared", null, .{ .session_id = session_id });
     }
 
+    // Look up the actual navigated frame. For main frame navigations this is
+    // the root frame; for iframes it is the child frame. Using the correct
+    // frame's JS context for inspector.contextCreated prevents re-registering
+    // the root context under a new id (which silently invalidates the
+    // previous id on the V8 side).
+    const frame = bc.session.findFrameByFrameId(event.frame_id) orelse root_frame;
```
karlseguin: Should we log this or error on this? 

Else everything else LGTM.

krichprollsch: I agree, it's better to return an error if the frame is missing.

navidemad:
Local validation against `cdp-double-frame-navigated-event` HEAD `806497c0` (build `1.0.0-dev.6157`).

**What the PR cures.** The pre-existing failure on `main` was `Cannot find context with specified id` raised from `Runtime.callFunctionOn` after a sibling-frame `Page.frameNavigated`. With this PR, that error no longer appears. The contextId churn caused by the duplicate `Page.frameNavigated` is genuinely fixed.

**New symptom on the same pages.** Across a batch CDP workload (multiple sequential `BrowserContext` create/dispose cycles, each navigating to a page with 4 top-level `<iframe>` + 2 nested), `document.querySelectorAll("#frameOne")` and `document.evaluate(".//iframe[@id='frameOne']", document, …)` evaluated against the **parent** document via `Runtime.evaluate` consistently return `Array(0)` — even though `Page.frameAttached` for those iframes has already arrived on the wire. Excerpt from one of these queries (timestamps relative, single CDP session):

```
◀ 2.092s  Page.loadEventFired                   (main frame)
▶ 2.093s  Runtime.evaluate document.querySelectorAll("#frameOne")
◀ 2.094s  Page.frameAttached                    (FID-...0006)
◀ 2.096s  Page.frameAttached                    (FID-...0007)
◀ 2.098s  → Array(0)
```

The query is retried ~25× over ~1 second with the same result — so this isn't just an early-query race, the iframe stays invisible to JS queries on the parent for the full retry window.

**Where I'd start looking** (same area as @karlseguin's review nit, but tied to ordering):

1. `page.zig:639` — `frameAttached` for non-main frames. If `frameNavigated` was previously the trigger that flushed the parent-side DOM insertion of the `<iframe>` element, removing the duplicate may have shifted ordering so the parent's JS-visible DOM doesn't contain the child element when `loadEventFired` fires.
2. `page.zig:677-699` (`OnInspectorEvent → executionContextCreated`) — if the parent's execution context is now signalled to CDP *before* the parent has finished walking its document subtree, JS queries against `document` will miss iframe children.
3. `page.zig:648` — order between `executionContextsCleared`, `frameNavigated` for the main frame, and the parent's HTML-tree insertion of child `<iframe>` nodes.

navidemad:
Dug into the symptom from my earlier comment. **Root cause is `Inspector.default_context` getting overwritten on every `inspector.contextCreated` call**, and the PR now triggers it for child-frame navigations.

Chain:

1. The PR changes `page.zig:648` so `frame` is the actually-navigated frame (root for main-frame navigations, child for iframe navigations) — that's the intended fix for issue #2400.
2. `page.zig:669-675` then calls `inspector.contextCreated(frame.js, …, true)` — the 6th arg (`is_default_context`) is **hard-coded to `true`** for both root and child frame navigations.
3. In `Inspector.zig:106-128`, that flag controls a session-wide field:
   ```zig
   if (is_default_context) {
       self.default_context = local.ctx.handle;   // overwrites on every call
   }
   ```
4. `Inspector.zig:431-438` exports `self.default_context` as V8's `ensureDefaultContextInGroup` callback target — i.e. what V8 returns when an inspector op (`Runtime.evaluate` without `contextId`, `Runtime.callFunctionOn` without an objectId-derived context) needs "a context".

So after main + N iframe navigations, `default_context` is whichever iframe most recently fired `contextCreated`. `Runtime.evaluate` without an explicit `contextId` then runs inside that iframe's V8 context — `document` resolves to that iframe's document.

**Reproducer (now deterministic with deeper nesting).** Mirroring a parent → 4 top-level iframes (including 2 that point at a `frame_parent.html`) → each `frame_parent` → `frame_child.html` → 2 grandchildren = 10 iframes total over 3 levels. On `cdp-double-frame-navigated-event` HEAD `806497c0` unpatched, a small Python `websockets` client gets:

```
querySelectorAll('#frameOne')         -> []
document.evaluate(XPath iframe)        -> []
document.getElementsByTagName('iframe') -> 0 elements
document.body.innerHTML preview         -> '<div id="divInFrameTwo">two</div>'
```

`document.body.innerHTML` is one of the grandchildren's body — confirms the evaluate ran inside the last iframe to navigate, not the parent.

**One-line patch.** `is_default_context` should follow the same is-root gating the PR already applies to `executionContextsCleared` and isolated worlds:

```diff
         bc.inspector_session.inspector.contextCreated(
             &ls.local,
             "",
             frame.origin orelse "",
             aux_data,
-            true,
+            is_root_frame,
         );
```

`aux_data`'s `"isDefault":true` stays as-is — per CDP, each frame's main world IS the default for that frame, and that's what the JSON describes. The 6th Zig arg is purely Lightpanda's internal `default_context` bookkeeping, and it should remain bound to the page's main frame so V8's `ensureDefaultContextInGroup` keeps returning the right context. This also matches pre-PR behavior: the old code always passed `currentFrame()` (the root) into `frame.js.localScope`, so `default_context` was effectively always the root's context — the new lookup is correct, but the flag wasn't updated to match.

**Validation against patched build.**

- The reproducer above now returns `qsa=['IFRAME']`, `all_iframes=4`, `body=parent`.
- The 13 frame-spec failures from my earlier comment go to 0 across 4 test seeds (24/24 each).
- Broader CDP-driven shared-spec battery (1399 examples touching navigation, cookies, modals, forms) — 0 failures across 3 different seeds. No regressions from the change.

src/cdp/domains/page.zig
```diff
@@ -666,22 +674,29 @@ pub fn frameNavigated(arena: Allocator, bc: *CDP.BrowserContext, event: *const N
             true,
```
navidemad: Per [my analysis above](https://github.com/lightpanda-io/browser/pull/2431#issuecomment-4440419799), gating this on `is_root_frame` keeps `Inspector.default_context` bound to the main frame (so V8's `ensureDefaultContextInGroup` keeps returning the page's main world for `Runtime.evaluate` without an explicit `contextId`), while still letting child-frame contexts register under their own ids — which is the point of this PR.

```suggestion
            is_root_frame,
```

krichprollsch: Good catch! thanks!

PR #2433, opened 2026-05-12
Implement webMCP API and webMCP cdp domain

krichprollsch:
demo e2e test: https://github.com/lightpanda-io/demo/pull/173

### DIFF
--- a/src/browser/webapi/ModelContext.zig
+++ b/src/browser/webapi/ModelContext.zig
@@ -17,17 +17,6 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 // WebMCP — https://webmachinelearning.github.io/webmcp/
-//
-// Exposes `navigator.modelContext`, the page-side surface for declaring MCP
-// tools to a browser agent. Lightpanda doesn't ship an agent yet; the
-// follow-ups will wire the registered tools through:
-//   1. CDP `WebMCP` domain (https://chromedevtools.github.io/devtools-protocol/tot/WebMCP/)
-//   2. Lightpanda's own MCP server forwarding tools to an external LLM.
-//
-// Both consumers reach into `tools()` / `findTool()` from Zig; the JS-side
-// surface (`registerTool`, `requestUserInteraction`) is the only part shipped
-// today.
-
 const std = @import("std");
 
 const js = @import("../js/js.zig");
@@ -63,9 +52,6 @@ pub const Tool = struct {
     input_schema: ?js.Object.Global,
     execute: js.Function.Global,
     annotations: Annotations,
-    // When present, the tool is considered unregistered once the signal
-    // fires. Checked lazily on each `tools()` / `findTool()` call — fine
-    // for headless usage where there's no synchronous observer to notify.
     signal: ?*AbortSignal,
 
     pub fn markAborted(self: *Tool, exec: *const Execution) !void {
@@ -106,9 +92,7 @@ pub fn registerTool(
         }
     }
 
-    // Reject duplicate names. The spec says `InvalidStateError`. We compact
-    // the list lazily here so a tool whose signal already aborted doesn't
-    // block re-registering under the same name.
+    // Reject duplicate names. The spec says `InvalidStateError`.
     for (self._tools.items) |existing| {
         if (std.mem.eql(u8, existing.name, tool.name)) {
             return error.InvalidStateError;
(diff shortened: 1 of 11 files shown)


src/cdp/domains/webmcp.zig
```diff
@@ -0,0 +1,540 @@
...
+        return cmd.sendResult(.{ .invocationId = &inv_id_str }, .{});
+    };
+
+    const callback = local.toLocal(tool.execute);
+
+    // ModelContextClient has no per-instance state today (see
+    // ModelContext.zig). We still build a fresh wrapper so the page-side
+    // signature `execute(input, client)` works as documented.
+    var client_inst = ModelContextClient{};
+    const client_value = try local.zigValueToJs(&client_inst, .{});
```
karlseguin: You're taking the address of a local, `client_inst`. I'm not really following the code, but presumably the JS callback can do anything with the corresponding value, including having it outlive the callback. You should either put on the heap (`frame._factory` would be the default way, unless there's a scope-appropriate arena), or you could pass `client_value` as-is, and the bridge will put it on the heap for.

Also, I don't think you need to call `zigValueToJs` here, `tryCall` will do it for you.

src/cdp/domains/webmcp.zig
```diff
@@ -0,0 +1,540 @@
...
+            try w.objectField("name");
+            try w.write(t.name);
+
+            try w.objectField("description");
+            try w.write(t.description);
+
+            try w.objectField("inputSchema");
+            if (t.input_schema) |schema_global| {
+                const schema_obj = schema_global.local(self.local);
+                const schema_json = schema_obj.toValue().toJson(self.arena) catch "{}";
```
karlseguin: I think `jsonStringify` was recently added to `js.Value`, can this be used here?

src/cdp/domains/webmcp.zig
```diff
@@ -0,0 +1,540 @@
...
+    };
+    try bc.webmcp_invocations.put(bc.arena, inv_id, invocation);
+
+    // Send toolInvoked event before we run the JS, so the client sees
+    // them in order even if the tool resolves synchronously.
+    const session_id = bc.session_id;
+    try cmd.sendEvent("WebMCP.toolInvoked", .{
+        .toolName = tool.name,
+        .frameId = id.toFrameId(frame_id),
+        .invocationId = &inv_id_str,
```
karlseguin, line 129: Is the order a problem? The event containing this new invocation id is sent before the result to indicate what the id is.

src/browser/webapi/ModelContext.zig
```diff
@@ -0,0 +1,246 @@
...
+    for (self._tools.items) |t| {
+        if (std.mem.eql(u8, t.name, name)) return t;
+    }
+    return null;
+}
+
+/// Walk the tool list and remove any whose `AbortSignal` has fired,
+/// dispatching `model_context_tool_removed` for each. Cheap when no
+/// signals fired (which is the common case).
+fn compactAborted(self: *ModelContext, frame: *Frame) void {
```
karlseguin: So this isn't run when the signal is aborted, it's run lazily (and maybe never). I'm fine with that, if you are, but when I added AbortSignal.any I had the same issue, and I ended up keeping a list of dependent signals. When a signal aborts, it notifies the dependent signal.

AbortSignal._dependent could become a union and the `ModelContext` could become a dependent. When AbortSignal is aborted, we iterate through the dependents (which we already do) and abort the dependent. But I don't know how important the `model_context_tool_removed` notification. So maybe not worth it.

src/cdp/domains/webmcp.zig
krichprollsch: I updated the e2e test to use [Puppeteer's builtin webmcp](https://pptr.dev/guides/webmcp) API instead of raw CDP calls.
It seems ok with it.

The minimal Chrome version (1.49) didn't land yet for me. So It's hard to compare.
Let's keep as it for now.

src/browser/webapi/ModelContext.zig
krichprollsch: I'm not sure how important it is. I could imagine the client wants to update the list of available tools, avoiding LLM to invoke a removed tool.
So I suppose it's important.

So thanks for the insights, I tried to implement that on f00c0ab276bb80a9bb4f783010f2f85ee96b39ba

karlseguin (commented):
There are now a number of stale comments that mention lazy compaction. Other than that, LGTM.

PR #2442, opened 2026-05-13
CI fixes, callback timing correctness

karlseguin:
The point of this PR is to bring stability to the CI and, in so doing, improve the correctness of when callbacks execut (i.e. always on the next tick).  This was driven by debugging CI failures, so it touches a few different things. One specific change is that `cache` is **now always disabled** in the e2e test matrix. The cache always results in incorrect timing, and that should be fixed, but it's cleaner to do in a separate PR. Also, testing caching behavior requires specific cache-aware tests, with specific headers and crucially, multiple requests to the same resources. Simply running existing tests with the cache enable, isn't the most useful.

1 - We buffer worker messages until the worker is ready. This is spec-correct and should eliminate some of the random timeouts we're seeing in CI (where postMessage is queued BEFORE the worker is setup, and the messages were just discarded)

2 - Prevents synchronous callback from happening when a request is made.  This change is probably not needed when https://github.com/lightpanda-io/browser/pull/2303 lands

The interaction between these two issues is interesting. In `main` (which this PR targets), (1) is *not* much of an issue specifically because sync callbacks happen. However, with (2) forcing callbacks to happen only on the next tick, (1) becomes important.

3 - Runner considers whether or not queued requests when deciding if processing was completed or not.

4 - Because 1 introduces some latency, we now also process queued messages after processing any pending messages. This helps avoid having to wait until the next tick to start working on queued messages (which, thanks to #1, we'll now have more of).


karlseguin:
ARGGG...CI is still flaky. But I do think it's a step in the right direction.

src/browser/tests/net/xhr_worker.html
```diff
@@ -0,0 +1,59 @@
+<!DOCTYPE html>
+<script src="../testing.js"></script>
+
+<script id=xhr_in_worker type=module>
+  {
+    const state = await testing.async();
+    const worker = new Worker('./xhr-worker.js');
+    worker.onmessage = (e) => state.resolve(e.data);
+    setTimeout(() => worker.postMessage({ kind: 'basic' }), 100);
```
navidemad: The pending-message buffer this PR adds makes the 100 ms `setTimeout` obsolete and slightly racy — on a loaded CI box it can expire either before or after the worker's script load, so the buffered path gets inconsistent coverage across runs. Posting immediately exercises it deterministically.

```suggestion
    worker.postMessage({ kind: 'basic' });
```

Same on lines 32 and 51 — separate suggestions below.

src/browser/tests/net/xhr_worker.html
```diff
@@ -0,0 +1,59 @@
...
+    });
+  }
+</script>
+
+<script id=xhr_arraybuffer_in_worker type=module>
+  {
+    const state = await testing.async();
+    const worker = new Worker('./xhr-worker.js');
+    worker.onmessage = (e) => state.resolve(e.data);
+    setTimeout(() => worker.postMessage({ kind: 'arraybuffer' }), 100);
```
navidemad:
```suggestion
    worker.postMessage({ kind: 'arraybuffer' });
```

src/browser/tests/net/xhr_worker.html
```diff
@@ -0,0 +1,59 @@
...
+    });
+  }
+</script>
+
+<script id=xhr_document_in_worker_unsupported type=module>
+  {
+    const state = await testing.async();
+    const worker = new Worker('./xhr-worker.js');
+    worker.onmessage = (e) => state.resolve(e.data);
+    setTimeout(() => worker.postMessage({ kind: 'document_unsupported' }), 100);
```
navidemad:
```suggestion
    worker.postMessage({ kind: 'document_unsupported' });
```

src/browser/Runner.zig
```diff
@@ -212,7 +212,7 @@ fn _tick(self: *Runner, comptime is_cdp: bool, opts: TickOpts) !CDPTickResult {
                 },
             }
 
-            if (http_active == 0 and http_client.ws_active == 0 and (comptime is_cdp == false)) {
+            if (http_active == 0 and http_client.ws_active == 0 and http_client.queue.first == null and (comptime is_cdp == false)) {
```
navidemad: Item 3 of the description checks `queue.first` but misses `ready_queue`. The new `makeRequest` wraps `handles.perform()` in a `performing=true` window — any synchronous libcurl callback that ends up in `trackConn` (e.g. JS creates a WebSocket → `WebSocket.zig:152` → `trackConn` → `performing==true` branch) appends to `ready_queue` instead of `in_use`. When `http_client.tick` returns, `ready_queue` can be non-empty and the runner falsely bails with `done`.

```suggestion
            if (http_active == 0 and http_client.ws_active == 0 and http_client.queue.first == null and http_client.ready_queue.first == null and (comptime is_cdp == false)) {
                // we don't need to consider http_client.intercepted here
                // because is_cdp is false, and that can only be
                // the case when interception isn't possible.
                //
                // ready_queue is also part of the check: makeRequest now
                // wraps its handles.perform() in a performing=true window,
                // and any synchronous libcurl callback that ends up
                // calling trackConn during that window (e.g. JS creating
                // a WebSocket) will append to ready_queue. Without this
                // check we could observe it non-empty after
                // http_client.tick returns.
```

navidemad (commented):
Three small follow-up suggestions inline. Validated with codex — first pass caught a real bug in my Worker.zig fix (importScripts reentrancy via `runMacrotasks`), the revised version below is what's actually safe.

Two regression tests deliberately left out, IMO better as a separate PR than bundled here:

- **importScripts ordering**: worker calls `importScripts(...)` in its outer script, parent posts during that window, worker sets `onmessage` after — assert FIFO delivery after load.
- **`ready_queue` idleness**: arrange a sync libcurl callback to create a WebSocket while `performing=true`, assert runner doesn't return `done` with the conn still in `ready_queue`.

src/browser/webapi/Worker.zig
```diff
@@ -185,6 +184,13 @@ fn httpDoneCallback(ctx: *anyopaque) !void {
 }
 
 fn loadInitialScript(self: *Worker, script: []const u8) !void {
+    // Mark loaded BEFORE eval: messages received during eval (e.g. while
+    // an importScripts() call pumps the CDP socket) should be scheduled
+    // normally, not re-buffered. drainPendingMessages below picks up any
+    // messages that arrived strictly before this point.
+    self._script_loaded = true;
+    defer self._worker_scope.drainPendingMessages();
```
navidemad: Current placement marks `_script_loaded` before eval and drains on `defer` *after* `runMacrotasks` — so any pre-eval buffered message gets a scheduler sequence number **higher** than during-eval messages that arrive while `importScripts()` pumps the CDP socket. Scheduler is FIFO by sequence (`Scheduler.zig:32`), so during-eval messages run first and pre-eval slip to the next runner tick. Inverts the spec-required postMessage order.

The obvious fix — drain *before* eval — is unsafe: `importScripts()` calls `ls.local.runMacrotasks()` mid-eval (`WorkerGlobalScope.zig:485`). If the outer script does `importScripts(...)` before assigning `onmessage`, early-drained callbacks fire inside that nested `runMacrotasks` and silently drop at the no-listener check.

Buffering through the entire outer eval (including its trailing `runMacrotasks`) handles both: pre-eval and during-eval messages all stay in `_pending_messages` in receive order, then drain together once eval is fully done.

```suggestion
    // Keep buffering throughout the entire outer eval (including any
    // runMacrotasks pumped by importScripts via the synchronous CDP path,
    // see WorkerGlobalScope.importScripts). The flip-and-drain happens
    // via defer so it runs after eval AND after the trailing
    // runMacrotasks below — by which point the outer script has had its
    // only chance to register onmessage. drainPendingMessages enqueues
    // messages in receive order, so pre-eval and during-eval messages
    // are delivered FIFO on the next runner tick, matching the spec.
    //
    // On eval-throw the defer still fires; the messages get scheduled
    // and then drop at the "no listener" check, mirroring the
    // httpErrorCallback path.
    defer {
        self._script_loaded = true;
        self._worker_scope.drainPendingMessages();
    }
```

krichprollsch (approved):

PR #2461, opened 2026-05-14
Surface at-rules through insertRule and replaceSync (fixes #2459 partial)

### LINKED ISSUE #2459: <style> elements accumulate without bound on CSS-in-JS pages (downstream of #1970/PR #1972)
## Summary

CSS-in-JS pages (styled-components, emotion, Stitches, Mantine, etc.) accumulate `<style>` elements monotonically across user interactions on Lightpanda. On a typical Shopify storefront, clicking variant swatches in sequence causes the document's `<style>` count to grow ~190 elements per click, with no upper bound until the page is closed. On Blink the count stays flat across the same workload.

This is downstream of the deliberate behavior added in PR #1972 / closes #1970, which made `CSSStyleSheet.insertRule` swallow at-rules silently to keep apps from crashing. The silent-success path is the right call for `insertRule`-throws-kills-module-eval, but it has a second-order effect: CSS-in-JS libraries that hash-and-deduplicate their stylesheet via read-back see the rules as missing, conclude the stylesheet is empty, and fall back to direct `<style>` element injection per render. The result is per-render `<style>` accumulation that's invisible to typical fast benchmarks (open-page-and-go) but lethal for long-lived sessions doing repeated DOM interaction.

## Reproducer

Any page using a major CSS-in-JS library will reproduce. We hit this on Allbirds (and presumably every other modern Shopify storefront on the same React stack):

```js
// Connect via CDP, navigate, then:
for (let i = 0; i < 50; i++) {
  const swatches = document.querySelectorAll('button[data-color-swatch]');
  swatches[i % swatches.length]?.click();
  await new Promise(r => setTimeout(r, 1500));
}
console.log('style count:', document.head.querySelectorAll('style').length);
```

URL: `https://www.allbirds.com/products/womens-wool-runners-dapple-grey`

## Measured behavior (per-click DOM tag breakdown over 56 clicks)

```
       tag   first   last   delta
     STYLE     403  11254  +10851   (94% of all DOM growth)
    IFRAME      64    812    +748
    SCRIPT     113    559    +446
      LINK     328    438    +110
   (svg, path, rect, stop, linearGradient, etc.)  unchanged
```

`<style>` climbs almost perfectly linearly across clicks: 403 -> 3163 -> 5923 -> 8688 -> 11254 (every datapoint is one quarter-mark sample of the run, ~14 clicks apart). Static structural elements stay constant, so the growth is concentrated in elements the page is generating per render rather than the underlying layout DOM.

On Blink the same workload yields:
- nodeCount: 3751 -> 3753 (+0%)
- htmlLen: 1.18 MB -> 1.18 MB (+0%)

So Blink is either (a) reusing the same persistent stylesheet via `insertRule` round-trip, or (b) cleaning up old `<style>` elements when the styled component unmounts. On Lightpanda neither is happening.

## Performance impact

On the same workload the per-click cost grows ~2.5x over 56 clicks (446 ms -> 1118 ms average), driven mostly by `outerHTML` cost (the captured payload grows from 2.77 MB to 11.46 MB) and `waitForReady` taking longer to settle as the page grows. On long renders (100+ clicks) we see individual click+settle cycles hitting 8-9 s because the network-quiet window keeps getting nudged by the growing page.

## Suspected mechanism (in `src/browser/webapi/css/CSSStyleSheet.zig`)

```zig
pub fn insertRule(self: *CSSStyleSheet, rule: []const u8, maybe_index: ?u32, frame: *Frame) !u32 {
    const requested_index = maybe_index orelse 0;
    var it = Parser.parseStylesheet(rule);
    const parsed_rule = it.next() orelse {
        if (it.has_skipped_at_rule) {
            log.debug(.not_implemented, "CSSStyleSheet.insertRule", .{});
            // Lightpanda currently skips at-rules (e.g., @keyframes, @media) in its
            // CSS parser. To prevent JS apps (like Expo/Reanimated) from crashing
            // during initialization, we simulate a successful insertion by returning
            // the requested index.
            return requested_index;
        }
        return error.SyntaxError;
    };
    ...
}
```

The silent-success was added to fix #1970 (Expo Web crash) and is correct for that case. The trade-off: a library that does

```js
const idx = sheet.insertRule(`@keyframes spin { ... }`);
const inserted = sheet.cssRules[idx];
if (!inserted || inserted.cssText !== expected) {
  // fallback path: append a fresh <style> element to <head>
}
```

will hit the fallback every render, because Lightpanda returns success but the rule never lands in `cssRules`.

## Suggested fix directions (any one would help; combinable)

1. **Implement at-rule insertion** for at minimum `@keyframes`, `@media`, `@supports`, `@font-face`. These are extraordinarily common in CSS-in-JS output. Even a no-op CSSRule subclass that just stores the source text (without applying it) would let the read-back path return the rule, defeating the per-render fallback.

2. **Throw a typed error for unsupported at-rules** (e.g. `DOMException("Lightpanda: at-rule not yet supported", "NotSupportedError")`) so libraries can feature-detect cleanly. Today they can't tell a successful-but-no-op insertion apart from a real one. PR #1972's contract change (success-with-empty-rules) is what makes the bug invisible at the call site.

3. **Garbage-collect detached `<style>` elements** as a backstop. Even if the library keeps appending fresh `<style>` elements, the old ones should be removable. (See sibling issue I'm filing about `Frame.removeNode` never freeing Node memory; that one is more general but a per-`<style>`-cleanup pass would also work.)

(1) is the cleanest fix for our case. (2) is the most correct contract. (3) is a backstop that helps even if (1) and (2) don't land.

## Environment

- Lightpanda: `1.0.0-nightly.6240+37391687` (binary from official channel)
- Repo HEAD at investigation time: `2f3a426fb0d3` (line numbers above against this rev)
- OS: macOS 14, arm64
- Test driver: puppeteer-core via CDP at `ws://127.0.0.1:9222`
- Page tested: Allbirds product page (Shopify storefront with React + CSS-in-JS)

Happy to provide raw per-click traces, the diagnostic instrumentation we used (just augments `page.evaluate(...)` with tag-count breakdown), or run targeted experiments to disambiguate hypotheses if useful.

staylor:
Closes #2459 (partial -- see "Scope" below).

## Summary

`CSSStyleSheet.insertRule` previously detected at-rules in the parser and returned a fake index without inserting anything (added in PR #1972 to keep at-rules from killing Expo Web's module eval). The silent-success contract has a second-order effect: CSS-in-JS libraries (emotion, styled-components, Stitches, Mantine, Linaria) round-trip through `cssRules` to deduplicate their stylesheets. When the rule is missing on read-back they conclude the stylesheet is empty and fall back to per-render `<style>` element injection, which leaks unboundedly.

This PR makes at-rules surface as opaque `CSSRule` placeholders in `cssRules`, so dedup works. The CSS engine still doesn't apply them (out of scope), but the JS-visible contract matches what libraries expect.

## A/B verification (synthetic CSS-in-JS dedup pattern)

50 calls into a `injectRule()` function that:
1. checks `cssRules` to dedup
2. calls `insertRule(css)`
3. falls back to a fresh `<style>` element if `cssRules.length` didn't grow

```
baseline (1.0.0-nightly.6240): styles=51 rules=0
patched  (this branch):        styles=1  rules=50
```

The 50 fallback `<style>` elements collapse into the single persistent stylesheet that was always intended.

The minimal repro HTML I used is small enough to inline:

```html
<!doctype html><html><head><title>UNSET</title></head><body><script>
function makeInjector() {
  const el = document.createElement('style');
  document.head.appendChild(el);
  const sheet = el.sheet;
  return function inject(css, key) {
    for (let i = 0; i < sheet.cssRules.length; i++) {
      if ((sheet.cssRules[i].cssText || '').indexOf(key) !== -1) return;
    }
    const before = sheet.cssRules.length;
    try { sheet.insertRule(css); } catch (e) {}
    if (sheet.cssRules.length === before) {
      const fallback = document.createElement('style');
      fallback.textContent = css;
      document.head.appendChild(fallback);
    }
  };
}
const inject = makeInjector();
for (let i = 0; i < 50; i++) {
  inject('@keyframes spin { from {} to {} }', 'spin');
}
document.title = 'styles=' + document.head.querySelectorAll('style').length;
</script></body></html>
```

`lightpanda fetch --dump <file>` and grep `<title>` for the result.

## Changes

- **`src/browser/css/Parser.zig`**: `RulesIterator.next` now returns `?Rule` where `Rule` is a union of `{ .style: StyleRule, .at_rule: AtRule }`. The at-rule variant carries the keyword (without `@`) and the full source span. The dead `skipAtRule` helper is removed; the new `consumeAtRule` is its replacement that records spans instead of discarding. `has_skipped_at_rule` is gone (only consumer was `CSSStyleSheet.insertRule`).

- **`src/browser/webapi/css/CSSRule.zig`**: Added `_text: []const u8` field and an `initAtRule(rule_type, text, frame)` constructor for at-rule placeholders. `getCssText` returns the stored text. `CSSStyleRule.getCssText` still wins for `.style` rules via the bridge's most-derived dispatch, so this doesn't change existing behavior.

- **`src/browser/webapi/css/CSSStyleSheet.zig`**: `insertRule` and `replaceSync` switch on the parser's `Rule` union -- regular rules become `CSSStyleRule` (unchanged), at-rules become opaque `CSSRule` placeholders via `CSSRule.initAtRule`. A small `atRuleTypeFor(keyword)` helper maps keywords to `CSSRule.Type` variants; vendor prefixes (`-webkit-`, `-moz-`, `-ms-`, `-o-`) are stripped before lookup. Unrecognized at-rule keywords fall back to `.media` -- this is a deliberate choice to avoid changing the `CSSRule.Type` enum's numeric layout (which is exposed as the spec `CSSRule.type` constant) just to add an `unknown` variant. CSS-in-JS dedup paths key on `length` and `cssText` rather than `type`, so the type fallback is harmless for the issue at hand.

- **`src/browser/StyleManager.zig`**: `addRawRules` switches on the new `Rule` union, processing only `.style` rules (it only filters on display/visibility/opacity at the top level). No semantic change.

- **`src/browser/webapi/css/CSSRule.zig`** (incidental fix): The CSSRule spec constants (`STYLE_RULE`, `KEYFRAMES_RULE`, `MEDIA_RULE`, ...) were declared as plain Zig consts inside `JsApi` but never wrapped with `bridge.property`, so they came back as `undefined` from JS. Wrapped them with `bridge.property(N, .{ .template = true })` matching the pattern in `Node.zig`'s `ELEMENT_NODE` etc. The regression tests need them; without this, every `CSSRule.X` reference in JS evaluates to `undefined`.

## Tests

- **Parser unit tests** (`src/browser/css/Parser.zig`): three new tests for the at-rule path (`surfaces statement at-rules`, `surfaces block at-rules`, `surfaces vendor-prefixed at-rules`). Existing tests updated for the union return shape.

- **HTML runner tests** (`src/browser/tests/css/stylesheet.html`): eight new subtests covering insertRule for each common at-rule keyword (`@keyframes`, `@media`, `@supports`, `@font-face`), vendor-prefixed keyframes, mixed style + at-rule insertion, replaceSync at-rule preservation, and the dedup-via-cssRules pattern that's the actual library code path.

- Full suite green: `make test` reports 644 of 644 tests passed.

## Scope

This PR fixes the specific `insertRule` contract issue identified in #2459 and unblocks the major CSS-in-JS libraries (which all use `insertRule` + cssRules dedup). The synthetic A/B above shows the leak-via-fallback pattern collapsing from 51 styles to 1.

The original Allbirds reproducer in #2459 turned out to involve a Vue.js + Yotpo widget that injects `<style>` elements via direct `document.head.appendChild` rather than through `insertRule`, so this PR doesn't fully eliminate the leak on that specific page. The Vue case appears to be a separate mechanism -- possibly related to `vue-style-loader`'s closure-based dedup interacting with how Lightpanda hosts long-lived JS module state, or to web-component lifecycle differences. I'll file that separately once I have a clean characterization.

### DIFF
--- a/src/browser/StyleManager.zig
+++ b/src/browser/StyleManager.zig
@@ -89,7 +89,13 @@ fn parseSheet(self: *StyleManager, sheet: *CSSStyleSheet) !void {
         const text = try style.asNode().getTextContentAlloc(self.arena);
         var it = CssParser.parseStylesheet(text);
         while (it.next()) |parsed_rule| {
-            try self.addRawRule(parsed_rule.selector, parsed_rule.block);
+            // StyleManager only filters on regular style rules (display,
+            // visibility, opacity). At-rules don't carry top-level
+            // declarations relevant here -- skip them.
+            switch (parsed_rule) {
+                .style => |s| try self.addRawRule(s.selector, s.block),
+                .at_rule => {},
+            }
         }
     }
 }
--- a/src/browser/css/Parser.zig
+++ b/src/browser/css/Parser.zig
@@ -294,9 +294,34 @@ fn isBang(token: Tokenizer.Token) bool {
     };
 }
 
-pub const Rule = struct {
+pub const StyleRule = struct {
     selector: []const u8,
     block: []const u8,
+};
+
+/// An at-rule (`@keyframes`, `@media`, `@supports`, `@font-face`, etc.).
+///
+/// We don't apply at-rules to the page (the CSS engine doesn't process them
+/// yet), but we do surface them so that JS-side reads via `cssRules` see
+/// what was inserted. CSS-in-JS libraries (styled-components, emotion,
+/// Stitches, Mantine) deduplicate their stylesheets by reading back
+/// `cssRules` after `insertRule` -- if the rule is missing they fall back to
+/// per-render `<style>` element injection, which leaks unboundedly. See
+/// lightpanda-io/browser#2459.
+pub const AtRule = struct {
+    /// At-keyword without the leading `@` (e.g., `"keyframes"`, `"media"`,
+    /// `"-webkit-keyframes"`). Borrowed from the input slice; copy if you
+    /// need to outlive the input.
+    keyword: []const u8,
+    /// Full at-rule source span starting at `@` and ending after the closing
+    /// brace (block at-rules) or semicolon (statement at-rules). Borrowed
+    /// from the input slice.
+    text: []const u8,
+};
+
+pub const Rule = union(enum) {
+    style: StyleRule,
+    at_rule: AtRule,
 };
 
 pub fn parseStylesheet(input: []const u8) RulesIterator {
@@ -306,7 +331,6 @@ pub fn parseStylesheet(input: []const u8) RulesIterator {
 pub const RulesIterator = struct {
     input: []const u8,
     stream: TokenStream,
-    has_skipped_at_rule: bool = false,
 
     pub fn init(input: []const u8) RulesIterator {
         return .{
@@ -352,18 +376,14 @@ pub const RulesIterator = struct {
                 var selector = self.input[selector_start.?..selector_end.?];
                 selector = std.mem.trim(u8, selector, &std.ascii.whitespace);
 
-                return .{
+                return .{ .style = .{
                     .selector = selector,
                     .block = self.input[block_start..block_end],
-                };
+                } };
             }
 
             if (peeked.token == .at_keyword) {
-                self.has_skipped_at_rule = true;
-                self.skipAtRule();
-                selector_start = null;
-                selector_end = null;
-                continue;
+                return .{ .at_rule = self.consumeAtRule() };
             }
 
             if (selector_start == null and (isWhitespaceOrComment(peeked.token) or isSemicolon(peeked.token))) {
@@ -395,29 +415,48 @@ pub const RulesIterator = struct {
         }
     }
 
-    fn skipAtRule(self: *RulesIterator) void {
-        _ = self.stream.next(); // consume @keyword
+    /// Consume a full at-rule (statement or block form) and return its
+    /// keyword (without `@`) and full source span. Mirrors `skipAtRule`'s
+    /// termination logic but records spans instead of discarding.
+    fn consumeAtRule(self: *RulesIterator) AtRule {
+        const at_span = self.stream.next() orelse unreachable; // caller peeked an at_keyword
+        const start = at_span.start;
+        const keyword = switch (at_span.token) {
+            .at_keyword => |name| name,
+            else => "",
+        };
+        var end = at_span.end;
         var depth: usize = 0;
         var saw_block = false;
 
         while (true) {
-            const peeked = self.stream.peek() orelse return;
+            const peeked = self.stream.peek() orelse break;
             if (!saw_block and isSemicolon(peeked.token) and depth == 0) {
-                _ = self.stream.next();
-                return;
+                const semi = self.stream.next() orelse break;
+                end = semi.end;
+                break;
             }
 
-            const span = self.stream.next() orelse return;
-            if (isWhitespaceOrComment(span.token)) continue;
+            const span = self.stream.next() orelse break;
+            if (isWhitespaceOrComment(span.token)) {
+                end = span.end;
+                continue;
+            }
 
+            end = span.end;
             if (span.token == .curly_bracket_block) {
                 depth += 1;
                 saw_block = true;
             } else if (span.token == .close_curly_bracket) {
                 if (depth > 0) depth -= 1;
-                if (saw_block and depth == 0) return;
+                if (saw_block and depth == 0) break;
             }
         }
+
+        return .{
+            .keyword = keyword,
+            .text = self.input[start..end],
+        };
     }
 };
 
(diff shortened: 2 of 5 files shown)


karlseguin:
@staylor thanks, just to flag it to you, I did change it to fallback to a new `unknown` type instead of `media`.

PR #2463, opened 2026-05-14
Send `Accept` header when navigating

nikneym:
Prefers Firefox's `Accept` header value, WAF challenges seem to require this. Only affects navigation requests.

### DIFF
--- a/src/Config.zig
+++ b/src/Config.zig
@@ -488,6 +488,9 @@ pub const HttpHeaders = struct {
     // normal client.
     pub const accept_language: [:0]const u8 = "Accept-Language: en-US,en;q=0.9";
 
+    // Document-navigation Accept value Chrome sends.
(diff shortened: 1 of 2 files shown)


karlseguin:
What website is this for? Did you try a less ambitious value? Firefox sends `text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8` and Curl just sends `*/*`. That seems safer than saying we support some signed cryptographic exchange, when we don't.

nikneym:
It was for IMDB, I've just preferred what Chrome sends but Firefox's also work. Made the change.

krichprollsch:
Relates wtih https://lightpanda.slack.com/archives/C05TRU6RBM1/p1774956090948619?thread_ts=1774622671.992309&cid=C05TRU6RBM1

PR #2464, opened 2026-05-14
Fix `Frame` use-after-free for `location.reload()`

nikneym:
Resolves the current root frame again via `session.currentFrame()` before dumping, so a script-driven reload (like WAF challenge) doesn't leave us reading freed memory.

Follow-up to #2463.


karlseguin:
fixed in https://github.com/lightpanda-io/browser/pull/2457

PR #2468, opened 2026-05-15
Fix Fetch.enable + http-proxy CDP send/recv deadlock (#2462)

### LINKED ISSUE #2462: Fetch.enable + --http-proxy deadlocks navigation (subresource events stop firing)
## Summary

Calling `Fetch.enable` (CDP) on a Lightpanda session that was started with `--http-proxy` deadlocks every subsequent navigation. Specifically, `Page.navigate` returns successfully but the document never finishes loading -- subresource `Network.requestWillBeSent` events stop firing, and the navigation hangs until the client-side timeout.

This is the CDP-layer root cause behind a behavior that puppeteer-core's `page.setRequestInterception(true)` exposes -- puppeteer enables the Fetch domain under the hood when interception is requested.

## Reproduction

CDP-level (no puppeteer):

1. Start lightpanda: `lightpanda serve --http-proxy http://127.0.0.1:8080 --port 9222 --host 127.0.0.1`
   (any HTTP CONNECT proxy on `127.0.0.1:8080` reproduces -- a local mitmproxy / squid is fine; it doesn't have to actually reach the upstream)
2. Connect a CDP client, attach to a target with `flatten: true`.
3. Send `Fetch.enable` with `patterns: [{ urlPattern: "*" }]` (or even `Fetch.enable {}` with no patterns).
4. Send `Page.navigate { url: "https://example.com" }`.

Observed: `Page.navigate` returns a frame id; no `Page.frameStoppedLoading`, no `Page.loadEventFired`, no further `Network.requestWillBeSent` after the initial document request. The navigation hangs until the client-side timeout.

Without `--http-proxy`: `Fetch.enable` works correctly -- subresource interception fires, navigation completes normally.

Without `Fetch.enable`: `--http-proxy` works correctly -- pages load through the proxy in the expected time.

The deadlock requires both at once.

## puppeteer-core surface

```js
// Same shape, easier to reproduce locally
const page = await browser.newPage();
await page.setRequestInterception(true);  // calls Fetch.enable internally
await page.goto('https://example.com');   // hangs until timeout
```

I empirically measured this against `https://www.allbirds.com/products/womens-wool-runners-dapple-grey` through a residential-proxy chain: direct fetch through Lightpanda (with `--http-proxy` set) completes in ~4.4 s; the same page with `setRequestInterception(true)` enabled hits the navigation timeout (30 s) and fails.

## Why this matters

`Fetch.enable` is the standard way to filter outbound resource requests at the browser level (block image / font / media / tracking-domain hosts before they fetch). For automation use cases this is critical:

- Scraping at scale: blocking unnecessary subresources cuts page load time 2-5x and saves substantial proxy bandwidth.
- The `--http-proxy` flag is also critical -- routing through residential proxies / proxy classifiers is a common production setup.

When the two can't coexist, callers have to choose: skip the bandwidth filter (heavier pages, more third-party JS executes -- including widgets like Yotpo that inject hundreds of `<style>` elements per interaction), or skip the proxy (loses the routing model entirely). Neither is acceptable for production scrape paths.

## Suspected mechanism

I haven't traced this in the Lightpanda source, but the symptom (subresource events stop firing, document never completes) suggests something in the CDP `Fetch.enable` path either:

1. Doesn't propagate the upstream proxy configuration to the intercepted-and-resumed request, so resumed requests stall at connect time, or
2. Holds a lock / pending-completion handle that the proxy round-trip doesn't release, or
3. Expects the `Fetch.requestPaused` / `Fetch.continueRequest` round-trip to complete on the *direct* request path and breaks when the request is going through the proxy

Happy to instrument and bisect if a maintainer can point me at the right entry point in `src/cdp/domains/fetch*` (or wherever the Fetch domain lives).

## Environment

- Lightpanda: `1.0.0-nightly.6240+37391687`
- OS: macOS 14, arm64
- Test driver: puppeteer-core via CDP at `ws://127.0.0.1:9222`

## Related

- Affects scraper setups that need both per-request filtering and proxy routing.
- Documented as a known constraint in our codebase (we skip resource filtering on Lightpanda only, with a multi-paragraph comment pointing at this exact deadlock) -- happy to share the workaround code if it's useful as a reference for what surface client code uses.

staylor:
Closes #2462.

## Summary

Five coordinated fixes for the `Fetch.enable` + `--http-proxy` deadlock. The original symptom (CONNECT/TLS proxy round-trips inflate concurrency, lightpanda's WS send fills the kernel buffer, lightpanda blocks on write while puppeteer's matching `Fetch.continueRequest` replies pile up unread) is the headline scenario, but along the way an empirical reproducer surfaced four more places that produce the same "lightpanda stops draining the CDP socket and/or crashes mid-cleanup" symptom. All five share the root pattern: a synchronous wait loop or shutdown path that re-enters JS without giving CDP a chance to make progress.

## A/B verification

Repro: puppeteer-core + `setRequestInterception(true)` + `req.continue()` against a localhost forward proxy, lightpanda started with `--http-proxy http://127.0.0.1:8080`.

| URL | before | after |
|---|---|---|
| `https://example.com` | OK 124 ms | OK 117 ms |
| `https://github.com` | HANG 20 s (12/82 continues processed) | OK 1410 ms (82/82) |
| `https://www.shopify.com` | HANG 20 s (1/4) | OK 1973 ms (66/66) |
| `https://www.allbirds.com` | HANG 20 s (12/53) | OK 3944 ms (372/372) |
| `https://www.allbirds.com/products/womens-wool-runners-dapple-grey` | HANG 20 s (12/53) | OK 6286 ms (459/459) |

Lightpanda survives 18 back-to-back navigations across the matrix above (3 per URL) without crashing on the patched binary. Pre-patch, the cleanup path triggered SIGSEGV (`ZigToCurlAllocator.free: incorrect alignment`) inside `BrowserContext.deinit` whenever the page actually completed before puppeteer disconnected.

(`https://www.nike.com` still doesn't fire the `load` event in either version. After this fix all 68 `Fetch.continueRequest` messages process cleanly; the remaining stall is third-party widgets keeping the page in `loading` state, not a CDP/HTTP deadlock. Out of scope.)

## Reproducer

The puppeteer-core script that drove all measurements:

```js

const browser = await puppeteer.connect({
  browserWSEndpoint: 'ws://127.0.0.1:9223',
  defaultViewport: null,
});
const page = await browser.newPage();
await page.setRequestInterception(true);
let seen = 0, continued = 0;
page.on('request', (req) => {
  seen++;
  req.continue().then(() => { continued++; }).catch(() => {});
});
const t0 = Date.now();
try {
  await page.goto('https://www.allbirds.com', { waitUntil: 'load', timeout: 20000 });
  console.log('OK ' + (Date.now() - t0) + 'ms');
} catch (e) {
  console.log('HANG: ' + e.message);
}
await new Promise(r => setTimeout(r, 1000));
console.log('seen=' + seen + ' continued=' + continued);
await browser.disconnect();
```

Lightpanda startup:
```
lightpanda serve --host 127.0.0.1 --port 9223 \
  --http-proxy http://127.0.0.1:8080
```

Any HTTP CONNECT-capable forward proxy on `127.0.0.1:8080` reproduces (vanilla Node `http.createServer` with a `connect` handler is enough).

### `src/network/WsConnection.zig` -- drain reads while waiting for write space

`WsConnection.send` previously switched the socket to blocking mode on `EWOULDBLOCK`. The header comment said this "should virtually never happen" -- under `Fetch.enable` + `--http-proxy` it is routine, and the blocking write deadlocks because the peer is also blocked trying to send to us.

New behavior: on `EWOULDBLOCK`, poll for `POLLOUT` *and* `POLLIN`, drain any incoming bytes into the reader buffer (without dispatching -- that would re-enter `send` and recurse). Buffered bytes are picked up by the next `CDP.readSocket`. Adds `tryRead` (variant of `read` that distinguishes "no new bytes" from "peer closed") and `bufferedBytes` accessor.

### `src/browser/HttpClient.zig` -- never starve CDP reads

`perform()` returned `.normal` early whenever `processMessages()` had work, so a steady stream of HTTP completions never gave the CDP socket a chance to be polled. Now: always do at least a non-blocking `curl_multi_poll` of the CDP socket, and return `.cdp_socket` immediately when the WS reader has bytes that were rescued during a backpressured send (via the new `has_buffered_input` callback on `CDPClient`).

Also makes `Transfer.deinit` idempotent. The error-callback path can cascade into re-entrant deinits (e.g. `Script.errorCallback -> manager.evaluate() -> JS execution -> Frame.deinit -> abortOwner -> Transfer.kill -> Transfer.deinit`); the second deinit then double-frees the Transfer's arena (`ArenaPool: double-free detected` panic). The fix uses `client.transfers.remove(self.id)` as a one-shot ownership claim -- if it returns false, this transfer was already torn down by the cascade and we early-return.

`Transfer.kill` is promoted from `fn` to `pub fn` so `BrowserContext.deinit` can use it (see CDP.zig change below). `RequestParams.deinit` / `Request.deinit` now take `*` instead of `*const` so they can call into the now-mutating `Headers.deinit`.

### `src/network/http.zig` -- make Headers.deinit idempotent

`Headers.deinit` did not null `self.headers` after `curl_slist_free_all`. Combined with the value-copy hazard `RobotsLayer` already documents (Headers wraps a raw `*curl_slist`, so a value copy shares the pointer), a second deinit on a copied Headers double-freed the curl slist. The symptom was an "incorrect alignment" panic inside `ZigToCurlAllocator.free` because the slist nodes were already returned to the pool. Nulling `self.headers` after the free makes the second call a no-op regardless of how the value-copy hazard is triggered.

### `src/browser/ScriptManagerBase.zig` -- drain CDP in `waitForImport`

`waitForImport` spun on `client.tick(200)` and discarded the `.cdp_socket` return. A script `import()` whose request was paused at the `InterceptionLayer` hung forever: the matching `Fetch.continueRequest` reply never reached `CDP.processMessage`, the request never resumed, and this loop spun until the client's navigation timeout fired. Mirrors the `syncRequest` pattern that already does this correctly.

The entry pointer is now re-fetched on every loop iteration. The CDP-drain step above re-enters JS, which can call back into `preloadImport` for a transitively-imported module; `preloadImport` calls `imported_modules.getOrPut`, which may rehash the map and invalidate every existing entry pointer. A cached `entry` was a use-after-free on the next `entry.value_ptr.state` access.

### `src/cdp/CDP.zig` -- wire backpressured-buffer drain + safer teardown

`readSocket` now uses `tryRead` (the new `WsConnection` API) and tolerates the `no_new_data` case so messages drained during a backpressured `send` still get processed.

`BrowserContext.deinit` aborts pending intercepts via `transfer.kill` instead of `transfer.abort`. `kill` fires `shutdown_callback` (or no-op for transfers without one); `abort` fires `error_callback`, which re-enters JS via XHR/script error handlers. The handlers crash because the inspector and V8 context this `BrowserContext` owns have already been torn down (`inspector.stopSession()` runs two lines above the abort loop). The original code already aborted "before closing the session/page since some of these might callback into the page/scriptmanager" -- `kill`'s shutdown path is the right primitive for that.

## Stack of failure modes I observed during debugging

For anyone hitting this region in the future, the symptoms shifted as I peeled layers:

1. `Fetch.requestPaused` events fire in lightpanda but `Fetch.continueRequest` replies never arrive. (Backpressure deadlock -- fixed by the WsConnection + perform changes.)
2. After unblocking the deadlock, the page navigates successfully, then puppeteer disconnects, then `BrowserContext.deinit` SIGSEGVs in `curl_slist_free_all`. (Headers double-free -- fixed by `Headers.deinit` idempotency.)
3. After fixing that, the same teardown path SIGSEGVs differently: `ArenaPool: double-free detected` in `Transfer.deinit`. (Cascade out of `error_callback` re-deinits the same transfer -- fixed by `Transfer.deinit` ownership claim.)
4. After fixing that, the teardown path SIGSEGVs again inside `_v8_inspector__Session__wrapObject`. (XHR/script `error_callback` re-enters JS through a torn-down inspector -- fixed by `kill` instead of `abort` in `BrowserContext.deinit`.)
5. After fixing that, the matrix runs clean.

### DIFF
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -176,6 +176,14 @@ fn handleConnection(self: *Server, socket: posix.socket_t) void {
         return;
     }
 
+    // From here on the CDP reader thread owns the recv side of the socket.
+    // It must be spawned _after_ the handshake (which reads from the same
+    // socket) to avoid two threads racing on socket reads.
+    cdp.startReader() catch |err| {
+        log.err(.app, "CDP startReader", .{ .err = err });
+        return;
+    };
+
     while (true) {
         const next = cdp.tick() catch |err| {
             log.err(.app, "cdp tick", .{ .err = err });
--- a/src/browser/HttpClient.zig
+++ b/src/browser/HttpClient.zig
@@ -170,6 +170,12 @@ pub const CDPClient = struct {
     blocking_read_start: *const fn (*anyopaque) bool,
     blocking_read: *const fn (*anyopaque) bool,
     blocking_read_end: *const fn (*anyopaque) bool,
+    // Returns true if the CDP client has bytes already buffered (typically
+    // rescued from the socket while a synchronous send was backpressured;
+    // see WsConnection.send). When true, perform() returns .cdp_socket so
+    // the runner drains them even if the OS recv buffer is empty. See
+    // lightpanda-io/browser#2462.
+    has_buffered_input: *const fn (*anyopaque) bool,
 };
 
 pub fn init(self: *Client, allocator: Allocator, network: *Network, cdp_client: ?CDPClient) !void {
@@ -413,7 +419,8 @@ pub fn _request(_: *anyopaque, transfer: *Transfer) !void {
 // inside this function, we free it before returning the error. Callers
 // must NOT pair `request()` with their own `errdefer headers.deinit()`
 // — that's a double-free.
-pub fn request(self: *Client, req: Request, owner: ?*Owner) !void {
+pub fn request(self: *Client, req_in: Request, owner: ?*Owner) !void {
+    var req = req_in;
     const arena = self.arena_pool.acquire(.small, "Request.arena") catch |err| {
         req.headers.deinit();
         return err;
@@ -660,21 +667,36 @@ fn perform(self: *Client, timeout_ms: c_int) anyerror!PerformStatus {
 
     // We're potentially going to block for a while until we get data. Process
     // whatever messages we have waiting ahead of time.
-    if (try self.processMessages()) {
-        return .normal;
-    }
+    const processed = try self.processMessages();
 
     var status = PerformStatus.normal;
     if (self.cdp_client) |cdp_client| {
+        // Bytes may have been rescued from the CDP socket while a
+        // synchronous send was backpressured (see WsConnection.send). The
+        // OS recv buffer is empty in that case, but we still owe the
+        // dispatcher a chance to process them. See
+        // lightpanda-io/browser#2462.
+        if (cdp_client.has_buffered_input(cdp_client.ctx)) {
+            return .cdp_socket;
+        }
+
+        // Even when we processed completion messages this round, do a
+        // non-blocking poll of the CDP socket so a steady stream of HTTP
+        // completions can't starve CDP reads. Without this, a flood of
+        // intercepted/proxied transfers can leave Fetch.continueRequest
+        // messages unread for seconds at a time.
+        const cdp_timeout: c_int = if (processed) 0 else timeout_ms;
         var wait_fds = [_]http.WaitFd{.{
             .fd = cdp_client.socket,
             .events = .{ .pollin = true },
             .revents = .{},
         }};
-        try self.handles.poll(&wait_fds, timeout_ms);
+        try self.handles.poll(&wait_fds, cdp_timeout);
         if (wait_fds[0].revents.pollin or wait_fds[0].revents.pollpri or wait_fds[0].revents.pollout) {
             status = .cdp_socket;
         }
+    } else if (processed) {
+        return .normal;
     } else if (running > 0) {
         try self.handles.poll(&.{}, timeout_ms);
     }
@@ -963,7 +985,7 @@ pub const Request = struct {
         return written.ptr[0..written.len :0];
     }
 
-    pub fn deinit(self: *const Request) void {
+    pub fn deinit(self: *Request) void {
         self.headers.deinit();
     }
 };
@@ -1133,6 +1155,22 @@ pub const Transfer = struct {
     }
 
     pub fn deinit(self: *Transfer) void {
+        // Use `transfers.remove` as a one-shot ownership claim. If it
+        // returns false, this transfer has already been deinit'd by a
+        // cascade out of `error_callback` (e.g. Script.errorCallback ->
+        // manager.evaluate() -> JS execution -> Frame.deinit ->
+        // abortOwner -> Transfer.kill -> Transfer.deinit). Returning
+        // here keeps us from double-freeing the arena. See
+        // lightpanda-io/browser#2462.
+        //
+        // Reading `self.id` and `self.client` after a prior deinit has
+        // released `self.arena` is technically a stale read, but
+        // arena_pool zombies the memory until the slot is handed out
+        // again. If by some race the slot were re-used between the two
+        // deinits, the new tenant's id would not be in `transfers`
+        // either, so the early-out still fires and we bail cleanly.
+        if (!self.client.transfers.remove(self.id)) return;
+
         if (self._conn) |conn| {
             self.client.removeConn(conn);
             self._conn = null;
@@ -1146,10 +1184,6 @@ pub const Transfer = struct {
             self.client.queue.remove(&self._node);
             self._queued = false;
         }
-
-        // Drop the id→*Transfer index entry before freeing the memory.
-        // Any concurrent CDP lookup by id will now see this transfer as gone.
-        _ = self.client.transfers.remove(self.id);
 
         self.req.deinit();
         if (self.owner) |o| {
@@ -1177,8 +1211,12 @@ pub const Transfer = struct {
 
     // Owner-driven teardown: fires shutdown_callback (not error_callback)
     // and otherwise behaves like abort. Called by Client.abortOwner /
-    // abortRequests when a Frame / WGS is being torn down.
-    fn kill(self: *Transfer) void {
+    // abortRequests when a Frame / WGS is being torn down, and from
+    // BrowserContext.deinit when the CDP connection drops with paused
+    // intercepts still outstanding -- in both cases firing error_callback
+    // would re-enter JS via XHR/script error handlers, and the inspector
+    // / V8 context the handler reaches into has already been torn down.
+    pub fn kill(self: *Transfer) void {
         if (self.req.shutdown_callback) |cb| {
             cb(self.req.ctx);
         }
--- a/src/browser/HttpClient.zig
+++ b/src/browser/HttpClient.zig
@@ -165,17 +165,15 @@ fn layerWith(self: anytype, next: Layer) Layer {
 // specifically when we're waiting for a request interception response to
 // a blocking script.
 pub const CDPClient = struct {
+    // Fd to poll for "the CDP client has work for us". With the dedicated
+    // CDP reader thread, this is the read end of the mailbox's wake pipe;
+    // it becomes readable when a message has been enqueued.
     socket: posix.socket_t,
     ctx: *anyopaque,
-    blocking_read_start: *const fn (*anyopaque) bool,
+    // Drain any pending CDP messages. Returns false if the CDP connection
+    // has been closed. Called from sync-wait paths (HttpClient.syncRequest,
+    // ScriptManagerBase.waitForImport) and from CDP.tick.
     blocking_read: *const fn (*anyopaque) bool,
-    blocking_read_end: *const fn (*anyopaque) bool,
-    // Returns true if the CDP client has bytes already buffered (typically
-    // rescued from the socket while a synchronous send was backpressured;
-    // see WsConnection.send). When true, perform() returns .cdp_socket so
-    // the runner drains them even if the OS recv buffer is empty. See
-    // lightpanda-io/browser#2462.
-    has_buffered_input: *const fn (*anyopaque) bool,
 };
 
 pub fn init(self: *Client, allocator: Allocator, network: *Network, cdp_client: ?CDPClient) !void {
@@ -671,20 +669,11 @@ fn perform(self: *Client, timeout_ms: c_int) anyerror!PerformStatus {
 
     var status = PerformStatus.normal;
     if (self.cdp_client) |cdp_client| {
-        // Bytes may have been rescued from the CDP socket while a
-        // synchronous send was backpressured (see WsConnection.send). The
-        // OS recv buffer is empty in that case, but we still owe the
-        // dispatcher a chance to process them. See
-        // lightpanda-io/browser#2462.
-        if (cdp_client.has_buffered_input(cdp_client.ctx)) {
-            return .cdp_socket;
-        }
-
-        // Even when we processed completion messages this round, do a
-        // non-blocking poll of the CDP socket so a steady stream of HTTP
-        // completions can't starve CDP reads. Without this, a flood of
-        // intercepted/proxied transfers can leave Fetch.continueRequest
-        // messages unread for seconds at a time.
+        // Always poll the CDP wake fd, even when we just processed HTTP
+        // completions. Without this, a steady stream of completions could
+        // starve CDP reads — the runner would never observe .cdp_socket
+        // and never call into the dispatcher. With the mailbox reader
+        // thread the wake fd is just a pipe, so the poll is cheap.
         const cdp_timeout: c_int = if (processed) 0 else timeout_ms;
         var wait_fds = [_]http.WaitFd{.{
             .fd = cdp_client.socket,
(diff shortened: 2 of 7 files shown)


krichprollsch:
I rebased + added a little change in #2489

karlseguin:
I've been talking about a possible solution for some of these issues. 

Currently, every CDP connection maps to 1 Browser and runs in a thread. I think having the CDP socket polled and read from the main thread (in Network.zig) and communicate messages to the worker via an Erlang OTP-style mailbox would be more robust.

In normal operations, the mailbox would be drained in Runner.tick, i.e. a safe point, never inside of a JS callback. For Request Interception, which needs to block current execution, it would scan the mailbox and only process "safe" messages (a simple allow list of commands) + a timeout.

This adds some complexity at the main <-> worker boundary. It removes the natural back-pressure on incoming CDP messages (which I don't think is really a problem, but there are pretty simple solutions to that if it is). The main upside is that CDP commands that mutate the environment (e.g. Target.closeTarget) are only ever executed in Runner.tick. I'm not sure if this 100% solves the blocking write issue, but it should unblock the peer send which, from the way you desribe it was at least part of the problem. Unrelated to the PR, another benefit is that we can detect a disconnected client even if the worker is stuck in a heavy JS loop.

Would really appreciate any feedback you have about this.

staylor:
let me resolve conflicts and throw Claude at it - this PR is the main blocker for Puppeteer and Playwright.

staylor:
@karlseguin — I took a swing at your mailbox idea as a second commit on top so it's easy to compare (and easy to drop if you'd rather we don't go this direction): [`5577679a`](https://github.com/lightpanda-io/browser/pull/2468/commits/5577679a). The branch is rebased on current `main`, no conflicts.

## Shape

- New `src/cdp/Mailbox.zig`: thread-safe FIFO + wake pipe. Producer is the reader thread (`push` / `close`); consumer is the worker (`pop` / `drainWake` / `freeMessage`). `HttpClient.perform` polls `Mailbox.wake_read` instead of the bare WS socket fd, so the runner still gets a `.cdp_socket` signal as soon as a message lands.
- `CDP.zig`: dedicated reader thread spawned via a new `startReader()` *after* the WS handshake (called from `Server.serveConnection`); spawning earlier races the handshake on socket reads. `deinit()` `ws.shutdown()`s the recv side to wake the reader's blocked poll (POLLHUP), joins, then tears down. `readSocket()` now drains the mailbox.
- `WsConnection.send`: serialized by a `send_mutex` because both threads can reach the send path (reader for protocol close + pong replies, worker for CDP events). Reverted EWOULDBLOCK handling to the simpler "flip to blocking, write, flip back" approach — safe again because the recv side is owned by a different thread.
- Deleted the read/write deadlock-breakers from the squashed parent commit (`waitWritableDrainingReads`, `drainAvailable`, `tryRead`, `bufferedBytes`, `has_buffered_input`) — the mailbox eliminates the deadlock structurally so all of that machinery becomes unnecessary.

Net: ~165 deletions in `WsConnection.zig` traded for ~120 lines of `Mailbox.zig`, plus the reader-thread plumbing in CDP.

## Verification against the PR repro

Local CONNECT proxy on `127.0.0.1:8080`, lightpanda with `--http-proxy http://127.0.0.1:8080`, the exact script from the PR description against `https://www.allbirds.com`, three back-to-back runs:

```
run 1   OK 2422ms   seen=378 continued=378
run 2   OK 2131ms   seen=379 continued=379
run 3   OK 3114ms   seen=393 continued=393
```

(For reference: `HANG 20s (12/53)` pre-fix; `OK 3944ms (372/372)` with just the squashed inline commit.)

All 692 in-tree tests still pass.

## What I didn't change

- The other fixes in the squashed parent commit (`Transfer.deinit` idempotency, `Headers.deinit` idempotency, `BrowserContext.deinit` switching to `transfer.kill`, `ScriptManagerBase.waitForImport` re-fetch + entry guard) are independent of the read path — they fix re-entrant cleanup crashes, not the deadlock — so they stay either way.
- I did **not** implement the "allow-list during sync wait" you described. With the mailbox we already drain everything at every safe point (including `syncRequest`'s tick loop), so paused intercepts always get a chance to resume even mid-XHR/mid-`import()`. If you'd rather tighten that scope — e.g. only allow `Fetch.continueRequest` / `failRequest` / `fulfillRequest` / `continueWithAuth` to dispatch while a sync wait is in flight — it would slot into `Mailbox.pop` or `CDP.readSocket` cleanly.
- Outbound writes are still inline on the worker thread. We could push them through a writer thread too for full-duplex decoupling, but the inbound decoupling alone already breaks the deadlock and keeps the diff small.

## How you'd like to proceed?

A few options I see:

1. **Land this as-is** (two commits on the same PR): the squashed inline-fix as belt, the mailbox as suspenders.
2. **Squash mailbox into the fix commit** and drop the now-redundant inline send/perform machinery from the message.
3. **Split**: keep the cleanup-path fixes (`Transfer.deinit` / `Headers.deinit` / `BC.deinit` / `ScriptManager` guards) on this PR; open a separate PR for the mailbox architecture so it can land on its own review timeline.

Happy to take whichever direction works for you — and to iterate on the prototype (allow-list, outbound writer thread, anything else).

karlseguin:
@staylor thanks for taking a stab at this. This is something I wanted to implement myself (with Claude's help, if I'm being honest). One notable difference with my approach is that I use the main thread (via Network.zig) to poll/read the CDP sockets, so it doesn't require spinning up any new threads.

The PR is https://github.com/lightpanda-io/browser/pull/2495. It doesn't address all of the issues in this PR, but it should address some (along with other issues we've been seeing) and hopefully serves as a better foundation. Would love if you'd be able to take a look and possibly try it.

staylor:
yeah I can took a look in a lil bit - I have no preference on the approach, I just want to unblock our own adoption. if your PR fixes the IO deadlock, we're probably fine. I don't think we actually need #2399 and #2476, but I'll confirm. #2476 sprang up intermediately while I was trying to track down the deadlock, which took multiple multi-hour sessions. Claude even gave up twice.

staylor:
stress-testing these scenarios:
 - A: PR #2495 alone vs 
 - B: PR #2495 + #2476                                                                                                                                                                                           
 - C: PR #2468 alone vs 
 - D: PR #2468 + #2476

staylor:
@karlseguin 🥇 
```
 In this smoke test, PR #2495 alone is the most reliable Playwright path 
against allbirds.com under --http-proxy
```

whatever we can do to get yours merged, let's do it!

 ### Observations                                                                                                                                                                                                                     
                                                                                                                                                                                                                                      
 1. No IO deadlocks observed in any branch. Every failure was either a V8 crash (SIGABRT) or allbirds latency (DOMContentLoaded firing ~4 s past the 20 s budget then everything completing). I never saw the "hang with no error"    
 signature.                                                                                                                                                                                                                           
 2. PR #2495 alone is the most reliable in our smoke test. Across two proxy runs the server stayed alive end‑to‑end (24/25 page loads succeeded; the one fail was an allbirds timing edge, not a server crash).                       
 3. PR #2468 (rebased onto main) consistently V8‑OOMs around iter 8–9 under proxy, whether or not #2476 is applied. The crash is V8 hitting its ~32 MB heap cap in RegExp.exec → oomCallback (Env.zig:563) which then nul‑derefs.     
 4. #2476 did not measurably improve reliability in this test against either base. The "drain deferred dispose" path isn't what's leaking heap here — the V8 isolate is being reused across BrowserContext cycles in both #2468 and   
 #2495 and growing past its cap on the heavy Shopify storefront.                                                                                                                                                                      
 5. The segfault on B & D iter 8 is log.err formatting a null/dangling string during the V8 resolveModuleCallback path — likely downstream of the OOM rather than caused by #2476 itself.                                             
                                                                                                                                                                                                                                      
 ### Caveats                                                                                                                                                                                                                          
                                                                                                                                                                                                                                      
 - Debug build, single page (allbirds.com/products/mens-wool-runners), one fixed wait strategy, on macOS. Different pages / release builds may shift the picture.                                                                     
 - The noop proxy is just a CONNECT tunnel. A real proxy with latency/jitter could expose different races.                                                                                                                            
 - "Deadlock" wasn't reproduced — the test budget (20 s goto, 45–60 s per iter) may not be long enough to expose a slow deadlock vs. a crash.                                                                                         
                                                                                                                                                                                                                                      
 ### Bottom line for your PR                                                                                                                                                                                                          
                                                                                                                                                                                                                                      
 In this smoke test, PR #2495 alone is the most reliable Playwright path against allbirds.com under --http-proxy. Your PR #2468 (rebased on current main) consistently crashes earlier under the same load, and #2476 doesn't fix it  
 — the failure mode is V8 heap exhaustion in the regex compiler, which neither PR addresses.

krichprollsch:
Are we agree #2495 closes this PR and issue #2462?

karlseguin:
Closing. Fixed (or at least significantly improved) by #2495

PR #2469, opened 2026-05-15
`Cookie`: honor SameSite=Strict on cross-site navigation

nikneym:


src/browser/HttpClient.zig
```diff
@@ -905,7 +905,7 @@ pub const RequestParams = struct {
     headers: http.Headers,
     body: ?[]const u8 = null,
     cookie_jar: ?*CookieJar,
-    cookie_origin: [:0]const u8,
+    cookie_origin: ?[:0]const u8 = null,
```
karlseguin: Is this change right? Can a cookie ever have no origin?

src/browser/Frame.zig
```diff
@@ -1356,9 +1366,8 @@ pub fn iframeAddedCallback(self: *Frame, iframe: *IFrame) !void {
 
     new_frame.navigate(url, .{
         .reason = .initialFrameNavigation,
-        // Iframe's initial src request carries the parent's URL as Referer.
```
karlseguin, line 1358: I thought this comment was useful?

src/browser/Frame.zig
```diff
@@ -1356,9 +1366,8 @@ pub fn iframeAddedCallback(self: *Frame, iframe: *IFrame) !void {
...
-        // Parent frame outlives this navigate() call, so the slice is safe.
         .referer = if (std.mem.startsWith(u8, self.url, "http")) self.url else null,
+        .initiator_url = if (std.mem.startsWith(u8, self.url, "http")) self.url else null,
```
karlseguin: duplicate from above, store in local and assign to both referer and initiator_url

Issue #2472, opened 2026-05-15
Frame ID generator resets on page teardown, causing CDP target ID collisions across BrowserContexts


staylor:
# Frame ID generator resets on page teardown, causing CDP target ID collisions across BrowserContexts

## Summary

`Session.frame_id_gen` is reset to `0` in `tearDownActivePage` (`src/browser/Session.zig:189`). Every page allocated after a teardown (whether in a fresh BrowserContext or otherwise) therefore receives `FID-0000000001` again, even though Playwright (and per CDP spec, Chrome) treats target IDs as unique for the lifetime of the CDP connection.

`playwright-core` records every attached target in an in-memory map and asserts on collision:

```
Error: Duplicate target FID-0000000001
    at CRBrowser._onAttachedToTarget (.../chromium/crBrowser.js:162:30)
```

The assertion is fatal: the connection cannot be reused. Clients must drop and re-establish the CDP connection between every BrowserContext, which defeats the point of pooling.

## Reproduction

`lightpanda` built from `fix/2462-fetch-proxy-deadlock` (commit `f05f4517`, version `1.0.0-dev.6261+5687a33d`). Does not require #2399 to be applied — the reset behaviour is independent.

```js
// repro.mjs — `node repro.mjs`
import { chromium } from "playwright-core";
import { spawn } from "node:child_process";

const lp = spawn("lightpanda", [
  "serve", "--host", "127.0.0.1", "--port", "9229",
  "--cdp-max-connections", "4",
  "--disable-subframes", "--disable-workers",
], { stdio: ["ignore", "inherit", "inherit"] });
await new Promise((r) => setTimeout(r, 800));

const browser = await chromium.connectOverCDP("ws://127.0.0.1:9229");

try {
  for (let i = 1; i <= 3; i++) {
    const ctx = await browser.newContext();
    const page = await ctx.newPage();
    await ctx.close();
    console.log(`cycle ${i} ok`);
  }
  console.log("ALL CYCLES OK");
} finally {
  await browser.close().catch(() => {});
  lp.kill("SIGTERM");
}
```

Output (against `f05f4517`):

```
cycle 1 ok
.../assert.js:26
    throw new Error(message || "Assertion error");
          ^
Error: Duplicate target FID-0000000001
    at CRBrowser._onAttachedToTarget (.../chromium/crBrowser.js:162:30)
```

[...]

PR #2476, opened 2026-05-15
Drain deferred disposeBrowserContext on next createBrowserContext (#2472 follow-up)

staylor:
Follow-up to #2472 / #2474. The frame-ID PR fixed the cross-BrowserContext target-ID collision. This one fixes a second `Cannot have more than one browser context at a time` failure that the same Playwright workload hits when the CDP connection is long-lived.

## Problem

`CDP.disposeBrowserContext` (`src/cdp/CDP.zig:378`) defers cleanup when a script eval is on the stack. The existing comment is right about why -- a CDP message drained inside `HttpClient.syncRequest` arrives mid-evaluate, and tearing down the bc here would free Session / V8 / inspector state the unwinding script frame is about to dereference. The comment also said cleanup was "deferred to `CDP.deinit` at connection close, by which time eval has unwound."

That assumption holds for short-lived CDP connections that close after each browser session. It breaks for long-lived ones. Playwright's `chromium.connectOverCDP` keeps the websocket open across many BrowserContexts: once a dispose hits the deferred branch, `browser_context` stays non-null, and the next `createBrowserContext` rejects with `error.AlreadyExists` -> `Cannot have more than one browser context at a time` for the rest of the connection. There is no path to recovery short of disconnecting and reconnecting, which defeats the point of pooling.

## Reproduction

Minimal Node + Playwright + a slow `<script src>` server:

```js
import { chromium } from "playwright-core";
import http from "node:http";
import { spawn } from "node:child_process";

// Slow script server: holds /slow.js open for 3 s so syncRequest is in flight.
  if (req.url === "/slow.js") {
    setTimeout(() => res.end("window.__loaded=true;"), 3_000);
    return;
  }
  res.writeHead(200, { "content-type": "text/html" });
  res.end(`<!doctype html><script src="/slow.js"></script>`);
});
await new Promise((r) => server.listen(0, "127.0.0.1", r));
const PAGE_URL = `http://127.0.0.1:${server.address().port}/`;

const lp = spawn("lightpanda", ["serve", "--host", "127.0.0.1", "--port", "9229"], { stdio: "inherit" });
await new Promise((r) => setTimeout(r, 1_500));
const browser = await chromium.connectOverCDP("ws://127.0.0.1:9229");

for (let i = 1; i <= 3; i++) {
  const ctx = await browser.newContext();
  page.goto(PAGE_URL).catch(() => {});           // start sync /slow.js fetch
  await new Promise((r) => setTimeout(r, 500));  // close mid-fetch
  await ctx.close();                             // -> deferred dispose
  await new Promise((r) => setTimeout(r, 4_000)); // wait long enough for eval to unwind
}
```

Before this PR (against `main`):

```
cycle 1 close() returned
FAILED: browser.newContext: Protocol error (Target.createBrowserContext):
  Cannot have more than one browser context at a time
```

After:

```
cycle 1 close() returned
cycle 2 close() returned
cycle 3 close() returned
ALL CYCLES OK
```

## Fix

Three small changes in `src/cdp/CDP.zig`:

  1. **New `pending_dispose: bool` flag on `BrowserContext`.** Set in the deferred branch of `disposeBrowserContext` instead of just returning. Marks the bc as logically gone but physically still alive while the eval frame above us unwinds.

  2. **New private `flushPendingDispose` helper.** Re-checks `anyScriptEvaluating()` and runs the same `bc.deinit` + `closeSession` + `browser_context = null` the non-deferred path would have if eval is now off the stack. Returns `false` if eval is still on the stack so the caller can surface that to the client.

  3. **`createBrowserContext` calls `flushPendingDispose` before checking `AlreadyExists`.** If the previous deferred dispose can be drained, the new context proceeds normally. If eval is still on the stack at the next create attempt (very tight cycle), the create still rejects with `AlreadyExists` -- waiting for eval to finish is the client's problem, but recovery now happens automatically once it does.

  4. **`isValidSessionId` treats `pending_dispose == true` as no bc.** Commands on the old session_id now reject with `Unknown sessionId` instead of silently routing into a half-alive context.

`Session.removePage` has the same deferred pattern but isn't in the client-visible recovery path (Playwright never calls page-level dispose, only context-level), so this PR doesn't touch it. Filing separately if it turns out to bite something.

## Tests

`zig build test`: 653 / 653 pass (652 existing + 1 new).

Added `cdp.target: createBrowserContext flushes deferred dispose once eval unwinds`:

  - Forces `is_evaluating = true` on the active page's frame.
  - Asserts `disposeBrowserContext` returns success and flips `pending_dispose`.
  - Asserts `createBrowserContext` rejects with `AlreadyExists` while eval is still on the stack.
  - Drops `is_evaluating = false` (eval has unwound).
  - Asserts the next `createBrowserContext` flushes the pending dispose and returns a distinct bc.

Verified the test catches the bug: reverting the CDP.zig changes (keeping only the test) produces a single-test failure, the rest of the suite still passes.

## Relationship to other PRs

  - #2399: Playwright `connectOverCDP` STARTUP-session promotion. Lets Playwright clients connect at all.
  - #2474 (#2472): Frame ID generator scoped to Browser. Lets Playwright clients open a second BrowserContext on the same connection without `Duplicate target FID-...`.
  - **This PR**: lets Playwright clients keep doing that even after a context-close lands during a sync script fetch.

Independent of both -- can land in any order. With all three, Playwright `connectOverCDP` becomes safe to use as a long-lived pooled connection.

### DIFF
--- a/src/cdp/CDP.zig
+++ b/src/cdp/CDP.zig
@@ -345,11 +345,40 @@ fn dispatchCommand(command: *Command, method: []const u8) !void {
 
 fn isValidSessionId(self: *const CDP, input_session_id: []const u8) bool {
     const browser_context = &(self.browser_context orelse return false);
+    // A pending-dispose bc is logically gone from the client's POV --
+    // commands on its session_id should reject like any other unknown
+    // session, not silently route into a half-alive context.
+    if (browser_context.pending_dispose) return false;
     const session_id = browser_context.session_id orelse return false;
     return std.mem.eql(u8, session_id, input_session_id);
 }
 
+// Finish a deferred `disposeBrowserContext` if one is pending and the
+// eval frame that blocked it has unwound. Returns true if the bc was
+// flushed (or there was none). Returns false if a pending dispose
+// remains (eval still on stack -- caller must surface that to the user).
+fn flushPendingDispose(self: *CDP) bool {
+    const bc = &(self.browser_context orelse return true);
+    if (!bc.pending_dispose) return true;
+    if (bc.session.currentPage()) |page| {
+        if (page.frame.anyScriptEvaluating()) return false;
+    }
+    bc.deinit();
+    self.browser.closeSession();
+    self.browser_context = null;
+    _ = self.browser_context_arena.reset(.{ .retain_with_limit = 1024 * 16 });
+    return true;
+}
+
 pub fn createBrowserContext(self: *CDP) ![]const u8 {
+    // Drain any deferred dispose first. Without this, a Playwright-style
+    // `ctx.close(); newContext()` sequence rejects with AlreadyExists
+    // forever once the original close hit the deferred branch in
+    // `disposeBrowserContext` -- the comment there says "deferred to
+    // CDP.deinit at connection close," but long-lived CDP connections
+    // (Playwright `connectOverCDP`) never close between contexts.
+    _ = self.flushPendingDispose();
+
     if (self.browser_context != null) {
         return error.AlreadyExists;
     }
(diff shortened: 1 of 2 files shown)


karlseguin:
Closing. Fixed (or at least significantly improved) by #2495

PR #2479, opened 2026-05-15
Implement Accessibility.queryAXTree CDP method (and fix latent frame-binding bug)

navidemad:
## Summary

Adds `Accessibility.queryAXTree`, the standard CDP primitive for finding AX nodes by role and/or accessible name without serializing the full tree. Motivating use case: agentic / MCP automation clients (Capybara drivers, Playwright + Stagehand, custom MCP integrations) that need *"find all buttons named Submit"* against complex pages where `getFullAXTree` round-trips multi-MB JSON.

Bundles a small fix to the pre-existing `axnodeWriter` while it's in scope — see "Cross-cutting frame-binding fix" below.

## What this PR changes

- **`src/cdp/AXNode.zig`** — `Writer` now takes an optional `Opts.filter`. When `filter == null`, behavior is unchanged (full AX tree, the `getFullAXTree` path). When set, `toJSON` branches into `walkQuery` / `emitMatch`: walks the DOM subtree rooted at the requested node, visits every descendant (including nodes that would be ignored in the AX tree, per the CDP spec), and emits a flat JSON array of matches. A shared `resolveRole` helper handles label-promotion for both the tree path (`writeNode`) and the query path (`emitMatch`) so the two can't drift.
- **`src/cdp/domains/accessibility.zig`** — New `queryAXTree` action in the dispatch enum + handler. Resolves the input node via `dom.getNode`, which already handles `nodeId` / `backendNodeId` / `objectId`. Threads the filter through `axnodeWriter`.
- **`src/cdp/domains/dom.zig`** — Promoted `getNode` from file-private to `pub` so accessibility can reuse it. No behavior change for the existing `DOM.*` callers.
- **`src/cdp/CDP.zig`** — `BrowserContext.axnodeWriter` is now the single entry point (both `getFullAXTree` and `queryAXTree` use it). Frame binding goes through `root.dom.ownerFrame(currentFrame)` (see fix section below).

## Label-promotion parity with `getFullAXTree`

`emitMatch` reuses the shared `resolveRole` helper, so when a `<label>` targets a CSS-hidden checkbox/radio (toggle-switch / CSS-only radio pattern), the label is emitted with the input's role. Without this, `queryAXTree(role="checkbox", name="Enable feature")` on a toggle-switch fixture would return the `display:none` input — a non-actionable target. Real automation clients (Capybara `find`, MCP agents) would then issue a click that silently fails.

## Cross-cutting frame-binding fix

`axnodeWriter` (used by both `getFullAXTree` and `queryAXTree` now) used to bind the writer to `session.currentFrame()`. That's wrong when the root node lives in a different frame — accessible-name resolution (via `Label.findLabelByFor` against `ownerDocument`) and visibility checks (via `frame._style_manager`) are per-frame, so a cross-frame query would compute names/ignored-state against the wrong document.

The writer now derives the frame via `root.dom.ownerFrame(currentFrame)`. The fallback (`currentFrame`) only matters for orphan / detached nodes; for any DOM-attached node, `ownerFrame` returns the right one. This also corrects a latent bug in `getFullAXTree` where the handler resolves the requested `frameId` for the *root document* but the writer then overrode with `currentFrame` for everything else.

## Performance posture

The headline benchmark (123MB peak / 9× faster than headless Chrome) is the reason this primitive matters at all — `getFullAXTree` is the expensive path on rich pages. `queryAXTree`:

- Walks the DOM subtree once
- Reuses the existing visibility / label caches built per call
- Filters before serialization — output is bounded by match count, not subtree size

No new retained state, no extra HTTP, no layout dependency. The frame-binding fix replaces one pointer assignment with one method call (`ownerFrame`) — no measurable overhead.

## Caveats

MVP scope. Each item is a mechanical follow-up PR:

1. **Matches emit empty `properties` and `childIds` arrays and no `parentId`.** Per CDP spec these fields are optional. Clients that need full properties call `Accessibility.getFullAXTree` on a matched nodeId. Cross-reference support is the obvious next PR.
2. **`frameId` not parsed by `queryAXTree`** — queries the frame derived from the supplied nodeId. Clients pass a nodeId already scoped to the frame they care about; explicit `frameId` would be redundant. `getFullAXTree` continues to honor `frameId` for consistency with its prior contract.

## Where to focus review

- `src/cdp/AXNode.zig` — `resolveRole` is the single source of truth for the label-promotion-aware role lookup, called by both `writeNode` (tree mode) and `emitMatch` (query mode). The `walkQuery` walk is intentionally different from `writeNodeChildren`: per the queryAXTree spec it visits AX-ignored nodes (and reports their state via `ignored: true`) rather than pruning hidden subtrees.
- `src/cdp/CDP.zig` — `axnodeWriter` is now the only AX writer entry point; the `ownerFrame(currentFrame)` change applies to both modes. `getFullAXTree` previously bound to `currentFrame` for non-root nodes, so this changes existing behavior on cross-frame queries. The previous behavior was buggy, but if anything in the test suite or downstream code depends on it, this will surface it.
- `src/cdp/domains/dom.zig` — `getNode` is now `pub`. No call-site changes.

## Test plan

- [x] 4 unit tests for query mode in `src/cdp/AXNode.zig`: filter by role, by name, combined-with-label-promotion, no match — all exercise the unified `Writer` with `Opts.filter` set
- [x] 2 CDP-wire integration tests in `src/cdp/domains/accessibility.zig`: missing node identifier (now via `dom.getNode → error.MissingParams`), unknown nodeId
- [x] Full suite: `make test` reports **658 of 658 tests pass** — including the existing `AXNode: writer` and `AXNode: writer prunes hidden and resolves labels` tests, which exercise the tree path post-refactor (verifies `resolveRole` extraction didn't change tree behavior)
- [x] `zig fmt --check ./*.zig ./**/*.zig` clean

### DIFF
--- a/src/cdp/AXNode.zig
+++ b/src/cdp/AXNode.zig
@@ -634,6 +634,119 @@ pub const Writer = struct {
     }
 };
 
+// Backing writer for Accessibility.queryAXTree. Walks the DOM subtree rooted
+// at `root`, computes role + accessible name for every node (including those
+// that would be ignored in the AX tree, per the CDP spec), and emits a flat
+// JSON array of nodes whose role/name match the optional filters.
+//
+// Intentionally MVP: emits an empty `properties` array and empty `childIds`.
+// Clients that need full properties call Accessibility.getFullAXTree on a
+// matched nodeId.
+pub const QueryWriter = struct {
+    root: *const Node,
+    registry: *Node.Registry,
+    frame: *Frame,
+    visibility_cache: *DOMNode.Element.VisibilityCache,
+    label_index: *Label.LabelByForIndex,
+    temp_arena: std.mem.Allocator,
+    accessible_name: ?[]const u8,
+    role: ?[]const u8,
+
+    pub const Opts = struct {
+        accessible_name: ?[]const u8 = null,
+        role: ?[]const u8 = null,
+    };
+
+    pub fn jsonStringify(self: *const QueryWriter, w: anytype) error{WriteFailed}!void {
+        self.toJSON(w) catch |err| {
+            log.err(.cdp, "queryAXTree toJSON", .{ .err = err });
+            return error.WriteFailed;
+        };
+    }
+
+    fn toJSON(self: *const QueryWriter, w: anytype) !void {
+        try w.beginArray();
+        try self.walk(self.root.dom, false, w);
+        return w.endArray();
+    }
+
+    fn walk(self: *const QueryWriter, node: *DOMNode, in_aria_hidden: bool, w: anytype) !void {
+        const axn = AXNode.fromNode(node);
+        try self.maybeEmit(axn, in_aria_hidden, w);
+
+        // <head>, <script>, <style> never expose AX content — skip recursion.
+        if (axn.ignoreChildren()) return;
+
+        const child_in_aria_hidden = in_aria_hidden or blk: {
+            const parent_el = node.is(DOMNode.Element) orelse break :blk false;
+            break :blk hasAriaHiddenTrue(parent_el);
+        };
+
+        var it = node.childrenIterator();
+        while (it.next()) |child| {
+            switch (child._type) {
+                .element, .cdata => {},
+                else => continue,
+            }
+            try self.walk(child, child_in_aria_hidden, w);
+        }
+    }
+
+    fn maybeEmit(self: *const QueryWriter, axn: AXNode, in_aria_hidden: bool, w: anytype) !void {
+        const role_str = axn.getRole() catch return;
+        if (self.role) |needle| {
+            if (!std.mem.eql(u8, needle, role_str)) return;
+        }
+
+        const name = (try axn.getName(self.frame, self.temp_arena)) orelse "";
+        if (self.accessible_name) |needle| {
+            if (!std.mem.eql(u8, needle, name)) return;
+        }
+
+        const node = try self.registry.register(axn.dom);
+        const ignored = axn.isIgnore(self.frame, self.visibility_cache, in_aria_hidden);
+
+        try w.beginObject();
+
+        try w.objectField("nodeId");
+        var buf: [10]u8 = undefined;
+        const id_str = try std.fmt.bufPrint(&buf, "{d}", .{node.id});
+        try w.write(id_str);
+
+        try w.objectField("backendDOMNodeId");
+        try w.write(node.id);
+
+        try w.objectField("ignored");
+        try w.write(ignored);
+
+        try w.objectField("role");
+        try w.beginObject();
+        try w.objectField("type");
+        try w.write("role");
+        try w.objectField("value");
+        try w.write(role_str);
+        try w.endObject();
+
+        try w.objectField("name");
+        try w.beginObject();
+        try w.objectField("type");
+        try w.write("computedString");
+        try w.objectField("value");
+        try w.write(name);
+        try w.endObject();
+
+        try w.objectField("properties");
+        try w.beginArray();
+        try w.endArray();
+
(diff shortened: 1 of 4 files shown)


src/cdp/domains/accessibility.zig
```diff
@@ -67,3 +69,84 @@ fn getFullAXTree(cmd: *CDP.Command) !void {
...
+    if (params.objectId != null and params.nodeId == null and params.backendNodeId == null) {
+        return cmd.sendError(-32000, "Accessibility.queryAXTree by objectId is not yet supported; use nodeId or backendNodeId", .{});
+    }
+
+    const input_id = params.nodeId orelse params.backendNodeId orelse {
+        return cmd.sendError(-32000, "Either nodeId, backendNodeId or objectId must be specified", .{});
+    };
+
+    const bc = cmd.browser_context orelse return error.BrowserContextNotLoaded;
+    const node = bc.node_registry.lookup_by_id.get(input_id) orelse return error.NodeNotFound;
```
krichprollsch: `dom.zig` has a `getNode` function which can also handle `objectId`. Why not using it?

src/cdp/AXNode.zig
```diff
@@ -634,6 +634,131 @@ pub const Writer = struct {
...
 
+// Backing writer for Accessibility.queryAXTree. Walks the DOM subtree rooted
+// at `root`, computes role + accessible name for every node (including those
+// that would be ignored in the AX tree, per the CDP spec), and emits a flat
+// JSON array of nodes whose role/name match the optional filters.
+//
+// Intentionally MVP: emits an empty `properties` array and empty `childIds`.
+// Clients that need full properties call Accessibility.getFullAXTree on a
+// matched nodeId.
+pub const QueryWriter = struct {
```
krichprollsch: Can't we reuse/change the `Writer` maybe with some options to handle that case? It sounds like both are many in common, but I could be wrong.

src/cdp/domains/accessibility.zig
navidemad: Good call — switched to dom.getNode (made it pub). queryAXTree now resolves nodeId / backendNodeId / objectId through the existing helper instead of returning the deferred-support error. 814ca8ab.

src/cdp/AXNode.zig
navidemad: Folded `QueryWriter` into `Writer` via `Opts.filter` (`?Filter`). When `filter == null`, behavior is unchanged (tree mode → `getFullAXTree`). When set, `toJSON` branches to `walkQuery` / `emitMatch` and emits the flat-match shape — same output as before, just under one struct.

Shared a `resolveRole` helper between the tree path (`writeNode`) and the query path (`emitMatch`), so the label-promotion logic can't drift between the two anymore.

The walks themselves intentionally stay separate (per the queryAXTree spec, the query walk visits all subtree nodes including AX-ignored ones, while the tree walk prunes hidden subtrees) — but they share the per-node role+name lookup. `axnodeQueryWriter` is gone; `axnodeWriter` now takes `Opts.filter`. Net -62 lines. 814ca8ab.

src/cdp/AXNode.zig
```diff
@@ -632,6 +666,83 @@ pub const Writer = struct {
...
+        try w.write("computedString");
+        try w.objectField("value");
+        try w.write(name);
+        try w.endObject();
+
+        try w.objectField("properties");
+        try w.beginArray();
+        try w.endArray();
+
+        try w.objectField("childIds");
```
krichprollsch, line 740: You don't fill `childIds` and  `properties` on purpose?

krichprollsch (approved):

arrufat (approved):

PR #2482, opened 2026-05-15
make: forward optional V8_PATH to zig build

navidemad:
Today the build/test targets run `zig build` without `-Dprebuilt_v8_path`, so V8 is rebuilt from source on every invocation (10+ minutes locally).

This PR adds an optional `ZIGFLAGS` variable that's forwarded to each `zig build` call. Contributors with a cached archive can short-circuit:

```
ZIGFLAGS=-Dprebuilt_v8_path=/path/to/libc_v8.a make test
```

When `ZIGFLAGS` is empty (default), behavior is unchanged — no env diff for CI or existing workflows. The variable also lets contributors pass any other `-D...` option (optimize level, snapshot path, etc.) without further Makefile churn.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,11 +6,10 @@ BC := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
 # option test filter make test F="server"
 F=
 
-# Optional path to a prebuilt V8 archive (e.g. V8_PATH=v8/libc_v8.a).
-# When set, every $(ZIG) build invocation below passes -Dprebuilt_v8_path,
-# skipping the multi-minute V8 source rebuild.
-V8_PATH ?=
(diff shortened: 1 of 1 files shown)


karlseguin:
I like this. Do you have any thoughts on making it generic, e.g. `ZIGFLAGS`, would be more useful? Or do you think the explicitness of distinct flags is better?

navidemad:
Good call — switched to `ZIGFLAGS`. It's the standard Make convention (`CFLAGS`/`LDFLAGS`/...) so contributors will guess it works, and it covers any future `-D...` option (`-Doptimize`, `-Dsnapshot_path`, ...) without another Makefile change. Documented the V8 use case in a comment above the variable so the original motivation stays discoverable.

PR #2483, opened 2026-05-15
Dockerfile: fix curl|sh pipefail; trim builder stage

navidemad:
## Summary

Five focused Dockerfile changes. The headline is a real correctness fix: `curl https://sh.rustup.rs | sh` could mask a failed download under `/bin/sh` (no `pipefail`), silently shipping a corrupted Rust install. The rest is build hygiene that shrinks the builder stage by ~540 MB without changing the published image.

## What changes

- **Fix `curl | sh` pipefail bug** in the rustup install. The load-bearing edit:

  ```diff
  -RUN curl https://sh.rustup.rs -sSf | sh -s -- --profile minimal -y
  +RUN curl --fail -sSL --retry 3 --retry-delay 2 -o /tmp/rustup.sh https://sh.rustup.rs && \
  +    sh /tmp/rustup.sh --profile minimal -y && \
  +    rm /tmp/rustup.sh
  ```

- **`--no-install-recommends` + `rm -rf /var/lib/apt/lists/*`** on both apt stages.
- **`curl --retry 3 --retry-delay 2`** on all 4 network downloads (rustup, minisign, zig × 2, V8) — handles transient GitHub/ziglang 5xx without invalidating the build.
- **`git clone --depth 1`** for the lightpanda repo.
- **Drop `v` from `tar`** for minisign and zig extractions — log noise only.

## Measured impact

Cold-cache `docker build --no-cache --pull` on `linux/arm64` (Apple silicon).

|                              |       Before |        After |                            Δ |
| ---------------------------- | -----------: | -----------: | ---------------------------: |
| Final image size             | 69,545,706 B | 69,545,962 B |  +256 B (compression noise) |
| Stage-0 packages installed   |          156 |          116 |             −40 pkgs (−26 %) |
| Stage-0 apt archive download |       194 MB |       117 MB |               −77 MB (−40 %) |
| Stage-0 apt on-disk          |      1144 MB |       605 MB |          **−539 MB (−47 %)** |
| `git clone` working tree     |        28 MB |        9.6 MB |              −18.4 MB (−66 %) |
| `git clone` `.git` dir       |        20 MB |        1.7 MB |              −18.3 MB (−92 %) |
| Total build wall time        |      448.7 s |      442.6 s |             −6.2 s (~1.4 %)  |

The **shipped image is unchanged** (within compression noise) — the published artifact has the same `debian:stable-slim` base + the same `lightpanda` binary + the same `tini`. The reduction lives in the **builder stage / build-cache footprint** on dev and CI machines, which is where the bulk of the wins show up.

The 6 s build-time delta is in the noise — the build is dominated by Rust `html5ever` (~170 s) and the two Zig release builds (~230 s combined).

## Why \`--retry 3 --retry-delay 2\`

Three retries with a 2 s back-off is the sweet spot for GitHub/ziglang.org — long enough to clear a transient 5xx or DNS blip, short enough that a genuinely-down host fails in under 10 s instead of dragging out the build. Matches the pattern in upstream `rustup-init.sh`'s own download retry.

## Where to focus review

The rustup edit (line 14–17). The other four changes are mechanical apt/curl/tar/git flags; this one changes shell structure. The new form is the official rustup non-pipe install pattern, but worth a second pair of eyes on the arg passing (`sh /tmp/rustup.sh --profile minimal -y` vs the old `sh -s -- --profile minimal -y`).

## Deliberately deferred

- **Supply-chain pinning** for the V8 archive and minisign tarball — today only Zig is minisign-verified. Worth a follow-up.
- **`COPY . /browser` vs `git clone`** — current behavior (Dockerfile lives in repo but clones `main` from GitHub) is preserved with `--depth 1`. Switching to `COPY` would be a structural call about Docker Hub publishing semantics.
- **BuildKit cache mounts** (`--mount=type=cache` for cargo and apt) — would cut CI rebuild time significantly but needs `# syntax=docker/dockerfile:1` and BuildKit-enabled environments.
- **Non-root user** in the final image and OCI labels — small hardening, separate PR.

## Test plan

- [x] Cold-cache `docker build` succeeds on `linux/arm64`
- [x] `docker run lightpanda:after` boots CDP server on `:9222`, accepts SIGTERM cleanly via tini
- [x] Image layer hashes verified (debian-slim base + 3 COPY, identical layout to `before`)
- [ ] `linux/amd64` cold-cache build — same Dockerfile, different `case` branch; would appreciate a `docker buildx --platform linux/amd64` confirmation

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,20 +7,25 @@ ARG ZIG_V8=v0.4.5
 ARG TARGETPLATFORM
 
 RUN apt-get update -yq && \
-    apt-get install -yq xz-utils ca-certificates \
+    apt-get install -yq --no-install-recommends xz-utils ca-certificates \
         pkg-config libglib2.0-dev \
-        clang make curl git
+        clang make curl git && \
+    rm -rf /var/lib/apt/lists/*
 
 # Get Rust
-RUN curl https://sh.rustup.rs -sSf | sh -s -- --profile minimal -y
+# Download then execute (rather than `curl | sh`) so a failed download is not
+# masked by sh's exit code under /bin/sh, which has no pipefail.
+RUN curl --fail -sSL --retry 3 --retry-delay 2 -o /tmp/rustup.sh https://sh.rustup.rs && \
+    sh /tmp/rustup.sh --profile minimal -y && \
+    rm /tmp/rustup.sh
 ENV PATH="/root/.cargo/bin:${PATH}"
 
 # install minisig
-RUN curl --fail -L -O https://github.com/jedisct1/minisign/releases/download/${MINISIG}/minisign-${MINISIG}-linux.tar.gz && \
-    tar xvzf minisign-${MINISIG}-linux.tar.gz -C /
+RUN curl --fail -L --retry 3 --retry-delay 2 -O https://github.com/jedisct1/minisign/releases/download/${MINISIG}/minisign-${MINISIG}-linux.tar.gz && \
+    tar xzf minisign-${MINISIG}-linux.tar.gz -C /
 
 # clone lightpanda
-RUN git clone https://github.com/lightpanda-io/browser.git
+RUN git clone --depth 1 https://github.com/lightpanda-io/browser.git
 WORKDIR /browser
 
 # install zig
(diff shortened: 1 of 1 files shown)


krichprollsch (approved):

krichprollsch:
thanks

Issue #2343, opened 2026-05-02
fork of Lightpanda that adds a minimal rendering path


andrewarrow:
I’ve been experimenting with a fork of Lightpanda that adds a minimal rendering path, mainly to support screenshot generation. The goal isn’t to turn it into a full browser engine, but to introduce a lightweight pipeline that can rasterize the existing DOM/CSSOM state into a PNG for common cases.

Before I go too far, I wanted to check:

* has anyone already explored a similar direction in a fork?
* are there internal plans around layout/paint that I should be aware of?
* any obvious architectural constraints that would make this a bad fit?

Happy to share progress or keep it as a separate experimental branch depending on feedback.

ivanUri:
Hi! This is a very interesting direction you're taking.

Lightpanda’s primary strength currently lies in its ultra-lightweight, "rendering-free" architecture, which allows for the high performance and low resource usage that the community values. However, adding a minimal rendering path for screenshots and Canvas rasterization is definitely a feature many users (including myself) have been thinking about.

Regarding your questions:

Architectural Constraints: Lightpanda is built to be fast by skipping the expensive layout/paint phases. The biggest challenge with a "minimal" path is often the Layout Engine. Even for a simple screenshot, you need to accurately calculate box models, text wrapping, and CSS positioning. Are you implementing a custom layout pass in Zig or integrating an existing library (like tiny-skia or yoga)?
Modular Approach: From my perspective, keeping this as an optional/modular pipeline would be ideal. It would allow users to opt-in for "visual" features like screenshots or real toDataURL() results without sacrificing the core engine's performance when they only need DOM automation.
Internal Plans: I haven't seen an official roadmap for a full rendering pipeline yet, as the current focus seems to be on Web API coverage and stability. Your experimental branch could be a great "proof of concept" for how Lightpanda might handle visual output in the future.
I’d love to hear more about your progress, especially how you’re handling the rasterization part. Are you using a specific software renderer for the PNG generation?

Looking forward to seeing where this goes!

andrewarrow:
Thanks, this matches how I’m thinking about it too. I definitely do not want to compromise Lightpanda’s rendering-free default path.

The current experiment is intentionally opt-in: fetch --make-png <path> <url>. Normal DOM automation should not pay for layout/paint, and I’m keeping the rendering module off the hot path unless that flag is used.

[...]

PR #2487, opened 2026-05-16
Add --enable-external-stylesheets flag with fetch + parse

navidemad:
## What this fixes

Lightpanda silently dropped every `<link rel="stylesheet">` — no fetch, no
parse, no entry on `document.styleSheets`, no contribution to the visibility
cascade. Inline `<style>` blocks worked end-to-end via StyleManager, but
external sheets were explicitly marked "out of scope for the headless engine"
(see `StyleManager.zig:124-125` on `main`). That gap hurts agentic and
testing use cases where accurate `checkVisibility()` / `getComputedStyle()`
matters, while leaving the cheap no-fetch fast path intact for pure
scraping/crawling.

## What this PR changes

Behind the opt-in `--enable-external-stylesheets` flag (or
`LP.configureLoading.externalStylesheets` per-session), `<link rel=stylesheet>`
elements — both parser-emitted and JS-created — now:

1. Resolve `href` against the frame base.
2. Fetch synchronously via `HttpClient.syncRequest` (same path
   `ScriptManager.addFromElement` uses).
3. Hand the body to `CSSStyleSheet.replaceSync`, which parses + populates
   `cssRules` + calls `StyleManager.sheetModified()`.
4. Register the sheet on `document.styleSheets` and fire `load` (or `error`
   on any failure — status non-2xx, oversize body, parse error).

The whole feature is gated on `session.load_external_stylesheets`. With the
flag off, behavior is unchanged from `main`: the synthetic load-event stub
in `linkAddedCallback` still fires for `rel=stylesheet/preload/modulepreload`,
no network traffic, no entries on `document.styleSheets`.

## Notable design choices

**Synchronous fetch from the parser callback.** Matches
`ScriptManager.addFromElement` precedent and how real browsers handle CSS
(render-blocking). Lets us reuse `_pending_loads` semantics without
bookkeeping — by the time `linkAddedCallback` returns, the sheet is live.

**`replaceSync` does the parse + cascade wiring.** No `StyleManager` or
`CSSStyleSheet` API changes; the existing inline-`<style>` ingestion path
already does exactly what's needed once we hand it the bytes.

**Cached sheet on `Link._sheet`.** Mirrors `Style._sheet`. On the first
load the sheet is created and registered on `document.styleSheets`; on
subsequent loads (e.g., mutating `link.href` on a connected element) the
existing sheet is reused via `replaceSync` so the styleSheets list stays
stable and old rules don't accumulate. Fetch failures leave the previous
sheet intact — matches browser behavior where a broken `href` doesn't
invalidate the previously loaded sheet until the link itself is removed.
Disconnect cleanup removes the sheet symmetrically so SPA theme-switch
patterns (append new sheet, remove old) don't leak phantom entries.

**`Build.created` on Link.** Pre-existing gap: static `<link>` elements
(void, no closing tag) never reached `nodeComplete` → `linkAddedCallback`
was a dead callback for static head links. Mirrors `Image.Build.created`.
Side effect: the synthetic `load` event now also fires for static
`<link rel=stylesheet>` in the disabled-flag case — matches browser
behavior, brings Link in line with Image.

**Fragment-mode gate.** `Build.created` fires for parser-instantiated
elements before attachment, including content parsed via `innerHTML` /
`outerHTML` / `insertAdjacentHTML` / `Range.createContextualFragment` /
`<template>` content / `DOMParser.parseFromString`. Without a guard these
would have fetched against the live document and registered phantom
sheets even when the parsed subtree was never attached.
`loadExternalStylesheet` short-circuits on `_parse_mode == .fragment`,
and `DOMParser.parseFromString` now brackets its `parser.parse` calls
with the same fragment parse-mode `parseHtmlAsChildren` uses (which also
incidentally stops DOMParser-emitted `<script>` tags from executing
against the live frame).

**`is_evaluating` bracket around `syncRequest`.** Sync HTTP from inside
the parser callback pumps the CDP socket, so a `Target.closeTarget`
arriving mid-fetch could otherwise drive `Session.removePage` while we
still hold `self` — UAF. The bracket every other `syncRequest` caller
uses is what `Session.removePage`'s reentrancy guard
(`anyScriptEvaluating()`) checks.

**`.stylesheet` ResourceType.** CDP Network events now report
`"Stylesheet"` per spec, instead of falling back to `"Fetch"`.

## Why 2 MiB cap (and what it actually protects)

The 2 MiB cap in `Frame.loadExternalStylesheet` is a **CSS-parser-arena
protection**, not a network protection. It bounds how much CSS the parser
will turn into `CSSRule` objects on the frame's medium arena.

Network-level streaming protection already exists via
`HttpClient.max_response_size`, enforced at `HttpClient.zig:1496/1534/1545`
both pre-request (Content-Length) and per-chunk during the body callback.
Drivers that need a tighter network cap should set
`config.httpMaxResponseSize`.

A per-request streaming cap specifically for stylesheets would let us
short-circuit large responses before the global cap, but it requires a new
field on the `Request` struct in `HttpClient.zig` — exactly the territory
#2303 is refactoring. Deferred to a follow-up after the network layer
stabilizes.

Tailwind's full `preflight + utilities` raw build is ~3 MiB; at that size
a site should already be splitting by route. 2 MiB is well above anything
seen on real production sites.

## Caveats (v1)

- No CORS check. Cross-origin sheets fetch regardless of the `crossorigin`
  attribute. Matches how scripts work today.
- No MIME enforcement. `SyncResponse` doesn't expose response headers, and
  changing that touches `HttpClient.zig` — #2303 territory. Follow-up.
- 2 MiB cap is post-buffer (see "Why 2 MiB" above). Network cap from
  `HttpClient.max_response_size` already streams.
- `CSSStyleSheet.replaceSync` itself is not atomic: it clears `cssRules`
  before its insert loop, so an OOM mid-insert can leave the cached sheet
  with old rules dropped and new rules partially loaded. `_href` is held
  back until `replaceSync` succeeds, so URL/content stay consistent — but
  full atomicity would require a scratch-list pattern in `replaceSync`.
- No `@import` recursion or `url(...)` font/image fetch.

## Where to focus review

- `Frame.loadExternalStylesheet` (size cap, `is_evaluating` bracket,
  `errdefer`/ownership-transfer on headers, `_href` ordering, cached-sheet
  reuse on re-fetch).
- `Link._sheet` field + `Build.created` + the disconnect-walker branch
  in `Frame.zig` — confirms the no-leak invariant under
  append/remove/href-mutation.
- `DOMParser.parseFromString` parse-mode bracket — incidentally also
  stops DOMParser scripts from running against the live frame.
- `src/browser/tests/css/external_stylesheet.html` — eight sub-tests
  covering static + dynamic load, cascade reaching StyleManager, 404,
  oversize, href-change-replaces-sheet, fragment-parse-skipped (innerHTML
  + DOMParser), disconnect-removes.

## Real-world validation

Built `1.0.0-dev.6282+fdbe2a269` (`ReleaseSafe`) and ran against the
[`capybara-lightpanda`](https://github.com/navidemad/capybara-lightpanda)
driver suite as a downstream sanity check:

| Scenario | Flag OFF | Flag ON | Result |
|---|---|---|---|
| **Synthetic page** — external sheet with `.hide-me { display: none }`, `checkVisibility()` on a matching div | `styleSheets=0`, `checkVisibility=true` | `styleSheets=1`, `checkVisibility=false` | ✅ Cascade reflects external CSS only when enabled |
| **`rake test:all`** — full capybara-lightpanda test suite (313 runs, 562 assertions) | 313/313, 4 skips | 313/313, same 4 skips | ✅ No regressions from new fetch path or `Build.created` |
| **`https://lightpanda.io/`** — production page load via the gem | `styleSheets=2` (inline only) | `styleSheets=3` (+1 external link) | ✅ Flag actually pulls in external CSS in the wild |

The driver was invoked via a small `LIGHTPANDA_EXTRA_ARGS` passthrough on
its `Process#build_args` so the same cached binary could be A/B-tested
without rebuilding. No driver-side code knows about the flag — purely a
CLI/CDP toggle.

## Compatibility with #2303

Keeping this PR in draft until #2303 lands, per @krichprollsch's note about
not piling new fetch sites onto the in-flight network refactor. Disjoint
files apart from `HttpClient.zig` (1 enum variant) and `cdp/domains/fetch.zig`
(2 switch arms). Rebase delta after #2303 lands is concrete and small:

### Concrete rebase delta

<details>
<summary>What #2303 changes that touches this PR</summary>

**1. `Request` struct splits into two types.** Post-2303 introduces a
`RequestParams` struct (input to `syncRequest` and `Client.request`) that's
a strict subset of `Request`:

| Field                  | Pre-2303 `Request` | Post-2303 `RequestParams` |
|------------------------|--------------------|---------------------------|
| `frame_id`, `loader_id`, `method`, `url`, `headers`, `body`, `cookie_jar`, `cookie_origin`, `resource_type`, `credentials`, `notification`, `timeout_ms` | yes | yes |
| `arena`, `request_id`  | n/a                | injected by `Client.request` |
| `protect_from_abort`   | n/a                | yes (navigation abort safety) |
| `ctx`, callback hooks (`start_callback`, `header_callback`, `data_callback`, `done_callback`, `error_callback`, `shutdown_callback`) | yes | n/a (sync doesn't need them) |
| `skip_robots`          | yes                | n/a |

Rebase change at the `loadExternalStylesheet` call site: the field set we
pass (`url`, `method`, `frame_id`, `loader_id`, `headers`, `cookie_jar`,
`cookie_origin`, `resource_type`, `notification`) is already a subset of
`RequestParams`, so the struct literal compiles unchanged. The type the
compiler infers becomes `RequestParams` instead of `Request`. No code edit
needed at the call site itself.

**2. `ResourceType` enum lives in two parallel places.** Post-2303 has
`RequestParams.ResourceType` AND `Request.ResourceType`, both with identical
variant lists. The rebase must add `.stylesheet` to *both* enums and to
both `string()` mappings (currently this PR only touches one because pre-2303
has a single enum).

**3. Threading model changes — transparent for this PR.** `HttpClient`
becomes a thin wrapper around `Network.Handle`; libcurl runs on a shared
main thread with cross-thread signaling. `syncRequest` from a CDP worker
thread (where `linkAddedCallback` already executes) keeps its blocking
semantics — the worker thread waits on a condition variable while the main
thread drives the transfer. We inherit the same stability guarantee
`ScriptManager.addFromElement` does; if #2303 breaks `syncRequest` under
the new threading model, scripts break too, so the whole browser is gated
on that path working.

**4. `HttpClient.process` internals change** (`network.getConnection()` →
`handle.getConnection()`) — doesn't affect our code, no rebase action.

**5. CDP `fetch.zig` switches.** Same `.stylesheet` arms apply pre- and
post-2303; the surrounding diff might shift line numbers but the change is
mechanically identical.

### Estimated rebase work

5–10 minutes once #2303 lands: add `.stylesheet` to the second
`ResourceType` enum + `string()` mapping in `HttpClient.zig`, re-run
`make test`. No semantic changes to the stylesheet fetch path.

</details>

## Test plan

- [x] `make test` — 694/694 pass
- [x] `zig fmt --check ./*.zig ./**/*.zig` clean
- [x] `WebApi: HTML.Link` — flag-disabled assertion confirms no sheet
      registered when off
- [x] `WebApi: HTML.Link external stylesheet` — flag-enabled covers
      static + dynamic load, cascade, 404, oversize,
      href-change-replaces, fragment-parse-skipped (innerHTML +
      DOMParser), disconnect-removes
- [x] Downstream driver: `capybara-lightpanda` `rake test:all` 313/313
      pass identically with flag on/off
- [x] Real-world: `https://lightpanda.io/` loads cleanly and picks up
      the external stylesheet with the flag enabled

Refs #2343

### DIFF
--- a/src/browser/Frame.zig
+++ b/src/browser/Frame.zig
@@ -1749,16 +1749,24 @@ pub fn loadExternalStylesheet(self: *Frame, link: *Element.Html.Link) !void {
         return;
     }
 
-    const sheet = try CSSStyleSheet.initWithOwner(element, self);
+    // Reuse the cached sheet on re-fetch (href mutation on a connected
+    // link) so `document.styleSheets` keeps a single entry per <link>
+    // instead of accumulating one per href change. On first load, create
+    // and register; on subsequent loads, replace content in place.
+    const sheet = link._sheet orelse blk: {
+        const new_sheet = try CSSStyleSheet.initWithOwner(element, self);
+        link._sheet = new_sheet;
+        const sheets = try self.document.getStyleSheets(self);
+        try sheets.add(new_sheet, self);
+        break :blk new_sheet;
+    };
+
     sheet._href = try self.arena.dupe(u8, resolved);
     sheet.replaceSync(response.body.items, self) catch |err| {
         log.warn(.browser, "external stylesheet parse", .{ .err = err, .url = resolved });
         try self.fireLinkEvent(link, comptime .wrap("error"));
         return;
     };
-
-    const sheets = try self.document.getStyleSheets(self);
-    try sheets.add(sheet, self);
 
     try self.fireLinkEvent(link, comptime .wrap("load"));
 }
--- a/src/browser/Frame.zig
+++ b/src/browser/Frame.zig
@@ -52,6 +52,7 @@ const AbstractRange = @import("webapi/AbstractRange.zig");
 const MutationObserver = @import("webapi/MutationObserver.zig");
 const IntersectionObserver = @import("webapi/IntersectionObserver.zig");
 const Worker = @import("webapi/Worker.zig");
+const CSSStyleSheet = @import("webapi/css/CSSStyleSheet.zig");
 const CustomElementDefinition = @import("webapi/CustomElementDefinition.zig");
 const PageTransitionEvent = @import("webapi/event/PageTransitionEvent.zig");
 const SubmitEvent = @import("webapi/event/SubmitEvent.zig");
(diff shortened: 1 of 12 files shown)


karlseguin (changes requested):
I don't want to merge this until the feature is in. Having the parameter show up in the `help` seems like it could cause confusion. I know Claude and friends can pick the help output pretty easily.

navidemad:
I wanted to make it small PR, but i guess instead could make a lightweight implementation with the feature 👍🏻

src/browser/Frame.zig
```diff
@@ -1679,6 +1680,131 @@ pub fn queueLoad(self: *Frame, html: *Element.Html) !void {
...
+    // mid-fetch would otherwise drive `Session.removePage` while this
+    // function still holds pointers to `self`. The check in
+    // `Session.removePage` (Session.zig:253) consults
+    // `frame.anyScriptEvaluating()`, which only sees this flag.
+    const sm = &self._script_manager.base;
+    const was_evaluating = sm.is_evaluating;
+    sm.is_evaluating = true;
+    defer sm.is_evaluating = was_evaluating;
+
+    var response = http_client.syncRequest(arena, .{
```
krichprollsch, line 1740: Do we need sync request for stylesheets? Can't we use async ones?

karlseguin: I don't think so.  There is preload, but that would be a different implemented differently (maybe paired with the cache).

There are still some gaps in the implementation. A stylesheet added dynamically via innerHTML = '<link...'  won't get loaded. And, as stated in the PR, `@import` also isn't supported.

karlseguin (approved):

PR #2494, opened 2026-05-19
feat: add --json to fetch command

marchelbling:
The `fetch` command is very practical to render pages without needing to have a long running browser instance.
It is however masking all details on the fetch, most importantly the HTTP status code. This is a big caveat when leveraging `lightpanda fetch` in a pipeline.

This introduces a `--json` option to provide a structured output that contains:
* url
* HTTP status code
* response headers
* rendered content as controlled by the `--dump` option

The proposal is to always output the same JSON format even when not using `--dump` with an option.

### DIFF
--- a/src/browser/Frame.zig
+++ b/src/browser/Frame.zig
@@ -257,7 +257,7 @@ _type: enum { root, frame }, // only used for logs right now
 _req_id: u32 = 0,
 _navigated_options: ?NavigatedOpts = null,
 _http_status: ?u16 = null,
-_http_headers: std.ArrayListUnmanaged(HttpHeader) = .empty,
+_http_headers: std.ArrayList(HttpHeader) = .empty,
 
 pub const HttpHeader = struct {
     name: []const u8,
--- a/src/browser/Frame.zig
+++ b/src/browser/Frame.zig
@@ -257,7 +257,12 @@ _type: enum { root, frame }, // only used for logs right now
 _req_id: u32 = 0,
 _navigated_options: ?NavigatedOpts = null,
 _http_status: ?u16 = null,
-_http_headers: std.StringArrayHashMapUnmanaged([]const u8) = .empty,
+_http_headers: std.ArrayListUnmanaged(HttpHeader) = .empty,
+
+pub const HttpHeader = struct {
+    name: []const u8,
+    value: []const u8,
+};
 
 pub fn init(self: *Frame, frame_id: u32, page: *Page, parent: ?*Frame) !void {
     if (comptime IS_DEBUG) {
@@ -452,14 +457,14 @@ pub fn getTitle(self: *Frame) !?[]const u8 {
 pub const HttpMetadata = struct {
     url: [:0]const u8,
     status: ?u16,
-    headers: std.StringArrayHashMapUnmanaged([]const u8),
+    headers: []const HttpHeader,
 };
 
 pub fn httpMetadata(self: *const Frame) HttpMetadata {
     return .{
         .url = self.url,
         .status = self._http_status,
-        .headers = self._http_headers,
+        .headers = self._http_headers.items,
     };
 }
 
@@ -1043,9 +1048,10 @@ fn frameHeaderDoneCallback(response: HttpClient.Response) !bool {
     self._http_status = response.status();
     var it = response.headerIterator();
     while (it.next()) |hdr| {
-        const name = try self.arena.dupe(u8, hdr.name);
-        const value = try self.arena.dupe(u8, hdr.value);
-        try self._http_headers.put(self.arena, name, value);
+        try self._http_headers.append(self.arena, .{
+            .name = try self.arena.dupe(u8, hdr.name),
+            .value = try self.arena.dupe(u8, hdr.value),
+        });
     }
 
     if (self._navigated_options) |no| {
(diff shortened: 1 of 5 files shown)


marchelbling:

marchelbling:
I hope it's fine to open a PR as a way to make a suggestion with a concrete implementation (the CONTRIBUTING.md didn't surface this).
In my current usage, not being able to get details on the underlying request is a blocker; I however understand that this extension may not be welcomed as it modifies more than the fetch logic. 

Happy to hear your thoughts or alternative solutions that could unlock me!

krichprollsch:
Hello @marchelbling,

Thanks for your contribution. It's fine to open a PR directly. But I'm not sure about a `json` option.

Generally speaking, we tend to follow the curl's behavior for the `fetch` command, even if many options are fairly different.

W/ curl, you get directly the HTTP response's body as output, and you can use `--dump-header -`  to dump response's headers (but it will dump all the successive headers in case of redirection for example: `curl  `--dump-header - -L  https://httpbin.io/absolute-redirect/6`)

May I ask why getting the status code is a blocker for you and how you want to use it?

Thanks!

src/lightpanda.zig
```diff
@@ -157,38 +158,101 @@ pub fn fetch(app: *App, browser: *Browser, url: [:0]const u8, opts: FetchOpts) !
...
+    try writeJsonString(writer, body);
+
+    try writer.writeAll("}\n");
+}
+
+fn writeJsonString(writer: *std.Io.Writer, s: []const u8) !void {
+    try writer.writeByte('"');
+    try std.json.Stringify.encodeJsonStringChars(s, .{}, writer);
+    try writer.writeByte('"');
+}
```
arrufat: Instead of handrolling the JSON, you could use `std.json.Stringify`.
Check: https://github.com/lightpanda-io/browser/blob/e61eddf9562ea9ed669555f28a3d0e9f07e5dc5b/src/cookies.zig#L114

src/browser/Frame.zig
```diff
@@ -256,6 +256,8 @@ _parent_notified: bool = false,
 _type: enum { root, frame }, // only used for logs right now
 _req_id: u32 = 0,
 _navigated_options: ?NavigatedOpts = null,
+_http_status: ?u16 = null,
+_http_headers: std.StringArrayHashMapUnmanaged([]const u8) = .empty,
```
arrufat: Some HTTP headers can be repeated, so something like an `ArrayList(struct { name, value })` would be more appropriate here.

src/lightpanda.zig
```diff
@@ -157,38 +158,101 @@ pub fn fetch(app: *App, browser: *Browser, url: [:0]const u8, opts: FetchOpts) !
...
+            try writeJsonString(writer, value);
+        }
+    }
+    try writer.writeAll("}");
+
+    try writer.writeAll(",\"dump\":");
+    try writeJsonString(writer, if (dump_mode) |mode| @tagName(mode) else "");
+
+    try writer.writeAll(",\"body\":");
+    try writeJsonString(writer, body);
```
arrufat: `body` is an overloaded name, which might be confused with the HTTP body.

marchelbling:
@krichprollsch Thanks for your quick feedback!
I recently started working on a PoC leveraging `lightpanda fetch` to introduce headless JS rendering in a web crawler. The status code is critical for me to determine how to continue my crawl. 

I initially considered a curl-style `-w` output, but for my use case a structured JSON response is easier and more reliable to consume programmatically, especially when combining status code and rendered content in one parseable payload. 

Happy to revisit this contribution if you think it's a better fit for the project!

src/lightpanda.zig
marchelbling: Renamed to `content` [review](https://github.com/lightpanda-io/browser/pull/2494/commits/fec2bbda7b0d1864e20badb509538a9d0b6c6d0a)

src/browser/Frame.zig
marchelbling: Good point, updated

marchelbling:
@arrufat thank you for your comments. I pushed an update that addresses all of them. I suppose this could overall be reworked and exposed as a configurable output more similar to what curl does (for `--dump-header` or `--write-out` as per @krichprollsch comment).
Let me know what you think!

karlseguin:
Opps..I started to review this before realizing my colleagues were already in on the fun.  I pushed a small commit that importantly adds help to the `--json` flag (the `help.zon` is easy to miss). It does some other trivial tweaks. Whatever shape this takes, make sure to update `help.zon` appropriately.

I personally like the feature. I agree there's a bit of awkwardness around the naming / interaction of --dump & --json, but I don't have a good solution, and they are composable, e.g. `--dump markdown --json` and `--dump html --json`

marchelbling:
Thank you @karlseguin!
My initial thought was to have a `--json` option that would be opinionated about the content format. Then I figured that having `--json` only controling the shape of the response and `--dump` controling the format of the content was the best way forward to keep responsibilities split and therefore better backward compatibility.
Then it's true that both option wording might convey a different semantic at first sight.

krichprollsch (approved):

marchelbling:
@krichprollsch thank you for accepting. 
Should I consider that I can merge now or would you like to think about this some more?

krichprollsch:
Thanks @marchelbling for your contribution.
FYI on `main` we have a crash currently with `fetch` command, but #2502 will fix it.

PR #2495, opened 2026-05-19
Main/Network reads CDP socket

karlseguin:
Previously, the CDP socket was added to the worker's multi and fully owned by the worker. While this is simple, it introduced some issues:

1 - Cannot detect a disconnected client during JS processing ( for(;;) )

2 - A blocked worker can cause back-pressure that blocks the client. This can
    cause a deadlock if the worker is blocked waiting for a CDP message

In addition to these 2 problems, there was 1 other serious CDP-related issue: arbitrary CDP messages could be processed during JavaScript callback. For example, a Worker calls importScripts while request interception is enabled, this requires us to tick the HttpClient waiting for the interception response. But, a client could sent Target.closeTarget, which we'd process and delete the frame..all while importScripts is still blocked. Assuming importScripts unblocks everything is a big UAF since the frame (and its workers) were cleared from closeTarget.

The CDP socket is now read from the network (main) thread and an OTP-style mailbox is used. The network thread posts message to the Worker's inbox and signals it to wakeup. This solves #1 and #2. It doesn't directly solve the reentrancy issue, but it provides the foundation. Specifically, in introduces a queue for of CDP message and more control over when/how that queue is processed. At "safe points" (Runner.tick, HttpClient.tick), any message can be processed. But, when inside a JavaScript callback, we can process only non- destructive/mutating message. Specifically, we can process only messages related to request interception.

### DIFF
--- a/src/browser/HttpClient.zig
+++ b/src/browser/HttpClient.zig
@@ -667,9 +667,14 @@ fn perform(self: *Client, timeout_ms: c_int) anyerror!void {
         try self.trackConn(conn);
     }
 
-    // Process completions queued from the curl_multi_perform above before
-    // we potentially block.
-    _ = try self.processMessages();
+    // We just processed completions; their done_callbacks may have
+    // scheduled microtasks (JS continuations) or queued new transfers.
+    // Return without polling so the caller (_tick) can run macrotasks
+    // and re-evaluate. Otherwise we'd sleep on cdp_link_active for up
+    // to timeout_ms while pending JS work sits idle.
+    if (try self.processMessages()) {
+      return;
+    }
 
     // Poll for HTTP I/O. The Network thread will call curl_multi_wakeup
     // on our multi handle whenever it pushes to our inbox, so we drop
--- a/src/browser/HttpClient.zig
+++ b/src/browser/HttpClient.zig
@@ -713,7 +713,7 @@ fn drainInbox(self: *Client, mode: DrainMode) !void {
             .cdp => |*c| cdp.onMessage(c) catch |err| {
                 // A single malformed/failed dispatch shouldn't poison
                 // the rest of the batch — log and continue.
-                log.warn(.cdp, "CDP dispatch", .{ .err = err });
+                log.err(.cdp, "CDP dispatch", .{ .err = err });
             },
             .ping => |body| cdp.onPing(body),
             .close => {
--- a/src/browser/HttpClient.zig
+++ b/src/browser/HttpClient.zig
@@ -31,6 +31,8 @@ const http = @import("../network/http.zig");
 const Robots = @import("../network/Robots.zig");
 const Network = @import("../network/Network.zig");
 
+const CDP = @import("../cdp/CDP.zig");
+const Inbox = @import("../Inbox.zig");
 const CachedResponse = @import("../network/cache/Cache.zig").CachedResponse;
 
 pub const CacheLayer = @import("../network/layer/CacheLayer.zig");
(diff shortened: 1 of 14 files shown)


src/browser/HttpClient.zig
```diff
@@ -127,7 +129,29 @@ obey_robots: bool,
...
 user_agent_header_override: ?[:0]const u8 = null,
 
-cdp_client: ?CDPClient = null,
+// The CDP layer we dispatch inbox messages to. Set in CDP.init for
+// `serve` mode; null in all other modes. Since this is set early, BEFORE the
+// CDP socket is registered with the network thread, we also have the
+// `cdp_link_active` boolean.
+cdp: ?*CDP = null,
+
+// True iff a producer (Server.handleConnection, after the worker
```
krichprollsch, line 138:
```suggestion
// True if a producer (Server.handleConnection, after the worker
```

src/browser/HttpClient.zig
```diff
@@ -659,29 +667,82 @@ fn perform(self: *Client, timeout_ms: c_int) anyerror!PerformStatus {
...
+            .sync_wait => self.inbox.popIf(allowDuringSyncWait),
+        } orelse return;
+
+        defer msg.deinit(self.arena_pool);
+
+        switch (msg.payload) {
+            .cdp => |*c| cdp.onMessage(c) catch |err| {
+                // A single malformed/failed dispatch shouldn't poison
+                // the rest of the batch — log and continue.
+                log.warn(.cdp, "CDP dispatch", .{ .err = err });
```
krichprollsch: is `warn` enough? shouldn't be `err`?

src/network/Network.zig
```diff
@@ -420,6 +494,195 @@ pub fn fireTicks(self: *Network) void {
...
+}
+
+// Synchronous teardown. Blocks the caller until this thread has
+// dropped the link from its poll set and won't invoke any of the
+// link's callbacks. Safe to call after Network has already dropped
+// the link unsolicited (state == .removed) — returns immediately in
+// that case.
+pub fn unregisterCdp(self: *Network, link: *CdpLink) void {
+    self.cdp_mutex.lock();
+    defer self.cdp_mutex.unlock();
```
krichprollsch, line 515: you hold the mutex during the whole function call here?

src/network/Network.zig
```diff
@@ -420,6 +494,195 @@ pub fn fireTicks(self: *Network) void {
...
+    }
+
+    while (link.state != .removed) {
+        // condition variable, waiting for a signal
+        self.cdp_unregister.wait(&self.cdp_mutex);
+    }
+}
+
+// Drop a link from the poll set. Caller must hold cdp_mutex.
+//   - on_disconnect is fired iff `notify` is true. Set notify=false
```
krichprollsch, line 529:
```suggestion
//   - on_disconnect is fired if `notify` is true. Set notify=false
```

src/browser/HttpClient.zig
karlseguin: Not sure if you're correcting a typo or the logic.

But iff = if and only if. I believe it's correct here because it's true only when the CDP socket has been registered with the network, and when it has been registered with the network, it's always true.

src/browser/HttpClient.zig
karlseguin: Thanks. My mental model is wrong.

My threshold for error/fatal is: can *WE* do anything about it? Like if we host a version of lightpanda, can we do anything about this error? Or if a user open a bug report and opens an issue, would *WE* want to see this? It's why we never log page JavaScript errors or network errors fetching. It just pollutes the logs with errors that we can't do anything about.

But that's a very SaaS way to look at it. If you're running your own browser, than YES, you probably DO want to see this error because it IS actionable.

src/network/Network.zig
karlseguin: `cdp_unregister` is a condition variable. The mutex gets released when `wait` is entered and re-acquired before it returns.

src/network/Network.zig
karlseguin: Same as above.

src/browser/HttpClient.zig
krichprollsch: oh ok, my bad :man_facepalming:

krichprollsch (approved):

PR #2501, opened 2026-05-20
Remove reentrency teardown protection

karlseguin:
This largely reverts 92607ad7650f19ab73a3b101239b6be35b10bf03 (captured in PR: https://github.com/lightpanda-io/browser/pull/2398).

https://github.com/lightpanda-io/browser/pull/2495 introduces protection against execution arbitrary CDP command during JavaScript callbacks. Claude initially made the case for keeping the existing code as a safety net, but sycophanted when I pushed by.

My reason for removing it is that it isn't a low-maintenance guard. It's a flag that serves a real purpose (ensuring 1 JS script is finished before executing another one), that has been expended to solve these issues. It needs to be set (and reverted) at every callsite that makes a blocking call, and it needs to be checked (recursively across all frames) in any place that can teardown the page/ frame.

Claude called the allowlist "load-bearing in a non-obvious way", but I think it's purpose built specifically for this case. Extended the comment atop `allowDuringSyncWait` so that future-selves remember this.

(Also removed some dead code that 2495 missed)

### DIFF
--- a/src/browser/Frame.zig
+++ b/src/browser/Frame.zig
@@ -1256,24 +1256,6 @@ pub fn isGoingAway(self: *const Frame) bool {
     return parent.isGoingAway();
 }
 
-// True if this frame, any descendant frame, or any worker owned by any
-// of those frames is currently inside script evaluation. Used as a
-// reentrancy guard before tearing down a page from a CDP message that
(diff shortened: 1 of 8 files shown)


karlseguin:
FYI @staylor

krichprollsch (approved):

PR #2505, opened 2026-05-20
Cleanup Transfer flag

karlseguin:
Replaces 4 boolean flags with a state. Makes it easier to figure out what the state of the transfer is, and removes the possibility of inconsistent flags .e.g queued + loop_owned.

loop_owned -> state != .created
_queued -> state == .queued
_perform -> state == .completing
aborted -> state == .aborted

### DIFF
--- a/src/browser/HttpClient.zig
+++ b/src/browser/HttpClient.zig
@@ -419,10 +419,16 @@ fn drainQueue(self: *Client) !void {
             self.queue.prepend(queue_node);
             return;
         };
-        // Cleared only after we've successfully obtained a connection;
-        // if we put the node back, _queued stays true.
-        transfer._queued = false;
-        try self.makeRequest(conn, transfer);
+        // Bridge state to .created so a failure inside makeRequest before
+        // any commit cleans up via the abort below. makeRequest flips to
+        // .inflight on a successful trackConn.
+        transfer.state = .created;
+        self.makeRequest(conn, transfer) catch |err| {
+            if (transfer.state == .created) {
+                transfer.abort(err);
+            }
+            return err;
+        };
     }
 }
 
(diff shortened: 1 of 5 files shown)


src/browser/HttpClient.zig
```diff
@@ -1153,67 +1167,120 @@ pub const Transfer = struct {
...
+        self.state = .{ .parked = by };
+    }
+
+    // Layer-facing: take the transfer out of .parked and return it to
+    // the request flow (.created). Caller is expected to re-enter the
+    // layer chain or call abort(); if neither happens the transfer is
+    // stranded.
+    pub fn unpark(self: *Transfer) void {
+        lp.assert(self.state == .parked, "Transfer.unpark", .{ .state = self.state });
+        self.state = .created;
```
krichprollsch, line 1267: Is the next state always `.created`? Even if the transfer was park by `.intercept_response`?

src/browser/HttpClient.zig
```diff
@@ -1153,67 +1167,120 @@ pub const Transfer = struct {
...
+        aborted,
+    };
+
+    pub const ParkedBy = enum {
+        // CDP Fetch interception, request phase.
+        intercept_request,
+
+        // CDP Fetch interception, response phase. Reserved for when
+        // response interception lands; not currently emitted.
+        intercept_response,
```
krichprollsch: not used for now?

src/browser/HttpClient.zig
karlseguin: Right now, yes. It's meant for layer to "park" prior to sending it in-flight (`park` has an assertion that the state is == .created).

There is a case for post-parking: Request Interception where requestStage == "response". This is something we don't currently support.  But I think more has to change to support that (e.g. I don't think the transfer is generally safe to use post-perform right now). I added a note about this in `unpark`..which relates to your other comment about the unused `intercept_response` (which I removed)

krichprollsch (approved):

PR #2506, opened 2026-05-20
Serve Cache from Next Client Tick

mookums:


src/network/layer/CacheLayer.zig
```diff
@@ -72,11 +73,21 @@ fn request(ptr: *anyopaque, transfer: *Transfer) anyerror!void {
             &.{ .transfer = transfer },
         );
 
-        // Cache hit: serve synchronously from the original callbacks, then
-        // tear down. On error, the transfer is still alive and Client.request's
-        // errdefer will handle cleanup (loop_owned is still false).
-        try serveFromCache(req, &cached);
-        transfer.deinit();
+        const CacheServeCtx = struct {
```
karlseguin: Is there a reason to wrap this?

src/network/layer/CacheLayer.zig
```diff
@@ -72,11 +73,21 @@ fn request(ptr: *anyopaque, transfer: *Transfer) anyerror!void {
...
+            cached: CachedResponse,
+        };
+
+        const ctx = try arena.create(CacheServeCtx);
+        ctx.* = .{ .cached = cached };
+
+        try transfer.client.runNextTick(transfer.id, ctx, struct {
+            fn run(t: *Transfer, ctx_ptr: *anyopaque) anyerror!void {
```
karlseguin: `t` is leaked.

src/network/layer/CacheLayer.zig
```diff
@@ -72,11 +73,21 @@ fn request(ptr: *anyopaque, transfer: *Transfer) anyerror!void {
...
-        try serveFromCache(req, &cached);
-        transfer.deinit();
+        const CacheServeCtx = struct {
+            cached: CachedResponse,
+        };
+
+        const ctx = try arena.create(CacheServeCtx);
+        ctx.* = .{ .cached = cached };
+
+        try transfer.client.runNextTick(transfer.id, ctx, struct {
```
karlseguin: The `fd` is leaked if the transfer is externally aborted before the next tick (I think)

src/browser/HttpClient.zig
```diff
@@ -412,6 +435,42 @@ pub fn tick(self: *Client, timeout_ms: u32, mode: DrainMode) !void {
     try self.drainInbox(mode);
 }
 
+pub fn runNextTick(self: *Client, transfer_id: u32, ctx: *anyopaque, run: NextTickNode.Run) !void {
+    const node = try self.allocator.create(NextTickNode);
```
karlseguin: Might be worth having a MemoryPool on the HttpClient for these. Not a huge deal though

src/browser/HttpClient.zig
```diff
@@ -412,6 +435,42 @@ pub fn tick(self: *Client, timeout_ms: u32, mode: DrainMode) !void {
...
+    var queue = self.next_tick_queue;
+    self.next_tick_queue = .{};
+
+    while (queue.popFirst()) |node| {
+        const n: *NextTickNode = @fieldParentPtr("node", node);
+        defer self.allocator.destroy(n);
+        defer self.next_tick_count -= 1;
+
+        if (self.findTransfer(n.transfer_id)) |t| {
+            try n.run(t, n.ctx);
```
karlseguin: If this fails, everything after it in the list is lost.

I was going to comment that, if you're going to do `self.next_tick_queue = .{};` then you can avoid the overhead of `popFirst()` and just walk the list without removing it.

BUT..if you keep `popFirst` and remove `self.next_tick_queue = .{};` then the 'everything is lost' is fixed.

The only issue is if drainNextTickQueue is called while `drainNextTickQueue` is running. Your current code handles that fine, at the cost of leaking on error.

Not sure if there's a 3rd solution that has neither drawbacks.

Also, related, but if this fails with something after it, `next_tick_queue` is empty, but `next_tick_count > 0`

src/browser/HttpClient.zig
```diff
@@ -1220,6 +1279,8 @@ pub const Transfer = struct {
         // Any concurrent CDP lookup by id will now see this transfer as gone.
         _ = self.client.transfers.remove(self.id);
 
+        self.client.cancelNextTick(self.id);
```
karlseguin: It's a shame we have to walk the list on _every_ transfer deinit.

`cancelNextTick` could short-circuit if `next_tick_count == 0`.  Even better, store `tick_node: ?std.DoublyLinkedList.Node` on `Transfer`.

In fact, I'd go even further and store everything you need for this on Transfer, avoiding allocations, the transfer_id lookup, the transfer.deinit list walk.

navidemad:
> [!NOTE]
> Not re-reviewing — @karlseguin already covered the substance. I verified his
> findings against the code (they all hold) and have two things to add: the exact
> leak mechanisms, and a concrete answer to the open `drainNextTickQueue`
> question. Codex also flagged the `t` leak independently.

## Confirming the two leaks, with the precise mechanism

**`t` is leaked (CacheLayer.zig:84)** — the old path did `serveFromCache(...);
transfer.deinit();`. The new `run` closure serves but never deinits, on *both*
the success and `catch` branches, so the transfer stays in `client.transfers` +
owner lists and its arena is never released. Codex independently called this out
(it trips the debug leak checks — one leaked transfer per cache hit).

**`fd` is leaked on abort (CacheLayer.zig:83)** — confirmed, and worth pinning
down *why* it's specifically the fd: `serveFromCache` closes the file in its
`defer` (CacheLayer.zig:125-128), so the fd is released whenever `run()` actually
executes. The ctx lives in `transfer.arena` (lines 60, 80), so when the transfer
is aborted first, the struct memory is reclaimed with the arena — but nothing
closes the fd, because `cancelNextTick` only `destroy`s the node. So it's a pure
fd leak, not a memory leak.

## These aren't four separate bugs — they share one root

`NextTickNode` has a `run` but no `cancel`/cleanup, and the ctx's resources
(the transfer, the cached fd) have no owner across the three ways a scheduled
node can end:

```mermaid
flowchart TD
    S["runNextTick: node scheduled<br/>ctx = CacheServeCtx { cached(fd) } in transfer.arena"] --> O{"how it ends"}
    O -->|"tick drains it"| D["run() → serveFromCache<br/>✓ closes fd in defer<br/>✗ never deinits transfer → t leaks"]
    O -->|"transfer aborted first"| C["cancelNextTick: destroy(node)<br/>✗ run() never runs → fd never closed<br/>(struct itself reclaimed via arena)"]
    O -->|"client.deinit"| K["deinit drain: destroy(node)<br/>✗ run() never runs → fd leaks"]
```

That's why your HttpClient.zig:1282 suggestion — *store the state on `Transfer`*
— is the strongest of the options on the table: it makes `Transfer.deinit` the
single owner, which collapses the `t` leak (deinit tears it down), the fd leak
(deinit closes the fd on every path), the O(N) `cancelNextTick` walk (→ O(1)),
*and* the per-node allocation into one fix. The MemoryPool idea (HttpClient.zig:439)
only addresses the allocation; this addresses all four.

## Re: `drainNextTickQueue` — a 3rd option (HttpClient.zig:469)

You asked for a solution with neither the "lost on error" nor the re-entrancy
drawback. Drain *all* nodes and surface the first error at the end, keeping the
swap for re-entrancy safety:

```zig
fn drainNextTickQueue(self: *Client) !void {
    var queue = self.next_tick_queue;
    self.next_tick_queue = .{};

    var first_err: ?anyerror = null;
    while (queue.popFirst()) |node| {
        const n: *NextTickNode = @fieldParentPtr("node", node);
        defer self.allocator.destroy(n);
        defer self.next_tick_count -= 1;

        if (self.findTransfer(n.transfer_id)) |t| {
            n.run(t, n.ctx) catch |err| {
                if (first_err == null) first_err = err;
            };
        }
    }
    if (first_err) |err| return err;
}
```

Every node is popped + destroyed (no leak), `next_tick_count` lands at 0 (no
inconsistency), the swap still isolates re-entrant scheduling, and the error
still propagates. The only semantic shift is that one failing task no longer
skips the others — which is what you'd want for independent transfers anyway.
(Today it's moot regardless: the cache `run` closure swallows its own error and
returns void, so `try n.run` never actually fails — but the `Run` signature
invites it, so hardening the drain is worthwhile.)

One more for the same root: the fd also leaks in this function's
`findTransfer == null` branch and in the `deinit` drain — both `destroy` the node
without running it. A node-level cleanup hook (or the store-on-`Transfer`
approach) closes all of these at once.

## Tiny

Re your CacheLayer.zig:76 question — no real reason to wrap; since `CacheServeCtx`
is a single `cached` field, `arena.create(CachedResponse)` + `ctx.* = cached`
works directly.

src/browser/HttpClient.zig
navidemad: A 3rd option that avoids both drawbacks: keep the swap (so re-entrancy is still isolated), but drain *every* node and surface the first error at the end instead of `try`-ing inside the loop:

```zig
fn drainNextTickQueue(self: *Client) !void {
    var queue = self.next_tick_queue;
    self.next_tick_queue = .{};

    var first_err: ?anyerror = null;
    while (queue.popFirst()) |node| {
        const n: *NextTickNode = @fieldParentPtr("node", node);
        defer self.allocator.destroy(n);
        defer self.next_tick_count -= 1;

        if (self.findTransfer(n.transfer_id)) |t| {
            n.run(t, n.ctx) catch |err| {
                if (first_err == null) first_err = err;
            };
        }
    }
    if (first_err) |err| return err;
}
```

Every node is popped + destroyed (nothing lost), `next_tick_count` lands at 0 (no `count > 0` while empty), the swap still isolates anything scheduled during the drain, and the error still propagates. The only behavioural shift is that one failing task no longer skips the rest — which is what you'd want for independent transfers anyway.

Worth noting it's moot for the current sole caller: the cache `run` closure catches its own error and returns void, so `try n.run` never actually fails today. But the `Run` signature allows it, so hardening here is cheap insurance.

src/browser/HttpClient.zig
mookums: `NextTickNode` is now owned by Transfer, seems like a nice solution.

src/browser/HttpClient.zig
mookums: Removed erroring from the `NextTickNode.Run` method. It should really be handled within the `Run` callback because that is also where it is expected for the `Transfer` to be deinit'd. Means that is should be fine to have the swap and the iterating through the detached queue.

src/browser/HttpClient.zig
```diff
@@ -416,6 +442,48 @@ pub fn tick(self: *Client, timeout_ms: u32, mode: DrainMode) !void {
...
+    ctx: *anyopaque,
+    params: struct { run: NextTickNode.Run, abort: ?NextTickNode.Abort = null },
+) !void {
+    transfer._next_tick_node = .{ .ctx = ctx, .run = params.run, .abort = params.abort };
+
+    self.next_tick_count += 1;
+    self.next_tick_queue.append(&transfer._next_tick_node.?.node);
+}
+
+fn cancelNextTick(self: *Client, transfer: *Transfer) void {
```
karlseguin, line 456: This should decrement `next_tick_count` ?

src/browser/HttpClient.zig
```diff
@@ -229,6 +244,16 @@ pub fn init(self: *Client, allocator: Allocator, network: *Network, cdp: ?*CDP)
 
 pub fn deinit(self: *Client) void {
     self.abort();
+
+    // Drain Next Tick Queue
+    while (self.next_tick_queue.popFirst()) |node| {
```
karlseguin: Is this necessary? Wasn't transfer.deinit called in the `self.abort()` which would have aborted its own next tick?

src/browser/HttpClient.zig
```diff
@@ -416,6 +442,48 @@ pub fn tick(self: *Client, timeout_ms: u32, mode: DrainMode) !void {
...
+
+fn drainNextTickQueue(self: *Client) !void {
+    var current = self.next_tick_queue.first;
+    self.next_tick_queue = .{};
+
+    while (current) |node| {
+        defer self.next_tick_count -= 1;
+
+        const n: *NextTickNode = @fieldParentPtr("node", node);
+        current = node.next;
```
karlseguin: The `ntn.run` a few lines down, can invalidate this `current`. `run` calls doneCallback which could result in page navigation / teardown. `abortFrame` is called, and all transfers are deinit'd. 

Don't see how you avoid a popFirst() drain, which would solve the issue as since`cancelNextTick` updates `self.next_tick_queue`.

mookums: Changed this so it uses the  `next_tick_count` to limit and then uses `popFirst` to iterate through.

PR #2509, opened 2026-05-20
Fix CDP server stall/SIGTERM hang in optimized builds (Network drops CDP sockets from poll set)

### LINKED ISSUE #2507: CDP server ignores SIGTERM after serving a connection (regression: 6323 OK → 6353 broken)
## Summary

Once `lightpanda serve` has accepted a single CDP WebSocket connection, the process **no longer terminates on `SIGTERM`** — it must be `SIGKILL`ed. An **idle** server (never connected) still exits on `SIGTERM` normally. This is a regression: the current published nightly **6323 is clean**, a `main` build at **6353 is broken**.

## Impact

Any CDP client that stops the browser with `SIGTERM` (the conventional graceful stop — Puppeteer, Playwright, chromedp, custom drivers) hangs waiting for the process to exit after it has driven a session. In the `capybara-lightpanda` driver this surfaced as a **~45-minute hang at test-suite teardown** (the driver sends `SIGTERM` then `waitpid`). We're adding a `SIGKILL` fallback on our side, but the browser ignoring `SIGTERM` is the root cause.

## Environment

- macOS aarch64 (signal handling here is platform-independent; expected on Linux too)
- Broken: `1.0.0-dev.6353+f1b0adf9` (built from `main`)
- Clean:  `1.0.0-nightly.6323+037db695` (current published nightly)

## Reproduction

Self-contained — only `bash` + `python3` (stdlib); no driver/gem involved:

```bash
#!/usr/bin/env bash
# Repro: lightpanda ignores SIGTERM once it has served a CDP connection.
set -u
BIN="${1:?usage: $0 /path/to/lightpanda}"
echo "binary: $("$BIN" version 2>&1)"

state() { # alive | zombie | gone   (zombie/gone = it exited)
  s=$(ps -o stat= -p "$1" 2>/dev/null | tr -d ' ')
  if   [ -z "$s" ];        then echo gone
  elif [ "${s:0:1}" = Z ]; then echo zombie
  else                          echo alive
  fi
}

cdp_touch() { # open one CDP WebSocket to 127.0.0.1:$1 and send one command
  python3 - "$1" <<'PY'
import socket, base64, os, sys
port = int(sys.argv[1])
s = socket.create_connection(("127.0.0.1", port), timeout=5)
key = base64.b64encode(os.urandom(16)).decode()
s.sendall(("GET / HTTP/1.1\r\nHost: 127.0.0.1\r\nUpgrade: websocket\r\n"
           "Connection: Upgrade\r\nSec-WebSocket-Key: %s\r\n"
           "Sec-WebSocket-Version: 13\r\n\r\n" % key).encode())
hdr = b""
while b"\r\n\r\n" not in hdr:
    hdr += s.recv(1)
assert hdr.startswith(b"HTTP/1.1 101"), hdr[:40]
payload = b'{"id":1,"method":"Target.getTargets"}'
mask = os.urandom(4)
masked = bytes(b ^ mask[i % 4] for i, b in enumerate(payload))
s.sendall(bytes([0x81, 0x80 | len(payload)]) + mask + masked)
try:
    s.recv(256)
except Exception:
    pass
print("  (CDP WebSocket connected + Target.getTargets sent)")
PY
}

PORT=39911
run() { # $1=label  $2=connect?(0/1)
  "$BIN" serve --host 127.0.0.1 --port "$PORT" --log_level info >/dev/null 2>&1 &
  PID=$!
  sleep 2
  if [ "$(state "$PID")" != alive ]; then echo "[$1] failed to start"; PORT=$((PORT + 1)); return; fi
  [ "$2" = 1 ] && cdp_touch "$PORT"
  kill -TERM "$PID"
  for _ in $(seq 1 12); do [ "$(state "$PID")" = alive ] || break; sleep 1; done
  if [ "$(state "$PID")" = alive ]; then
    echo "[$1] SIGTERM IGNORED — still running 12s after SIGTERM"
    kill -KILL "$PID"; sleep 1
    echo "[$1]   -> SIGKILL: now $(state "$PID")"
  else
    echo "[$1] exited on SIGTERM (ok)"
  fi
  wait "$PID" 2>/dev/null
  PORT=$((PORT + 1))
}

run "idle (no CDP connection)" 0
run "used (one CDP connection)" 1
```

### Output

**Broken — build 6353:**

```
binary: 1.0.0-dev.6353+f1b0adf9
[idle (no CDP connection)] exited on SIGTERM (ok)
[used (one CDP connection)] SIGTERM IGNORED — still running 12s after SIGTERM
[used (one CDP connection)]   -> SIGKILL: now gone
```

**Clean — build 6323:**

```
binary: 1.0.0-nightly.6323+037db695
[idle (no CDP connection)] exited on SIGTERM (ok)
[used (one CDP connection)] exited on SIGTERM (ok)
```

(Downstream the "used" process was observed alive 45+ minutes after `SIGTERM`, so this is a true hang, not slow shutdown.)

## Expected vs actual

- **Expected:** `SIGTERM` terminates the process whether or not a CDP client has connected (as at 6323, and as the idle case still does at 6353).
- **Actual (6353):** after ≥1 CDP connection, `SIGTERM` is ignored; only `SIGKILL` stops it.

## Regression window / suspect

`git merge-base` over `main`:

- Clean at build **6323** = `037db695` (#2492 "Re-organization CDP connection", rename-only).
- Broken at build **6353** = `f1b0adf9`.
- Prime suspect: **#2495 "Main/Network reads CDP socket"** (`875c1477`, ~build 6331) — it moved CDP socket ownership/reading into the main/network loop; it is in the broken window and absent at 6323. (Suspected, not bisect-confirmed.)

Reads like a CDP socket-reader loop/thread that blocks and isn't interrupted by the shutdown signal once a connection exists. Happy to bisect or test a patch.

### LINKED ISSUE #2508: PR #2495 introduces CDP hang in release builds (Playwright/Puppeteer stop responding after first messages)
## Summary

PR #2495 (`cdp_inbox`, merged as `a314984b`) introduces a CDP regression that **only manifests in optimized builds** (`-Doptimize=ReleaseFast` and `-Doptimize=ReleaseSafe`). Debug builds work normally.

After a CDP client connects and exchanges the first one or two browser-session messages, the server stops responding to further browser-session commands. The reader/inbox path appears to silently stall — no error is logged, no panic fires, the process keeps running but never answers another command.

Because the nightly release build uses `-Doptimize=ReleaseFast`, the **next nightly that picks up #2495 will be unusable for Playwright and Puppeteer**.

## Repro

```bash
git checkout a314984b   # or current main (same result)

# Exact nightly recipe (.github/workflows/release.yml):
zig build -Dprebuilt_v8_path=v8/libc_v8.a -Doptimize=ReleaseFast snapshot_creator -- src/snapshot.bin
zig build -Dprebuilt_v8_path=v8/libc_v8.a -Doptimize=ReleaseFast -Dsnapshot_path=../../snapshot.bin -Dversion=nightly

./zig-out/bin/lightpanda serve --host 127.0.0.1 --port 9222 --log-level info &
URL=https://example.com node cdp-repros/repro-puppeteer.mjs
# hangs forever after "[pup] connect ws://127.0.0.1:9222"
```

Playwright via `connectOverCDP` shows the same hang. CDP traffic captured with `DEBUG=pw:protocol` (3 runs, behaviour is non-deterministic on which command stalls):

**Run A:** `Browser.getVersion` (id=1) ✓ → `Target.setAutoAttach` (id=2) ✗ → never responds.

**Run B:** `Browser.getVersion` (id=1) ✓ → `Target.setAutoAttach` (id=2) ✓ → `Browser.setDownloadBehavior` (id=3) ✓ → all subsequent `Page.*` / `Runtime.*` sent on the `STARTUP` session are dropped.

In both shapes, only **browser-session commands** (no `sessionId`) make it through reliably; once the inbox stalls, all further messages are ignored. No log output is emitted by the server (info level shows literally just `client connected` plus the SIGTERM line, or nothing).

## Verified working / not working

| Commit | Mode | Result |
|---|---|---|
| `29d5a7ae` (parent of #2495 merge) | ReleaseFast | ✅ Puppeteer + example.com loads in 134 ms |
| `a314984b` (#2495 merge) | ReleaseFast | ❌ deterministic hang (6/6 attempts) |
| `a314984b` | ReleaseSafe | ❌ same hang, no panic emitted |
| `a314984b` | Debug | ✅ Playwright + allbirds.com loads in 10.8 s |
| `5905319e` (current main HEAD) | ReleaseFast | ❌ same hang |
| `5905319e` | Debug | ✅ works |

Bisect points squarely at PR #2495.

## Hypothesis

Almost certainly a startup race in the new dedicated reader/inbox thread introduced by #2495 (`Re-organization CDP connection` + `Main/Network reads CDP socket` + commit `97c8ca38` "when work is done, don't keep polling, return to process it"). Debug-mode timing (slower code, asserts, log overhead) masks the race; optimized builds expose it.

Non-determinism between runs (sometimes id=2 stalls, sometimes id=3 stalls) is consistent with a missed wake-up / signal-already-consumed race between the reader thread and the worker thread around the wake pipe or condition variable.

The fact that **no debug logs appear at all** in the release-build run — not even from before the hang — also suggests the worker is never processing messages off the inbox (or the inbox handoff itself is the broken edge), since logs in CDP message handling never fire.

## Environment

- macOS 26, arm64 (Apple Silicon)
- Zig 0.15.2
- V8 14.0.365.4 (using the `release_4fa7ab4e` prebuilt)
- playwright-core 1.59.1, puppeteer-core 24.42.0

## Impact

- Next nightly will be broken for any CDP client (Playwright, Puppeteer, raw CDP).
- Suggest reverting #2495 from the next nightly tag, or holding the nightly cron, until a fix lands.

## Repro scripts in the repo

`cdp-repros/repro-puppeteer.mjs` + `cdp-repros/smoke-newctx.mjs` both reproduce. `repro-puppeteer.mjs` against `https://example.com` is the simplest:

```js
const ctx = await browser.createBrowserContext();
const resp = await page.goto('https://example.com', { waitUntil: 'load', timeout: 30000 });
// release-fast build of #2495+ never reaches here
```

navidemad:
## What this fixes

Two regressions from #2495, both reproducing **only in optimized builds** (`-Doptimize=ReleaseFast`/`ReleaseSafe`, i.e. the nightly):

- **#2508** — after the first one or two CDP commands the server stops answering. In a minimal repro `Browser.getVersion` (id=1) succeeds, then `Target.setAutoAttach` (id=2) never gets a reply — a hard stall (no recovery, no logs). This makes the next nightly unusable for Playwright/Puppeteer.
- **#2507** — once a CDP connection has been served, the process ignores `SIGTERM` and must be `SIGKILL`ed (20/20 in a release build).

Both have a single root cause.

## Root cause

`preparePollFds` (`src/network/Network.zig`) rebuilds the curl portion of `pollfds` each loop iteration, but sliced to the **end** of the array:

```zig
const curl_fds = self.pollfds[PSEUDO_POLLFDS..];   // runs into the CDP region
@memset(curl_fds, .{ .fd = -1, .events = 0, .revents = 0 });
```

`pollfds` is three contiguous regions:

```
[0]                            wakeup pipe
[1]                            listener
[PSEUDO_POLLFDS .. cdp_start]  curl multi fds   (httpMaxConcurrent)
[cdp_start .. end]             CDP socket fds   (maxConnections)   <-- wiped every iteration
```

`prepareCdpPollFds` owns `[cdp_start..]` and only rebuilds it when `cdp_dirty` is set (a steady-state optimization). So the `@memset` reset every live CDP socket fd to `-1` on each iteration and they were never re-added — the Network thread silently stopped polling them.

This only triggers once Network owns a curl `multi` handle. **The only caller of `network.submitRequest` is telemetry** (`telemetry/lightpanda.zig:108`), and **telemetry is disabled in Debug builds** (`telemetry/telemetry.zig:14`). Regular HTTP/navigation runs on the *worker's* handles, not Network's multi. That's why Debug always worked and only optimized builds broke — it's telemetry creating the multi, not the optimizer.

With the CDP socket unwatched:
- Network never reads further client messages → **#2508**.
- Network never sees the client's EOF / `Connection.shutdown` → the worker (blocked in `curl_multi_poll`) is never told to disconnect, so it loops forever and `Server.deinit` waits on it → **#2507**.

`lldb` on a stalled process confirmed it: the Network thread parked in `poll(timeout=-1)` while id=2 sat unread on the socket.

## What this PR changes

```zig
// preparePollFds: clear/rebuild ONLY the curl region, never the CDP region
const curl_fds = self.pollfds[PSEUDO_POLLFDS..self.cdp_start];
```

Plus a defensive timeout fix in the same loop: `curl_multi_timeout` returns `-1` when curl is idle, so `@min(250, curl_timeout)` was `@min(250, -1) = -1` — `poll()` blocked forever. That starved `onTick` (telemetry's periodic flush) and turned any missed wakeup into a permanent hang. Treat `curl_timeout <= 0` as "no curl deadline" and fall back to the 250 ms cap.

## Why these constants

- **`cdp_start` boundary** — `httpMaxConcurrent` (default 10) is the designed size of the curl region, and Network's multi only ever serves telemetry (a single request), so the region has ample headroom.
- **250 ms** — the existing `min_timeout`; it bounds onTick latency and now also bounds any missed wakeup, so this whole class of bug can no longer hard-hang.

## Where to focus review

- The `preparePollFds` slice boundary, and that the `cdp_dirty` reuse optimization in `prepareCdpPollFds` is now sound again (nothing else touches `[cdp_start..]` between rebuilds).

## Test plan

- [x] New unit test `Network: preparePollFds leaves the CDP fd region untouched` — confirmed it **fails** on the old slice and **passes** on the fix.
- [x] `make test` — 718/718 pass.
- [x] `zig fmt --check` clean.
- [x] #2507 repro on a ReleaseFast build: **0/25** hang (was 20/20).
- [x] #2508 repro on a ReleaseFast build: full Puppeteer-style startup sequence completes (was stalling at id=2).

Fixes #2507
Fixes #2508

### DIFF
--- a/src/network/Network.zig
+++ b/src/network/Network.zig
@@ -722,8 +722,15 @@ pub fn run(self: *Network) void {
                 break :blk min_timeout;
             }
 
+            // curl_multi_timeout reports -1 when curl has no timeout
+            // preference (idle) and 0 when it wants to be serviced
+            // immediately. Treat both as "no curl-imposed deadline" and
+            // fall back to min_timeout — otherwise @min(min_timeout, -1)
+            // would be -1, i.e. poll() blocks forever, starving onTick
+            // (telemetry's periodic flush) and removing the safety net
+            // that bounds any missed wakeup to min_timeout.
             const curl_timeout = self.getCurlTimeout();
-            if (curl_timeout == 0) {
+            if (curl_timeout <= 0) {
                 break :blk min_timeout;
             }
 
@@ -860,7 +867,16 @@ fn acceptConnections(self: *Network) void {
 }
 
 fn preparePollFds(self: *Network, multi: *libcurl.CurlM) void {
-    const curl_fds = self.pollfds[PSEUDO_POLLFDS..];
+    // Only the curl slice — NOT through to the end of pollfds. The CDP
+    // socket fds live in [cdp_start..] and are owned by
+    // prepareCdpPollFds, which only rebuilds them when cdp_dirty is set
+    // (a steady-state optimization). Slicing to the end here would
+    // @memset those fds to -1 every iteration once a multi exists (which
+    // happens as soon as telemetry sends its first request), silently
+    // dropping every live CDP socket from the poll set — Network then
+    // never reads another CDP message (#2508) nor observes peer
+    // EOF/shutdown (#2507).
+    const curl_fds = self.pollfds[PSEUDO_POLLFDS..self.cdp_start];
     @memset(curl_fds, .{ .fd = -1, .events = 0, .revents = 0 });
 
     var fd_count: c_uint = 0;
@@ -1052,4 +1068,41 @@ fn loadCerts(allocator: Allocator) !libcurl.CurlBlob {
         .data = result.ptr,
         .flags = 0,
     };
+}
+
+const testing = @import("../testing.zig");
+
+test "Network: preparePollFds leaves the CDP fd region untouched" {
+    // Regression for #2507 / #2508. Once a multi exists (telemetry creates
+    // one in optimized builds), preparePollFds runs every loop iteration.
+    // It rebuilds only the curl slice [PSEUDO_POLLFDS..cdp_start]; the CDP
+    // region [cdp_start..] is owned by prepareCdpPollFds, which keeps its
+    // entries across iterations and only rebuilds when cdp_dirty is set.
+    // A slice that ran to the end of pollfds @memset those CDP sockets to
+    // -1, silently dropping every live CDP connection from the poll set —
+    // so Network stopped reading CDP messages (#2508) and never observed
+    // peer EOF/shutdown (#2507). curl global is initialized by the test
+    // harness (App.init -> Network.init).
+    const multi = libcurl.curl_multi_init() orelse return error.FailedToInitMulti;
+    defer libcurl.curl_multi_cleanup(multi) catch {};
+
+    const curl_slots = 4;
+    const cdp_slots = 3;
+    var pollfds: [PSEUDO_POLLFDS + curl_slots + cdp_slots]posix.pollfd = undefined;
+    @memset(&pollfds, .{ .fd = -1, .events = 0, .revents = 0 });
+
+    // preparePollFds only reads self.pollfds and self.cdp_start.
+    var nw: Network = undefined;
+    nw.pollfds = &pollfds;
+    nw.cdp_start = PSEUDO_POLLFDS + curl_slots;
+
+    // Two live CDP sockets parked in the CDP region, mimicking the steady
+    // state between cdp_dirty rebuilds.
+    pollfds[nw.cdp_start] = .{ .fd = 4242, .events = posix.POLL.IN, .revents = 0 };
+    pollfds[nw.cdp_start + 1] = .{ .fd = 4243, .events = posix.POLL.IN, .revents = 0 };
+
+    nw.preparePollFds(multi);
+
+    try testing.expectEqual(@as(posix.fd_t, 4242), pollfds[nw.cdp_start].fd);
+    try testing.expectEqual(@as(posix.fd_t, 4243), pollfds[nw.cdp_start + 1].fd);
 }


staylor:
Independently verified on macOS arm64 against a ReleaseFast build of `e1e49c8a` (this PR's head), matching the nightly recipe exactly (`zig build -Dprebuilt_v8_path=… -Doptimize=ReleaseFast -Dsnapshot_path=…`).

The root cause analysis is on the money — the Debug-works / Release-hangs split was the most confusing part of #2508 from the outside, and "telemetry creates the multi → preparePollFds clobbers CDP fds" explains it cleanly.

### #2508 — CDP stall

| Client | Page | Before #2509 (release) | After #2509 (release) |
|---|---|---|---|
| Puppeteer | example.com | ∞ hang at `[pup] connect` | ✅ 360 ms total, status=200, title + body |
| Playwright `connectOverCDP` + `newContext` | allbirds.com/products/mens-wool-runners | ∞ hang at `Target.setAutoAttach` (id=2) | ✅ DOMContentLoaded in 7.8 s, full title + 911 KB body, clean close |

Stress runs against allbirds (10 iterations each, ReleaseFast, `newContext`+`newPage` per iter):

- **No proxy:** 10/10 PASS, avg 7.9 s, server alive end-to-end.
- **`--http-proxy http://127.0.0.1:9999`** (a noop CONNECT-tunneling proxy): **10/10 PASS, avg 8.1 s, server alive.** Same matrix on post-#2495 main was 8/15 with the process crashing under proxy load.

### #2507 — SIGTERM hang

Send `SIGTERM` to the server right after a CDP client has connected + run:
- Before: process ignores SIGTERM, has to be SIGKILLed.
- After #2509: clean exit in **540 ms**.

### The 250 ms `curl_timeout <= 0` follow-up fix

Worth highlighting separately — `@min(250, curl_timeout)` with `curl_timeout == -1` collapsing to `-1` (block forever) is the kind of latent bug that's effectively invisible until something else changes the wake-up pattern. The `<= 0` guard plus the 250 ms floor turns "any missed wakeup → permanent hang" into "any missed wakeup → 250 ms latency", which is a real defence-in-depth win independent of the `pollfds` slice fix.

karlseguin:
Thanks, and sorry about that.

karlseguin:
To add context. The goal is to move all networking to the main thread and have a single multi (it can significantly reduce memory usage and provides better connection-reuse in multi-threaded scenarios). So Network.zig having an optional multi was always a temporary thing, and an optimization for when telemetry is disabled.

PR #2511, opened 2026-05-20
network: terminate live CDP connections on shutdown

### LINKED ISSUE #2510: CDP server ignores a single SIGTERM while a connection is live (distinct from #2507)
# CDP server ignores a single SIGTERM while a connection is live (distinct from #2507)

## Summary

`lightpanda serve` ignores a **single** `SIGTERM` whenever a CDP WebSocket
connection is **still open**. The process only terminates if the CDP socket is
closed (client EOF) *before* the signal, or if `SIGTERM` is sent **three** times.
A conventional one-shot graceful stop (`kill -TERM <pid>` then `waitpid`) hangs.

This is **separate from #2507 / #2509** (the telemetry curl-multi poll-fd bug):

- It reproduces with **telemetry disabled** (`LIGHTPANDA_DISABLE_TELEMETRY=true`),
  i.e. with no curl `multi` handle, so the #2509 `preparePollFds` fix does not
  touch it.
- It reproduces on a build that already includes #2509.
- No navigation or CDP command is required — a bare WebSocket upgrade is enough.

## Environment

- macOS aarch64 (signal handling is platform-independent; expected on Linux too)
- Reproduced on `1.0.0-dev.6353+f1b0adf9` **and** `1.0.0-dev.6354+e1e49c8a`
  (the latter = main + #2509), both `-Doptimize=ReleaseFast`.
- Not bisected; untested whether pre-#2495 (`6323`) also exhibits it.

## Reproduction (Lightpanda + raw CDP only)

```python
#!/usr/bin/env python3
# usage: repro.py /path/to/lightpanda
import socket, base64, os, sys, subprocess, time, signal

BIN = sys.argv[1]
print("binary:", subprocess.run([BIN, "version"], capture_output=True, text=True).stdout.strip())

def ws_connect(port):
    s = socket.create_connection(("127.0.0.1", port), timeout=5)
    key = base64.b64encode(os.urandom(16)).decode()
    s.sendall(("GET / HTTP/1.1\r\nHost: x\r\nUpgrade: websocket\r\nConnection: Upgrade\r\n"
               "Sec-WebSocket-Key: %s\r\nSec-WebSocket-Version: 13\r\n\r\n" % key).encode())
    buf = b""
    while b"\r\n\r\n" not in buf:
        buf += s.recv(1)
    assert buf.startswith(b"HTTP/1.1 101"), buf[:40]
    return s

def run(label, close_before_sigterm, port):
    env = dict(os.environ); env["LIGHTPANDA_DISABLE_TELEMETRY"] = "true"
    p = subprocess.Popen([BIN, "serve", "--host", "127.0.0.1", "--port", str(port), "--log_level", "error"],
                         stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, env=env, start_new_session=True)
    time.sleep(2)
    s = ws_connect(port)               # one CDP WebSocket; no commands sent
    if close_before_sigterm:
        s.close(); time.sleep(0.3)
    p.send_signal(signal.SIGTERM)      # exactly ONE SIGTERM
    deadline = time.time() + 12
    while time.time() < deadline:
        if p.poll() is not None:
            print("[%s] exited on SIGTERM (ok)" % label); break
        time.sleep(0.25)
    else:
        print("[%s] *** SIGTERM IGNORED — still alive 12s after one SIGTERM ***" % label)
        os.killpg(os.getpgid(p.pid), signal.SIGKILL)
    try: s.close()
    except OSError: pass
    try: p.wait(timeout=3)
    except Exception:
        try: os.killpg(os.getpgid(p.pid), signal.SIGKILL)
        except Exception: pass

run("WS closed before SIGTERM", True, 39920)
run("WS still open at SIGTERM", False, 39921)
```

### Output (identical on 6353 and on 6354 = main + #2509)

```
binary: 1.0.0-dev.6354+e1e49c8a
[WS closed before SIGTERM] exited on SIGTERM (ok)
[WS still open at SIGTERM] *** SIGTERM IGNORED — still alive 12s after one SIGTERM ***
```

## Broken flow

```mermaid
sequenceDiagram
    participant C as CDP client
    participant S as Sighandler thread (sigwait)
    participant M as Server/Network loop
    participant W as Connection worker (handleConnection)
    C->>M: WS upgrade → spawnWorker (active_threads=1)
    Note over C: connection left OPEN
    C->>S: SIGTERM (1st)
    S->>M: attempt=1 → run graceful listeners → Server.deinit()
    M->>W: Server.shutdown(): conn.shutdown()
    Note over W: not unwound while the CDP socket is open
    loop forever
        M->>M: while active_threads>0 { sleep 10ms }
    end
    Note over S: a 2nd SIGTERM re-runs listeners, only a 3rd does process.exit(1)
    Note over C,M: client sent ONE SIGTERM ⇒ HANG
```

## Expected flow

```mermaid
sequenceDiagram
    participant C as CDP client
    participant S as Sighandler thread (sigwait)
    participant M as Server/Network loop
    participant W as Connection worker
    C->>M: WS upgrade → spawnWorker (active_threads=1)
    C->>S: SIGTERM (1st)
    S->>M: run graceful listeners → Server.deinit()
    M->>W: Server.shutdown(): conn.shutdown() (interrupts the worker)
    W-->>M: worker returns (active_threads→0)
    M-->>C: process exits cleanly
```

## Root cause (pointers)

- `src/Sighandler.zig` `sighandle`: the 1st/2nd `SIGTERM` only run the graceful
  listeners; only the **3rd** (`attempt > 1`) calls `std.process.exit(1)`.
- `src/Server.zig` `deinit`: after `shutdown()` it spins
  `while (active_threads.load() > 0) sleep(10ms)`, waiting for each
  `handleConnection` worker to return.
- `Server.shutdown()` calls `conn.shutdown()` per live CDP connection, but that
  does not unwind a worker whose connection socket is still open, so
  `active_threads` never reaches 0 and the graceful path blocks indefinitely.

The single SIGTERM is therefore consumed by the graceful path that can't
complete, instead of terminating the process.

## Expected vs actual

- **Expected:** one `SIGTERM` terminates `lightpanda serve` whether or not a CDP
  connection is open (as already happens when the socket is closed first).
- **Actual:** with a live CDP connection, the first (and second) `SIGTERM` are
  absorbed; only `SIGKILL` or a third `SIGTERM` stops it.

## Suggested direction

On the first termination signal, force-close live CDP connections (and/or make
`conn.shutdown()` interrupt the worker's blocking wait) so `Server.deinit`'s
drain loop can complete — i.e. the first `SIGTERM` should always lead to exit.

## Why it matters

Any client that stops the browser with a single graceful `SIGTERM` (the
conventional path — Puppeteer/Playwright/chromedp/custom drivers) hangs after
driving a session. In `capybara-lightpanda` this surfaced as a multi-minute /
indefinite hang at suite teardown when process cleanup fell to a GC finalizer
that SIGTERMs without first closing the CDP socket.

navidemad:
## What

`lightpanda serve` ignored a single `SIGTERM` while a CDP connection was live — the process only exited if the socket was closed before the signal, or after a third signal. A conventional one-shot graceful stop (`kill -TERM <pid>` then `waitpid`) hung after the client had driven a session.

Closes #2510.

## Root cause

On shutdown the sighandler runs `Network.stop` (sets `shutdown`, lets the run loop exit) **before** `Server.shutdown`. A live CDP worker parks in `curl_multi_poll` and is woken **only** by the Network thread via `dropCdp -> handles.wakeup()`. Once the run loop exits with links still live, nothing can wake those workers, so `Server.deinit()`'s `while (active_threads > 0)` loop spins forever.

```mermaid
flowchart TD
    A1[SIGTERM] --> A2["Network.stop: shutdown = true"]
    A2 --> A3{"run loop: shutdown and running_handles == 0"}
    A3 -->|break| A4["Network thread exits<br/>(live CDP links abandoned)"]
    A4 --> A5["Server.shutdown: conn.shutdown()"]
    A5 --> A6["worker still parked in curl_multi_poll<br/>(nothing left to wake it)"]
    A6 --> A7["Server.deinit: while active_threads &gt; 0<br/>spins forever"]
```

## Change

`Network.run`, on observing `shutdown`, now calls a new `shutdownCdpLinks()` that drops every still-live CDP link via the existing peer-EOF path (`dropCdp(notify=true)` — pushes `.disconnect` into the worker's inbox and wakes it out of `curl_multi_poll`). The worker's `cdp.tick()` then returns false and it exits before the loop breaks. The helper is idempotent (a no-op once the links are drained).

```mermaid
flowchart TD
    B1[SIGTERM] --> B2["Network.stop: shutdown = true"]
    B2 --> B3["run loop: shutdownCdpLinks()<br/>dropCdp(notify=true) per live link"]
    B3 --> B4["worker woken: .disconnect in inbox<br/>+ curl_multi_wakeup"]
    B4 --> B5["cdp.tick() returns false<br/>worker exits, active_threads -&gt; 0"]
    B3 --> B6[run loop breaks, Network exits]
    B5 --> B7["Server.deinit returns cleanly"]
```

## Test plan

- [x] `zig build test` — 717/717 pass.
- [x] End-to-end (ReleaseFast build): the reproducer below — a raw CDP client that opens a WebSocket and sends **one** `SIGTERM` with the connection still open — now **exits cleanly**; before the fix it stayed alive 12s+ until `SIGKILL`. Closing the socket before `SIGTERM` already worked and still does.

No in-process unit test: the shutdown path is multi-threaded (sighandler thread → Network run loop → per-connection worker), and the without-fix failure mode is a `Server.deinit()` spin rather than an assertion. The link's lifetime is guarded by `unregisterCdp`'s synchronous handoff *on the Network thread*, so dropping links from any other thread (e.g. a test) risks a use-after-free on `cdp_poll_snapshot`. A race-free test would need a dedicated `Server`/`Network` instance plus a watchdog — happy to add that if you'd like it in the suite.

## Reproducer

Self-contained (Lightpanda + Python stdlib, no extra deps), telemetry disabled to show this is independent of #2509:

```python
import socket, base64, os, sys, subprocess, time, signal

BIN = sys.argv[1]
print("binary:", subprocess.run([BIN, "version"], capture_output=True, text=True).stdout.strip())

def ws_connect(port):
    s = socket.create_connection(("127.0.0.1", port), timeout=5)
    key = base64.b64encode(os.urandom(16)).decode()
    s.sendall(("GET / HTTP/1.1\r\nHost: x\r\nUpgrade: websocket\r\nConnection: Upgrade\r\n"
               "Sec-WebSocket-Key: %s\r\nSec-WebSocket-Version: 13\r\n\r\n" % key).encode())
    buf = b""
    while b"\r\n\r\n" not in buf:
        buf += s.recv(1)
    assert buf.startswith(b"HTTP/1.1 101"), buf[:40]
    return s

def run(label, close_before_sigterm, port):
    env = dict(os.environ); env["LIGHTPANDA_DISABLE_TELEMETRY"] = "true"
    p = subprocess.Popen([BIN, "serve", "--host", "127.0.0.1", "--port", str(port), "--log_level", "error"],
                         stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, env=env, start_new_session=True)
    time.sleep(2)
    s = ws_connect(port)               # one CDP WebSocket; no commands sent
    if close_before_sigterm:
        s.close(); time.sleep(0.3)
    p.send_signal(signal.SIGTERM)      # exactly ONE SIGTERM
    deadline = time.time() + 12
    while time.time() < deadline:
        if p.poll() is not None:
            print("[%s] exited on SIGTERM (ok)" % label); break
        time.sleep(0.25)
    else:
        print("[%s] *** SIGTERM IGNORED — still alive 12s after one SIGTERM ***" % label)
        os.killpg(os.getpgid(p.pid), signal.SIGKILL)
    try: s.close()
    except OSError: pass
    try: p.wait(timeout=3)
    except Exception:
        try: os.killpg(os.getpgid(p.pid), signal.SIGKILL)
        except Exception: pass

run("WS closed before SIGTERM", True, 39920)
run("WS still open at SIGTERM", False, 39921)
```

Before this PR the second case prints `SIGTERM IGNORED`; after it, both print `exited on SIGTERM (ok)`.

### DIFF
--- a/src/network/Network.zig
+++ b/src/network/Network.zig
@@ -697,12 +697,11 @@ fn shutdownCdpLinks(self: *Network) void {
 
     var it = self.cdp_links.first;
     while (it) |node| {
-        const next = node.next;
+        it = node.next;
         const link: *CdpLink = @fieldParentPtr("node", node);
         if (link.state == .live) {
             self.dropCdp(link, null, true);
         }
-        it = next;
     }
 
     self.cdp_unregister.broadcast();
--- a/src/network/Network.zig
+++ b/src/network/Network.zig
@@ -683,6 +683,35 @@ fn processCdpEvents(self: *Network) void {
     }
 }
 
+// On shutdown, force-disconnect every still-live CDP link. Each link's
+// worker thread blocks in curl_multi_poll and is woken ONLY by this
+// (Network) thread via dropCdp -> handles.wakeup(). If the run loop
+// exits with links still live, those workers never wake and
+// Server.deinit() spins on active_threads forever (issue #2510).
+// Mirrors the peer-EOF path in processCdpEvents: dropCdp(notify=true)
+// pushes a .disconnect into the worker's inbox and wakes it, so
+// cdp.tick() returns false and the worker exits.
+fn shutdownCdpLinks(self: *Network) void {
+    self.cdp_mutex.lock();
+    defer self.cdp_mutex.unlock();
+
+    var any_removed = false;
+    var it = self.cdp_links.first;
+    while (it) |node| {
+        const next = node.next;
+        const link: *CdpLink = @fieldParentPtr("node", node);
+        if (link.state == .live) {
+            self.dropCdp(link, null, true);
+            any_removed = true;
+        }
+        it = next;
+    }
+
+    if (any_removed) {
+        self.cdp_unregister.broadcast();
+    }
+}
+
 pub fn run(self: *Network) void {
     var drain_buf: [64]u8 = undefined;
     var running_handles: c_int = 0;
@@ -760,13 +789,24 @@ pub fn run(self: *Network) void {
 
         self.fireTicks();
 
-        if (self.shutdown.load(.acquire) and running_handles == 0) {
-            // Check if fireTicks submitted new requests (e.g. telemetry flush).
-            // If so, continue the loop to drain and send them before exiting.
-            self.submission_mutex.lock();
-            const has_pending = self.submission_queue.first != null;
-            self.submission_mutex.unlock();
-            if (!has_pending) break;
+        if (self.shutdown.load(.acquire)) {
+            // Force-disconnect any still-live CDP clients so their worker
+            // threads wake from curl_multi_poll and exit. A worker is woken
+            // only by this (Network) thread; if the loop exits with links
+            // still live, those workers block forever and Server.deinit()
+            // spins on active_threads (issue #2510). Idempotent — a no-op
+            // once the links are drained.
+            self.shutdownCdpLinks();
+
+            if (running_handles == 0) {
+                // Check if fireTicks submitted new requests (e.g. telemetry
+                // flush). If so, continue the loop to drain and send them
+                // before exiting.
+                self.submission_mutex.lock();
+                const has_pending = self.submission_queue.first != null;
+                self.submission_mutex.unlock();
+                if (!has_pending) break;
+            }
         }
     }
 
--- a/src/network/Network.zig
+++ b/src/network/Network.zig
@@ -787,7 +787,7 @@ pub fn run(self: *Network) void {
 
         if (self.shutdown.load(.acquire)) {
             // Drain any live CDP links so their workers can exit (issue #2510).
-            // Idempo tent — no-op once drained, safe to call every iteration
+            // Idempotent — no-op once drained, safe to call every iteration
             self.shutdownCdpLinks();
 
             if (running_handles == 0) {
--- a/src/network/Network.zig
+++ b/src/network/Network.zig
@@ -695,21 +695,17 @@ fn shutdownCdpLinks(self: *Network) void {
     self.cdp_mutex.lock();
     defer self.cdp_mutex.unlock();
 
-    var any_removed = false;
     var it = self.cdp_links.first;
     while (it) |node| {
         const next = node.next;
         const link: *CdpLink = @fieldParentPtr("node", node);
         if (link.state == .live) {
             self.dropCdp(link, null, true);
-            any_removed = true;
         }
         it = next;
     }
 
-    if (any_removed) {
-        self.cdp_unregister.broadcast();
-    }
+    self.cdp_unregister.broadcast();
 }
 
 pub fn run(self: *Network) void {
@@ -790,12 +786,8 @@ pub fn run(self: *Network) void {
         self.fireTicks();
 
         if (self.shutdown.load(.acquire)) {
-            // Force-disconnect any still-live CDP clients so their worker
-            // threads wake from curl_multi_poll and exit. A worker is woken
-            // only by this (Network) thread; if the loop exits with links
-            // still live, those workers block forever and Server.deinit()
-            // spins on active_threads (issue #2510). Idempotent — a no-op
-            // once the links are drained.
+            // Drain any live CDP links so their workers can exit (issue #2510).
+            // Idempo tent — no-op once drained, safe to call every iteration
             self.shutdownCdpLinks();
 
             if (running_handles == 0) {
@@ -805,7 +797,10 @@ pub fn run(self: *Network) void {
                 self.submission_mutex.lock();
                 const has_pending = self.submission_queue.first != null;
                 self.submission_mutex.unlock();
-                if (!has_pending) break;
+
+                if (!has_pending) {
+                    break;
+                }
             }
         }
     }
--- a/src/Inbox.zig
+++ b/src/Inbox.zig
@@ -39,6 +39,13 @@ const Inbox = @This();
 mutex: std.Thread.Mutex = .{},
 queue: DoublyLinkedList = .{},
 
+// One-way latch, set by the worker's drainInbox the first time it
+// observes a .disconnect (or .close) and never cleared. Ensures that, on
+// multiple drains, the terminated state is preserved / communicated. This is
+// specifically meant to handle the case where a disconnect is captured during
+// a syncRequest and we want the following non-nested tick to pick it up again.
+terminated: bool = false,
+
 pub fn deinit(self: *Inbox, arena_pool: *ArenaPool) void {
     self.mutex.lock();
     defer self.mutex.unlock();
--- a/src/browser/HttpClient.zig
+++ b/src/browser/HttpClient.zig
@@ -398,6 +398,10 @@ pub fn abortRequests(_: *Client, owner: *Owner) void {
 const DrainMode = enum { all, sync_wait };
 
 pub fn tick(self: *Client, timeout_ms: u32, mode: DrainMode) !void {
+    if (self.inbox.terminated) {
+        return error.ClientDisconnected;
+    }
+
     try self.drainQueue();
     try self.perform(@intCast(timeout_ms));
     // perform/processMessages just released a batch of connections back to
@@ -536,6 +540,10 @@ const SyncContext = struct {
 };
 
 pub fn syncRequest(self: *Client, allocator: Allocator, req: Request) !SyncResponse {
+    if (self.inbox.terminated) {
+        return error.ClientDisconnected;
+    }
+
     var sync_ctx = SyncContext{ .allocator = allocator, .body = .empty };
     errdefer sync_ctx.body.deinit(allocator);
 
@@ -719,10 +727,12 @@ fn drainInbox(self: *Client, mode: DrainMode) !void {
             .close => {
                 cdp.onClose();
                 cdp.onDisconnect(null);
+                self.inbox.terminated = true;
                 return error.ClientDisconnected;
             },
             .disconnect => |err| {
                 cdp.onDisconnect(err);
+                self.inbox.terminated = true;
                 return error.ClientDisconnected;
             },
         }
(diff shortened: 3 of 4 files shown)


karlseguin:
@krichprollsch mind reviewing my changes? They are somewhat independent but serve the same purpose. The original commit looks good, but, for me, it wasn't enough to cleanly stop under all circumstances. Only the two sets of changes combined were enough.

navidemad:
Thanks @karlseguin — good catch, I'd missed the `syncRequest` path. Agreed the two changes are complementary, not redundant. Writing out the split for @krichprollsch's review:

- **My commit** handles the *idle* worker — parked at the top-level `perform`/poll. `shutdownCdpLinks` → `dropCdp(notify=true)` pushes a `.disconnect` and wakes it, so `drainInbox` returns `error.ClientDisconnected` and the worker exits.
- **Your commit** handles the *busy* worker — mid-`syncRequest`. The `.disconnect` is drained inside the `.sync_wait` loop and `error.ClientDisconnected` propagates out of `syncRequest`, but the JS caller treats it as a failed fetch rather than a fatal disconnect. The worker resumes, re-parks with the link already dropped and nothing left to wake it. The `terminated` latch makes that one-shot error sticky so the next `tick`/`syncRequest` exits cleanly.

So: mine delivers + wakes, yours makes the signal survive a swallowed error. Both needed. (Your half also fixes this *outside* shutdown — a client disconnecting mid-synchronous-fetch would leak its worker thread the same way, and on a single-connection server wedge the next connection.)

I checked the latch's safety since it's never cleared: `terminated` is only touched on the worker thread (the Network thread only touches `inbox.queue` under the mutex, so no atomic needed), and it can't poison the next pooled connection because `cdp.init` → `browser.init` → `http_client.init` rebuilds the inbox fresh (`.inbox = .{}`).

**One real issue in `Capture disconnect/close in Worker`:** the early return in `syncRequest` (`HttpClient.zig:543`) returns *before* `self.request(r, null)`, which takes ownership of `req.headers` and frees the curl_slist on every path. Callers (`Frame.zig` external stylesheet, `ScriptManager` blocking scripts, worker `importScripts`) rely on `request()` to free them — so after a latched disconnect, any nested fetch leaks its header list. I pushed a one-line fix mirroring `request()`'s own failure paths:

```zig
if (self.inbox.terminated) {
    req.headers.deinit();
    return error.ClientDisconnected;
}
```

Tests: I pushed a `tick`-latch regression test (fails against the pre-latch `HttpClient.zig` and passes with it) and a behavioral `syncRequest`-after-disconnect test. The header leak above isn't directly assertable — curl_slist is C-allocated through the tracking allocator, outside the per-test leak check — so that one's verified by review. Full suite green (719/719).

One thing, take or leave: the now-unconditional `cdp_unregister.broadcast()` — is that closing a specific missed-wakeup race, or defensive? Harmless either way; a one-line comment naming the race would help future readers if there is one.

@krichprollsch over to you for the combined diff.

src/network/Network.zig
```diff
@@ -683,6 +683,31 @@ fn processCdpEvents(self: *Network) void {
...
+// Mirrors the peer-EOF path in processCdpEvents: dropCdp(notify=true)
+// pushes a .disconnect into the worker's inbox and wakes it, so
+// cdp.tick() returns false and the worker exits.
+fn shutdownCdpLinks(self: *Network) void {
+    self.cdp_mutex.lock();
+    defer self.cdp_mutex.unlock();
+
+    var it = self.cdp_links.first;
+    while (it) |node| {
+        const next = node.next;
```
krichprollsch: maybe it's silly and I see we do this way on other loops, but why not directly having `it = node.next;`?

krichprollsch (approved):
One non-blocking comment

src/network/Network.zig
navidemad: Not silly at all. The direct form isn't safe here because this loop removes nodes while it iterates, so reading `node.next` after `dropCdp` breaks two ways:

1. `dropCdp` calls `self.cdp_links.remove(&link.node)` (Network.zig:535). Zig's `DoublyLinkedList.remove` doesn't clear the removed node's own `next` (it just relinks the neighbors), so `node.next` still points at the old successor and we'd keep walking a node that's already out of the list.
2. With `notify=true`, `dropCdp` also wakes the worker, which can then exit and free the `CdpLink` that `node` lives inside. By the time the iteration resumes, `node` itself may be gone, so it's a real use-after-free and not just a stale read.

Grabbing `next` before the `dropCdp` call avoids both. The loops that advance with `: (it = node.next)` directly (like `prepareCdpPollFds`, Network.zig:571) only read the list and never drop nodes, so they're fine. This one matches the `processCdpEvents` first pass (Network.zig:601), which removes nodes for the same reason.

karlseguin: Yes, but you can grab next BEFORE the drop INTO `it`

```zig
while (it) |node| {
    it = node.next;
    const link: *CdpLink = @fieldParentPtr("node", node);
    if (link.state == .live) {
        self.dropCdp(link, null, true);
    }
}
```

navidemad: Good call @karlseguin, applied.

Reading `node.next` straight into `it` before the drop keeps the same safety (successor captured while the node's still linked) and drops the extra local.

PR #2512, opened 2026-05-21
Ensure about:blank frame visibility

karlseguin:
Our navigate on about:blank short-circuits most of the complex logic and loads the content then and there, not asynchronously. This results in notifications for the frame navigation/creation that happen immediately. With the current code ordering though, the frame isn't entered in child_frames until AFTER navigation, resulting in these events trigger on a frame which can't be found via Session.findFrameByFrameId

This code adds the frame to child_frames BEFORE navigate (and removes it on error).

Note: about:blank iframe navigation is more common than you probably think. As soon as a frame is [dynamically] created, it navigates to about:blank, e.g.:

let iframe = document.createElement('iframe');
// IMPLICIT navigation to about:blank happens here

// and this will be another navigation event
iframe.src = "keemun.php";

### DIFF
--- a/src/browser/Frame.zig
+++ b/src/browser/Frame.zig
@@ -1393,7 +1393,7 @@ pub fn iframeAddedCallback(self: *Frame, iframe: *IFrame) !void {
         .referer = parent_url,
         .initiator_url = parent_url,
     }) catch |err| {
-        // extra defensive..maybe navigate added a new fame, and the index it
+        // extra defensive..maybe navigate added a new frame, and the index it
         // was added at was removed. Or maybe this frame was removed somehow
         // (which I don't think is possible)
         if (std.mem.indexOfScalar(*Frame, self.child_frames.items, new_frame)) |idx| {
--- a/src/browser/Frame.zig
+++ b/src/browser/Frame.zig
@@ -1384,6 +1384,11 @@ pub fn iframeAddedCallback(self: *Frame, iframe: *IFrame) !void {
     // frame in self.child_frames.
     try self.child_frames.append(self.arena, new_frame);
 
+    // navigate() may run JS that reads window[N]; flag the list unsorted until
+    // we've verified ordering post-navigate.
+    const was_sorted = self.child_frames_sorted;
+    self.child_frames_sorted = false;
+
     // Iframe's initial src request carries the parent's URL as Referer and
     // as the SameSite initiator. Parent frame outlives this navigate()
     // call, so the slice is safe.
@@ -1412,11 +1417,12 @@ pub fn iframeAddedCallback(self: *Frame, iframe: *IFrame) !void {
     const frames_len = self.child_frames.items.len;
     if (frames_len == 1) {
         // this is the only frame, it must be sorted.
+        self.child_frames_sorted = true;
         return;
     }
 
-    if (self.child_frames_sorted == false) {
-        // the list already wasn't sorted, it still isn't
+    if (!was_sorted) {
+        // it was already unsorted; leave flag false
         return;
     }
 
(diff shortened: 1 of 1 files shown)


src/browser/Frame.zig
```diff
@@ -1388,20 +1393,22 @@ pub fn iframeAddedCallback(self: *Frame, iframe: *IFrame) !void {
         .referer = parent_url,
         .initiator_url = parent_url,
     }) catch |err| {
+        // extra defensive..maybe navigate added a new fame, and the index it
```
krichprollsch:
```suggestion
        // extra defensive..maybe navigate added a new frame, and the index it
```

krichprollsch (approved):

navidemad:
> [!NOTE]
> Had a look since synchronous about:blank navigation is fiddly. The core fix
> reads correctly to me — appending before `navigate()` is the right call for the
> `findFrameByFrameId` path. I did trace one residual edge case around `window[N]`
> ordering; it's narrow and I think it was actually *more* broken before this PR,
> so not a blocker — just flagging in case you want to close it cheaply here.

## 1 · The core fix looks right

`findFrameBy` recurses `child_frames` by id (`Page.zig:285`), so it's
order-independent — appending the frame before `navigate()` (`Frame.zig:1385`)
is exactly what makes it findable during the synchronous about:blank
notifications (`frame_navigate`/`frame_navigated`/`frame_loaded` fired at
`Frame.zig:589/606/980`). The error-path `swapRemove` (`Frame.zig:1399-1401`)
cleans up on failure. Makes sense.

## 2 · One residual edge case — `window[N]` during a reentrant `onload`

`getFrame` skips sorting when `child_frames_sorted == true` and returns
append-order (`Window.zig:613`). That flag is only re-evaluated *after*
`navigate()` returns (`Frame.zig:1412-1433`). The catch is that about:blank
dispatches the iframe's `load` event synchronously *inside* `navigate()`
(`documentIsComplete` → `notifyParentLoadComplete` → `iframeCompletedLoading`,
`Frame.zig:922`), which can run parent in-page JS while the flag is still stale:

```mermaid
sequenceDiagram
    participant JS as Parent JS
    participant P as Parent Frame
    participant C as New Frame (about:blank)
    participant GF as getFrame / window[N]

    JS->>P: insertBefore(iframe, existing)  ← out of doc order
    activate P
    P->>P: child_frames.append(new)
    Note over P: list now unsorted,<br/>but flag still = true
    P->>C: navigate("about:blank")
    activate C
    C->>C: documentIsComplete → notifyParentLoadComplete
    C->>P: iframeCompletedLoading(iframe)
    P-->>JS: dispatch "load" on iframe (sync)
    JS->>GF: read window[0]
    Note over GF: flag == true → no sort →<br/>returns append-order (WRONG)
    deactivate C
    P->>P: child_frames_sorted maintenance
    Note over P: sets flag = false — but the<br/>onload already read stale order
    deactivate P
```

So if an iframe is inserted **out of document order** (`insertBefore`) into a
previously-sorted list **and** its `onload` reads `window[N]`/`frames[N]`, that
read sees append-order instead of document-order. It self-corrects once
`navigate()` returns (`Frame.zig:1432`). And to be clear this isn't a regression
— pre-PR the same handler would have seen the new frame *missing entirely*
(it wasn't in `child_frames` yet), so this PR is a net improvement either way.

## 3 · If you want to close it (cheap)

The sorted-check at `Frame.zig:1412-1433` only depends on the two iframes'
DOM positions, which are fixed before `iframeAddedCallback` even runs —
`navigate()` doesn't touch them. So the whole block can move to right after the
append, before `navigate()`:

```mermaid
graph TD
    subgraph now["now"]
        A1["append new_frame"] --> A2["navigate() — sync onload reads stale flag"] --> A3["update child_frames_sorted"]
    end
    subgraph proposed["proposed"]
        B1["append new_frame"] --> B2["update child_frames_sorted"] --> B3["navigate() — onload sees correct flag"]
    end
```

One thing to watch if you do: the error-path `swapRemove` would then run after
the flag's been set, so it'd want a reset (or just set the flag conservatively
`false` after append and restore the optimization after `navigate()` succeeds).
That's only an optimization concern, not correctness.

## 4 · Nits

| Where | Nit |
|-------|-----|
| `Frame.zig:1363` | `frame_child_frame_created` still dispatches *before* the append — intentional that "created" fires pre-list while the navigation events fire post-append? Just checking it's deliberate. |

karlseguin:
@navidemad  thanks, that caught a real issue, appreciate that! You think the verbosity can be toned down in the future though?

PR #2514, opened 2026-05-21
Performance WebAPI on Worker

karlseguin:
This includes EventCounts and PerformanceObserver. This change is like any other WebAPI on Worker change (e.g. Frame -> js.Execution), but we need the performance observer hooks in addition to that. Thankfully, every Frame/Worker only has 1 Performance instance, so we can move the observers and delivery from the Frame to the Performance instance so that both Frame and WGS gain the behavior.

### DIFF
--- a/src/browser/Frame.zig
+++ b/src/browser/Frame.zig
@@ -47,7 +47,6 @@ const ShadowRoot = @import("webapi/ShadowRoot.zig");
 const Performance = @import("webapi/Performance.zig");
 const Screen = @import("webapi/Screen.zig");
 const VisualViewport = @import("webapi/VisualViewport.zig");
-const PerformanceObserver = @import("webapi/PerformanceObserver.zig");
 const AbstractRange = @import("webapi/AbstractRange.zig");
 const MutationObserver = @import("webapi/MutationObserver.zig");
 const IntersectionObserver = @import("webapi/IntersectionObserver.zig");
@@ -173,11 +172,6 @@ _intersection_delivery_scheduled: bool = false,
 // Slots that need slotchange events to be fired
 _slots_pending_slotchange: std.AutoHashMapUnmanaged(*Element.Html.Slot, void) = .{},
 _slotchange_delivery_scheduled: bool = false,
-
-/// List of active PerformanceObservers.
-/// Contrary to MutationObserver and IntersectionObserver, these are regular tasks.
-_performance_observers: std.ArrayList(*PerformanceObserver) = .{},
-_performance_delivery_scheduled: bool = false,
 
 // Lookup for customized built-in elements. Maps element pointer to definition.
 _customized_builtin_definitions: std.AutoHashMapUnmanaged(*Element, *CustomElementDefinition) = .{},
@@ -321,7 +315,7 @@ pub fn init(self: *Frame, frame_id: u32, page: *Page, parent: ?*Frame) !void {
         ._proto = undefined,
         ._document = self.document,
         ._location = &default_location,
-        ._performance = Performance.init(),
+        ._performance = .init(),
         ._screen = screen,
         ._visual_viewport = visual_viewport,
         ._cross_origin_wrapper = undefined,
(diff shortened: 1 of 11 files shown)


krichprollsch (approved):

navidemad:
> [!NOTE]
> The refactor itself reads really well — decoupling `Performance` from `Frame`
> onto `Execution` and moving the observer machinery onto `Performance` so both
> globals share it is clean, and the `exec.arena` lifetime looks right to me (no
> dangling observer/entry refs that I could find). I have one thing I'd flag for
> a decision — the worker API surface — plus a couple of minor notes.

## 1 · What I checked and liked

- Observer registration moved off `Frame` onto `Performance` (`registerObserver`
  / `notifyObservers` / `scheduleDelivery`); each global owns its own
  `Performance`, so workers get observers for free. Nice.
- Entries/marks/measures allocate from `exec.arena` / `exec._factory`, matching
  the owning context's lifetime. No cross-context capture that I could spot.

## 2 · The one thing worth a decision — Window-only surface leaks onto workers

Registering the whole `Performance` interface in `WorkerJsApis` (`bridge.zig:987`)
also exposes its Window-only accessors. It's not just `eventCounts` — `timing`
and `navigation` come along too (`Performance.zig:368-370`). Per spec all three
live on Window-only partials, while the base interface is `[Exposed=(Window,Worker)]`:

```mermaid
graph LR
    subgraph shared["base Performance — Exposed Window + Worker"]
        A["now / mark / measure"]
        B["getEntries* / clearMarks / clearMeasures"]
        D["timeOrigin"]
    end
    subgraph winonly["Window-only partials (spec)"]
        E["timing"]
        F["navigation"]
        G["eventCounts"]
    end
    shared --> W["worker.performance"]
    winonly -. this PR also lands these on workers .-> W
```

`eventCounts` is the most concrete case: even setting the spec aside, on a worker
it's permanently empty. The only place it's incremented is `EventManager.zig:106`,
which is hardcoded to `self.frame.window._performance` — workers never hit that
path, so `worker.performance.eventCounts` always reads zero. The new
`worker_event_counts` test (`performance.html:574`) only asserts shape + initial
zeros, so it passes even though the feature is inert.

The reason it's one decision and not three separate fixes: the bridge gates
exposure per *interface*, not per *member* (`bridge.zig:739` vs `:947` are
type lists). So the options as I see them:

- **Split the interface** — base `Performance` shared + a Window-only piece
  adding `timing`/`navigation`/`eventCounts`. Matches how the spec is structured.
- **Add member-level context gating to the bridge** — bigger change, but reusable.
- **Accept the drift** and just drop the `worker_event_counts` test so we're not
  locking in always-zero behavior.

You'd know which fits the bridge best — flagging the tradeoff rather than
prescribing one.

## 3 · Minor — `notifyObservers` always schedules a delivery task

`notifyObservers` calls `scheduleDelivery` unconditionally (`Performance.zig:312`/
`:321`), so every `mark()`/`measure()` queues a scheduler task even when
`_observers` is empty or nobody's interested. The `_delivery_scheduled` guard
stops double-scheduling, but not the empty case. Cheap to skip the schedule
unless at least one entry was actually queued.

(Also tiny: that error log uses `lp.log.err(.frame, ...)` inside what's now a
context-agnostic component — slightly off for the worker case.)

## 4 · Pre-existing, out of scope — `measure()` with all three timing fields

Noticed while reading: `measure(name, { start, end, duration })` with all three
present silently prefers `duration` (`Measure.init`, `Performance.zig:519`),
where User Timing L3 wants a `TypeError`. This isn't introduced by your PR — the
refactor only swapped `frame` → `exec` here — so purely a possible follow-up.

## Nits

| Where | Nit |
|-------|-----|
| `performance.html:574` | `worker_event_counts` only checks shape + zeros — won't catch the inert increment path |
| `notifyObservers` | `lp.log.err(.frame, …)` scope on a now-shared component |

karlseguin:
Thanks @navidemad I had no idea that Window and Worker shared the same types but that they could have different shapes. I thought this would always be represented as distinct types (e.g. Location vs WorkerLocation). I've added an `exposed = both/window/worker` to the bridge.

PR #2515, opened 2026-05-21
Add watchdog to Network thread for abandoned & stuck workers

karlseguin:
If a worker is in some heavy JS e.g. `for(;;)` it will be stuck forever, even if the peer closes the CDP connection.

With CDP reads now owned by the network thread, we now correctly detect the disconnect and simply need to force the worker to shutdown. To achieve this, on socket close, the CdpLink held by the network is given a terminate_ms (five seconds from now) and added to a linked list. On every wakeup, the network thread can check the list + timestamp and, if necessary, call Isolate::Terminate (which is safe to call on a different thread).

### DIFF
--- a/src/network/Network.zig
+++ b/src/network/Network.zig
@@ -536,6 +536,8 @@ fn dropCdp(self: *Network, link: *CdpLink, err: ?anyerror, notify: bool) void {
     link.state = .removed;
     self.cdp_dirty = true;
     if (notify) {
+        link.cdp.terminateFromNetwork();
+
         // notify=true means the worker hasn't been told yet — push the
         // disconnect into the inbox and break it out of curl_multi_poll.
         // notify=false paths have already woken the worker (close frame
--- a/src/network/Network.zig
+++ b/src/network/Network.zig
@@ -22,7 +22,6 @@ const builtin = @import("builtin");
 
 const App = @import("../App.zig");
 const Config = @import("../Config.zig");
-const datetime = @import("../datetime.zig");
 
 const CDP = @import("../cdp/CDP.zig");
 const libcurl = @import("../sys/libcurl.zig");
@@ -66,23 +65,11 @@ pub const CdpLink = struct {
     handles: http.Handles,
     node: DoublyLinkedList.Node = .{},
 
-    // Network thread will call cdp.terminateFromNetwork() when this is reached.
-    // The timing guarantee is poor, but it will be at some point AFTER this
-    // time and before an accept (ensure abandoned workers can't block new
-    // connections)
-    terminate_ms: u64 = std.math.maxInt(u64),
-
     pub const State = enum {
         live,
         // Worker called unregisterCdp; Network will drop the link on
         // its next loop iteration and signal cdp_unregister.
         unregistering,
-        // Network dropped the socket and parked the link in
-        // closed_cdp_linkss. Worker's unregisterCdp OR the watchdog
-        // deadline transitions it to .removed. Must NOT be .removed
-        // while still in closed_cdp_linkss — worker would free out
-        // from under the next tick.
-        terminating,
         // Network has dropped the link from its poll set. The worker
         // can safely free anything the link's callbacks closed over.
         removed,
@@ -139,13 +126,6 @@ callbacks_mutex: std.Thread.Mutex = .{},
 cdp_links: DoublyLinkedList = .{},
 cdp_mutex: std.Thread.Mutex = .{},
 cdp_unregister: std.Thread.Condition = .{},
-// After the CDP socket is closed, the CdpLink is moved from cdp_links to
-// closed_cdp_links. At that point we act as a watchdog. If the worker doesn't
-// unregister the CDP link within 5 seconds (e.g. stuck in a endless JS loop?)
-// we can force a termination from the worker. (This is a bit like FetchTerminator
-// in main)
-closed_cdp_links: DoublyLinkedList = .{},
-
 // Per-iteration snapshot of CdpLinks whose sockets are in pollfds.
 // Sized at maxConnections at init time so we never allocate inside
 // run(). Parallel to pollfds[cdp_start..cdp_start + cdp_poll_count].
@@ -539,14 +519,6 @@ pub fn unregisterCdp(self: *Network, link: *CdpLink) void {
         self.wakeupPoll();
     }
 
-    if (link.state == .terminating) {
-        // Network parked us in closed_cdp_links. Unpark ourselves and
-        // skip the watchdog path.
-        self.closed_cdp_links.remove(&link.node);
-        link.state = .removed;
-        return;
-    }
-
     while (link.state != .removed) {
         // condition variable, waiting for a signal
         self.cdp_unregister.wait(&self.cdp_mutex);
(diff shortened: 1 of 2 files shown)


src/network/Network.zig
```diff
@@ -65,11 +66,23 @@ pub const CdpLink = struct {
...
+    // connections)
+    terminate_ms: u64 = std.math.maxInt(u64),
+
     pub const State = enum {
         live,
         // Worker called unregisterCdp; Network will drop the link on
         // its next loop iteration and signal cdp_unregister.
         unregistering,
+        // Network dropped the socket and parked the link in
+        // closed_cdp_linkss. Worker's unregisterCdp OR the watchdog
```
krichprollsch:
```suggestion
        // closed_cdp_links. Worker's unregisterCdp OR the watchdog
```

src/network/Network.zig
```diff
@@ -65,11 +66,23 @@ pub const CdpLink = struct {
...
+        // deadline transitions it to .removed. Must NOT be .removed
+        // while still in closed_cdp_linkss — worker would free out
```
krichprollsch:
```suggestion
        // while still in closed_cdp_links — worker would free out
```

krichprollsch (approved):
I like it, it works well.
2 questions:

1. Finally do we really need to give a chance to the worker to terminate itself? what about immediately terminate from the main thread?

2. if we need the delay, why 5s?

navidemad:
> [!NOTE]
> The watchdog adds a new cross-thread path, so I spent some time tracing it and
> figured I'd share what I found in case it's useful. As far as I can tell the
> design is sound — the `.terminating` state closes the disconnect→teardown
> window cleanly. One thing I wasn't sure about below, plus a doc idea and a
> couple of nits.

## 1 · The cross-thread terminate looks safe

The part that gave me pause is the watchdog calling
`link.cdp.terminateFromNetwork()` up to 5s after socket close — a cross-thread
deref of `link.cdp.browser.env` that didn't exist before this PR. Tracing it, it
holds up because of two interlocking facts:

- **Mutual exclusion** — `terminateAbandonedWorkers` (`Network.zig:872`) and
  `unregisterCdp` (`Network.zig:533`) both hold `cdp_mutex` for their *entire*
  body. Whichever reaches the link first transitions it out of
  `closed_cdp_links`; the other sees `.removed` and is a no-op.
- **Teardown ordering** — in `handleConnection` the `unregisterCdp` defer is
  registered last (`Server.zig:247`), so it runs *before* `cdp.deinit()`
  (`Server.zig:186`) and `cdp_pool.destroy` (`Server.zig:179`). The Env is
  therefore always alive when the watchdog terminates it.

```mermaid
sequenceDiagram
    participant W as Worker
    participant Mtx as cdp_mutex
    participant N as Network (watchdog)

    Note over W,N: link in closed_cdp_links · state = .terminating

    N->>Mtx: lock()
    activate Mtx
    W-->>Mtx: unregisterCdp() blocks here
    N->>N: node = n.next  (advance BEFORE remove)
    N->>N: link.cdp.terminateFromNetwork()
    Note right of N: env alive — worker hasn't<br/>cleared unregisterCdp yet
    N->>N: remove(node) · state = .removed
    N->>Mtx: unlock()
    deactivate Mtx
    Mtx-->>W: acquired
    W->>W: state == .removed → return (no-op)
    W->>W: cdp.deinit() / free
    Note right of W: Network already dropped the<br/>link — nothing dangles
```

So I don't think the "free callbacks while Network still references the link"
scenario can occur: the worker is parked on the mutex for the whole watchdog
iteration, and `node = n.next` is captured before `remove`, so the terminated
link is never touched again. The reverse interleaving (worker wins the lock
first) is symmetric — it hits the `.terminating` branch at `Network.zig:542`,
unparks itself, and the watchdog never sees the node.

<details>
<summary>Line-by-line trace of both interleavings</summary>

**Watchdog wins the lock**
1. `terminateAbandonedWorkers` locks `cdp_mutex`, worker blocks in `unregisterCdp`.
2. `node = n.next` captured; `terminateFromNetwork()` → `env.terminate()` (env alive).
3. `closed_cdp_links.remove(n)`; `state = .removed`; loop continues on `node`, not the removed link.
4. Unlock → worker acquires, sees `.removed`, the `while (state != .removed)` guard is already false → returns.
5. Defers unwind: `cdp.deinit()` / `destroy` run only now, after Network is provably done.

**Worker wins the lock**
1. `unregisterCdp` locks, hits `if (link.state == .terminating)` at `Network.zig:542`.
2. `closed_cdp_links.remove(&link.node)`; `state = .removed`; returns.
3. Watchdog later locks, the node is gone from `closed_cdp_links` → never iterated.

Either way the link leaves `closed_cdp_links` under the lock before any free.
</details>

## 2 · The one thing I wasn't sure about — `cancelTerminate` asymmetry

`TerminateExecution` sets a *sticky* flag, and `Env.cancelTerminate`'s own doc
(`Env.zig:509`) warns that cleanup-time V8 calls "keep tripping over the
terminating-state asserts" until it's cleared. The only caller of
`cancelTerminate` is the fetch path (`main.zig:196`). The server path — both the
existing graceful shutdown (`Server.zig:74`) and now the watchdog — calls
`env.terminate()` but nothing seems to clear it before `cdp.deinit()`.

```mermaid
graph TD
    subgraph fetch["fetch path · main.zig"]
        A["env.terminate()"] --> B["releaseBrowser → env.cancelTerminate() ✅"]
    end
    subgraph server["server path · Server.zig"]
        C["watchdog / shutdown → env.terminate()"] --> D["cdp.deinit() · no cancelTerminate ❓"]
    end
```

**Question — you'd know this better than me:** is `cdp.deinit()` guaranteed
JS-free after a termination? If it makes any V8 call, this looks like it'd trip
the asserts — and the watchdog now exercises this path mid-session, not just at
shutdown, so the exposure seems wider than the existing shutdown-only case. If
deinit can touch V8, maybe a `cancelTerminate()` once the worker observes the
disconnect (or at the top of teardown) would cover it?

## 3 · A few ideas (take or leave)

- **Maybe worth pinning the invariant in a comment.** The watchdog's safety
  rests on the `cdp_mutex` serialization *and* the defer order in
  `handleConnection`. Both are correct today, but a future reorder of those
  defers would turn it into a UAF with no local signal. A one-liner on
  `terminateFromNetwork` / `terminateAbandonedWorkers` — "safe because
  `unregisterCdp` completes before `cdp.deinit`, both under `cdp_mutex`" — might
  save someone later.
- **A test for the motivating case.** The message cites `for(;;)`.
  `TerminateExecution` interrupts at loop back-edges so an empty busy loop
  *should* die, but since that's the exact reason this exists, a test that opens
  a session, spins a busy loop, drops the socket, and asserts the worker is
  reaped within ~5s could be nice to have.
- **Scope note.** This only fires after disconnect — a runaway loop on a
  *still-connected* session isn't bounded here. Assuming that's intentional
  (don't kill a live client), a sentence in the PR description might stop anyone
  reading it as general runaway-JS protection.

## 4 · Nits

| Where | Nit |
|-------|-----|
| `Network.zig:888` | `log.warn(.cdp, "watchdog terminate", .{})` carries no connection/link id — might be handy to correlate a forced kill to a session |

karlseguin:
> I like it, it works well. 2 questions:
> 
>     1. Finally do we really need to give a chance to the worker to terminate itself? what about immediately terminate from the main thread?
> 
>     2. if we need the delay, why 5s?

I DON'T KNOW?!   Check the revised version...a lot simpler. I tested it and if the worker isn't in JS, but then enters a JS loop after this has been called, it still works. That Terminate call doesn't just kill a current execution, it causes any future ones to throw a Js exception, which is what we want.

Thanks for the sanity check!

krichprollsch (approved):

PR #2516, opened 2026-05-21
docs(contributing): document prebuilt V8 to skip the source build

navidemad:
CONTRIBUTING's Development section points contributors at `make test` / `make build`, but doesn't mention that on a fresh checkout these compile V8 from source (several minutes) when no prebuilt library is present. CI already avoids this by downloading a prebuilt archive from the `zig-v8-fork` releases, so this adds a short note showing how to do the same locally via `ZIGFLAGS=-Dprebuilt_v8_path=...`, the flag the Makefile already forwards.


navidemad:
Superseded by #2517, which adds a `make download-v8` target plus a concise CONTRIBUTING note that points at it. Closing in favor of the consolidated PR.

PR #2517, opened 2026-05-21
Add `make download-v8` to fetch prebuilt V8 and skip the source build

navidemad:
Building V8 from source on a fresh checkout takes 10+ minutes. CI avoids this by downloading a prebuilt archive from the `zig-v8-fork` releases and passing `-Dprebuilt_v8_path` (see `.github/actions/install/action.yml` and `.github/workflows/zig-test.yml`). This brings the same shortcut to local builds:

- `make download-v8` fetches the archive matching the `v8` / `zig-v8` versions (read straight from `action.yml`, so they can't drift from CI) and symlinks it to `v8/libc_v8.a`, the same path CI uses.
- When `v8/libc_v8.a` exists and the caller hasn't set `ZIGFLAGS`, the Makefile links against it instead of building V8 from source. So after one `make download-v8`, plain `make build` / `make test` skip the source build with no extra flags.
- A short CONTRIBUTING note points contributors at the target.

Note the V8 dependency has no default lookup: it builds from source unless `-Dprebuilt_v8_path` is passed. That's why the Makefile sets the flag rather than relying on the symlink alone.

Verified on macOS arm64: `make download-v8` creates the symlink, and `make build-dev` with `ZIGFLAGS` unset builds against the prebuilt archive without a source rebuild. `v8/` is added to `.gitignore`.

Supersedes #2516, which documented the manual path on its own.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,13 +43,12 @@ V8_VERSION := $(shell awk -F\' '/^  v8:/{f=1} f&&/default:/{print $$2; exit}' $(
 ZIG_V8_TAG := $(shell awk -F\' '/^  zig-v8:/{f=1} f&&/default:/{print $$2; exit}' $(V8_ACTION))
 V8_ARCHIVE := libc_v8_$(V8_VERSION)_$(OS)_$(ARCH).a
 V8_CACHE   := .lp-cache/prebuilt-v8/$(V8_ARCHIVE)
-V8_LINK    := v8/libc_v8.a
 
-# If the prebuilt archive is in place and the caller hasn't set ZIGFLAGS, link
-# against it (the same flag CI passes) rather than building V8 from source.
+# If the prebuilt archive is in place and the caller hasn't set ZIGFLAGS, point
+# the build at it rather than building V8 from source.
 ifeq ($(strip $(ZIGFLAGS)),)
-  ifneq ($(wildcard $(V8_LINK)),)
-    ZIGFLAGS := -Dprebuilt_v8_path=$(V8_LINK)
+  ifneq ($(wildcard $(V8_CACHE)),)
+    ZIGFLAGS := -Dprebuilt_v8_path=$(V8_CACHE)
   endif
 endif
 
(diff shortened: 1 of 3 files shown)


karlseguin:
There are cases where I'd use this, but I'll defer to @krichprollsch

Makefile
```diff
@@ -33,6 +33,27 @@ else
...
+# -----------
+# Building V8 from source takes 10+ minutes. `make download-v8` fetches the
+# matching prebuilt archive from the zig-v8-fork releases instead. The versions
+# are read from the install action so they can't drift from CI.
+V8_ACTION := .github/actions/install/action.yml
+V8_VERSION := $(shell awk -F\' '/^  v8:/{f=1} f&&/default:/{print $$2; exit}' $(V8_ACTION))
+ZIG_V8_TAG := $(shell awk -F\' '/^  zig-v8:/{f=1} f&&/default:/{print $$2; exit}' $(V8_ACTION))
+V8_ARCHIVE := libc_v8_$(V8_VERSION)_$(OS)_$(ARCH).a
+V8_CACHE   := .lp-cache/prebuilt-v8/$(V8_ARCHIVE)
+V8_LINK    := v8/libc_v8.a
```
krichprollsch: Why having both `.lp-cache/prebuilt-v8` and `v8/libc_v8.a`? Why not directly using `.lp-cache/prebuilt-v8/libc_v8.a`?

krichprollsch (commented):
I'm OK with the change, just having question about adding a new `v8/` dir.

Makefile
navidemad: Good call — dropped the `v8/` symlink. `make download-v8` now downloads straight into `.lp-cache/prebuilt-v8/` and the Makefile points `-Dprebuilt_v8_path` at that file directly, so there's no second location and no new `v8/` dir (I also removed the `/v8/` line this PR had added to `.gitignore`).

I kept the version in the cached filename (`libc_v8_<version>_<os>_<arch>.a`) instead of a fixed `libc_v8.a`: it's keyed off the V8 pin in `action.yml`, so bumping the pin fetches a new file rather than clobbering, and switching back needs no re-download.

Verified locally: with `ZIGFLAGS` unset, `make build-dev` links against the cached archive with no source rebuild.

krichprollsch (approved):

PR #2519, opened 2026-05-21
telemetry: document env var, add --no-telemetry flag

staylor:
## Problem

`LIGHTPANDA_DISABLE_TELEMETRY` already opts out of telemetry, but:

- It is missing from `lightpanda --help` (any subcommand).
- The README only mentions it in one sentence and shows it as `=true`, which suggests the value matters — it does not. The current implementation uses `std.process.hasEnvVarConstant`, so the variable just needs to be present (any value, including empty).
- There is no CLI equivalent, so users who do not want to set environment variables (CI configs, one-off invocations, MCP launchers) have no way to opt out.

## Changes

- **Add `--no-telemetry` as a common option** for `serve`, `fetch`, and `mcp`. Wired into the existing `CommonOptions` block in `Config.zig` so it picks up the standard `--kebab-case` / `--snake_case` handling and unknown-flag validation for free.
- **Make the crash reporter honor the flag.** `crash_handler.zig` calls `telemetry.isDisabled()` from a panic context where the `App`/`Config` are not reachable, so I added a module-level `cli_disabled` flag + `setDisabledByCli` setter that `App.init` populates from `config.noTelemetry()` before `Telemetry.init` runs. This guarantees `--no-telemetry` suppresses crash reports too.
- **Document both opt-outs.** `src/help.zon` now lists `--no-telemetry` under common options and references the env var. The README clarifies that the env var only needs to be present (no specific value required) and that debug builds always disable telemetry.
- **Extend the existing telemetry test** to assert that the CLI override path also short-circuits `isDisabled()`.

## Verification

```
zig fmt --check ./*.zig ./**/*.zig    # clean
make test                              # 720 of 720 tests passed

./zig-out/bin/lightpanda help fetch | grep -A4 no-telemetry
  --no-telemetry
          Disable anonymous usage telemetry and crash reporting.
          Equivalent to setting the LIGHTPANDA_DISABLE_TELEMETRY
          environment variable (any value, including empty).
          See https://lightpanda.io/privacy-policy for details.

./zig-out/bin/lightpanda fetch http://example.com --no-telemetryX
FATAL app : unknown argument
      mode = fetch
      arg = --no-telemetryX
```

## Notes

- No existing behavior changes when neither the flag nor env var is set.
- The semantics of `LIGHTPANDA_DISABLE_TELEMETRY` are unchanged; only its documentation is updated to match the actual implementation.
- Debug/test builds remain hard-disabled, matching the prior invariant covered by `telemetry: always disabled in debug builds`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,7 +170,12 @@ Add to your MCP configuration:
 A skill is available in [lightpanda-io/agent-skill](https://github.com/lightpanda-io/agent-skill).
 
 ### Telemetry
-By default, Lightpanda collects and sends usage telemetry. This can be disabled by setting an environment variable `LIGHTPANDA_DISABLE_TELEMETRY=true`. You can read Lightpanda's privacy policy at: [https://lightpanda.io/privacy-policy](https://lightpanda.io/privacy-policy).
+By default, Lightpanda collects and sends anonymous usage telemetry and crash reports. You can opt out in either of two equivalent ways:
+
+- Pass `--no-telemetry` to any command (`serve`, `fetch`, `mcp`).
+- Set the environment variable `LIGHTPANDA_DISABLE_TELEMETRY` to any value (e.g. `LIGHTPANDA_DISABLE_TELEMETRY=1`). The variable only needs to be present; its value is not inspected.
+
+Telemetry is also always disabled in debug builds. You can read Lightpanda's privacy policy at: [https://lightpanda.io/privacy-policy](https://lightpanda.io/privacy-policy).
 
 ## Status
 
--- a/src/App.zig
+++ b/src/App.zig
@@ -22,7 +22,8 @@ const lp = @import("lightpanda");
 const Config = @import("Config.zig");
 const Snapshot = @import("browser/js/Snapshot.zig");
 const Platform = @import("browser/js/Platform.zig");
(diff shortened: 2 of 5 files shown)


karlseguin:
I _do_ like that this is more discoverable. My only reservation is that there's a window (before App.init is called) where the flag is ignored. So it would be possible for --no-telemetry to be set but a crash report would still happen. I assume this is likely common. Punting this one to @krichprollsch also.

krichprollsch:
Thanks for the contribution. After discussion and while we understand your points, as an open source project, telemetry is vital to sustaining development in the long run. That's why we prefer to close this change and keep only the env var for now.

> There is no CLI equivalent, so users who do not want to set environment variables (CI configs, one-off invocations, MCP launchers) have no way to opt out.

For CI config it's common to setup env vars, [we do that ourself](https://github.com/lightpanda-io/browser/blob/main/.github/workflows/e2e-test.yml#L8).

On-off invocation is also quite easy to do: `LIGHTPANDA_DISABLE_TELEMETRY=true lightpanda`

And in mcp configuration, for example, with Claude [you can setup env vars easily](https://code.claude.com/docs/en/mcp#what-you-can-do-with-mcp).

So IMO the env var is well supported and good enough in all these cases.

PR #2520, opened 2026-05-21
script: unify PandaScript to slash commands

arrufat:
Unifies PandaScript syntax by replacing uppercase keywords with slash
commands (e.g., `/goto`, `/click`, `/fill`). Refactors the `Command`
representation to use a generic tool call structure backed by schemas.

BREAKING CHANGE: Uppercase PandaScript keywords (GOTO, CLICK, TYPE, etc.)
are no longer supported. All scripts must use slash commands.


src/agent/CommandRunner.zig
```diff
@@ -20,41 +20,83 @@ const std = @import("std");
...
+    const needsSub = struct {
+        fn f(is_fill_: bool, key: []const u8, val: std.json.Value) bool {
+            if (is_fill_ and std.mem.eql(u8, key, "value")) return false;
+            return val == .string and std.mem.indexOf(u8, val.string, "$LP_") != null;
+        }
+    }.f;
+
+    var needs_any = false;
+    var it = v.object.iterator();
+    while (it.next()) |entry| {
```
karlseguin:
```zig
var it = v.object.iterator();
while (it.next()) |entry| {
    if (needsSub(is_fill, entry.key_ptr.*, entry.value_ptr.*)) {
        break;
    }
} else {
  return v;
}
```

src/agent/SlashCommand.zig
```diff
@@ -16,63 +16,40 @@
...
     hint: []const u8,
     /// Tab-completion candidates for the first positional arg.
     values: []const [:0]const u8,
-};
 
-pub const meta_commands = [_]MetaCommand{
-    .{ .name = "help", .hint = "", .values = &.{} },
-    .{ .name = "quit", .hint = "", .values = &.{} },
-    .{ .name = "verbosity", .hint = "<low|medium|high>", .values = &.{ "low", "medium", "high" } },
+    pub const Kind = enum { help, quit, verbosity };
```
karlseguin:
```suggestion
    const Kind = enum { help, quit, verbosity };
```

src/script/schema.zig
```diff
@@ -0,0 +1,489 @@
...
+};
+
+/// Split a slash-command body into `<name> <rest>`. Returns null on empty input.
+pub fn splitNameRest(input: []const u8) ?Split {
+    const trimmed = std.mem.trim(u8, input, &std.ascii.whitespace);
+    if (trimmed.len == 0) return null;
+    const name_end = std.mem.indexOfAny(u8, trimmed, &std.ascii.whitespace) orelse trimmed.len;
+    return .{
+        .name = trimmed[0..name_end],
+        .rest = std.mem.trim(u8, trimmed[name_end..], &std.ascii.whitespace),
```
karlseguin:
```suggestion
        .rest = std.mem.trimStart(u8, trimmed[name_end..], &std.ascii.whitespace),
```

the tail is already trimmed by the first line in the function.

src/script/command.zig
```diff
@@ -525,963 +231,454 @@ pub const Command = union(enum) {
...
             .triple_double => "\"\"\"",
             .triple_single => "'''",
         };
     }
 
-    /// Pick the triple-quote delimiter that does not collide with `body`.
-    /// Defaults to `triple_single`; swaps to `triple_double` only when the
-    /// body already contains `'''`.
-    fn pickFor(body: []const u8) QuoteType {
+    /// Default `'''`; swaps to `"""` only when the body already contains `'''`.
```
karlseguin: It could contain both. I'm just poking around, so I'm not sure how strict the rest is with respect to potentially invalid scripts. But if invalid scripts are generally rejected, I think it might make sense to do so here too.

karlseguin (commented):
I just poked around the code a little, all my comments are trivial.

Here are additional observation from using it. Again, none are very significant and if you do think there's some valid feedback here, it might be better in a follow up PR anyways.


1.
```
> /goto http://www.lightpanda.io/ waitUntil=x
goto failed: InvalidParams
```

Not clear indication of what's wrong, or what waitUtil wants.

2.
Also, I think a  `waitForScript` would be a nice addition. The runner supports it.

3.
```
> /waitForSelector "a.asdsad" timeout=0
  waitForSelector failed: InternalErr
```

Honestly don't know what I did wrong here. If the issue is that it couldn't be found, then when I pass timeout=10000 I'd expect it to wait, but the error is still returned immediately.

Also, if the issue IS that the selector isn't found, the error is inconsistent with `/hover selector="a.asdsad"`

4. 
For my old eyes, on my terminal, the /help text is very low contrast with the background.

5.
Is it worth having an explicit `/prompt`. I typed `help` instead of `/help` more than a few times.

6.
Inside of of a multi-line string, the "help" attribute list shows up
```
> /eval """
console.warn(1)  
console.warn(a) [url=...] [timeout=...] [waitUntil=...]
```
Yes, it's greyed out, but still weird/wrong.

7.
Enter inside a multi-line string executes the command. I'm not sure what other tools do, but i feel that it should enter a newline. 

8.
Hitting up (to go back in history) brings me to the start of a multi-line input. I dunno if this is right or wrong. I thin I expected it to take me to the end of it, but I bet if it went to the end I'd complain that it didn't go to the start!

9.
```
> /eval "setTimeout(() => document.querySelector('a'), 1000)"
● /eval "setTimeout(() => document.querySelector('a'), 1000)": missing required argument. Try /help /eval "setTimeout(() => document.querySelector('a'), 1000)".
```
I don't know why that's invalid, and the help message is wrong  (if I type what it tells me to, it complains). 

10.
ctrl-c doesn't exit (and when run from `zig build run` it requires a `kill -9` on the lightpanda process to recover)

11.
/help split between "Slash commands' and "Meta" seems arbitrary (meta are also slash commands from my point of view). 

12.
`/exit` can?

13.
`/help` commands sorted by name?

14.
```
> /goto http::www
⡇ [goto http::www]ERROR frame : navigate failed . . . . . . . . . . . . . . . . [+0ms]
      err = UrlMalformat
      type = root
      url = http::www

Navigated successfully.
```

(a) The two logs share the same line
(b) was it really "Navigated successfully" ?

15.
```
> /goto https://www.openmymind.net/
Navigated successfully.
> /waitForSelector 'main a'
Element found. backendNodeId: 1
> /click backendNodeId=1
Clicked element (backendNodeId: 1). Page url: https://www.openmymind.net/Is-Zigs-New-Io-Unsafe/, title: Is Zig's New Writer Unsafe?
> /click backendNodeId=1
⠏ [click backendNodeId=1]Segmentation fault at address 0xaaaaaaaaaaaaaaaa
....
```

16.
I think I'd prefer it if the commands and argument names were case insensitive.

17.
`/press` couldn't figure out how to "press" enter. Tried '13', 'enter' and '\n'. (Fundamentally, I was on DDG and coudln't figure out how to submit the form, short of using /eval)

18.
/goto SOME_URL_WITH_COOKIES
/goto SOME_URL_WITHOUT_COOKIES
/getCookies <-- shows cookies for the first URL

Dunno if that's how it's suppposed to be.

19.
No way to dump the full HTML?

karlseguin:
add `.lp-history` to `.gitignore`

arrufat:
Thanks a lot for the review, @karlseguin. I merge the unification, and tackle those usability issues in a separate PR.

PR #2537, opened 2026-05-24
feat(webapi): implement W3C File API

Ar-maan05:
Implements the W3C `File` Web API, which inherits from `Blob`.

        - Exposes `validateMimeType` and `writePartWithEndings` in `Blob.zig` to be reused.
        - Implements constructor (`new File(bits, name, options)`), name, and lastModified properties.
        - Maps `lastModified` as an `f64` to prevent JS from seeing it as a `bigint`.
        - Adds HTML integration tests in `file.html`.


Ar-maan05:

PR #2542, opened 2026-05-25
Immediately run deferred inline scripts after "load" is fired

karlseguin:
Currently, any inline deferred script, e.g:

<script type=module>
...
</script>

That happens AFTER load, never executes. Unclear how serious an issue this is, but it _does_ cause problems for some WPT tests which use document.write to inject a <script type="module">...</script> block after an iframe is loaded.

### DIFF
--- a/src/browser/ScriptManager.zig
+++ b/src/browser/ScriptManager.zig
@@ -210,7 +210,13 @@ pub fn addFromElement(self: *ScriptManager, comptime from_parser: bool, script_e
     };
 
     const is_blocking = mode == .normal;
-    if (is_blocking == false) {
+
+    // Once parsing is done, the deferred-script batch has already drained and
+    // won't run again, so a non-blocking script inserted afterwards would go
+    // unprocessed. Run it immediately instead. Remote scripts still need to
+    // be queued so they execute when their fetch completes.
+    const run_immediately = is_blocking or (self.base.static_scripts_done and remote_url == null);
+    if (run_immediately == false) {
         self.base.scriptList(script).append(&script.node);
     }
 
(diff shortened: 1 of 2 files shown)


src/browser/ScriptManager.zig
```diff
@@ -210,7 +210,13 @@ pub fn addFromElement(self: *ScriptManager, comptime from_parser: bool, script_e
     };
 
     const is_blocking = mode == .normal;
-    if (is_blocking == false) {
+
+    // Once parsing is done, the deferred-script batch has already drained and
+    // won't run again, so a non-blocking script inserted afterwards would go
+    // unprocessed. Run it immediately instead. Remote scripts still need to
+    // be queued so they execute when their fetch completes.
+    const run_immediately = is_blocking or (self.base.static_scripts_done and remote_url == null);
```
krichprollsch, line 218: it's not clear to me: `run_immediatly` is `true` when `is_blocking` is `true`?
but when the title of the PR says: 

> Immediately run deferred inline scripts after "load" is fired

 I understand it will not be run immedatly when the browser is blocking... 

But maybe I'm just wrong?

karlseguin: `is_blocking` refers to whether or not the script tag should be a blocking load. Before this commit, `is_blocking` and ONLY `is_blocking` meant `run_immediatly`. This commits expands the cases where we'll `run_immediatly`.

The issue came up because of a `<script type=module>...</script>` which was dynamically loaded AFTER the `load` event fired. If it happens _before_ the `load` event fires, it isn't a problem. But when it happens after, there's a chance we _never_ run the script. This is what the PR title is referring to. So even after a load happens, we're still executing all scripts (including deferred inline scripts).

krichprollsch: oh OK, git it now. thanks.

krichprollsch (approved):

PR #2546, opened 2026-05-26
Unify Data URLs

karlseguin:
Previously, the logic for data/blob URLs were spread out at every http_client caller. fetch/XHR/ScriptManager all had their own "if this is a data url, ..." logic.

This causes two issues.
1 - Duplication, particularly as we try to cover more edge cases that need to
    be handled in all places.

2 - Correctness because data/blob URLs are still URLs and still need to be
    "fetched" (from memory). They should still fire with the same timing as any
    other URL. That means that for fetch/XHR, they should fire asynchronously
    (i.e. on the next tick). And for ScriptManager they should fire depending
    on the type of script (normal/defer/async).

This PR relies on the infrastructure added to:
https://github.com/lightpanda-io/browser/pull/2506 in order to fulfilled a synthetic response on the next tick.

This also fixes issues with our base64 decoding of data urls. WPT tests brought up various edge cases that `std` doesn't handle, because it's supposed to be a "Forgiving-base64 decode" algorithm. I had claude write the implementation based on the WPT tests and the spec. So while the code generally removed duplicate, the added LOC are still more due to this custom base64 and new tests.

Frame.navigate is excluded from this refactor. For one, about:blank must be special-cased and run synchronously (one of the few places where this is strictly required) and even blob URLs are a bit different: the blob URL list is the parent frame, not self, and there's more we need to do (set origin). Potentially there _is_ some improvement here, but it's both less significant and less simple.

### DIFF
--- a/src/browser/HttpClient.zig
+++ b/src/browser/HttpClient.zig
@@ -567,6 +567,21 @@ fn requestT(self: *Client, req: Request, owner: ?*Owner) !*Transfer {
     // via transfer.abort which fires error_callback and deinits. `.created`
     // means no commit happened — anything else is held by an owner that
     // will clean up.
+
+    // Synthetic schemes never touch the network or the layer chain — they skip
+    // robots/cache/interception and deliver on the next tick
+    if (Synthetic.isSynthetic(req.url)) {
+        // The 2nd transfer is the callback context. We don't actually use it,
+        // we're just sticking transfer in there to have something.
+        self.runNextTick(transfer, transfer, .{ .run = Synthetic.run }) catch |err| {
+            if (transfer.state == .created) {
+                transfer.abort(err);
+            }
+            return err;
+        };
+        return transfer;
+    }
+
     self.entry_layer.request(transfer) catch |err| {
         if (transfer.state == .created) {
             transfer.abort(err);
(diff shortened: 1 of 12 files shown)


src/browser/HttpClient.zig
```diff
@@ -567,6 +567,21 @@ fn requestT(self: *Client, req: Request, owner: ?*Owner) !*Transfer {
     // via transfer.abort which fires error_callback and deinits. `.created`
     // means no commit happened — anything else is held by an owner that
     // will clean up.
+
+    // Synthetic schemes never touch the network or the layer chain — they skip
+    // robots/cache/interception and deliver on the next tick
```
krichprollsch, line 572: I think interception must be called w/ `data:` urls.

I tried w/ `src/browser/tests/frames/data_url_iframe.html` and I get an event for the `data:text/html` url:

```
< {"method":"Network.requestWillBeSent","params":{"requestId":"1EBE5127A5AA76BBC8A9FFAAB4BD1609","loaderId":"1EBE5127A5AA76BBC8A9FFAAB4BD1609","documentURL":"data:text/html,<p>hello</p><script>parent.postMessage(document.querySelector('p').textContent, '*
')</script>","request":{"url":"data:text/html,<p>hello</p><script>parent.postMessage(document.querySelector('p').textContent, '*')</script>","method":"GET","headers":{"Referer":"http://127.0.0.1:1234/","Upgrade-Insecure-Requests":"1"},"mixedContentType":"
none","initialPriority":"VeryHigh","referrerPolicy":"strict-origin-when-cross-origin","isSameSite":false},"timestamp":20435.808338,"wallTime":1779793000.293844,"initiator":{"type":"script","stack":{"callFrames":[{"functionName":"","scriptId":"7","url":"ht
tp://127.0.0.1:1234/frames/data_url_iframe.html","lineNumber":16,"columnNumber":29}]}},"redirectHasExtraInfo":false,"type":"Document","frameId":"5E8E7E8972A515D1F9790D154A1C7A4D","hasUserGesture":false},"sessionId":"3283BF41A5EDD25878C234FFC68180D1"}
< {"method":"Page.domContentEventFired","params":{"timestamp":20435.807414},"sessionId":"3283BF41A5EDD25878C234FFC68180D1"}
```

karlseguin: You mean in Chrome, you get them?  They aren't passed now to HttpClient, so this isn't a new bug. Could be done in a follow up.

krichprollsch: > You mean in Chrome, you get them?

Yes I get them w/ Chrome, I agree we don't dispatch them correctly right now. A followup is good for me.

PR #2562, opened 2026-05-27
Implement CookieStore web API

krichprollsch:
https://developer.mozilla.org/en-US/docs/Web/API/CookieStore

### DIFF
--- a/src/browser/webapi/event/CookieChangeEvent.zig
+++ b/src/browser/webapi/event/CookieChangeEvent.zig
@@ -1,0 +1,90 @@
+// Copyright (C) 2023-2026  Lightpanda (Selecy SAS)
+//
+// Francis Bouvier <francis@lightpanda.io>
+// Pierre Tachoire <pierre@lightpanda.io>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+const std = @import("std");
+const lp = @import("lightpanda");
+
+const js = @import("../../js/js.zig");
+const Frame = @import("../../Frame.zig");
+
+const Event = @import("../Event.zig");
+const CookieStore = @import("../storage/CookieStore.zig");
+
+const String = lp.String;
+const Allocator = std.mem.Allocator;
+
+// https://developer.mozilla.org/en-US/docs/Web/API/CookieChangeEvent
+const CookieChangeEvent = @This();
+
+_proto: *Event,
+_changed: []*CookieStore.CookieListItem,
+_deleted: []*CookieStore.CookieListItem,
+
+const CookieChangeEventOptions = struct {
+    changed: ?[]*CookieStore.CookieListItem = null,
+    deleted: ?[]*CookieStore.CookieListItem = null,
+};
+
+const Options = Event.inheritOptions(CookieChangeEvent, CookieChangeEventOptions);
+
+pub fn init(typ: []const u8, _opts: ?Options, frame: *Frame) !*CookieChangeEvent {
+    const arena = try frame.getArena(.tiny, "CookieChangeEvent");
+    errdefer frame.releaseArena(arena);
+    const type_string = try String.init(arena, typ, .{});
+
+    const opts = _opts orelse Options{};
+
+    const event = try frame._factory.event(
+        arena,
+        type_string,
+        CookieChangeEvent{
+            ._proto = undefined,
+            ._changed = opts.changed orelse &.{},
+            ._deleted = opts.deleted orelse &.{},
+        },
+    );
+
+    Event.populatePrototypes(event, opts, false);
+    return event;
+}
+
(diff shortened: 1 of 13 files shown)


src/browser/webapi/event/CookieChangeEvent.zig
```diff
@@ -0,0 +1,139 @@
...
+        .secure = snapshot.secure,
+        .sameSite = switch (snapshot.same_site) {
+            .strict => .strict,
+            .lax => .lax,
+            .none => .none,
+        },
+        .partitioned = false,
+    };
+
+    const items = try arena.dupe(*CookieStore.CookieListItem, &.{item});
```
karlseguin: If the goal is to turn the single value into a slice, you can do it allocation-free, since a single item and a slice.len == 1 have the same memory layout. In fact, I'm pretty sure Zig will let you do:

```zig
._changed = if (kind == .changed) item[0..1] else &.{},
._deleted = if (kind == .deleted) item[0..1] else &.{},
```

src/browser/webapi/storage/CookieStore.zig
```diff
@@ -0,0 +1,499 @@
...
+// not delivered.
+_frame: ?*Frame = null,
+
+pub fn asEventTarget(self: *CookieStore) *EventTarget {
+    return self._proto;
+}
+
+/// Registers this CookieStore as a listener for jar-change notifications on
+/// the given frame's session. Must be called once after construction by the
+/// owning Window. Idempotent on re-call (replaces the frame).
```
karlseguin: "replaces the frame" ??

src/browser/webapi/storage/CookieStore.zig
```diff
@@ -0,0 +1,499 @@
...
+
+const FunctionSetter = union(enum) {
+    func: js.Function.Global,
+    anything: js.Value,
+};
+
+// https://developer.mozilla.org/en-US/docs/Web/API/CookieStore/get
+const GetOptions = struct {
+    name: ?[]const u8 = null,
+    url: ?[]const u8 = null,
```
karlseguin, line 205: This is never used..is that intentional?

src/browser/webapi/storage/CookieStore.zig
```diff
@@ -0,0 +1,499 @@
...
+pub const JsApi = struct {
+    pub const bridge = js.Bridge(CookieStore);
+
+    pub const Meta = struct {
+        pub const name = "CookieStore";
+        pub const prototype_chain = bridge.prototypeChain();
+        pub var class_id: bridge.ClassId = undefined;
+    };
+
+    pub const get = bridge.function(CookieStore.get, .{});
```
karlseguin: I think most (all?) need `.dom_exception = true`.

src/browser/webapi/storage/CookieStore.zig
```diff
@@ -0,0 +1,499 @@
...
+    secure: bool,
+    same_site: Cookie.SameSite,
+
+    fn cancelled(ctx: *anyopaque) void {
+        const self: *ChangeCallback = @ptrCast(@alignCast(ctx));
+        self.releaseArena();
+    }
+
+    fn releaseArena(self: *ChangeCallback) void {
+        if (self.cookie_store._frame) |frame| {
```
karlseguin: There's a potential leak on shutdown here. If detach is called, it sets `_frame = null`. So detach is called first, then cancelled is called, the arena will never be freed.

The solution might be to hold Session so you can release the arena.

src/browser/webapi/event/CookieChangeEvent.zig
```diff
@@ -0,0 +1,139 @@
...
+_deleted: []*CookieStore.CookieListItem,
+
+const CookieChangeEventOptions = struct {
+    changed: ?[]*CookieStore.CookieListItem = null,
+    deleted: ?[]*CookieStore.CookieListItem = null,
+};
+
+const Options = Event.inheritOptions(CookieChangeEvent, CookieChangeEventOptions);
+
+pub fn init(typ: []const u8, _opts: ?Options, frame: *Frame) !*CookieChangeEvent {
```
karlseguin: This will crash if called from a `Worker`. Last parameter should be a `*js.Execution`

src/browser/webapi/storage/CookieStore.zig
```diff
@@ -0,0 +1,499 @@
...
+// https://developer.mozilla.org/en-US/docs/Web/API/CookieStore
+const CookieStore = @This();
+
+_proto: *EventTarget,
+_on_change: ?js.Function.Global = null,
+// Owning Frame, used to filter incoming change notifications by document
+// scope and to schedule async event dispatch. Null when the store has no
+// frame context (e.g. WorkerGlobalScope), in which case change events are
+// not delivered.
+_frame: ?*Frame = null,
```
karlseguin: Events _should_ be delivered to Workers. I think you should take/store the `*js.Execution`. 

I've tried to improve the ergonomics of Execution in https://github.com/lightpanda-io/browser/pull/2564

src/browser/webapi/storage/CookieStore.zig
krichprollsch: well, idk :shrug:  removed.

src/browser/webapi/event/CookieChangeEvent.zig
krichprollsch: It worked but I had to use slice of `CookieStore.CookieListItem` not `*CookieStore.CookieListItem`.

src/browser/webapi/storage/CookieStore.zig
krichprollsch: It's implemented now, thanks.
It's useful only with workers.

krichprollsch:
I updated the PR. I'm sorry I pushed to fast yesterday, I should set as draft until I auto-review/fix some changes.
But thanks again for the review.

PR #2570, opened 2026-05-28
Cache Revalidation

mookums:
This adds Cache Revalidation, properly checking `no-cache` and stale entries.

### DIFF
--- a/src/network/cache/Cache.zig
+++ b/src/network/cache/Cache.zig
@@ -55,9 +55,9 @@ pub fn evict(self: *Cache, url: []const u8) void {
     };
 }
 
-pub fn renew(self: *Cache, arena: std.mem.Allocator, url: []const u8, timestamp: i64) !void {
+pub fn renew(self: *Cache, arena: std.mem.Allocator, req: RenewRequest) !void {
     return switch (self.kind) {
-        inline else => |*c| c.renew(arena, url, timestamp),
+        inline else => |*c| c.renew(arena, req),
     };
 }
 
@@ -173,12 +173,38 @@ pub const CachedMetadata = struct {
     pub fn hasValidators(self: CachedMetadata) bool {
         return self.etag != null or self.last_modified != null;
     }
+
+    pub fn renew(self: *CachedMetadata, req: RenewRequest) void {
+        self.stored_at = req.timestamp;
+        self.age_at_store = 0;
+
+        for (req.headers) |h| {
+            const name = h.name;
+            const value = h.value;
+
+            if (std.ascii.eqlIgnoreCase("Age", name)) {
+                self.age_at_store = std.fmt.parseInt(u64, value, 10) catch 0;
+            } else if (std.ascii.eqlIgnoreCase("Cache-Control", name)) {
+                self.cache_control = CacheControl.parse(value) orelse continue;
+            } else if (std.ascii.eqlIgnoreCase("ETag", name)) {
+                self.etag = value;
+            } else if (std.ascii.eqlIgnoreCase("Last-Modified", name)) {
+                self.last_modified = value;
+            }
+        }
+    }
 };
 
 pub const CacheRequest = struct {
     url: []const u8,
     timestamp: i64,
     request_headers: []const Http.Header,
+};
+
+pub const RenewRequest = struct {
+    url: []const u8,
+    timestamp: i64,
+    headers: []const Http.Header,
 };
 
 pub const CachedData = union(enum) {
--- a/src/network/cache/Cache.zig
+++ b/src/network/cache/Cache.zig
@@ -69,6 +69,7 @@ pub fn clear(self: *Cache) !void {
 
 pub const CacheControl = struct {
     max_age: u64,
+    must_revalidate: bool = false,
 
     pub fn parse(value: []const u8) ?CacheControl {
         var cc: CacheControl = .{ .max_age = undefined };
@@ -88,7 +89,8 @@ pub const CacheControl = struct {
                 return null;
             }
             if (std.mem.eql(u8, directive, "no-cache")) {
-                return null;
+                cc.must_revalidate = true;
+                continue;
             }
             if (std.mem.eql(u8, directive, "private")) {
                 return null;
@@ -132,15 +134,17 @@ pub const CachedMetadata = struct {
     vary_headers: []const Http.Header,
 
     // Validators for conditional requests.
-    etag: ?[]const u8 = null,
-    last_modified: ?[]const u8 = null,
+    etag: ?[]const u8,
+    last_modified: ?[]const u8,
 
     pub fn format(self: CachedMetadata, writer: *std.Io.Writer) !void {
-        try writer.print("url={s} | status={d} | content_type={s} | max_age={d} | vary=[", .{
+        try writer.print("url={s} | status={d} | content_type={s} | max_age={d} | etag={s} | last-modified={s} | vary=[", .{
             self.url,
             self.status,
             self.content_type,
             self.cache_control.max_age,
+            self.etag orelse "null",
+            self.last_modified orelse "null",
         });
 
         // Logging all headers gets pretty verbose...
@@ -156,6 +160,7 @@ pub const CachedMetadata = struct {
     }
 
     pub fn isStale(self: CachedMetadata, timestamp: i64) bool {
+        if (self.cache_control.must_revalidate) return true;
         const age = (timestamp - self.stored_at) + @as(i64, @intCast(self.age_at_store));
         return age >= @as(i64, @intCast(self.cache_control.max_age));
     }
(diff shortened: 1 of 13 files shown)


src/network/cache/Cache.zig
```diff
@@ -55,6 +55,12 @@ pub fn evict(self: *Cache, url: []const u8) void {
     };
 }
 
+pub fn revalidate(self: *Cache, arena: std.mem.Allocator, url: []const u8, timestamp: i64) !void {
```
krichprollsch: I find this name confusing. I was expecting a mechanism to check the cache's validity, but as far as i can read from FSCache.revalidate, `revalidate` prorogates the cache.
I think I would rename into something like prorogate + add a explanation comment.

src/network/layer/CacheLayer.zig
```diff
@@ -67,48 +67,79 @@ fn request(ptr: *anyopaque, transfer: *Transfer) anyerror!void {
...
-        try transfer.client.runNextTick(transfer, ctx, .{
-            .run = struct {
-                fn run(t: *Transfer, ctx_ptr: *anyopaque) void {
-                    defer t.deinit();
-
-                    const c: *CachedResponse = @ptrCast(@alignCast(ctx_ptr));
-                    serveFromCache(&t.req, c) catch |err| {
-                        t.req.error_callback(t.req.ctx, err);
-                    };
+        if (cached.expired) {
```
krichprollsch: I think you can change some `if` order to have less imbrications.
The first one could be to manage the cache missing case first:
```zig
    const _cache = transfer.client.network.cache.?.get(arena, .{
        .url = req.url,
        .timestamp = std.time.timestamp(),
        .request_headers = req_header_list.items,
    }));
    if (_cache == null) {
        // Cache miss: install wrappers so we can inspect the response and decide
        // whether to write the body into the cache when it's done.
        try installCacheContext(arena, transfer, null);
        return self.next.request(transfer);
    }
```

karlseguin:
I tested it and it doesn't work. Wrong results + crash.

krichprollsch:
@mookums ~Do you have a corresponding PR on demo to add more tests?~ found https://github.com/lightpanda-io/demo/pull/185

src/network/layer/CacheLayer.zig
```diff
@@ -98,15 +100,41 @@ fn request(ptr: *anyopaque, transfer: *Transfer) anyerror!void {
...
+            try req.headers.add(header_value);
+        } else if (cached.metadata.last_modified) |lm| {
+            log.debug(.cache, "revalidate with last-modified", .{ .url = req.url, .last_modified = lm });
+            const header_value = try std.fmt.allocPrintSentinel(arena, "If-Modified-Since: {s}", .{lm}, 0);
+            try req.headers.add(header_value);
+        }
+
+        try installCacheContext(arena, transfer, cached);
+    } else {
+        // If it is expired w/o validators, evict from Cache.
```
karlseguin, line 119: I think this leaks the FD.

src/network/layer/CacheLayer.zig
```diff
@@ -177,6 +243,7 @@ const CacheContext = struct {
     req_url: [:0]const u8,
     req_headers: @import("../http.zig").Headers,
     pending_metadata: ?*CachedMetadata = null,
+    stale_entry: ?CachedResponse = null,
```
karlseguin, line 249: What happens to this on the error/shutdown path?

src/network/layer/CacheLayer.zig
```diff
@@ -98,15 +100,41 @@ fn request(ptr: *anyopaque, transfer: *Transfer) anyerror!void {
...
     }
 
-    // Cache miss: install wrappers so we can inspect the response and decide
-    // whether to write the body into the cache when it's done.
+    if (cached.metadata.hasValidators()) {
+        if (cached.metadata.etag) |etag| {
+            log.debug(.cache, "revalidate with etag", .{ .url = req.url, .etag = etag });
+            const header_value = try std.fmt.allocPrintSentinel(arena, "If-None-Match: {s}", .{etag}, 0);
+            try req.headers.add(header_value);
+        } else if (cached.metadata.last_modified) |lm| {
```
krichprollsch: It seems Chrome sends both if available:

Image

Tested w/ https://httpbin.io/cache

src/network/cache/Cache.zig
```diff
@@ -82,7 +89,8 @@ pub const CacheControl = struct {
                 return null;
             }
             if (std.mem.eql(u8, directive, "no-cache")) {
-                return null;
+                cc.must_revalidate = true;
```
krichprollsch, line 97: if we get only `Cache-Control: no-cache`, we will never cache the file b/c we won't have max-age isn't it?
What about setting `max_age=0` if not already set in this case?

krichprollsch:
I think we should apply immediately https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Caching#heuristic_caching

But it may be challenging to parse the `Last-Modified` header. Maybe we could go simpler, w/ a short max-age (10min?) + consider the cache control is `no-cache` and only *if* the response contains either a `Last-Modifed` or an `etag`.

mookums:
Duration regression is from the cache tests being added to puppeteer. Any objections to just upping the average time in the CI test?

krichprollsch:
Duration is about `puppeteer/cdp.js` test only. Your change shouldn't have an impact on it, cache is not enable :thinking:

src/network/layer/Forward.zig
```diff
@@ -73,3 +74,25 @@ pub fn forwardErr(self: Forward, e: anyerror) void {
 pub fn forwardShutdown(self: Forward) void {
     if (self.shutdown) |cb| cb(self.ctx);
 }
+
+pub fn noop() Forward {
```
krichprollsch: Claude complains it's deadcode 🤔 
> Forward.noop() is still dead code with a stale signature (Forward.zig:83 returns anyerror!bool, not HeaderResult). Compiles only because it's unreferenced; will break the moment anyone calls it. Delete it or change to return .proceed.

src/network/cache/FsCache.zig
```diff
@@ -294,6 +298,57 @@ pub fn evict(self: *FsCache, url: []const u8) void {
...
+        arena,
+        &json_reader,
+        .{ .allocate = .alloc_always },
+    ) catch |e| {
+        log.warn(.cache, "renew parse", .{ .url = url, .err = e });
+        return e;
+    };
+
+    parsed.metadata.stored_at = timestamp;
+    parsed.metadata.age_at_store = 0;
```
krichprollsch: Claude says we should update more than max-age from the revalidation request:

> 304 does not update stored metadata/headers (RFC 9111 §4.3.4)
>
>  src/network/cache/FsCache.zig:301-350 (renew) only resets stored_at and age_at_store = 0. A 304 Not Modified can carry updated Cache-Control (new max-age), a new ETag, updated Expires, etc., which per spec should be merged into the stored entry. Here
  they're discarded, so a server extending/shortening freshness or rotating the ETag won't take effect until the entry is fully re-fetched. Functionally serves correct content, but freshness drifts from what the origin intends. Worth at least a tracking
  note.

krichprollsch (commented):
Here is another comment from Claude about DeferredLayer (lines are outdated)
Not sure if it's correct r not 🤔 

>   - DeferringLayer doesn't propagate .handled — its deferred-replay paths (DeferringLayer.zig:283- and :333-) only branch on .abort vs not, treating .handled like .proceed. If a CacheLayer 304 (.handled) ever sits inside a DeferringLayer, the replay would  forward header and continue replaying data/done, double-forwarding. Likely not a current combination, but the asymmetry is a latent trap.

src/network/layer/CacheLayer.zig
```diff
@@ -200,11 +270,37 @@ const CacheContext = struct {
...
         const conn = transfer._conn.?;
-        const vary = if (conn.getResponseHeader("vary", 0)) |h| h.value else null;
-
         var rh = &transfer.res.header.?;
+
+        if (self.stale_entry != null and rh.status == 304) {
+            const stale = self.stale_entry.?;
+            self.stale_entry = null;
+
+            var iter = self.req_headers.iterator();
```
krichprollsch: You should use the response's headers instead of the request's ones to renew the cache isn't it?

src/network/cache/Cache.zig
```diff
@@ -153,6 +201,12 @@ pub const CacheRequest = struct {
     request_headers: []const Http.Header,
 };
 
+pub const RenewRequest = struct {
```
krichprollsch: I think it would be more accurate to rename it in `RenewResponse`

krichprollsch (approved):

krichprollsch:
I think we are good to merge. Anything to change from your POV @karlseguin?

karlseguin:
@krichprollsch LGTM

PR #2571, opened 2026-05-28
storage: persist localStorage/sessionStorage across navigations, fix quota

arrufat:
Move the per-Window storage bucket to an origin-keyed Shed on the Session
so localStorage and sessionStorage survive navigation within an origin,
matching the Web Storage spec.

Also fixes two pre-existing bugs surfaced by this work:
  - setItem's quota counter was incremented on every call, never
    decremented on overwrite — five same-key overwrites tripped the cap
    spuriously. Now subtracts the old value's length first.
  - Shed.getOrPut used allocator.free on a single-pointer allocation
    where allocator.destroy was required, and inserted into _origins
    before its dependent allocations could fail. Reordered so the entry
    is only put once both key dupe and bucket creation have succeeded.

Adds an MCP test that round-trips localStorage between two origins via
the eval tool to lock in the persistence + isolation contract.

### DIFF
--- a/src/browser/webapi/Window.zig
+++ b/src/browser/webapi/Window.zig
@@ -71,7 +71,6 @@ _model_context: ModelContext = .init,
 _screen: *Screen,
 _visual_viewport: *VisualViewport,
 _performance: Performance,
-_storage_bucket: storage.Bucket = .{},
 _on_load: ?js.Function.Global = null,
 _on_pageshow: ?js.Function.Global = null,
 _on_popstate: ?js.Function.Global = null,
@@ -233,12 +232,19 @@ pub fn getPerformance(self: *Window) *Performance {
     return &self._performance;
 }
 
+fn bucketForOrigin(self: *Window) *storage.Bucket {
+    return self._frame._session.storage_shed.getOrPut(
+        self._frame._session.browser.app.allocator,
+        self.getOrigin(),
+    ) catch @panic("OOM");
+}
+
 pub fn getLocalStorage(self: *Window) *storage.Lookup {
-    return &self._storage_bucket.local;
+    return &self.bucketForOrigin().local;
 }
 
 pub fn getSessionStorage(self: *Window) *storage.Lookup {
-    return &self._storage_bucket.session;
+    return &self.bucketForOrigin().session;
 }
 
 pub fn getOrigin(self: *const Window) []const u8 {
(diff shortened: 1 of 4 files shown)


src/browser/webapi/Window.zig
```diff
@@ -233,12 +232,19 @@ pub fn getPerformance(self: *Window) *Performance {
     return &self._performance;
 }
 
+fn bucketForOrigin(self: *Window) *storage.Bucket {
+    return self._frame._session.storage_shed.getOrPut(
+        self._frame._session.browser.app.allocator,
+        self.getOrigin(),
```
karlseguin: I think you should use `self._frame.js.origin.key` here, else 2 opaque origins (about:blank) will share the same bucket. `self._frame.js.origin.key` will be a UUID for these cases.

src/browser/webapi/storage/storage.zig
```diff
@@ -35,64 +35,100 @@ pub const Shed = struct {
...
         }
         self._origins.deinit(allocator);
     }
 
     pub fn getOrPut(self: *Shed, allocator: Allocator, origin: []const u8) !*Bucket {
-        const gop = try self._origins.getOrPut(allocator, origin);
-        if (gop.found_existing) {
-            return gop.value_ptr.*;
-        }
+        if (self._origins.get(origin)) |existing| return existing;
```
karlseguin: I don't mind this change, and I would have merged it as-is..but since I had another more important comment (the origin)...

This does result in 2 hash lookups when it could be one. That's the beauty of `getOrPut`. I think the only reason to avoid `getOrPut` is if the map can be altered between the get/write operation, which invalidates the gop pointers.

src/browser/webapi/storage/storage.zig
```diff
@@ -35,64 +35,100 @@ pub const Shed = struct {
...
+    }
+
     pub fn getItem(self: *const Lookup, key_: ?[]const u8) ?[]const u8 {
         const k = key_ orelse return null;
         return self._data.get(k);
     }
 
     pub fn setItem(self: *Lookup, key_: ?[]const u8, value: []const u8, frame: *Frame) !void {
         const k = key_ orelse return;
+        const allocator = frame._session.browser.app.allocator;
```
karlseguin: Not against storing an Allocator in the Lookup (self.allocator).

PR #2575, opened 2026-05-29
Improve WPT /url/ tests

karlseguin:
This is a bit all over the place.

1 - Replace libidn2 with rust-idna. It looks like there are different idna
    profiles, and rust-idna (from the servo project) implements the whatwg
    one. libidn2 would be too strict in some cases and not strict enough in
    others. (Gemini says I could use libidn2 for this, but what it suggested
    didn't work, and I couldn't figure it out myself, and claude insisted it
    _did not_ have the correct implementation for what we want).

2 - We previously only ran a URL through idna if it wasn't ascii. Turns out
    we also need to run it if there's a "xn--" (aka, an IDNA ACE prefix) in
    there. This helps us pass hundreds of WPT cases, and it's pretty cheap.

3 - Implement more of the Area WebAPI. Mostly copied from Anchor.

4 - Add username/password accessor to Anchor/Area

5 - window.open validates the URL (i.e. tries to resolve it and handles the
    error)

6 - Invalid idna conversion maps to a TypeError

7 - Cleanup closed popups on the next tick (like destroyed pages), rather than
    at an interval or on shutdown. This one seems unrelated, but some of these
    tests are opening hundreds (thousands?) of popups and then closing them.


src/browser/Session.zig
```diff
@@ -70,7 +70,8 @@ _active: ?*Page = null,
 // In-flight root navigation
 _pending: ?*Page = null,
 
-_queued_destroy: std.ArrayList(*Page) = .{},
+_page_descruction_queue: std.ArrayList(*Page) = .{},
```
krichprollsch:
```suggestion
_page_destruction_queue: std.ArrayList(*Page) = .{},
```

src/browser/Session.zig
```diff
@@ -137,12 +138,27 @@ pub fn deinit(self: *Session) void {
...
+    }
+
+    {
+        const queue = self._page_descruction_queue.items;
+        if (queue.len > 0) {
+            for (queue) |page| {
+                page.deinit();
+                self.browser.page_pool.destroy(page);
+            }
+            self._page_descruction_queue.clearRetainingCapacity();
```
krichprollsch:
```suggestion
            self._page_destruction_queue.clearRetainingCapacity();
```

src/browser/Session.zig
```diff
@@ -161,8 +177,12 @@ fn allocatePage(self: *Session, frame_id: u32) !*Page {
 }
 
 // Tear down and free a Page allocated via allocatePage.
-fn destroyPage(self: *Session, page: *Page) void {
-    self._queued_destroy.append(self.arena, page) catch @panic("OOM");
+fn queuePageDestruction(self: *Session, page: *Page) void {
+    self._page_descruction_queue.append(self.arena, page) catch @panic("OOM");
```
krichprollsch:
```suggestion
    self._page_destruction_queue.append(self.arena, page) catch @panic("OOM");
```

krichprollsch (approved):

PR #2580, opened 2026-05-30
agent: add arrow-key navigation to picker

francisbouvier:
Make the numbered agent choice prompt interactive on TTYs: Up/Down moves the selected row, Enter confirms it, and numeric input still works as before.

Keep the line-based numbered prompt as the non-interactive fallback, restore terminal settings after the raw-mode picker exits, and render raw-mode output with CRLF so menu rows stay aligned. Dim the picker hint text to match existing terminal command hints.

### DIFF
--- a/src/agent/Terminal.zig
+++ b/src/agent/Terminal.zig
@@ -47,6 +47,7 @@ pub const ansi = struct {
     pub const yellow = "\x1b[33m";
     pub const red = "\x1b[31m";
     pub const clear_eol = "\x1b[K";
+    pub const clear_line = "\x1b[2K";
 };
 
 const Verbosity = Config.AgentVerbosity;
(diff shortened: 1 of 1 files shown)


francisbouvier:
There is a bug when the terminal is very small in high and the list is very long (like the model choice).

```
src/agent/Terminal.zig:845: the interactive renderer redraws every item and moves up items.len + 1 lines on each keypress. This is OK for provider selection, but pickModel can pass a long model list.

If the list is taller than the terminal, redraws will not be stable and arrow navigation can leave stale/garbled output. Consider falling back to the line prompt for large lists or implementing a viewport.
```

I think it's an edge case and I fear that handling that will required a lot of changes that could introduce others bugs, but maybe I'm wrong.

arrufat:
Locally, I have added a fallback to the numbered list when it doesn't fit in the terminal, but I am experimenting on how hard it is to implement a viewport. If I am happy with it, I will push it here in a bit.

arrufat:
Closing in favor of #2581

PR #2582, opened 2026-05-30
WebSocket: attach Cookie header on upgrade request

pinetops:
## Summary

WebSocket upgrade requests are HTTP/1.1 requests (RFC 6455 §4.1) and per RFC 6265 §5.4 cookies attach to *any* HTTP request matching the cookie's domain/path. Today `WebSocket.init` builds the upgrade headers without consulting the session jar, so cookie-authenticated WS endpoints (Phoenix LiveView, anything session-gated) reject the upgrade because their auth cookie never arrives.

This PR threads the session jar into the upgrade: read matching cookies via `exec.session.cookie_jar.forRequest` with the same opts shape `HTTPDocument` uses (`is_http: true, is_navigation: false`) and add a `Cookie:` header if any apply.

## Changes

- **`src/browser/webapi/net/WebSocket.zig`** — call into the cookie jar before sending the upgrade; add `Cookie:` header iff any cookies match.
- **`src/TestWSServer.zig`** — capture the upgrade's `Cookie` header value so a fixture can ask the server "what did you receive?" via a new `get-cookie` command. Case-insensitive header lookup.
- **`src/browser/tests/net/websocket.html`** — five regression tests:
  1. **Happy path**: cookie set on the page lands on the upgrade.
  2. **Path scoping (negative)**: cookie scoped to `/restricted` does NOT attach to a WS upgrade for `/`. Guards against over-sending.
  3. **`SameSite=Strict`**: cookie attaches on same-site upgrades. Pins same-site detection.
  4. **Empty jar**: no `Cookie:` header added when nothing matches (exercises the guard clause).
  5. **Multiple cookies**: joined with `; ` per RFC 6265 §5.4.

## Why

Real-world impact: a Phoenix LiveView app behind Lightpanda can't establish its session because the auth cookie is missing from the upgrade. Same applies to any WS endpoint that gates on a cookie. Chromium/Firefox both send cookies on the upgrade — this brings Lightpanda in line.

## Verified

- \`make test\` passes (all 5 new cookie tests green plus existing WebSocket suite).
- \`zig fmt --check\` clean on touched files.
- Built against current \`main\` — \`zig build -Dprebuilt_v8_path=v8/libc_v8.a -Doptimize=ReleaseFast snapshot_creator\` succeeds.

### DIFF
--- a/src/TestWSServer.zig
+++ b/src/TestWSServer.zig
@@ -95,6 +95,22 @@ fn handleClient(client: posix.socket_t) void {
     const key_end = std.mem.indexOfScalarPos(u8, request, key_line_start, '\r') orelse return;
     const key = request[key_line_start..key_end];
 
+    // Capture the request's Cookie header value (if any) so the test
+    // fixture can ask for it back via the `get-cookie` command. Copy out
+    // of `request` because the WS frame loop below reuses `buf` for
+    // incoming frames, invalidating the slice. Buffer is sized to match
+    // `buf` since any cookie that fit in the request fits here.
+    var cookie_buf: [4096]u8 = undefined;
+    var cookie_len: usize = 0;
+    const cookie_header = "Cookie: ";
+    if (std.mem.indexOf(u8, request, cookie_header)) |cookie_start| {
+        const value_start = cookie_start + cookie_header.len;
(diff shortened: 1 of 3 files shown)


pinetops:

karlseguin:
Thanks

PR #2583, opened 2026-05-31
browser/URL: ignore query and fragment slashes in URL resolve

jschaf:
Resolve relative request URLs against only the path component of the base URL. Previously, URL.resolve searched for the last slash in the whole base URL after the authority, so slashes inside a query string or hash route could be mistaken for path directory separators.

The observed failure was a hash-routed page loaded at `http://127.0.0.1:8123/#/login`. When that page ran `fetch("api/users/login", { method: "POST" })`, browser-compatible URL resolution should have requested `http://127.0.0.1:8123/api/users/login`. Instead, Lightpanda 0.3.1 reported the response URL as `http://127.0.0.1:8123/#/api/users/login`, and the HTTP server received `POST /`.

That meant the server returned the single-page app shell instead of the JSON login response. The failure broke a RealWorld/Conduit-style SPA benchmark after login because the app used relative API URLs such as api/users/login. 

The same issue was not limited to fragments. A base URL such as `https://example/app/page?next=/foo/bar` also contains slashes after the path component. Those slashes must not affect how path-relative inputs such as api/users/login or ../api/users/login are merged with the base path.

This change bounds the directory calculation at the first `?` or `#` in the base URL and searches for the last path slash only inside that path component. Root-absolute inputs still keep only the scheme and authority, query-only inputs still replace the query on the current path, and fragment-only inputs still replace the fragment on the current path.

This matches the behavior of the WHATWG URL [algorithm](https://url.spec.whatwg.org/#relative-state), Chromium's relative URL [canonicalization](https://chromium.googlesource.com/chromium/src/+/HEAD/url/url_canon_relative.cc#409), for the cases covered here. All of those implementations split the base URL into components before merging a path-relative reference, so slashes in the query or fragment do not change the base directory.

### DIFF
--- a/src/browser/URL.zig
+++ b/src/browser/URL.zig
@@ -121,7 +121,8 @@ pub fn resolve(allocator: Allocator, base: [:0]const u8, source_path: anytype, o
 
     const scheme_end = std.mem.indexOf(u8, base, "://");
     const authority_start = if (scheme_end) |end| end + 3 else 0;
-    const path_start = std.mem.indexOfScalarPos(u8, base, authority_start, '/') orelse base.len;
+    const path_start = std.mem.indexOfAnyPos(u8, base, authority_start, "/?#") orelse base.len;
+    const path_end = std.mem.indexOfAnyPos(u8, base, path_start, "?#") orelse base.len;
 
     if (path[0] == '/') {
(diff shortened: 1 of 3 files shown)


jschaf:

karlseguin:
Thanks

PR #2593, opened 2026-06-01
Clone data into the transfer's arena

karlseguin:
There can be cases where the data referenced by Request does not live for the duration of a Transfer. The most obvious case is a QueuedNavigation..this is generally a problem, but it's particularly problematic under load when the request gets queued in the HttpClient.

Most of the data is small, so just _always_ duping it is worth it. The body can be large, so we provide a flag for callers to set to turn off the body duping (essentially saying "I promise to keep the body alive for the duration of the transfer). Currently, only XHR is able to make this guarantee.


krichprollsch (approved):
💯

PR #2596, opened 2026-06-01
Robots Deny on Next Tick

mookums:
This fires the `RobotsBlocked` error when the RobotsLayer synchronously denies a request on the next tick.

### DIFF
--- a/src/network/layer/RobotsLayer.zig
+++ b/src/network/layer/RobotsLayer.zig
@@ -70,8 +70,21 @@ fn request(ptr: *anyopaque, transfer: *Transfer) anyerror!void {
                 const path = URL.getPathname(url);
 
                 if (!robots.isAllowed(path)) {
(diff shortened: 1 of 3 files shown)


src/network/layer/RobotsLayer.zig
```diff
@@ -73,8 +73,21 @@ fn request(ptr: *anyopaque, transfer: *Transfer) anyerror!void {
                 const path = URL.getPathname(url);
 
                 if (!robots.isAllowed(path)) {
-                    log.warn(.http, "blocked by robots", .{ .url = url });
-                    return error.RobotsBlocked;
+                    try transfer.client.runNextTick(
+                        transfer,
+                        null,
```
karlseguin, line 75: In `Synthetic`, I passed`transfer` and then ignore it.  If you prefer this way (making it optional), than please also update `Synthetic` so we aren't using both approaches.

PR #2602, opened 2026-06-02
Add Script cache to Runner.waitForScript

karlseguin:
This adds script compilation to the Runner.waitForScript loop. The goal here isn't really to improve waitForScript - we generally expect these scripts to be very simple. The goal is to introduce script caching so that it could be used in more important places (e.g. ScriptManager).

Depends on https://github.com/lightpanda-io/zig-v8-fork/pull/181

### DIFF
--- a/src/browser/js/Local.zig
+++ b/src/browser/js/Local.zig
@@ -172,6 +172,28 @@ pub fn compileFunction(
 }
 
 pub fn compileAndRun(self: *const Local, src: []const u8, name: ?[]const u8) !js.Value {
+    const script = try self.compile(src, name);
+    return script.run();
+}
+
+// Compile `src` into a context-bound Script
+pub fn compile(self: *const Local, src: []const u8, name: ?[]const u8) !js.Script {
+    const result = try self.compileWithCache(src, name, null);
+    return result.script;
+}
+
+pub const CompileResult = struct {
+    script: js.Script,
+    // True only when `cached_data` was supplied AND V8 rejected it (source,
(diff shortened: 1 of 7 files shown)


src/browser/js/Local.zig
```diff
@@ -171,29 +171,61 @@ pub fn compileFunction(
...
+    // unique_ptr), so we must not delete it ourselves — Source__DESTRUCT does.
+    const cached: ?*v8.ScriptCompilerCachedData = if (cached_data) |bytes|
+        v8.v8__ScriptCompiler__CachedData__NEW(bytes.ptr, @intCast(bytes.len))
+    else
+        null;
+
     // Create ScriptCompilerSource
     var script_comp_source: v8.ScriptCompilerSource = undefined;
-    v8.v8__ScriptCompiler__Source__CONSTRUCT2(script_source, &origin, null, &script_comp_source);
+    v8.v8__ScriptCompiler__Source__CONSTRUCT2(script_source, &origin, cached, &script_comp_source);
```
krichprollsch, line 213: Does it make sense to pass a cache here if `v8__ScriptCompiler__CachedData__NEW` rejected it?
Or is it `v8__ScriptCompiler__Source__CONSTRUCT2` which set the `rejected` field?

krichprollsch (approved):

src/browser/js/Local.zig
karlseguin: It's `v8__ScriptCompiler__Compile` (that gets called on the `script_comp_source` we initialized here) that sets it.

PR #2611, opened 2026-06-02
feat(webapi): reflect name and value on HTMLParamElement

rohitsux:
HTMLParamElement is already registered as a concrete element type, but its \`name\` and \`value\` attributes were not reflected. This adds them as plain DOMString reflections per HTML 15.3 (https://html.spec.whatwg.org/multipage/obsolete.html#htmlparamelement), following the same idiom as HTMLTimeElement and HTMLDataElement.

## Changes

- \`src/browser/webapi/element/html/Param.zig\`: add \`name\` and \`value\` getter/setter accessors (empty string default when the attribute is absent).
- \`src/browser/tests/element/html/param.html\`: test fixture covering attribute-to-property and property-to-attribute reflection plus the empty-string default for both attributes.

## Why

Moves the html/dom/reflection-obsolete.html WPT row closer to green.

## Verified

- \`make test\` passes (764/764, no leaks); the new \`WebApi: HTML.Param\` test is included.
- \`zig fmt --check ./*.zig ./**/*.zig\` clean.
- \`make build\` succeeds.

### DIFF
--- a/src/browser/tests/element/html/param.html
+++ b/src/browser/tests/element/html/param.html
@@ -1,0 +1,36 @@
+<!DOCTYPE html>
+<script src="../../testing.js"></script>
+
+<param id="p1" name="n0" value="v0">
+
+<script id="name">
+  {
+    const p = document.getElementById('p1');
+    testing.expectEqual('n0', p.name);
+
+    p.name = 'n1';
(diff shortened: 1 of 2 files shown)


rohitsux:

karlseguin:
Thanks

PR #2621, opened 2026-06-03
build: strip debug symbols in release builds

arrufat:
Strip binaries when optimizing for ReleaseFast or ReleaseSmall to reduce output size. From 144 to 57 MB on Linux.


karlseguin:
How will the crash reports look after this?

arrufat:
Not very good...

arrufat:
I guess people can always `strip` the builds themselves, then.

PR #2623, opened 2026-06-03
Fix double-free of async <script> arena on retried/aborted (204) fetches

### LINKED ISSUE #2622: Double-free / unbounded memory leak: async <script> arena released twice on retried/aborted (204) fetches
## Summary

An `async` `<script>` whose fetch is **retried/aborted with a non-200 (observed: `204`)** can have its terminal callbacks fire **twice** for the same `Script`. Because `Script` is allocated *inside* the arena it later releases, the second terminal callback is a use-after-free that ends in a **double release of the pooled arena**, GP-faulting in `ArenaPool.release`.

- **Debug build:** `General protection exception` → SIGABRT at ~2 GiB, clean trace (below).
- **ReleaseFast:** no fault — the double-release corrupts the `ArenaPool` free-list, so arenas leak unbounded; RSS climbs to **~18 GiB** before the process dies (SIGSEGV/SIGABRT). This is the "memory leak" symptom.

Heavily reproducible on **Internet Archive / Wayback** pages, whose injected scripts (e.g. `https://apollo.archive.org/js/container_*.js`) frequently return 204 / get aborted/retried. Observed ~13 crashes across a 53-task agent benchmark run; correlates 1:1 with sessions that navigate `web.archive.org`.

## Symbolized backtrace (Debug build)

```
WARN http: script fetch error  err=Abort  status=204  req=https://apollo.archive.org/js/container_kzRD6OIl.js  extra=frame
INFO browser: executing script              src=https://apollo.archive.org/js/container_kzRD6OIl.js   ← same script also evaluated
General protection exception (no address available)
  src/ArenaPool.zig:175            release            _ = arena.reset(.{ .retain_with_limit = bucket.retain_bytes });
  src/browser/Session.zig:358      releaseArena
  src/browser/ScriptManagerBase.zig:190  releaseArena
  src/browser/ScriptManagerBase.zig:539  deinit       self.manager.releaseArena(self.arena);
  src/browser/ScriptManagerBase.zig:420  evaluate     defer script.deinit();      ← release #1 (.frame async drained from ready_scripts)
  ...
  src/browser/ScriptManagerBase.zig:692  errorCallback  self.deinit();             ← release #2 (err=Abort path)
```

## Root cause

- `Script` is created in its own pooled arena: `arena.create(Script)` (`ScriptManagerBase.zig:225,325`).
- `Script.deinit()` releases that arena back to the pool (`deinit` → `releaseArena` → `Session.releaseArena` → `ArenaPool.release`).
- Both terminal paths can run for one `Script`:
  - `doneCallback` appends `.frame` async scripts to `ready_scripts`; `evaluate()` drains and `defer script.deinit()`s them (`ScriptManagerBase.zig:420`).
  - `errorCallback` also `self.deinit()`s (`ScriptManagerBase.zig:692`).
- For the same script to be **both executed and errored** (see logs), a terminal callback fired **more than once** — consistent with transfer retries. The code already carries `debug_transfer_id` / `debug_transfer_tries` and an assert commented *"trying to figure out why … headerCallback is being called twice"* (`ScriptManagerBase.zig` ~line 568), so duplicate callbacks in this path are already suspected.
- Second `deinit()` → `ArenaPool.release` does `@fieldParentPtr("arena", …)` + `arena.reset()` on an `Entry` already returned to the pool → GP fault (Debug) / free-list corruption + unbounded arena leak (ReleaseFast).

## Reproduce

Any session that loads a Wayback page with aborted/retried async scripts. Deterministic-ish via the agent on a task that browses `web.archive.org`:

- Debug build: crashed on the **1st** attempt (GP fault, trace above).
- ReleaseFast: crashes ~3/4 attempts; RSS balloons to ~18 GiB first.

## Proposed fix

1. **Primary (transfer layer):** enforce *exactly one terminal callback per request* — once `doneCallback` or `errorCallback` has fired for a `Script`/transfer, drop subsequent done/error for it. (Ties into the existing `header_callback_called` / `debug_transfer_tries` investigation.)
2. **Defense-in-depth (`ArenaPool.release`):** run the existing double-free check (currently ~line 182, `@panic("ArenaPool: double-free detected")`) **before** `arena.reset()` at line 175, so a double-release fails loud (panic) instead of GP-faulting and silently corrupting the pool into an 18 GiB leak.

## Notes

- Not agent-specific: faulting commits `c2a1b6fd` / `a591956dc` ("Split ScriptManager to support worker module imports") are in `main`. The agent harness just exercises it heavily (long multi-navigation sessions hitting archived pages).

arrufat:
Fixes #2622

## Problem

A `Transfer` could deliver **two terminal callbacks** (done *and* error) to the same `ctx` over its lifetime:

- `_notified_fail` only dedups *errors within a single attempt* and is **reset on retry** (`Transfer.reset`).
- `done_callback` was **not latched at all**.

For an `async` `<script>` whose fetch is retried/aborted with a non-200 (observed: `204`), both `Script.doneCallback` (→ `evaluate` → `script.deinit`) and `Script.errorCallback` (→ `script.deinit`) could run for the same `Script`. Since each `deinit` releases the `Script`'s pooled arena, the second release double-frees it:

- **Debug:** GP fault in `ArenaPool.release` (`@fieldParentPtr` + `arena.reset` on a dangling `Entry`).
- **ReleaseFast:** free-list corruption → arenas leak unbounded (RSS → ~18 GiB).

## Fix

**Primary (transfer layer).** Add a `_terminated` latch on `Transfer` that guarantees exactly one terminal callback over the transfer's whole lifetime. Unlike `_notified_fail`, it is **never reset** — legitimate retries/redirects all call `reset()` *before* any terminal callback runs, so they're unaffected. `done_callback` now goes through a new `requestComplete()` that shares the latch with `requestFailed()`, so done and error can't both reach the ctx.

**Defense-in-depth (`ArenaPool.release`).** Run the existing debug double-free check **before** `arena.reset()`, so any future stray double-release panics loudly (`ArenaPool: double-free detected`) instead of GP-faulting on a dangling `Entry` and silently corrupting the pool.

## Testing

- `make test` — 768/768 pass.
- `zig fmt --check` clean on both changed files.

### DIFF
--- a/src/ArenaPool.zig
+++ b/src/ArenaPool.zig
@@ -171,13 +171,11 @@ pub fn release(self: *ArenaPool, allocator: Allocator) void {
     const entry: *Entry = @fieldParentPtr("arena", arena);
     const bucket = entry.bucket;
 
-    // Reset the arena before acquiring the lock to minimize lock hold time
-    _ = arena.reset(.{ .retain_with_limit = bucket.retain_bytes });
-
-    self.mutex.lock();
-    defer self.mutex.unlock();
-
     if (IS_DEBUG) {
+        // Must precede arena.reset below: a double-release hits a dangling
+        // Entry, so resetting first GP-faults before this panic can fire.
+        self.mutex.lock();
+        defer self.mutex.unlock();
         if (self._leak_track.getPtr(entry.debug)) |count| {
             count.* -= 1;
             if (count.* < 0) {
@@ -189,6 +187,12 @@ pub fn release(self: *ArenaPool, allocator: Allocator) void {
             @panic("ArenaPool: release of untracked arena");
         }
     }
+
+    // Reset the arena before acquiring the lock to minimize lock hold time.
+    _ = arena.reset(.{ .retain_with_limit = bucket.retain_bytes });
+
+    self.mutex.lock();
+    defer self.mutex.unlock();
 
     if ((comptime SAFETY) or bucket.free_list_len >= bucket.free_list_max) {
         // In Debug, we never pool. It can mask UAF bugs.
--- a/src/browser/HttpClient.zig
+++ b/src/browser/HttpClient.zig
@@ -1131,7 +1131,7 @@ fn processOneMessage(self: *Client, msg: http.Handles.MultiMessage, transfer: *T
     self.removeConn(msg.conn);
     transfer._conn = null;
 
-    try transfer.req.done_callback(transfer.req.ctx);
+    try transfer.requestComplete();
 
     return true;
 }
@@ -1510,6 +1510,12 @@ pub const Transfer = struct {
 
     _notified_fail: bool = false,
 
+    // Latches on the first terminal callback (done or error) and, unlike
+    // _notified_fail, is never reset across retries. Ensures a ctx that frees
+    // itself in that callback (e.g. a Script releasing its pooled arena) gets
+    // exactly one terminal callback over the transfer's lifetime, never two.
+    _terminated: bool = false,
+
     // Set when conn is temporarily detached from transfer during redirect
     // reconfiguration. Used by processMessages to release the orphaned conn
     // if reconfiguration fails. Transient inside the redirect path only.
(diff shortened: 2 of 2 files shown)


karlseguin:
The double trigger is fixed in https://github.com/lightpanda-io/browser/pull/2625  and https://github.com/lightpanda-io/browser/pull/2624 might also be necessary to add stability.

The ArenaPool fix is good. I wonder if we should make your `_terminated` check (which I think is still a good failsafe) `unreachable` in debug?

karlseguin:
Fixed by a combination of https://github.com/lightpanda-io/browser/pull/2625 , https://github.com/lightpanda-io/browser/pull/2624 and https://github.com/lightpanda-io/browser/pull/2667

PR #2624, opened 2026-06-03
Fix two issues with DeferringLayer

karlseguin:
The first is that it can outlive the Transfer and thus has to dupe anything it'll use (e.g. the frame_id off the request).

The second is that flushFrame is reentrant (flushFrame -> fire -> flushFrame) and that results in the active list and list pointers becoming invalid.


krichprollsch (approved):

karlseguin:
Just to add context. The WPT `/fledge/` tests would crash in debug mode. The crash didn't appear to be deterministic, but it would reliably crash at some point during the runs. Weirdly, it wouldn't crash with -concurrency 1 ,but would with 4. There's nothing multi-threaded about this, so I assume it was just the added network latency that helped generate the right timing.

PR #2631, opened 2026-06-03
Sqlite Cache Backend

mookums:
This is a new SQLite backend for the Cache. This is built off of the Cache Revalidation PR (#2570)


src/network/cache/Cache.zig
```diff
@@ -29,6 +30,7 @@ pub const Cache = @This();
 
 kind: union(enum) {
     fs: FsCache,
```
krichprollsch: I'm not sure we should keep the FsCache here. It seems pretty useless now, WDYT?

src/network/Network.zig
```diff
@@ -238,21 +239,38 @@ pub fn init(allocator: Allocator, app: *App, config: *const Config) !Network {
...
+            return e;
+        };
+
+        const cache_path = try std.fmt.allocPrintSentinel(
+            allocator,
+            "{s}/cache.db",
+            .{trimmed_dir},
+            0,
+        );
+        defer allocator.free(cache_path);
```
krichprollsch: What about moving this logic into  config directly? We could create something like `config.httpCacheSqlitePath()` returning an nullable path name? I would remove this code from `Network` which is not really its place and maybe allow us to have a config flag to customize this path.

src/storage/sqlite/Sqlite.zig
```diff
@@ -54,6 +54,85 @@ pub fn deinit(self: *Sqlite, allocator: Allocator) void {
...
+                    }
+                }
+            }
+
+            if (start == migrations.len) {
+                return start;
+            }
+
+            try conn.begin();
+            errdefer conn.rollback() catch {};
```
krichprollsch: no risk to corrupt anything by running rollback after a commit error?

src/storage/sqlite/Sqlite.zig
```diff
@@ -54,6 +54,85 @@ pub fn deinit(self: *Sqlite, allocator: Allocator) void {
...
+                defer rows.deinit();
+
+                while (try rows.next()) |row| {
+                    const id = row.get(i64, 0);
+                    const hash = row.get(i64, 1);
+                    const idx: usize = @intCast(id - 1);
+                    const stored_hash = hashes[idx];
+
+                    if (hash != stored_hash) {
+                        log.err(.storage, "migration hash mismatch", .{
```
krichprollsch: I would like to have a telemetry report in this case...

krichprollsch: Moreover, maybe we could help the use, suggestion in the log the DB is corrupted and he could remove it manually for example.

src/network/cache/SqliteCache.zig
```diff
@@ -0,0 +1,805 @@
...
+    const last_modified = if (entry.get(?[]const u8, 6)) |opt| try arena.dupe(u8, opt) else null;
+
+    var header_rows = try conn.rows(
+        "select name, value, vary from header where url = $1",
+        .{url},
+    );
+    defer header_rows.deinit();
+
+    var headers: std.ArrayList(Http.Header) = .empty;
+    var vary_headers: std.ArrayList(Http.Header) = .empty;
```
krichprollsch: What about allocating immediately the good size of headers? maybe metadata could save the two values? or is it premature optimization?

src/network/cache/SqliteCache.zig
```diff
@@ -0,0 +1,805 @@
...
+            try vary_headers.append(arena, h);
+        } else {
+            try headers.append(arena, h);
+        }
+    }
+
+    return CachedMetadata{
+        .url = try arena.dupeZ(u8, url),
+        .content_type = blk: {
+            for (headers.items) |h| {
```
krichprollsch: what about saving this value during the `while (try header_rows.next()) |row| {` loop to avoid another loop?

src/network/cache/SqliteCache.zig
```diff
@@ -0,0 +1,805 @@
...
+}
+
+fn loadBody(conn: Conn, arena: std.mem.Allocator, url: []const u8) ![]const u8 {
+    const body_entry = try conn.row(
+        "select data from body where url = $1",
+        .{url},
+    ) orelse @panic("valid metadata must have a body");
+    defer body_entry.deinit();
+
+    return try arena.dupe(u8, body_entry.get([]const u8, 0));
```
krichprollsch: Body must be a slice? Any chance, maybe in a following PR, to return a closable reader instead?

src/network/cache/SqliteCache.zig
```diff
@@ -0,0 +1,805 @@
...
+const Conn = @import("../../storage/sqlite/Sqlite.zig").Conn;
+const Migrations = @import("../../storage/sqlite/Sqlite.zig").Migrations;
+
+pub const SqliteCache = @This();
+
+allocator: std.mem.Allocator,
+pool: Pool,
+
+const SqliteCacheMigrations = Migrations(&.{
+    \\ create table metadata (
```
krichprollsch: What about using strict tables? https://sqlite.org/stricttables.html

src/network/cache/SqliteCache.zig
```diff
@@ -0,0 +1,805 @@
...
+    \\      last_modified     text
+    \\ )
+    ,
+    \\ create table body (
+    \\      url               text not null primary key,
+    \\      data              blob not null,
+    \\      foreign key (url) references metadata(url) on delete cascade
+    \\ )
+    ,
+    \\ create table header (
```
krichprollsch, line 61: I'm wondering if we shouldn't concat the headers directly into the `metadata` row:
* we always write headers and metadata at the same time
* we load both at the same time

we could split headers and vary_headers in 2 different columns

src/network/cache/SqliteCache.zig
```diff
@@ -0,0 +1,805 @@
...
+                .reason = "vary mismatch",
+                .header = vary_hdr.name,
+                .expected = vary_hdr.value,
+                .got = incoming,
+            });
+            return null;
+        }
+    }
+
+    const expired = metadata.isStale(req.timestamp);
```
krichprollsch: Can we avoid body load if the data is expired?

src/network/Network.zig
```diff
@@ -238,21 +239,38 @@ pub fn init(allocator: Allocator, app: *App, config: *const Config) !Network {
     else
         null;
 
-    const cache = if (config.httpCacheDir()) |cache_dir_path|
-        Cache{
+    const cache = if (config.httpCacheDir()) |cache_dir_path| blk: {
```
krichprollsch: We must update the http-cache-dir description to explain we will create/use an sqlite file `<DIR>/cache.db` to save cache. And we must explicitly warn against using the same directory across multiple LP processes. Maybe we can detect that? idk

src/network/cache/SqliteCache.zig
```diff
@@ -0,0 +1,805 @@
...
+    \\      vary              bool not null,
+    \\      primary key (url, name),
+    \\      foreign key (url) references metadata(url) on delete cascade
+    \\ )
+    ,
+    "create index header_url on header(url)",
+});
+
+pub fn init(allocator: std.mem.Allocator, path: [:0]const u8) !SqliteCache {
+    var pool = try Pool.init(allocator, path);
```
krichprollsch: `errdefer pool.deinit(allocator)` ?

src/network/Network.zig
```diff
@@ -238,21 +239,38 @@ pub fn init(allocator: Allocator, app: *App, config: *const Config) !Network {
...
-                .fs = FsCache.init(cache_dir_path) catch |e| {
+                .sqlite = SqliteCache.init(allocator, cache_path) catch |e| {
                     log.err(.cache, "failed to init", .{
-                        .kind = "FsCache",
-                        .path = cache_dir_path,
+                        .kind = "SqliteCache",
+                        .path = cache_path,
                         .err = e,
                     });
                     return e;
```
krichprollsch, line 242: Maybe we could show an error + disable the cache in case of error to avoid process hard stop?
but I'm not sure it's a good thing 🤔

src/network/cache/SqliteCache.zig
```diff
@@ -0,0 +1,805 @@
...
+        .etag = etag,
+        .last_modified = last_modified,
+    };
+}
+
+fn loadBody(conn: Conn, arena: std.mem.Allocator, url: []const u8) ![]const u8 {
+    const body_entry = try conn.row(
+        "select data from body where url = $1",
+        .{url},
+    ) orelse @panic("valid metadata must have a body");
```
krichprollsch: idk about that one. I would prefer logging an error and return a null body instead.

src/network/cache/SqliteCache.zig
```diff
@@ -0,0 +1,805 @@
...
+    });
+
+    try conn.exec(
+        "insert into body (url, data) values ($1, $2)",
+        .{ meta.url, body },
+    );
+
+    var lower_name: [256]u8 = undefined;
+    for (meta.headers) |h| {
+        const name = std.ascii.lowerString(lower_name[0..h.name.len], h.name);
```
krichprollsch, line 258: Shouldn't we handle `h.name.len > 256` ?

src/network/cache/SqliteCache.zig
```diff
@@ -0,0 +1,805 @@
...
+        try conn.exec(
+            "insert into header (url, name, value, vary) values ($1, $2, $3, false)",
+            .{ meta.url, name, h.value },
+        );
+    }
+    for (meta.vary_headers) |h| {
+        const name = std.ascii.lowerString(lower_name[0..h.name.len], h.name);
```
krichprollsch, line 266: same about h.name length

src/storage/sqlite/Sqlite.zig
```diff
@@ -54,6 +54,85 @@ pub fn deinit(self: *Sqlite, allocator: Allocator) void {
     self.pool.deinit(allocator);
 }
 
+pub fn Migrations(comptime migrations: []const [:0]const u8) type {
```
krichprollsch: Should we keep src/storage/sqlite/migrations.zig?

krichprollsch:
This is nice to be able to use duckdb CLI to explore the cache data :)

```console
cache D select * from metadata where must_revalidate != 0;
┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┬────────┬────────────┬──────────────┬─────────┬─────────────────┬──────────────────────────────────────┬───────────────────────────────┐
│                                                         url                                                         │ status │ stored_at  │ age_at_store │ max_age │ must_revalidate │                 etag                 │         last_modified         │
│                                                       varchar                                                       │ int64  │   int64    │    int64     │  int64  │      int64      │               varchar                │            varchar            │
├─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┼────────┼────────────┼──────────────┼─────────┼─────────────────┼──────────────────────────────────────┼───────────────────────────────┤
│ https://improving.duckduckgo.com/t/page_home_tagline_impression?2117232&b=unknown&d=d&l=en&p=other                  │    200 │ 1782224467 │            0 │       0 │               1 │ NULL                                 │ Mon, 28 Sep 1970 06:00:00 GMT │
│ https://improving.duckduckgo.com/t/page_home_searchbox_ai_impression?3479694&b=unknown&d=d&l=en&p=other&atb=v539-7… │    200 │ 1782224467 │            0 │       0 │               1 │ NULL                                 │ Mon, 28 Sep 1970 06:00:00 GMT │
│ https://improving.duckduckgo.com/t/page_home_impression?8744613&theme=light&b=unknown&d=d&l=en&p=other&atb=v539-7&… │    200 │ 1782224467 │            0 │       0 │               1 │ NULL                                 │ Mon, 28 Sep 1970 06:00:00 GMT │
│ https://duckduckgo.com/t.js?q=lightpanda&l=us-en&p=1&s=0&dl=en&ct=FR&bing_market=en-US&p_ent=&dp=xXvfRl5mggeuTWh4h… │    200 │ 1782224476 │            0 │       0 │               1 │ NULL                                 │ NULL                          │
│ https://hn.algolia.com/public/main-6e634771f729331a3c3c.js                                                          │    200 │ 1782226454 │            0 │       0 │               1 │ "1755549896.0-2608202-407047307"     │ Mon, 18 Aug 2025 20:44:56 GMT │
│ https://js.driftt.com/include/1782224700000/6nmy7z65gsxc.js                                                         │    200 │ 1782224599 │            0 │       0 │               1 │ W/"d45d26b9900d0bece786bd59b23c5006" │ Thu, 26 Mar 2026 08:37:12 GMT │
│ https://config.gorgias.chat/gorgias-chat-bundle-loader.js?applicationId=15954                                       │    200 │ 1782226616 │         3537 │       0 │               1 │ W/"71a-vOzeu8cLY+IjFLWstPX9XvL10zs"  │ NULL                          │
│ https://www.googleadservices.com/pagead/conversion/971805850/?random=1782224689183&cv=11&fst=1782224689183&fmt=7&b… │    200 │ 1782224690 │            0 │       0 │               1 │ NULL                                 │ NULL                          │
│ https://config.gorgias.chat/bundle-loader/01GYCCC51FF4CA6RW7XJH50SCK?source=shopify1click&shop=kit-and-ace.myshopi… │    200 │ 1782226638 │         3557 │       0 │               1 │ W/"74d-X+O5AqWK9m1LWrn+UvfhH5u8pSY"  │ NULL                          │
│ https://www.googleadservices.com/pagead/conversion/971805850/?random=1782224718058&cv=11&fst=1782224718058&fmt=7&b… │    200 │ 1782224719 │            0 │       0 │               1 │ NULL                                 │ NULL                          │
│ https://www.googleadservices.com/pagead/conversion/971805850/?random=1782224718143&cv=11&fst=1782224718143&fmt=7&b… │    200 │ 1782224720 │            0 │       0 │               1 │ NULL                                 │ NULL                          │
│ https://www.googleadservices.com/pagead/conversion/971805850/?random=1782224755238&cv=11&fst=1782224755238&fmt=7&b… │    200 │ 1782224756 │            0 │       0 │               1 │ NULL                                 │ NULL                          │
│ https://www.googleadservices.com/pagead/conversion/971805850/?random=1782224755314&cv=11&fst=1782224755314&fmt=7&b… │    200 │ 1782224756 │            0 │       0 │               1 │ NULL                                 │ NULL                          │
│ https://improving.duckduckgo.com/t/page_home_tagline_impression?9740447&b=unknown&d=d&l=en&p=other                  │    200 │ 1782225829 │            0 │       0 │               1 │ NULL                                 │ Mon, 28 Sep 1970 06:00:00 GMT │
│ https://improving.duckduckgo.com/t/page_home_searchbox_ai_impression?5392774&b=unknown&d=d&l=en&p=other&atb=v539-7… │    200 │ 1782225829 │            0 │       0 │               1 │ NULL                                 │ Mon, 28 Sep 1970 06:00:00 GMT │
│ https://improving.duckduckgo.com/t/page_home_impression?1190434&theme=light&b=unknown&d=d&l=en&p=other&atb=v539-7&… │    200 │ 1782225829 │            0 │       0 │               1 │ NULL                                 │ Mon, 28 Sep 1970 06:00:00 GMT │
│ https://improving.duckduckgo.com/t/page_home_tagline_impression?8429873&b=unknown&d=d&l=en&p=other                  │    200 │ 1782226412 │            0 │       0 │               1 │ NULL                                 │ Mon, 28 Sep 1970 06:00:00 GMT │
│ https://improving.duckduckgo.com/t/page_home_searchbox_ai_impression?9690011&b=unknown&d=d&l=en&p=other&atb=v539-7… │    200 │ 1782226412 │            0 │       0 │               1 │ NULL                                 │ Mon, 28 Sep 1970 06:00:00 GMT │
│ https://duckduckgo.com                                                                                              │    200 │ 1782226421 │            0 │       0 │               1 │ "6a3a9d36-9c21"                      │ NULL                          │
│ https://improving.duckduckgo.com/t/page_home_tagline_impression?7683633&b=unknown&d=d&l=en&p=other                  │    200 │ 1782226424 │            0 │       0 │               1 │ NULL                                 │ Mon, 28 Sep 1970 06:00:00 GMT │
│ https://improving.duckduckgo.com/t/page_home_searchbox_ai_impression?8002091&b=unknown&d=d&l=en&p=other&atb=v539-7… │    200 │ 1782226424 │            0 │       0 │               1 │ NULL                                 │ Mon, 28 Sep 1970 06:00:00 GMT │
│ https://improving.duckduckgo.com/t/page_home_impression?585745&theme=light&b=unknown&d=d&l=en&p=other&atb=v539-7&p… │    200 │ 1782226424 │            0 │       0 │               1 │ NULL                                 │ Mon, 28 Sep 1970 06:00:00 GMT │
│ https://duckduckgo.com/ac/?q=light&kl=wt-wt&vertical=homepage                                                       │    200 │ 1782226427 │            0 │       0 │               1 │ NULL                                 │ NULL                          │
│ https://duckduckgo.com/ac/?q=lightp&kl=wt-wt&vertical=homepage                                                      │    200 │ 1782226427 │            0 │       0 │               1 │ NULL                                 │ NULL                          │
│ https://duckduckgo.com/ac/?q=lightpa&kl=wt-wt&vertical=homepage                                                     │    200 │ 1782226428 │            0 │       0 │               1 │ NULL                                 │ NULL                          │
│ https://duckduckgo.com/ac/?q=lightpan&kl=wt-wt&vertical=homepage                                                    │    200 │ 1782226428 │            0 │       0 │               1 │ NULL                                 │ NULL                          │
│ https://duckduckgo.com/ac/?q=lightpand&kl=wt-wt&vertical=homepage                                                   │    200 │ 1782226429 │            0 │       0 │               1 │ NULL                                 │ NULL                          │
│ https://duckduckgo.com/ac/?q=lightpanda&kl=wt-wt&vertical=homepage                                                  │    200 │ 1782226429 │            0 │       0 │               1 │ NULL                                 │ NULL                          │
│ https://duckduckgo.com/t.js?q=lightpanda&l=us-en&s=0&dl=en&ct=FR&bing_market=en-US&p_ent=&ex=-1&dp=J9gRog2IaKXlQqQ… │    200 │ 1782226432 │            0 │       0 │               1 │ NULL                                 │ NULL                          │
│ https://duckduckgo.com/post3.html                                                                                   │    200 │ 1782226432 │            0 │       0 │               1 │ "6a3a9d69-88"                        │ NULL                          │
│ https://www.amazon.com/s/ref=nb_sb_noss?url=search-alias%3Daps&field-keywords=ces+jours+qui+disparaissent           │    200 │ 1782226530 │            0 │       0 │               1 │ NULL                                 │ NULL                          │
│ https://js.driftt.com/include/1782226800000/6nmy7z65gsxc.js                                                         │    200 │ 1782226539 │            0 │       0 │               1 │ W/"d45d26b9900d0bece786bd59b23c5006" │ Thu, 26 Mar 2026 08:37:12 GMT │
│ https://www.googleadservices.com/pagead/conversion/971805850/?random=1782226623304&cv=11&fst=1782226623304&fmt=7&b… │    200 │ 1782226624 │            0 │       0 │               1 │ NULL                                 │ NULL                          │
└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┴────────┴────────────┴──────────────┴─────────┴─────────────────┴──────────────────────────────────────┴───────────────────────────────┘
  33 rows
  ```

File: src/network/cache/SqliteCache.zig [hunk 18]
```diff
@@ -0,0 +1,805 @@
...
+const Migrations = @import("../../storage/sqlite/Sqlite.zig").Migrations;
+
+pub const SqliteCache = @This();
+
+allocator: std.mem.Allocator,
+pool: Pool,
+
+const SqliteCacheMigrations = Migrations(&.{
+    \\ create table metadata (
+    \\      url               text not null primary key,
```
krichprollsch, line 48: I'm wondering if using the same hash as migration as primary key wouldn't be better?

src/network/cache/Cache.zig
mookums: I think its fine to get rid of the `FsCache` one as this one serves pretty much the same purpose. It would also reduce maintenance burden.

src/network/Network.zig
karlseguin: I don't think there should be this much caching logic in Network.zig.

src/storage/sqlite/Sqlite.zig
karlseguin: I'm not sure I see the reason for this to be a generic, nor what the hash provides. What case does the hash protect against? The DB getting corrupted? The developer (us) making a build-time mistake? Also, it isn't cryptographic, so whatever it does, it only does sometimes.

It might not matter for a small/simple DB schema, but I think migrations should be code, not text. There are some migrations that simply require pulling data into code.

src/network/cache/SqliteCache.zig
karlseguin: what about:

```sql
create table cache (
  url text not null primary key,
  vary bool not null,
  response text not null,
  body blob not null
)
```

Only `url` and `vary`are queried. `response` can be a JSON representation of the cache.  `body` is extracted from `response` to avoid the base64 bloating of any binary data.

src/network/cache/Cache.zig
krichprollsch: I agree, let's remove it.

src/network/cache/SqliteCache.zig
mookums: At the moment, if the server returns a 304, we immediately serve using the body that we got from the database.

src/network/cache/SqliteCache.zig
mookums: SQLite will return the full row so we'll have the body in memory until we load the next row so the only way I can see it right now is to have it as a slice that we take ownership of

src/storage/sqlite/Sqlite.zig
mookums: There isn't a risk, it'll just throw a SQLITE_ERROR which we catch.

src/storage/sqlite/Sqlite.zig
mookums: We no longer track hash and we also support either just SQL or code as the migration.

src/network/cache/SqliteCache.zig
krichprollsch: what about assert then to get the crash report at least?

krichprollsch:
If we want to cleanup the db at some point, for example to reduce its size, do we have an SQL request to do it easily?

krichprollsch:
I had a crash with integration tests + warmup cache (and a rebased branch)
```
Segmentation fault at address 0x77bc847ab930
/home/pierre/wrk/browser-claude/src/network/cache/Cache.zig:218:42: 0x47343c9 in deinit (lightpanda)
    pub fn deinit(self: CachedData) void {
                                         ^
/home/pierre/wrk/browser-claude/src/network/layer/CacheLayer.zig:203:29: 0x481d669 in serveFromCache (lightpanda)
    defer cached.data.deinit();
                            ^
/home/pierre/wrk/browser-claude/src/network/layer/CacheLayer.zig:84:35: 0x4733cee in run (lightpanda)
                    serveFromCache(t, c) catch |err| {
                                  ^
/home/pierre/wrk/browser-claude/src/browser/HttpClient.zig:508:16: 0x38218ff in drainNextTickQueue (lightpanda)
        ntn.run(transfer, ntn.ctx);
               ^
/home/pierre/wrk/browser-claude/src/browser/HttpClient.zig:456:32: 0x38229ab in tick (lightpanda)
    try self.drainNextTickQueue();
                               ^
/home/pierre/wrk/browser-claude/src/browser/Runner.zig:302:29: 0x44afd00 in _tick__anon_516032 (lightpanda)
        try http_client.tick(@intCast(ms_to_wait), .all);
                            ^
/home/pierre/wrk/browser-claude/src/browser/Runner.zig:150:39: 0x44b00a2 in _wait__anon_516026 (lightpanda)
        const tick_result = self._tick(is_cdp, 200, conditions) catch |err| {
                                      ^
/home/pierre/wrk/browser-claude/src/browser/Runner.zig:75:23: 0x44b050b in waitForFrameCDP (lightpanda)
    _ = try self._wait(true, timeout_ms, &conditions);
                      ^
/home/pierre/wrk/browser-claude/src/cdp/CDP.zig:269:34: 0x44b0685 in pageWait (lightpanda)
    return runner.waitForFrameCDP(page.frame_id, ms, .done);
                                 ^
/home/pierre/wrk/browser-claude/src/cdp/CDP.zig:246:18: 0x44b077b in tick (lightpanda)
    self.pageWait(wait_ms) catch |wait_err| switch (wait_err) {
                 ^
/home/pierre/wrk/browser-claude/src/Server.zig:259:30: 0x435828d in handleConnection (lightpanda)
        const next = cdp.tick() catch |err| {
                             ^
/usr/local/zig-0.15.2/lib/std/Thread.zig:509:13: 0x41f4890 in callFn__anon_441681 (lightpanda)
            @call(.auto, f, args);
            ^
/usr/local/zig-0.15.2/lib/std/Thread.zig:781:30: 0x40e1cc3 in entryFn (lightpanda)
                return callFn(f, args_ptr.*);
                             ^
./nptl/pthread_create.c:447:8: 0x77bd0d69caa3 in start_thread (pthread_create.c)
../sysdeps/unix/sysv/linux/x86_64/clone3.S:78:0: 0x77bd0d729c6b in __clone3 (../sysdeps/unix/sysv/linux/x86_64/clone3.S)
???:?:?: 0x0 in ??? (???)
run
└─ run exe lightpanda failure
error: the following command terminated unexpectedly:
./.zig-cache/o/cdeffea3ddf78e9d9796e5606def3d0f/lightpanda serve --http-cache-dir /tmp/lp-cache

Build Summary: 30/32 steps succeeded; 1 failed
run transitive failure
└─ run exe lightpanda failure

.zig-cache/o/d20cc142e18b434fc8b7b6f062898d2d/build /usr/local/zig-0.15.2/zig /usr/local/zig-0.15.2/lib /home/pierre/wrk/browser-claude .zig-cache /home/pierre/.cache/zig --seed 0x8ae37990 -Zc4c25bdb8f119cba run -- serve --http-cache-dir /tmp/lp-cache
```

src/network/cache/SqliteCache.zig
krichprollsch: this comment has not been addressed AFAIK

src/network/cache/SqliteCache.zig
```diff
@@ -0,0 +1,813 @@
...
+    \\      foreign key (url) references metadata(url) on delete cascade
+    \\ )
+    },
+    .{ .sql = 
+    \\ create table header (
+    \\      url               text not null,
+    \\      name              text not null,
+    \\      value             text not null,
+    \\      vary              bool not null,
+    \\      primary key (url, name),
```
krichprollsch: using url + header name as PK is a problem for response containing the same header multiple time.

src/network/HttpClient.zig
```diff
@@ -837,7 +837,7 @@ fn cacheLookup(self: *Client, transfer: *Transfer) !bool {
     var iter = req.headers.iterator();
     const req_headers = try iter.collect(arena);
 
-    const cached = cache.get(arena, .{
+    const cached = try cache.get(arena, .{
```
krichprollsch: You should maybe catch the error and return a missing cache + a log instead of propagating the error.

src/Config.zig
krichprollsch: dead code now

krichprollsch: Dead code, we should remove `App.storage` and `Storage.zig` I think.

krichprollsch: We should also remove FSCache.zig

src/network/cache/SqliteCache.zig
```diff
@@ -0,0 +1,813 @@
...
+        .{url},
+    ) orelse @panic("valid metadata must have a body");
+    defer body_entry.deinit();
+
+    return try arena.dupe(u8, body_entry.get([]const u8, 0));
+}
+
+fn insertMetadata(conn: Conn, meta: CachedMetadata, body: []const u8) !void {
+    try conn.exec(
+        \\ insert or replace into metadata
```
krichprollsch: Claude complaints the FK w/ headers depends on `pragma foreign_keys=on` conn:

```
 4. put correctness rests on an implicit REPLACE→FK-cascade interaction. insert or replace
  into metadata (:173) only clears the old body/header rows if pragma foreign_keys=on took
  effect on that connection; otherwise the plain insert into body (:188) hits a PK
  conflict. Make it explicit — delete from metadata where url = $1 first — rather than
  depending on a subtle corner of SQLite semantics.
```

src/network/HttpClient.zig
mookums: We catch where it would be missing and handle that separately within the SqliteCache impll. The error being propagated here would be something going wrong with Sqlite.

karlseguin:
On an expired cache hit, we pull the meta data (dupe from sqlite), pull the body (dupe from sqlite), then detect that it's expired and free everything. Because it's an arena, this may or may not be freed, so in the end we're 2x peak memory (the expired body we discarded + the new body we fetch).

This is just the most extreme example, but there are others:
1 - Same applies to header as applies to body
2 - No need to collect the request headers if the URL isn't even in the DB
3 - I don' think there's ever a need to dupe the metadata
4 - renew in particular doesn't need to fetch the metadata (with duping) just to update the row

I think there's a few more. There's a hint of more tightly coupled and more efficient code with the early exit around 'vary mismatch', but even this has duped every header by the time it's checked.

Aside from that, there's still some unused `.file` references and `renew` does a select, followed by update within the same transaction. This is a recipe for an BUSY error.  Should probably use `begin immediate`

src/network/cache/SqliteCache.zig
```diff
@@ -0,0 +1,893 @@
...
+
+    while (try header_rows.next()) |row| {
+        const name = try arena.dupe(u8, row.get([]const u8, 0));
+        const value = try arena.dupe(u8, row.get(Blob, 1).data);
+        if (std.ascii.eqlIgnoreCase(name, "content-type")) content_type = value;
+        try headers.append(arena, .{ .name = name, .value = value });
+    }
+
+    const metadata = CachedMetadata{
+        .url = try arena.dupeZ(u8, req.url),
```
karlseguin: Couldn't find where this is used.

src/network/cache/SqliteCache.zig
```diff
@@ -0,0 +1,893 @@
...
+        .content_type = content_type,
+        .status = status,
+        .stored_at = stored_at,
+        .age_at_store = @intCast(age_at_store),
+        .cache_control = .{
+            .max_age = max_age,
+            .must_revalidate = must_revalidate,
+        },
+        .headers = headers.items,
+        .vary_headers = vary_headers.items,
```
karlseguin: Couldn't find where this is used.

src/network/cache/SqliteCache.zig
```diff
@@ -0,0 +1,893 @@
...
+        last_modified,
+        req.url,
+    });
+
+    var lower_name: [256]u8 = undefined;
+    for (req.headers) |h| {
+        if (h.name.len > lower_name.len) return error.HeaderNameTooLong;
+        const name = std.ascii.lowerString(lower_name[0..h.name.len], h.name);
+        try conn.exec(
+            "insert into header (url, name, value, vary) values ($1, $2, $3, false)",
```
karlseguin, line 363: Does this just keep appending more (potentially duplicate) headers to the same url?

src/network/HttpClient.zig
karlseguin: Which still isn't a reason to fail the request

src/network/cache/SqliteCache.zig
```diff
@@ -0,0 +1,893 @@
...
+
+pub fn deinit(self: *SqliteCache) void {
+    self.pool.deinit(self.allocator);
+}
+
+pub fn get(self: *SqliteCache, arena: std.mem.Allocator, req: CacheRequest) !?CachedResponse {
+    const conn = try self.pool.acquire();
+    defer self.pool.release(conn);
+
+    try conn.begin(.immediate);
```
karlseguin: Serializes all all cache access. In renew, it made sense because you always turned it into an write (and, ideally, writes are less common than reads). Here, the write is the edge-case. It's better to commit the expired read, and then issue the write under a distinct transaction. You can guard against another thread updating the row by changing the delete to included `stored_at = $2` and passing the read `stored_at`.

src/network/cache/SqliteCache.zig
```diff
@@ -0,0 +1,893 @@
...
+        .vary_headers = vary_headers.items,
+        .etag = if (raw_etag) |v| try arena.dupe(u8, v) else null,
+        .last_modified = if (raw_last_modified) |v| try arena.dupe(u8, v) else null,
+    };
+
+    var body_entry = try conn.row(
+        "select data from body where url = $1",
+        .{req.url},
+    ) orelse {
+        log.debug(.cache, "miss", .{ .url = req.url, .reason = "missing body " });
```
karlseguin: But the metadata row remains. This makes my previous comment about using a more concurrent-friendly transaction more complicated to implement (since deleting the metadata here would introduce a 2nd (but still very infrequent, write path). Still possible to clean this up AND have tighter transactions, but not if that end result ends up too spaghetti.

src/Config.zig
```diff
@@ -513,9 +510,29 @@ pub fn userAgent(self: *const Config) ?[]const u8 {
...
-pub fn httpCacheDir(self: *const Config) ?[]const u8 {
+/// This returns the constructed path of the Sqlite DB for the HTTP Cache.
+pub fn httpCacheSqlitePath(self: *const Config, allocator: std.mem.Allocator) !?[:0]const u8 {
     return switch (self.mode) {
-        inline .serve, .fetch, .mcp, .agent => |opts| opts.http_cache_dir,
+        inline .serve, .fetch, .mcp, .agent => |opts| {
+            const cache_dir = opts.http_cache_dir orelse return null;
+
+            // Ensure path exists for SQLite Cache Directory
+            std.fs.cwd().makePath(cache_dir) catch |e| {
```
karlseguin: Belongs in the SqliteCache.init (just like it was in FsCache.init). No other Config getter has this sort of side effect.

src/network/cache/SqliteCache.zig
```diff
@@ -0,0 +1,893 @@
...
+}
+
+pub fn renew(self: *SqliteCache, _: std.mem.Allocator, req: RenewResponse) !void {
+    const conn = try self.pool.acquire();
+    defer self.pool.release(conn);
+
+    try conn.begin(.immediate);
+    defer conn.rollback() catch {};
+
+    const exists = try conn.row("select 1 from metadata where url = $1", .{req.url});
```
karlseguin: This isn't necessary. You can do the update and then check the number of changed rows via `sqlite3_changes`.

src/network/cache/SqliteCache.zig
mookums: Just split it back to how it was where `get` is meant just to get and it now returns an enum where we evict after if needed.

src/network/cache/SqliteCache.zig
mookums: Merged the body into the other row so they're just colocated together now. This should prevent this issue of one existing while the other doesn't.

mookums:
Will fail integration tests until #3044 is merged.

karlseguin (approved):

PR #2635, opened 2026-06-04
Implement input type=file support (FileList, input.files/value, DOM.setFileInputFiles)

### LINKED ISSUE #2175: Implement input type=file support
Implement the ability to upload files on input type=file.

Blocks #1203

Ar-maan05:
## What

Implements `<input type=file>` support, building on the W3C `Blob`/`File` API in the first commit.

- **`FileList`** now backs a real `[]*File` (was a zero-size stub): `length` and `item(i)` work.
- **`input.files`** returns a live, identity-stable `FileList` for `type=file`, and `null` for every other input type (per the IDL).
- **`input.value`** returns the spec `C:\fakepath\<name>` string for file inputs; setting it to anything but `""` throws.
- Required file inputs report `valueMissing` only when empty (was a hardcoded always-empty `TODO`).
- **CDP `DOM.setFileInputFiles`** loads files from disk into a file input, sniffs MIME by extension, and fires `input` + `change`: the upload entry point for WebDriver/automation.

## Memory ownership
  
`File`/`Blob` own a reference-counted pooled arena, normally released when V8 finalizes the JS wrapper. Files set via CDP may never be read from JS, so the owning `Frame` now tracks each input's `FileList`, `acquireRef`s the files it holds, and releases them at teardown, mirroring how `URL.createObjectURL` blobs are managed. Without this they leak an arena (caught by the test runner's leak detector). A scoped `errdefer` also frees partially-created files if one path in a multi-file set fails to load.
  
## Not in scope

- multipart/form-data encoding of files on form submit
- drag-and-drop / `DataTransfer`

## Tests

- `element/html/input_file.html`: `files` identity/length/item, `null` for non-file inputs, `value` getter/setter.
- `cdp.dom: setFileInputFiles`: happy path, non-input node, missing node id, and partial-failure (no-leak) path.
- `make test` green; `zig fmt --check` clean.
  
Closes #2175

### DIFF
--- a/src/browser/webapi/element/html/Input.zig
+++ b/src/browser/webapi/element/html/Input.zig
@@ -31,6 +31,8 @@ const Event = @import("../../Event.zig");
 const InputEvent = @import("../../event/InputEvent.zig");
 const ValidityState = @import("ValidityState.zig");
 const popover = @import("../popover.zig");
+const File = @import("../../File.zig");
+const FileList = @import("../../FileList.zig");
 
 const String = lp.String;
 
@@ -87,6 +89,7 @@ _indeterminate: bool = false,
 _custom_validity: ?[]const u8 = null,
 _validity: ?*ValidityState = null,
 _popover_target: ?*Element = null,
+_files: ?*FileList = null,
 
 _selection_start: u32 = 0,
 _selection_end: u32 = 0,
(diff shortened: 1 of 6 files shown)


Ar-maan05:
Friendly ping! Resolved all merge conflicts. Let me know if any changes are needed.

src/browser/webapi/element/html/Input.zig
```diff
@@ -226,6 +229,54 @@ pub fn getValidity(self: *Input, frame: *Frame) !*ValidityState {
...
+    try frame._event_manager.dispatch(self.asElement().asEventTarget(), change_evt);
+}
+
+/// JS-binding wrapper for the `value` getter: for type=file, return the spec
+/// "C:\\fakepath\\<name>" string; otherwise delegate to plain getValue().
+pub fn getValueForJS(self: *const Input, frame: *Frame) ![]const u8 {
+    if (self._input_type == .file) {
+        const fl = self._files orelse return "";
+        if (fl._files.len == 0) return "";
+        return try std.fmt.allocPrint(frame.arena, "C:\\fakepath\\{s}", .{fl._files[0]._name});
```
karlseguin:
```suggestion
        return std.fmt.allocPrint(frame.call_arena, "C:\\fakepath\\{s}", .{fl._files[0]._name});
```

Better to use the `call_arena`, lives just long enough for the value to get transferred to v8.

karlseguin:
Thank you.

PR #2636, opened 2026-06-04
feat(cdp): fire hover events on Input.dispatchMouseEvent mouseMoved

rohitsux:
CDP `Input.dispatchMouseEvent` with `type: "mouseMoved"` was silently ignored, so `element.hover()` over a Playwright/Puppeteer CDP connection fired no events — breaking hover-dependent flows (tooltips, dropdown menus, `:hover` states). This wires it up. Reported in #2043.

## Changes

- `src/browser/Frame.zig`: add `triggerMouseMove(x, y)` — hit-tests the point with `elementFromPoint` and dispatches trusted `mousemove`, `mouseover` and `mouseenter` events on the target, mirroring `triggerMouseClick` and the event semantics already used by `actions.zig` `hover()`.
- `src/cdp/domains/input.zig`: route `dispatchMouseEvent` `mouseMoved` to `triggerMouseMove` instead of returning early, and add a test that dispatches a `mouseMoved` at an element's position and asserts `mousemove`/`mouseover`/`mouseenter` fire.

This is the enter sequence; `mouseout`/`mouseleave` on leaving a previous element (which needs hover-target tracking) can follow separately.

## Verified

- `make test` passes (769/769, no leaks); new test `cdp.input: dispatchMouseEvent mouseMoved fires hover events`.
- `make build` succeeds.

### DIFF
--- a/src/browser/Frame.zig
+++ b/src/browser/Frame.zig
@@ -3929,6 +3929,44 @@ pub fn triggerMouseClick(self: *Frame, x: f64, y: f64) !void {
     try self._event_manager.dispatch(target.asEventTarget(), mouse_event.asEvent());
 }
 
+pub fn triggerMouseMove(self: *Frame, x: f64, y: f64) !void {
+    const target = (try self.window._document.elementFromPoint(x, y, self)) orelse return;
+    if (comptime IS_DEBUG) {
+        log.debug(.frame, "frame mouse move", .{
+            .url = self.url,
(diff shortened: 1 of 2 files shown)


krichprollsch (approved):

krichprollsch:
thanks!

krichprollsch:
Would be great to have the mouse wheel and mouse release in followup PRs :grimacing:

Issue #2610, opened 2026-06-02
Missing XMLHttpRequest's overrideMimeType and upload used by htmx block login


fabian57:
##  Bug Description

`Set-Cookie` headers from XHR/fetch responses are not being stored in Lightpanda's cookie jar. Cookies set via page navigation work, but cookies set by JavaScript-initiated requests (XMLHttpRequest, fetch) are never persisted.

### Setup

1. A local server at `http://localhost:8080` with an htmx-based login form:
   ```html
   <form hx-post="/api/login" hx-target="this" hx-swap="outerHTML">
     <input type="text" name="username" required>
     <input type="password" name="password" required>
     <button type="submit">Sign In</button>
   </form>
   ```

2. The `/api/login` endpoint responds with `Set-Cookie: session=abc123...; Path=/; HttpOnly; SameSite=Strict` on success.

### Steps to reproduce

```bash
# Use Lightpanda engine
agent-browser --engine lightpanda open http://localhost:8080/login

# Fill credentials
agent-browser fill "[type='text']" user
agent-browser fill "[type='password']" pass

# Click sign in (htmx sends XHR POST to /api/login)
agent-browser click @e4

# Wait for XHR to complete
sleep 2

# Check cookies — returns empty
agent-browser cookies
# (no output)
```

### Expected behavior

`agent-browser cookies` should show the `session` cookie set by the server's `Set-Cookie` response header from the XHR request.

### Actual behavior

`agent-browser cookies` returns nothing. Navigating to an auth-required page redirects to `/login` (no cookie means no session).

### Works with Chrome engine

```bash
agent-browser --engine chrome open http://localhost:8080/login
# (same login flow)
agent-browser cookies
# session  abc123...  (present)
```

##  Environment

- OS: Linux
- agent-browser: 0.27.1
- Lightpanda: 0.3.1

##  Notes

- Cookies from page-level navigation (`document.location` redirects) work fine
- Only XHR/fetch-initiated `Set-Cookie` headers are affected
- The server uses `HttpOnly` and `SameSite=Strict` on the cookie (standard session cookie attributes)
- Tested with a Go + htmx server, but the framework shouldn't matter — any XHR response with a `Set-Cookie` header should store cookies

## Disclaimer:

Everything up to this point is generated by pi. I let it vibe code a small app for testing purposes, so I also let it describe the problem since it probably knows better than I do what happened.
But I also tried running the whole workflow myself. Here is the output

```sh
agent-browser snapshot
[...]

PR #2637, opened 2026-06-04
Implement XMLHttpRequest.overrideMimeType()

krichprollsch:
Adds the overrideMimeType(mime) method.

Wires the final MIME type (override ?? response) into responseXML for the default response type (""): when the final MIME is text/xml, the body is lazily parsed into a Document and cached in a new _response_xml field. Other XML MIME types (application/xml, image/svg+xml) land in Mime.ContentType.other whose backing slices are unsafe to read after Mime.parse returns; supporting them is left for a follow-up.

The override does not affect response / responseText today — responseText charset decoding and the document responseType's HTML-vs-XML parser choice are noted as follow-ups in the code.

Relates with #2610

### DIFF
--- a/src/browser/tests/net/xhr.html
+++ b/src/browser/tests/net/xhr.html
@@ -306,6 +306,98 @@
   }
 </script>
 
+<script id=xhr_override_mime type=module>
+  {
+    // overrideMimeType is callable in state UNSENT (no open() yet)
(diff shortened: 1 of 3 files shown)


karlseguin:
@krichprollsch LGTM, please take a look at my small commit.

PR #2638, opened 2026-06-04
Implement XMLHttpRequest.upload

krichprollsch:
Return an XMLHttpRequestUpload (inheriting XMLHttpRequestEventTarget) from the lazily-cached `upload` attribute. Fixes htmx login flows that called `xhr.upload.addEventListener(...)`.

Relates with #2610


src/browser/webapi/net/XMLHttpRequest.zig
```diff
@@ -289,6 +271,18 @@ pub fn send(self: *XMLHttpRequest, body_: ?BodyInit, exec_: *const Execution) !v
...
 }
 
+// https://xhr.spec.whatwg.org/#the-upload-attribute
+// The XMLHttpRequestUpload object is created lazily and cached: scripts expect
+// the same instance on every access so their event listeners stick.
+pub fn getUpload(self: *XMLHttpRequest) !*XMLHttpRequestUpload {
+    if (self._upload) |upload| {
+        return upload;
+    }
+    const upload = try self._exec._factory.xhrEventTarget(self._arena, XMLHttpRequestUpload{ ._proto = undefined });
```
karlseguin: I agree with using `self._arena`, BUT..it means `XMLHttpRequestUpload` has to participate in `XMLHttpRequest`'s rc.

It needs 3 things:
1 - a reference to XMLHttpRequest
2 - a `acquireRef` that calls `self._xhr.acquireRef();`
3 - a `releaseRef` that decrements `self._xhr.releaseRef();`

krichprollsch: Good catch :+1:  updated

PR #2641, opened 2026-06-04
feat(cdp): fire wheel events on Input.dispatchMouseEvent mouseWheel

rohitsux:
Follow-up to #2636, completing the pair @krichprollsch requested in that review: "Would be great to have the mouse wheel and mouse release in followup PRs". Mouse release is in #2640; this is the mouse wheel half.

CDP `Input.dispatchMouseEvent` with `type: "mouseWheel"` was silently ignored. This wires it to dispatch a `wheel` event at the given point.

## Changes

- `src/browser/Frame.zig`: add `triggerMouseWheel(x, y, deltaX, deltaY)` — hit-tests the point with `elementFromPoint` and dispatches a trusted `wheel` event on the target. If the event is not cancelled, it applies the scroll offset (`scrollLeft`/`scrollTop` += delta) and fires a `scroll` event, mirroring the wheel handling in `WebDriver.zig`.
- `src/cdp/domains/input.zig`: accept `deltaX`/`deltaY` params, route `dispatchMouseEvent` `mouseWheel` to `triggerMouseWheel` instead of returning early, and add a test asserting the `wheel` event fires with the right delta.

## Verified

- `make test` passes (772/772, no leaks); new test `cdp.input: dispatchMouseEvent mouseWheel fires wheel event`.
- `make build` succeeds.

### DIFF
--- a/src/browser/Frame.zig
+++ b/src/browser/Frame.zig
@@ -61,6 +61,7 @@ const popover = @import("webapi/element/popover.zig");
 const NavigationKind = @import("webapi/navigation/root.zig").NavigationKind;
 const KeyboardEvent = @import("webapi/event/KeyboardEvent.zig");
 const MouseEvent = @import("webapi/event/MouseEvent.zig");
+const WheelEvent = @import("webapi/event/WheelEvent.zig");
 
 const HttpClient = @import("HttpClient.zig");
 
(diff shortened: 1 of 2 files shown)


krichprollsch (approved):

src/browser/Frame.zig
```diff
@@ -3941,6 +3942,49 @@ pub fn triggerMouseMove(self: *Frame, x: f64, y: f64) !void {
...
+    wheel_event.asEvent().acquireRef();
+    defer _ = wheel_event.asEvent().releaseRef(self._page);
+    try self._event_manager.dispatch(target.asEventTarget(), wheel_event.asEvent());
+
+    if (wheel_event.asEvent()._prevent_default) {
+        return;
+    }
+
+    // Apply the scroll and fire a trusted scroll event, mirroring WebDriver wheel.
+    const new_left: i32 = @as(i32, @intCast(target.getScrollLeft(self))) + @as(i32, @intFromFloat(delta_x));
```
karlseguin: I know this code mirrors WebDriver. But WebDriver is only exposed to the [trusted] WPT.  The @intFromFloat could fail (NaN) and the addition could overflow. I think a saturating addition is fine (`+|`) so something like:

```zig
fn deltaToScroll(d: f64) i32 {
 if (std.math.isNan(d)) return 0;
 return @intFromFloat(std.math.clamp(d, minInt(i32), maxInt(i32)));
}
...
const new_left: i32 = @as(i32, @intCast(target.getScrollLeft(self))) +| deltaToScroll(delta_x);
```

rohitsux:
Thanks for the careful review. You're right that CDP deltas are untrusted unlike the WebDriver/WPT path. I've applied the saturating addition with a deltaToScroll helper that guards NaN and clamps to the i32 range, and rebased on main (the mouse-release change from #2640 had landed in the meantime).

PR #2644, opened 2026-06-04
telemetry: send agent_replay mode on non-interactive usage

krichprollsch:
Introduce an `interactive()` func from Config which is sent to the telemetry.
Use the `interactive` field in `agent` mode to send a custom `agent_replay` mode to the telemetry.  

add a new `llm` event w/ `provider` and `model`.

Relates with https://github.com/lightpanda-io/telemetry/pull/40

### DIFF
--- a/src/telemetry/telemetry.zig
+++ b/src/telemetry/telemetry.zig
@@ -28,7 +28,7 @@ fn TelemetryT(comptime P: type) type {
 
         const Self = @This();
 
-        pub fn init(app: *App, run_mode: Config.RunMode) !Self {
+        pub fn init(app: *App, run_mode: Config.RunMode, interactive: bool) !Self {
             const disabled = isDisabled();
             if (builtin.mode != .Debug and builtin.is_test == false) {
                 log.info(.telemetry, "telemetry status", .{ .disabled = disabled });
@@ -39,7 +39,7 @@ fn TelemetryT(comptime P: type) type {
             const provider = try app.allocator.create(P);
             errdefer app.allocator.destroy(provider);
 
-            try P.init(provider, app, iid, run_mode);
+            try P.init(provider, app, iid, run_mode, interactive);
 
             return .{
                 .disabled = disabled,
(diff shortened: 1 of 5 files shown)


src/telemetry/telemetry.zig
```diff
@@ -1,4 +1,5 @@
 const std = @import("std");
+const zenai = @import("zenai");
```
karlseguin:
```suggestion
```

karlseguin (approved):

src/telemetry/telemetry.zig
```diff
@@ -113,6 +118,40 @@ pub const Event = union(enum) {
...
         dropped: u32,
     };
+
+    const LLM = struct {
+        provider: [:0]const u8,
+        model: ?Model,
+
+        const Model = struct {
+            len: u8,
+            buffer: [32]u8,
```
arrufat, line 128: some models like `claude-haiku-4-5-20251001` have 25 characters, maybe 64 is a safer bet?

arrufat (approved):

src/telemetry/telemetry.zig
arrufat: I found this one: `gemini-2.5-computer-use-preview-10-2025` at 39 characters...

krichprollsch: Then we will truncate it to 32.
The DB column is limited to chars 32 also, I think it's enough for model info.
WHDY @francisbouvier ?

arrufat: Yes, for telemetry it's fine, we can already guess the model based on the 32 first chars.

gaby:
@krichprollsch @karlseguin @arrufat This could potentially leak internal company model names and unreleased models. Given the difficulty in disabling telemetry within `lightpanda`.

krichprollsch:
@gaby thanks for the feedback, but can you elaborate about why you said it's difficult to disable the telemetry?

PR #2654, opened 2026-06-05
form: encode file inputs as multipart/form-data on submit

krichprollsch:
collectForm now emits File entries for <input type="file"> (one per selected file, or a synthetic empty File with application/octet-stream when none is selected, per WHATWG). multipartEncodeEntry writes the filename, Content-Type, and file bytes per RFC 7578; Value.asString / format fall back to the filename for urlencoded / text-plain encodings.

Follow-up for #2635

### DIFF
--- a/src/browser/webapi/net/FormData.zig
+++ b/src/browser/webapi/net/FormData.zig
@@ -112,14 +112,14 @@ pub fn init(form_: ?*Form, submitter: ?*Element, exec: *const Execution) !*FormD
 pub fn deinit(_: *FormData, _: *Page) void {}
 
 pub fn releaseRef(self: *FormData, page: *Page) void {
-    self._rc.release(self, page);
-
     for (self._entries.items) |entry| {
         switch (entry.value) {
             .file => |file| file.releaseRef(page),
             else => {},
         }
     }
+
+    self._rc.release(self, page);
 }
 
 pub fn acquireRef(self: *FormData) void {
--- a/src/browser/webapi/net/FormData.zig
+++ b/src/browser/webapi/net/FormData.zig
@@ -47,14 +47,17 @@ pub const Entry = struct {
         fn asString(self: *const Value) []const u8 {
             return switch (self.*) {
                 .string => |*s| s.str(),
-                .file => unreachable, // nothing currently creates this type of value
+                // Per WHATWG, file entries serialized in a non-multipart encoding
+                // (application/x-www-form-urlencoded, text/plain) collapse to the
+                // file's name only — body is dropped.
+                .file => |f| f.getName(),
             };
         }
 
         pub fn format(self: Value, writer: *std.Io.Writer) !void {
             return switch (self) {
                 .string => |s| s.format(writer),
-                .file => unreachable, // nothing currently creates this type of value
+                .file => |f| writer.writeAll(f.getName()),
             };
         }
     };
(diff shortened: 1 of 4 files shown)


src/browser/webapi/net/FormData.zig
```diff
@@ -351,6 +370,21 @@ fn collectForm(arena: Allocator, form_: ?*Form, submitter_: ?*Element, frame: *F
                         continue;
                     }
                 }
+                if (input_type == .file) {
+                    // WHATWG: a file input with zero selected files contributes a
+                    // single entry whose value is an empty File of MIME
+                    // application/octet-stream; otherwise, one entry per file.
+                    const files = if (input._files) |fl| fl._files else &.{};
+                    if (files.len == 0) {
+                        const empty = try File.init(null, "", .{ .type = "application/octet-stream" }, frame._page);
```
karlseguin: This leaks.

I suggest you add a 3rd type to Entry.value: `no_file: void`

src/browser/webapi/net/FormData.zig
```diff
@@ -351,6 +370,21 @@ fn collectForm(arena: Allocator, form_: ?*Form, submitter_: ?*Element, frame: *F
...
+                        try appendFile(&list, arena, name, empty);
+                    } else {
+                        for (files) |file| {
+                            try appendFile(&list, arena, name, file);
```
karlseguin, line 436: These files can get released.

The FormData will need to participate in reference counting (calling acquireRef and releaseRef on every file when it, itself is acquired/freed).

We can land this separately if you're eager to get these reachable - `unreachable` out of main. The change of a File being GC'd while the FormData still exists is pretty small.

krichprollsch: I tried to fix that with 8c1c3c93f837fb8c23ce3e7a54ada48dfdb1c293

src/browser/webapi/net/FormData.zig
```diff
@@ -33,6 +34,8 @@ const Allocator = std.mem.Allocator;
 
 const FormData = @This();
 
+_rc: lp.RC(u32),
```
karlseguin:
```suggestion
_rc: lp.RC(u8),
```

Only v8 can reference this (and it can only have 1 reference)

src/browser/webapi/net/FormData.zig
```diff
@@ -96,6 +109,30 @@ pub fn init(form_: ?*Form, submitter: ?*Element, exec: *const Execution) !*FormD
     return form_data;
 }
 
+pub fn deinit(_: *FormData, _: *Page) void {}
+
+pub fn releaseRef(self: *FormData, page: *Page) void {
+    self._rc.release(self, page);
```
karlseguin: Should be called _after_ the iteration. Now it doesn't matter, because `deinit` is noop. But if that ever changes, FormData will be cleared and then self._entries will be called.

karlseguin (approved):

navidemad (commented):
I just closed my duplicate (#2663) in favour of this one. It's the more complete take on #2662. I read through all three commits and checked the encoding and the lifetime handling end to end. A few notes in case they're useful.

## The `.no_file` representation is the right call

Modelling an unselected file input as a dedicated zero-width union tag, instead of synthesising an empty `File`, means the empty-input case allocates nothing. My version reached for `File.init(null, "", …)` there, and that leaks a `.large` Blob arena on every submit of an unfilled file input: the Blob comes off `session.getArena(.large, …)` and nothing releases it, since the synthetic File sits in no FileList the frame tracks. The tag avoids all of that and still produces the same `filename=""` + `application/octet-stream` bytes Chrome sends.

## Lifetime coupling is wired correctly

I went looking for whether `FormData.acquireRef`/`releaseRef` were actually reachable, since defining them isn't enough on its own. They are: declaring `acquireRef` makes `FormData` a finalizer type (`findFinalizerType` walks the chain for it), so `Local.zig` acquires when the instance first crosses into JS and the weak finalizer releases at GC or page teardown. A `new FormData(form)` genuinely co-owns the files it pulled out of the inputs' FileLists. Nice.

## One asymmetry worth a look: refcount on mutation

`acquireRef` and `releaseRef` each iterate `_entries`, but at two different moments: acquire at first JS exposure, release at GC. If a `.file` entry is removed in between via `delete()` or `set()` (`set` → `deleteByName` → append a string), `releaseRef` never sees it, so the ref `acquireRef` took on its Blob is never given back and the arena leaks.

```js
const fd = new FormData(formWithAFileInput);
fd.delete('thefile');   // entry dropped from _entries
// fd is GC'd → releaseRef iterates the *current* entries, the file is gone, its ref is never released
```

It's narrow today because `append`/`set` are string-only, so file entries can only come from `collectForm`. But the mirror case turns into a crash rather than a leak the moment a `Blob` overload for `append` lands (the spec has one): a file appended after exposure is never acquired, so at GC `releaseRef` calls `release` on a refcount of 0 and trips the `prev > 0` assert.

Acquiring and releasing at the add/remove sites (`appendFile` / `deleteByName`) instead of snapshot-iterating would make it robust to mutation and to a future Blob `append`. Happy to send a patch if that helps.

## Minor

- `multipartEncodeEntry` reads `file._proto._mime` directly. `file._proto.getType()` is the public accessor and keeps the field private. Trivial, take it or leave it.
- Unit coverage is solid. Two gaps I'd fill: an html test that submits a real `<form enctype="multipart/form-data">` with a file set through `DOM.setFileInputFiles` (nothing currently exercises `collectForm` and submission together), and a body with embedded NUL bytes to pin down that binary uploads survive. The libcurl path is length-delimited (`POSTFIELDSIZE` is set before `COPYPOSTFIELDS`), so it should, but that's exactly the kind of thing that earns a regression test.

krichprollsch:
I tried to address the leak on deletion with 845dac86bac754b170d5ceeb2bbc0f6ca578c248

karlseguin (approved):

PR #2657, opened 2026-06-05
feat(cdp): support mouse button and clickCount in Input.dispatchMouseEvent

rohitsux:
Continues the CDP mouse work from #2636 (hover), #2640 (mouseReleased) and #2641 (mouseWheel). `Input.dispatchMouseEvent` still ignored the `button` and `clickCount` parameters, and `mousePressed` only fired a `click` event — it never fired `mousedown`, and there was no `contextmenu`, `auxclick` or `dblclick` anywhere.

## Changes

- `src/cdp/domains/input.zig`: parse the `button` (left/middle/right/back/forward) and `clickCount` params and map the button to the DOM `MouseEvent.button` numbering.
- `src/browser/Frame.zig`: `triggerMousePress` fires `mousedown` carrying the button; `triggerMouseRelease` fires `mouseup` then the button-appropriate activation event — `click` (main), `auxclick` (auxiliary), or `contextmenu` (secondary/right) — and a `clickCount` of 2 also fires `dblclick`. The button-aware dispatch mirrors the existing helper in `WebDriver.zig`.

## Why

Right-click context menus, middle-click and double-click are core Playwright/Puppeteer interaction patterns that previously did nothing over CDP.

## Verified

- `make test` passes (782/782, no leaks); new test `cdp.input: dispatchMouseEvent right button fires contextmenu, double-click fires dblclick`.
- `make build` succeeds.

### DIFF
--- a/src/browser/Frame.zig
+++ b/src/browser/Frame.zig
@@ -3904,10 +3904,12 @@ fn findFrameByName(frame: *Frame, name: []const u8) ?*Frame {
 
 // DOM MouseEvent.button values.
 // https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button
-const mouse_button = struct {
-    const main: i32 = 0; // left
-    const auxiliary: i32 = 1; // middle
-    const secondary: i32 = 2; // right
+pub const mouse_button = struct {
+    pub const main: i32 = 0; // left
+    pub const auxiliary: i32 = 1; // middle
+    pub const secondary: i32 = 2; // right
+    pub const fourth: i32 = 3; // back
+    pub const fifth: i32 = 4; // forward
 };
 
 // Dispatch a single trusted mouse event of the given type on `target`, carrying
--- a/src/browser/Frame.zig
+++ b/src/browser/Frame.zig
@@ -3902,25 +3902,35 @@ fn findFrameByName(frame: *Frame, name: []const u8) ?*Frame {
     return null;
 }
 
-pub fn triggerMouseClick(self: *Frame, x: f64, y: f64) !void {
-    const target = (try self.window._document.elementFromPoint(x, y, self)) orelse return;
-    if (comptime IS_DEBUG) {
-        log.debug(.frame, "frame mouse click", .{
-            .url = self.url,
-            .node = target,
-            .x = x,
-            .y = y,
-            .type = self._type,
-        });
-    }
-    const mouse_event: *MouseEvent = try .initTrusted(comptime .wrap("click"), .{
+// Dispatch a single trusted mouse event of the given type on `target`, carrying
+// the pressed button and pointer position. `detail` is the click count (used for
+// click/dblclick); 0 for events where it does not apply.
+fn dispatchMouseEventOn(self: *Frame, target: *Element, comptime typ: []const u8, x: f64, y: f64, button: i32, detail: u32) !void {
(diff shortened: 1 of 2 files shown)


rohitsux:
@karlseguin  please review

src/browser/Frame.zig
```diff
@@ -3961,25 +3971,39 @@ pub fn triggerMouseMove(self: *Frame, x: f64, y: f64) !void {
...
+    const detail: u32 = if (click_count > 0) @intCast(click_count) else 1;
+
+    try self.dispatchMouseEventOn(target, "mouseup", x, y, button, detail);
+
+    // After mouseup, the activation event depends on the button:
+    //   main (0)      -> click
+    //   auxiliary (1) -> auxclick
+    //   secondary (2) -> contextmenu
+    switch (button) {
+        0 => {
```
krichprollsch: Creating constants would auto-document the switch.

src/cdp/domains/input.zig
```diff
@@ -92,15 +94,36 @@ fn dispatchMouseEvent(cmd: *CDP.Command) !void {
...
     const frame = bc.session.currentFrame() orelse return;
+
+    // Map the CDP button name to the DOM MouseEvent.button numbering
+    // (left/main = 0, middle/auxiliary = 1, right/secondary = 2, ...).
+    const button: i32 = switch (params.button) {
+        .none, .left => 0,
+        .middle => 1,
+        .right => 2,
+        .back => 3,
+        .forward => 4,
```
krichprollsch: same about using contants.

src/browser/Frame.zig
```diff
@@ -3902,25 +3902,35 @@ fn findFrameByName(frame: *Frame, name: []const u8) ?*Frame {
     return null;
 }
 
-pub fn triggerMouseClick(self: *Frame, x: f64, y: f64) !void {
+// Dispatch a single trusted mouse event of the given type on `target`, carrying
+// the pressed button and pointer position. `detail` is the click count (used for
+// click/dblclick); 0 for events where it does not apply.
+fn dispatchMouseEventOn(self: *Frame, target: *Element, comptime typ: []const u8, x: f64, y: f64, button: i32, detail: u32) !void {
```
krichprollsch, line 3918: Does this function dispatch mouse event for all kind of mouse event or only for pres/release?
If it's the first one, why not using it for other events triggered in Frame.zig?

src/browser/Frame.zig
rohitsux: Done — added a mouse_button constant block (main/auxiliary/secondary) and switched on those instead of the bare 0/1/2.

src/cdp/domains/input.zig
rohitsux: Done — the CDP-button to DOM-button mapping now uses named constants too.

src/browser/Frame.zig
rohitsux: Right now it's used for the press/release family (mousedown, mouseup, click, auxclick, contextmenu, dblclick), which all share the same options (bubbles + cancelable + composed). The other events differ slightly: mousemove/mouseover match, but mouseenter/mouseleave are non-bubbling and wheel is a WheelEvent rather than a MouseEvent, so I kept the helper scoped to the click family to avoid changing their behavior. Happy to generalize it (take the event options as a param) and route triggerMouseMove through it too if you'd prefer the consolidation lmk.

krichprollsch (approved):

PR #2658, opened 2026-06-05
`URL`: Rework URL web API with rust-url

nikneym:
This PR reworks our Web API URL implementation by taking advantage of rust-url package.

### DIFF
--- a/src/browser/Frame.zig
+++ b/src/browser/Frame.zig
@@ -349,7 +349,11 @@ pub fn init(self: *Frame, frame_id: u32, page: *Page, parent: ?*Frame) !void {
     });
     errdefer browser.env.destroyContext(self.js);
 
-    self.window._location = try Location.init("about:blank", self);
+    const location = try Location.init("about:blank", self);
+    // We're holding a reference in Zig-side.
+    location.acquireRef();
+    errdefer location.releaseRef(page);
+    self.window._location = location;
 
     document._frame = self;
 
@@ -429,6 +433,9 @@ pub fn deinit(self: *Frame) void {
         if (document._fonts) |f| {
             f.releaseRef(page);
         }
+
+        // Release our reference to location.
+        self.window._location.releaseRef(page);
     }
 
     const browser = page.session.browser;
@@ -559,7 +566,14 @@ pub fn navigate(self: *Frame, request_url: [:0]const u8, opts: NavigateOpts) !vo
         // have to do this to make sure window.location is at a unique _address_.
         // If we don't do this, multiple window._location will have the same
         // address and thus be mapped to the same v8::Object in the identity map.
-        self.window._location = try Location.init(self.url, self);
+        //
+        // We don't hold a reference to old location anymore.
+        self.window._location.releaseRef(self._page);
+        // Create new location.
+        const location = try Location.init(self.url, self);
+        location.acquireRef();
+        errdefer location.releaseRef(self._page);
+        self.window._location = location;
 
         if (is_blob) {
             // strip out blob:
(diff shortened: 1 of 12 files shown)


src/browser/webapi/History.zig
```diff
@@ -62,7 +62,8 @@ pub fn pushState(_: *History, state: js.Value, _: ?[]const u8, _url: ?[]const u8
     _ = try frame._session.navigation.pushEntry(url, .{ .source = .history, .value = json }, frame, true);
 
     frame.url = url;
-    frame.window._location._url = try URL.init(url, null, &frame.js.execution);
+    // setHref == reinitializing.
+    try frame.window._location._url.setHref(url, &frame.js.execution);
```
karlseguin, line 64: Is this correct? In the previous version, we created a new URL. Now we mutate the existing one. Fine. Exception, in `replaceState` we maintain the "create a new URL" pattern. This seems like it's either fixing an error in the old code, or it's a bit inconsistent with itself.

src/browser/Frame.zig
```diff
@@ -557,11 +562,18 @@ pub fn navigate(self: *Frame, request_url: [:0]const u8, opts: NavigateOpts) !vo
...
+        // even though about:blank navigations may share the same _data_, we
         // have to do this to make sure window.location is at a unique _address_.
         // If we don't do this, multiple window._location will have the same
         // address and thus be mapped to the same v8::Object in the identity map.
-        self.window._location = try Location.init(self.url, self);
+        //
+        // We don't hold a reference to old location anymore.
+        self.window._location.releaseRef(self._page);
+        // Create new location.
+        const location = try Location.init(self.url, self);
```
karlseguin, line 568: If this fails, e.g. an invalid url, it returns and `window._location` is no longer valid due to the line above. Swapping the two lines solves the issue:

```
const location = try Location.init(self.url, self);
location.acquireRef();
errdefer location.releaseRef(self._page);

self.window._location.releaseRef(self._page);
self.window._location = location;
```

This same pattern happens 3 times in Frame.zig and once in History.zig. Maybe a `window.swapLocation` since it's 4 places.

src/browser/webapi/URL.zig
```diff
@@ -17,234 +17,324 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 const std = @import("std");
+const lp = @import("lightpanda");
 const js = @import("../js/js.zig");
 
-const U = @import("../URL.zig");
+const U_ = @import("../../sys/url.zig");
```
karlseguin:
```suggestion
const U = @import("../../sys/url.zig");
```

src/browser/webapi/URL.zig
```diff
@@ -17,234 +17,324 @@
...
-pub fn init(url: [:0]const u8, base_: ?[:0]const u8, exec: *const Execution) !*URL {
-    const arena = exec.arena;
-    const context_url = exec.url.*;
+pub fn init(url: []const u8, maybe_base: ?[]const u8, exec: *const Execution) !*URL {
+    // NOTE: about:blank address is valid in rust-url.
+
+    var err: i32 = 0;
+    if (maybe_base) |base| {
+        const base_url = U_.url_parse(base.ptr, base.len, &err) orelse return error.TypeError;
+        errdefer U_.url_free(base_url);
```
karlseguin:
```suggestion
        defer U_.url_free(base_url);
```

and remove the ` U_.url_free(base_url);` later...it's a double-free on error as-is.

karlseguin:
Small feedback inline. Is the plan to replace URL.resolve in a follow up PR? Is there a risk of having the two live at the same time, e.g. new URL(...)  and top.location = '...' end up behaving differently?

nikneym:
> Small feedback inline. Is the plan to replace URL.resolve in a follow up PR?

Indeed, that's how I planned. It breaks couple of small cases (mostly related to encoding).

src/browser/webapi/History.zig
nikneym: Yes, we also never expose `URL` directly to JS so it's safe to do. Turns out `replaceState` was left non-spec-compliant so I've applied the same.

src/browser/Frame.zig
nikneym: I'm unsure of adding a function for this, though I've updated freeing order for `window._location`.

src/browser/Frame.zig
```diff
@@ -353,6 +349,12 @@ pub fn init(self: *Frame, frame_id: u32, page: *Page, parent: ?*Frame) !void {
     });
     errdefer browser.env.destroyContext(self.js);
 
+    const location = try Location.init("about:blank", self);
+    // We're holding a reference in Zig-side.
+    location.acquireRef();
+    errdefer location.releaseRef(page);
```
karlseguin: Are these right? If anything after this fails, we releaseRef, but we'll still releaseRef on Frame.deinit. I think this `errdefer` can simply be removed (and the others like it).

PR #2664, opened 2026-06-07
feat(cdp): honor viewport in Emulation.setDeviceMetricsOverride

rohitsux:
`Emulation.setDeviceMetricsOverride` was a noop, so `window.innerWidth`/`innerHeight` always reported the hardcoded 1920x1080 regardless of the viewport a driver set. This is a CDP-automation gap from #2043 (responsive/mobile layout testing).

## Changes

- `src/browser/webapi/Window.zig`: store the viewport (width/height, default 1920x1080) on the Window; expose it via `innerWidth`/`innerHeight`, which become accessors backed by `getInnerWidth`/`getInnerHeight`. `getInnerHeight` (already used by `scrollIntoViewIfNeeded`) now reads the stored value. The properties remain writable from JS, mirroring `scrollX`/`scrollY`.
- `src/cdp/domains/emulation.zig`: implement `setDeviceMetricsOverride` to parse `width`/`height` and set the current frame window's viewport (a 0 dimension means "don't override", per the CDP default). Mirrors the existing `setUserAgentOverride` shape.

## Why

`setViewport` is a core Playwright/Puppeteer call for responsive and mobile testing; it previously had no effect.

## Verified

- `make test` passes (782/782, no leaks); new test `cdp.Emulation: setDeviceMetricsOverride updates window.innerWidth/innerHeight`.
- `make build` succeeds.

### DIFF
--- a/src/browser/Page.zig
+++ b/src/browser/Page.zig
@@ -25,6 +25,7 @@ const v8 = js.v8;
 const Frame = @import("Frame.zig");
 const Session = @import("Session.zig");
 const Factory = @import("Factory.zig");
+const Viewport = @import("Viewport.zig");
 
 const Allocator = std.mem.Allocator;
 const IS_DEBUG = builtin.mode == .Debug;
@@ -117,6 +118,18 @@ popups: std.ArrayList(*Frame) = .empty,
 // branch on this to stamp `is_pending_root` on the frame_navigate
 // notification (so CDP doesn't reset its node registry yet) and
 _state: enum { active, pending } = .active,
+
+// Runtime viewport override set via Emulation.setDeviceMetricsOverride and
+// cleared via Emulation.clearDeviceMetricsOverride. Null means use the
+// compile-time Viewport.default. Read every viewport consumer through
+// getViewport so they all observe the same (possibly overridden) value.
+viewport_override: ?Viewport = null,
+
+// The viewport every consumer should read: the runtime override if set,
+// otherwise the compile-time default.
+pub fn getViewport(self: *const Page) Viewport {
+    return self.viewport_override orelse Viewport.default;
+}
 
 // Initialize a Page and its root Frame.
 pub fn init(self: *Page, session: *Session, frame_id: u32) !void {
--- a/src/browser/StyleManager.zig
+++ b/src/browser/StyleManager.zig
@@ -20,7 +20,6 @@ const std = @import("std");
 const lp = @import("lightpanda");
 
 const Frame = @import("Frame.zig");
-const Viewport = @import("Viewport.zig");
 
 const CssParser = @import("css/Parser.zig");
 const MediaQuery = @import("css/MediaQuery.zig");
@@ -148,7 +147,7 @@ fn applyMediaAtRule(self: *StyleManager, text: []const u8, depth: u8) !void {
     const query = std.mem.trim(u8, rest[0..open], &std.ascii.whitespace);
     const inner = rest[open + 1 .. close];
 
-    if (!MediaQuery.matches(query, Viewport.default)) return;
+    if (!MediaQuery.matches(query, self.frame._page.getViewport())) return;
 
     var it = CssParser.parseStylesheet(inner);
     while (it.next()) |nested_rule| {
--- a/src/browser/webapi/Element.zig
+++ b/src/browser/webapi/Element.zig
@@ -1543,7 +1543,7 @@ pub fn scrollIntoViewIfNeeded(self: *Element, center_if_needed: ?bool, frame: *F
     _ = center_if_needed;
     const y = calculateDocumentPosition(self.asNode());
     const scroll_y: f64 = @floatFromInt(frame.window.getScrollY());
-    const viewport_height: f64 = @floatFromInt(frame.window.getInnerHeight());
+    const viewport_height: f64 = @floatFromInt(frame.window.getInnerHeight(frame));
     if (y >= scroll_y and y <= scroll_y + viewport_height) {
         return;
     }
--- a/src/browser/webapi/IntersectionObserver.zig
+++ b/src/browser/webapi/IntersectionObserver.zig
@@ -22,7 +22,6 @@ const js = @import("../js/js.zig");
 
 const Page = @import("../Page.zig");
 const Frame = @import("../Frame.zig");
-const Viewport = @import("../Viewport.zig");
 
 const Node = @import("Node.zig");
 const Element = @import("Element.zig");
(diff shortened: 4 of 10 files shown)


karlseguin:
I have my doubts about this. Unfortunately, 1080 & 1920 are hard-coded in a few different places, I'm afraid changing them in one, but not the others, will break more than it fixes.

I'd rather fix the hardcoded-ness in one commit first, then add support for this. But..is this really something CDP clients use/care about?

krichprollsch:
I suppose it could have an impact for 2 things:
* testing in mobile env, maybe some js can have different behavior depending this value
* impact on fingerprinting?

But IMO the noop is good enough until we have a proof it's an issue.
@navidemad maybe you have a point on that?

rohitsux:
Makes sense, agreed. You're right that a partial override would be inconsistent: 1920/1080 is also hard-coded in Page.getLayoutMetrics, the MediaQuery viewport (so matchMedia wouldn't reflect it), VisualViewport, and Screen, so changing only innerWidth/innerHeight would create more confusion than it solves. And without a concrete CDP client need, the noop is reasonable. I'm happy to leave this until there's a real use case, or to take a first pass at centralizing the hard-coded viewport into one default if that groundwork is useful on its own. Whatever you prefer.

navidemad:
@karlseguin you're right to want the centralize-first path, and the client side backs you up.

On "does anyone actually use this": not me, not today. capybara-lightpanda doesn't implement Capybara's window-sizing API at all, so `current_window.resize_to(w, h)` just raises `NotImplementedError` and nothing ever reaches `setDeviceMetricsOverride`/`setViewport`. The `window_size` option I expose is vestigial: stored, never sent to the browser. I route breakpoint specs to Cuprite on purpose in the dual-driver setup.

Where I land matches your instinct: an innerWidth-only override would be worse than the current noop for me. #2478 already wired `@media`/matchMedia to the same hardcoded 1920×1080, so honoring it on `innerWidth` alone gives a split-brain viewport — `window.innerWidth` reports 375 while matchMedia, `getLayoutMetrics`, `VisualViewport` and `Screen` all still say 1920. I'd rather have everything wrong-but-consistent than half of it "fixed."

The thing I'd actually use isn't innerWidth, it's one settable viewport. The day `setDeviceMetricsOverride` drives a single centralized value that matchMedia/checkVisibility/innerWidth/getLayoutMetrics all read, I can wire that dead `window_size` through and pull the breakpoint specs back off Cuprite. Useful but not urgent: the noop holds until the groundwork's there.

So: centralization first gets my vote, the partial patch doesn't.

navidemad:
To make "centralize first" concrete rather than hand-wavy — the 1920×1080 is duplicated across six independent sites right now:

- `src/browser/css/MediaQuery.zig:50` (`Viewport.default`) — already flagged in-code as the rewire point: *"when viewport emulation lands, this is the single helper to rewire so the cascade and matchMedia move together."* Natural home for the source of truth.
- `src/browser/webapi/Window.zig:59` + `:1013` — `innerHeight` / `innerWidth` (what this PR touches).
- `src/browser/webapi/Screen.zig:57-59` — `screen.width` / `height` / `availWidth`.
- `src/browser/webapi/VisualViewport.zig:54-55` — `visualViewport.width` / `height`.
- `src/cdp/domains/page.zig:909-910` — `Page.getLayoutMetrics`, feeding all four layout/visual/css viewport blocks.
- `src/browser/webapi/IntersectionObserver.zig:214-215` — root viewport bounds for intersection rects.

Rough shape: point those six at one stored viewport seeded from `MediaQuery.Viewport.default`, then `Emulation.setDeviceMetricsOverride` writes it (and `clearDeviceMetricsOverride` resets to default), keeping the 0 = "don't override" rule. Scope it per-page so matchMedia (per-document) and getLayoutMetrics (per-target) read the same value. There's also a `width == 5.0 → 1920` fallback at `Element.zig:1216` that's probably a separate quirk, not part of this.

Once that's in place, honoring innerWidth/innerHeight falls out for free — and it's the point where I could finally wire `window_size` through on the gem side.

rohitsux:
Thanks @navidemad, that's a really clear breakdown. I'll take the centralize-first path: introduce one stored viewport seeded from `MediaQuery.Viewport.default` and point the six sites (MediaQuery, Window inner*, Screen, VisualViewport, Page.getLayoutMetrics, IntersectionObserver) at it, no behavior change. Then `setDeviceMetricsOverride`/`clearDeviceMetricsOverride` writing that single value (with 0 = don't override) can land as a clean follow-up. I'll open the centralization PR separately and leave this one as the discussion thread.

krichprollsch:
@rohitsux can you please rebase the PR now #2694 is merged?

rohitsux:
Rebased onto main and reworked to use the centralized viewport from #2694 (thanks @navidemad for the six-site breakdown earlier).

Instead of the earlier innerWidth-only approach, the override now lives on the \`Page\` and every consumer reads it via \`Page.getViewport()\`: \`window.innerWidth\`/\`innerHeight\`, \`screen.width\`/\`height\`/\`availWidth\`, \`visualViewport.width\`/\`height\`, \`Page.getLayoutMetrics\`, the \`IntersectionObserver\` root rect, and \`matchMedia\` / the \`StyleManager\` media cascade. So \`setDeviceMetricsOverride\` moves all of them together — no split-brain.

- \`setDeviceMetricsOverride\` writes the override (0 width/height keeps that dimension, per the CDP default).
- \`clearDeviceMetricsOverride\` resets to the default.
- \`deviceScaleFactor\`, \`mobile\`, \`scale\`, \`screenWidth\`/\`screenHeight\` are accepted but logged as not_implemented for now.

\`make test\` passes (the only failing test, \`WebApi: Frames\`, fails identically on a clean \`main\` for me — it looks environment-dependent and unrelated to this change; happy to look if it's not expected). \`zig fmt --check\` clean, \`make build\` OK.

src/browser/Page.zig
```diff
@@ -118,6 +119,18 @@ popups: std.ArrayList(*Frame) = .empty,
 // notification (so CDP doesn't reset its node registry yet) and
 _state: enum { active, pending } = .active,
 
+// Runtime viewport override set via Emulation.setDeviceMetricsOverride and
+// cleared via Emulation.clearDeviceMetricsOverride. Null means use the
+// compile-time Viewport.default. Read every viewport consumer through
+// getViewport so they all observe the same (possibly overridden) value.
+viewport_override: ?Viewport = null,
```
krichprollsch, line 126: Having the viewport_override at `Page` level works only for the current page. But we will loose the change on next page... I think it would make more sense to set the `viewport` at `Browser` level.

But we can change that in a followup PR.

krichprollsch (approved):

PR #2665, opened 2026-06-08
fix: map password inputs to textbox accessibility role

mvanhorn:
## Summary

A `<input type="password">` now reports the accessibility role `textbox` in the CDP accessibility tree, matching Chrome. It previously reported `none`, so accessibility-driven automation could not locate password fields.

## Why this matters

In `src/cdp/AXNode.zig` the `.input` role-mapping switch mapped `.password, .hidden => .none`. Chrome reports `textbox "Password" [required]` while Lightpanda reported `none "Password" [required]` (#2610, reproduced by @karlseguin on 2026-06-08, who named the exact line). This moves `.password` into the `.text, .email, .tel, .url => .textbox` branch and leaves `.hidden => .none` intact, aligning the role mapping with the existing name handling that already groups `.password` with the text-like inputs. A regression test covers the password-input role.

This is the secondary half of #2610; the primary `XMLHttpRequest` cause was already fixed by #2637 and #2638.

## Testing

- Password input maps to `textbox`; hidden input stays `none`.
- The existing hidden-input label-promotion test still passes.

Refs #2610

### DIFF
--- a/src/cdp/AXNode.zig
+++ b/src/cdp/AXNode.zig
@@ -803,7 +803,7 @@ pub const AXRole = enum(u8) {
                 .input => {
                     const input = el.as(DOMNode.Element.Html.Input);
                     return switch (input._input_type) {
-                        .tel, .url, .email, .text => .textbox,
+                        .tel, .url, .email, .text, .password => .textbox,
                         .image, .reset, .button, .submit => .button,
                         .radio => .radio,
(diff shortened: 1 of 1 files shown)


krichprollsch (approved):

krichprollsch:
Waiting until the GH outage is fixed.
https://www.githubstatus.com/

PR #2666, opened 2026-06-08
WPT /WebCryptoAPI/

karlseguin:
I threw Claude at WPT's /WebCryptoAPI/ which represents a significant number of failing cases.

There were a few very low-hanging fruit, like enabling CryptoKey on Worker and supporting `{name: string}` in addition to just `string` for some functions.

Some of these "fixes" just handle the error case / validation, which tends to represent a greater number of WPT cases, e.g. some of the importKey algorithms aren't fully supported, but they still validate the input and return the correct errors.

To that end, while there should be considerably more passing cases (~42K), some of these changes merely unlocked more failing cases, so our total case count should go up.

Claude's summary:

  1. digest {name} object fix (16→80)
  2. Symmetric importKey/exportKey/generateKey (AES, HMAC; raw + jwk) + CryptoKey accessors + DataError
  3. EC/OKP importKey usage validation (failure-path)
  4. PBKDF2 + HKDF deriveBits, then deriveKey
  5. ECDH generateKey + import (spki/pkcs8) + deriveBits/deriveKey
  6. AES encrypt/decrypt (CBC/CTR/GCM)

### DIFF
--- a/src/browser/js/Local.zig
+++ b/src/browser/js/Local.zig
@@ -1411,12 +1411,6 @@ pub fn rejectPromise(self: *const Local, err: js.PromiseResolver.RejectError) js
     return resolver.promise();
 }
 
-pub fn rejectErrorPromise(self: *const Local, value: js.PromiseResolver.RejectError) !js.Promise {
-    var resolver = js.PromiseResolver.init(self);
-    resolver.rejectError("Local.rejectPromise", value);
-    return resolver.promise();
-}
-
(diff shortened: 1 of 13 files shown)


karlseguin:
I should point out that this is all just wrapping libcrypto (which we import via libcurl).

krichprollsch (approved):

karlseguin:
I'd like to merge this AFTER the WPTRunner changes land/stabilize. Then this. Then we can rebase `wpt`. That way we have some indication of what each change is doing.

PR #2671, opened 2026-06-08
webapi: DataTransfer / DataTransferItem / DataTransferItemList + DragEvent (drag-and-drop file upload)

navidemad:
## What this enables

Drag-and-drop file upload. Page JS and test drivers build a `DataTransfer`, add `File` objects to it, and dispatch a `drop`/`dragover` event whose `.dataTransfer` exposes those files as a `FileList`:

```js
const dt = new DataTransfer();
dt.items.add(new File(["…"], "report.pdf", { type: "application/pdf" }));
el.dispatchEvent(new DragEvent("drop", { dataTransfer: dt }));
// handler reads e.dataTransfer.files[0] -> the File
```

Until now `DataTransfer` / `DataTransferItem` were `undefined` (TODO stubs), so that whole pattern threw. This was the last untracked gap in #2043, and the capybara-lightpanda driver needs it for drag-and-drop file-upload system tests.

## What this adds

- **`DataTransfer`** — constructible; `dropEffect`, `effectAllowed`, `types`, `files` (`FileList`), `items` (`DataTransferItemList`), `getData`/`setData`/`clearData`, `setDragImage` (no-op).
- **`DataTransferItem`** — `kind`, `type`, `getAsFile()`, `getAsString(cb)`.
- **`DataTransferItemList`** — `length`, indexed access, `add(data, type)` / `add(file)`, `remove(i)`, `clear()`.
- **`DragEvent`** (extends `MouseEvent`) carrying `dataTransfer`; also resolves the `InputEvent.dataTransfer` TODO.

## How it fits together

`DataTransfer` keeps one ordered drag-data-store of items (string- or file-kind). `.items` is a live view that mutates the store, `.files` is a `FileList` rebuilt from the file-kind items, and `.types` is derived. So `items.add(file)` shows up in `.files` right away and the two never drift.

Memory follows the existing `<input type=file>` path: the backing `FileList` is created on the factory slab and frame-tracked, `add` takes a ref on each `File`, and `remove`/`clear`/teardown drop it, so the leak-checking test runner stays green. The `DataTransfer`'s own arena is refcounted like `Blob`, so the GC finalizer reclaims it. `DataTransferItem`, `DataTransferItemList`, and the item-list iterator all live in that arena but are handed to JS on their own, so each forwards `acquireRef`/`releaseRef` to the owning `DataTransfer` (the same way `XMLHttpRequestUpload` forwards to its `XMLHttpRequest`) — holding any of them from JS keeps the store alive even if the `DataTransfer` wrapper is collected first. A `DragEvent`/`InputEvent` that carries a `DataTransfer` holds a ref the same way, released in `deinit` like `MessageEvent` does with a `Blob`.

`getData`/`setData` apply the spec's format shorthands (`text` becomes `text/plain`, `url` becomes `text/uri-list`). `effectAllowed`/`dropEffect` reject values outside their allow-lists, `setData` replaces or appends, and `clearData()` drops string data but keeps files.

> [!NOTE]
> Deferred to a follow-up: CDP `Input.dispatchDragEvent` injection. The JS-constructible `DataTransfer` plus the drop-event path is what unblocks the common driver pattern; CDP-side synthesis is additive and can land on its own. The v1 drag-data-store is always read/write. It does not model the spec's event-phase read-only/protected mode gating, because drivers build a `DataTransfer` and dispatch without ever entering the restricted phases.

## Where to focus review

- `DataTransfer.zig` / `DataTransferItem.zig` / `DataTransferItemList.zig` — the `File` acquire/release balance across `addFileItem` / `removeItem` / `clearItems` against the frame-tracked `FileList`, and the `acquireRef`/`releaseRef` forwarding from the item, list, and iterator to the owning `DataTransfer` so its arena outlives any JS-held wrapper.
- `DragEvent.zig` / `InputEvent.zig` — these reach the shared refcount through `asEvent()._rc` (their proto chain is deeper than `MessageEvent`'s), acquire the `DataTransfer` on store, and release it in `deinit`.
- `MouseEvent.zig` — the new `drag_event` arm in the type union and `is()`.

## Test plan

- [x] TDD fixtures first: `new DataTransfer()`, `items.add(file)` reflected in `.files`, `getData`/`setData` round-trip with normalization, `remove`/`clear`, item iteration (`data_transfer.html`); a dispatched `drop` `DragEvent` whose `dataTransfer.files` is readable, plus a refcount-balance stress block (`event/drag.html`).
- [x] `make test` — 791/791, no leaks.
- [x] `zig fmt --check ./*.zig ./**/*.zig` clean.
- [x] `zig build` succeeds.

Refs #2043

### DIFF
--- a/src/browser/webapi/DataTransfer.zig
+++ b/src/browser/webapi/DataTransfer.zig
@@ -98,8 +98,7 @@ fn normalizeFormat(arena: Allocator, format: []const u8) ![]const u8 {
     if (std.ascii.eqlIgnoreCase(format, "url")) {
         return "text/uri-list";
     }
-    const buf = try arena.dupe(u8, format);
-    return std.ascii.lowerString(buf, buf);
+    return std.ascii.allocLowerString(arena, format);
 }
 
 pub fn getData(self: *const DataTransfer, format: []const u8, frame: *Frame) ![]const u8 {
@@ -114,15 +113,15 @@ pub fn getData(self: *const DataTransfer, format: []const u8, frame: *Frame) ![]
 
 pub fn setData(self: *DataTransfer, format: []const u8, data: []const u8) !void {
     const norm = try normalizeFormat(self._arena, format);
-    const dup = try self._arena.dupe(u8, data);
+    const owned_data = try self._arena.dupe(u8, data);
     for (self._items.items) |it| {
         if (it._kind == .string and std.mem.eql(u8, it._type, norm)) {
-            it._payload = .{ .string = dup };
+            it._payload = .{ .string = owned_data };
             return;
         }
     }
     const it = try self._arena.create(DataTransferItem);
-    it.* = .{ ._kind = .string, ._type = norm, ._payload = .{ .string = dup } };
+    it.* = .{ ._data_transfer = self, ._kind = .string, ._type = norm, ._payload = .{ .string = owned_data } };
     try self._items.append(self._arena, it);
 }
 
@@ -159,11 +158,10 @@ pub fn addItem(self: *DataTransfer, data: js.Value, type_: ?[]const u8, frame: *
         return try self.addFileItem(file, frame);
     } else |_| {}
 
-    const s = try data.toZig([]const u8);
+    const owned_data = try data.toStringSliceWithAlloc(self._arena);
     const norm = try normalizeFormat(self._arena, type_ orelse "");
-    const dup = try self._arena.dupe(u8, s);
     const it = try self._arena.create(DataTransferItem);
-    it.* = .{ ._kind = .string, ._type = norm, ._payload = .{ .string = dup } };
+    it.* = .{ ._data_transfer = self, ._kind = .string, ._type = norm, ._payload = .{ .string = owned_data } };
     try self._items.append(self._arena, it);
     return it;
 }
@@ -171,7 +169,7 @@ pub fn addItem(self: *DataTransfer, data: js.Value, type_: ?[]const u8, frame: *
 fn addFileItem(self: *DataTransfer, file: *File, frame: *Frame) !*DataTransferItem {
     file._proto.acquireRef();
     const it = try self._arena.create(DataTransferItem);
-    it.* = .{ ._kind = .file, ._type = file._proto.getType(), ._payload = .{ .file = file } };
+    it.* = .{ ._data_transfer = self, ._kind = .file, ._type = file._proto.getType(), ._payload = .{ .file = file } };
     try self._items.append(self._arena, it);
     try self.rebuildFiles(frame);
     return it;
--- a/src/browser/webapi/DataTransfer.zig
+++ b/src/browser/webapi/DataTransfer.zig
@@ -1,0 +1,291 @@
+// Copyright (C) 2023-2026  Lightpanda (Selecy SAS)
+//
+// Francis Bouvier <francis@lightpanda.io>
+// Pierre Tachoire <pierre@lightpanda.io>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+const std = @import("std");
+const lp = @import("lightpanda");
+
+const js = @import("../js/js.zig");
+const Frame = @import("../Frame.zig");
+const Page = @import("../Page.zig");
+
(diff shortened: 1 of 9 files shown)


src/browser/webapi/DataTransfer.zig
```diff
@@ -0,0 +1,291 @@
...
+// https://html.spec.whatwg.org/multipage/dnd.html#dom-datatransfer-getdata
+// "text" and "url" are shorthands the spec maps onto MIME types.
+fn normalizeFormat(arena: Allocator, format: []const u8) ![]const u8 {
+    if (std.ascii.eqlIgnoreCase(format, "text")) {
+        return "text/plain";
+    }
+    if (std.ascii.eqlIgnoreCase(format, "url")) {
+        return "text/uri-list";
+    }
+    const buf = try arena.dupe(u8, format);
```
karlseguin:
```zig
return std.ascii.allocLowerString(arena, format);
```

src/browser/webapi/DataTransfer.zig
```diff
@@ -0,0 +1,291 @@
...
+
+// --- DataTransferItemList delegation ---
+
+// add(File) -> file item ; add(DOMString, DOMString) -> string item.
+pub fn addItem(self: *DataTransfer, data: js.Value, type_: ?[]const u8, frame: *Frame) !?*DataTransferItem {
+    if (data.toZig(*File)) |file| {
+        return try self.addFileItem(file, frame);
+    } else |_| {}
+
+    const s = try data.toZig([]const u8);
```
karlseguin: I think a union works for data here:

```zig
const AddItemData = union(enum) {
   file: *File,
   string: []const u8,  // order matters, anything can be coerced to string, so it should go last
}
pub fn addItem(self: *DataTransfer, data: AddItemData, ....) !?*DataTransferItem {
  switch (data) {
      ...
   }
}
```

If there is a reason to keep it as-is, then you should be able to avoid the data  -> s -> dup double copy:

```zig
const s = try data.toStringSliceWithAlloc(self._arena)
```

src/browser/webapi/DataTransfer.zig
```diff
@@ -0,0 +1,291 @@
...
+        if (it._kind == .string and std.mem.eql(u8, it._type, norm)) {
+            return it._payload.string;
+        }
+    }
+    return "";
+}
+
+pub fn setData(self: *DataTransfer, format: []const u8, data: []const u8) !void {
+    const norm = try normalizeFormat(self._arena, format);
+    const dup = try self._arena.dupe(u8, data);
```
karlseguin:
```suggestion
    const owned_data = try self._arena.dupe(u8, data);
```

Sorry, minor, but I was sure this was duping `norm` at first, and I couldn't quite figure it out. I'm sure I've named things `dup` also mind you.

src/browser/webapi/DataTransfer.zig
```diff
@@ -0,0 +1,291 @@
...
+// The canonical drag-data-store: one ordered list of items (string- or
+// file-kind). `.items` is a live DataTransferItemList view; `.files` is a
+// FileList rebuilt from the file-kind items so the two stay in sync. Per v1
+// scope the store is always read/write (no event-phase mode gating).
+const DataTransfer = @This();
+
+pub fn registerTypes() []const type {
+    return &.{
+        DataTransfer,
+        DataTransferItem,
```
karlseguin, line 44: I think `DataTransferItem`, `DataTransferItemList` and `DataTransferItemList.Iterator` all need an acquireRef and releaseRef. They all forward it to their `self._data_transfer` (which means that has to be added to `DataTransferItem`). For The Iterator, it can forward it to its `self.list` (which will forward it to `_data_transfer`).

src/browser/webapi/DataTransfer.zig
navidemad: Done. Switched to `std.ascii.allocLowerString(arena, format)`, which folds the dupe and the lowercasing into one call.

src/browser/webapi/DataTransfer.zig
navidemad: Went with the second option. `addItem` now reads the string straight into the persistent arena via `data.toStringSliceWithAlloc(self._arena)`, so the `data -> s -> dup` hop is gone.

I did look at the union. It reads well, but it wouldn't actually drop the copy: the bridge materializes a `[]const u8` field through `call_arena` (`String.toSlice`), so I'd still have to dupe it into `_arena`. It would also push the union type up into `DataTransferItemList.add`, which forwards `data` down to here. So I kept `js.Value` and just removed the extra copy.

src/browser/webapi/DataTransfer.zig
navidemad: Renamed to `owned_data` in `setData`, and the same in the new `addItem` path.

src/browser/webapi/DataTransfer.zig
navidemad: Added. `DataTransferItem` now carries a `_data_transfer` and forwards `acquireRef`/`releaseRef` to it. `DataTransferItemList` forwards to its own `_data_transfer`, and the iterator's inner struct forwards to `self.list`, which the generic `Entry` iterator picks up through its `_inner` acquire/release. Same shape as `NodeList.Iterator` and `XMLHttpRequestUpload`.

PR #2674, opened 2026-06-09
test,perf: Fix leaky thread in test runner

karlseguin:
The TestHTTPServer leaks a thread for every connection that shuts down. This commit correctly releases it. It also closes a leaking CDP socket.

Full test went from 20s -> 10s for me


krichprollsch:
50s -> 35s here

krichprollsch (approved):

PR #2675, opened 2026-06-09
perf,http: Support for script preloading

karlseguin:
Implement <link rel=preload as=script href=...>. The implementation is similar to preloadImport / waitForImport. Consider a website that does something like:

```
// pseudo html
<head>
<link preload script1.js>
<link preload script2.js>
<link preload script3.js>
<link preload script4.js>

<script script1.js></script>
<script script2.js></script>
<script script3.js></script>
<script script4.js></script>
```

Then, without preloading, we hit <script script1.js></script> and block while we load it + execute it. Repeat for 2, 3, 4 ...

With preloading, by the time we block on <script script1.js></script> all the scripts are already being downloaded in the background.

I opted to remove the script on first use. If a script happens to be used twice (we have seen this happen for imports, but I guess it's more rare on blocking scripts), then it'll get re-downloaded the 2nd time, just like before (and just like before, the http cache is a better mechanism to rely on here).

airbnb preloads 41 scripts.

### DIFF
--- a/src/browser/ScriptManager.zig
+++ b/src/browser/ScriptManager.zig
@@ -61,19 +61,20 @@ pub fn init(allocator: Allocator, http_client: *HttpClient, frame: *Frame) Scrip
 }
 
 pub fn deinit(self: *ScriptManager) void {
-    self.freeDonePreloads();
+    self.freePreloads();
     self.base.deinit();
     self.preloaded_scripts.deinit(self.base.allocator);
 }
 
 pub fn reset(self: *ScriptManager) void {
-    self.freeDonePreloads();
+    self.freePreloads();
     self.preloaded_scripts.clearRetainingCapacity();
     self.base.reset();
     self.frame_notified_of_completion = false;
 }
 
-fn freeDonePreloads(self: *ScriptManager) void {
+// Frees every preloaded Script
+fn freePreloads(self: *ScriptManager) void {
     var it = self.preloaded_scripts.valueIterator();
     while (it.next()) |preload_script| {
         preload_script.deinit();
@@ -107,7 +108,7 @@ pub fn preloadScript(self: *ScriptManager, url: []const u8) !void {
     }
 
     const frame = self.frame;
-    const arena = try frame.getArena(.medium, "SM.preloadScript");
+    const arena = try frame.getArena(.large, "SM.preloadScript");
     errdefer frame.releaseArena(arena);
 
     const owned_url = try arena.dupeZ(u8, url);
@@ -123,24 +124,14 @@ pub fn preloadScript(self: *ScriptManager, url: []const u8) !void {
         .extra = .preload,
     };
 
-    try self.preloaded_scripts.putNoClobber(self.base.allocator, owned_url, .{});
+    try self.preloaded_scripts.putNoClobber(self.base.allocator, owned_url, .{ .state = .{ .loading = script } });
     errdefer _ = self.preloaded_scripts.remove(owned_url);
 
     if (comptime IS_DEBUG) {
         log.debug(.http, "script queue", .{ .url = owned_url, .ctx = "preload" });
     }
 
-    // Tracked in async_scripts only while in flight so shutdown/reset can free a
-    // never-consumed preload; preloadDoneCallback moves ownership to the map.
-    self.base.async_scripts.append(&script.node);
-
-    // Guard against a synchronous completion re-entering evaluate() mid-parse,
-    // the same reason getAsyncImport does.
-    const was_evaluating = self.base.is_evaluating;
-    self.base.is_evaluating = true;
-    defer self.base.is_evaluating = was_evaluating;
-
-    frame.makeRequest(.{
+    try frame.makeRequest(.{
         .ctx = script,
         .url = owned_url,
         .method = .GET,
@@ -156,10 +147,7 @@ pub fn preloadScript(self: *ScriptManager, url: []const u8) !void {
         .data_callback = Script.dataCallback,
         .done_callback = PreloadedScript.doneCallback,
         .error_callback = PreloadedScript.errorCallback,
-    }) catch |err| {
-        self.base.async_scripts.remove(&script.node);
-        return err;
-    };
+    });
 }
 
 fn waitForPreload(self: *ScriptManager, url: [:0]const u8) ?*Script {
@@ -185,7 +173,6 @@ fn waitForPreload(self: *ScriptManager, url: [:0]const u8) ?*Script {
                 _ = self.preloaded_scripts.remove(url);
                 return script;
             },
-            .err => return null,
         }
     }
 }
@@ -430,18 +417,16 @@ pub fn staticScriptsDone(self: *ScriptManager) void {
 }
 
 const PreloadedScript = struct {
-    state: State = .loading,
+    state: State,
 
     const State = union(enum) {
-        err,
-        loading,
+        loading: *Script,
         done: *Script,
     };
 
     pub fn deinit(self: PreloadedScript) void {
         switch (self.state) {
-            .done => |script| script.deinit(),
-            else => {},
+            inline else => |script| script.deinit(),
         }
     }
 
@@ -453,12 +438,7 @@ const PreloadedScript = struct {
         }
 
         const self: *ScriptManager = @fieldParentPtr("base", script.manager);
-        // Hand ownership to the map; the blocking path adopts the body via
-        // waitForPreload, and reset() frees the .done entry.
-        self.base.async_scripts.remove(&script.node);
         self.preloaded_scripts.getPtr(script.url).?.state = .{ .done = script };
-
-        self.base.evaluate();
     }
 
     fn errorCallback(ctx: *anyopaque, err: anyerror) void {
@@ -470,12 +450,7 @@ const PreloadedScript = struct {
         }
 
         const self: *ScriptManager = @fieldParentPtr("base", script.manager);
-        self.base.async_scripts.remove(&script.node);
         _ = self.preloaded_scripts.remove(script.url);
         script.deinit();
-
-        if (self.base.shutdown == false) {
-            self.base.evaluate();
-        }
     }
 };
--- a/src/browser/ScriptManager.zig
+++ b/src/browser/ScriptManager.zig
@@ -46,23 +46,38 @@ frame: *Frame,
 // "load" event).
 frame_notified_of_completion: bool,
 
+// scripts loaded based on a <link rel=preload as=script href=...> found during parsing
+preloaded_scripts: std.StringHashMapUnmanaged(PreloadedScript),
+
 pub fn init(allocator: Allocator, http_client: *HttpClient, frame: *Frame) ScriptManager {
     var base = ScriptManagerBase.init(allocator, http_client, .{ .frame = frame });
     base.tail_hook = tailHook;
     return .{
-        .frame = frame,
         .base = base,
+        .frame = frame,
+        .preloaded_scripts = .empty,
         .frame_notified_of_completion = false,
     };
 }
 
 pub fn deinit(self: *ScriptManager) void {
+    self.freeDonePreloads();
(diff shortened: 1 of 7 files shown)


krichprollsch (approved):

navidemad:
Nice — overlapping the fetches is the right call, and the ownership handover (preload → `waitForPreload` adoption → caller frees) is handled carefully. I traced the arena lifetimes and the `is_evaluating` re-entrancy guards against the existing `preloadImport`/`waitForImport` path and they line up, including the re-fetch of the map entry after each `client.tick`. A few small things, each also dropped as an inline suggestion on the relevant line, plus one question worth a sentence from you.

**Suggestions (all minor):**

- **Dead `.err` state** — `PreloadedScript.errorCallback` *removes* the map entry rather than ever setting `.err`, so the variant and its `waitForPreload` arm are unreachable. Dropped it. The remove-then-fall-back-to-`syncRequest` behaviour is actually the nicer error path (a failed preload re-surfaces the error on the normal blocking path), so I added a comment saying so instead.
- **Two stale doc references** — the `unreachable` comments in `ScriptManagerBase` and the `async_scripts.append` comment refer to `preloadDoneCallback` / `preloadErrorCallback`; the real symbols are `PreloadedScript.doneCallback` / `errorCallback`.
- **Arena bucket** — `preloadScript` uses `.medium` while the sibling remote-script fetches (blocking `addFromElement`, `preloadImport`, `getAsyncImport`) all use `.large`. A consumed preload becomes a blocking script's body anyway, so matching `.large` keeps the four paths consistent.

**Tests** — the PR ships without coverage, and this ownership transfer is exactly what the debug leak detector is good at catching. The patch below (`git am`) adds `element/html/script/preload.html` + two `.js` assets: one preload consumed by a blocking `<script>` (asserts in-order execution — exercises adopt-and-free), and one preload with no consumer (fetched, never executed, freed on `reset()`). Both pass locally under the leak detector; the `element/html/script` directory runner picks them up with no registration.

<details>
<summary>Test patch — <code>git am &lt; patch</code></summary>

```diff
From a52d5e430830bd17934c983fe48c16f268ce3d27 Mon Sep 17 00:00:00 2001
From: Navid EMAD <navid.emad@yespark.fr>
Date: Tue, 9 Jun 2026 13:02:43 +0200
Subject: [PATCH] test,http: cover <link rel=preload as=script> loading

Exercises ScriptManager.waitForPreload adopting a preloaded body for a
blocking <script>, plus a preload with no consumer (fetched, never executed,
freed on reset). Both run under the debug-build leak detector.

Picked up automatically by the element/html/script directory runner.
---
 .../tests/element/html/script/preload.html    | 38 +++++++++++++++++++
 .../tests/element/html/script/preload.js      |  2 +
 .../element/html/script/preload_unused.js     |  4 ++
 3 files changed, 44 insertions(+)
 create mode 100644 src/browser/tests/element/html/script/preload.html
 create mode 100644 src/browser/tests/element/html/script/preload.js
 create mode 100644 src/browser/tests/element/html/script/preload_unused.js

diff --git a/src/browser/tests/element/html/script/preload.html b/src/browser/tests/element/html/script/preload.html
new file mode 100644
index 00000000..1a20ab7b
--- /dev/null
+++ b/src/browser/tests/element/html/script/preload.html
@@ -0,0 +1,38 @@
+<!DOCTYPE html>
+<script src="../../../testing.js"></script>
+
+<!--
+  The preload hint is declared before the <script> that consumes it, so by the
+  time the parser blocks on `preload.js` the fetch is already in flight (or
+  done). ScriptManager.waitForPreload adopts that body instead of issuing a
+  fresh syncRequest, then the consumed entry is freed by the caller. Both halves
+  of that ownership handover run under the leak detector here.
+-->
+<link rel="preload" as="script" href="preload.js">
+
+<!--
+  A preload with no matching <script> is fetched but never executed. It lingers
+  as a `.done` (or still-loading) entry until reset() frees it — exercises
+  freeDonePreloads / async_scripts cleanup under the leak detector — and it must
+  not run on its own.
+-->
+<link rel="preload" as="script" href="preload_unused.js">
+
+<script id="setup">
+    var order = '';
+    testing.expectEqual('', order);
+</script>
+
+<!-- Blocking remote script: its body is adopted from the preload above. -->
+<script id="preloaded" src="preload.js"></script>
+
+<script id="after">
+    // The preloaded blocking script executed, in document order, before us.
+    order += 'b';
+    testing.expectEqual('ab', order);
+</script>
+
+<script id="unused_check">
+    // A preload hint that nothing consumes must never execute.
+    testing.expectEqual(undefined, window.unused_preload_ran);
+</script>
diff --git a/src/browser/tests/element/html/script/preload.js b/src/browser/tests/element/html/script/preload.js
new file mode 100644
index 00000000..658e1e9f
--- /dev/null
+++ b/src/browser/tests/element/html/script/preload.js
@@ -0,0 +1,2 @@
+order += 'a';
+testing.expectEqual('a', order);
diff --git a/src/browser/tests/element/html/script/preload_unused.js b/src/browser/tests/element/html/script/preload_unused.js
new file mode 100644
index 00000000..f0d04b8a
--- /dev/null
+++ b/src/browser/tests/element/html/script/preload_unused.js
@@ -0,0 +1,4 @@
+// Nothing consumes this preload, so it must never be evaluated. If it runs, the
+// flag trips the assertion in preload.html (and this fail() fires directly).
+window.unused_preload_ran = true;
+testing.fail('an unconsumed <link rel=preload as=script> must not execute');
-- 
2.50.1 (Apple Git-155)

```

</details>

> [!NOTE]
> **One thing to confirm:** an in-flight preload sits in `async_scripts`, which gates `scriptsCompletedLoading` (the load event) via `tailHook`. So a `<link rel=preload as=script>` with no consumer now delays the frame's load event until its fetch completes — previously preload links never hit the network. Defensible, since it's a genuine in-flight fetch, but it's an observable change for anything keying off load timing. Intended, or should preloads be tracked outside the load-gating list?

I only ran the `WebApi: Script` suite locally — not the full suite or CI.

navidemad (commented):
Inline suggestions for the three minor items from the summary above — drop-in ready.

src/browser/ScriptManager.zig
```diff
@@ -86,6 +101,95 @@ fn getHeaders(self: *ScriptManager) !HttpClient.Headers {
     return self.base.getHeaders();
 }
 
+pub fn preloadScript(self: *ScriptManager, url: []const u8) !void {
+    if (self.preloaded_scripts.contains(url)) {
+        return;
+    }
+
+    const frame = self.frame;
+    const arena = try frame.getArena(.medium, "SM.preloadScript");
```
navidemad: Bucket: match the sibling remote-fetch paths (`preloadImport` / `getAsyncImport` / blocking `addFromElement` all use `.large`).

```suggestion
    // .large to match the other remote-script fetch paths (addFromElement's
    // blocking branch, preloadImport, getAsyncImport): a consumed preload
    // becomes a blocking script's body, which would have used .large anyway.
    const arena = try frame.getArena(.large, "SM.preloadScript");
```

src/browser/ScriptManager.zig
```diff
@@ -86,6 +101,95 @@ fn getHeaders(self: *ScriptManager) !HttpClient.Headers {
...
+
+    try self.preloaded_scripts.putNoClobber(self.base.allocator, owned_url, .{});
+    errdefer _ = self.preloaded_scripts.remove(owned_url);
+
+    if (comptime IS_DEBUG) {
+        log.debug(.http, "script queue", .{ .url = owned_url, .ctx = "preload" });
+    }
+
+    // Tracked in async_scripts only while in flight so shutdown/reset can free a
+    // never-consumed preload; preloadDoneCallback moves ownership to the map.
```
navidemad: Fixes the `doneCallback` symbol reference, and flags the load-event gating for the next reader.

```suggestion
    // Tracked in async_scripts only while in flight so shutdown/reset can free a
    // never-consumed preload; PreloadedScript.doneCallback moves ownership to
    // the map once the fetch finishes. While in flight a preload keeps
    // async_scripts non-empty, so it also delays the frame's load event (see
    // tailHook) — the same gating any other in-flight async script gets.
```

src/browser/ScriptManager.zig
```diff
@@ -311,3 +428,54 @@ pub fn addFromElement(self: *ScriptManager, comptime from_parser: bool, script_e
...
     self.base.staticScriptsDone();
 }
+
+const PreloadedScript = struct {
+    state: State = .loading,
+
+    const State = union(enum) {
+        err,
+        loading,
+        done: *Script,
```
navidemad, line 424: `.err` is never assigned — `errorCallback` removes the entry instead — so drop it from the union.

```suggestion
    const State = union(enum) {
        loading,
        done: *Script,
```

src/browser/ScriptManager.zig
```diff
@@ -86,6 +101,95 @@ fn getHeaders(self: *ScriptManager) !HttpClient.Headers {
...
+                _ = client.tick(200, .sync_wait) catch return null;
+                continue;
+            },
+            .done => |script| {
+                // Preload scripts are single-use. We return it and it becomes
+                // the caller's responsibility to free.
+                _ = self.preloaded_scripts.remove(url);
+                return script;
+            },
+            .err => return null,
```
navidemad: …and remove its now-unreachable arm in `waitForPreload` (a removed entry already returns `null` via the `orelse` above).

```suggestion
            },
```

src/browser/ScriptManager.zig
```diff
@@ -311,3 +428,54 @@ pub fn addFromElement(self: *ScriptManager, comptime from_parser: bool, script_e
...
+    fn errorCallback(ctx: *anyopaque, err: anyerror) void {
+        const script: *Script = @ptrCast(@alignCast(ctx));
+        if (script.status == 404) {
+            log.info(.http, "script 404", .{ .req = script.url, .extra = "preload" });
+        } else {
+            log.warn(.http, "script fetch error", .{ .err = err, .req = script.url, .extra = "preload", .status = script.status });
+        }
+
+        const self: *ScriptManager = @fieldParentPtr("base", script.manager);
+        self.base.async_scripts.remove(&script.node);
```
navidemad: Document why a failed preload is dropped from the map rather than marked — it lets a later blocking `<script>` miss and fall back to `syncRequest`.

```suggestion
        const self: *ScriptManager = @fieldParentPtr("base", script.manager);
        // Drop the entry entirely rather than marking it failed: a later
        // blocking <script> then misses in waitForPreload and falls back to a
        // fresh syncRequest, which re-surfaces the error on the normal path.
        self.base.async_scripts.remove(&script.node);
```

src/browser/ScriptManagerBase.zig
```diff
@@ -646,6 +650,7 @@ pub const Script = struct {
                 entry.state = .{ .done = self };
                 entry.buffer = self.source.remote;
             },
+            .preload => unreachable, // preloads use ScriptManager.preloadDoneCallback
```
navidemad: Real symbol is `PreloadedScript.doneCallback`.

```suggestion
            .preload => unreachable, // preloads use ScriptManager.PreloadedScript.doneCallback
```

src/browser/ScriptManagerBase.zig
```diff
@@ -688,6 +693,7 @@ pub const Script = struct {
                 entry.state = .err;
             },
             .frame => self.executeCallback(comptime .wrap("error")),
+            .preload => unreachable, // preloads use ScriptManager.preloadErrorCallback
```
navidemad: Real symbol is `PreloadedScript.errorCallback`.

```suggestion
            .preload => unreachable, // preloads use ScriptManager.PreloadedScript.errorCallback
```

karlseguin:
Thanks @navidemad, I applied these in a follow up commit. I think `async_scripts` gating `scriptsCompletedLoading` was a more serious issue, so that was fixed. There's no guarantee that a preloaded script is ever going to get consumed, so load might never fire.

PR #2676, opened 2026-06-09
perf; Optimize IntersectionObserver

karlseguin:
`checkIntersection` can be called a lot...on every domChanged. And it can call a known expensive API: `getBoundingClientRect`.

This commit reworks the code to be able to exit from checkIntersection early. `_previous_states` is now seeded with the target so that, once reported, the lack of entry can be used to quickly exit.

When profiling airbnb, `checkIntersection` was reported 841 times. With this change, it was reported twice.

### DIFF
--- a/src/browser/webapi/IntersectionObserver.zig
+++ b/src/browser/webapi/IntersectionObserver.zig
@@ -140,8 +140,8 @@ pub fn observe(self: *IntersectionObserver, target: *Element, frame: *Frame) !vo
         try frame.registerIntersectionObserver(self);
     }
 
-    // Don't initialize previous state yet - let checkIntersection do it
-    // This ensures we get an entry on first observation
+    // false = tracked but not yet reported
+    try self._previous_states.put(self._arena, target, false);
 
     // Check intersection for this new target and schedule delivery
     try self.checkIntersection(target, frame);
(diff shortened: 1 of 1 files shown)


krichprollsch (approved):

navidemad:
Nice optimization — the early-exit on the `_previous_states` miss is clean, and deferring `getBoundingClientRect` until an actual transition is the right move.

One thing I want to confirm is intentional: once a target is reported intersecting it's removed from `_previous_states`, so the observer is effectively one-shot. A target detached after being reported (`parentNode` → null) won't emit `isIntersecting: false` anymore, and a re-attach won't re-fire. Since intersection here is just parent-presence, this only affects detach/reattach — niche, and lazy-load callers that `unobserve` after the first hit won't notice. So I think it's fine; I just want to make sure dropping that transition is deliberate rather than a side effect. If it is, a one-line comment (or a note in the PR body) that observers are one-shot would keep it from reading as a silent behavior change — the PR comment says "the only reachable transition is false → intersecting", but `true → false` is technically reachable via detach, it's just being dropped on purpose.

Two optional nits:
- `IntersectionData.is_intersecting` looks dead now — `checkIntersection` recomputes `is_now_intersecting` itself and nothing else reads the field.
- `_previous_states` only ever stores `false` (the `true` state is now represented by absence), so the value is vestigial. Not worth changing here, just slightly misleading.

Might also be worth a test asserting that a second `domChanged` after a report produces no new entries — that's the core of this change and nothing currently guards against a regression reintroducing repeated reporting.

karlseguin:
I swapped it to a set to better communicate the intent.
Yes, the change was intentional. We see too many sites where IntersectionObserver fires _way_ too much, often causing issues. I'm hoping this approach better balances the need that some sites have for IO to fire in order to progress versus over-triggering.

Issue #2680, opened 2026-06-09
Surface server-side agent-discovery signals: Link response header (RFC 8288) and robots.txt Content-Signal


navidemad:
Lightpanda drives AI agents, which puts it in a good spot to read the machine-readable signals some sites now publish specifically for agents. Two of them are standardized and cheap for a server to emit, and Lightpanda currently reads neither.

## 1. The `Link` response header (RFC 8288)

A server can advertise where its docs and API description live with a response header:

```
Link: <https://docs.example.com/>; rel="service-doc"
Link: <https://api.example.com/openapi.json>; rel="service-desc"
```

`service-doc` and `service-desc` are registered in RFC 8631, and `api` is in the IANA Link Relations registry. An agent that reads these can jump straight to the docs instead of scraping the page to guess where they are.

Today `src/browser/structured_data.zig` pulls link relations only from HTML `<link>` elements in the parsed document (`links`, `alternate`, alongside json-ld and open-graph). The HTTP `Link:` response header is never parsed, so a `rel="service-doc"` sent as a header is invisible to the driving agent.

## 2. `Content-Signal` in `robots.txt`

The contentsignals.org convention lets an operator state how content may be used:

```
User-agent: *
Content-Signal: ai-train=no, search=yes, ai-input=yes
```

`src/network/Robots.zig` (used by `RobotsLayer`, gated on `obey_robots`) parses `Allow`/`Disallow` only. `Content-Signal` lines are dropped, so an agent has no way to see the operator's stated preference.

## Proposal

- Parse the `Link:` response header where response headers are already processed, and surface the registered, agent-useful relations (at least `service-doc`, `service-desc`, `api`) through `StructuredData`, so they sit next to the existing `<link rel>` data and flow out via `src/mcp/`. The header is comma-separated and can repeat (RFC 8288 §3); a header-level rel should merge with any DOM-level link carrying the same rel rather than overwrite it.
- Extend the `robots.txt` parser to capture `Content-Signal` directives and expose the declared preferences to the agent. This is advisory metadata about usage, not a crawl gate, so it should stay separate from the `Allow`/`Disallow` enforcement that `obey_robots` already drives.

Both are read-only additions to extraction surfaces that already exist. Neither changes navigation or fetch behavior.

## References

- RFC 8288 — Web Linking
- RFC 8631 — Link Relation Types for Web Services (`service-doc`, `service-desc`, `service-meta`)
- IANA Link Relations registry
[...]

PR #2682, opened 2026-06-09
Surface server-side agent-discovery signals: Link response header (RFC 8288) + robots.txt Content-Signal

navidemad:
Implements #2680. Lightpanda now consumes the two machine-readable discovery signals that well-behaved sites publish for agents, instead of dropping them on the floor.

## Link response header (RFC 8288) → structured data

`collectStructuredData` already pulls `<link rel>` out of the DOM but never looked at the HTTP `Link:` response header. It now also parses that header from the frame's retained response headers (`frame._http_headers`) and surfaces the registered, agent-useful relations — `service-doc`, `service-desc` (RFC 8631) and `api` — under a new `linkHeaders` field:

```json
"linkHeaders": [
  { "rel": "service-doc", "href": "https://docs.example.com/" },
  { "rel": "service-desc", "href": "https://api.example.com/openapi.json" }
]
```

This reaches the agent through both existing transports with no extra plumbing: the MCP `structuredData` tool and the CDP `LP.getStructuredData` command. The field is emitted only when non-empty, so the structured-data shape is unchanged for sites without such a header (see Caveats).

The parser (`LinkHeaderIterator` + `extractRel` in `src/browser/structured_data.zig`) splits comma-separated link-values while respecting commas inside `<...>` targets and `"..."` quoted params (honouring RFC 7230 backslash escapes), resolves relative targets against the frame base, and de-dupes. The scan is iterative rather than recursive so stack usage stays bounded on a hostile header value.

## Content-Signal in robots.txt → CDP

`src/network/Robots.zig` parsed only `Allow`/`Disallow`. It now also captures the `Content-Signal` directive (https://contentsignals.org):

```
User-agent: *
Content-Signal: ai-train=no, search=yes, ai-input=yes
```

Parsed preferences live on a new `Robots.content_signals` field, follow the same specific-user-agent-else-wildcard precedence as the crawl rules, and are read back via `RobotStore.getContentSignals`. A new `LP.getContentSignal` CDP command returns them for the current document's host:

```json
{ "available": true, "contentSignals": [ { "name": "ai-train", "value": "no" } ] }
```

`Content-Signal` is advisory metadata. It is parsed into a separate field and never reaches `isAllowed`, so it cannot move the crawl gate — there is a test that asserts exactly that.

## Caveats

> [!IMPORTANT]
> `LP.getContentSignal` reports `available: false` unless robots.txt has already been fetched for the host, which today only happens when `obey_robots` is on (the `RobotsLayer` is what populates the store — `HttpClient.zig:233`). Fetching robots.txt purely to read content signals when `obey_robots` is off is a separate decision I kept out of this PR; happy to follow up if you'd want it.

- Adding `linkHeaders` to `StructuredData` also changes the `LP.getStructuredData` result shape. It is conditionally emitted (absent when there is no `Link:` header), so clients that do not look for it are unaffected — but worth flagging for anyone asserting the exact key set.
- Header parsing is limited to `service-doc`/`service-desc`/`api`. Other relations are ignored on purpose, to keep the output to what an agent can act on. Widening the set later is a one-line change.

## Where to focus review

- `collectLinkHeaders` / `LinkHeaderIterator` — the RFC 8288 splitting, in particular the quoted-string and backslash handling and the iterative scan over an attacker-controlled header value.
- `parseRulesWithUserAgent` in `Robots.zig` — the specific-vs-wildcard selection done in lockstep with the crawl rules, and the free/deinit paths for the duped signal strings (the test runner fails any test that leaks).

## Test plan

- [x] `make test` — full suite green (800 tests), no leaks reported by the runner.
- [x] RFC 8288 parser units: quoted and unquoted rel, multiple links, comma inside a quoted param, escaped quote, stylesheet filtered out, empty header.
- [x] Link relations surfaced through `collectStructuredData` (integration test that injects response headers on the frame).
- [x] Content-Signal: capture, casing normalisation, specific-over-wildcard precedence, `RobotStore` round-trip, and the gating-unaffected assertion.
- [x] `LP.getContentSignal` wired end to end (returns the documented shape; `available: false` when no robots.txt has been fetched).

Refs #2680

### DIFF
--- a/src/browser/structured_data.zig
+++ b/src/browser/structured_data.zig
@@ -223,12 +223,6 @@ pub fn collectStructuredData(
     };
 }
 
-/// Registered link relations we surface from the HTTP `Link:` response header
-/// (RFC 8288). `service-doc`/`service-desc` come from RFC 8631; `api` is in the
-/// IANA Link Relations registry. These let an agent discover a site's developer
-/// docs / API description without scraping.
-const header_link_rels = [_][]const u8{ "service-doc", "service-desc", "api" };
-
 /// Parse every `Link:` response header retained on the frame, surfacing only the
 /// registered, agent-useful relations. Relative targets are resolved against the
 /// frame's base URL, mirroring `collectLink`.
@@ -237,8 +231,14 @@ fn collectLinkHeaders(
     frame: *Frame,
     out: *std.ArrayList(LinkRel),
 ) !void {
+    // Registered link relations we surface from the HTTP `Link:` response header
+    // (RFC 8288).
+    const header_link_rels = [_][]const u8{ "service-doc", "service-desc", "api" };
+
     for (frame._http_headers.items) |header| {
-        if (!std.ascii.eqlIgnoreCase(header.name, "link")) continue;
+        if (!std.ascii.eqlIgnoreCase(header.name, "link")) {
+            continue;
+        }
 
         var it: LinkHeaderIterator = .{ .value = header.value };
         while (it.next()) |entry| {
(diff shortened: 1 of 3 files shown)


karlseguin:
@arrufat I'm fine with the mechanics of the code, but the result impacts you more

1. Adds a lp.getContentSignal method to CDP to return content signals extracted from robots.txt (requires --obey-robots, https://contentsignals.org/)

2. Adds a new `linkHeaders` field to structured data based on specific "Link" HTTP response headers (RFC 8288)

arrufat:
I'm good with the mechanics :)

PR #2688, opened 2026-06-10
script: Force pending script evaluation after evalutate finishes

karlseguin:
This commit is in the same vein as 0d9482ccbf7e856a74abfde860a11cb00e1c273f. A dynamic async script which completes during an evalute() does not evaluate if `is_evaluating == true`. That's good, but we need to make sure that it _does_ eventually get evaluated. As-is, we rely on a done callback to call evaluate() again which works for all but the last script.

This commit creates a more robust reentrancy flow. is_evaluating still guards against reentrancy, but a new flag, `evaluate_pending` is set to true. Whenever `is_evaluating` is set to false if `evaluate_pending` is true, we evaluate().

Similalry, during evaluate(), we better handle completed scripts during evaluate() and process them immediately (via an outer loop).

### DIFF
--- a/src/browser/Frame.zig
+++ b/src/browser/Frame.zig
@@ -1762,7 +1762,7 @@ pub fn loadExternalStylesheet(self: *Frame, link: *Element.Html.Link, href: []co
     const sm = &self._script_manager.base;
     const was_evaluating = sm.is_evaluating;
     sm.is_evaluating = true;
-    defer sm.is_evaluating = was_evaluating;
+    defer sm.endEvaluationWindow(was_evaluating);
 
     var response = http_client.syncRequest(arena, .{
         .url = resolved,
--- a/src/browser/ScriptManager.zig
+++ b/src/browser/ScriptManager.zig
@@ -157,7 +157,7 @@ fn waitForPreload(self: *ScriptManager, url: [:0]const u8) ?*Script {
 
     const was_evaluating = self.base.is_evaluating;
     self.base.is_evaluating = true;
-    defer self.base.is_evaluating = was_evaluating;
+    defer self.base.endEvaluationWindow(was_evaluating);
 
     var client = self.base.client;
     while (true) {
@@ -333,7 +333,7 @@ pub fn addFromElement(self: *ScriptManager, comptime from_parser: bool, script_e
 
         const was_evaluating = self.base.is_evaluating;
         self.base.is_evaluating = true;
-        defer self.base.is_evaluating = was_evaluating;
+        defer self.base.endEvaluationWindow(was_evaluating);
 
         if (is_blocking) {
             if (self.waitForPreload(url)) |pre| {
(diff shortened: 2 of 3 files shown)


navidemad (commented):
The reentrancy mechanism looks sound to me: `evaluate_pending` is cleared at the top of each outer-loop iteration and only set by reentrant calls, so the loop converges (an iteration that produces no new work exits). Swapping the defer order in `addFromElement` (`script.deinit()` before `endEvaluationWindow`) also looks right, since closing the window can now run arbitrary JS.

A few things I noticed:

**1. `Frame.loadExternalStylesheet` opens the same window but never drains the flag** (`Frame.zig:1763-1765`). It sets `is_evaluating` around a `syncRequest` for the same reason `addFromElement` does, but restores with a plain `defer sm.is_evaluating = was_evaluating`. The sync fetch pumps the shared HTTP client, so an async script's `doneCallback` can fire mid-fetch, get swallowed by the guard, and set `evaluate_pending`, which then goes stale if no later `doneCallback` arrives. That's the same "last script" hang this PR fixes. Should it use `sm.endEvaluationWindow(was_evaluating)` too? (Mitigating: `load_external_stylesheets` is off by default.)

**2. The `tail_hook` can leave the flag stale.** The hook runs after the loop breaks but before `defer self.is_evaluating = false` fires. It dispatches DOMContentLoaded/load handlers, i.e. arbitrary JS. If that JS triggers a synchronous `evaluate()`, e.g. `document.open/write/close` → `scriptCreatedParseDone()` → `evaluate()`, it sets `evaluate_pending = true`, and the defer then clears `is_evaluating` without consulting it. Deferred scripts produced by a script-created parser inside a load handler would never run, which is the scenario from 0d9482cc. Changing the defer to `defer self.endEvaluationWindow(false)` would close this; the recursion is bounded since `evaluate()` clears the flag each iteration, and `documentIsLoaded` / `scriptsCompletedLoading` are already idempotent.

**3. Question on `getAsyncImport` / `waitForImport`.** The comment above the window in `getAsyncImport` says the point is to avoid a call from v8 resulting in script evaluation. With `endEvaluationWindow`, when `was_evaluating == false`, we now call `evaluate()` at window close. That's still inside the v8 callstack (import resolution), just before unwinding. If the original concern was v8 reentrancy itself, this reintroduces it; if it was only about evaluation during request setup, this is fine. Either way the comment probably deserves an update.

**4. Minor: `reset()` doesn't clear `evaluate_pending`.** A flag left set survives navigation and triggers a spurious `evaluate()` on the next page. Harmless today (queues are empty and `static_scripts_done` is false, so it returns before the hook), but one line in `reset()` would keep it from becoming a question later.

**5. Typos:** "reentrency" in the field comment (`ScriptManagerBase.zig:99`), "we're done the outer-evaluate" (line 413).

On CI: `demo-runner (true, false, true)` was cancelled after 15m while the other combos finish in ~1m40s, so the runner hung. That's the exact failure mode a bug in this code would produce (load never firing). I checked recent main runs and the job is flaky there too (two recent failures on different matrix combos), so it's probably not this PR, but given the subject a re-run before merge seems worth it.

PR #2697, opened 2026-06-11
ci: Include v8 version in Snapshot cache key

karlseguin:


karlseguin:
@krichprollsch I think we ran into this before, and I ran into it again. I think this is right? Does require bumping a 3rd v8 version though.

krichprollsch (approved):

PR #2702, opened 2026-06-11
agent: improve save system prompt

francisbouvier:
Avoid redundancy between general instructions and skill-like docs + be more clear on not using big evaluate in favor of extract + top-level JS logic.


arrufat (approved):

arrufat:
@francisbouvier are you ok with me merging this?

PR #2703, opened 2026-06-11
`URL`: Rework `resolve` with new URL implementation

nikneym:
`resolve` now uses rust-url; also introduces fixes for various `URL` and `URLSearchParams` WPT tests.


karlseguin:
Haven't looked it through, but there seems to be issues (timeouts)/regression with WPT /url/a-element*  tests (e..g /url/a-element-origin.html).

karlseguin (commented):
The PR is good, except for 2 issues.

First, `lightpanda fetch lightpanda.io` no longer works. It used to, and that was intentional (to mimic curl). This is true for any callers into Frame.navigate, so I assume this is also broken for CDP.

Second, while WPT tests are generally up (yay), at least a couple are behaving poorly. `/url/a-element-origin.html` is a good example, if you fetch it with `--wait-script "window.report && report.complete"` (which is what I normally use, and what our WPT runner does), it hangs forever. Something has broken the test.

src/browser/webapi/Node.zig
```diff
@@ -586,6 +586,15 @@ pub fn resolveURL(self: *const Node, url: anytype, frame: *Frame, opts: ResolveU
     return URL.resolve(allocator, owner_frame.base(), url, .{ .encoding = owner_frame.charset });
 }
 
+// Same as `resolveURL` but can't return `TypeError`, this is needed for multiple
+// getters throught codebase. Returns back passed `url` for `TypeError`s.
```
karlseguin:
```suggestion
// getters throughout codebase. Returns provided `url` on `TypeError`.
```

karlseguin (approved):

PR #2705, opened 2026-06-11
Agent feedback fixes

arrufat:


src/browser/Session.zig
```diff
@@ -401,6 +401,20 @@ pub fn runner(self: *Session, opts: Runner.Opts) !Runner {
     return Runner.init(self, opts);
 }
 
+/// Page transfers run on the session thread's curl multi; left unserviced
+/// while the frontend waits on input they die on curl's wall-clock timeout.
+/// Returns how long the caller may block before pumping again.
+pub fn idleSlice(self: *Session) u31 {
+    const quiet_ms = 250;
+    self.processDestroyQueues();
```
karlseguin, line 409: Move `session.processDestroyQueues();` to `Runner._tick` so it becomes more self-contained?

karlseguin: Nope, claude pointed out that opens a UAF, and it's right.

karlseguin (approved):

Issue #2704, opened 2026-06-11
CDP: cannot set a browser User-Agent — Mozilla UAs rejected by setUserAgentOverride and ignored by setExtraHTTPHeaders


staylor:
## Summary

On current `main` (`1.0.0-dev.6709+d695ce10`), a CDP client **cannot set a realistic browser `User-Agent`** through either Playwright mechanism. Two independent behaviors combine so that any UA beginning with `Mozilla/5.0` (i.e. every real Chrome/Firefox/Safari UA) is silently dropped and the origin server receives `User-Agent: Lightpanda/1.0`:

1. **`Emulation.setUserAgentOverride` rejects any UA containing `"Mozilla"`** (the go-rod workaround), so Playwright's `newContext({ userAgent })` is a no-op for browser-like UAs.
2. **`Network.setExtraHTTPHeaders` does not apply a caller-supplied `User-Agent`** (it works for every *other* header), so the documented fallback of forcing the header doesn't work either.

The net effect: there is no CDP path to present a normal browser User-Agent to origins. This defeats the "drop-in Chrome/Chromium replacement" use case, where matching `navigator.userAgent`, the outbound header, and the TLS fingerprint to a real browser is the whole point.

## Environment

- Lightpanda `1.0.0-dev.6709+d695ce10` (built from `main`)
- Driver: `playwright-core` 1.59.1 via `chromium.connectOverCDP(...)`

## Reproduction

Start a server, then run the script below (needs `playwright-core` installed):

```sh
./zig-out/bin/lightpanda serve --host 127.0.0.1 --port 9222 --log-level warn
node repro-ua-issue.mjs
```

<details><summary><code>repro-ua-issue.mjs</code></summary>

```js
import http from "node:http";
import { chromium } from "playwright-core";

const CDP = process.env.CDP || "http://127.0.0.1:9222";
const CONTROL = "X-Repro-Control"; // non-UA control header

async function probe({ label, ua, viaExtraHeaders }) {
  let received = "<no request>";
  let control = "<none>";
  const server = http.createServer((req, res) => {
    if (req.url === "/page") {
      received = req.headers["user-agent"] ?? "<absent>";
      control = req.headers[CONTROL.toLowerCase()] ?? "<absent>";
    }
    res.writeHead(200, { "content-type": "text/html" });
    res.end("<!doctype html><title>ok</title>body");
  });
  await new Promise((r) => server.listen(0, "127.0.0.1", r));
  const { port } = server.address();

[...]

PR #2706, opened 2026-06-11
fix(cdp): let setExtraHTTPHeaders override built-in headers like User-Agent

staylor:
## What

`Network.setExtraHTTPHeaders` could not override the request `User-Agent`. Adds `Headers.set` (replace-by-name) and uses it when applying CDP extra headers, so a caller-supplied header overrides a built-in default of the same name.

Addresses the `setExtraHTTPHeaders` half of #2704.

## Why

Requests are seeded with a default `User-Agent: Lightpanda/1.0` in `http.Headers.init`. `httpRequestStart` then applied each CDP extra header with a plain `curl_slist` **append**, so a caller-supplied `User-Agent` produced two `User-Agent` entries in the slist. libcurl keeps the first, so the override was silently dropped and origins always saw `Lightpanda/1.0`. Every *other* header worked, because it had no built-in default to collide with.

This is the standard mechanism a Playwright/Puppeteer client uses to set a header, and Chrome's `Network.setExtraHTTPHeaders` overrides default request headers — so the append-only behavior diverges from Chrome.

## How

`Headers.set` rebuilds the header list, dropping any existing entry whose name matches (case-insensitive) the new header, then appends the new one. `httpRequestStart` now calls `set` instead of `add` for `bc.extra_headers`.

## Verification

Unit tests added in `src/network/http.zig` (replace, case-insensitive match, add-new-preserves-defaults). Full suite green (`794 of 794`).

End-to-end against a local echo server, driven by `playwright-core` over `connectOverCDP`, before vs. after this change:

| `setExtraHTTPHeaders({"User-Agent": ua})` | before | after |
|---|---|---|
| `ua = "CustomAgent/1.0"` | server saw `Lightpanda/1.0` | server saw `CustomAgent/1.0` |
| `ua = "Mozilla/5.0 … Chrome/120"` | server saw `Lightpanda/1.0` | server saw `Mozilla/5.0 … Chrome/120` |

A non-UA control header (`X-Repro-Control`) arrived correctly both before and after, confirming the bug was specific to colliding with the default `User-Agent`.

## Scope note

This does **not** touch the `Emulation.setUserAgentOverride` "ignore UA containing Mozilla" heuristic (the other half of #2704). That behavior is deliberate (keeps go-rod working), so I've left it as a maintainer decision; with this change, `setExtraHTTPHeaders` is a working, Chrome-compatible way to set a browser-like `User-Agent` on the wire.

### DIFF
--- a/src/cdp/domains/network.zig
+++ b/src/cdp/domains/network.zig
@@ -307,9 +307,11 @@ pub fn httpRequestStart(bc: *CDP.BrowserContext, msg: *const Notification.Reques
     const frame_id = req.frame_id;
     const frame = bc.session.findFrameByFrameId(frame_id) orelse return;
 
-    // Modify request with extra CDP headers
+    // Modify request with extra CDP headers. Use set (replace by name) so a
+    // caller-supplied header overrides a built-in default of the same name
+    // (e.g. User-Agent) instead of producing a duplicate libcurl drops.
     for (bc.extra_headers.items) |extra| {
-        try req.headers.add(extra);
+        try req.headers.set(extra);
     }
 
     // We're missing a bunch of fields, but, for now, this eems like enough
--- a/src/cdp/domains/network.zig
+++ b/src/cdp/domains/network.zig
@@ -108,6 +108,16 @@ fn setExtraHTTPHeaders(cmd: *CDP.Command) !void {
     try extra_headers.ensureTotalCapacity(arena, params.headers.map.count());
     var it = params.headers.map.iterator();
     while (it.next()) |header| {
+        // Reject a non-printable User-Agent, but (unlike
+        // Emulation.setUserAgentOverride) allow Mozilla: this is the
+        // intended escape hatch for a real browser-like UA (#2704).
+        if (std.ascii.eqlIgnoreCase(header.key_ptr.*, "user-agent")) {
+            for (header.value_ptr.*) |c| {
+                if (!std.ascii.isPrint(c)) {
+                    return cmd.sendError(-32602, "User agent contains non-printable characters", .{});
+                }
+            }
+        }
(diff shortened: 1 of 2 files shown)


karlseguin:
If user-agent is set, this needs to go through the same `Config.validateUserAgent(ua) ` as `Emulation.setUserAgentOverride()` and the --user-agent flag.

staylor:
Updated in 4deff4fb. `Network.setExtraHTTPHeaders` now rejects a non-printable `User-Agent` with the same `-32602` CDP error that `Emulation.setUserAgentOverride()` and `--user-agent` use.

One intentional deviation: I did **not** apply `validateUserAgent`'s reserved (Mozilla) check on this path. Applying the full check would drop any `User-Agent` containing "Mozilla", which is exactly the real browser-like UA this PR is meant to let through (see the `Mozilla/5.0 … Chrome/120` row in the verification table). The Mozilla heuristic lives in `Emulation.setUserAgentOverride` deliberately (keeps go-rod Lightpanda-branded); `setExtraHTTPHeaders` is the deliberate, Chrome-compatible escape hatch around it. So I kept the printable check but left out the reserved check — happy to apply the full `validateUserAgent` instead if you'd rather have consistency over the escape-hatch behavior.

karlseguin:
From https://github.com/lightpanda-io/browser/pull/2100 and other conversations, we do not want to allow "Mozilla" in the user-agent. I created https://github.com/lightpanda-io/browser/pull/2735 as an alternative to this PR, rather than commit over this PR, since 2735 changes the core intent.

@krichprollsch

krichprollsch:
I agree with #2735, the `user-agent` provided via `setExtraHTTPHeaders` must use the same validation path that other methods to override it.

PR #2717, opened 2026-06-12
cdp: raise inbound WebSocket message cap to 100MB

### LINKED ISSUE #2716: cdp: WebSocket closes (1009) on inbound messages over 512KB — large Runtime.evaluate payloads kill the connection
## Summary

Any inbound CDP WebSocket message larger than `CDP_MAX_MESSAGE_SIZE` (512KB + 154 bytes of frame overhead, `src/Config.zig`) makes the server close the connection with WebSocket close code 1009. No JSON-RPC error is sent — the socket just dies, and every pending command plus all session state (targets, sessions, enabled domains) is lost with it.

The most common trigger is `Runtime.evaluate` with a large `expression`: injecting any bundled JS library over CDP (a widely-used accessibility-audit bundle is 553KB minified, so it crosses the threshold on its own) kills the connection. Any method whose serialized JSON crosses the threshold hits the same wall — `Page.addScriptToEvaluateOnNewDocument` with a big script, `Fetch.fulfillRequest` with a base64 body, etc. The client gets zero diagnostics: from its side, the browser connection drops mid-command.

Chrome's DevTools server accepts inbound messages up to 100MB ([`kReceiveBufferSizeForDevTools` in `content/browser/devtools/devtools_http_handler.cc`](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/devtools/devtools_http_handler.cc)), so CDP clients and tooling reasonably assume payloads of this size are fine.

## Today's behavior

Empirical bisect on nightly 6703 (macOS aarch64) via `Runtime.evaluate` expression payloads: a ~512,056-byte message succeeds; a ~600,056-byte message gets no response and the server closes the socket with code 1009. The cap is enforced by the WS framer — `src/cdp/Connection.zig` instantiates `WS.Reader(true, Config.CDP_MAX_MESSAGE_SIZE)`, and `next()` in `src/network/WS.zig` returns `error.TooLarge` as soon as the frame header declares a length above the cap, which maps to the 1009 close frame.

```mermaid
sequenceDiagram
    participant LP as Lightpanda CDP server
    participant V8 as V8
    Client->>LP: Runtime.evaluate with 600KB expression
    LP->>LP: frame header declares length above 512KB cap
    LP-->>Client: WS close frame code 1009
    Note over Client,LP: no JSON-RPC response, pending commands and session state lost
    Note over LP,V8: V8 never sees the script
```

## Expected behavior

The message is framed and dispatched like any other; the client gets its JSON-RPC response. This matches Chrome, which accepts inbound DevTools messages up to 100MB.

```mermaid
sequenceDiagram
    participant LP as Lightpanda CDP server
    participant V8 as V8
    Client->>LP: Runtime.evaluate with 600KB expression
    LP->>LP: reader buffer grows on demand
    LP->>V8: evaluate script
    V8-->>LP: result
    LP-->>Client: JSON-RPC result for the command id
```

## Proposed fix

The reader buffer is already lazily allocated — it starts at 16KB and grows on demand (`growBuffer` in `src/network/WS.zig`) — so `CDP_MAX_MESSAGE_SIZE` is purely a hostile-input ceiling, not a per-connection allocation. Well-behaved clients pay nothing for a higher ceiling.

Proposal: raise `CDP_MAX_MESSAGE_SIZE` to 100MB, matching Chrome's inbound ceiling, and keep it as a ceiling (a declared-length cap is still worth having against garbage/hostile input).

Alternative considered: replying with a JSON-RPC error (`"message too large"`) instead of closing. That can't be done cleanly — the command `id` lives inside the JSON payload, so it can't be echoed back without buffering the very frame we're refusing, and silently skipping a frame would desync the client anyway. Raising the ceiling is simpler and matches client expectations.

## Reproducer

Self-contained: one shell script + one Node CDP driver (raw `ws`, installed via `npm install --no-save ws`). No fixture page needed — the bug is at the framing layer, `about:blank` suffices.

`repro.sh`:
```bash
#!/usr/bin/env bash
# Reproducer: CDP WebSocket closes (code 1009) on inbound messages > 512KB.
#
# Prerequisites: node (any recent). `npm install --no-save ws` runs once below.
# Run:           bash repro.sh           (uses lightpanda from $PATH)
#                LIGHTPANDA_BIN=/path/to/lightpanda bash repro.sh
#
# Expected today (bug):  the ~400KB control Runtime.evaluate returns, the
#                        ~600KB one gets no response — the server closes the
#                        socket with code 1009. "FAIL: ..." and exit 1.
# Expected after fix:    both evaluates return. "OK: ..." twice and exit 0.
set -u
cd "$(dirname "$0")"

LP="${LIGHTPANDA_BIN:-lightpanda}"
command -v node >/dev/null || { echo "node not found"; exit 2; }
[ -d node_modules/ws ] || npm install --no-save ws >/dev/null 2>&1

# Free the port from any previous run.
lsof -ti :9333 2>/dev/null | xargs kill 2>/dev/null
sleep 0.3

LIGHTPANDA_DISABLE_TELEMETRY=true "$LP" serve --host 127.0.0.1 --port 9333 >lightpanda.log 2>&1 &
LP_PID=$!
trap 'kill $LP_PID 2>/dev/null' EXIT
sleep 2

node repro.js
exit $?
```

`repro.js`:
```js
// Reproducer: inbound CDP WebSocket messages larger than ~512KB close the
// connection (WS close code 1009) instead of being processed.
//
// Control: a ~400KB Runtime.evaluate succeeds on both broken and fixed builds.
// Bug:     a ~600KB Runtime.evaluate gets no JSON-RPC response; the server
//          closes the socket with code 1009 and all session state is lost.
//
// Usage: node repro.js   (expects `lightpanda serve` on ws://127.0.0.1:9333)

const ws = new WebSocket('ws://127.0.0.1:9333/');
let nextId = 1;
let closeCode = null;
const pending = new Map();

ws.on('message', (raw) => {
    pending.delete(m.id);
    else resolve(m.result);
  }
});
ws.on('close', (code) => {
  closeCode = code;
  for (const { reject } of pending.values()) {
    reject(new Error(`WebSocket closed (code ${code}) before a response arrived`));
  }
  pending.clear();
});
ws.on('error', (e) => {
  for (const { reject } of pending.values()) reject(e);
  pending.clear();
});

function send(method, params = {}, sessionId) {
  const id = nextId++;
  });
}

// Build a Runtime.evaluate whose JSON message is approximately `bytes` long
// (a big string literal assignment, like a client injecting a JS bundle).
function evalOfSize(bytes, sessionId) {
  const expression = `window.x = "${'x'.repeat(bytes)}"; window.x.length`;
  return send('Runtime.evaluate', { expression, returnByValue: true }, sessionId);
}

async function main() {
  await new Promise((r) => ws.on('open', r));
  const { targetId } = await send('Target.createTarget', { url: 'about:blank' });
  const { sessionId } = await send('Target.attachToTarget', { targetId, flatten: true });

  // Control: ~400KB message, under the limit. Works on every build.
  const small = await evalOfSize(400 * 1024, sessionId);
  console.log(`OK: ~400KB Runtime.evaluate returned (string length ${small.result.value})`);

  // ~600KB message, over the 512KB limit.
  try {
    const big = await evalOfSize(600 * 1024, sessionId);
    console.log(`OK: ~600KB Runtime.evaluate returned (string length ${big.result.value})`);
    process.exit(0);
  } catch (e) {
    console.log(`FAIL: ~600KB Runtime.evaluate — ${e.message}`);
    if (closeCode !== null) console.log(`      server closed the WebSocket with code ${closeCode}`);
    process.exit(1);
  }
}

main().catch((e) => {
  console.log(`FAIL: ${e.message}`);
  process.exit(1);
});
```

**Run**: `bash repro.sh`

Today (nightly 6703):
```
OK: ~400KB Runtime.evaluate returned (string length 409600)
FAIL: ~600KB Runtime.evaluate — WebSocket closed (code 1009) before a response arrived
      server closed the WebSocket with code 1009
```
exit 1.

Expected after the fix:
```
OK: ~400KB Runtime.evaluate returned (string length 409600)
OK: ~600KB Runtime.evaluate returned (string length 614400)
```
exit 0.

## Likely fix location

`src/Config.zig` — `CDP_MAX_MESSAGE_SIZE`. The buffer growth machinery in `src/network/WS.zig` already handles arbitrary sizes; only the ceiling needs to move. The existing 1009 assertion in `src/Server.zig` ("Client: read invalid websocket message") encodes the current limit and needs to move with it.

## Environment

- Lightpanda: nightly 6703 (`b0426bab`); constant confirmed in `main` HEAD `e7e605b6`
- OS: macOS aarch64 (also reproduced on Linux x86_64)
- CDP client: Node `ws` (raw WebSocket)

navidemad:
## What

Inbound CDP WebSocket messages larger than `CDP_MAX_MESSAGE_SIZE` (512KB + 154 bytes of frame overhead) close the connection with code 1009 and no JSON-RPC response — every pending command and all session state is lost. A `Runtime.evaluate` whose expression is a bundled JS library (553KB minified is a common real-world size) kills the connection on its own. This PR raises the cap to 100MB, matching Chrome's inbound DevTools ceiling ([`kReceiveBufferSizeForDevTools`](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/devtools/devtools_http_handler.cc)).

Closes #2716.

## Root cause

`src/cdp/Connection.zig` instantiates `WS.Reader(true, Config.CDP_MAX_MESSAGE_SIZE)`. In `src/network/WS.zig`, `Reader.next()` returns `error.TooLarge` as soon as a frame header declares a length above the cap; `errorReply` maps that to the 1009 close frame and the worker tears the connection down. The cap predates the lazily-growing reader buffer: since `init` starts at 16KB and `growBuffer` expands on demand, the constant is purely a ceiling — it no longer protects any preallocated buffer, it just rejects messages that Chrome would accept.

```mermaid
flowchart LR
    A[Runtime.evaluate 600KB message] --> B[frame length above 512KB cap]
    B --> C[error.TooLarge]
    C --> D[WS close 1009, session lost]
    style B fill:#fdd
    style C fill:#fdd
    style D fill:#fdd
```

## Fix

- `src/Config.zig` — `CDP_MAX_MESSAGE_SIZE`: raised from 512KB to 100MB (Chrome's inbound ceiling), with a comment explaining it is a hostile-input ceiling, not a per-connection allocation. The `+14 +140` frame/control overhead is unchanged.
- `src/Server.zig` — updated the 1009 assertion in `Client: read invalid websocket message` to declare a 128MB frame (above the new cap — the ceiling is kept; a declared-length check against garbage input is still worth having).
- `src/Server.zig` — new test `Client: large websocket message`: a single 600KB text frame carrying a real `Browser.getVersion` command is framed, dispatched, and answered end-to-end instead of being rejected.

```mermaid
flowchart LR
    A[Runtime.evaluate 600KB message] --> B[reader buffer grows on demand]
    B --> C[frame dispatched to CDP]
    C --> D[JSON-RPC result returned]
    style B fill:#dfd
    style C fill:#dfd
    style D fill:#dfd
```

## Test

- **Unit tests**: `src/Server.zig` — `Client: large websocket message` (fails on `main` with a 1009 close, passes with this commit) and the updated 1009 ceiling assertion in `Client: read invalid websocket message`. Full `zig build test`: 792/792.
- **Reproducer**: the `repro.sh` from #2716 — a ~400KB control `Runtime.evaluate` plus a ~600KB one — exits 1 on nightly 6703 (600KB gets a 1009 close) and exits 0 with this commit (both return results).

## Notes

- Memory: the reader buffer starts at 16KB and only grows when a large frame actually arrives, so well-behaved clients pay nothing for the higher ceiling. A grown buffer is retained for the connection's lifetime (same behavior as today).
- A JSON-RPC "message too large" error instead of the close was considered and rejected: the command `id` lives inside the payload being refused, so it can't be echoed back without buffering the frame, and skipping a frame silently would desync the client. The 1009 close stays for messages above the new ceiling.

### DIFF
--- a/src/Config.zig
+++ b/src/Config.zig
@@ -32,9 +32,13 @@ const Allocator = std.mem.Allocator;
 pub const CDP_MAX_HTTP_REQUEST_SIZE = 4096;
 
 // max message size
+// 100MB matches Chrome's inbound DevTools ceiling (kReceiveBufferSizeForDevTools
+// in content/browser/devtools/devtools_http_handler.cc). This is only a
+// hostile-input ceiling, not a per-connection allocation: the reader buffer
+// starts at 16KB and grows on demand (see network/WS.zig).
 // +14 for max websocket payload overhead
 // +140 for the max control packet that might be interleaved in a message
-pub const CDP_MAX_MESSAGE_SIZE = 512 * 1024 + 14 + 140;
+pub const CDP_MAX_MESSAGE_SIZE = 100 * 1024 * 1024 + 14 + 140;
 
 // TCP keepalive parameters applied to accepted CDP connections.
 // Detection window ≈ IDLE + CNT * INTVL = 4 + 3*2 = 10s.
--- a/src/Server.zig
+++ b/src/Server.zig
@@ -427,8 +427,8 @@ test "Client: read invalid websocket message" {
         );
     }
 
-    // length of message is 0000 0810, i.e: 1024 * 512 + 265
-    try assertWebSocketError(1009, &.{ 129, 255, 0, 0, 0, 0, 0, 8, 1, 0, 'm', 'a', 's', 'k' });
+    // declared payload length is 0x0800 0000, i.e. 128MB — above CDP_MAX_MESSAGE_SIZE
+    try assertWebSocketError(1009, &.{ 129, 255, 0, 0, 0, 0, 8, 0, 0, 0, 'm', 'a', 's', 'k' });
 
     // continuation type message must come after a normal message
     // even when not a fin frame
@@ -507,6 +507,45 @@ test "Client: close message" {
         // fin | close, masked | len, 4-byte mask
         &.{ 136, 128, 0, 0, 0, 0 },
     );
+}
+
+test "Client: large websocket message" {
+    // A message bigger than the historical 512KB ceiling must be framed
+    // and dispatched, not answered with a 1009 close. CDP clients
+    // routinely send Runtime.evaluate / Page.addScriptToEvaluateOnNewDocument
+    // payloads of this size (e.g. injecting a bundled JS library).
+    const allocator = testing.allocator;
+
+    const prefix = "{\"id\":9,\"method\":\"Browser.getVersion\",\"pad\":\"";
+    const suffix = "\"}";
+    const payload_len = 600 * 1024;
+
+    const payload = try allocator.alloc(u8, payload_len);
+    defer allocator.free(payload);
+    @memcpy(payload[0..prefix.len], prefix);
+    @memset(payload[prefix.len .. payload_len - suffix.len], 'a');
+    @memcpy(payload[payload_len - suffix.len ..], suffix);
+
+    // fin | text, masked | 127 (8-byte extended length), big-endian
+    // length, all-zero mask (XOR no-op)
+    var header = [_]u8{ 129, 255, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 };
+    std.mem.writeInt(u64, header[2..10], payload_len, .big);
+
+    var c = try createTestClient();
+    defer c.deinit();
+
+    try c.handshake();
+    try c.stream.writeAll(&header);
+    try c.stream.writeAll(payload);
+
+    const msg = try c.readWebsocketMessage() orelse return error.NoMessage;
+    defer if (msg.cleanup_fragment) {
+        c.reader.cleanup();
+    };
+
+    try testing.expectEqual(.text, msg.type);
+    try testing.expect(std.mem.indexOf(u8, msg.data, "\"id\":9") != null);
+    try testing.expect(std.mem.indexOf(u8, msg.data, "protocolVersion") != null);
 }
 
 test "server: 404" {


karlseguin:
I can see 512KB being too small. @krichprollsch this raise any concern for you?

krichprollsch:
from 512kb to 100MB sounds like a big jump :/
I'm not sure about that, can we start w/ 10MB only?
@navidemad did you have the issue or is it purely about reducing the diff w/ chrome?

karlseguin:
Closed in favor of https://github.com/lightpanda-io/browser/pull/2760

PR #2719, opened 2026-06-12
css: apply @layer block rules to the cascade

### LINKED ISSUE #2718: css: @layer block rules are dropped from the cascade (checkVisibility / getComputedStyle report Tailwind v4 hidden elements as visible)
## Summary

Style rules wrapped in a CSS Cascade Layers block (`@layer name { … }` or anonymous `@layer { … }`, [css-cascade-5 §6](https://www.w3.org/TR/css-cascade-5/#layering)) never participate in the cascade. `StyleManager.parseSheet` applies top-level `.style` rules and recurses into `@media`, but `@layer` falls into the silent `else => {}` arm — so every declaration inside the block is dropped. A `display: none` rule inside `@layer` leaves the element reporting `checkVisibility() === true` and `getComputedStyle(el).display === "block"`, where Chrome reports hidden.

This bites any page built with Tailwind CSS v4, which emits its **entire** compiled stylesheet inside `@layer` blocks (`@layer theme, base, components, utilities;` + one block per layer). Every utility — including `.hidden { display: none }` — silently no-ops, so closed dropdown menus, collapsed panels, etc. all report visible. Since `checkVisibility()` and `getComputedStyle()` both consume this cascade, the gap affects agentic/scraping visibility filtering, not just styling.

## Today's behavior

A CDP client navigating to a page whose stylesheet contains `@layer utilities { .layer-hidden { display: none } }` and evaluating `document.querySelector('.layer-hidden').checkVisibility()` via `Runtime.evaluate` gets `true`. The identical rule written at the top level (no layer) correctly yields `false`. The statement form (`@layer a, b;`) parses harmlessly — only block contents are lost.

```mermaid
sequenceDiagram
    participant LP as Lightpanda
    Client->>LP: Page.navigate
    LP->>SM: parseSheet walks stylesheet
    SM->>SM: style rule added to cascade
    SM->>SM: media block recursed into
    SM--xSM: layer block hits else arm and is dropped
    Client->>LP: Runtime.evaluate checkVisibility
    LP-->>Client: true — element wrongly visible
    Note over SM: src/browser/StyleManager.zig parseSheet
```

## Expected behavior

Per [css-cascade-5 §6.1](https://www.w3.org/TR/css-cascade-5/#layer-block): "block at-rules inside of `@layer` … specify styles in the cascade" — the inner rules must apply. Chrome hides the element (`checkVisibility() === false`, computed `display === "none"`).

```mermaid
sequenceDiagram
    participant LP as Lightpanda
    Client->>LP: Page.navigate
    LP->>SM: parseSheet walks stylesheet
    SM->>SM: style rule added to cascade
    SM->>SM: media block recursed into
    SM->>SM: layer block flattened into cascade
    Client->>LP: Runtime.evaluate checkVisibility
    LP-->>Client: false — element hidden as in Chrome
```

Suggested v1 scope (matching how `@media` is handled today): treat `@layer` blocks as transparent wrappers — parse the inner rules into the cascade with source-order tie-breaking, recursing into nested `@media`/`@layer` both ways. Full layer-priority ordering (css-cascade-5 §6.4, where earlier-declared layers lose to later ones regardless of specificity) can follow separately; flattening already fixes the common failure mode (rules vanishing entirely).

On the CSSOM side, `cssRules`-inserted `@layer` rules (via `insertRule`/`replaceSync`) currently map to the `.unknown` rule type; they need their own type so the cascade pass can recognize them. Their JS-visible `type` should stay `0` — `@layer` postdates the legacy numeric constants ([CSSOM §6.4.1](https://drafts.csswg.org/cssom/#dom-cssrule-type), matching Chrome's `CSSLayerBlockRule`).

## Reproducer

`repro.html`:
```html
<!DOCTYPE html>
<html>
<head>
  <title>layer cascade repro</title>
  <style>
    .plain-hidden { display: none; }
    @layer utilities {
      .layer-hidden { display: none; }
    }
  </style>
</head>
<body>
  <div class="plain-hidden">hidden by a top-level rule (control)</div>
  <div class="layer-hidden">hidden by the same rule inside an @layer block</div>
</body>
</html>
```

`repro.sh`:
```bash
#!/usr/bin/env bash
# Asserts that rules inside CSS `@layer` blocks participate in the cascade
# (checkVisibility / getComputedStyle), like Chrome.
#
# Run:           ./repro.sh
# Expected today (bug): ".layer-hidden checkVisibility(): true"   exit 1
# Expected after fix:   ".layer-hidden checkVisibility(): false"  exit 0

set -euo pipefail

CDP_PORT="${CDP_PORT:-9222}"

PIDS=()
cleanup() {
}
trap cleanup EXIT

cd "$REPRO_DIR"
PIDS+=("$!")

PIDS+=("$!")

for _ in $(seq 1 50); do
  sleep 0.1
done

fi

```

`probe.js`:
```js
// Asserts that a `display:none` rule wrapped in `@layer` participates in the
// cascade: `.layer-hidden` must report checkVisibility() === false, exactly
// like the unlayered `.plain-hidden` control rule does.

(async () => {
  const c = await connect();

  const plain = await c.eval(`document.querySelector('.plain-hidden').checkVisibility()`);
  const layered = await c.eval(`document.querySelector('.layer-hidden').checkVisibility()`);
  const display = await c.eval(`getComputedStyle(document.querySelector('.layer-hidden')).display`);

  console.log(`.plain-hidden checkVisibility(): ${plain} (expected false — control)`);
  console.log(`.layer-hidden checkVisibility(): ${layered} (expected false)`);
  console.log(`.layer-hidden computed display:  ${JSON.stringify(display)} (expected "none")`);

  c.close();
  if (plain !== false) {
    console.log('BROKEN HARNESS: the unlayered control rule did not apply.');
    process.exit(2);
  }
  if (layered !== false || display !== 'none') {
    console.log('FAIL: @layer block rules are dropped from the cascade.');
    process.exit(1);
  }
  console.log('PASS: @layer block rules participate in the cascade.');
  process.exit(0);
```

<details>

```js
// than this is long).

  let nextId = 1;
  const pending = new Map();
  const handlers = new Map();
    pending.clear();
  };
  ws.on('message', (raw) => {
      pending.delete(m.id);
      else resolve(m.result);
    } else if (m.method) {
    }
  });
  return {
      const id = nextId++;
      });
    },
    ready() {
        ws.on('open', resolve);
      });
    },
    close() { ws.close(); },
  };
}

  await client.ready();

  return {
    sessionId,
    close: () => client.close(),
    async eval(expression) {
      }, sessionId);
    },
    async navigate(url2) {
      let loaded = false;
        try {
          }, sessionId);
        } catch (e) {
        }
      }
    },
  };
}

module.exports = { connect };
```

</details>

**Run**: `bash repro.sh`

Today (nightly `1.0.0-nightly.6703+b0426bab`):
```
.plain-hidden checkVisibility(): false (expected false — control)
.layer-hidden checkVisibility(): true (expected false)
.layer-hidden computed display:  "block" (expected "none")
FAIL: @layer block rules are dropped from the cascade.
```
exit 1.

Expected (verified against a local build with the fix):
```
.plain-hidden checkVisibility(): false (expected false — control)
.layer-hidden checkVisibility(): false (expected false)
.layer-hidden computed display:  "none" (expected "none")
PASS: @layer block rules participate in the cascade.
```
exit 0.

## Likely fix location

`src/browser/StyleManager.zig` — `parseSheet` (both the `_css_rules` path and the `<style>` text path) plus the nested at-rule recursion shared with `applyMediaAtRule`; and `src/browser/webapi/css/CSSStyleSheet.zig` — `atRuleTypeFor` so `cssRules`-inserted layer rules are recognizable.

## Environment

- Lightpanda: nightly `1.0.0-nightly.6703+b0426bab` (also confirmed in source at `main` `e7e605b6`)
- OS: macOS aarch64

navidemad:
## What

Style rules wrapped in CSS Cascade Layers blocks (`@layer name { … }`, anonymous `@layer { … }`, dotted sub-layer names like `@layer theme.dark { … }`) never participated in the cascade. A `display: none` inside `@layer` left the element reporting `checkVisibility() === true` and computed `display === "block"`, where Chrome reports it hidden. This PR flattens `@layer` block contents into the cascade the same way a matching `@media` body is applied today, on both stylesheet paths (`<style>` text and `cssRules` via `insertRule`/`replaceSync`), recursing into nested `@media`/`@layer` in both directions. The statement form (`@layer a, b;`) declares ordering only and stays inert.

Closes #2718.

## Root cause

`StyleManager.parseSheet` handled `.style` rules and recursed into `@media` only. `@layer` fell into the silent `else => {}` arm on the `_css_rules` path, and was skipped by the keyword check on the `<style>` text path. On the CSSOM side, `atRuleTypeFor` had no `layer` mapping, so `cssRules`-inserted layer rules degraded to `.unknown` and were unrecognizable to the cascade pass even if it had wanted to recurse.

```mermaid
flowchart LR
    A[stylesheet parsed] --> B[style rules added]
    A --> C[media blocks recursed]
    A --> D[layer blocks hit else arm]
    D --> E[layered rules vanish from cascade]
    style D fill:#fdd,color:#000
    style E fill:#fdd,color:#000
```

## Fix

- `src/browser/StyleManager.zig`: in `parseSheet`, dispatch `@layer` on both paths. A new `applyLayerAtRule` flattens the block body (the statement form has no block, so it's a no-op). Two helpers are now shared with the existing `@media` path: `applyInnerRules` (the nested-rule loop) and `atRuleBlock` (the comment-aware block-boundary scan), so `@media`-in-`@layer` and `@layer`-in-`@media` both work under the shared depth cap (`MAX_MEDIA_NESTING` generalized to `MAX_AT_RULE_NESTING`). The mutually-recursive group carries an explicit `Allocator.Error` error set, since inferred sets can't resolve the cycle.
- `src/browser/webapi/css/CSSRule.zig`: new `.layer` variant in `Type`; `getType()` returns `0` for it ([CSSOM §6.4.1](https://drafts.csswg.org/cssom/#dom-cssrule-type): `@layer` postdates the legacy numeric constants, matching Chrome's `CSSLayerBlockRule`). The variant sits after the last legacy-numbered type, so no existing `rule.type` value shifts.
- `src/browser/webapi/css/CSSStyleSheet.zig`: `atRuleTypeFor` maps `layer` to `.layer`.

```mermaid
flowchart LR
    A[stylesheet parsed] --> B[style rules added]
    A --> C[media blocks recursed]
    A --> D[layer blocks flattened]
    D --> E[layered rules join cascade]
    style D fill:#dfd,color:#000
    style E fill:#dfd,color:#000
```

### Deliberate scope

Layers are flattened. Full layer-priority ordering ([css-cascade-5 §6.4](https://www.w3.org/TR/css-cascade-5/#layer-ordering), where earlier-declared layers lose to later ones regardless of specificity) is **not** implemented: ties keep breaking on specificity + document order, and the layer-name prelude is not retained. The gap this leaves is narrow. `StyleManager`'s cascade only tracks the visibility filter (`display`, `visibility`, `opacity`, `pointer-events`), so a missing inter-layer priority can only matter when two of *those* declarations collide across layers with inverted specificity, and it never touches `getComputedStyle` for any other property. Flattening already fixes the common failure mode, layered rules vanishing entirely (Tailwind v4, for instance, emits its whole compiled output inside `@layer`). Proper inter-layer ordering can follow separately if a real page needs it.

## Test

- **Unit test**: `src/browser/webapi/css/CSSStyleSheet.zig` → `test "WebApi: layer @-rule cascade"` runs the new fixture `src/browser/tests/css/layer_at_rule_cascade.html` (named/anonymous/dotted blocks, statement form, `@media`-in-`@layer` matching and non-matching, `@layer`-in-`@media`, `@layer`-in-`@layer`, `!important`, `visibility: hidden`, `insertRule`/`replaceSync` paths, `rule.type === 0`, 50-deep nesting cap). Every assertion targets a rule that only reaches the cascade once `@layer` is flattened, so the fixture is meaningless without the fix.
- **Suite**: `zig build test` on this branch passes 978/979, including the new test. The single failure (`WebApi: Frames`, an iframe-over-HTTP test) is pre-existing on `main` — clean `main` is 977/978 with the same failure — and isn't touched by this PR, which only changes CSS files.
- **Reproducer**: issue #2718 carries a standalone CDP repro (one HTML fixture plus a small Node driver, no Ruby) that shows the same gap at the protocol level: `checkVisibility()` returns `true` for a `display:none` rule inside `@layer utilities`, where the identical unlayered rule returns `false`.

## Notes

`src/browser/tests/css/media_at_rule_cascade.html` has a one-line comment touch-up (`MAX_MEDIA_NESTING` → `MAX_AT_RULE_NESTING`) tracking the constant rename.


navidemad:
Bumping this since it's been quiet for a few weeks. The impact keeps getting more common: Tailwind v4 emits most of its utilities inside `@layer`, so on any Tailwind-v4 page an element hidden via a layered `display: none` still reports `checkVisibility() === true`, and automation clients end up acting on elements the page actually hides.

The branch is still conflict-free against main. Happy to rebase onto the recent StyleManager/computed-style work (#2836, #2864) and re-verify, or rework the approach if you'd prefer a different shape.

arrufat (approved):
Hi @navidemad, sorry for letting this sit so long, and thanks for the bump.

I reviewed it and verified against today's main. The branch is ~490 commits behind by now, but it merges cleanly, the new fixture passes as-is (including the deep nesting cap), and the full suite is green on the merged tree (1068/1068 with leak detection on). The pre-existing `WebApi: Frames` failure you mentioned has since been fixed on main, so everything is clean.

The approach looks right: `@layer` goes through the same flattening path as `@media`, with the shared helpers handling nesting in both directions under one depth cap. Flattening without layer-priority ordering is a reasonable scope, since StyleManager only tracks the visibility filter the remaining gap is much rarer than the current behavior, where layered rules vanish entirely and every Tailwind v4 page misreports visibility. Matching Chrome with `type === 0` is correct too.

LGTM, thanks for the contribution.

karlseguin:
Will look at it tomorrow

PR #2722, opened 2026-06-12
feat(cdp): implement Browser.setDownloadBehavior file downloads

### LINKED ISSUE #2701: Browser.setDownloadBehavior is a noop: file downloads not implemented (no disk write, no download events)
## Summary

`Browser.setDownloadBehavior` is dispatched but its implementation is a noop — every parameter is commented out and it returns a bare success. As a result Lightpanda has no file-download path at all: a response carrying `Content-Disposition: attachment` is treated as a normal navigation (and for a non-displayable MIME its body bytes are discarded), the configured `downloadPath` is ignored, nothing is written to disk, and neither `Page.downloadWillBegin` nor `Browser.downloadProgress` is ever emitted. A CDP client therefore cannot trigger, await, or locate a download.

## Today's behavior

`setDownloadBehavior` at `src/cdp/domains/browser.zig:73-82` parses none of its params and just calls `sendResult(null, …)`. On navigation, `Frame.frameHeaderDoneCallback` (`src/browser/Frame.zig:1045`) commits the response as a page regardless of `Content-Disposition`; `frameDataCallback`/`frameDoneCallback` then route a non-HTML/-text/-image body into `_parse_state = .raw` (`Frame.zig:1171`) whose bytes are appended (`:1193`) and then **discarded** when an empty document is parsed in their place (`Frame.zig:1259-1264`). There is no `download_*` notification type and no `Page.downloadWillBegin` / `Browser.downloadProgress` emission anywhere in `src/`.

```mermaid
sequenceDiagram
    participant LP as Lightpanda
    participant Frame
    Client->>LP: Browser.setDownloadBehavior allow, downloadPath
    LP-->>Client: success but params are ignored
    Client->>LP: Page.navigate to attachment URL
    LP->>Frame: response with Content-Disposition attachment
    Frame->>Frame: treat as page load, buffer then discard body
    Frame-->>LP: empty document committed
    LP--xClient: no Page.downloadWillBegin
    LP--xClient: no Browser.downloadProgress
    Note over LP,Frame: nothing written to downloadPath
```

## Expected behavior

Per the CDP protocol, opting in via `Browser.setDownloadBehavior` should make a download response land on disk and surface progress events:

- [`Browser.setDownloadBehavior`](https://chromedevtools.github.io/devtools-protocol/tot/Browser/#method-setDownloadBehavior) — honor `behavior` (`allow` / `allowAndName` / `deny` / `default`), write downloaded files under `downloadPath`, and (when `eventsEnabled: true`) emit download events.
- [`Page.downloadWillBegin`](https://chromedevtools.github.io/devtools-protocol/tot/Page/#event-downloadWillBegin) — `{ frameId, guid, url, suggestedFilename }`, fired when a download starts.
- [`Browser.downloadProgress`](https://chromedevtools.github.io/devtools-protocol/tot/Browser/#event-downloadProgress) — `{ guid, totalBytes, receivedBytes, state }` where `state` advances `inProgress` → `completed`.

A response is a download when it carries `Content-Disposition: attachment` (and, in Chrome, when its MIME type is one the browser will not render). Clicking an `<a download>` anchor is the same operation initiated from script. No rendering or layout is required for any of this.

```mermaid
sequenceDiagram
    participant LP as Lightpanda
    participant Frame
    participant Disk
    Client->>LP: Browser.setDownloadBehavior allow, downloadPath
    LP->>LP: store behavior and downloadPath on the connection
    LP-->>Client: success
    Client->>LP: Page.navigate to attachment URL
    LP->>Frame: response with Content-Disposition attachment
    LP-->>Client: Page.downloadWillBegin guid, suggestedFilename
    Frame->>Disk: write body bytes under downloadPath
    LP-->>Client: Browser.downloadProgress state inProgress
    LP-->>Client: Browser.downloadProgress state completed
```

## Reproducer

Four files in one directory. No Chrome, no framework. Requires `node >= 18`.

```bash
npm install --no-save ws     # the only dependency
bash repro.sh                # or: LIGHTPANDA_BIN=/path/to/lightpanda bash repro.sh
```

`server.js` — serves the attachment:

```js
const http = require('http');

const PORT = process.env.HTTP_PORT || 9580;
const BODY = 'col1,col2\nhello,world\n42,1337\n';

  if (req.url === '/report.csv') {
    res.writeHead(200, {
      'Content-Type': 'text/csv',
      'Content-Disposition': 'attachment; filename="report.csv"',
      'Content-Length': Buffer.byteLength(BODY),
    });
    res.end(BODY);
    return;
  }
  res.writeHead(404, { 'Content-Type': 'text/plain' });
  res.end('not found');
});

server.listen(PORT, '127.0.0.1', () => console.error(`http server on 127.0.0.1:${PORT}`));
```

`probe.js` — set download behavior, navigate at the attachment, check four signals:

```js
const fs = require('fs');
const path = require('path');
const os = require('os');
const { connect, sleep } = require('./cdp.js');

const EXPECTED = 'col1,col2\nhello,world\n42,1337\n';

(async () => {
  const dlDir = fs.mkdtempSync(path.join(os.tmpdir(), 'lp-dl-'));
  const c = await connect();

  const events = [];
  c.on('Page.downloadWillBegin', (p) => events.push(['downloadWillBegin', p]));
  c.on('Browser.downloadProgress', (p) => events.push(['downloadProgress', p]));

  await c.send('Browser.setDownloadBehavior', {
    behavior: 'allow',
    downloadPath: dlDir,
    eventsEnabled: true,
  });

  await c.send('Page.navigate', { url: `http://127.0.0.1:${HTTP_PORT}/report.csv` });

  let file = null;
  for (let i = 0; i < 20; i++) {
    const completed = events.some(([n, p]) => n === 'downloadProgress' && p && p.state === 'completed');
    const files = fs.readdirSync(dlDir).filter((f) => !f.startsWith('.') && !f.endsWith('.crdownload'));
    if (files.length) file = path.join(dlDir, files[0]);
    if (completed && file) break;
    await sleep(100);
  }

  c.close();

  const beginFired = events.some(([n]) => n === 'downloadWillBegin');
  const completedFired = events.some(([n, p]) => n === 'downloadProgress' && p && p.state === 'completed');
  const fileExists = !!file && fs.existsSync(file);
  const contentOk = fileExists && fs.readFileSync(file, 'utf8') === EXPECTED;

  console.log(`Page.downloadWillBegin fired:               ${beginFired}`);
  console.log(`Browser.downloadProgress {state:completed}: ${completedFired}`);
  console.log(`file written to downloadPath:               ${fileExists}${file ? ' (' + path.basename(file) + ')' : ''}`);
  console.log(`file content matches response body:         ${contentOk}`);

  try { fs.rmSync(dlDir, { recursive: true, force: true }); } catch {}

  const ok = beginFired && completedFired && fileExists && contentOk;
  console.log(ok
    ? 'PASS: setDownloadBehavior wrote the attachment to disk and emitted download events.'
    : 'FAIL: attachment response produced no download — no events fired, nothing written to disk.');
  process.exit(ok ? 0 : 1);
})().catch((e) => { console.error(e); process.exit(2); });
```

`repro.sh` — orchestration:

```bash
#!/usr/bin/env bash
set -euo pipefail

CDP_PORT="${CDP_PORT:-9222}"
export HTTP_PORT

lsof -ti:"$CDP_PORT"  2>/dev/null | xargs kill 2>/dev/null || true
lsof -ti:"$HTTP_PORT" 2>/dev/null | xargs kill 2>/dev/null || true

PIDS=()
trap cleanup EXIT

cd "$REPRO_DIR"
if [ ! -d node_modules/ws ]; then npm install --no-save --silent ws; fi

node server.js >server.log 2>&1 &
PIDS+=("$!")

PIDS+=("$!")

for _ in $(seq 1 50); do
  sleep 0.1
done

exec node probe.js
```

<details>
<summary><code>cdp.js</code> — raw-<code>ws</code> CDP helper (Target.createTarget + attachToTarget boilerplate)</summary>

```js

  let nextId = 1;
  const pending = new Map();
  const handlers = new Map();
    pending.clear();
  };
  ws.on('message', (raw) => {
      pending.delete(m.id);
      else resolve(m.result);
    } else if (m.method) {
    }
  });
  return {
      const id = nextId++;
      });
    },
    close() { ws.close(); },
  };
}

  const startUrl = opts.startUrl || 'about:blank';
  await client.ready();
  const { targetId } = await client.send('Target.createTarget', { url: startUrl });

  return {
    sessionId,
    targetId,
    raw: client,
    send: (method, params = {}) => client.send(method, params, sessionId),
    on: (method, fn) => client.on(method, fn),
    close: () => client.close(),
    async eval(expression, { awaitPromise = false } = {}) {
      const send = () => client.send('Runtime.evaluate', {
        expression, returnByValue: true, awaitPromise,
      }, sessionId);
      let r;
      for (let attempt = 0; attempt < 3; attempt++) {
        try { r = await send(); break; }
        catch (e) {
          if (attempt === 2) throw e;
          await sleep(150);
        }
      }
    },
    async navigate(url) {
      let loaded = false;
      await client.send('Page.navigate', { url }, sessionId);
        if (loaded) break;
        await sleep(100);
        try {
          }, sessionId);
        } catch (e) {
        }
      }
      await this.eval('1');
    },
  };
}

function sleep(ms) { return new Promise(r => setTimeout(r, ms)); }

module.exports = { connect, makeClient, sleep };
```

</details>

### Run output today

```
Page.downloadWillBegin fired:               false
Browser.downloadProgress {state:completed}: false
file written to downloadPath:               false
file content matches response body:         false
FAIL: attachment response produced no download — no events fired, nothing written to disk.
```

Exit code `1`. After the fix, all four lines read `true`, it prints `PASS`, and exits `0`. Reproduced identically against the current public nightly and a local build of `main` (`d695ce10`).

## Likely fix location

- `src/cdp/domains/browser.zig` — `setDownloadBehavior`: parse and store `behavior` / `downloadPath` / `eventsEnabled`.
- `src/cdp/CDP.zig` (`BrowserContext`) — hold the per-connection download policy.
- `src/browser/Frame.zig` — `frameHeaderDoneCallback` (recognize `Content-Disposition: attachment`), `frameDataCallback` / `frameDoneCallback` (stream the body to disk instead of discarding the `.raw` buffer), and `handleClick` (the `a.download` branch at `Frame.zig:4099`, which currently only logs a warning).
- `src/Notification.zig` — add `download_will_begin` / `download_progress` event types.
- `src/cdp/domains/page.zig` + `src/cdp/domains/browser.zig` — subscribe and emit `Page.downloadWillBegin` / `Browser.downloadProgress`.

## Proposed approach

Smallest correct slice:

1. **Store the policy.** Parse `behavior` / `downloadPath` / `eventsEnabled` in `setDownloadBehavior` and stash them on the per-connection `BrowserContext`.
2. **Detect the download.** In the navigation response path, a response whose `Content-Disposition` is `attachment` is a download rather than a document. (Optionally also a non-displayable MIME with no explicit disposition, matching Chrome.)
3. **Divert the body to disk.** Stream the bytes to `<downloadPath>/<suggestedFilename>` — filename from the `Content-Disposition` `filename=` parameter, else the URL basename — instead of buffering-then-discarding them as the `.raw` path does today.
4. **Emit the events.** Add the two notification types and have the CDP layer emit `Page.downloadWillBegin` and a terminal `Browser.downloadProgress { state: "completed" }`.

The `<a download>` click trigger (`Frame.handleClick`) reuses the same machinery and `behavior: deny` could land as a follow-up, keeping the first PR focused on the `Content-Disposition` navigation path.

> [!IMPORTANT]
> **One design question before I open a PR.** In Chrome, navigating to an `attachment` response does **not** change the current document — the page stays put and the download runs in the background. Lightpanda commits the pending page in `frameHeaderDoneCallback` as soon as headers arrive, and the `Content-Disposition` header is available there, so the navigation *can* be recognized as a download before commit. Which behavior do you prefer?
>
> - **(A) Preserve the current page** (Chrome-faithful): detect the attachment in the header callback, skip the page commit, keep the previous document live, run the download as a side effect. More correct; touches the pending/active commit path.
> - **(B) Let the navigation commit to a blank document** and run the download as a side effect. Simpler and contained to the response-body path; since there is no rendering the post-download page state is rarely observed, but it diverges from Chrome's "page unchanged" semantics.
>
> I'm happy to implement either — (A) is more faithful, (B) is a smaller, lower-risk first PR. Glad to follow your preference before writing the navigation-lifecycle code.

## Environment

- Lightpanda: current public nightly + local build of `main` at `d695ce10` (both reproduce identically)
- OS: macOS (aarch64); the download path is platform-independent
- CDP client: Node 18 + `ws` (raw WebSocket; see `cdp.js` above)

Ar-maan05:
Fixes #2701

## Summary

`Browser.setDownloadBehavior` was a noop (every param commented out, bare success returned), so Lightpanda had no file-download path at all: a response carrying `Content-Disposition: attachment` was treated as a normal navigation,
its body discarded, `downloadPath` ignored, and no download events emitted.

This wires up the full path. When `Browser.setDownloadBehavior` opts in (`allow` / `allowAndName`) and a navigation response carries `Content-Disposition: attachment`, the body is streamed to disk under `downloadPath` instead of being parsed as a page, and, when `eventsEnabled: true`, `Page.downloadWillBegin` and `Browser.downloadProgress` (`inProgress` -> `completed`) are emitted.

## Approach

This follows the existing notification pattern: `Frame` dispatches notifications and the CDP domains consume them and emit CDP events (the same shape as `frame_navigated` -> `Page.frameNavigated`).

- **`Session.zig`**: download config (`download_behavior`, `download_path`, `download_events_enabled`), scoped per CDP connection.
- **`Notification.zig`**: new `DownloadWillBegin` / `DownloadProgress` event
- **`Frame.zig`**: a `.download` parse state that holds the open file; detect the attachment in `frameHeaderDoneCallback`, stream the body to disk in `frameDataCallback`, close + emit `completed` in `frameDoneCallback`, then commit an empty document so the navigation still finishes cleanly (mirrors
  the existing `.raw` path). Content-Disposition parsing handles the quoted, unquoted, and RFC 5987 (`filename*=`) forms, and strips path components to guard against traversal.
- **`Runner.zig`**: `.download` handled as a loading state.
- **`CDP.zig`**: idempotent registration for the download notifications.
- **`browser.zig` / `page.zig`**: `setDownloadBehavior` now parses and stores
  its params and toggles event registration; `Browser.downloadProgress` and `Page.downloadWillBegin` emitters.

## Behavior notes / scope

- **Navigation-initiated downloads only.** Script-initiated `<a download>` clicks are the same operation from JS but go through a different entry point; left as a follow-up.
- **`deny`** falls through to the prior behavior (the response is not written to disk); no cancellation event is emitted.
- RFC 5987 `filename*` values are taken verbatim (no percent-decoding yet).

## Testing

- New end-to-end CDP test: configures `setDownloadBehavior` with a temp `downloadPath`, navigates to a `Content-Disposition: attachment` response, and asserts both `Page.downloadWillBegin` and `Browser.downloadProgress` fire and that the file lands on disk with the correct bytes.
- New tests for config storage, registration toggling, and rejection of an unknown `behavior`.
- New unit tests for the Content-Disposition / filename parsing helpers.
- Full suite green (`make test`), formatting clean (`zig fmt --check`).

## Checklist

- [x] Tests pass (`make test`).
- [x] Formatting is clean (`zig fmt --check ./*.zig ./**/*.zig`).
- [x] CLA signed.

### DIFF
--- a/src/browser/Frame.zig
+++ b/src/browser/Frame.zig
@@ -1198,7 +1198,165 @@ fn frameHeaderDoneCallback(response: HttpClient.Response) !HttpClient.HeaderResu
         });
     }
 
+    // If the response is a file download, stream its body to disk instead of
+    // parsing it as a page. This sets _parse_state to .download, which the
+    // data/done callbacks below special-case.
+    _ = try self.maybeStartDownload(response);
+
     return .proceed;
+}
+
+// Returns true when the response was set up as a file download. A response is
+// treated as a download when Browser.setDownloadBehavior opted in
+// (allow/allowAndName) and the response carries Content-Disposition: attachment.
+// See issue #2701.
+fn maybeStartDownload(self: *Frame, response: HttpClient.Response) !bool {
+    const session = self._session;
+    switch (session.download_behavior) {
+        .allow, .allow_and_name => {},
+        .default, .deny => return false,
+    }
+
+    const disposition = blk: {
+        var it = response.headerIterator();
+        while (it.next()) |hdr| {
+            if (std.ascii.eqlIgnoreCase(hdr.name, "content-disposition")) {
+                break :blk hdr.value;
+            }
+        }
+        return false;
+    };
+    if (isAttachment(disposition) == false) {
+        return false;
+    }
+
+    const download_path = session.download_path orelse {
+        log.warn(.frame, "download without downloadPath", .{ .url = self.url });
+        return false;
+    };
+
+    var guid_buf: [36]u8 = undefined;
+    @import("../id.zig").uuidv4(&guid_buf);
+    const guid = try self.arena.dupe(u8, &guid_buf);
+
+    const suggested = dispositionFilename(disposition) orelse urlBasename(self.url) orelse guid;
+    const suggested_filename = try self.arena.dupe(u8, suggested);
+
+    // allowAndName stores the file under its guid; allow uses the suggested name.
+    const on_disk_name = switch (session.download_behavior) {
+        .allow_and_name => guid,
+        else => suggested_filename,
+    };
+
+    std.fs.cwd().makePath(download_path) catch |err| {
+        log.err(.frame, "download makePath", .{ .err = err, .path = download_path });
+        return false;
+    };
+    var dir = std.fs.cwd().openDir(download_path, .{}) catch |err| {
+        log.err(.frame, "download openDir", .{ .err = err, .path = download_path });
+        return false;
+    };
+    defer dir.close();
+    const file = dir.createFile(on_disk_name, .{ .truncate = true }) catch |err| {
+        log.err(.frame, "download createFile", .{ .err = err, .name = on_disk_name });
+        return false;
+    };
+
+    const total: ?u64 = if (response.contentLength()) |cl| cl else null;
+
+    self._parse_state = .{ .download = .{
+        .guid = guid,
+        .file = file,
+        .filename = suggested_filename,
+        .received = 0,
+        .total = total,
+    } };
+
+    if (session.download_events_enabled) {
+        session.notification.dispatch(.download_will_begin, &.{
+            .frame_id = self._frame_id,
+            .guid = guid,
+            .url = self.url,
+            .suggested_filename = suggested_filename,
+        });
+        session.notification.dispatch(.download_progress, &.{
+            .guid = guid,
+            .total_bytes = total orelse 0,
+            .received_bytes = 0,
+            .state = .in_progress,
+        });
+    }
+
+    return true;
+}
+
+// True when a Content-Disposition value's type token is "attachment".
+fn isAttachment(disposition: []const u8) bool {
+    const semi = std.mem.indexOfScalar(u8, disposition, ';') orelse disposition.len;
+    const token = std.mem.trim(u8, disposition[0..semi], " \t");
+    return std.ascii.eqlIgnoreCase(token, "attachment");
+}
+
+// Extracts the filename from a Content-Disposition value, handling the quoted,
+// unquoted, and RFC 5987 (filename*=charset''value) forms. Path components are
+// stripped so the result is always a bare basename.
+fn dispositionFilename(disposition: []const u8) ?[]const u8 {
+    // Prefer the extended filename*= form when present, per RFC 6266.
+    if (paramValue(disposition, "filename*")) |ext| {
+        // charset'lang'value — take everything after the second quote.
+        if (std.mem.indexOfScalar(u8, ext, '\'')) |first| {
+            if (std.mem.indexOfScalarPos(u8, ext, first + 1, '\'')) |second| {
+                return basename(ext[second + 1 ..]);
+            }
+        }
+        return basename(ext);
+    }
+    if (paramValue(disposition, "filename")) |name| {
+        return basename(name);
+    }
+    return null;
+}
+
+// Returns the (optionally quoted) value of `key=` within a parameter list.
+fn paramValue(disposition: []const u8, key: []const u8) ?[]const u8 {
+    var rest = disposition;
+    while (std.mem.indexOfScalar(u8, rest, ';')) |semi| {
+        const param = std.mem.trim(u8, rest[semi + 1 ..][0 .. (std.mem.indexOfScalarPos(u8, rest, semi + 1, ';') orelse rest.len) - semi - 1], " \t");
+        if (std.mem.indexOfScalar(u8, param, '=')) |eq| {
+            const name = std.mem.trim(u8, param[0..eq], " \t");
+            if (std.ascii.eqlIgnoreCase(name, key)) {
+                var value = std.mem.trim(u8, param[eq + 1 ..], " \t");
+                if (value.len >= 2 and value[0] == '"' and value[value.len - 1] == '"') {
+                    value = value[1 .. value.len - 1];
+                }
+                if (value.len > 0) {
+                    return value;
+                }
+            }
+        }
+        rest = rest[semi + 1 ..];
+    }
+    return null;
+}
+
+// Strips any directory components, guarding against path traversal.
+fn basename(name: []const u8) ?[]const u8 {
+    var out = name;
+    if (std.mem.lastIndexOfAny(u8, out, "/\\")) |i| {
+        out = out[i + 1 ..];
+    }
+    if (out.len == 0 or std.mem.eql(u8, out, ".") or std.mem.eql(u8, out, "..")) {
+        return null;
+    }
+    return out;
+}
+
+// Derives a filename from a URL's last path segment, ignoring query/fragment.
+fn urlBasename(url: []const u8) ?[]const u8 {
+    var path = url;
+    if (std.mem.indexOfScalar(u8, path, '?')) |i| path = path[0..i];
+    if (std.mem.indexOfScalar(u8, path, '#')) |i| path = path[0..i];
+    return basename(path);
 }
 
 fn frameDataCallback(response: HttpClient.Response, data: []const u8) !void {
@@ -1279,6 +1437,13 @@ fn frameDataCallback(response: HttpClient.Response, data: []const u8) !void {
             }
         },
         .raw, .image => |*buf| try buf.appendSlice(self.arena, data),
+        .download => |*download| {
+            download.file.writeAll(data) catch |err| {
+                log.err(.frame, "download write", .{ .err = err, .guid = download.guid });
+                return;
+            };
+            download.received += data.len;
+        },
         .pre => unreachable,
         .complete => unreachable,
         .err => unreachable,
@@ -1371,6 +1536,30 @@ fn frameDoneCallback(ctx: *anyopaque) !void {
 
             parser.parse(html);
             self._parse_state = .complete;
+            self.documentIsComplete();
+        },
+        .download => |*download| {
+            download.file.close();
+
+            // Capture before invalidating the union below.
+            const guid = download.guid;
+            const received = download.received;
+            const total = download.total orelse download.received;
+            self._parse_state = .complete;
+
+            const session = self._session;
+            if (session.download_events_enabled) {
+                session.notification.dispatch(.download_progress, &.{
+                    .guid = guid,
+                    .total_bytes = total,
+                    .received_bytes = received,
+                    .state = .completed,
+                });
+            }
+
+            // The body went to disk; commit an empty document so the frame
+            // navigation still completes cleanly (mirrors the .raw path).
+            parser.parse("<html><head><meta charset=\"utf-8\"></head><body></body></html>");
             self.documentIsComplete();
         },
         else => unreachable,
@@ -3793,13 +3982,32 @@ const ParseState = union(enum) {
     image: std.ArrayList(u8),
     raw: std.ArrayList(u8),
     raw_done: []const u8,
+    download: Download,
 
     fn deinit(self: *ParseState, frame: *Frame) void {
         switch (self.*) {
             .html => |html| frame.releaseArena(html.arena),
+            // Only reached when a frame is torn down mid-download (the normal
+            // completion path in frameDoneCallback already closes the file and
+            // transitions to .complete).
+            .download => |*download| download.file.close(),
             else => {},
         }
     }
+};
+
+// An in-flight file download (Content-Disposition: attachment under an
+// allow/allowAndName Browser.setDownloadBehavior). The response body is
+// streamed straight to `file` rather than parsed as a page. See issue #2701.
+const Download = struct {
+    // uuidv4, arena-owned. Matches the guid reported in the CDP events.
+    guid: []const u8,
+    file: std.fs.File,
+    // suggested filename surfaced to the client, arena-owned.
+    filename: []const u8,
+    received: u64,
+    // from Content-Length, when the response advertised one.
+    total: ?u64,
 };
 
 const LoadState = enum {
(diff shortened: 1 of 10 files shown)


Ar-maan05:
Hi, I have really enjoyed contributing to this repo so far. Always happy to help with other open issues if useful.

Ar-maan05:
I have rebased this branch to the latest `main`, let me know if any changes are needed.

krichprollsch:
@Ar-maan05 thanks for the PR, It seems that the tests are failing w/ playwright.
You can replay them locally via https://github.com/lightpanda-io/demo/
1. install deps with npm install
2. start lightpanda serve 
3. run the test suite `go run runner/main.go`

Ar-maan05:
I have fixed the merge conflicts and ensured all the tests pass. CI should be green now. Thank you for the review. Always happy to help with any other issues if useful.

src/browser/Frame.zig
```diff
@@ -1167,9 +1167,167 @@ fn frameHeaderDoneCallback(response: HttpClient.Response) !bool {
...
+    }
+
+    const download_path = session.download_path orelse {
+        log.warn(.frame, "download without downloadPath", .{ .url = self.url });
+        return false;
+    };
+
+    var guid_buf: [36]u8 = undefined;
+    @import("../id.zig").uuidv4(&guid_buf);
+    const guid = try self.arena.dupe(u8, &guid_buf);
```
krichprollsch, line 1243: I'm sorry but I don't get what's "guid" stands for 🤔 I read "group user id" but it's obviously wrong....

EDIT: found it from CDP docs: "Global unique identifier"

src/browser/Frame.zig
```diff
@@ -1167,9 +1167,167 @@ fn frameHeaderDoneCallback(response: HttpClient.Response) !bool {
...
+                }
+            }
+        }
+        rest = rest[semi + 1 ..];
+    }
+    return null;
+}
+
+// Strips any directory components, guarding against path traversal.
+fn basename(name: []const u8) ?[]const u8 {
```
krichprollsch: Can't we use `std.fs.path.basename` instead?

src/browser/Frame.zig
```diff
@@ -1167,9 +1167,167 @@ fn frameHeaderDoneCallback(response: HttpClient.Response) !bool {
...
+        out = out[i + 1 ..];
+    }
+    if (out.len == 0 or std.mem.eql(u8, out, ".") or std.mem.eql(u8, out, "..")) {
+        return null;
+    }
+    return out;
+}
+
+// Derives a filename from a URL's last path segment, ignoring query/fragment.
+fn urlBasename(url: []const u8) ?[]const u8 {
```
krichprollsch: I would use a real url parsing instead. see `webapi/URL.zig`.

src/browser/Frame.zig
```diff
@@ -1167,9 +1167,167 @@ fn frameHeaderDoneCallback(response: HttpClient.Response) !bool {
...
+        return basename(ext);
+    }
+    if (paramValue(disposition, "filename")) |name| {
+        return basename(name);
+    }
+    return null;
+}
+
+// Returns the (optionally quoted) value of `key=` within a parameter list.
+fn paramValue(disposition: []const u8, key: []const u8) ?[]const u8 {
```
krichprollsch: Is it specific to an HTTP header value isn't it?
Maybe we could move it into `Header` struct from `http.zig`.
Than you can use directly the `Header` you get from the response's header iterator.

krichprollsch: Maybe you could also implement a values iterator into `Header` struct, loop hover them and get the value you need directly: attachement, filename, ...

src/cdp/CDP.zig
```diff
@@ -797,6 +801,36 @@ pub const BrowserContext = struct {
...
+    }
+
+    pub fn onDownloadWillBegin(ctx: *anyopaque, msg: *const Notification.DownloadWillBegin) !void {
+        const self: *BrowserContext = @ptrCast(@alignCast(ctx));
+        return @import("domains/page.zig").downloadWillBegin(self, msg);
+    }
+
+    pub fn onDownloadProgress(ctx: *anyopaque, msg: *const Notification.DownloadProgress) !void {
+        const self: *BrowserContext = @ptrCast(@alignCast(ctx));
+        return @import("domains/browser.zig").downloadProgress(self, msg);
```
krichprollsch, line 831: I'm not sure I get why you send `Browser.downloadProgress` and `Page.downloadWillBegin` and not also `Browser.downloadWillBegin` + `Page.downloadProgress`?

krichprollsch: Afterall, since we set `Browser.setDownloadBehavior` I think we should only dispatch `Browser.XXX` events and not `Pages.XXX`

src/browser/Session.zig
```diff
@@ -105,6 +105,17 @@ _console_capture: bool = false,
...
 load_external_stylesheets: bool = false,
 
+// Download handling configured via the `Browser.setDownloadBehavior` CDP
+// method (see issue #2701). When `download_behavior` is `.allow` or
+// `.allow_and_name`, a navigation whose response carries
+// `Content-Disposition: attachment` is written to `download_path` instead
+// of being parsed as a page, and (when `download_events_enabled` is set)
+// `Page.downloadWillBegin` / `Browser.downloadProgress` events are emitted.
+// `download_path` is duped into the Session arena.
+download_behavior: DownloadBehavior = .default,
```
krichprollsch: I think `default` is useless, it should be removed and we should set directly `deny` instead.
And if the CDP command sends a `default`, then explicitly set the behavior as `deny`.

src/cdp/domains/browser.zig
```diff
@@ -69,18 +71,69 @@ fn getVersion(cmd: *CDP.Command) !void {
...
 fn setDownloadBehavior(cmd: *CDP.Command) !void {
-    // const params = (try cmd.params(struct {
-    //     behavior: []const u8,
-    //     browserContextId: ?[]const u8 = null,
-    //     downloadPath: ?[]const u8 = null,
-    //     eventsEnabled: ?bool = null,
-    // })) orelse return error.InvalidParams;
+    const params = (try cmd.params(struct {
+        behavior: []const u8,
+        browserContextId: ?[]const u8 = null,
```
krichprollsch, line 83: If `browserContextId` write a `not_implemented` log.

```zig
if (params.browserContextId != null) {
    log.warn(.not_implemented, "Browser.setDownloadBehavior", .{ .param = "browserContextId" });
}
```

src/browser/Frame.zig
```diff
@@ -1248,6 +1406,13 @@ fn frameDataCallback(response: HttpClient.Response, data: []const u8) !void {
             }
         },
         .raw, .image => |*buf| try buf.appendSlice(self.arena, data),
+        .download => |*download| {
+            download.file.writeAll(data) catch |err| {
+                log.err(.frame, "download write", .{ .err = err, .guid = download.guid });
+                return;
```
krichprollsch, line 1433: I think we should handle the error here more correctly here, but idk exactly how.
We can't set the `parse_state` to `err` directly b/c we could hit the `.err => unreachable` on next chunck.

So we should either:
* change that unreachable on error
* set a failure/error on download

And maybe remove the eventually partial file written?

I suggest to write a `TODO` here for a follow up instead.

src/browser/Frame.zig
Ar-maan05: Yep, "Global Unique Identifier" from the CDP download events. Added a comment at the definition so it's not cryptic.

src/browser/Frame.zig
Ar-maan05: Done. Replaced it with `std.fs.path.basename` (the helper is now `sanitizeFilename`). I kept a backslash strip on top of it because a Content-Disposition can carry Windows separators even on a Linux host, where the posix `basename` wouldn't split them, plus the ./../empty guard.

src/browser/Frame.zig
Ar-maan05: Switched to real URL parsing: it now goes through rust-url (`sys/url.zig url_parse` + `url_get_path`) and takes the basename of the parsed path, so query/fragment are handled there rather than by hand.

src/browser/Frame.zig
Ar-maan05: Moved this onto the `Header` struct in `network/http.zig`. It now exposes `firstValue()` (the value before the first ;), `param(key)`, and a `params()` `ParamIterator`. `dispositionFilename` takes the `Header` straight from the response header iterator and uses `.param("filename")` / `.param("filename*")`, so the Content-Disposition-specific string slicing is gone from `Frame`.

src/cdp/CDP.zig
Ar-maan05: Agreed. Since the opt-in is `Browser.setDownloadBehavior`, I dropped the `Page.*` events and now emit `Browser.downloadWillBegin` + `Browser.downloadProgress` only. The `downloadWillBegin` emitter moved from `page.zig` to `browser.zig` alongside `downloadProgress`.

src/browser/Session.zig
Ar-maan05: Removed the `.default` variant; the field now defaults to `.deny`, and a CDP `behavior: "default"` is explicitly mapped to `.deny`.

src/cdp/domains/browser.zig
Ar-maan05: Added the `not_implemented` log as suggested.

src/browser/Frame.zig
Ar-maan05: Left a `TODO` here for a follow-up, as you suggested. Noted the two open parts: it can't set `parse_state = .err` without tripping the `.err => unreachable` on the next chunk, and it should emit a `canceled` progress event and remove the partial file.

Ar-maan05:
@krichprollsch pushed a round addressing all the review comments. Full suite green (`make test`), `zig fmt --check` clean.

krichprollsch:
Thanks for the changes, LGMT. I'm preparing a demo/test script for the feature into https://github.com/lightpanda-io/demo/

Ar-maan05:
Sounds good! Let me know if I can be of any help with this or with any issues/feature implementations.

PR #2727, opened 2026-06-12
feat(cdp): honor Browser.grantPermissions / setPermission / resetPermissions

rohitsux:
`Browser.grantPermissions`, `Browser.setPermission` and `Browser.resetPermissions` were noops, so a CDP automation client (Puppeteer's `browserContext.grantPermissions(['geolocation'])`, Playwright's `context.grantPermissions`) could not influence what `navigator.permissions.query()` returns. The query always reported `"prompt"` regardless. This is a gap for automation flows that need to pre-grant or deny permissions (geolocation, notifications, clipboard, etc.) before a page runs.

## Changes

- `src/browser/Page.zig`: store permission state on the Page, keyed by permission name, with a small `setPermission` helper. Keys and values are duped into the Page arena, so the state lives for (and resets with) the Page.
- `src/cdp/domains/browser.zig`: implement the three methods. `grantPermissions` sets each listed permission to `"granted"`; `setPermission` sets a single permission to an explicit `"granted"`/`"denied"`/`"prompt"` state; `resetPermissions` clears the map.
- `src/browser/webapi/Permissions.zig`: `navigator.permissions.query()` now reads the stored state and falls back to `"prompt"` when unset (previously hardcoded `"prompt"`).

## Scope note

Permission state is stored on the active `Page`, so it resets on navigation. CDP scopes these methods to a browser context (state persists across navigations); persisting across navigations would mean storing on the CDP `BrowserContext` and re-applying on each new Page, which I left as a follow-up to keep this change small. The `origin` / `browserContextId` parameters are accepted and ignored for now.

## Verified

- `make test` passes (796/796, no leaks); new test `cdp.browser: grant/set/reset permissions reach navigator.permissions` drives all three methods through `processMessage` and asserts the stored state.
- `make build` succeeds.

### DIFF
--- a/src/cdp/domains/browser.zig
+++ b/src/cdp/domains/browser.zig
@@ -17,7 +17,11 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 const std = @import("std");
+const lp = @import("lightpanda");
 const CDP = @import("../CDP.zig");
+
+const log = lp.log;
+const PermissionState = @import("../../browser/webapi/Permissions.zig").State;
 
 // TODO: hard coded data
 const PROTOCOL_VERSION = "1.3";
@@ -97,7 +101,8 @@ fn setWindowBounds(cmd: *CDP.Command) !void {
 }
 
 // Grant the listed permissions so navigator.permissions.query() reports them
-// as "granted". State is stored on the active Page and resets on navigation.
+// as "granted". State is stored on the Browser, so it persists across page
+// navigations.
 fn grantPermissions(cmd: *CDP.Command) !void {
     const params = (try cmd.params(struct {
         permissions: []const []const u8,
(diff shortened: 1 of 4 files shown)


src/cdp/domains/browser.zig
```diff
@@ -96,19 +96,54 @@ fn setWindowBounds(cmd: *CDP.Command) !void {
...
+    const params = (try cmd.params(struct {
+        permissions: []const []const u8,
+        origin: ?[]const u8 = null,
+        browserContextId: ?[]const u8 = null,
+    })) orelse return error.InvalidParams;
+
+    const bc = cmd.browser_context orelse return error.BrowserContextNotLoaded;
+    const page = bc.session.currentPage() orelse return error.PageNotLoaded;
+    for (params.permissions) |name| {
+        try page.setPermission(name, .granted);
```
krichprollsch: Shouldn't we set permissions at the browser's level?
Here, any permission set via CDP will be lost with the page.

Also regarding the `origin` support, if we don't support it right now, we should at least write down a not_implemented log.
Same about `browserContextId`.

rohitsux:
Thanks for the review. Addressed all three points:

- Moved the permission store from the active \`Page\` to the \`Browser\` (mirroring how \`http_client\` holds the user-agent override), so permissions now persist across page navigations instead of being lost with the page. \`navigator.permissions.query()\` reads the browser-level state via \`frame._session.browser\`.
- Added \`not_implemented\` log warnings for \`origin\` and \`browserContextId\` on both \`grantPermissions\` and \`setPermission\`, since they are accepted but not yet honored.

The test now asserts the state on the browser. \`make test\` passes (796/796, no leaks), \`zig fmt --check\` is clean, and \`make build\` succeeds.

rohitsux:
Rebased the browser-level change on top of your `PermissionState` enum refactor (it had gotten clobbered in an earlier merge on my end). The browser store is now `StringHashMapUnmanaged(PermissionState)`, `setPermission` takes the enum, and `Browser.setPermission` from CDP parses the setting via `std.meta.stringToEnum`. Force-pushed. `make test` is green (796/796, no leaks).

PR #2737, opened 2026-06-13
script: make goto async and support top-level await

arrufat:
Wraps agent scripts in an async IIFE to allow top-level `await` and explicit `return` statements for output. The `goto` primitive is now asynchronous, returning a Promise that is settled by driving the browser event loop. All other primitives remain synchronous.

### DIFF
--- a/docs/agent-script.md
+++ b/docs/agent-script.md
@@ -1,11 +1,12 @@
 # Agent JavaScript scripts
 
 `lightpanda agent <script.js>` runs a JavaScript file that drives a
-Lightpanda browser session. Navigation goes through a global `goto(url)` that
-returns a **page object**; every other browser action is a method on that page
-(`page.extract(...)`, `page.click(...)`, …). The format is intentionally plain
-JavaScript: use normal variables, functions, loops, objects, arrays,
-`JSON.parse`, `JSON.stringify`, and other standard ECMAScript built-ins.
+Lightpanda browser session. Pages are created with the global `Page` class —
+`new Page()` makes a page and `await page.goto(url)` navigates it; every other
+browser action is a method on that page (`page.extract(...)`, `page.click(...)`,
+…). The format is intentionally plain JavaScript: use normal variables,
+functions, loops, objects, arrays, `JSON.parse`, `JSON.stringify`, and other
+standard ECMAScript built-ins.
 
 ```console
 ./lightpanda agent session.js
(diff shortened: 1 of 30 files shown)


docs/agent-script.md
```diff
@@ -32,18 +32,19 @@ web page's JavaScript context.
...
-  (`.then`) only run after the script body finishes, never in between
-  primitive calls.
-- Tool failures throw JavaScript `Error` exceptions and stop execution unless
-  you catch them.
-- The script's completion value — its last top-level expression — is printed
-  automatically (objects and arrays as JSON; other values coerced). End a
-  script with the bare expression you want as output, e.g. a final
-  `extract({ ... });` or `results;`. `console.log(...)` is for extra or debug
-  output and does not JSON-format objects.
+- `goto(url)` is asynchronous and resolves to a **page handle** — always
```
karlseguin:
```suggestion
- `goto(url)` is asynchronous and resolves to a **page handle** — always use
```

Maybe it's too early, but I didn't really get it at first.

docs/agent.md
```diff
@@ -353,27 +353,32 @@ session).
...
 click({ selector: "a.login" });
-evaluate("document.title");
+return evaluate("document.title");
 ```
 
-The primitives are **synchronous and blocking**: each returns its
-result directly, so write `const data = extract(…)`, not
-`await extract(…)`. (`evaluate(...)` can run async JS inside the page,
-but the `evaluate(...)` call itself still returns synchronously.)
+`goto(url)` is **asynchronous** and resolves to a **page handle** — always
```
karlseguin:
```suggestion
`goto(url)` is **asynchronous** and resolves to a **page handle** — always use
```

same as above, maybe I'm wrong and it's better for an agent as-is.

src/script/Runtime.zig
```diff
@@ -43,6 +44,13 @@ console_data: [std.enums.values(ConsoleMethod).len]ConsoleData,
 /// of colliding with the indicator; the line still goes to stdout/stderr.
 console_observer: ?ConsoleObserver = null,
 
+/// In-flight `goto` navigations the driver loop settles. Concurrent gotos open
+/// popup frames so they coexist; a sequential goto replaces the page.
+pending_gotos: std.ArrayListUnmanaged(PendingGoto) = .empty,
```
karlseguin:
```suggestion
pending_gotos: std.ArrayList(PendingGoto) = .empty,
```

src/script/Runtime.zig
```diff
@@ -273,19 +307,79 @@ pub fn runSource(self: *Runtime, source: []const u8, name: []const u8) RunError!
...
+        if (self.env.terminatePending()) return try self.dupeError("terminated");
+
+        // Nothing in flight → the script awaits something we can't settle.
+        if (self.pending_gotos.items.len == 0) break;
+
+        var next_tick_ms: u32 = 0;
+        {
+            self.session.browser.env.isolate.enter();
+            defer self.session.browser.env.isolate.exit();
+            if (runner == null) {
```
karlseguin: Is there a reason to do this lazily? Can this be called when there's no page (which is the only reason `session.runner()` can fail). If we assume it can get here and runner fails (there's no page) can't think possibly turn into an endless tight loop?

tl;dr, I think having this above the loop simplifies things:

```zig
var runner = try self.session.runner(.{});
```

karlseguin (commented):
It all looks ok. The only broad concern is the use of v8.* directly. If we ever do plan on supporting another JS engine, it's going to have to be moved to the js.* abstraction.

karlseguin:
I see that the resolved handle is the frame_id. So i assume that access is always done via `Session.findFrameByFrameId`. I was worried that it would be possible to reference a handle/frame after another navigation, but as long as it all goes through that `findFrameByFrameId` indirection, it should be ok.

PR #2740, opened 2026-06-13
fix(cdp): allow setting a browser User-Agent via CDP

### LINKED ISSUE #2704: CDP: cannot set a browser User-Agent — Mozilla UAs rejected by setUserAgentOverride and ignored by setExtraHTTPHeaders

Ar-maan05:
## What

Fixes #2704: a CDP client cannot set a realistic browser `User-Agent`. Both Playwright mechanisms are broken, so any UA beginning with `Mozilla/5.0` (i.e. every real Chrome/Firefox/Safari UA) is silently dropped and origins always receive `User-Agent: Lightpanda/1.0`.

There are two independent bugs; this addresses both.

### Bug 1: `Emulation.setUserAgentOverride` blanket-ignores any "Mozilla" UA

`Config.validateUserAgent` returned `error.Reserved` for any UA containing `"mozilla"` (case-insensitive), and `setUserAgentOverride` turned that into a warning + `sendResult(null)` without applying the UA. The intent (commit 920ae57f) was narrow: go-rod activates device emulation by default (`Devices.LaptopWithMDPIScreen`) and pushes that device's `Mozilla/...` UA over CDP on every connect, so the maintainers chose to ignore it rather than error (erroring breaks go-rod). But a blanket substring match also discards every legitimate `newContext({ userAgent })` call.

Fix: scope the workaround to go-rod's **exact** default device UA string instead of any UA containing `"Mozilla"`. go-rod's `LaptopWithMDPIScreen` UA is a hardcoded constant (`...Chrome/114.0.0.0 Safari/537.36`), so go-rod keeps working while any other deliberately-set UA is honored. The non-printable-character rejection is unchanged.

### Bug 2: `Network.setExtraHTTPHeaders` appends `User-Agent` instead of replacing it

Each request's header list is seeded with the default `User-Agent: Lightpanda/1.0` in `http.Headers.init`. `httpRequestStart` then applied the CDP extra headers with `req.headers.add`, a plain `curl_slist_append`. The result was two `User-Agent` entries; curl keeps the first (`Lightpanda/1.0`), so the caller's value was silently dropped (every other extra header worked, since they had no default to collide with).

Fix: add `Headers.set`, which drops any existing same-name (case-insensitive) header before appending, and use it for CDP extra headers. Extra headers now override request defaults instead of producing a duplicate curl discards. This restores the documented `setExtraHTTPHeaders({ "User-Agent": ... })` fallback.

## Why it matters

For a drop-in Chrome/Chromium replacement, presenting a normal browser User-Agent to origins is essential, and keeps `navigator.userAgent`, the outbound header, and the TLS fingerprint mutually consistent. With both fixes, the natural `newContext({ userAgent })` path and the `setExtraHTTPHeaders` fallback both work for realistic UAs; go-rod's auto-injected UA is still ignored.

## Tests

- `cdp.Emulation: setUserAgentOverride honors a deliberate Mozilla user agent` (was "ignores mozilla", inverted to assert the new behavior).
- `cdp.Emulation: setUserAgentOverride ignores the go-rod default user agent` (asserts the scoped workaround still fires for go-rod's exact UA).
- `Headers.set replaces a same-name header instead of duplicating` (one `User-Agent` entry carrying the overriding value; unrelated seeded defaults preserved).

Full suite green: `797 of 797 tests passed`; `zig fmt --check` clean.

### DIFF
--- a/src/Config.zig
+++ b/src/Config.zig
@@ -592,8 +592,4 @@ pub fn validateUserAgent(ua: []const u8) !void {
             return error.NonPrintable;
         }
     }
-
-    if (std.ascii.indexOfIgnoreCase(ua, "mozilla") != null) {
-        return error.Reserved;
-    }
 }
--- a/src/cdp/domains/emulation.zig
+++ b/src/cdp/domains/emulation.zig
@@ -24,6 +24,13 @@ const Config = @import("../../Config.zig");
 
 const log = lp.log;
 
+// go-rod activates device emulation by default (Devices.LaptopWithMDPIScreen)
+// and pushes that device's User-Agent over CDP on every connect. Honoring it
+// would make every go-rod session silently impersonate Chrome, so we ignore
+// this one exact value while accepting any other UA a client deliberately sets.
+// See https://github.com/lightpanda-io/browser/issues/2704.
+const gorod_default_user_agent = "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.0.0 Safari/537.36";
+
 pub fn processMessage(cmd: *CDP.Command) !void {
     const action = std.meta.stringToEnum(enum {
         setEmulatedMedia,
@@ -91,11 +98,12 @@ pub fn setUserAgentOverride(cmd: *CDP.Command) !void {
     const ua = params.userAgent;
     Config.validateUserAgent(ua) catch |err| switch (err) {
         error.NonPrintable => return cmd.sendError(-32602, "User agent contains non-printable characters", .{}),
-        error.Reserved => {
-            log.warn(.not_implemented, "Emulation.setUserAgentOverride", .{ .param = "userAgent", .value = ua, .info = "User agent must not contain Mozilla" });
-            return cmd.sendResult(null, .{});
-        },
     };
+
+    if (std.mem.eql(u8, ua, gorod_default_user_agent)) {
+        log.warn(.not_implemented, "Emulation.setUserAgentOverride", .{ .param = "userAgent", .value = ua, .info = "ignoring go-rod default user agent" });
+        return cmd.sendResult(null, .{});
+    }
 
     const bc = cmd.browser_context orelse return error.BrowserContextNotLoaded;
     const http_client = &cmd.cdp.browser.http_client;
@@ -121,10 +129,7 @@ test "cdp.Emulation: setUserAgentOverride with valid user agent" {
     try ctx.expectSentResult(null, .{ .id = 1 });
 }
 
-test "cdp.Emulation: setUserAgentOverride ignores mozilla" {
-    const filter: testing.LogFilter = .init(&.{.not_implemented});
-    defer filter.deinit();
-
+test "cdp.Emulation: setUserAgentOverride honors a deliberate Mozilla user agent" {
     var ctx = try testing.context();
     defer ctx.deinit();
     _ = try ctx.loadBrowserContext(.{ .id = "BID-UA2" });
@@ -135,11 +140,11 @@ test "cdp.Emulation: setUserAgentOverride ignores mozilla" {
         .params = .{ .userAgent = "Mozilla/5.0 (Windows NT 10.0)" },
     });
 
-    try ctx.expectSentResult(null, .{});
-    try testing.expectEqual(false, ctx.cdp().browser_context.?.user_agent_changed);
+    try ctx.expectSentResult(null, .{ .id = 2 });
+    try testing.expectEqual(true, ctx.cdp().browser_context.?.user_agent_changed);
 }
 
-test "cdp.Emulation: setUserAgentOverride ignores mozilla case insensitive" {
+test "cdp.Emulation: setUserAgentOverride ignores the go-rod default user agent" {
     const filter: testing.LogFilter = .init(&.{.not_implemented});
     defer filter.deinit();
 
@@ -150,7 +155,7 @@ test "cdp.Emulation: setUserAgentOverride ignores mozilla case insensitive" {
     try ctx.processMessage(.{
         .id = 3,
         .method = "Emulation.setUserAgentOverride",
-        .params = .{ .userAgent = "MOZILLA/5.0 test" },
+        .params = .{ .userAgent = gorod_default_user_agent },
     });
 
     try ctx.expectSentResult(null, .{});
--- a/src/cdp/domains/network.zig
+++ b/src/cdp/domains/network.zig
@@ -307,9 +307,11 @@ pub fn httpRequestStart(bc: *CDP.BrowserContext, msg: *const Notification.Reques
     const frame_id = req.frame_id;
     const frame = bc.session.findFrameByFrameId(frame_id) orelse return;
 
-    // Modify request with extra CDP headers
+    // Modify request with extra CDP headers. Use `set` so a caller-supplied
+    // header (e.g. User-Agent) overrides the request's default instead of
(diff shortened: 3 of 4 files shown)


karlseguin:
The first case isn't a bug, it's the current design of the system that any user agent containing "mozilla", whether set automatically by a tool, or manually in a script, is rejected.

The second case is handled by https://github.com/lightpanda-io/browser/pull/2735 and https://github.com/lightpanda-io/browser/pull/2706

Ar-maan05:
Thanks for the context. Closing this as redundant.

PR #2741, opened 2026-06-13
fix: Element.innerText emits block-level line breaks and skips display:none descendants

### LINKED ISSUE #2734: dom: Element.innerText emits no line breaks around block-level boxes and includes display:none descendants
## Summary

`Element.innerText` does not implement the HTML spec's [rendered-text collection steps](https://html.spec.whatwg.org/multipage/dom.html#the-innertext-idl-attribute): block-level boxes get no required line breaks (sibling `<div>`s concatenate into one run), and descendants that are not being rendered (`display: none`) are included in the output. The walker emits `\n` only for `<br>`. For text-heavy consumers — scraping, agentic text extraction, anything diffing rendered text against expectations — the output diverges from Chrome on essentially every multi-block element.

## Today's behavior

A CDP client evaluating `el.innerText` via `Runtime.evaluate` on `<div><div>first</div><div>second</div><span>tail</span></div>` gets `"firstsecondtail"`. With a `display:none` child (`visible<span style="display:none">secret</span>`), it gets `"visiblesecret"`.

```mermaid
sequenceDiagram
    participant LP as Lightpanda
    participant DOM as DOM walker
    Client->>LP: Runtime.evaluate innerText
    LP->>DOM: recurse children and emit text
    DOM-->>LP: newline only for br elements
    DOM-->>LP: hidden descendants included
    LP-->>Client: firstsecondtail
    Note over LP,DOM: src/browser/webapi/element/Html.zig getInnerText
```

## Expected behavior

Per [HTML §the-innerText-IDL-attribute](https://html.spec.whatwg.org/multipage/dom.html#the-innertext-idl-attribute): items at block-level boundaries contribute required line breaks (step: "If node's computed value of 'display' is 'block'... append a required line break count of 1"), nodes that are not being rendered are skipped entirely (step 2 of the inner-text collection steps), and consecutive required line breaks collapse. Chrome returns `"first\nsecond\ntail"` and `"visible"` for the fixture below.

```mermaid
sequenceDiagram
    participant LP as Lightpanda
    participant DOM as DOM walker
    Client->>LP: Runtime.evaluate innerText
    LP->>DOM: recurse children and emit text
    DOM-->>LP: required line breaks around block boxes
    DOM-->>LP: hidden descendants skipped
    LP-->>Client: first newline second newline tail
```

## Scope question (before any PR)

The full spec algorithm wants the **computed** value of `display` for both halves (line breaks and skipping), which intersects with #2733 (computed-style cascade resolution is currently limited to `display`/`visibility`). Two possible v1 scopes:

1. **Reuse what exists today** — `StyleManager` already answers `display:none` truthfully (it backs `checkVisibility()`), so hidden-descendant skipping needs no new cascade machinery. Line breaks resolve from the UA-default display per tag (`div`, `p`, `li`, `section`, ... → block) plus any author `display` override the StyleManager already tracks. This covers virtually all real pages without waiting on full cascade resolution.
2. **Full spec algorithm** — including the line-collapsing pass (required line breaks adjacent to empty blocks collapse, so `a<div></div>b` yields `"a\nb"` not `"a\n\nb"`), `display: table-cell`/`table-row` separators, and `text-transform` application. More faithful, but the tail of it depends on #2733's outcome.

Option 1 with the line-collapsing pass included seems like the right cut — it fixes the divergence on real pages and doesn't block on cascade work. Happy to implement once the scope is agreed.

## Reproducer

`repro.html`:
```html
<!DOCTYPE html>
<html>
<head>
  <title>innerText line-break repro</title>
</head>
<body>
  <div id="blocks"><div>first</div><div>second</div><span>tail</span></div>
  <div id="hidden-child">visible<span style="display: none">secret</span></div>
</body>
</html>
```

`repro.sh`:
```bash
#!/usr/bin/env bash
# Asserts that Element.innerText inserts required line breaks around
# block-level boxes and skips display:none descendants (HTML spec
# "rendered text" algorithm).
#
# Run:           ./repro.sh
# Expected today (bug): "firstsecondtail" / "visiblesecret"   exit 1
# Expected after fix:   "first\nsecond\ntail" / "visible"     exit 0

set -euo pipefail

CDP_PORT="${CDP_PORT:-9222}"

PIDS=()
cleanup() {
}
trap cleanup EXIT

cd "$REPRO_DIR"
PIDS+=("$!")

PIDS+=("$!")

for _ in $(seq 1 50); do
  sleep 0.1
done

fi

```

`probe.js`:
```js
// Asserts two halves of the HTML innerText algorithm:
//  1. required line breaks around block-level boxes ("first\nsecond\ntail")
//  2. skipping descendants that are not being rendered (display:none)

(async () => {
  const c = await connect();

  const blocks = await c.eval(`document.getElementById('blocks').innerText`);
  const hidden = await c.eval(`document.getElementById('hidden-child').innerText`);

  console.log(`block boxes innerText: ${JSON.stringify(blocks)} (expected "first\\nsecond\\ntail")`);
  console.log(`hidden child innerText: ${JSON.stringify(hidden)} (expected "visible")`);

  c.close();
  const breaksOk = blocks === 'first\nsecond\ntail';
  const hiddenOk = hidden === 'visible';
  if (!breaksOk || !hiddenOk) {
    if (!breaksOk) console.log('FAIL: no required line breaks around block-level boxes.');
    if (!hiddenOk) console.log('FAIL: display:none descendant leaks into innerText.');
    process.exit(1);
  }
  console.log('PASS: innerText follows the HTML rendered-text algorithm.');
  process.exit(0);
```

<details>

```js
// than this is long).

  let nextId = 1;
  const pending = new Map();
  const handlers = new Map();
    pending.clear();
  };
  ws.on('message', (raw) => {
      pending.delete(m.id);
      else resolve(m.result);
    } else if (m.method) {
    }
  });
  return {
      const id = nextId++;
      });
    },
    ready() {
        ws.on('open', resolve);
      });
    },
    close() { ws.close(); },
  };
}

  await client.ready();

  return {
    sessionId,
    close: () => client.close(),
    async eval(expression) {
      }, sessionId);
    },
    async navigate(url2) {
      let loaded = false;
        try {
          }, sessionId);
        } catch (e) {
        }
      }
    },
  };
}

module.exports = { connect };
```

</details>

**Run**: `bash repro.sh`
- Today (nightly `1.0.0-nightly.6736+63665f81`):
  ```
  block boxes innerText: "firstsecondtail" (expected "first\nsecond\ntail")
  hidden child innerText: "visiblesecret" (expected "visible")
  FAIL: no required line breaks around block-level boxes.
  FAIL: display:none descendant leaks into innerText.
  ```
  exits 1.
- Expected: prints `PASS: innerText follows the HTML rendered-text algorithm.` and exits 0.

Chrome control (same fixture, chrome-headless-shell): `"first\nsecond\ntail"` and `"visible"`.

## Likely fix location

`src/browser/webapi/element/Html.zig` — `getInnerText` / `_getInnerText`: the walker recurses children and emits `\n` only for `<br>` (the `// TODO check if elt is hidden` comment marks the missing skip). Needs block-boundary line breaks, hidden-descendant skipping, and the line-collapsing pass — see the scope question above.

## Environment

- Lightpanda: nightly `1.0.0-nightly.6736+63665f81` (macOS aarch64); walker behavior confirmed in source on `main`
- OS: darwin (macOS 15, aarch64)
- CDP client: raw `ws` (Node 26)

mvanhorn:
## Summary

Element.innerText emits block-level line breaks and skips display:none descendants

## Why this matters

`Element.innerText` does not implement the HTML spec's rendered-text collection steps. Sibling block-level boxes (e.g. adjacent `<div>`s) concatenate into a single run with no separating line break, and descendants that are not rendered (`display: none`) are still included in the output. Today the walker (`_getInnerText` in `src/browser/webapi/element/Html.zig`) only emits a `\n` for `<br>` elements and recurses into every child unconditionally (there is even a literal `// TODO check if elt is hidden.` comment at the child-recursion site). The result diverges from Chrome on essentially every multi-block element, which breaks scraping and agentic text extraction that diffs rendered text against expectations. The issue was filed by maintainer @navidemad with an exact reproduction and the precise function reference.

## Changes

Extend `_getInnerText` / `innerTextState` in `src/browser/webapi/element/Html.zig` to follow the spec's rendered-text steps for the two named gaps. First, skip descendants whose resolved `display` is `none` (and other non-rendered cases the existing StyleManager/computed-style path can already answer for `display`) before recursing into them, replacing the `// TODO check if elt is hidden.` branch. Second, insert a required line break around block-level boundaries: track whether the current node contributes a block box and append a single `\n` between consecutive block-level runs (collapsing leading/trailing/duplicate required breaks per the spec's "required line break count" handling) rather than emitting newlines only for `<br>`. Keep the change confined to the existing walker and its state struct; reuse the already-present display lookup that `checkVisibility`/computed-style consults so no new cascade machinery is introduced.

## Testing

- Happy path: `<div><div>first</div><div>second</div><span>tail</span></div>` yields `first\nsecond\ntail` (block boundaries break, inline tail does not).
- display:none: `visible<span style="display:none">secret</span>` yields `visible` (hidden descendant excluded).
- Edge: leading/trailing and consecutive block boundaries collapse to a single required break (no doubled or leading `\n`); `<br>` still emits its newline; pure-inline content (`<span>a</span><span>b</span>`) stays `ab`.
- Regression: existing inline-only innerText tests in `Html.zig` and `AXNode.zig` consumers continue to pass.

Fixes #2734

### DIFF
--- a/src/browser/webapi/element/Html.zig
+++ b/src/browser/webapi/element/Html.zig
@@ -219,13 +219,50 @@ pub fn asEventTarget(self: *HtmlElement) *@import("../EventTarget.zig") {
 // innerText represents the **rendered** text content of a node and its
 // descendants.
 pub fn getInnerText(self: *HtmlElement, writer: *std.Io.Writer) !void {
-    var state = innerTextState{};
+    var state = innerTextState{ .frame = ownerFrame(self.asNode()) };
+    return try self._getInnerText(writer, &state);
+}
+
+fn getInnerTextWithFrame(self: *HtmlElement, writer: *std.Io.Writer, frame: *Frame) !void {
+    var state = innerTextState{ .frame = frame };
     return try self._getInnerText(writer, &state);
 }
 
 const innerTextState = struct {
+    frame: ?*Frame = null,
     pre_w: bool = false,
     trim_left: bool = true,
+    pending_break: bool = false,
+    wrote_output: bool = false,
+    last_was_break: bool = false,
+
+    fn requireBreak(self: *innerTextState) void {
+        self.pending_break = true;
+        self.pre_w = false;
+        self.trim_left = true;
+    }
+
+    fn flushBreak(self: *innerTextState, writer: *std.Io.Writer) !void {
+        if (!self.pending_break) return;
+
+        self.pending_break = false;
+        self.pre_w = false;
+        self.trim_left = true;
+
+        if (self.wrote_output and !self.last_was_break) {
+            try writer.writeByte('\n');
+            self.last_was_break = true;
+        }
+    }
+
+    fn emitLineBreak(self: *innerTextState, writer: *std.Io.Writer) !void {
+        self.pending_break = false;
+        try writer.writeByte('\n');
+        self.pre_w = false;
+        self.trim_left = true;
+        self.wrote_output = true;
+        self.last_was_break = true;
+    }
 };
 
 fn _getInnerText(self: *HtmlElement, writer: *std.Io.Writer, state: *innerTextState) !void {
@@ -235,15 +272,21 @@ fn _getInnerText(self: *HtmlElement, writer: *std.Io.Writer, state: *innerTextSt
             .element => |e| switch (e._type) {
                 .html => |he| switch (he._type) {
                     .br => {
-                        try writer.writeByte('\n');
-                        state.pre_w = false; // prevent a next pre space.
-                        state.trim_left = true;
+                        if (!he.isRendered(state.frame)) continue;
+                        try state.emitLineBreak(writer);
                     },
                     .script, .style, .template => {
                         state.pre_w = false; // prevent a next pre space.
                         state.trim_left = true;
                     },
-                    else => try he._getInnerText(writer, state), // TODO check if elt is hidden.
+                    else => {
+                        if (!he.isRendered(state.frame)) continue;
+
+                        const is_block = he.asElement().getTag().isBlock();
+                        if (is_block) state.requireBreak();
+                        try he._getInnerText(writer, state);
+                        if (is_block) state.requireBreak();
+                    },
                 },
                 .svg => {},
             },
@@ -253,8 +296,18 @@ fn _getInnerText(self: *HtmlElement, writer: *std.Io.Writer, state: *innerTextSt
                     state.trim_left = true;
                 },
                 .text => {
-                    if (state.pre_w) try writer.writeByte(' ');
+                    const text = c.getData().str();
+                    const has_text = hasRenderedText(text);
+                    if (state.pending_break and !has_text) continue;
+                    try state.flushBreak(writer);
+
+                    const wrote_pre_space = state.pre_w;
+                    if (wrote_pre_space) try writer.writeByte(' ');
                     state.pre_w = try c.render(writer, .{ .trim_left = state.trim_left });
+                    if (wrote_pre_space or has_text or state.pre_w) {
+                        state.wrote_output = true;
+                        state.last_was_break = false;
+                    }
                     // if we had a pre space, trim left next one.
                     state.trim_left = state.pre_w;
                 },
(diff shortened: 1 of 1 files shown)


karlseguin:
Thanks for the changes. Unfortunately, the test doesn't seem to pass. Also, I'm not sure if this test approach was chosen for a reason, but they should probably have been written in https://github.com/lightpanda-io/browser/blob/main/src/browser/tests/element/inner.html which would have the advantage that we can open that file in Firefox and quickly verify the correctness of the new cases. 

I'm also concerned about performance, as this add a reasonable amount of weight to a call that could be in a hot path. The visibility lookup can be cached, which is important for this type of treewalk.

With the tests failing, and the tests not following the established pattern, I haven't looked too deeply into the change. If you'd like to make adjustments to (a) fix the tests (b) move them to .html files (c) look at performance, that would be appreciated. Otherwise, we can close the PR and I'll take a look at it sometime this week (because I agree it's important to fix).

PR #2742, opened 2026-06-15
uaf, webapi: Window reuse

karlseguin:
Depends on https://github.com/lightpanda-io/zig-v8-fork/pull/182

This PR addresses 1 of 2 UAFs.

A few things going on. The simpler change is that a Window maintains its identity even when an iframe/popup navigates. So the Window of an iframe before and after a navigation is the same window (both in Zig and in JS).

While this is correct on the JS side, I think it's a minor compliance win and, on its own, possibly not worth the complexity. However, it also addresses 1 UAF bug.

After [rust-url integration](https://github.com/lightpanda-io/browser/pull/2658) was merged, I started to investigate new WPT crashes. It's a bit complicated. Fundamentally, we have 2 memory models: things live forever (until page end) or are reference counted. Because these are bi-directional graphs, you can't really mix the two memory models. If you do RC, then you need to do RC across everything in that graph. 

For that reason, most things don't do RC. It's just so much easier to lean on the page's arena. `Window` doesn't do RC. And that used to mean that if we did something like:

```
var w = iframe.contentWindow.
iframe.src = 'spice';
console.log(w.location)
```

Everything would work. Although we "destroy" the Frame, that Window continues to exist and can be reached from JS. 

The rust-url integration broke this by making URL reference counted. This is fine for JavaScript created URLs, but it breaks for window.location which is Zig-owned. In the above example, that abandoned Window, which is still valid, references a freed rust-url (freed in Page.deinit). 

One approach might have been to try to detach the URL's lifetime from the frame. I believe this would have fixed the UAF, but it would have resulted in incorrect data. e.g. the above snippet should print the new url, not the old one.

This PR addresses UAF caused by a navigation. It does not address the UAF from a `popup.close()`. I'll address that separately, likely after this is merged.

### DIFF
--- a/.github/actions/install/action.yml
+++ b/.github/actions/install/action.yml
@@ -13,7 +13,7 @@ inputs:
   zig-v8:
     description: 'zig v8 version to install'
     required: false
-    default: 'v0.4.7'
+    default: 'v0.4.8'
   v8:
     description: 'v8 version to install'
     required: false
--- a/.github/actions/v8-snapshot/action.yml
+++ b/.github/actions/v8-snapshot/action.yml
@@ -13,7 +13,7 @@ inputs:
   zig-v8:
     description: 'zig-v8 release tag the prebuilt lib came from'
     required: false
-    default: 'v0.4.7'
+    default: 'v0.4.8'
 
 runs:
   using: "composite"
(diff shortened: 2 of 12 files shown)


krichprollsch:
Did you see the tests don't pass?

karlseguin:
I did, but I ignored it. I thought I was just because a new V8 version needs to be tagged. Looking into it now, I'm not so sure...

EDIT: it is just the stale v8, we need to tag a new one https://github.com/lightpanda-io/zig-v8-fork/pull/182

PR #2743, opened 2026-06-15
js: migrate the agent script runtime onto the js.* abstraction

arrufat:
The agent script runtime (`src/script/Runtime.zig`) gave itself a bare, WebAPI-less V8 context driven by ~130 raw `v8.*` C bindings. This moves all of that onto the `js.*` wrappers in `src/browser/js/`, so **`Runtime.zig` now uses zero raw `v8.*`**.


karlseguin (approved):
I am OK with bare and the unreachables everywhere!

PR #2753, opened 2026-06-16
webapi: fire iframe load on next tick

karlseguin:
Depends on: https://github.com/lightpanda-io/browser/pull/2724

Unblocks some WPT tests, but more generally, adds correctness so that the onload on an iframe doesn't fire before subsequent scripts. Imagine something like:

```
<iframe src=blah onload=done>
<script>
function done() {
 //...
}
</script>
```

(Which is an approximation of what some WPT tests do)...kind of important that we don't fire the onload while we're still parsing.


krichprollsch (approved):

krichprollsch:
We didn't dispatch the `load` event for iframes until now?

karlseguin:
We did, but not on the next tick.

PR #2767, opened 2026-06-17
implement BroadcastChannel

### LINKED ISSUE #2761: fetch --dump markdown crashes with ReferenceError: BroadcastChannel is not defined on Huawei developer pages
When using the CLI `fetch` command on certain Huawei developer documentation pages, Lightpanda throws a JavaScript evaluation error because `BroadcastChannel` is not defined in its JS runtime. The process ultimately returns an `Unknown` exception instead of the page content.

## To Reproduce

Run the following command:

```bash
lightpanda fetch --dump markdown https://developer.huawei.com/consumer/cn/doc/harmonyos-guides/ide-insight-description
```

## Expected behavior

The page should be fetched and dumped as markdown, or at least gracefully handle the missing `BroadcastChannel` API without an unrecoverable exception.

## Actual behavior

The CLI emits a JS warning and then crashes with an `Unknown` exception:

```
$time=1781671265620 $scope=js $level=warn $msg="evaluate module" stack=
 specifier=https://developer.huawei.com/consumer/cn/doc/main.b330376941674c63.js message="ReferenceError: BroadcastChannel is not defined"
$time=1781671265620 $scope=js $level=warn $msg="eval script" url=https://developer.huawei.com/consumer/cn/doc/main.b330376941674c63.js
caught.exception=Unknown caught.stack=na caught.line=null caught.caught=false cacheable=true
```

## Environment

- Lightpanda version: nightly (installed via `scripts/install.sh` around 2025-06-17)
- OS: macOS
- Command: `lightpanda fetch --dump markdown <url>`

## Additional context

- The offending script (`main.b330376941674c63.js`) appears to be a core bundle for the Huawei docs site.
- `BroadcastChannel` is a standard Web API for cross-context messaging, but it seems unimplemented in Lightpanda's JS runtime. A stub or minimal implementation might prevent this crash.

krichprollsch:
Fix #2761


karlseguin:
@krichprollsch I updated the BroadcastChannel to be available on workers and, more substantially, allow messages to be posted on frames/popups/workers of the same origin.

krichprollsch:
Seems good to me, thanks for the change. I'm just not sure I fully understand the `sequence` mechanism purpose :thinking:

karlseguin:
There's a WPT test that adds a broadcastchannel in its onmessage callback. It should not receive the message. There's a lot of things like that actually, for mutation observer and slotchanges, we capture them in 1 iteration, then loop through the capture and fire off the events. It's a bit more complicated here since we need to traverse the entire tree to find channels. So easier with the sequence cutoff.

PR #2773, opened 2026-06-18
Internal Flag on Request

### LINKED ISSUE #2749: `fetch --obey-robots` hangs forever when robots.txt returns a non-2xx status (301/403)
**Version:** 1.0.0-nightly.6762+3984082c

### Behavior
`fetch <url>` works fine. Add `--obey-robots` and it never returns.

### example
```bash
# Works fine
./lightpanda fetch https://driscollmotors.co.nz --dump html

# Never return
./lightpanda fetch https://driscollmotors.co.nz --obey-robots --log-level debug
```

### Likely cause: non-2xx robots.txt

The page loads assets from `static.cdn-website.com`, and when I killed the run it was fetching that host's robots.txt. Both robots.txt endpoints return non-2xx, and both respond quickly:

```
curl https://driscollmotors.co.nz/robots.txt    => 301 in 0.6s
curl https://static.cdn-website.com/robots.txt  => 403 in 1.1s
```

mookums:
This adds an `internal` boolean on the `Request` struct, replacing the `skip_robots`. This `internal` boolean signals to layers that this `Request` is internal to the browser and should skip correctness layers (such as Robots or Deferring). This fixes #2749.

### DIFF
--- a/src/browser/HttpClient.zig
+++ b/src/browser/HttpClient.zig
@@ -1324,7 +1324,10 @@ pub const Request = struct {
     credentials: ?[:0]const u8 = null,
     notification: *Notification,
     timeout_ms: u32 = 0,
-    skip_robots: bool = false,
+
+    // Requests that are internal to the browser and skip various layers,
+    // these do not need to be deferred and do not obey robots.txt.
+    internal: bool = false,
 
     // When false, the caller does not guarantee that the body outlives the
     // transfer, and thus we'll need to dupe it.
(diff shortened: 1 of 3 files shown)


krichprollsch:
Why running the request through the DeferringLayer causes the issue?

karlseguin:
Does it relate to the failed assertion Transfer.unpark assertion we saw on the 16th (which appears to be a UAF in RobotsLayer, but I can't tell if it's related)?

krichprollsch:
It relates with #2749 I didn't have crash on reproduction, idk if it has an impact on `Transfer.unpark` assertion.

krichprollsch (approved):

PR #2784, opened 2026-06-18
agent: add --save flag to synthesize scripts from tasks

arrufat:


src/help.zon
```diff
@@ -201,7 +202,14 @@
...
     \\--task <STRING>            One-shot mode: run a single user turn, print the
     \\                           final answer to stdout, and exit. Conflicts with
-    \\                           the positional script.
+    \\                           the positional script. With --save, the answer is
+    \\                           suppressed and a script is written instead.
+    \\
+    \\--save <PATH>              Synthesize a replayable .js script from the --task
+    \\                           run and write it to PATH, instead of printing the
+    \\                           answer. Replay it later with `agent PATH` (no LLM
+    \\                           calls). Overwrites PATH if it exists. PATH must be a
```
karlseguin: "Path must not be a path"  😕

karlseguin (approved):
Not  sure about that help text, but LGTM othewise.

src/help.zon
arrufat: Right, it reads weird, fixing that.

This was because the MCP can also save files (shared save code), and since there's an agent driving that, I only allowed it to save in the current directory.

For the flag (human-driven) we can allow any path.

PR #2788, opened 2026-06-19
uaf: Extend closed popup's lifetime from Frame to Page

karlseguin:
Follow up to https://github.com/lightpanda-io/browser/pull/2742

Where 2742 dealt with window-reuse across navigation, this commit addresses window lifetime after a .close() (so, specifically for popups). I was hoping for a fancier solution which would let us maintain eager release with safety, but instead opted for the much simpler "window lifetime is tied to the page".

There's a couple reasons for this. It _is_ more consistent with how everything else works (Workers, and iframes), but the real reason is that we lack the infrastructure to do this safely/cleanly.

### DIFF
--- a/src/browser/Page.zig
+++ b/src/browser/Page.zig
@@ -110,6 +110,11 @@ frame: Frame,
 // to the original page like this.
 popups: std.ArrayList(*Frame) = .empty,
 
+// Popup Frames that have been closed. The window can still be referenced / used
+// from JS, so we defer shutting them down until page tear down (which isn't
+// ideal from a memory point of view).
+closed_frames: std.ArrayList(*Frame) = .empty,
+
 // Lifecycle state. A Page is `.pending` while we hold it as the in-flight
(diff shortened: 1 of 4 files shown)


krichprollsch (approved):
Can't we use a `fn deinit()` to cleanup resources earlier once the JS garbage collect it?

karlseguin:
> Can't we use a `fn deinit()` to cleanup resources earlier once the JS garbage collect it? 

Possibly, but, how safe is it? The real win is killing the v8.Context.
```
var div = null;
{
  var opener = window.open('blah');
  opener.document.querySelector('div');
  opener.close();
}
<-- v8 signals window (i.e. opener) finalized

// we can still access it
div.ownerDocument.defaultView.postMessage('...');
```

In general, this is safe for our code to do, because most things live in the slab allocator which lives for the entire page. But if we start tearing down things before Page.deinit, I dunno. None of the code is written with that robustness in mind. We would need to make the `js: *Context` field nullable in Frame and then guard its use everywhere.

PR #2789, opened 2026-06-22
design: multi-page session

karlseguin:
Replaces the 1 (+1 inflight) page design of session for an unlimited number of pages. The main goal is to support a more efficient async goto (1). Without this commit, async goto has two implementation: multiple browser, which is resource intensive (thread + isolate per page), bolting it onto sub-frames (like iframe or popups). The issue with the 2nd version (which I originally suggested) is that most resources are tied to the Page. So even if an async goto "page" (which would map to a Frame) is released, most things stay in memory, including the DOM (page.factory) and the V8::Context.

This new approach adds multiple page support to sessions. The advantage is pretty clear: the existing memory model (page-tied resources) becomes a strength of the design, rather than a weakness.

This change is not tirivla, but the diff is inflated by 2 large mechanial changes, so it isn't _that_ big either. That said, I'd divide this into four parts.

1 - The old concept of _active/_pending is now baked into the Page itself. A Page has a `replaces: ?*Page = null` and `replacement: ?*Page = null` field.

2 - Because of #1 above, the Session now just has an `pages: ArrayList(*Page)`. As much as possible, single-page APIs, like `removePage` no longer exists. We're trying to present a consistent multi-page API. See #3.

3 - References to *Page and *Frame have always been dangerous. For example, `lightpanda.fetch` has a block to scope `frame`:

```zig
{
    const frame = try session.createPage();
    // frame isn't safe to use after navigate, it can be swapped out
    _ = try frame.navigate(...)
}
```

While we still hand out *Page and *Frame (user-beware), some of the more important APIs now take a frame_id which Session can resolve. Furthermore, createPage now returns a PageHandle which is a safe wrapper around a page/frame.

4 - Two large mechanical changes were made:
  a - Many tests were superficially changed to account for new naming or use
      a new test-helper to preserve the single-page illusion (because 99% of
      tests _are_ single-page)
  b - Many CDP changes where `bc.session.currentFrame()` -> `bc.mainFrame()`).
      This isn't to say CDP changes are meaningless, but it's mostly 1 change
      about how the "main" target/frame_id is tracked (by CDP itself, rather
      than the Session) that required a number of superficial changes to
      accomodate

`Runner` remains largely single-page focused. Runner and some MCP/Agent tools continue to be tied to the "currentFrame" or "primaryPage". The Runner is something I want to address in a follow-up PR, but we need to figure out what it means to "wait" for multiple pages.
As much as possible, Session becomes multi-page native and has no concept of a special/first/primary/main page. Users of session become responsible for tracking pages of interest.

This change is not trivial, but the diff is inflated by 2 large mechanical changes:
1 - Many tests

(1) https://github.com/lightpanda-io/browser/pull/2759

### DIFF
--- a/src/browser/Runner.zig
+++ b/src/browser/Runner.zig
@@ -33,20 +33,34 @@ const IS_DEBUG = builtin.mode == .Debug;
 
 const Runner = @This();
 
+// The "main" or "primary" frame which is used against our `until` checks. This
+// is, hopefully, temporary during our transition to multi-page sessions. It
+// helps preserve existing behavior for single-page sessions (which CDP and
+// fetch use exclusively). But once we start adding multi-page to Agent, and
+// maybe CDP and fetch, then "completeness" should gain a clearer meaning.
 frame: *Frame,
+
 session: *Session,
 http_client: *HttpClient,
 
 pub const Opts = struct {};
 
 pub fn init(session: *Session, _: Opts) !Runner {
-    const frame = session.currentFrame() orelse return error.NoPage;
+    const frame = firstFrame(session) orelse return error.NoPage;
 
     return .{
         .frame = frame,
         .session = session,
         .http_client = &session.browser.http_client,
     };
+}
+
+// See the `frame` field. Hopefully this goes away and we make Runner truly
+// multi-page aware.
+fn firstFrame(session: *Session) ?*Frame {
+    const page = (session.primaryPage() orelse return null).page().?;
+    const target = session.replacementOf(page) orelse page;
+    return &target.frame;
 }
 
 pub const WaitOpts = struct {
@@ -156,12 +170,22 @@ pub fn tick(self: *Runner, opts: TickOpts) !TickResult {
 }
 
 fn _tick(self: *Runner, comptime is_cdp: bool, opts: TickOpts) !TickResult {
+    const session = self.session;
+    const http_client = self.http_client;
+
+    // Drain queued navigations across every live page (one page per call)
+    // A navigation can swap a frame pointer or the page set,
+    // so restart the tick to re-resolve cleanly.
+    if (try session.processQueuedNavigation()) {
+        self.frame = firstFrame(session) orelse return .done;
+        return .{ .ok = 0 };
+    }
+
     // Refresh self.frame from session. In case of pending page, we want to
     // take its state while loading. If we use only the current frame, we will
     // return a .done result immediately.
-    self.frame = self.session.pendingOrCurrentFrame() orelse return .done;
-    const frame = self.frame;
-    const http_client = self.http_client;
+    const frame = firstFrame(session) orelse return .done;
+    self.frame = frame;
 
     switch (frame._parse_state) {
         .pre, .raw, .text, .image, .download => {
@@ -177,14 +201,6 @@ fn _tick(self: *Runner, comptime is_cdp: bool, opts: TickOpts) !TickResult {
             return .{ .ok = 0 };
         },
         .html, .complete => {
-            const session = self.session;
-            if (session.currentPage()) |page| {
-                if (page.queued_navigation.items.len != 0) {
-                    try session.processQueuedNavigation();
-                    self.frame = session.currentFrame().?; // might have changed
-                    return .{ .ok = 0 };
-                }
-            }
             const browser = session.browser;
 
             // The HTML page was parsed. We now either have JS scripts to
@@ -381,35 +397,34 @@ test "Runner: no page" {
 }
 
 test "Runner: waitForSelector timeout" {
-    const frame = try testing.pageTest("runner/runner1.html", .{});
-    defer frame._session.removePage();
+    const page = try testing.pageTest("runner/runner1.html", .{});
+    defer page.close();
 
-    var runner = try frame._session.runner(.{});
+    var runner = try page.session.runner(.{});
     try testing.expectError(error.Timeout, runner.waitForSelector("#nope", 10));
 }
 
 test "Runner: waitForSelector" {
     defer testing.reset();
-    const frame = try testing.pageTest("runner/runner1.html", .{});
-    defer frame._session.removePage();
+    const page = try testing.pageTest("runner/runner1.html", .{});
 
-    var runner = try frame._session.runner(.{});
+    var runner = try page.session.runner(.{});
     const el = try runner.waitForSelector("#sel1", 10);
     try testing.expectEqual("selector-1-content", try el.asNode().getTextContentAlloc(testing.arena_allocator));
 }
 
 test "Runner: waitForScript timeout" {
-    const frame = try testing.pageTest("runner/runner1.html", .{});
-    defer frame._session.removePage();
+    const page = try testing.pageTest("runner/runner1.html", .{});
+    defer page.close();
 
-    var runner = try frame._session.runner(.{});
+    var runner = try page.session.runner(.{});
     try testing.expectError(error.Timeout, runner.waitForScript("document.querySelector('#nope')", 10));
 }
 
 test "Runner: waitForScript" {
-    const frame = try testing.pageTest("runner/runner1.html", .{});
-    defer frame._session.removePage();
+    const page = try testing.pageTest("runner/runner1.html", .{});
+    defer page.close();
 
-    var runner = try frame._session.runner(.{});
+    var runner = try page.session.runner(.{});
     try runner.waitForScript("document.querySelector('#sel1')", 10);
 }
--- a/src/browser/Session.zig
+++ b/src/browser/Session.zig
@@ -334,8 +334,6 @@ fn installNewActivePage(self: *Session, frame_id: u32) !*Frame {
     return frame;
 }
 
-// NOTE: the caller is not the owner of the returned value,
-// the pointer on Frame is just returned as a convenience
 pub fn createPage(self: *Session) !PageHandle {
     // Drain any pending Page deinits now, while we're at a known-safe point
     self.processDestroyQueues();
--- a/src/browser/Session.zig
+++ b/src/browser/Session.zig
@@ -27,10 +27,10 @@ const storage = @import("webapi/storage/storage.zig");
 const Navigation = @import("webapi/navigation/Navigation.zig");
 const History = @import("webapi/History.zig");
 
-const Frame = @import("Frame.zig");
 const Page = @import("Page.zig");
-pub const Runner = @import("Runner.zig");
+const Frame = @import("Frame.zig");
 const Browser = @import("Browser.zig");
+pub const Runner = @import("Runner.zig");
 const Notification = @import("../Notification.zig");
 const QueuedNavigation = Frame.QueuedNavigation;
 
@@ -40,13 +40,10 @@ const Allocator = std.mem.Allocator;
 const IS_DEBUG = builtin.mode == .Debug;
 
 // A Session represents a browsing context group (cookie jar, session storage,
-// navigation history) within a Browser. It hosts one Page at a time — the
+// navigation history) within a Browser. It owns a set of live Pages — each a
 // root Frame and all of its descendants — and is responsible for Page
 // lifecycle (create, remove, replace on root navigation).
-//
-// Multiple concurrent Pages (e.g. an old Page retiring while a new provisional
-// Page is loading) are not yet supported; see Page.zig for the intended
-// direction.
+
 const Session = @This();
 
 browser: *Browser,
@@ -65,14 +62,14 @@ inject_scripts: []const []const u8 = &.{},
 // Shared allocator. Used by Session itself and borrowed by Pages.
 arena_pool: *ArenaPool,
 
-// The currently-active Page
-// flips this pointer.
-_active: ?*Page = null,
-
-// In-flight root navigation
-_pending: ?*Page = null,
+// All live top-level Pages. During a root navigation this transiently holds
+// both the live page and its in-flight replacement.
+pages: std.ArrayList(*Page) = .{},
 
 _page_destruction_queue: std.ArrayList(*Page) = .{},
+
+// Round-robin cursor for fair page iteration (processQueuedNavigation)
+_nav_cursor: usize = 0,
 
 // Loader IDs are scoped to the Session: each new BrowserContext gets a
 // fresh counter. Frame IDs (`frame_id_gen`) live on `Browser` instead so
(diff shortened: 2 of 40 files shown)


src/browser/Session.zig
```diff
@@ -651,7 +730,11 @@ fn processRootQueuedNavigation(self: *Session) !void {
 // queued-navigation path (processRootQueuedNavigation) and the CDP entry point
 // (initiateRootNavigation); each caller owns any arena tied to `url`/`opts`.
 fn replaceRootImmediate(self: *Session, frame_id: u32, url: [:0]const u8, opts: Frame.NavigateOpts) !void {
-    self.tearDownActivePage();
+    if (self.livePage(frame_id)) |page| {
```
navidemad, line 739: **Double-free / use-after-free: synthetic re-navigation while an HTTP replacement is in flight.**

`replaceRootImmediate` tears down the live page via `tearDownPage` → `retire` (`Session.zig:266`) **without first discarding an in-flight replacement**. Compare `initiateRootNavigation` (`Session.zig:758`), which explicitly does `if (self.replacementOf(live)) |old| self.discardPendingPage(old)` before re-navigating. The synthetic queued path skips that guard.

Trigger (single-page, reachable today): page `P` starts an HTTP root nav → replacement `R` (`pages = [P, R]`, `R.replaces = P`, `P.replacement = R`). Before `R`'s headers arrive, `P`'s JS schedules an `about:blank`/`blob:` navigation (a timer/microtask — `P` is still the live, JS-active page). Next tick `processQueuedNavigation` picks `P` → `processRootQueuedNavigation` → `replaceRootImmediate`. `retire(P)` only reconciles `page.replaces` (null for `P`), so `R.replaces` is never cleared, and `P` is `queuePageDestruction`'d. `R` now dangles at the to-be-freed `P`. When `R`'s headers arrive, `commitPendingPage(R)` reads `old = R.replaces` = `P`, writes `old.replacement = null`, then `queuePageDestruction(P)` **a second time** → double-free (or UAF if `P` was already drained).

Deep fix (preferred over a per-caller discard): make `retire`/`tearDownPage` self-consistent — when retiring a page that still holds a `.replacement`, sever/discard it too, so no caller can leave a half-linked pair regardless of entry path.

src/browser/interactive.zig
```diff
@@ -496,8 +496,8 @@ fn getListenerTypes(target: *EventTarget, listener_targets: ListenerTargetMap) [
 const testing = @import("../testing.zig");
 
 fn testInteractive(html: []const u8) ![]InteractiveElement {
-    const frame = try testing.test_session.createPage();
-    defer testing.test_session.removePage();
+    const frame = try testing.createFrame();
+    defer testing.test_session.closeAllPages();
```
navidemad: **Use-after-free: `defer closeAllPages()` frees the page synchronously, but the returned slices live in `frame.call_arena`.**

`testInteractive` returns `collectInteractiveElements(..., frame.call_arena, frame)` (line 506) while this `defer testing.test_session.closeAllPages()` fires on return. Unlike the old `removePage()` (which only *queued* destruction), `closeAllPages` → `processDestroyQueues()` runs inline and calls `page.deinit()` immediately, releasing `call_arena`. The caller (`test "browser.interactive: button"`, etc.) then reads `elements[0].tag_name` from freed memory.

Currently masked — the released arena bytes aren't reused before the read, so the tests pass — but it's a real UAF and a regression from the `removePage`→`closeAllPages` semantic change (deferred vs. immediate destruction). Fix: don't free inside the helper. Return the page/handle and let the *test* close it after its assertions (as `forms.zig`/`structured_data.zig` now do).

src/browser/Session.zig
```diff
@@ -370,35 +379,76 @@ pub fn releaseArena(self: *Session, allocator: Allocator) void {
...
 
-pub fn pendingOrCurrentFrame(self: *Session) ?*Frame {
-    const page = self.pendingPage() orelse self.currentPage() orelse return null;
-    return &page.frame;
+// DEPRECATED.
+// Needed by Runner so long as Runner is largel single-page driven.
+// Ultilatemy the goal is that runner is full multi-page, but during this
+// transition to multi-page sessions, we maintain the idea that one page is
+// the "main" page.
+pub fn primaryPage(self: *Session) ?PageHandle {
```
navidemad, line 422: **Latent (multi-page): `pages.items[0]` is not guaranteed to be the live primary page.**

`primaryPage()` here and `currentFrame()` (line 430) assert `pages.items[0].replaces == null` and treat index 0 as primary. But `swapRemove` in `retire` (~line 276) and `commitPendingPage` (~line 865) moves the *last* element into the vacated slot. With ≥2 simultaneous live pages: `pages = [P0, P1, R1]` (`R1.replaces = P1`); closing `P0` → `swapRemove(0)` → `pages = [R1, P1]`, so `pages[0] = R1` with `replaces != null`. The debug assert then panics, and in release `currentFrame()` returns the not-yet-committed replacement frame — `Runner.firstFrame` (called every tick) drives the wrong page.

Not reachable today (single live page), but the data structure this PR introduces enables it. Consider a dedicated primary-page field/handle rather than a positional convention duplicated across `primaryPage`/`currentFrame`/`firstFrame`.

src/browser/Session.zig
```diff
@@ -370,35 +379,76 @@ pub fn releaseArena(self: *Session, allocator: Allocator) void {
...
 }
 
+// Multi-page aware: frame ids are globally unique (monotonic on `Browser`), so
+// search every live page's frame tree. During a root navigation the old and
+// replacement pages share the root frame_id; the old page sits at index 0 and
+// is matched first — the addressable page until commit.
 pub fn findFrameByFrameId(self: *Session, frame_id: u32) ?*Frame {
-    const page = self.currentPage() orelse return null;
-    return page.findFrameByFrameId(frame_id);
+    for (self.pages.items) |page| {
```
navidemad, line 452: **Latent (multi-page): root frame_id resolved by array order, not the `replaces` link.**

During an in-flight root nav the old and replacement pages share the root frame_id. `livePage` resolves this correctly via `page.replaces orelse page`, but `findFrameByFrameId` returns the *first page in array order* whose tree contains the id — the comment hard-codes "the old page sits at index 0 and is matched first." As in the `primaryPage` finding, `swapRemove` can reorder `pages` so the replacement precedes the old page, and this would then hand CDP the still-loading replacement frame mid-commit. Route root-frame resolution through `livePage` so both paths agree under any array order.

src/browser/Runner.zig
```diff
@@ -156,12 +170,22 @@ pub fn tick(self: *Runner, opts: TickOpts) !TickResult {
 }
 
 fn _tick(self: *Runner, comptime is_cdp: bool, opts: TickOpts) !TickResult {
+    const session = self.session;
+    const http_client = self.http_client;
+
+    // Drain queued navigations across every live page (one page per call)
+    // A navigation can swap a frame pointer or the page set,
+    // so restart the tick to re-resolve cleanly.
+    if (try session.processQueuedNavigation()) {
```
navidemad, line 179: **Behavior change: queued navigations are now drained at the top of every `_tick`, for all parse states.**

Previously this ran only inside the `.html, .complete` switch arm (after `.pre/.raw/.text/.image/.download` returned early) and only against `currentPage()`. It's now unconditional and round-robins all pages, so a frame still streaming its body (`.raw`/`.download`) that carries a queued navigation has it processed *before* reaching `.complete` instead of after. Likely benign (queued navs normally originate post-parse from JS), but it's an observable semantics change and it's precisely what re-enables the `replaceRootImmediate` double-free above. Worth a regression test: queue a nav while the root frame is mid-load and confirm it isn't drained prematurely.

src/browser/tools.zig
```diff
@@ -1423,8 +1423,10 @@ fn mapActionError(err: anytype) ToolError {
 /// If the previous action queued a navigation (form submit, link click,
 /// Enter on an input), drive the runner until it completes or times out.
 fn awaitQueuedNavigation(session: *lp.Session) ToolError!void {
-    const page = session.currentPage() orelse return;
-    if (page.queued_navigation.items.len == 0) return;
+    const navigated = session.processQueuedNavigation() catch return ToolError.InternalError;
```
navidemad, line 1426: **`awaitQueuedNavigation` now *processes* the navigation (and advances the session-global `_nav_cursor`) before building the runner, rather than checking and letting the runner drive it.**

`processQueuedNavigation()` can synchronously tear down/replace the page (the synthetic `replaceRootImmediate` path dispatches `frame_remove`/`frame_created` and swaps the page) *outside* any wait loop. Two concerns: (1) the page mutation is committed before `session.runner(.{})` is even constructed — if that fails the swap already happened with no wait; (2) it consumes/advances the shared round-robin cursor, so once an agent drives >1 page the action's *own* queued nav may not be the one processed/awaited here. Equivalent for today's single-page agent, but the coupling is fragile.

src/browser/forms.zig
```diff
@@ -289,14 +288,15 @@ fn testForms(html: []const u8) ![]FormInfo {
...
     defer testing.reset();
-    defer testing.test_session.removePage();
     const forms = try testForms(
         \\<form action="/login" method="POST">
         \\  <input type="email" name="email" required placeholder="Email">
         \\  <input type="password" name="password" required>
         \\  <input type="submit" value="Log In">
         \\</form>
     );
+    defer testing.test_session.closeAllPages();
```
navidemad, line 299: **Test cleanup regression (error path): `defer closeAllPages()` is registered *after* `try testForms(...)`.**

If `testForms` errors, this `defer` was never registered, so the page created on the shared `test_session` leaks — and since `testing.reset()` only resets the arena (it doesn't drain `session.pages`), the leak detector fails the test for an unrelated reason, masking the real failure. Same shape in `structured_data.zig` (which additionally *deleted* its helper-internal `errdefer removePage()`) and `links.zig`. Register the cleanup before the call, or give the helper an `errdefer`.

src/cdp/CDP.zig
```diff
@@ -729,14 +735,32 @@ pub const BrowserContext = struct {
...
+    // True while this context's page is mid-commit (a root navigation's
+    // replacement is being promoted).
+    pub fn inCommit(self: *const BrowserContext) bool {
+        const page = self.mainPage() orelse return false;
+        return self.session.replacementOf(page) != null;
+    }
+
+    // The live Page for this single-target context (resolved by frame_id),
+    // or null if no page is loaded.
+    pub fn mainPage(self: *const BrowserContext) ?*Page {
```
navidemad, line 752: **`BrowserContext` re-implements `PageHandle` inline instead of holding one.**

`mainPage()` is `session.livePage(self.frame_id orelse return null)` — byte-for-byte `PageHandle.page()` (`Session.zig:921`), and `frame_id: ?u32` + `session` is exactly `PageHandle`'s state. This PR introduces `PageHandle` as "a stable, navigation-safe reference to a top-level browsing context," yet the most important CDP consumer keeps a parallel hand-rolled copy. Any future change to how a handle resolves its live page (the index-0/`replaces` resolution, popup/multi-target handling) must now be made in two places. Consider having `BrowserContext` hold a `PageHandle` and delegate `mainPage`/`mainFrame`/`inCommit` to it.

src/cdp/domains/page.zig
```diff
@@ -240,7 +240,8 @@ fn close(cmd: *CDP.Command) !void {
         bc.session_id = null;
     }
 
-    bc.session.removePage();
+    if (bc.frame_id) |frame_id| bc.session.closePage(frame_id);
```
navidemad: zig-style §3 ("Always brace control-flow bodies. Even a single-statement `continue` / `return` / `break`."): this capture body is unbraced. The equivalent close path in `target.zig:308` braces it.

```suggestion
    if (bc.frame_id) |frame_id| {
        bc.session.closePage(frame_id);
    }
```

navidemad:
Non-blocking follow-ups from the review (cleanup + polish — captured here rather than as separate inline threads to keep the diff readable):

### Cleanup
- **Duplicated `swapRemove` loop** — the `for (self.pages.items, 0..) |p, i| { if (p == page) { _ = swapRemove(i); break; } }` block is copy-pasted in `retire` (`Session.zig:274`) and `commitPendingPage` (`Session.zig:849`). Extract a `removePageFromList(page)` built on `std.mem.indexOfScalar`.
- **`primaryPage` / `currentFrame` near-duplicate** (`Session.zig:418` / `430`) — same `len==0` guard + `pages.items[0]` + identical debug assert, differing only in return shape. Have one call the other.
- **`Runner.firstFrame` hot path** (`Runner.zig:60`) — resolves one frame by walking `self.pages` 2–3× (`primaryPage()` → `.page()`/`livePage` → `replacementOf`), called from `init` and every `_tick`. `pages[0].replacement orelse pages[0]` gives it directly.
- **Redundant `replaces`/`replacement` pair** (`Page.zig:135`) — `replacement` is derivable (`p where p.replaces == page`). The hand-synced pair plus the per-call debug scan in `replacementOf` (`Session.zig:399`, also on the `firstFrame` hot path) is the cost of storing both. Keep one and compute the other.

### Behavior / robustness (low severity)
- **`Session.zig:364`** — `closeAllPages` (now the sole `deinit` teardown) skips the `frame_remove` dispatch / `navigation.onRemoveFrame()` that `removePage` did. Benign today (CDP unregisters listeners before session deinit) but a silent contract change for the shared `test_session`.
- **`mcp/tools.zig:1029`** — `session.pages.items[0]` replaced `currentPage().?`; an empty page set now panics on index rather than taking a defined null path (test-only).

### Polish (zig-style §1 + typos)
- Multi-line roadmap comments on one-line fields: `Runner.zig:36`, `Session.zig:413`.
- Typos in added comments: `largel` → `largely` and `Ultilatemy` → `Ultimately` (`Session.zig:414-415`); `the "live" on` → `one` (`Session.zig:387`).
- Missing terminal periods: `Session.zig:775`, `Page.zig:122`.

src/browser/Session.zig
karlseguin: These methods are deprecated and used only as we transition to multi-page. Current users are only single-page and once/if they transition, these functions will be removed.

src/browser/Runner.zig
karlseguin: Runner is going to become more multi-page aware in a follow up PR.

src/browser/tools.zig
karlseguin: Runner (and agent/mcp/tools) are going to become more multi-page aware in a follow up PR.

src/browser/Session.zig
```diff
@@ -332,34 +336,45 @@ fn installNewActivePage(self: *Session, frame_id: u32) !*Frame {
 
 // NOTE: the caller is not the owner of the returned value,
 // the pointer on Frame is just returned as a convenience
```
krichprollsch: Maybe we should update this comment too?

src/browser/Session.zig
```diff
@@ -792,3 +891,32 @@ pub fn nextLoaderId(self: *Session) u32 {
...
+
+    pub fn frame(self: PageHandle) ?*Frame {
+        return &(self.page() orelse return null).frame;
+    }
+
+    pub fn page(self: PageHandle) ?*Page {
+        return self.session.livePage(self.frame_id);
+    }
+
+    pub fn inCommit(self: PageHandle) bool {
```
krichprollsch, line 907: dead code? for future usage?

krichprollsch (approved):

src/browser/Session.zig
karlseguin: `inCommit` ?  Called from https://github.com/lightpanda-io/browser/blob/92d8de94cdcb98142ea39405587c80e09f7bb7d2/src/cdp/CDP.zig#L747

krichprollsch: oh my bad

PR #2790, opened 2026-06-22
webapi: add HashChangeEvent and Window hashchange event

krichprollsch:
Implement the HashChangeEvent interface and fire a `hashchange` event on the window whenever the URL fragment changes without a document reload, mirroring the existing PopStateEvent/onpopstate wiring.


src/browser/webapi/event/HashChangeEvent.zig
```diff
@@ -0,0 +1,103 @@
...
+fn initWithTrusted(arena: Allocator, typ: String, _opts: ?Options, trusted: bool, frame: *Frame) !*HashChangeEvent {
+    const opts = _opts orelse Options{};
+
+    const event = try frame._factory.event(
+        arena,
+        typ,
+        HashChangeEvent{
+            ._proto = undefined,
+            ._old_url = opts.oldURL,
+            ._new_url = opts.newURL,
```
navidemad: **Use-after-free: the option strings aren't duped into the event arena.**

`opts.oldURL`/`opts.newURL` are decoded from JS into the per-call `call_arena` (`String.toSlice` → `self.local.call_arena`, `js/String.zig:51`), which is reset when the constructor call returns (`Caller.deinit`). The event itself lives in the persistent `.tiny` arena, so storing the slices raw leaves them dangling once the call returns.

Both sibling events dupe for exactly this reason — `MessageEvent.zig:78` (`._origin = ... try arena.dupe(u8, str)`) and `ErrorEvent.zig:72-73`. Repro via the new `hashChangeDispatch` test: `new HashChangeEvent('hashchange', {oldURL, newURL})` then a separate `window.dispatchEvent(event)` reuses `call_arena`, so reading `received.oldURL` returns freed/garbage memory (flaky). The trusted Frame/History callers pass arena-stable slices so they don't trip it — only the JS constructor path does.
```suggestion
            ._old_url = try arena.dupe(u8, opts.oldURL),
            ._new_url = try arena.dupe(u8, opts.newURL),
```

src/browser/webapi/History.zig
```diff
@@ -98,6 +99,13 @@ fn goInner(delta: i32, frame: *Frame) !void {
                 const event = (try PopStateEvent.initTrusted(comptime .wrap("popstate"), .{ .state = entry._state.value }, frame)).asEvent();
                 try frame._event_manager.dispatchDirect(target, event, frame.window._on_popstate, .{ .context = "Pop State" });
             }
+
+            // Queue hashchange when traversing between entries that share a
+            // document but differ only by fragment.
+            const is_fragment_change = !std.mem.eql(u8, frame.url, url) and URL.eqlDocument(frame.url, url);
+            if (is_fragment_change) {
+                try frame.queueHashChange(frame.url, url);
```
navidemad: On history traversal `popstate` is dispatched synchronously (line 100) but `hashchange` is queued here. The HTML "update document for history step application" algorithm fires both within the *same* task; splitting `hashchange` into a later scheduler task lets microtasks / other queued tasks run in between — observable to a page whose `popstate` handler schedules work and then checks whether `hashchange` has fired. Relative order (popstate→hashchange) is preserved, so this is a fidelity deviation, not an inversion. (The queued model is correct for the `location.hash`/anchor path in `Frame.zig`; only history traversal fires synchronously per spec.)

Separately: this traversal branch (`history.back()/forward()/go()` across a fragment) has **no test** — `hashchange.html` only drives the `location.hash`/anchor path through `scheduleNavigationWithArena`. The whole branch could be deleted and `make test` would still pass.

src/browser/webapi/History.zig
```diff
@@ -98,6 +99,13 @@ fn goInner(delta: i32, frame: *Frame) !void {
                 const event = (try PopStateEvent.initTrusted(comptime .wrap("popstate"), .{ .state = entry._state.value }, frame)).asEvent();
                 try frame._event_manager.dispatchDirect(target, event, frame.window._on_popstate, .{ .context = "Pop State" });
             }
+
+            // Queue hashchange when traversing between entries that share a
+            // document but differ only by fragment.
+            const is_fragment_change = !std.mem.eql(u8, frame.url, url) and URL.eqlDocument(frame.url, url);
```
navidemad: This re-derives the fragment-vs-document decision from the raw `entry._url`, whereas `navigateInner` independently computes `is_same_document` from the *resolved* URL (`Navigation.zig:292`: `URL.eqlDocument(URL.resolve(arena, frame.url, entry._url), frame.url)`). The two agree only while entries store canonical absolute URLs. If they ever diverge — an entry whose `_url` isn't already absolute/canonical — you'd queue a `hashchange` while `navigateInner` instead schedules a full document reload (event fires for a navigation that didn't happen as a fragment change), or vice-versa. Computing the decision once and threading it into `navigateInner` removes the latent inconsistency.

src/browser/Frame.zig
```diff
@@ -1931,6 +1936,55 @@ pub fn queueElementEvent(self: *Frame, element: *Element.Html, kind: QueuedEvent
...
+    }
+};
+
+pub fn queueHashChange(self: *Frame, old_url: []const u8, new_url: []const u8) !void {
+    const callback = try self._factory.create(HashChangeCallback{
+        .frame = self,
+        .old_url = old_url,
+        .new_url = new_url,
+    });
+    try self.js.scheduler.add(callback, HashChangeCallback.run, 0, .{
```
navidemad, line 1976: If `scheduler.add` fails (e.g. OOM growing the queue), `callback` — already allocated via `_factory.create` just above — is never destroyed and never finalized, leaking the slab allocation (which the debug test runner flags). An `errdefer self._factory.destroy(callback)` before the `add` closes the gap.

Note: `MessagePort.postMessage` (`MessagePort.zig:86-96`) has the same pattern, so this matches existing precedent — low severity, flagging for completeness.

src/browser/Frame.zig
```diff
@@ -1931,6 +1936,55 @@ pub fn queueElementEvent(self: *Frame, element: *Element.Html, kind: QueuedEvent
...
+        const target = frame.window.asEventTarget();
+        if (!frame._event_manager.hasDirectListeners(target, "hashchange", frame.window._on_hashchange)) {
+            return null;
+        }
+
+        const event = (try HashChangeEvent.initTrusted(comptime .wrap("hashchange"), .{
+            .oldURL = self.old_url,
+            .newURL = self.new_url,
+        }, frame)).asEvent();
+        try frame._event_manager.dispatchDirect(target, event, frame.window._on_hashchange, .{ .context = "Hash Change" });
```
navidemad, line 1965: `run` uses `try HashChangeEvent.initTrusted(...)` and `try ...dispatchDirect(...)`, so a dispatch failure propagates to the scheduler and is logged under `log.warn(.scheduler, ...)` with `name=frame.hashChange`. The sibling `MessagePort.PostMessageCallback.run` (`MessagePort.zig:172-180`) instead `catch |err|`s both and logs under `log.err(.dom, ...)`. Per `.claude/rules/zig-style.md` Rule 9 — *"Log scope must match the subsystem"* — catch here and log under the DOM/event scope so the failure is attributed correctly. The `defer destroy` already runs on the error path, so there's no leak either way; this is purely about attribution.

src/browser/Frame.zig
```diff
@@ -1931,6 +1936,55 @@ pub fn queueElementEvent(self: *Frame, element: *Element.Html, kind: QueuedEvent
     }
 }
 
+// Per spec, `hashchange` is fired by queuing a task rather than synchronously,
```
navidemad: `.claude/rules/zig-style.md` Rule 1 — *"Trim comments hard — terse inline remarks, not essays… No multi-paragraph function-header doc blocks"* — and *"If a comment is longer than the code it documents, it's too long."* This 6-line block reads as an essay. Worth trimming to a one-liner.

More importantly, its rationale ("both already point to arena memory that outlives the task… a synchronous reassignment… leaves the captured blocks intact") holds only for the *trusted queue* callers — it's the exact reasoning that masks the constructor-path use-after-free flagged in `HashChangeEvent.zig` (where the slices live in `call_arena`, not a durable arena).

src/browser/tests/event/hashchange.html
```diff
@@ -0,0 +1,83 @@
...
+    newURLs.push(e.newURL);
+    if (newURLs.length === 2) state.resolve();
+  };
+
+  location.hash = '#bar';
+
+  await state.done(() => {
+    testing.expectEqual([root, root + '#foo'], oldURLs);
+    testing.expectEqual([root + '#foo', root + '#bar'], newURLs);
+    window.onhashchange = null;
```
navidemad, line 80: `window.onhashchange` is registered persistently (no `{ once: true }`) and only cleared inside the `state.done` success callback, which runs only after exactly 2 events arrive. If a regression makes the 2nd event never fire (or fire >2×), `state.resolve()` / the cleanup never runs and the global handler leaks past this test — and since this is the last `<script>`, a `hashchange` fired during later teardown/navigation would re-enter a dead test's handler. The sibling `hashChangeDispatch` test (line 35) uses `{ once: true }`; mirroring that (or clearing the handler unconditionally) keeps teardown symmetric.

navidemad:
**Review note (`Navigation.zig` isn't in this diff, so posting here rather than inline).**

Same-document navigations via the **Navigation API** don't fire `hashchange`. `navigateInner`'s `is_same_document` push/replace/traverse branches (`Navigation.zig:306 / 319 / 334`) assign `frame.url = new_url` directly and never call `queueHashChange`, nor do they route through `scheduleNavigationWithArena`'s fragment short-circuit. So after this PR, `navigation.navigate('#x')` / `navigation.back()` / `navigation.traverseTo()` over a fragment-only change update the URL but dispatch **zero** `hashchange` events, while the equivalent `location.hash` and `history.back()` paths now fire one. Per HTML a same-document fragment navigation should fire `hashchange`.

This may be deliberate scope for the PR (the Navigation API is still partial — several `todo: Fire navigate event` markers), but flagging the inconsistency since it leaves `hashchange` coverage uneven across the same logical operation.

src/browser/Frame.zig
karlseguin: I kind of wish Zig had a -DOOM=panic build flag that turned every OOM into a panic and made these OOM-only error functions infallible.  For now, we're inconsistent about how we handle this, but that's largely because it doesn't matter. If we "leak" `callback` on OOM, we'll be crashing soon enough anyways.

src/browser/Frame.zig
karlseguin: We aren't strict on this. I think the driving factor isn't about log subsystems, but rather the API that we want to present. Since the scheduler _does_ log an error, users are free to bubble up to it. The main reason to catch would be to add more context or make a function infallible.

karlseguin (approved):

PR #2795, opened 2026-06-22
webapi: innerText  uses StyleManager's visibility

karlseguin:
innerText now uses the StyleManager's visibility check (isHidden) to skip rendering children which are hidden.

Should allow us to close
https://github.com/lightpanda-io/browser/pull/2741 and https://github.com/lightpanda-io/browser/issues/2734

Note that ~1000 WPT test still fail because of other style-driven behavior, most significantly is `white-space`. The isHidden change only fixes a couple WPT cases, but I think, for end users, it's the more important change (and it's also the simpler for us to implement). Having wrong spacing is usually less of a problem than showing the wrong content.

### DIFF
--- a/src/browser/tests/element/inner.html
+++ b/src/browser/tests/element/inner.html
@@ -185,4 +185,19 @@
   // hidden via a CSS class). A hidden block does not contribute a line break.
   testing.expectEqual("abc", $('#d5').innerText);
   testing.expectEqual("beforeafter", $('#d6').innerText);
+
+  // .innerText read ON a display:none root equals textContent. The own-element
+  // check must keep visible child elements (an ancestor-walk would prune the
+  // <span> while keeping its sibling text nodes, yielding "ac").
+  d1.innerHTML = '<div id=dn style="display:none">a<span>b</span>c</div>';
+  testing.expectEqual("abc", $('#dn').innerText);
+
+  // A display:none <br> contributes no line break.
+  d1.innerHTML = 'a<br style="display:none">b';
+  testing.expectEqual("ab", d1.innerText);
+
+  // Guard: visibility:hidden text stays INCLUDED (the check is display-only),
+  // so a future switch to a visibility-aware check can't silently over-exclude.
+  d1.innerHTML = '<span style="visibility:hidden">X</span>Y';
+  testing.expectEqual("XY", d1.innerText);
 </script>
--- a/src/browser/tests/element/inner.html
+++ b/src/browser/tests/element/inner.html
@@ -11,6 +11,10 @@
 <p id=d4><span>Hello</span>
   <span>World</span></p>
 
+<style>.ihide { display: none }</style>
+<div id=d5>a<span class=ihide>X</span>b<span style="display:none">Y</span>c</div>
+<div id=d6>before<div class=ihide>deep <span>nested</span></div>after</div>
+
 <script id=innerHTML>
   const d1 = $('#d1');
   testing.expectEqual('hello <em>world</em>', d1.innerHTML);
@@ -175,4 +179,10 @@
   testing.expectEqual("This is a\ntext", d2.innerText);
   testing.expectEqual("Hello World", $('#d3').innerText);
   testing.expectEqual("Hello World", $('#d4').innerText);
+
+  // display:none subtrees are excluded from innerText, whether the rule comes
+  // from an inline style or a stylesheet (the common real-world case: content
+  // hidden via a CSS class). A hidden block does not contribute a line break.
+  testing.expectEqual("abc", $('#d5').innerText);
+  testing.expectEqual("beforeafter", $('#d6').innerText);
 </script>
--- a/src/browser/webapi/element/Html.zig
+++ b/src/browser/webapi/element/Html.zig
@@ -1331,9 +1331,6 @@ const InnerTextState = struct {
     // Needed to reach the StyleManager for CSS-driven visibility (display:none).
     frame: *Frame,
 
-    // Tree-local visibility cache shared across the whole walk (see getInnerText).
-    cache: Element.VisibilityCache = .{},
-
     // number of line breaks we've accumulated for the block. Emitted lazily that
     // leading/trailing breaks aren't written and so that we can emit the max
     // requested, which can change as we render children.
@@ -1449,16 +1446,20 @@ fn handleChildElement(
     saw_cell: *bool,
     saw_row: *bool,
 ) !void {
+    // doesn't use StyleManger's isHidden, because we don't care if the element
+    // is hidden through its parent. If you can el.innerText on an element, the
+    // visibility of el.parent doesn't matter. So we only care about visibility
+    // on the element itself and then on each child. This is much simpler too.
+    if (state.frame._style_manager.hasDisplayNone(he.asElement())) {
+        return;
+    }
+
     if (he._type == .br) {
         try state.flushBreaks();
         try state.writer.writeByte('\n');
         state.wrote_any = true;
         state.pre_w = false;
         state.trim_left = true;
-        return;
-    }
-
-    if (state.frame._style_manager.isHidden(he.asElement(), &state.cache, .{})) {
         return;
     }
 
--- a/src/browser/webapi/element/Html.zig
+++ b/src/browser/webapi/element/Html.zig
@@ -216,13 +216,14 @@ pub fn asEventTarget(self: *HtmlElement) *@import("../EventTarget.zig") {
     return self._proto._proto._proto;
 }
 
-pub fn getInnerText(self: *HtmlElement, writer: *std.Io.Writer) !void {
+pub fn getInnerText(self: *HtmlElement, writer: *std.Io.Writer, frame: *Frame) !void {
     const tag = self.asElement().getTag();
     switch (innerTextDisplay(self, tag)) {
         .skip, .replaced => return,
         else => {},
     }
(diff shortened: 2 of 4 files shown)


src/browser/webapi/element/Html.zig
```diff
@@ -1450,6 +1458,10 @@ fn handleChildElement(
         return;
     }
 
+    if (state.frame._style_manager.isHidden(he.asElement(), &state.cache, .{})) {
```
navidemad: **Regression: `isHidden` walks ancestors *above* the innerText root.** `StyleManager.isHidden` (StyleManager.zig:286) follows `parentElement()` to the document root with no stop at `self`. So when `.innerText` is read on an element that is itself inside (or is) a `display:none` subtree, every child *element* is pruned while direct text-node children are still written — yielding garbled/partial text:

```js
// <div style="display:none">a<span>b</span>c</div>
el.innerText // => "ac"   (was "abc"; Chrome returns textContent "abc")
```

Per spec, a not-rendered element's `innerText` getter returns the same value as `textContent`. The same gap hits **detached** subtrees too: a freshly `createElement`'d node still matches the live stylesheet's `.class { display:none }` rule and loses that text even though it isn't rendered.

The ancestor walk is also redundant here: `collectInnerText` descends top-down, so a hidden ancestor's subtree is never reached anyway. An **own-element** check is both correct and cheaper — e.g.

```zig
if (state.frame._style_manager.hasDisplayNone(he.asElement())) {
    return;
}
```

which also lets the `cache` field on `InnerTextState` be removed entirely (no ancestor walk left to amortize).

Two same-line cleanups this also resolves: (a) the check duplicates the UA-stylesheet `display:none` detection `innerTextDisplay` already does via `isHiddenByUaStylesheet` (lines 1523-1524), making that `.skip` arm unreachable for hidden tags on the child path; (b) it bypasses the `Element.checkVisibilityCached` wrapper the sibling tree-walks use (Document.zig:653, SemanticTree.zig:134).

src/cdp/AXNode.zig
```diff
@@ -1011,7 +1011,7 @@ fn writeName(
 
                     if (doc.getElementById(trimmed_id, frame)) |referenced_el| {
                         // Get the text content of the referenced element
-                        try referenced_el.getInnerText(&buf.writer);
+                        try referenced_el.getInnerText(&buf.writer, frame);
```
navidemad, line 1014: **CDP accessible-name regression from the new `display:none` skipping.** When an `aria-labelledby` target is itself `display:none` (the canonical "visually-hidden label" pattern) and its text is wrapped in child elements, `getInnerText` now returns `""`. But `has_content` is set `true` unconditionally because the element merely *exists*, and a trailing space is written (line 1015) — so the computed name becomes `" "` and is returned as `.aria_labelledby` (line 1022), shadowing the whole fallback chain (aria-label / label / name-from-contents).

```html
<button aria-labelledby="lbl"></button>
<span id="lbl" style="display:none"><b>Close</b></span>
```

Name was `"Close "`, now `" "` (blank). Per the accname spec, a directly-referenced node is used even when hidden. `writeLabelInnerText` (line 1265) has the same regression for a `display:none` `<label>`. Fixing the root cause in `Html.getInnerText` (own-element check rather than the ancestor walk — see the Html.zig comment) resolves this path too.

src/browser/webapi/element/Html.zig
navidemad: **A `display:none` `<br>` still emits a newline.** The `<br>` branch just above (lines 1452-1459) flushes breaks, writes `'\n'` and `return`s *before* this `isHidden` check, so:

```js
// <div>a<br style="display:none">b</div>
el.innerText // => "a\nb"  (expected "ab"; the <br> is not rendered)
```

Move the visibility check above the `<br>` handling (or apply it to `<br>` too) so a non-rendered `<br>` contributes no line break.

src/browser/tests/element/inner.html
```diff
@@ -175,4 +179,10 @@
   testing.expectEqual("This is a\ntext", d2.innerText);
   testing.expectEqual("Hello World", $('#d3').innerText);
   testing.expectEqual("Hello World", $('#d4').innerText);
+
+  // display:none subtrees are excluded from innerText, whether the rule comes
+  // from an inline style or a stylesheet (the common real-world case: content
+  // hidden via a CSS class). A hidden block does not contribute a line break.
+  testing.expectEqual("abc", $('#d5').innerText);
```
navidemad, line 186: These tests pin `display:none` *exclusion*, but nothing pins that `visibility:hidden` / `opacity:0` text stays *included* — which the spec requires, and which the `.{}` options (display-only) here correctly preserve. Without a guard test, an easy "consistency" edit to `.{ .check_visibility = true }` would silently over-exclude visible text and CI would stay green (Rule 9: tests should encode *why* the behavior matters). Worth adding a case like `<span style="visibility:hidden">X</span>` asserting `X` is still present in `innerText`.

navidemad:
A patch for the fixes from the review above, in case it saves you a round-trip. It swaps the ancestor-walking `isHidden` for the own-element `hasDisplayNone`, moves the check above the `<br>` branch, and drops the now-unused `cache` field. It also adds a handful of regression assertions to `inner.html`.

Against the #2734 fixtures it still gives `"first\nsecond\ntail"` and `"visible"`, and it clears the three regressions I flagged:

- `.innerText` on a `display:none` root now returns `"abc"` (was `"ac"` — the ancestor walk pruned the child `<span>` but kept its sibling text nodes)
- a `display:none` `<br>` no longer emits a newline: `"ab"` (was `"a\nb"`)
- the `aria-labelledby` → `display:none` label path in `AXNode.zig` stops blanking the accessible name

`zig build test` is green (AXNode + element suites). Based on 23e72cf5.

Save the block below as `fix.patch` and `git apply fix.patch`:

<details><summary>fix.patch</summary>

```diff
diff --git a/src/browser/tests/element/inner.html b/src/browser/tests/element/inner.html
index f8088357..8f660973 100644
--- a/src/browser/tests/element/inner.html
+++ b/src/browser/tests/element/inner.html
@@ -185,4 +185,27 @@
   // hidden via a CSS class). A hidden block does not contribute a line break.
   testing.expectEqual("abc", $('#d5').innerText);
   testing.expectEqual("beforeafter", $('#d6').innerText);
+
+  // #2734 bug A: required line breaks around block-level boxes.
+  d1.innerHTML = '<div>first</div><div>second</div><span>tail</span>';
+  testing.expectEqual("first\nsecond\ntail", d1.innerText);
+
+  // #2734 bug B: display:none descendant excluded.
+  d1.innerHTML = 'visible<span style="display:none">secret</span>';
+  testing.expectEqual("visible", d1.innerText);
+
+  // Review #1: .innerText read ON a display:none root equals textContent.
+  // The own-element check must keep visible child elements (not prune them as
+  // an ancestor-walk would, which yielded "ac").
+  d1.innerHTML = '<div id=dn style="display:none">a<span>b</span>c</div>';
+  testing.expectEqual("abc", $('#dn').innerText);
+
+  // Review #3: a display:none <br> contributes no line break.
+  d1.innerHTML = 'a<br style="display:none">b';
+  testing.expectEqual("ab", d1.innerText);
+
+  // Review #4 guard: visibility:hidden text stays INCLUDED (display-only check),
+  // so a future switch to a visibility-aware check can't silently over-exclude.
+  d1.innerHTML = '<span style="visibility:hidden">X</span>Y';
+  testing.expectEqual("XY", d1.innerText);
 </script>
diff --git a/src/browser/webapi/element/Html.zig b/src/browser/webapi/element/Html.zig
index 4571d71c..97470993 100644
--- a/src/browser/webapi/element/Html.zig
+++ b/src/browser/webapi/element/Html.zig
@@ -1331,9 +1331,6 @@ const InnerTextState = struct {
     // Needed to reach the StyleManager for CSS-driven visibility (display:none).
     frame: *Frame,
 
-    // Tree-local visibility cache shared across the whole walk (see getInnerText).
-    cache: Element.VisibilityCache = .{},
-
     // number of line breaks we've accumulated for the block. Emitted lazily that
     // leading/trailing breaks aren't written and so that we can emit the max
     // requested, which can change as we render children.
@@ -1449,6 +1446,16 @@ fn handleChildElement(
     saw_cell: *bool,
     saw_row: *bool,
 ) !void {
+    // Not rendered (display:none): skip the subtree entirely. Own-element check,
+    // not an ancestor walk: collectInnerText already descends top-down, so a
+    // hidden ancestor's subtree is never reached. The ancestor walk would also
+    // mis-handle .innerText read on a display:none root (it prunes child
+    // elements while keeping their sibling text nodes). Placed above the <br>
+    // branch so a display:none <br> contributes no line break either.
+    if (state.frame._style_manager.hasDisplayNone(he.asElement())) {
+        return;
+    }
+
     if (he._type == .br) {
         try state.flushBreaks();
         try state.writer.writeByte('\n');
@@ -1458,10 +1465,6 @@ fn handleChildElement(
         return;
     }
 
-    if (state.frame._style_manager.isHidden(he.asElement(), &state.cache, .{})) {
-        return;
-    }
-
     switch (innerTextDisplay(he, tag)) {
         // Not rendered: skip the subtree entirely (script/style/metadata/...).
         .skip => {},
```

</details>

src/browser/webapi/element/Html.zig
karlseguin: That is a very good catch. Thanks.

krichprollsch:
> Having wrong spacing is usually less of a problem than showing the wrong content.

:100:

krichprollsch (approved):

PR #2796, opened 2026-06-22
cli: add strip-mode invisible

karlseguin:
Adds a new strip-mode "invisible" which consults the StyleManager to remove hidden elements. However, unlike the general isHidden check in StyleManager which is spec-compliant and will automatically hide a number of tags, e.g. "meta", this check is explicitly against explicit styles. This seems more practical to me, while it might be useful to remove some tags by default, e.g. template, some tags like head, title and meta are too important to strip out.

### DIFF
--- a/src/browser/dump.zig
+++ b/src/browser/dump.zig
@@ -368,3 +368,60 @@ fn writeEscapedByte(input: []const u8, index: usize, writer: *std.Io.Writer) ![]
     }
     return input[index + 1 ..];
 }
+
+const testing = @import("../testing.zig");
+
+// A fresh page per assertion: `with_base` mutates the document (it inserts a
+// <base> element), so reusing one frame across opts would leak that mutation
+// into later dumps.
+fn expectDump(opts: Opts, expected: []const u8) !void {
+    var frame = try testing.pageTest("dump.html", .{});
+    defer testing.reset();
+    defer frame._session.removePage();
+
+    var aw: std.Io.Writer.Allocating = .init(testing.arena_allocator);
+    try root(frame.window._document, opts, &aw.writer, frame);
+    try testing.expectString(expected, aw.written());
+}
+
+test "dump: default dumps the whole document" {
+    try expectDump(.{},
+        \\<!DOCTYPE html>
+        \\<html><head><style>.hidden{display:none}</style><link rel="stylesheet" href="data:text/css,"><script>var a=1;</script></head><body><h1>Title</h1><p class="hidden">secret</p><img><svg></svg><noscript>nojs</noscript><p>visible &amp; well</p></body></html>
+    );
+}
+
+test "dump: with_base injects a <base> element" {
+    try expectDump(.{ .with_base = true },
+        \\<!DOCTYPE html>
+        \\<html><head><base base="http://127.0.0.1:9582/src/browser/tests/dump.html"></base><style>.hidden{display:none}</style><link rel="stylesheet" href="data:text/css,"><script>var a=1;</script></head><body><h1>Title</h1><p class="hidden">secret</p><img><svg></svg><noscript>nojs</noscript><p>visible &amp; well</p></body></html>
+    );
+}
+
+test "dump: strip.js removes script and noscript" {
+    try expectDump(.{ .strip = .{ .js = true } },
+        \\<!DOCTYPE html>
+        \\<html><head><style>.hidden{display:none}</style><link rel="stylesheet" href="data:text/css,"></head><body><h1>Title</h1><p class="hidden">secret</p><img><svg></svg><p>visible &amp; well</p></body></html>
+    );
+}
+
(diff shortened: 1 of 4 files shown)


src/help.zon
```diff
@@ -63,10 +63,11 @@
...
     \\          Defaults to no-strip.
     \\          Allowed values:
-    \\              js   script and link[as=script, rel=preload].
-    \\              ui   Includes img, picture, video, CSS and SVG.
-    \\              css  Includes style and link[rel=stylesheet].
-    \\              full Strip everything.
+    \\              js        Script and link[as=script, rel=preload].
+    \\              ui        Includes img, picture, video, CSS and SVG.
+    \\              css       Includes style and link[rel=stylesheet].
+    \\              invisible Invisible (via CSS) elements.
```
navidemad: `invisible` strips only **author `display:none`** (an inline `style=` or a `<style>`/author-sheet rule), but the help text — and the flag name — read as *all* CSS-invisible elements. These stay in the dump:

- `visibility:hidden` and `opacity:0` — `hasAuthorDisplayNone` calls `isElementHidden(.{ .ua_display_none = false })`, which leaves `check_visibility` / `check_opacity` at their `false` defaults, so only `display` is evaluated (StyleManager.zig:335, CheckVisibilityOptions defaults ~857).
- the `hidden` attribute and closed `<details>` content — these are UA-origin rules, deliberately excluded by `ua_display_none = false`.

Concretely: `--strip-mode invisible` over `<div style="visibility:hidden">secret</div>` (or `<div hidden>secret</div>`) still emits `secret`.

Either narrow the doc to say `display:none`, or widen the implementation to cover the other invisibility sources. Not a blocker — just align the wording with what the code does.

src/browser/dump.zig
```diff
@@ -306,6 +312,10 @@ fn shouldStripElement(el: *const Node.Element, opts: Opts) bool {
         if (std.mem.eql(u8, tag_name, "iframe")) return true;
     }
 
+    if (opts.strip.invisible and frame._style_manager.hasAuthorDisplayNone(el)) {
```
navidemad, line 315: Default-config gap worth a note: `hasAuthorDisplayNone` only sees rules the `StyleManager` actually parsed, and external `<link rel="stylesheet">` sheets are **not** loaded unless `--enable-external-stylesheets` is passed (off by default — Session.zig:105, short-circuited at Frame.zig:1987). So a `.hidden { display:none }` living in an external CSS file won't strip its elements under `--strip-mode invisible`; only inline `style=` and inline `<style>` blocks take effect.

It's the same limitation `getComputedStyle` / `checkVisibility` already carry, so nothing new is broken here — but since this is the first dump mode that *depends* on computed visibility, a one-liner in the help (or PR description) would save users from expecting external-CSS-hidden nodes to disappear.

src/browser/dump.zig
```diff
@@ -265,7 +266,12 @@ fn isVoidElement(el: *const Node.Element) bool {
     };
 }
 
-fn shouldStripElement(el: *const Node.Element, opts: Opts) bool {
+fn shouldStripElement(el: *Node.Element, opts: Opts, frame: *Frame) bool {
+    // Fast path: with no strip flags set (every innerHTML/outerHTML call)
+    if (@as(u4, @bitCast(opts.strip)) == 0) {
```
navidemad, line 271: Minor: the `u4` here hard-codes the width of `Opts.Strip`. Add a 5th strip flag and `packed struct(u4)` becomes `u5`, but this `@bitCast` target won't follow — it's a compile error rather than a silent bug, so it's caught, but it's an avoidable second place to edit. Deriving the width from the type keeps them in lockstep (same `backing_integer` idiom used in cli.zig:548):

```suggestion
    // Fast path: with no strip flags set (every innerHTML/outerHTML call)
    if (@as(@typeInfo(Opts.Strip).@"struct".backing_integer.?, @bitCast(opts.strip)) == 0) {
        return false;
    }
```

src/browser/dump.zig
```diff
@@ -265,7 +266,12 @@ fn isVoidElement(el: *const Node.Element) bool {
     };
 }
 
-fn shouldStripElement(el: *const Node.Element, opts: Opts) bool {
+fn shouldStripElement(el: *Node.Element, opts: Opts, frame: *Frame) bool {
```
navidemad, line 269: No test covers the new behavior. Nothing exercises `shouldStripElement` with `strip.invisible` set, nor the CLI parse of `--strip-mode invisible` or the `full` → `u4` change (`full` now flips the 4th bit via `maxInt(u4)`). A regression in the `ua_display_none` gating or the `@bitCast` fast path would ship silently.

A small test would lock in the intended boundary: dump a document containing an author `display:none` node, a `[hidden]` node, and a `<script>`, then assert the first is stripped while the latter two survive (the UA-vs-author distinction this PR is built around).

krichprollsch (approved):

PR #2797, opened 2026-06-22
fix: work around libcurl Brotli trailing-byte rejection (CURLE_WRITE_ERROR)

krichprollsch:
Some CDNs (e.g. CloudFront/Brave Search) emit Brotli streams with 1+
trailing bytes after the logical end. The Brotli decoder succeeds, but
libcurl's brotli_do_write() (content_encoding.c:439) rejects unconsumed
input bytes as CURLE_WRITE_ERROR — even though all decompressed data was
delivered to the write callback.

Browsers accept such responses, so match that: when CURLE_WRITE_ERROR
arrives but our callback never errored and bytes were received, downgrade
to success.

Reproduced w/ brave test: https://github.com/lightpanda-io/demo/pull/199/changes

### DIFF
--- a/src/browser/HttpClient.zig
+++ b/src/browser/HttpClient.zig
@@ -1057,7 +1057,28 @@ fn isRedirectStatus(status: u16) bool {
 
 fn processOneMessage(self: *Client, msg: http.Handles.MultiMessage, transfer: *Transfer) !bool {
     // State at entry: .inflight = conn (multi just delivered a completion).
-    if (msg.err == null or msg.err.? == error.RecvError) {
+
+    // Workaround for libcurl Brotli trailing-byte rejection.
+    //
+    // Some CDNs (e.g. CloudFront serving Brave Search) emit a Brotli stream
+    // whose compressed payload has 1+ trailing bytes after the logical end.
+    // The Brotli decoder reports BROTLI_DECODER_RESULT_SUCCESS, but libcurl's
+    // brotli_do_write() (content_encoding.c:439) treats any unconsumed input
+    // bytes as CURLE_WRITE_ERROR. All decompressed body data has already been
+    // delivered to our write callback successfully.
+    //
+    // Browsers accept such responses (the decompressed content is valid), so
+    // we match that behavior: when CURLE_WRITE_ERROR arrives but our callback
+    // never errored and bytes were received, treat it as success.
+    const effective_err: ?anyerror = if (msg.err) |err| blk: {
+        if (err == error.WriteError and transfer.res.callback_error == null and transfer.res.bytes_received > 0) {
+            log.debug(.http, "WriteError downgraded", .{ .url = transfer.req.url, .bytes = transfer.res.bytes_received });
+            break :blk null;
+        }
+        break :blk err;
+    } else null;
+
+    if (effective_err == null or effective_err.? == error.RecvError) {
         transfer.detectAuthChallenge(msg.conn);
     }
 
@@ -1090,7 +1111,7 @@ fn processOneMessage(self: *Client, msg: http.Handles.MultiMessage, transfer: *T
     // Handle redirects: reuse the same connection to preserve TCP state.
     // A redirect status without a Location header is not a redirect, it's a
     // final response and falls through so its body is delivered.
-    if (msg.err == null) {
+    if (effective_err == null) {
         const status = try msg.conn.getResponseCode();
         if (isRedirectStatus(status)) {
             if (msg.conn.getResponseHeader("location", 0)) |location| {
@@ -1127,7 +1148,7 @@ fn processOneMessage(self: *Client, msg: http.Handles.MultiMessage, transfer: *T
     // of the response per HTTP/1.1 when there is no Content-Length).
     // We must check this before endTransfer, which may reset the easy handle.
     const is_conn_close_recv = blk: {
-        const err = msg.err orelse break :blk false;
+        const err = effective_err orelse break :blk false;
         if (err != error.RecvError) break :blk false;
         const hdr = msg.conn.getResponseHeader("connection", 0) orelse break :blk true;
         break :blk std.ascii.eqlIgnoreCase(hdr.value, "close");
@@ -1138,8 +1159,8 @@ fn processOneMessage(self: *Client, msg: http.Handles.MultiMessage, transfer: *T
     // from .inflight; nothing to set here.)
     transfer.state = .completing;
 
-    if (msg.err != null and !is_conn_close_recv) {
-        transfer.requestFailed(transfer.res.callback_error orelse msg.err.?, true);
+    if (effective_err != null and !is_conn_close_recv) {
+        transfer.requestFailed(transfer.res.callback_error orelse effective_err.?, true);
         return true;
     }
 


src/browser/HttpClient.zig
```diff
@@ -1057,7 +1057,28 @@ fn isRedirectStatus(status: u16) bool {
...
+    // The Brotli decoder reports BROTLI_DECODER_RESULT_SUCCESS, but libcurl's
+    // brotli_do_write() (content_encoding.c:439) treats any unconsumed input
+    // bytes as CURLE_WRITE_ERROR. All decompressed body data has already been
+    // delivered to our write callback successfully.
+    //
+    // Browsers accept such responses (the decompressed content is valid), so
+    // we match that behavior: when CURLE_WRITE_ERROR arrives but our callback
+    // never errored and bytes were received, treat it as success.
+    const effective_err: ?anyerror = if (msg.err) |err| blk: {
+        if (err == error.WriteError and transfer.res.callback_error == null and transfer.res.bytes_received > 0) {
```
navidemad, line 1074: **Downgrade also fires for aborted transfers, not just the Brotli case.**

`dataCallback`'s abort path returns `http.writefunc_error` *without* setting `callback_error`:

```zig
// ~line 2117
if (transfer.state == .aborted) {
    return http.writefunc_error;
}
```

So a transfer aborted mid-body during an in-perform re-entrant abort (`abortRequests` → `kill` → `detachInPerform`, which sets `state = .aborted` while libcurl keeps draining) surfaces `error.WriteError` with `callback_error == null` and `bytes_received > 0` — exactly the downgrade condition. Instead of the pre-diff short-circuit (`requestFailed` + `return true` at line 1162), it now falls through the success path:

- `headerDoneCallback` runs (line 1170) and writes the aborted response's `Set-Cookie` headers into the cookie jar (lines 2038-2050) — cookies for a request the caller explicitly cancelled.
- The body-replay's `if (transfer.state == .aborted)` guard at line 1187 is now dead: line 1160 already overwrote the state to `.completing`.
- For a 401/407 aborted under Fetch auth interception, `detectAuthChallenge` (now reached via the `effective_err == null` gate at line 1081) plus the auth block can re-park the already-detached transfer as `.intercept_auth` and `return false`, so `processMessages` never deinits it — a zombie holding `interception_layer.intercepted++`.

Excluding the aborted state keeps the workaround scoped to the genuine Brotli case:

```suggestion
        if (err == error.WriteError and transfer.res.callback_error == null and transfer.res.bytes_received > 0 and transfer.state != .aborted) {
```

src/browser/HttpClient.zig
```diff
@@ -1057,7 +1057,28 @@ fn isRedirectStatus(status: u16) bool {
...
+            log.debug(.http, "WriteError downgraded", .{ .url = transfer.req.url, .bytes = transfer.res.bytes_received });
```
navidemad, line 1075: **Gate this `log.debug` behind `if (comptime IS_DEBUG)` to match the file's convention.**

The other three `log.debug` calls in this file are all wrapped in `if (comptime IS_DEBUG)` (lines 1095-1096, 2020-2021, 2190-2191); this new one isn't. `log.debug` is not compiled out in release builds, so per AGENTS.md ("Mirror the patterns in neighboring files") it should be gated too:

```suggestion
            if (comptime IS_DEBUG) {
                log.debug(.http, "WriteError downgraded", .{ .url = transfer.req.url, .bytes = transfer.res.bytes_received });
            }
```

src/browser/HttpClient.zig
```diff
@@ -1057,7 +1057,28 @@ fn isRedirectStatus(status: u16) bool {
 
 fn processOneMessage(self: *Client, msg: http.Handles.MultiMessage, transfer: *Transfer) !bool {
     // State at entry: .inflight = conn (multi just delivered a completion).
-    if (msg.err == null or msg.err.? == error.RecvError) {
+
+    // Workaround for libcurl Brotli trailing-byte rejection.
```
navidemad, line 1061: **House style: trim this comment.** Per `.claude/rules/zig-style.md` Rule 1 — *"No multi-paragraph function-header doc blocks ... If a comment is longer than the code it documents, it's too long"* and *"never a paragraph"* of citations. This is a 12-line, 3-paragraph block (including the `content_encoding.c:439` line-number citation, which will rot on the next libcurl bump) documenting ~7 lines of code. Consider collapsing to a one- or two-line remark naming the cause: libcurl rejects a Brotli stream with trailing bytes as `CURLE_WRITE_ERROR` even though the decode reported success.

src/browser/HttpClient.zig
navidemad: Following up with a cleaner fix than the `state != .aborted` patch above, plus a second gap on this same predicate.

`callback_error == null` is carrying more weight than it looks. Two of our own `writefunc_error` returns leave it null — the aborted path (`HttpClient.zig:2117`) and the `getResponseCode()` failure on first data (`:2081-2084`) — so "no callback error" doesn't actually mean "the write error came from curl rather than us." A provenance flag is tighter: set `res.write_callback_failed = true` at every `return http.writefunc_error` in `dataCallback`, then downgrade only when it's `false`. That cleanly identifies a curl-internal write error (the Brotli case) and excludes both abort and the getResponseCode failure, without needing the separate `state` check.

It also closes a second gap: `bytes_received > 0` rejects a valid **empty** Brotli body followed by trailing bytes. Nothing reaches `dataCallback`, so `bytes_received` stays `0` and the response still fails — the same bug class this PR fixes, just with a zero-length body. With the provenance flag you can drop the `bytes_received > 0` clause entirely (the empty-body case is just `write_callback_failed == false`).

One related cleanup: this guard leans on a wrong comment. `bytes_received` (struct def at `:2171`) is documented as "the response status line, the headers, and the [encoded] body," but it's only ever incremented in the write callback (`:2105`), so it counts **decoded body bytes** only. Worth correcting the comment since the predicate now depends on what it actually measures.

Optional hardening: also require a parsed `Content-Encoding` containing `br` (case-insensitive, comma-tokenized — not an exact `"br"` match, which misses stacked encodings) so a future curl-internal `WriteError` from some other writer can't be silently downgraded.

navidemad:
One thing beyond the inline threads: this change exists to compensate for a third-party decoder/completion quirk, but it ships without a test for that interaction, and `processOneMessage` has no direct coverage today.

A regression fixture worth adding: have the local test server return `Content-Encoding: br`, a correctly-encoded body plus one trailing byte, and a matching encoded `Content-Length`. Assert the exact decoded bytes, exactly one `done` callback, and no error callback. If empty-body support is intended (see the inline thread on the predicate), add a zero-length-payload variant.

If the downgrade decision gets pulled into a small classifier (e.g. `shouldDowngradeWriteError(...)`), a table test covering callback-error / aborted-state / non-Brotli encoding would lock in the intent without needing a live curl handle, matching the pure-helper test pattern already in this file.

karlseguin:
I was hesitant about this at first, but `curl -v -H "Accept-Encoding: br" --compressed "https://search.brave.com/search?q=lightpanda&source=web&="`  also reports a "curl: (23) Failed writing received data to disk/application". Based on similar issues (but not identical) reported to curl (e.g. https://github.com/curl/curl/issues/5200) this is the type of thing curl understandably doesn't want to special case.

karlseguin (approved):

krichprollsch:
@navidemad I get your concerns about this change, let's try to improve in a followup PR

PR #2803, opened 2026-06-23
Add `make wpt` to run Web Platform Tests locally

navidemad:
Running the Web Platform Tests locally currently means following the README by hand: build with `-Dwpt_extensions`, start `./wpt serve`, run the demo repo's `wptrunner`, then kill the server. This adds a `make wpt` target that does all of it, mirroring the `make end2end` convention (both expect the sibling clones in `../wpt` and `../demo`).

```
make wpt                          # whole suite (summary output)
make wpt T=Node-childNodes.html   # one test
```

It checks the prerequisites first (the two clones, the `web-platform.test` hosts entry, the `MANIFEST.json`) and prints the exact command to fix whichever is missing. Then it builds Lightpanda, serves WPT, runs the suite via `-lpd-path` so the runner spawns the binary the same way CI does, and kills the server on exit with a `trap`. The one-time `sudo` for the custom hosts is left to you; the target detects the missing entry rather than running sudo on your behalf.

The runner invocation reuses the two documented forms: `-lpd-path` from the CI job in `.github/workflows/wpt.yml`, and the positional single-test name from the README. I couldn't run it end to end here (neither sibling repo is cloned on this machine, and the hosts file needs sudo), so passing `-lpd-path` together with a positional test name is the one thing worth confirming on a first real run.

The README's WPT section gains a short "Quick local run" note pointing at the target, with the existing manual steps kept below as the detailed reference.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,8 @@ ZIG := zig
 BC := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
 # option test filter make test F="server"
 F=
+# single-test filter for `make wpt` (e.g. make wpt T=Node-childNodes.html)
+T=
 
 # Extra flags forwarded to every `$(ZIG) build` invocation. Most commonly used
 # to point at a prebuilt V8 archive and skip the multi-minute source rebuild:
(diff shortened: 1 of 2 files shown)


navidemad:
Closing in favor of https://github.com/lightpanda-io/browser/pull/2804. We decided the ~3h full WPT suite belongs in CI rather than a local runner; that PR instead surfaces a per-area failure breakdown in the CI job summary so it's clear where to focus. Happy to revisit a local runner later if useful.

krichprollsch:
@navidemad maybe the `demo` repository could integrate an easy way to install and run `wpt`.
So you will save 1 install step.

PR #2804, opened 2026-06-23
ci: summarize WPT results by area in the job summary

navidemad:
The nightly WPT run produces `wpt.json`, but it only lands as a raw artifact plus the Slack diff. There's no at-a-glance view of *which areas* are failing most, which is exactly what you need to decide where to spend effort to pass more tests.

This adds one step to the `run-wpt` job: after `wpt.json` is written, it parses the results and appends a per-area table to the GitHub job summary (`$GITHUB_STEP_SUMMARY`), sorted by failing tests. What runs is unchanged: the same full nightly suite.

Sample from the 2026-06-23 run:

> ## WPT results by area
>
> **6,231 / 37,266 tests passing (16.7%)**, 31,035 failing across 263 areas.
>
> | Area | Tests | Passed | Failed | Pass rate | Subtests |
> |------|------:|-------:|-------:|----------:|---------:|
> | `/html/` | 8,379 | 1,444 | 6,935 | 17.2% | 65,251/100,862 |
> | `/css/` | 7,198 | 920 | 6,278 | 12.8% | 80,624/169,807 |
> | `/referrer-policy/` | 1,390 | 193 | 1,197 | 13.9% | 902/8,469 |
> | `/fetch/` | 906 | 89 | 817 | 9.8% | 2,612/8,363 |
> | `/IndexedDB/` | 813 | 2 | 811 | 0.2% | 32/2,682 |

Top 25 areas show inline; the rest collapse into a `<details>`. The subtests column separates deeply-broken areas (`/IndexedDB/` at 0.2%, almost no subtests passing) from near-misses worth a push (areas where most subtests already pass but the test file doesn't flip green yet).

Notes:

- Inline `python3` rather than a committed script, because `run-wpt` checks out the wpt fork repo, not this one. `python3` is already on that runner for `./wpt serve`.
- `if: always()` plus a `test -s wpt.json` guard and a `try/except` around the parse, so a missing or partial `wpt.json` writes a short note instead of failing the job.

### DIFF
--- a/.github/workflows/wpt.yml
+++ b/.github/workflows/wpt.yml
@@ -116,6 +116,47 @@ jobs:
           ./wptrunner -lpd-path ./lightpanda -json -concurrency 5 -pool 5 --mem-limit 400 > wpt.json
           kill `cat WPT.pid`
 
+      - name: summarize results by area
+        if: always()
+        run: |
+          test -s wpt.json || { echo "No wpt.json produced; nothing to summarize." >> "$GITHUB_STEP_SUMMARY"; exit 0; }
+          python3 - wpt.json >> "$GITHUB_STEP_SUMMARY" <<'PY'
+          import collections, json, sys
+          try:
+              with open(sys.argv[1]) as fh:
+                  results = json.load(fh)
+          except Exception as e:
+              print(f"Could not parse {sys.argv[1]}: {e}")
+              sys.exit(0)
+          total = len(results)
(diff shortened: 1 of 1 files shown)


karlseguin:
Have you seen https://perf.lightpanda.io/wpt which groups them by category? I'm not sure I'd go digging through every action run to see this versus using the web view we already have.

navidemad:
I was not aware of that webview very interesting, way better indeed. No need of this PR then 👍🏻

PR #2809, opened 2026-06-24
uaf: Prevent double-free on BrowserContext deinit with pending interc…

karlseguin:
…eption

Adds a abortParked (to be used instead of abort for a parked transfer) which guards against a double-free by setting the state to .completing BEFORE running the requestFailed callback. Without this, requestFail could itself cause the transfer to be cleared.

Also added a guard to try to catch double transfer deinit's in Debug. Because the memory could get re-used between the first and second free, the lack of failure doesn't prove there is no UAF. But it's cheap to do and debug only.


mookums (approved):
Tests passed and it looks good to me.

PR #2816, opened 2026-06-25
initial implementation for `update` command

nikneym:
This PR adds `update` command which allows checking if newer version of Lightpanda is out. Self-updating is still open for discussion, hence not implemented in this one.

### DIFF
--- a/src/Updater.zig
+++ b/src/Updater.zig
@@ -111,9 +111,43 @@ fn resetConnection(self: *Updater) !void {
     self.conn_err = .none;
 }
 
+const SortContext = struct {
+    object: std.json.ObjectMap,
+
+    pub fn lessThan(ctx: SortContext, a_index: usize, b_index: usize) bool {
+        const keys = ctx.object.keys();
+        const a_version = std.SemanticVersion.parse(keys[a_index]) catch unreachable;
+        const b_version = std.SemanticVersion.parse(keys[b_index]) catch unreachable;
+        return a_version.order(b_version).compare(.gt);
+    }
+};
+
 /// Informs about running version to given `Writer` by desired `Channel`.
-pub fn inform(self: *Updater, channel: Channel, writer: std.Io.Writer) void {
+pub fn inform(self: *Updater, channel: Channel, writer: *std.Io.Writer) !void {
     const versions = try self.getVersions();
+    defer self.resetConnection() catch {};
+
+    switch (channel) {
+        .stable => {
+            var stable = versions.object;
+            // Get rid of `nightly`.
+            lp.assert(stable.swapRemove("nightly"), "Updater.inform: incorrect JSON", .{});
+
+            // Sort and retrieve the newest.
+            stable.sort(SortContext{ .object = stable });
+            // Newest is at the beginning.
+            const newest = stable.values()[0].object;
+            const version = (newest.get("version") orelse return error.IncorrectJson).string;
+            try writer.print("Latest stable Lightpanda version is {s}.\n", .{version});
+            return writer.flush();
+        },
+        .nightly => {
+            const nightly = (versions.object.get("nightly") orelse return error.IncorrectJson).object;
+            const version = (nightly.get("version") orelse return error.IncorrectJson).string;
+            try writer.print("Latest nightly Lightpanda version is {s}.\n", .{version});
+            return writer.flush();
+        },
+    }
 }
 
 /// Invoked by `Connection` when there are body bytes.
--- a/src/Updater.zig
+++ b/src/Updater.zig
@@ -74,25 +74,9 @@ pub fn deinit(self: *Updater) void {
     self.arena.deinit();
 }
 
-// TODO: Update this with actual versions.json when ready.
-const Versions = struct {
-    master: Entry,
-    @"0.16.0": Entry, // Consider this as stable.
-
-    /// Single version record.
-    const Entry = struct {
-        version: []const u8,
-        src: struct {
-            tarball: []const u8,
-            shasum: []const u8,
-            size: u64,
-        },
-    };
-};
-
 /// Returns Lightpanda versions; call `resetConnection` after you're done with
 /// `Versions` to do further requests.
-fn getVersions(self: *Updater) !Versions {
+fn getVersions(self: *Updater) !std.json.Value {
     try self.conn.setURL(VersionsUrl);
     try self.conn.setGetMode();
     try self.conn.setFollowLocation(true);
@@ -109,7 +93,7 @@ fn getVersions(self: *Updater) !Versions {
     }
 
     return std.json.parseFromSliceLeaky(
-        Versions,
+        std.json.Value,
         self.arena.allocator(),
         self.conn_read_buffer.items,
         .{
--- a/src/Updater.zig
+++ b/src/Updater.zig
@@ -119,15 +119,22 @@ pub fn inform(self: *Updater, writer: *std.Io.Writer) !void {
     try libcurl.curl_easy_setopt(conn._easy, .write_data, &ctx);
 
     // Make a request.
-    const status_int = conn.request(&self.config.http_headers) catch |err| {
+    const status_int = conn.perform() catch |err| {
         ctx.err catch |ctx_err| return ctx_err;
         return err;
     };
     const status: std.http.Status = @enumFromInt(status_int);
-    if (status != .ok) {
-        return error.UnexpectedStatus;
+    switch (status) {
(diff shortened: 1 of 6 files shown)


karlseguin:
Did we consider passing the current version to the server and letting it figure things out? Two benefits I can think of:

1 - we can track the version
2 - we can alter the response / logic 

Imagine someone is on version 0.3.3. With this code, all we can do is print "Running an older...". But if the logic is on the server, TODAY we can return "Running an older.." but tomorrow, we can change it to "A critical security vulnerability exists with your version of Lightpanda....".

nikneym:
> Did we consider passing the current version to the server and letting it figure things out?

I believe we didn't, it might make sense if we like to inform the user about why they should update.

krichprollsch:
We didn't. The current uses is a static json. But we can handle a response from the server. Give me a couple of days to prepare it.

krichprollsch:
@nikneym here is the endpoint: `GET https://telemetry.lightpanda.io/v/{version}`
It returns a plain text message for the version check result.

You can get a `400` if the versions format is invalid.

It can failed with `500` in case of issue, or `503` if the server can't get the latest versions list (stored in cache, updated every hours)

:warning: the current latest nightly value is invalid b/c the update [CI failed](https://github.com/lightpanda-io/browser/actions/runs/28417928619/job/84205928073). I merged a [fix](https://github.com/lightpanda-io/browser/pull/2846), that should be ok tomorrow.  


```
$ curl -i https://telemetry.lightpanda.io/v/0.3.3
HTTP/2 200
date: Tue, 30 Jun 2026 13:19:00 GMT
content-length: 33
content-type: text/plain; charset=utf-8
server: Fly/a02c95d2c (2026-06-29)
via: 2 fly.io, 2 fly.io
fly-request-id: 01KWCAXPTVSRQQ2YH14586Z1A5-cdg

You run the latest version 0.3.3
```

```
$ curl -i https://telemetry.lightpanda.io/v/0.3.2
HTTP/2 200
date: Tue, 30 Jun 2026 13:19:13 GMT
content-length: 137
content-type: text/plain; charset=utf-8
server: Fly/a02c95d2c (2026-06-29)
via: 2 fly.io, 2 fly.io
fly-request-id: 01KWCAY3FNQ71WC490YRKHYAAR-cdg

You run an old release (0.3.2), latest is 0.3.3.
Update via one-liner:
curl -fsSL https://pkg.lightpanda.io/install.sh | bash -s "0.3.3"
```

```
$ curl -i https://telemetry.lightpanda.io/v/1.0.0-nightly.7667+fe3e53c3
HTTP/2 200
date: Tue, 30 Jun 2026 13:21:36 GMT
content-length: 55
content-type: text/plain; charset=utf-8
server: Fly/a02c95d2c (2026-06-29)
via: 2 fly.io, 2 fly.io
fly-request-id: 01KWCB2F0JT00E9TKB112N66RA-cdg

You run the latest version 1.0.0-nightly.7667+fe3e53c3
```

src/main.zig
```diff
@@ -68,6 +68,10 @@ fn run(allocator: Allocator, main_arena: Allocator) !void {
             try lp.Agent.listModels(allocator, opts);
             return std.process.cleanExit();
         },
+        .update => {
```
krichprollsch: Are we agree w/ moving that in a `version --check` option instead?

nikneym: I'm fine with the change 👍

nikneym:
Adapted to new version check endpoint and rebased.

src/Updater.zig
```diff
@@ -0,0 +1,140 @@
...
+    defer conn.deinit();
+
+    const allocator = self.arena.allocator();
+    const url = try std.fmt.allocPrintSentinel(
+        allocator,
+        "https://telemetry.lightpanda.io/v/{s}",
+        .{lp.build_config.version},
+        0,
+    );
+    defer allocator.free(url);
```
karlseguin: why?

src/Updater.zig
```diff
@@ -0,0 +1,140 @@
...
+                return 0;
+            };
+            return chunk.len;
+        }
+    };
+
+    try libcurl.curl_easy_setopt(conn._easy, .write_function, ReceiverContext.onBytes);
+    // Set receiver context.
+    var ctx = ReceiverContext{ .allocator = allocator };
+    defer ctx.buffer.deinit(allocator);
```
karlseguin: Same, you know allocator is an arena. You might say there's no harm, but I think it hurts readability. When I read this, I start thinking "ok, we need to manage the memory" and I start tracing allocations and de-allocations. Just a distraction.

src/Updater.zig
```diff
@@ -0,0 +1,140 @@
...
+    if (comptime std.mem.eql(u8, "nightly", kind)) {
+        try writer.print("Running a nightly version of Lightpanda ({s}).\n", .{lp.build_config.version});
+        return writer.flush();
+    }
+
+    const conn = try http.Connection.init(self.x509_store, self.config, null);
+    defer conn.deinit();
+
+    const allocator = self.arena.allocator();
+    const url = try std.fmt.allocPrintSentinel(
```
karlseguin: You already do the versioning at comptime. Does this work??

const url = "https://telemetry.lightpanda.io/v/" ++ lp.build_config.version;

src/Updater.zig
```diff
@@ -0,0 +1,140 @@
...
+
+    // Make a request.
+    const status_int = conn.perform() catch |err| {
+        ctx.err catch |ctx_err| return ctx_err;
+        return err;
+    };
+    const status: std.http.Status = @enumFromInt(status_int);
+    switch (status) {
+        // Write what's received.
+        .ok => try writer.writeAll(ctx.buffer.items),
```
karlseguin: I was wondering if you should just _always_ write the response? Might contain some useful information. Plus, you no longer need to buffer anything, just stream to the writer.  If I'm right about the URL at comptime, the you don't need the allocator/arena anymore. `Updater` can just be a single `update` function (as-is, the struct is mostly ceremony anyways).

src/Updater.zig
nikneym: It does and its more elegant honestly, made the change for it.

src/Updater.zig
nikneym: > I was wondering if you should just always write the response?

Asked @krichprollsch about this; he said he's fine with it, so we now directly drain bytes to stdout.

src/Updater.zig
```diff
@@ -0,0 +1,123 @@
...
+pub fn inform(self: *Updater, writer: *std.Io.Writer) !void {
+    const kind = comptime versioning();
+    if (comptime std.mem.eql(u8, "dev", kind)) {
+        try writer.print("Running a development version of Lightpanda ({s}).\n", .{lp.build_config.version});
+        return writer.flush();
+    }
+    if (comptime std.mem.eql(u8, "nightly", kind)) {
+        try writer.print("Running a nightly version of Lightpanda ({s}).\n", .{lp.build_config.version});
+        return writer.flush();
+    }
```
krichprollsch: I don't want to skip the request in these cases: we handle them in the server and we could change the message in these case in the future.

krichprollsch: Moreover when I tried to remove these conditions, I get a compilation error.
So I think you didn't test in "real" condition.

```
src/sys/libcurl.zig:642:51: error: expected type '*const fn ([*]const u8, usize, usize, *anyopaque) callconv(.c) usize', found '*const fn ([*]const u8, usize, usize, ?*anyopaque) usiz
e'
        .write_function => @as(CurlWriteFunction, value),
                                                  ^~~~~
src/sys/libcurl.zig:642:51: note: pointer type child 'fn ([*]const u8, usize, usize, ?*anyopaque) usize' cannot cast into pointer type child 'fn ([*]const u8, usize, usize, *anyopaque
) callconv(.c) usize'
src/sys/libcurl.zig:642:51: note: calling convention 'auto' cannot cast into calling convention 'x86_64_sysv'
```

src/Updater.zig
```diff
@@ -0,0 +1,123 @@
...
+        return err;
+    };
+    const status: std.http.Status = @enumFromInt(status_int);
+    return switch (status) {
+        // We expect any of those.
+        .ok,
+        .bad_request,
+        .internal_server_error,
+        .service_unavailable,
+        => writer.flush(),
```
karlseguin: We're going to change this to just always flush, right?

src/Updater.zig
nikneym: This got broken after curl refactor PR, updating.

PR #2819, opened 2026-06-26
Telemetry smaller payload

karlseguin:
First, this adds 1 small piece of data to the navigate event: whether the navigate was a page, frame or popup.

It also adds a session id, but as far as I'm concerned, this isn't "new" information, or any new tracking/insight into users. Between the iid and the "run" event, a "session" was always trackable. By giving it an explicit value, we can shrink the size of all other messages.

This change reduces the telemetry payload by ~70% (despite the extra nav field). I'm hoping this might remove a reason some people would consider turning it off.

It hits a /v2/ endpoint. The changes:

1 - A header is the first message in a session and contains all of the static data, as well as a session id
2 - Every event is encoded as an array, [$SID, "event-type, params...]

```
{"sid":"92e98210a141f497","iid":"$UUID","mode":"fetch","os":"macos","arch":"aarch64","version":"$VERSION","proxy":false}
["92e98210a141f497","run"]
["92e98210a141f497","nav",false,"page"]
```

(the driver=cdp field was removed from nav, because it was always cdp).

Some notes for the server:

1 - The server can tell a header from an event based on the first character, '{' vs '['
2 - The SID is 8 bytes, enough to be unique, but not globally unique. The iid + sid + time window is how a events for the same SID can be grouped.
3 - A valid event is always an array of 2+ items, index 0 = SID, index 1 = type

Although positional data isn't expressive, it's still extendable.

The 4 events:

run, no parameters (mostly just used to flush the header now)
["sid", "run"]

// nav, tls, page/popup/iframe
["sid","nav",true,"page"]

// bof, # of lost telemetry events
["sid","bof",42]

// llm, provider, model (nullable)
["sid","llm","anthropic","claude"]

This also adds more aggressive batching.

### DIFF
--- a/src/browser/Frame.zig
+++ b/src/browser/Frame.zig
@@ -652,13 +652,7 @@ pub fn navigate(self: *Frame, request_url: [:0]const u8, opts: NavigateOpts) !vo
             .timestamp = timestamp(.monotonic),
         });
 
-        // Record telemetry for navigation
-        session.browser.app.telemetry.record(.{
-            .navigate = .{
-                .tls = false, // about:blank and blob: are not TLS
-                .proxy = session.browser.app.config.httpProxy() != null,
-            },
-        });
+        self.recordNavigateTelemetry(false);
 
         session.notification.dispatch(.frame_navigated, &.{
             .req_id = req_id,
@@ -738,10 +732,7 @@ pub fn navigate(self: *Frame, request_url: [:0]const u8, opts: NavigateOpts) !vo
     });
 
     // Record telemetry for navigation
-    session.browser.app.telemetry.record(.{ .navigate = .{
-        .tls = std.ascii.startsWithIgnoreCase(self.url, "https://"),
-        .proxy = session.browser.app.config.httpProxy() != null,
-    } });
+    self.recordNavigateTelemetry(std.ascii.startsWithIgnoreCase(self.url, "https://"));
 
     session.navigation._current_navigation_kind = opts.kind;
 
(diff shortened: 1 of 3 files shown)


krichprollsch:
API looks good to me. 
I will start the `v2` implementation on the server side.

karlseguin:
@krichprollsch updated based on your feedback:

```
["8800df58-a5d5-4ca5-9a06-d7691a2a3780","H","fetch",0,"macos","aarch64","1.0.0-dev.7609+88b1bc671"]
["8800df58-a5d5-4ca5-9a06-d7691a2a3780","R"]
["8800df58-a5d5-4ca5-9a06-d7691a2a3780","N",1,"P"]
```

H => Header
R => Run
N => Navigate
B => Buffer Overflow / dropped
L => LLM

Navigate context are
P => Page
O => Open (popup)
I => Iframe

krichprollsch:
Here is the server update: https://github.com/lightpanda-io/telemetry/pull/46

I pushed bf68720855839f71ef95aadcf9247ebca8695201 to force the header for each send + remove iid from each line.
I need the header for each send to be able to correctly fill the records in storage.

So we have something like:

```json
["8800df58-a5d5-4ca5-9a06-d7691a2a3780","H","fetch",0,"macos","aarch64","1.0.0-dev.7609+88b1bc671"]
["R"]
```

```json
["8800df58-a5d5-4ca5-9a06-d7691a2a3780","H","fetch",0,"macos","aarch64","1.0.0-dev.7609+88b1bc671"]
["N", 0, "P"]
["N", 0, "P"]
["N", 0, "P"]
["N", 0, "P"]
["N", 0, "P"]
```

We could remove the `\n`.

karlseguin:
It does make it bigger, but not by too much. 

I made the header smaller, shrinking the mode/arch/os

https://github.com/lightpanda-io/telemetry/pull/47

krichprollsch (approved):

PR #2821, opened 2026-06-26
webapi, fake: Navigator.sendBeacon

karlseguin:
This is a noop implementation of navigator.sendBeacon. It often shows up in the logs. I believe that returning "true" to signal successful queuing is correct as it'll prevent any attempts to fallback. However, I'm less sure that noop'ing the entire thing is better than just implementing it.


nikneym:
I believe having an actual implementation was discussed in the past but I don't think its crucial since it don't return application specific data (like response object) but a boolean.

krichprollsch (approved):

PR #2833, opened 2026-06-28
Fix parent URL resolution for drive-like path segments

### LINKED ISSUE #2794: url-constructor: '..' against a non-special base must pop the last path segment
This is a URL parsing case that fails in Lightpanda; the bug originates from rust-url (https://github.com/servo/rust-url/issues/1130), nice to have it here also.

```html
<script id=wpt_nonspecial_dot_dot_pops_segment>
  // url-constructor: '..' against a non-special base must pop the last path
  // segment (here the "C:/" segment), it is not a Windows drive letter.
  {
    const url = new URL('..', 'abc://x/y/z/C:/');
    testing.expectEqual('abc://x/y/z/', url.href);
  }
</script>
```

sjh9714:
## Summary

Fixes #2794.

`rust-url` treats a terminal `C:/` or `C|/` path segment as a drive root during parent-directory resolution, so `new URL('..', 'abc://x/y/z/C:/')` kept `C:/` instead of popping it. This adds a narrow workaround for parent-directory references against those terminal path segments, while preserving true `file:///C:/` drive-root clamping.

## Changes

- Route URL joins through a helper that adjusts only parent-directory resolution from drive-letter-like terminal path segments.
- Add Rust regression coverage for `..`, `../foo`, `../../foo`, query/fragment variants, and file drive-root behavior.
- Add browser URL coverage for the reported constructor behavior.

## Testing

- `$HOME/.cargo/bin/cargo test --manifest-path src/html5ever/Cargo.toml url_resolve_parent_dir` (failed before the fix, passes after the fix)
- `$HOME/.cargo/bin/cargo test --manifest-path src/html5ever/Cargo.toml`
- `$HOME/.cargo/bin/rustfmt --check src/html5ever/url.rs --edition 2021`
- `$HOME/.cargo/bin/cargo fmt --manifest-path src/html5ever/Cargo.toml -- --check` (fails on existing rustfmt diffs in `src/html5ever/sink.rs` and `src/html5ever/types.rs`)

Not run locally: `make test F="WebApi: parse_with_base"` / `zig fmt --check ./`; Zig 0.15.2 is not installed in this environment, and the local toolchain download from ziglang.org stalled at 74%.

Note: I used Codex while preparing this change, reviewed the final diff, and ran the listed checks locally.

### DIFF
--- a/src/browser/tests/url.html
+++ b/src/browser/tests/url.html
@@ -100,6 +100,16 @@
     testing.expectEqual('http://example.com/a/b/foo', url.toString());
   }
 
+  // Parent directory references pop drive-letter-like path segments when they
+  // are regular URL path segments, not file drive roots.
+  {
+    testing.expectEqual('abc://x/y/z/', new URL('..', 'abc://x/y/z/C:/').href);
+    testing.expectEqual('abc://x/y/z/foo', new URL('../foo', 'abc://x/y/z/C:/').href);
+    testing.expectEqual('abc://x/y/foo', new URL('../../foo', 'abc://x/y/z/C|/').href);
+    testing.expectEqual('file:///C:/', new URL('..', 'file:///C:/').href);
+    testing.expectEqual('file:///x/', new URL('..', 'file:///x/C:/').href);
+  }
+
   // Over-popping ".." clamps at the root.
   {
     const base = 'http://example.com/a/b/c/d';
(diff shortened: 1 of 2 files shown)


nikneym:
Hi, thank you for the PR. We believe that the issue belongs to rust-url rather than this project, hence made a report to their issue tracker.

https://github.com/servo/rust-url/issues/1130

sjh9714:
Thanks for checking this. Since this appears to belong in rust-url rather than lightpanda, and you have opened servo/rust-url#1130 to track it there, I am closing this PR to avoid keeping the fix in the wrong repository.

PR #2835, opened 2026-06-28
script-runtime: support concurrent page navigations

arrufat:
Changes `page.goto` to return a pending Promise instead of blocking synchronously. Introduces a driver loop in `Runtime` to tick the browser and settle pending navigations. This allows parallel gotos and routes tool calls to their respective frames.


src/script/Runtime.zig
```diff
@@ -42,6 +42,10 @@ console_data: [std.enums.values(ConsoleMethod).len]ConsoleData,
 /// clear the live spinner so script output starts on a clean line instead
 /// of colliding with the indicator; the line still goes to stdout/stderr.
 console_observer: ?ConsoleObserver = null,
+/// In-flight async `goto`s; emptied before each `runSource` returns.
+pending_gotos: std.ArrayListUnmanaged(PendingGoto),
```
karlseguin:
```suggestion
pending_gotos: std.ArrayList(PendingGoto),
```

src/browser/tools.zig
```diff
@@ -1826,24 +1826,56 @@ fn ensurePage(session: *lp.Session, registry: *CDPNode.Registry, url: ?[:0]const
...
     _ = page.frame().?.navigate(url, .{
         .reason = .address_bar,
         .kind = .{ .push = null },
     }) catch return ToolError.NavigationFailed;
+    return page;
+}
+
+/// Start a navigation without waiting for it to load, so the agent script Runtime
+/// can drive several gotos concurrently. `receiver_frame_id`, when set, is the
+/// page a re-goto on the same `Page` object replaces — only that one is closed.
```
karlseguin: > `receiver_frame_id`, when set, is the page a re-goto on the same `Page` object replaces — only that one is closed.

but it resets the registry for everything. I'm not sure if that's an issue, just pointing it out.

karlseguin (approved):
This is good as-is....but.....

The hope is to make Session not have any concept of special pages (e.g. items.pages[0]). I'm hoping we can remove `primaryPage` and `currentFrame` and have callers work on frame_id only.  We don't have to get there today, but it would be nice. Superficially tracing through it, it seems that instead of setting setToolFrameOverride, you could call `callTo` with the `frame_id` and then pass the `frame_id` to `call` as a first step.

src/browser/tools.zig
arrufat: I introduced `Registry.resetFrame` for that, can you check?

karlseguin (approved):

PR #2836, opened 2026-06-28
fix: getComputedStyle resolves inline style= declarations

### LINKED ISSUE #2733: css: getComputedStyle returns "" for all properties except display/visibility (stylesheet and inline declarations both dropped)
## Summary

`getComputedStyle()` returns the empty string for every CSS property except `display` and `visibility`. This affects declarations from **both** sources: stylesheet rules never reach the computed style, and even inline `style=` declarations — which are correctly readable through `el.style` — come back as `""` through the computed-style object. Per [CSSOM §resolved values](https://drafts.csswg.org/cssom/#resolved-values), `getComputedStyle()` must return the resolved value of every supported property; Chrome returns `"uppercase"` / `"700"` for the fixture below, and returns the property's initial value (`"none"`, `"400"`) when nothing matches.

Downstream CDP clients use `getComputedStyle` for far more than visibility: text-transform-aware text matching, font-weight assertions, color extraction, accessibility tree heuristics. With every property except `display`/`visibility` resolving to `""`, all of those silently degrade.

## Today's behavior

A CDP client evaluating `getComputedStyle(el).textTransform` via `Runtime.evaluate` gets `""` — whether the declaration comes from a stylesheet rule (`#styled { text-transform: uppercase }`) or from the element's own `style` attribute. The same inline declaration reads fine via `el.style.textTransform`. Only `display` and `visibility` are special-cased to consult the stylesheet cascade.

```mermaid
sequenceDiagram
    participant LP as Lightpanda
    Client->>LP: Runtime.evaluate getComputedStyle textTransform
    LP->>SM: computed branch consults cascade
    SM-->>LP: only display and visibility are resolved
    LP-->>Client: empty string for every other property
    Note over LP,SM: src/browser/webapi/css/CSSStyleDeclaration.zig getPropertyValue
```

## Expected behavior

Per [CSSOM §dom-window-getcomputedstyle](https://drafts.csswg.org/cssom/#dom-window-getcomputedstyle), the returned declaration block exposes the resolved value of supported properties. Chrome resolves the full cascade: author stylesheet rules by specificity, then inline declarations, then inheritance/initial values.

```mermaid
sequenceDiagram
    participant LP as Lightpanda
    Client->>LP: Runtime.evaluate getComputedStyle textTransform
    LP->>SM: computed branch consults cascade
    SM-->>LP: matched declarations resolved by specificity
    LP-->>Client: uppercase as in Chrome
```

## Scope question (before any PR)

The current architecture explains the gap: `StyleManager` tracks matched rules as three booleans (`display_none` / `visibility_hidden` / `opacity_zero`) rather than retaining the matched declarations, and `CSSStyleDeclaration.getPropertyValue`'s `_is_computed` branch special-cases exactly those two properties. Generalizing means deciding how much cascade to keep:

1. **Per-property extension** — keep the boolean-style fast path, add the handful of properties downstream clients consult most (`text-transform`, `font-weight`, `color`, `background-color`, `opacity`, ...), resolved from matched rules by specificity. Cheap, but each property is a new special case.
2. **General declaration retention** — `StyleManager` stores the full declaration list per matched rule; `getPropertyValue` resolves any property by walking matches in specificity/source order, falling back to inline, then to a small initial-value table. Bigger change, no per-property whack-a-mole.

Also worth deciding: whether unmatched properties should keep returning `""` or synthesize initial values (`"none"`, `"400"`) as Chrome does — the latter is what the spec's resolved-value text implies.

Happy to implement either direction — filing this first to agree on the approach before writing code.

## Reproducer

`repro.html`:
```html
<!DOCTYPE html>
<html>
<head>
  <title>computed-style cascade repro</title>
  <style>
    #styled { text-transform: uppercase; font-weight: 700; }
  </style>
</head>
<body>
  <p id="styled">hello</p>
  <p id="inline" style="text-transform: uppercase">hello (inline control)</p>
</body>
</html>
```

`repro.sh`:
```bash
#!/usr/bin/env bash
# Asserts that getComputedStyle resolves properties from stylesheet rules
# (cascade) and inline declarations, not only display/visibility.
#
# Run:           ./repro.sh
# Expected today (bug): computed textTransform: ""            exit 1
# Expected after fix:   computed textTransform: "uppercase"   exit 0

set -euo pipefail

CDP_PORT="${CDP_PORT:-9222}"

PIDS=()
cleanup() {
}
trap cleanup EXIT

cd "$REPRO_DIR"
PIDS+=("$!")

PIDS+=("$!")

for _ in $(seq 1 50); do
  sleep 0.1
done

fi

```

`probe.js`:
```js
// Asserts getComputedStyle resolves cascade properties (from stylesheet
// rules and inline declarations). Today it returns "" for everything except
// display/visibility. Control: the inline declaration is readable via
// el.style, so the value exists — only the computed-style path drops it.

(async () => {
  const c = await connect();

  const control = await c.eval(`document.getElementById('inline').style.textTransform`);
  const fromSheet = await c.eval(`getComputedStyle(document.getElementById('styled')).textTransform`);
  const fromInline = await c.eval(`getComputedStyle(document.getElementById('inline')).textTransform`);
  const display = await c.eval(`getComputedStyle(document.getElementById('styled')).display`);

  console.log(`el.style.textTransform (control):   ${JSON.stringify(control)} (expected "uppercase")`);
  console.log(`computed textTransform, stylesheet: ${JSON.stringify(fromSheet)} (expected "uppercase")`);
  console.log(`computed textTransform, inline:     ${JSON.stringify(fromInline)} (expected "uppercase")`);
  console.log(`computed display, stylesheet:       ${JSON.stringify(display)} (expected "block" — already works)`);

  c.close();
  if (control !== 'uppercase') {
    console.log('BROKEN HARNESS: the inline declaration is not readable via el.style.');
    process.exit(2);
  }
  if (fromSheet !== 'uppercase' || fromInline !== 'uppercase') {
    console.log('FAIL: getComputedStyle returns "" for cascade properties.');
    process.exit(1);
  }
  console.log('PASS: getComputedStyle resolves cascade properties.');
  process.exit(0);
```

<details>

```js
// than this is long).

  let nextId = 1;
  const pending = new Map();
  const handlers = new Map();
    pending.clear();
  };
  ws.on('message', (raw) => {
      pending.delete(m.id);
      else resolve(m.result);
    } else if (m.method) {
    }
  });
  return {
      const id = nextId++;
      });
    },
    ready() {
        ws.on('open', resolve);
      });
    },
    close() { ws.close(); },
  };
}

  await client.ready();

  return {
    sessionId,
    close: () => client.close(),
    async eval(expression) {
      }, sessionId);
    },
    async navigate(url2) {
      let loaded = false;
        try {
          }, sessionId);
        } catch (e) {
        }
      }
    },
  };
}

module.exports = { connect };
```

</details>

**Run**: `bash repro.sh`
- Today (nightly `1.0.0-nightly.6736+63665f81`):
  ```
  el.style.textTransform (control):   "uppercase" (expected "uppercase")
  computed textTransform, stylesheet: "" (expected "uppercase")
  computed textTransform, inline:     "" (expected "uppercase")
  computed display, stylesheet:       "block" (expected "block" — already works)
  FAIL: getComputedStyle returns "" for cascade properties.
  ```
  exits 1.
- Expected: all four lines resolve, prints `PASS: getComputedStyle resolves cascade properties.` and exits 0.

Chrome control (same fixture, chrome-headless-shell): `textTransform: "uppercase"`, `fontWeight: "700"` for the stylesheet element.

## Likely fix location

- `src/browser/webapi/css/CSSStyleDeclaration.zig` — `getPropertyValue`: the `_is_computed` branch special-cases only `display`/`visibility`.
- `src/browser/StyleManager.zig` — the cascade walk reduces matched rules to three booleans (`display_none`/`visibility_hidden`/`opacity_zero`); resolving other properties needs the matched declarations retained in some form (see scope question above).

## Environment

- Lightpanda: nightly `1.0.0-nightly.6736+63665f81` (macOS aarch64); special-casing confirmed in source on `main`
- OS: darwin (macOS 15, aarch64)
- CDP client: raw `ws` (Node 26)

mvanhorn:
## Summary

`getComputedStyle(el)` now returns the value of properties set through the element's inline `style=` attribute, instead of returning an empty string for everything except `display` and `visibility`.

## Why this matters

`CSSStyleDeclaration.init` skips parsing the inline `style=` attribute when the declaration is in computed mode, and `getPropertyValue` only special-cased `display`/`visibility` before falling through to defaults or `""`. So `getComputedStyle(el).getPropertyValue('text-transform')` came back `""` even though `el.style.textTransform` read the inline value correctly. Per the CSSOM resolved-value rules these should agree for inline declarations. See `src/browser/webapi/css/CSSStyleDeclaration.zig:79` (the computed branch) and `:47` (the `if (!is_computed)` guard in `init`).

This change is scoped to the inline `style=` path only. Resolving stylesheet-cascade rules is left out, since that is the contested part maintainers floated gating behind an opt-in.

The lookup reads the element's parsed inline declarations directly and honors `!important` precedence, so a normal declaration does not override an earlier important one in the same block.

## Testing

Added a `computedStyleInline` block to `src/browser/tests/element/styles.html` covering: inline `text-transform`/`font-weight`/`color` resolving through `getComputedStyle`, the non-computed `el.style` path staying unchanged, a property with no inline declaration still returning `""`, `display` still resolving to its default, and a duplicate `color: red !important; color: blue` resolving to `red`.

`zig fmt --check ./` passes on the pinned toolchain version.

Fixes #2733

### DIFF
--- a/src/browser/webapi/css/CSSStyleDeclaration.zig
+++ b/src/browser/webapi/css/CSSStyleDeclaration.zig
@@ -243,11 +243,18 @@ fn findInlineValue(element: *const Element, normalized_name: []const u8, frame: 
     const attr_value = element.getAttributeSafe(comptime .wrap("style")) orelse return null;
     var it = CssParser.parseDeclarationsList(attr_value);
     var result: ?[]const u8 = null;
+    var result_important = false;
     while (it.next()) |declaration| {
-        if (std.ascii.eqlIgnoreCase(declaration.name, normalized_name)) {
-            // Keep the last matching declaration, matching CSS declaration order.
-            result = normalizePropertyValue(frame.call_arena, normalized_name, declaration.value) catch declaration.value;
+        if (!std.ascii.eqlIgnoreCase(declaration.name, normalized_name)) {
+            continue;
         }
+        // A later declaration wins, except a normal one never overrides an
+        // earlier !important declaration (CSS cascade precedence).
+        if (result_important and !declaration.important) {
+            continue;
+        }
+        result = normalizePropertyValue(frame.call_arena, normalized_name, declaration.value) catch declaration.value;
+        result_important = declaration.important;
     }
     return result;
 }
--- a/src/browser/webapi/css/CSSStyleDeclaration.zig
+++ b/src/browser/webapi/css/CSSStyleDeclaration.zig
@@ -49,7 +49,7 @@ pub fn init(element: ?*Element, is_computed: bool, frame: *Frame) !*CSSStyleDecl
             if (el.getAttributeSafe(comptime .wrap("style"))) |attr_value| {
                 var it = CssParser.parseDeclarationsList(attr_value);
                 while (it.next()) |declaration| {
-                    try self.setPropertyImpl(declaration.name, declaration.value, declaration.important, frame);
+                    try self.applyParsedDeclaration(declaration, frame);
                 }
             }
         }
@@ -97,7 +97,9 @@ pub fn getPropertyValue(self: *const CSSStyleDeclaration, property_name: []const
         // Only return default values for computed styles
         if (self._is_computed) {
             if (self._element) |element| {
-                if (findInlineValue(element, normalized, frame)) |value| return value;
+                // Resolve inline `style=` declarations through the element's
+                // parsed inline style, so computed values match `el.style`.
+                if (frame._style_manager.inlineStyleValue(element, wrapped)) |value| return value;
             }
             return getDefaultPropertyValue(self, wrapped);
         }
@@ -125,6 +127,19 @@ pub fn setProperty(self: *CSSStyleDeclaration, property_name: []const u8, value:
     try self.setPropertyImpl(property_name, value, important, frame);
 
     try self.syncStyleAttribute(frame);
+}
+
+// Apply one declaration parsed from a `style=` block. Unlike the imperative
+// setProperty path, within a single declaration block a normal declaration must
+// not override an earlier !important one (CSS cascade precedence).
+fn applyParsedDeclaration(self: *CSSStyleDeclaration, declaration: CssParser.Declaration, frame: *Frame) !void {
+    if (!declaration.important) {
+        const normalized = normalizePropertyName(declaration.name, &frame.buf);
+        if (self.findProperty(.wrap(normalized))) |existing| {
+            if (existing._important) return;
+        }
+    }
+    try self.setPropertyImpl(declaration.name, declaration.value, declaration.important, frame);
 }
 
 fn setPropertyImpl(self: *CSSStyleDeclaration, property_name: []const u8, value: []const u8, important: bool, frame: *Frame) !void {
@@ -210,7 +225,7 @@ pub fn setCssText(self: *CSSStyleDeclaration, text: []const u8, frame: *Frame) !
     // Parse and set new properties
     var it = CssParser.parseDeclarationsList(text);
     while (it.next()) |declaration| {
-        try self.setPropertyImpl(declaration.name, declaration.value, declaration.important, frame);
+        try self.applyParsedDeclaration(declaration, frame);
     }
     try self.syncStyleAttribute(frame);
 }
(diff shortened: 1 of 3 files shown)


krichprollsch:
I added 0614795fc5b54213c7f3559aa5a7ab524aa30a5a to move logic into the style manager.

src/browser/webapi/css/CSSStyleDeclaration.zig
```diff
@@ -124,6 +129,19 @@ pub fn setProperty(self: *CSSStyleDeclaration, property_name: []const u8, value:
     try self.syncStyleAttribute(frame);
 }
 
+// Apply one declaration parsed from a `style=` block. Unlike the imperative
+// setProperty path, within a single declaration block a normal declaration must
+// not override an earlier !important one (CSS cascade precedence).
```
arrufat:
```suggestion
/// Apply one declaration parsed from a `style=` block. Unlike the imperative
/// setProperty path, within a single declaration block a normal declaration must
/// not override an earlier !important one (CSS cascade precedence).
```

arrufat (approved):

PR #2841, opened 2026-06-29
webapi: IndexedDB base

karlseguin:
Add IndexedDB  support

This API is large and introduces new interaction points (e.g. with the 
`js.Context` and events) and new lifetime concerns.

There are two design decisions worth noting:
1 - Only in-memory [sqlite] databases are currently supported. This simplifies
    thing like migrations (don't need) and expiring/reclaiming space (don't 
    need).
2 - The store is tied to the Session, so shared between pages/frames but not
    CDP connections

(IndexedDB (IDB) is accessible on both Frame and WorkerGlobalScope, butI'll just 
use "Frame" here for simplicity.)

Besides the API itself, there are two complicated aspects to the design. Because
of the in-memory database, each database can only have 1 open connection. This
is represented by Engine.zig, and Session maintains an origin->Engine lookup.
The Engine must keep a queue of transactions and execute them serially. This
 causes a lifetime issue as Transactions are Frame-bound. On Context.deinit, the
 Engine is notified and removes related transactions. It's pretty minor, but
 it's one of the few cases where the JS engine reaches into the WebAPI.

The other complexity is the memory management. IDB is callback-heavy, with
actions being done and results coming on the next tick. Even moderate IDB usage
has potential to require many allocations (including dupes out of sqlite), so
page allocators aren't suitable. On the flip side, it's a small but not trivial
graph with bi-directional node that makes fine memory tracking complicated. 
In the end, some things remain tied to the Page while everything else is tied
to an IDBTransaction. The IDBTransaction has an arena, and participates in 
reference counting + v8 finalization. Unlike any other type before, the
IDBTransaction tracks its own v8::Globals. In this way (and only in this way)
it acts as a js.Context. When the IDBTransaction is freed, it releases all of
its global.

The IDBTransaction has the most complicated RC of any type in the system. Every
type created from an IDBTransaction forwards the v8 acquire/release to it 
(like an iterator does to the underlying list). But the IDBTransaction itself 
can be internally referenced from either a scheduled task OR from being parked 
in the engine (OR BOTH!). So it's like fetch+response, but more complicated.

Some smaller notes:

1. An exception in a callback requires a transaction to be aborted. This required
some small changes to Event/EventManager.
2. Dynamic SQL is used extensively, but only with compile-time string literals.
3. Currently no limit on how big a database can get. Might be worth adding command line argument in a follow up PR


src/Config.zig
```diff
@@ -85,36 +85,7 @@ fn logLevelValidator(_: Allocator, args: *std.process.ArgIterator) !?log.Level {
...
...ame = "http_max_response_size", .type = ?usize }, .{ .name = "ws_max_concurrent", .type = ?u8 }, .{ .name = "insecure_disable_tls_host_verification", .type = bool }, .{ .name = "log_level", .type = ?log.Level, .validator = logLevelValidator }, .{ .name = "log_format", .type = ?log.Format }, .{ .name = "log_filter_scopes", .type = log.FilterRule, .multiple = true, .validator = logFilterScopesValidator }, .{ .name = "user_agent_suffix", .type = ?[]const u8 }, .{ .name = "http_cache_dir", .type = ?[]const u8 }, .{ .name = "web_bot_auth_key_file", .type = ?[]const u8 }, .{ .name = "web_bot_auth_keyid", .type = ?[]const u8 }, .{ .name = "web_bot_auth_domain", .type = ?[]const u8 }, .{ .name = "user_agent", .type = ?[]const u8 }, .{ .name = "block_private_networks", .type = bool }, .{ .name = "block_cidrs", .type = ?[]const u8 }, .{ .name = "cookie", .type = ?[]const u8 }, .{ .name = "cookie_jar", .type = ?[]const u8 }, .{ .name = "storage_engine", .type = ?Storage.EngineType }, .{ .name = "storage_sqlite_path", .type = ?[:0]const u8 }, .{ .name = "disable_subframes", .type = bool }, .{ .name = "disable_workers", .type = bool }, .{ .name = "enable_external_stylesheets", .type = bool } };
```
arrufat: Isn't it more readable with an entry per line?
Maybe you deleted a comma in the last entry, and it got mangled?

karlseguin: Ya..I had added  a new option, but then took it out...and the trailing comma went with it.

PR #2849, opened 2026-06-30
perf(css): cache parsed selectors per frame

arrufat:
## What

`querySelector` / `querySelectorAll` / `matches` / `closest` re-parsed the
selector string on **every** call — no cache.

This adds `Selector.Cache`, a parsed-selector cache **on the Browser** (a parsed
selector references no Frame/Context, so entries are shared across the browser's
pages and survive navigation). It's a `StringArrayHashMap` with **per-entry
arenas** (so eviction can free an individual entry, which a shared arena can't)
and **FIFO eviction** of the oldest entry past a capacity — so it's bounded.

The public entry points cache. SelectorPath's synthesized one-off candidates
(from `nodeDetails`) use new `*Uncached` variants that parse into a transient
arena, so they never churn the cache. StyleManager rule parsing is untouched.

## Impact

The win is on **repeated identical single-element matches** — event-delegation /
framework polling (`e.target.closest('.foo')`, `el.matches(sel)` in a loop) —
where parsing is a real fraction of the call. On a generated ~14k-element page,
100k repeated `matches()` with the same selector:

| | ns / call |
|---|---:|
| before | 4,682 |
| after | **463** |
| | **10.1× faster** |

Honest scope: `querySelectorAll` and `nodeDetails`/`SelectorPath` are **unchanged**
— there a full-document tree walk dominates and parsing is a rounding error, so
the cache doesn't help those paths.

## Notes

- Lock-free: a Browser is one-per-connection / single-session.
- App-level cross-thread sharing was considered but skipped — it needs locking on
  the hot `matches()`/`querySelector` read path and a parse is sub-µs, so it'd
  likely erase the win. Left as a possible follow-up.

### DIFF
--- a/src/browser/webapi/selector/Selector.zig
+++ b/src/browser/webapi/selector/Selector.zig
@@ -51,22 +51,43 @@ pub fn parseLeaky(arena: Allocator, input: []const u8) !Parsed {
     return .{ .selectors = try Parser.parseList(arena, input) };
 }
 
+// SelectorPath synthesizes many one-off selectors, so past the cap we parse
+// into a per-call arena rather than grow the frame-lifetime cache unbounded.
+const max_cache_entries = 1024;
+
+/// The parsed AST borrows slices of its input, so on a miss the key is duped
+/// into `frame.arena` and parsed against that copy; both share the frame's
+/// lifetime. `fallback` backs the parse once the cache is full.
+fn cachedParse(frame: *Frame, fallback: Allocator, input: []const u8) ![]const Selector {
+    if (input.len == 0) {
+        return error.SyntaxError;
+    }
+    if (frame._selector_cache.get(input)) |selectors| {
+        return selectors;
+    }
+    if (frame._selector_cache.count() >= max_cache_entries) {
+        return Parser.parseList(fallback, input);
+    }
+
+    const arena = frame.arena;
+    const owned = try arena.dupe(u8, input);
+    const selectors = try Parser.parseList(arena, owned);
+    try frame._selector_cache.put(arena, owned, selectors);
+    return selectors;
+}
+
 pub fn querySelector(root: *Node, input: []const u8, frame: *Frame) !?*Node.Element {
-    const parsed = try parseLeaky(frame.call_arena, input);
+    const parsed = Parsed{ .selectors = try cachedParse(frame, frame.call_arena, input) };
     return parsed.query(root, frame);
 }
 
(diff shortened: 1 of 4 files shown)


src/browser/webapi/selector/Selector.zig
```diff
@@ -51,22 +51,43 @@ pub fn parseLeaky(arena: Allocator, input: []const u8) !Parsed {
     return .{ .selectors = try Parser.parseList(arena, input) };
 }
 
+// SelectorPath synthesizes many one-off selectors, so past the cap we parse
+// into a per-call arena rather than grow the frame-lifetime cache unbounded.
+const max_cache_entries = 1024;
```
mookums: Just curious but any reason you settled on 1024 as the max cache entry count?

arrufat: 1024 was just a round-number guess, not measured, so I removed it. The cap was only ever there to stop SelectorPath (which synthesizes a unique selector per node in nodeDetails) from filling a frame-lifetime cache with strings that never repeat. 

Instead of bounding the cache, we now split the paths: the public querySelector/querySelectorAll/matches/closest entry points cache (selectors that genuinely recur: page scripts, waitForSelector), and SelectorPath's one-off candidates use new *Uncached variants that parse into a transient arena.
The cache now only ever holds reusable selectors, so it needs no size limit.

karlseguin: Just because something uses `querySelector` doesn't mean that it's re-used. Also,`Runner.waitForSelector` doesn't re-parse the selector.

I like this idea, but I think it (a) has to be bound and (b) don't love locking it in to the first 1024.

I'm tempted to add a real cache with an eviction policy. I also think the parsed result is independent of all transient instances (Frames, v8::Context, ...) and could live on the Browser or even the App and be shared across threads.

Not caching things we're _sure_ won't be re-used is fine.  For merging this more or less as-is, what do you think of:

1 - moving it to the browser (assuming I'm right that nothing Selector points back to Frame or anything)
2 - on full, evict some random keys, or switch to an ArrayHashMap and evict the first N keys

?

arrufat: Fixed, thank you for the review :)

karlseguin:
I captured the hit/total ratio on github and reddit, and the hit ratio was crazy high in both cases. I forget the exact number, but like 9800/9900 in github, and 3200/3300  on reddit.

PR #2851, opened 2026-07-01
networking: prefer `X509_STORE` instead of `ca_blob`

nikneym:
This PR refactors our certificate loading logic for libcurl. Currently we're using a CA blob that gets parsed for each connection. What this does instead is to modify `SSL_CTX` that come from BoringSSL directly to share a cert store that's being created on process start; since its shared, no copy happens in the path also.

Breaks #2816 but I prefer this being merged earlier so I can rebase for the other one.

### DIFF
--- a/src/sys/libcrypto.zig
+++ b/src/sys/libcrypto.zig
@@ -343,6 +343,26 @@ pub extern fn EVP_MD_CTX_free(ctx: ?*EVP_MD_CTX) void;
 pub const struct_evp_md_ctx_st = opaque {};
 pub const EVP_MD_CTX = struct_evp_md_ctx_st;
 
+pub extern fn X509_free(x509: ?*X509) void;
+pub extern fn d2i_X509(out: [*c]?*X509, inp: *[*]const u8, len: c_long) ?*X509;
+
+//pub const struct_x509_store_ctx_st = opaque {};
+//pub const X509_STORE_CTX = struct_x509_store_ctx_st;
+pub const struct_x509_store_st = opaque {};
+pub const X509_STORE = struct_x509_store_st;
+
+pub extern fn X509_STORE_new() ?*X509_STORE;
+pub extern fn X509_STORE_free(store: *X509_STORE) void;
+pub extern fn X509_STORE_add_cert(store: ?*X509_STORE, x: ?*X509) c_int;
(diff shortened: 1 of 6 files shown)


src/sys/libcrypto.zig
```diff
@@ -343,6 +343,26 @@ pub extern fn EVP_MD_CTX_free(ctx: ?*EVP_MD_CTX) void;
...
+
+//pub const struct_x509_store_ctx_st = opaque {};
+//pub const X509_STORE_CTX = struct_x509_store_ctx_st;
+pub const struct_x509_store_st = opaque {};
+pub const X509_STORE = struct_x509_store_st;
+
+pub extern fn X509_STORE_new() ?*X509_STORE;
+pub extern fn X509_STORE_free(store: *X509_STORE) void;
+pub extern fn X509_STORE_add_cert(store: ?*X509_STORE, x: ?*X509) c_int;
+pub extern fn X509_STORE_add_crl(store: ?*X509_STORE, x: ?*X509_CRL) c_int;
```
karlseguin: Not used?

src/sys/libcrypto.zig
```diff
@@ -343,6 +343,26 @@ pub extern fn EVP_MD_CTX_free(ctx: ?*EVP_MD_CTX) void;
...
+
+pub const struct_ssl_ctx_st = opaque {};
+pub const SSL_CTX = struct_ssl_ctx_st;
+/// SSL_CTX takes ownership.
+pub extern fn SSL_CTX_set0_verify_cert_store(ctx: ?*SSL_CTX, store: ?*X509_STORE) c_int;
```
karlseguin: Not used?

src/sys/libcrypto.zig
```diff
@@ -343,6 +343,26 @@ pub extern fn EVP_MD_CTX_free(ctx: ?*EVP_MD_CTX) void;
 pub const struct_evp_md_ctx_st = opaque {};
 pub const EVP_MD_CTX = struct_evp_md_ctx_st;
 
+pub extern fn X509_free(x509: ?*X509) void;
+pub extern fn d2i_X509(out: [*c]?*X509, inp: *[*]const u8, len: c_long) ?*X509;
+
+//pub const struct_x509_store_ctx_st = opaque {};
+//pub const X509_STORE_CTX = struct_x509_store_ctx_st;
```
karlseguin:
```suggestion

```

PR #2852, opened 2026-07-01
mem: add local_arena

karlseguin:
The `call_arena` is currently our shortest-lived arena. Its promise is that it will be valid for at least 1 v8 -> zig -> v8 function call, which makes it ideal for getting values from v8 into zig, temporary work, and getting values from zig to v8.

VBut `call_arena`'s lifetime is actually much longer. It's only reset when the call_depth reaches 0. And the reason for that are Zig functions that invoke v8 callbacks. If you just do it when a function ends, then if you allocate in ZigA and then call CallbackA which calls ZigB, then when ZigB ends, your ZigA allocation is cleared. This is something we could solve by reaching into Zig's ArenaAllocator to capture a position to rollback from.

So, `call_arena` can end up living relatively long and accumulating quite a bit of memory. But most calls _don't_ invoke callbacks. Hence, `local_arena` IS reset at the end of every function.

Using `local_arena` _is_ dangerous, mostly for the case where some of our APIs receive a *Frame (or *Execution) and thus might use a `local_arena` because they know they aren't invoking a JS callback. BUT, those APIs might not know that they're also invoked by some other Zig code that _could_ be invoking JS.

Dangerous? Sure. But, github has code that looks like:

```js
function onDelegatedClick(e) {
  for (const el of document.querySelectorAll('[data-action]')) {
    if (el.matches('.menu > .item:not(.disabled) a[href]')) {
      handle(el);
    }
    el.closest('.panel');
  }
}
```

Anything allocated in the `call_arena` will only be freed when the caller of `onDelegatedClick` ends. The `querySelectorAll` returns hundreds of elements and thus builds hundreds of parsed CSS and other scrap (x2 for `matches` and closest`). `call_arena` peaks at 15MB. With the local_arena? 1MB. If 10x more elements were returned, the peak would be 10x higher. With local_arena, it stays 1MB.

### DIFF
--- a/src/browser/Frame.zig
+++ b/src/browser/Frame.zig
@@ -251,9 +251,14 @@ js: *JS.Context,
 // An arena for the lifetime of the frame.
 arena: Allocator,
 
-// An arena with a lifetime guaranteed to be for 1 invoking of a Zig function
-// from JS. Best arena to use, when possible.
+// An arena with a lifetime for at least the scope of one Zig invocation from
+// JS. Prefer local_arena where possible. Use call_arena when allocations may
+// need to call back into JS (event dispatch, forEach callback, ....)
 call_arena: Allocator,
+
+// An arena with a lifetime guaranteed to be for exactly 1 invoking of a Zig
+// function from JS. Best arena to use, when possible.
+local_arena: Allocator,
 
 parent: ?*Frame,
 window: *Window,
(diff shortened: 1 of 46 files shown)


krichprollsch (approved):
How did you identify where to switch to local_arena?

karlseguin:
> How did you identify where to switch to local_arena?

I searched call_arena and looked at each one. Then had claude review my changes. Almost all usage of `call_arena` is safe as `local_arena`. `call_arena` is already a short lifetime, so if you need memory for an indeterminate time, you already had to dupe into something longer than `call_arena`. That just leaves the few cases where a Zig WebAPI calls v8, things like forEach.

PR #2855, opened 2026-07-01
Always clear buffered events on `firePartial`

mookums:
This makes it so that the buffered events in the `DeferredLayer` will be cleared even on a `firePartial` that ends up forwarding an error. Previously, these events would remain buffered if the loop did not properly succeed.


karlseguin:
No longer needed since https://github.com/lightpanda-io/browser/pull/2889

PR #2863, opened 2026-07-02
reclaim FinalizerCallback identities at page teardown

krichprollsch:
Page teardown resets the identity-map Globals, which cancels their pending weak callbacks — the only place fc_identity_pool nodes were freed. Any object still referenced at page close leaked one Identity per navigation until the connection closed. Destroy them in FinalizerCallback.deinit instead; releaseRef unlinks itself when it fires, so everything still linked there has a cancelled callback.

### DIFF
--- a/src/browser/js/js.zig
+++ b/src/browser/js/js.zig
@@ -414,13 +414,24 @@ pub const FinalizerCallback = struct {
 
     // Called during Page teardown to force cleanup regardless of identities.
     pub fn deinit(self: *FinalizerCallback, page: *Page) void {
-        // Mark all identities as done so stale V8 weak callbacks
-        // won't find the wrong FC if resolved_ptr_id is reused.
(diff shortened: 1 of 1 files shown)


karlseguin:
When I push my own claude on this, it also says this is safe, but I'm 90% sure it isn't.

https://github.com/lightpanda-io/browser/pull/2548 was the PR that more or less ended a constant stream of UAFs. I believe your PR circumvents the 2548 mechanism and brings it back to model we had earlier which was prone to UAFs. I think Claude is under-estimating the non-deterministic nature of the v8 GC in the face of the v8::Inspector.

I'll admit I was never able to fully reason what the inspector was doing, but what I did observe were finalizers being called after page teardown. The only thing I'd feel comfortable doing is adding a CDP flag and only doing this cleanup  when `is_cdp != true`

krichprollsch:
Let's close this one then.

PR #2866, opened 2026-07-02
networking: faster certificate loading for Linux and BSDs

nikneym:
Satisfies the TODO comment by taking advantage of `X509_STORE_load_locations`; we can't do this on macOS since there aren't any direct certs we can reach.

### DIFF
--- a/src/network/Network.zig
+++ b/src/network/Network.zig
@@ -749,6 +749,8 @@ fn createX509Store(allocator: Allocator) CreateX509StoreError!*crypto.X509_STORE
                     break :blk;
                 }
             }
+
+            log.warn(.app, "No system certificates", .{});
         },
         else => {
             // Prefer stdlib's cert scanner.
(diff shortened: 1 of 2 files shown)


src/network/Network.zig
```diff
@@ -718,41 +718,84 @@ pub fn newConnection(self: *Network) ?*http.Connection {
...
+                var ptr: [*]const u8 = bytes.ptr + index.*;
+                const x509 = crypto.d2i_X509(null, &ptr, @intCast(bytes.len - index.*)) orelse {
+                    log.warn(.app, "Skipping unparseable system cert", .{});
+                    continue;
+                };
+                defer crypto.X509_free(x509); // add_cert takes its own ref; drop ours.
 
-    const bytes = bundle.bytes.items;
-    if (bytes.len == 0) {
-        log.warn(.app, "No system certificates", .{});
```
karlseguin, line 737: Do you think this kind of notification is important to keep? I think it's common for constrained environments to not have any CA certs by default and it's easy to forget/overlook.

nikneym: I agree, it was there before, removed the check for it.

karlseguin: My comment was that we should keep it.

PR #2867, opened 2026-07-02
`libcurl`: refactor `curl_easy_setopt`

nikneym:
Removes lying dead code here and simplifies `curl_easy_setopt` type checking. Also marks passed callbacks as with C calling convention in order to get away from additional function wrapping.

### DIFF
--- a/src/browser/webapi/net/WebSocket.zig
+++ b/src/browser/webapi/net/WebSocket.zig
@@ -545,16 +545,11 @@ fn dispatchCloseEvent(self: *WebSocket, code: u16, reason: []const u8, was_clean
     }
 }
 
-fn sendDataCallback(
-    buffer: [*c]u8,
-    buf_count: usize,
-    buf_len: usize,
-    raw_connection: ?*anyopaque,
-) callconv(.c) usize {
+fn sendDataCallback(buffer: [*]u8, buf_count: usize, buf_len: usize, data: *anyopaque) callconv(.c) usize {
     if (comptime IS_DEBUG) {
         std.debug.assert(buf_count == 1);
     }
-    const conn: *http.Connection = @ptrCast(@alignCast(raw_connection));
+    const conn: *http.Connection = @ptrCast(@alignCast(data));
     return _sendDataCallback(conn, buffer[0..buf_len]) catch |err| {
         log.warn(.websocket, "send callback", .{ .err = err });
         return http.readfunc_pause;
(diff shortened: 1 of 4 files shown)


src/browser/webapi/net/WebSocket.zig
```diff
@@ -545,11 +545,16 @@ fn dispatchCloseEvent(self: *WebSocket, code: u16, reason: []const u8, was_clean
     }
 }
 
-fn sendDataCallback(buffer: [*]u8, buf_count: usize, buf_len: usize, data: *anyopaque) usize {
+fn sendDataCallback(
+    buffer: [*c]u8,
+    buf_count: usize,
+    buf_len: usize,
+    raw_connection: ?*anyopaque,
```
karlseguin: Here you renamed `data` -> `raw_connection` but in `receivedDataCallback` you kept it as `data`.

The changes to this file and HttpClient.zig seem unnecessary / arbitrary.

src/sys/libcurl.zig
```diff
@@ -595,172 +592,58 @@ pub fn curl_easy_perform(easy: *Curl) Error!void {
...
-                    .ssl_verify_host, .proxy_ssl_verify_host => if (value) 2 else 0,
-                    else => if (value) 1 else 0,
-                },
-                else => @compileError("expected bool|integer for " ++ @tagName(option) ++ ", got " ++ @typeName(@TypeOf(value))),
-            };
-            break :blk c.curl_easy_setopt(easy, opt, n);
-        },
+        => @as(c_long, @intFromBool(value)),
+
+        // VERIFYHOST takes 0 or 2 (1 is rejected since curl 7.66).
```
karlseguin: > 1 is rejected since curl 7.66

Really? From the linked page: "From 7.66.0: libcurl treats 1 and 2 to this option the same."

src/sys/libcurl.zig
```diff
@@ -31,22 +33,22 @@ pub const CurlCode = c.CURLcode;
...
 
 pub const CURLE = struct {
     pub const OK = c.CURLE_OK;
     pub const ABORTED_BY_CALLBACK = c.CURLE_ABORTED_BY_CALLBACK;
 };
 
-pub const CurlDebugFunction = fn (*Curl, CurlInfoType, [*c]u8, usize, *anyopaque) c_int;
-pub const CurlHeaderFunction = fn ([*]const u8, usize, usize, *anyopaque) usize;
-pub const CurlWriteFunction = fn ([*]const u8, usize, usize, *anyopaque) usize;
+pub const HeaderFunction = *const fn ([*]const u8, usize, usize, *anyopaque) callconv(.c) usize;
```
karlseguin: I don't mind this change, but doesn't it make it very inconsistent since only a few dropped the `Curl` prefix?

PR #2869, opened 2026-07-03
perf: used cachedParse for closest and waitForSelector

karlseguin:
closest was already hitting the cache, but it was doing it inside the loop. Moving it outside the loop avoids the cache lookup.

Inversely, Runner.waitForSelector was parsing outside the loop, but it wasn't caching. Repeated calls to waitForSelector with the same selector would not leverage the cache.

### DIFF
--- a/src/browser/Runner.zig
+++ b/src/browser/Runner.zig
@@ -301,13 +301,10 @@ fn _tick(self: *Runner, comptime is_cdp: bool, timeout_ms: u32, conditions: []Wa
     return .done;
 }
 
-pub fn waitForSelector(self: *Runner, frame_id: u32, selector: [:0]const u8, timeout_ms: u32) !*Node.Element {
+pub fn waitForSelector(self: *Runner, frame_id: u32, input: [:0]const u8, timeout_ms: u32) !*Node.Element {
     const session = self.session;
-    const arena = try session.getArena(.small, "Runner.waitForSelector");
-    defer session.releaseArena(arena);
-
     var timer = try std.time.Timer.start();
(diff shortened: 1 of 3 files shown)


krichprollsch (approved):

src/browser/Runner.zig
```diff
@@ -301,13 +301,10 @@ fn _tick(self: *Runner, comptime is_cdp: bool, timeout_ms: u32, conditions: []Wa
...
 
-pub fn waitForSelector(self: *Runner, frame_id: u32, selector: [:0]const u8, timeout_ms: u32) !*Node.Element {
+pub fn waitForSelector(self: *Runner, frame_id: u32, input: [:0]const u8, timeout_ms: u32) !*Node.Element {
     const session = self.session;
-    const arena = try session.getArena(.small, "Runner.waitForSelector");
-    defer session.releaseArena(arena);
-
     var timer = try std.time.Timer.start();
-    const parsed_selector = try Selector.parseLeaky(arena, selector);
+    const selector = try Selector.cachedParse(session.browser, input);
```
krichprollsch: Claude suggests to move it into the while loop b/c you take the risk the selector been evicted as the oldest during the loop.

karlseguin: Yes! good catch. I've reverted the Runner to how it was before. I think it's the faster option for the most common case (a single call to waitForSelector which loops a lot).

krichprollsch: I agree

krichprollsch (approved):

PR #2874, opened 2026-07-03
Add agent regression suite to CI (consumed from lightpanda-io/demo)

arrufat:
Adds CI coverage for `lightpanda agent`, consuming the regression suite that landed in lightpanda-io/demo#206 — same model as the runner/integration/wpt suites (checkout demo, download the built binary, drive it).

## What this adds

- **`agent-deterministic` job in `e2e-test.yml`** (every PR, no API key): replays golden PandaScripts against the demo sites (campfire-commerce single-page extract, amiibo front→item crawl — both JS-rendered, so the layer exercises script execution, fetch, and DOM manipulation) and diffs the returned JSON exactly against the committed goldens.
- **`agent-regression.yml`** (manual `workflow_dispatch` only with an optional `model` input): runs the deterministic layer plus the live LLM layer — closed-form Q&A over campfire-commerce, and a live Hacker News scrape saved via `--save`, replayed token-free and validated against a jq shape invariant. Uses the `GEMINI_API_KEY` secret (default model `gemini-3.5-flash`, pinned in the suite) and reuses `MASSIVE_PROXY_RESIDENTIAL_US` for the live HN call.
- **`make test-agent`** wraps `../demo/agent/run.sh` for local runs, mirroring `end2end`.

The binary is passed via `LPD_PATH` (the same variable wptrunner uses). Per-task `$usage` token accounting and a ceiling guard against runaway agent loops; agent stderr is surfaced on failures.

## Validation

- Full pipeline green on this branch via a temporary push trigger (since dropped): build → deterministic 2/2 → live 6/6, run 28646957472 (~294k live-layer tokens, HN save 280k).
- Earlier end-to-end validation of the pre-move suite: run 28576642458.
- `make test-agent LAYER=deterministic` green locally against demo@main.


arrufat:
Pushed `f3f731ff` — `fix: setChecked double-toggled checkables via click activation`. The new `checked.js` golden in lightpanda-io/demo#207 caught this: `actions.setChecked` set the state and then dispatched a trusted click whose activation behavior toggled it right back (with doubled `input`/`change`). Merge order: this PR first, then demo#207 — its golden pins the fixed behavior.

Issue #2877, opened 2026-07-03
Add a `--cacert` and `--capath` options to load TLS root certificates from a custom PEM file/dir


krichprollsch:
Lightpanda currently trusts whatever the OS trust store contains: the `X509_STORE` shared by all connections is built via `std.crypto.Certificate.Bundle.rescan`, which scans the platform's default certificate locations (e.g. `/etc/ssl/certs/ca-certificates.crt` on Debian-based systems).

But Chrome and FF contains their own CA store. So we can have differences with LP and Chrome behavior for the same domain.

`--cacert` and `--capath` are Curl's current option for the similar feature.
https://curl.se/docs/sslcerts.html

PR #2891, opened 2026-07-06
Support loading custom CAs (`--ca-cert`, `--ca-path`)

nikneym:
Addresses #2877, custom CA certs and directories can be provided via `--ca-cert` and `--ca-path`.

In order to test, you can grab Mozilla CA store in PEM format from [curl's website](https://curl.se/docs/caextract.html); then execute this:
```sh
lightpanda fetch "https://lightpanda.io/" --ca-cert <PEM> --dump > output.html
```

### DIFF
--- a/src/Config.zig
+++ b/src/Config.zig
@@ -118,6 +118,9 @@ const CommonOptions = .{
     .{ .name = "v8_flags_unsafe", .type = ?[]const u8 },
     .{ .name = "v8_max_heap_mb", .type = ?u32 },
     .{ .name = "watchdog_ms", .type = ?u32 },
+    .{ .name = "ca_cert", .type = [:0]const u8, .multiple = true },
+    .{ .name = "ca_path", .type = [:0]const u8, .multiple = true },
+    .{ .name = "disable_root_certificates", .type = bool },
 };
 
 fn dumpValidator(_: Allocator, args: *std.process.Args.Iterator) !?DumpFormat {
@@ -384,6 +387,14 @@ pub fn v8Flags(self: *const Config) ?[]const u8 {
 pub fn v8MaxHeapMb(self: *const Config) ?u32 {
     return switch (self.mode) {
         inline .serve, .fetch, .mcp, .agent => |opts| opts.v8_max_heap_mb,
+        else => unreachable,
+    };
+}
+
+pub fn disableRootCertificates(self: *const Config) bool {
+    return switch (self.mode) {
+        inline .serve, .fetch, .mcp, .agent => |opts| opts.disable_root_certificates,
+        .version => false,
         else => unreachable,
     };
 }
--- a/src/Config.zig
+++ b/src/Config.zig
@@ -147,7 +147,6 @@ const CommonOptions = .{
     .{ .name = "watchdog_ms", .type = ?u32 },
     .{ .name = "ca_cert", .type = [:0]const u8, .multiple = true },
     .{ .name = "ca_path", .type = [:0]const u8, .multiple = true, .validator = caPathValidator },
-    .{ .name = "disable_root_certificates", .type = bool },
 };
 
 fn dumpValidator(_: Allocator, args: *std.process.Args.Iterator) !?DumpFormat {
@@ -414,14 +413,6 @@ pub fn v8Flags(self: *const Config) ?[]const u8 {
 pub fn v8MaxHeapMb(self: *const Config) ?u32 {
     return switch (self.mode) {
         inline .serve, .fetch, .mcp, .agent => |opts| opts.v8_max_heap_mb,
-        else => unreachable,
-    };
-}
-
-pub fn disableRootCertificates(self: *const Config) bool {
-    return switch (self.mode) {
-        inline .serve, .fetch, .mcp, .agent => |opts| opts.disable_root_certificates,
-        .version => false,
         else => unreachable,
     };
 }
--- a/src/Config.zig
+++ b/src/Config.zig
@@ -74,7 +74,7 @@ fn logFilterScopesValidator(allocator: Allocator, args: *std.process.Args.Iterat
     }
 }
 
-fn logLevelValidator(_: Allocator, args: *std.process.ArgIterator, target: *?log.Level) !void {
+fn logLevelValidator(_: Allocator, args: *std.process.Args.Iterator, target: *?log.Level) !void {
     const str = args.next() orelse return error.MissingArgument;
     if (std.mem.eql(u8, str, "error")) {
         target.* = .err;
@@ -116,7 +116,7 @@ const Cert = struct {
 
 fn caCertValidator(
     _: Allocator,
-    args: *std.process.ArgIterator,
+    args: *std.process.Args.Iterator,
     cert: *Cert,
 ) !void {
     const file_name = args.next() orelse return error.MissingArgument;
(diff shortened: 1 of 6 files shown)


src/network/Network.zig
```diff
@@ -720,10 +720,38 @@ const CreateX509StoreError = std.crypto.Certificate.Bundle.RescanError || error{
...
+                    log.warn(.app, "Invalid CA cert", .{ .ca_cert = ca_cert });
+                }
+            }
+        },
+        .help => unreachable,
+        .version => {}, // Don't load custom CA for version command.
+    }
+
+    // Early return if root certificates are unwanted.
+    if (config.disableRootCertificates()) {
```
karlseguin: Is the idea here that someone can specify `--ca-cert` WITHOUT `--insecure-disable-tls-host-verification` so that we still load the system certificates, or WITH it so that we only load their certs? Most flexible that way.

src/help.zon
```diff
@@ -306,6 +306,17 @@
     \\  --insecure-disable-tls-host-verification
     \\          Disables host verification on all HTTP requests.
     \\          Only set this if you understand and accept the risk.
+    \\  --ca-cert <PATH>
```
karlseguin: Help options are now sorted alphabetically, which these new entries are not.

src/Config.zig
```diff
@@ -84,6 +84,35 @@ fn logLevelValidator(_: Allocator, args: *std.process.ArgIterator) !?log.Level {
...
 
+pub fn isHashedDirectory(dir: []const u8) bool {
+    var handle = std.fs.openDirAbsolute(dir, .{ .iterate = true }) catch return false;
+    defer handle.close();
+
+    var hashed = false;
+    var it = handle.iterate();
+    while (it.next() catch return false) |entry| {
+        if (std.mem.endsWith(u8, entry.name, ".0")) {
+            hashed = true;
```
karlseguin: can you just return true here and then return false outside the loop? why the local?

src/network/Network.zig
nikneym: > Is the idea here that someone can specify --ca-cert WITHOUT --insecure-disable-tls-host-verification so that we still load the system certificates

This is the idea, indeed.

src/network/Network.zig
```diff
@@ -720,9 +720,43 @@ const CreateX509StoreError = std.crypto.Certificate.Bundle.RescanError || error{
...
 /// NEVER give full ownership of store to `SSL_CTX`, always rely on ref counting.
 /// Allocations made through passed `allocator` are freed before this function returns.
-fn createX509Store(allocator: Allocator) CreateX509StoreError!*crypto.X509_STORE {
+fn createX509Store(allocator: Allocator, config: *const Config) CreateX509StoreError!*crypto.X509_STORE {
     const store = crypto.X509_STORE_new() orelse return error.FailedToCreateX509Store;
     errdefer crypto.X509_STORE_free(store);
+    // Report back if no certificates loaded.
+    defer {
+        const num_of_certs = crypto.getCertCount(store);
+        if (num_of_certs == 0) {
```
karlseguin: If --ca-path is hash directory, this will incorrectly warn.

src/network/Network.zig
```diff
@@ -720,9 +720,43 @@ const CreateX509StoreError = std.crypto.Certificate.Bundle.RescanError || error{
...
+            log.warn(.app, "No certificates loaded", .{});
+        }
+    }
+
+    load_custom_ca: switch (config.mode) {
```
karlseguin: In the end, did we decide that if --ca-path or -ca-cert was specified, we'd continue to load the system certs too?

src/Config.zig
```diff
@@ -116,6 +143,8 @@ const CommonOptions = .{
     .{ .name = "enable_external_stylesheets", .type = bool },
     .{ .name = "v8_flags_unsafe", .type = ?[]const u8 },
     .{ .name = "v8_max_heap_mb", .type = ?u32 },
+    .{ .name = "ca_cert", .type = [:0]const u8, .multiple = true },
+    .{ .name = "ca_path", .type = [:0]const u8, .multiple = true, .validator = caPathValidator },
```
karlseguin: It's a bit inconsistent that a bad ca_path stops the app from starting, but a bad ca_cert only warns at a later time.

src/Config.zig
```diff
@@ -84,6 +84,33 @@ fn logLevelValidator(_: Allocator, args: *std.process.ArgIterator) !?log.Level {
...
+    args: *std.process.ArgIterator,
+    list: *std.ArrayList([:0]const u8),
+) !void {
+    const dir = args.next() orelse return error.MissingArgument;
+    if (!isHashedDirectory(dir)) {
+        log.fatal(.app, "invalid CA path", .{ .arg = "--ca-path", .value = dir });
+        return error.InvalidArgument;
+    }
+
+    return list.append(allocator, dir);
```
karlseguin: in cli.zig, the documentation says that strings are "duped from argv". So I assume if they have to be duped there, it has to be duped here?

karlseguin:
`zig build run -- fetch --ca-path ca-test https://example.com`   crashes

src/network/Network.zig
nikneym: I think we didn't, loading only what's provided would probably be the desired behavior.

src/Config.zig
nikneym: Agreed. I think we can either try parsing `ca_cert` in `Config` and store or make both warn at a later time.

Parsing before creating the store requires us to store the `X509` pointers in somewhere though; so, if we can add something like this to `Config`, we can distinguish internal representation and CLI expectation.
```zig
.{
    .name = "ca_cert",
    .type = .{
        .cli = [:0]const u8,
        .memory = *crypto.X509,
    },
    .multiple = true,
},
```

Which would expect `[:0]const u8` from CLI but prefer `std.ArrayList(*crypto.X509)` in internal store.

Or, we can just parse 2 times; one in `Config` and the other at X509 store creation time.

src/network/Network.zig
```diff
@@ -724,9 +724,48 @@ const CreateX509StoreError = std.crypto.Certificate.Bundle.RescanError || error{
...
+
+            for (files) |ca_cert| {
+                if (crypto.X509_STORE_load_locations(store, ca_cert, null) != 1) {
+                    log.warn(.app, "Invalid CA cert", .{ .ca_cert = ca_cert });
+                }
+            }
+        },
+        .help => unreachable,
+        .version => {}, // Don't load custom CA for version command.
+    }
```
krichprollsch: Move the switch into a Config's accessors instead keeping it in Network.
Somehting like `caCerts` and `caCertDirs`.

src/sys/libcrypto.zig
```diff
@@ -353,6 +353,19 @@ pub extern fn X509_STORE_new() ?*X509_STORE;
...
+pub const OPENSSL_sk_cmp_func = ?*const fn ([*c]?*const anyopaque, [*c]?*const anyopaque) callconv(.c) c_int;
+pub const struct_stack_st = extern struct {
+    num: usize,
+    data: [*c]?*anyopaque,
+    sorted: c_int,
+    num_alloc: usize,
+    comp: OPENSSL_sk_cmp_func,
+};
+pub const _STACK = struct_stack_st;
+pub extern fn sk_num(sk: [*c]const _STACK) usize;
```
krichprollsch: Claude complaints about this use:
```
  2. The sk_num binding is unmodified translate-c output and uses a deprecated symbol. (src/sys/libcrypto.zig:356-368, 395-404)

  - struct_stack_st is declared but never dereferenced — only used as a @ptrCast target. It's a silent trap if someone later reads a field and the layout has drifted. Delete it.
  - sk_X509_OBJECT_num is Zig-only but declared callconv(.c) with a pointless const sk = arg_sk;. Compare findDigest at the bottom of the same file for the hand-written style.
  - sk_num is OPENSSL_DEPRECATED (stack.h:302). OPENSSL_sk_num (stack.h:257) is the supported name with the identical size_t signature, and returns 0 for NULL — so the whole thing is
  three lines with no null check needed.
```

WDYT?

src/Config.zig
krichprollsch: We still have inconsistency here:
* invalid `ca-path`:  the process doesn't start
* invalid `ca-cert`: the process starts w/ a warning "Invalid CA cert" but idk if we loaded the system's cert or if we disabled tls validation.

I think we should stop in both cases.

src/sys/libcrypto.zig
nikneym: I think we can do counting by ourselves, though there shouldn't be any deprecation notice on `sk_num` on BoringSSL.

src/Config.zig
nikneym: > I think we should stop in both cases.

Agreed, will match the behavior of both.

PR #2895, opened 2026-07-07
perf: Iframes with loading=lazy don't delay parent's "load" event

karlseguin:
When an iframe has "loading=lazy" it no longer delays the parent's "load" event. It's still loaded (subject to `--disable-subframes` parameter).

Many sites use loading=lazy for non-critical content (e.g. ads). On Chrome/ Firefox, these are only loaded when they're in the viewport, so sites usually can't depend/rely on them being loaded.

### DIFF
--- a/src/browser/Session.zig
+++ b/src/browser/Session.zig
@@ -655,14 +655,13 @@ fn _processFrameNavigation(self: *Session, frame: *Frame, qn: *QueuedNavigation)
 
     errdefer iframe._window = null;
 
-    // Should this frame's navigation delay the parent's load event? If it was
-    // already holding it, then that just carries over. If it wasn't delaying
-    // it (maybe because it completed already), then it should start, UNLESS
-    // the iframe is lazy loaded.
-    const had_hold = !frame._parent_notified and frame._delays_parent_load;
-    const delays_load = had_hold or !iframe.isLazyLoading();
-    const new_hold = delays_load and !had_hold;
-    if (new_hold) {
+    // was the previous navigation of this frame delaying the paren't load event?
+    const was_delaying = frame._delays_parent_load and !frame._parent_notified;
+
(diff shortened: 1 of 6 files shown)


karlseguin:
@krichprollsch We could replace `--disable-subframes` with something like: `--iframe-loading all/none/skip-lazy`  So that people can load iframes, but not lazy iframes (which, by nature of being lazy, we could say are likely to be even less critical).

I don't know how you feel about having a lot of configurable options in general.

krichprollsch:
do not delay the `load` event for lazy iframes is great! thanks for the change.
About the option, I'm not sure havinga `skip-lazy` is really useful since you can use events or selector to stop waiting.

At blueboard, we used to use selector b/c we wanted iframe data, so I would prefer having a way to wait for lazy options. But iframes on/off is good enough for now. Let's wait for a user's demand.

src/browser/Session.zig
```diff
@@ -655,9 +655,14 @@ fn _processFrameNavigation(self: *Session, frame: *Frame, qn: *QueuedNavigation)
...
-    const parent_notified = frame._parent_notified;
-    if (parent_notified) {
-        // we already notified the parent that we had loaded
+    // Should this frame's navigation delay the parent's load event? If it was
+    // already holding it, then that just carries over. If it wasn't delaying
+    // it (maybe because it completed already), then it should start, UNLESS
+    // the iframe is lazy loaded.
+    const had_hold = !frame._parent_notified and frame._delays_parent_load;
+    const delays_load = had_hold or !iframe.isLazyLoading();
+    const new_hold = delays_load and !had_hold;
```
krichprollsch: I have hard time trying to understand these condtions :/

krichprollsch (approved):

src/browser/Session.zig
karlseguin: I think I improved it.

PR #2896, opened 2026-07-07
feature: add watchdog for JS hangs

karlseguin:
Adds a new command line argument `--watchdog-ms` which, when set, will terminate any JS that appears to be hung.

When configured, a new thread is started. Workers heartbeat this thread to signal activity (e.g. not stuck in a JS loop). However, workers also block for their own reason (e.g. network polling), so they can signal the watchdog that they are "entering a wait" and, when complete, that they are "existing a wait". During such waits, the watchdog will not signal the isolate to terminate.

Obviously, it's important for workers to signal aliveness and whenever they plan on doing a non-JS wait. So you could say we introduce safe points where the watchdog (for that browser) is disabled. We could do the opposite: enable the watchdog whenever we enter JS ("hey, I'm about to execute JS, monitor me). But there are _a lot_ more place where this happens.

### DIFF
--- a/src/help.zon
+++ b/src/help.zon
@@ -360,7 +360,7 @@
     \\          stuck in an endless script loop). In serve mode the CDP
     \\          connection is then closed. Unlike fetch's --terminate-ms, an idle
     \\          or normally-working page never trips this.
-    \\          Defaults to disabled.
+    \\          Defaults to [a very generous] 30000, disable by setting to 0.
     \\  --web-bot-auth-domain <DOMAIN>
     \\          Your domain, e.g. yourdomain.com.
     \\  --web-bot-auth-key-file <PATH>
--- a/src/help.zon
+++ b/src/help.zon
@@ -354,6 +354,13 @@
     \\          Maximum V8 heap size in megabytes, per browser instance. Values
     \\          below ~16 are clamped by V8 and all behave the same.
     \\          Defaults to the V8 default (based on available memory).
+    \\  --watchdog-ms <INT>
+    \\          Terminates JavaScript when the browser stalls — stays busy for
+    \\          this long without returning to its network poll (e.g. a page
+    \\          stuck in an endless script loop). In serve mode the CDP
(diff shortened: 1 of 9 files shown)


src/help.zon
```diff
@@ -338,6 +338,13 @@
     \\          still sends Sec-Ch-Ua. Incompatible with --user-agent-suffix.
     \\  --user-agent-suffix <STRING>
     \\          Suffix appended to the Lightpanda/X.Y User-Agent.
+    \\  --watchdog-ms <INT>
+    \\          Terminates JavaScript when the browser stalls — stays busy for
+    \\          this long without returning to its network poll (e.g. a page
+    \\          stuck in an endless script loop). In serve mode the CDP
+    \\          connection is then closed. Unlike fetch's --terminate-ms, an idle
+    \\          or normally-working page never trips this.
+    \\          Defaults to disabled.
```
krichprollsch: I tink we should recommand (or give an example) of value. IMO using something too short is not good, for https://www.kitandace.com/, having less than `30000` won't work.

krichprollsch (approved):

src/help.zon
karlseguin: I just tried with `--watchdog-ms 10000` and it worked. The watchdog doesn't look at page/frame load, it more or less looks at JS run without hitting safe-point (Runner.Tick, HttpClient.poll). 30 seconds seems high, but maybe a safe default to initially recommend...but if some JS code executes for 10 seconds, without ever yielding to one of our two safepoints, I have a hard time imagining waiting another 20 seconds will unblock it.

krichprollsch: :thinking: maybe I should try to understand a bit more why I have failures locally with a lower value.

krichprollsch: but waiting 30s vs waiting forever is still improvement IMO

karlseguin: I changed the default to 30 seconds. But it would be interesting to know why it doesn't work with 10 seconds on www.kitandace.com.

krichprollsch (approved):

PR #2897, opened 2026-07-07
v8, mem: Add v8 configuration option + heap limit protection

karlseguin:
Depends on https://github.com/lightpanda-io/zig-v8-fork/pull/186

The main addition in this commit is that we hook into the Isolate's AddNearHeapLimitCallback callback and try to force the isolate to shutdown rather than letting v8 hit an OOM which would take down the entire process.

In support of this, we now support a `--v8-max-heap-mb` command line option to set an explicit heap limit. As a simple way to test this feature, load a relatively heavy JS page with `--v8-max-heap-mb 1`.

There's also a `--v8-flags-unsafe` which is a mechanism to pass arbitrary flags to v8 via its `SetFlagsFromString`. The parameter is called `unsafe` because some [of the many] configurable flags could conflict with how the snapshot is built and result in crashes. The snapshot creator also gains a `--v8-flags-unsafe` flag, so advance users COULD create their snapshot and run lightpanda with the same set of flags.

### DIFF
--- a/.github/actions/install/action.yml
+++ b/.github/actions/install/action.yml
@@ -13,7 +13,7 @@ inputs:
   zig-v8:
     description: 'zig v8 version to install'
     required: false
-    default: 'v0.5.0'
+    default: 'v0.5.1'
   v8:
     description: 'v8 version to install'
     required: false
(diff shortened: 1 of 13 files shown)


krichprollsch (approved):

krichprollsch:
you wait for the zig-v8 merge to change `build.zig.zon` right?

karlseguin:
> you wait for the zig-v8 merge to change `build.zig.zon` right?

Yes.

PR #2899, opened 2026-07-08
agent: add script healing capability

arrufat:
Adds a `--heal` CLI flag and interactive REPL prompt to automatically diagnose and repair failing scripts using the LLM. The agent attempts to fix the script, then validates the revision in a fresh session before overwriting the original file.


src/agent/Agent.zig
```diff
@@ -1472,18 +1666,371 @@ fn runScript(self: *Agent, path: []const u8) bool {
...
+        \\verify selectors against the live page, never guess. If the failing
+        \\step gated the rest of the script (a login, a navigation), carry on
+        \\far enough to show the script's goal is reachable again.
+    , .{ path, source, error_detail });
+}
+
+/// Heal synthesis instruction; rides on the regular save revision system prompt.
+const heal_revision_prompt =
+    \\Fix the script so it replays successfully against the current site: the
+    \\error and this session's working tool calls identify the repair. Keep
```
karlseguin: "and this session's working tool calls identify the repair"

Is that right?

src/agent/Agent.zig
```diff
@@ -1472,18 +1666,371 @@ fn runScript(self: *Agent, path: []const u8) bool {
...
+    return false;
+}
+
+const verdict_system_prompt =
+    \\You judge replays of saved browser-automation scripts. Given the script
+    \\and facts about a run that completed without errors, decide whether the
+    \\empty output means the script is broken (stale selectors after a site
+    \\change) or legitimate (the page genuinely has no such data right now).
+    \\A `
+++ std.mem.trimRight(u8, Baseline.marker, " ") ++
```
karlseguin:
```suggestion
++ std.mem.trimEnd(u8, Baseline.marker, " ") ++
```

Help 0.16.

src/agent/Agent.zig
```diff
@@ -1472,18 +1666,371 @@ fn runScript(self: *Agent, path: []const u8) bool {
...
+        \\```js
+        \\{s}
+        \\```
+    , .{ path, suspicion.detail, suspicion.source }) catch return null;
+    const raw = self.soloTurn(arena, "verdict", verdict_system_prompt, user_msg, 1024, self.effort) orelse return null;
+    return parseVerdict(arena, raw);
+}
+
+fn parseVerdict(arena: std.mem.Allocator, raw: []const u8) ?Verdict {
+    const start = std.mem.indexOfScalar(u8, raw, '{') orelse return null;
```
karlseguin: This seems frail.

Are you worried it'll add extra context? Could you ask it to put a more explicit market? `VERDICT: {...}`

karlseguin (commented):
Don't really understand, a few simple comments...feel free to ignore.  Looks fine mechanically.

src/agent/Agent.zig
arrufat: This is correct: the synthesis turn sees the diagnose turn's tool calls because it rides on the full conversation. However, "session" was unclear because the new browser session was used for validation, so I changed it to "the diagnosis tool calls above."

src/agent/Agent.zig
arrufat: Fixed! Thank you!

PR #2920, opened 2026-07-12
cli: add `run` command for script replay

arrufat:
## What

Adds `run` as a dedicated, script-only command:

```
lightpanda run script.js      # replay a saved script, token-free (no LLM)
lightpanda run                # error: needs a script
lightpanda run --task "..."   # parse error: --task is not a `run` flag
```

`lightpanda agent <script>` keeps working unchanged — `run` is just the natural spelling for what that path already does.

## Why

Users find `lightpanda agent <script>` awkward: replaying a saved script under a subcommand named `agent` reads wrong, especially since that path is fully token-free. `run` is the natural verb.

Scope is deliberately narrow — `run` is the *script-execution* command, not a full alias of `agent`. It owns the `[SCRIPT]` positional + common options only; the LLM-session surface (`--task`, `--save`, `--list-models`, REPL) stays on `agent`.

## Follow-up

`--heal` (on the unmerged `agent-self-heal` branch) will be rebased to hang off `run` once this lands — purely additive to the `run` descriptor.


src/telemetry/lightpanda.zig
```diff
@@ -82,6 +82,7 @@ pub fn init(self: *LightPanda, app: *App, iid: ?[36]u8, run_mode: Config.RunMode
             .fetch => "F",
             .serve => "S",
             .agent => if (interactive == false) "AR" else "A",
+            .run => "R",
```
arrufat, line 85: @krichprollsch we can make this "AR", if you prefer.

karlseguin (changes requested):
PR is fine, but blocking it until the telemetry server is updated.

krichprollsch:
telemetry is updated

karlseguin (approved):

PR #2925, opened 2026-07-13
http: Change default max http size to 1GB (from unlimited)

karlseguin:


src/help.zon
```diff
@@ -314,7 +314,7 @@
     \\  --http-max-response-size <INT>
     \\          Limits the acceptable response size for any request
     \\          e.g. XHR, fetch, script loading.
-    \\          Defaults to no limit.
+    \\          Defaults to 1GB.
```
arrufat: I know I might be a bit picky, but shouldn't it be 1 GiB?
```suggestion
    \\          Defaults to 1 GiB.
```

PR #2929, opened 2026-07-13
script: generate the PandaScript skill from the tool schemas

arrufat:
## Why

The PandaScript documentation that teaches an LLM to write agent scripts lived as a hand-maintained comptime string in `Agent.zig` (`script_skill`). Its primitives table drifted whenever a tool changed — the published [agent-skill](https://github.com/lightpanda-io/agent-skill) SKILL.md shows where that ends up (still documents `semantic_tree`, lists about half of the current tools).

## What

**`src/script/skill.zig`** now owns the skill. The prose (mental model, extract schema, best practices, common errors) is unchanged; the primitives reference is rendered from `Schema.all()` at first use:

- signatures from `positionalFor` + the remaining option fields (`backendNodeId` excluded, `extract` special-cased to `page.extract(schema)`)
- `await` marker from `Tool.isAsync()`
- enum values (so `waitForState` tracks `Config.WaitUntil`), boolean defaults, and per-parameter schema descriptions (`Schema.FieldEntry` now retains them)

Curated per-tool notes the schema can't express (goto's "a timeout does NOT reject", press's selector-first trap) live behind an **exhaustive `switch` on `Tool`** — adding or renaming a tool is a compile error until its doc entry exists. That's the anti-staleness mechanism.

Three consumers share the rendered text:

- the `/save` and revision system prompts (built lazily in `Agent.zig`, replacing the deleted string)
- `mcp://skill/pandascript` — MCP clients can pull the current doc live
- **`zig build skills`** → `zig-out/skills/pandascript/SKILL.md` with Claude Code frontmatter, via a registry-based generator exe (modeled on `snapshot_creator`); a future skill is one registry entry


src/script/skill.zig
```diff
@@ -0,0 +1,341 @@
...
+    \\   ```js
+    \\   const list = new Page();
+    \\   await list.goto(listUrl);
+    \\   const { items } = list.extract({ items: [{ selector: "a.row", fields: { url: { attr: "href" } } }] });
+    \\
+    \\   const pages = items.map(() => new Page());
+    \\   await Promise.all(pages.map((p, i) => p.goto(items[i].url)));   // all in flight at once
+    \\   return pages.map((p, i) => ({ ...items[i], ...p.extract({ /* schema */ }) }));
+    \\   ```
+    \\   - Concurrency is bounded by the HTTP connection pool (40 by default, `--http-max-concurrent`); extra navigations queue rather than fail. For long lists, fan out in batches and `page.close()` each page once read so its memory is reclaimed.
```
karlseguin: Worth mentioning `--http-max-host-open`??  (defaults to 6, same as browsers). Raising it too much could cause issues on the server though,

karlseguin (commented):
Some of the wording/formatting confused me a little, but Opus said it could see how my tiny little brain might get confused, but it understood fine. Since that's the intended audience of the document. I guess it's fine.

karlseguin (approved):

PR #2932, opened 2026-07-13
agent: render markdown in REPL output

arrufat:
## What

The interactive `lightpanda agent` REPL prints the model's answer through `Terminal.printAssistant`, which until now wrote the raw text verbatim — so markdown the model emits (`**bold**`, `# headings`, `- lists`, `[links](url)`) showed up as literal markers.

This adds a small markdown → ANSI renderer (`src/agent/md_term.zig`) and styles the assistant's answer in the REPL.

Handled:
- **Inline:** `**bold**`/`__bold__`, `*italic*`/`_italic_`, `` `code` `` (cyan), `~~strike~~`
- **Blocks:** `#`..`######` headings (bold), `-`/`*`/`+` bullets (dim `•`), numbered lists (marker kept), ` ``` ` fenced code (dim), blockquotes (`> ` → dim bar), horizontal rules (`---`/`***`/`___` → dim line)
- **Links:** `[text](url)` → **OSC 8 hyperlink** (clickable in supporting terminals; ignored elsewhere) with a dim `(url)` fallback so no terminal loses the URL. A bare link (label == url) drops the redundant suffix.
- **Backslash escapes:** `\*` etc. unescape, but only for markdown-special chars — `C:\Users` / `\d` are left intact.
- Unterminated markers (e.g. `**hi` with no close) stay literal — never swallow the rest of the line.

## Scope

Rendering is gated to the interactive REPL on a real terminal (`isRepl() and stdout_is_tty`). Non-interactive paths — `--task`, piped, or otherwise consumed output — stay **verbatim** so they remain machine-readable. Falls back to the raw text on any render failure, so the answer is never dropped.

## Changes

- New `src/agent/md_term.zig` — the renderer (no existing markdown parser to reuse; `browser/markdown.zig` goes the other direction, HTML→markdown).
- `src/agent/Terminal.zig` — `underline`/`strike` added to the `ansi` struct, a `stdout_is_tty` field (sibling of `stderr_is_tty`, set once in `init`), and the gated render in `printAssistant`.

## Testing

- 8 unit tests in `md_term.zig` cover every construct (incl. OSC 8 link + bare-link dedup, blockquote, horizontal rule, escapes) plus the unterminated-marker cases; leak detector clean (`make test F="md_term"`).
- Verified end-to-end with real LLM turns: REPL/TTY path emits correct ANSI; `--task` under a pty stays raw markdown.


src/agent/md_term.zig
```diff
@@ -0,0 +1,245 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+const std = @import("std");
+const ansi = @import("Terminal.zig").ansi;
+
+/// Render markdown `src` to an ANSI-styled terminal string owned by `arena`.
+pub fn render(arena: std.mem.Allocator, src: []const u8) ![]const u8 {
```
karlseguin: Isn't the better API to take an `*std.Io.Writer`?  More flexibility to the caller, e.g. could write directly to stdout.

src/agent/md_term.zig
```diff
@@ -0,0 +1,245 @@
...
+        else => false,
+    };
+}
+
+/// A line of 3+ identical `-`, `*` or `_` markers (spaces ignored).
+fn isHorizontalRule(s: []const u8) bool {
+    var marker: ?u8 = null;
+    var count: usize = 0;
+    for (s) |c| switch (c) {
+        ' ', '\t' => {},
```
karlseguin: The only caller for this has already left-trimmed the input. If you want to make it more re-usable, ok, but it would be more readable to trim upfront,then check if remaining.len >= 3 and then check if (first ==  '-' or first == '*' or first '_') and remaining[1] == first and remaining[2] == first).

The loop hides the fact that this is a prefix match.

karlseguin:
Screenshot 2026-07-14 at 12 10 41

I know it's partially how my terminal is configured, but maybe \x1b[2m  (dim) should never be used. It's almost unreadable.

arrufat:
I've made dim use the same gray as isocline dim, instead of being terminal-dependent.

karlseguin:
Screenshot 2026-07-15 at 09 34 59

i don't want to hold the PR up because of my terminal color, but with someone with awful eyesight....The above low-contrast image + what Gemini said about `\x1b[38;5;244m`:

Because of its low visibility, it is actually incredibly useful—just not for important data. It is best used for:
    - De-emphasized text: Timestamps in log files, borders, or markdown syntax characters (like bracket lines).
    - Hints or placeholders: Help text that you want available but not distracting.
    - Secondary information: Paths or metadata that the user doesn't need to read at first glance.
    
None of which seems applicable to code.

arrufat:
Ah, I misunderstood you. Now dimmed is only used for bullet points, table borders, etc, never content. Thank you :)

karlseguin (approved):

arrufat:
The rendering is no longer basic

PR #2940, opened 2026-07-14
feat(webapi): reflect HTMLAreaElement attributes

rohitsux:
## What

`HTMLAreaElement` previously implemented only the URL / `HTMLHyperlinkElementUtils` mixin (`href`, `origin`, `protocol`, `host`, `pathname`, ...). Its area-specific reflected IDL attributes were missing, so `alt`, `coords`, `shape`, `target`, `download`, `rel`, `referrerPolicy` and `relList` all read back as `undefined`.

## Change

Reflect the missing attributes, following the existing `HTMLAnchorElement` / `HTMLLinkElement` implementations in the same directory:

- `alt`, `coords`, `shape`, `target`, `download`, `rel` — plain string reflection via `getAttributeSafe` / `setAttributeSafe`.
- `referrerPolicy` — enumerated reflection via `reflectEnumerated` against the referrer-policy token set (same list as `HTMLLinkElement`).
- `relList` — the shared `DOMTokenList` helper (`element.getRelList`), guarded to the html/svg namespaces, matching `HTMLAnchorElement`.

`ping` is intentionally left out to match the current `HTMLAnchorElement` precedent (which also does not implement it).

## Test

Adds `src/browser/tests/element/html/area.html`: a parse-from-HTML block asserting each reflected attribute (including `relList` via `.length` / `.item()` / `.contains()`), a defaults block for a bare `<area>`, and a `createElement('area')` round-trip block. Wired via a `test "WebApi: HTML.Area"` block in `Area.zig`.

## Verification

- `make test F="Area"` → `2 of 2 tests passed` (`WebApi: HTML.Area` green).
- `zig fmt --check` clean on the changed file.
- Custom test runner reports no leaked allocations.

### DIFF
--- a/src/browser/tests/element/html/area.html
+++ b/src/browser/tests/element/html/area.html
@@ -1,0 +1,68 @@
+<!DOCTYPE html>
+<script src="../../testing.js"></script>
+
+<map name="m">
+  <area id="a1" alt="logo" coords="0,0,82,126" shape="rect" target="_blank" download="pic" rel="noopener noreferrer" referrerpolicy="no-referrer" href="/page">
+  <area id="a2">
+</map>
+
+<script id="area-from-html">
+  {
+    const a1 = document.getElementById('a1');
+    testing.expectEqual('HTMLAreaElement', a1.constructor.name);
+    testing.expectEqual('logo', a1.alt);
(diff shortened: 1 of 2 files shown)


karlseguin:
Thanks

PR #2942, opened 2026-07-14
webapi: AbortSignal conformance fixes (dependent signals, null reason, timeout after detach)

francisbouvier:
First of a series of stacked PRs splitting the `web_api_robot-dom` WPT work into reviewable chunks. This one fixes three AbortSignal conformance issues found by the WPT `/dom/abort/` suite.

## Commits

**webapi: fix AbortSignal dependent-signal abort semantics**
- Reentrant aborts (a source aborting another source from inside an abort listener) made a shared dependent signal fire its `abort` event twice; per spec only not-yet-aborted dependents are appended to the dispatch list.
- `signal.reason` re-wrapped a fresh `DOMException` on every access, so `source.reason === dependent.reason` failed. The reason is now a single `*DOMException` allocated at abort time and identity-mapped by the JS bridge.

**webapi: keep explicit null as AbortSignal abort reason**
- `controller.abort(null)` fell back to the default `AbortError` because the JS→Zig conversion collapsed explicit `null` into a missing argument. The reason parameter now preserves the missing vs explicit-null distinction; only missing/`undefined` gets the default.

**webapi: don't fire AbortSignal.timeout() after frame detach**
- The scheduler task kept running after the iframe that created the signal was detached, so the signal aborted anyway. The callback now skips the abort when a hosting iframe in the frame's ancestor chain is no longer connected, per the spec's fully-active requirement for queued global tasks.

## WPT coverage

|---|---|---|
| /dom/abort/abort-signal-any.any.html | 11/14 | 14/14 |
| /dom/abort/abort-signal-any.any.worker.html | 11/14 | 14/14 |
| /dom/abort/event.any.html | 15/16 | 16/16 |
| /dom/abort/event.any.worker.html | 15/16 | 16/16 |
| /dom/abort/abort-signal-timeout.html | 0/1 | 1/1 |

### DIFF
--- a/src/browser/webapi/AbortSignal.zig
+++ b/src/browser/webapi/AbortSignal.zig
@@ -35,14 +35,18 @@ const Dependend = union(enum) {
     signal: *AbortSignal,
     model_context_tool: *ModelContextTool,
 
-    fn markAborted(self: Dependend, reason_: ?Reason, exec: *const Execution) !void {
+    // Returns false if the dependent was already aborted, in which case no
+    // abort event must be dispatched for it.
+    fn markAborted(self: Dependend, reason_: ?Reason, exec: *const Execution) !bool {
         switch (self) {
             .signal => |dep| {
-                if (dep._aborted) return;
+                if (dep._aborted) return false;
                 try dep.markAborted(reason_, exec);
+                return true;
             },
             .model_context_tool => |dep| {
                 try dep.markAborted(exec);
+                return true;
             },
         }
     }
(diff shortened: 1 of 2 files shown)


src/browser/webapi/AbortSignal.zig
```diff
@@ -218,11 +238,33 @@ const TimeoutCallback = struct {
...
     }
+
+    fn timeoutAbort(self: *TimeoutCallback) !void {
+        // Per spec, the abort is queued as a global task on the signal's
+        // global: it must not run when the global's document is no longer
+        // fully active (e.g. the iframe that created the signal was detached).
+        switch (self.exec.js.global) {
+            .frame => |frame| {
+                var current: ?@TypeOf(frame) = frame;
+                while (current) |f| : (current = f.parent) {
```
karlseguin, line 254: First time I've seen the need for this, if it shows up again, we should create a `frame.isConnected()` helper or something.

PR #2943, opened 2026-07-14
webapi: HTMLCollection & named-collection property semantics

francisbouvier:
Stacked PR 2/22 (based on #2942). Makes `/dom/collections` fully green (53/53) by implementing the Web IDL legacy platform object semantics for HTMLCollection, NamedNodeMap and DOMStringMap.

## Commits

**webapi: supported HTMLCollection properties can't be deleted**
- `delete collection[0]` / `delete collection.name` succeeded; supported properties are backed by the live collection, so deletion must fail (TypeError in strict mode). HTMLCollection now registers indexed/named deleter interceptors.

**webapi: enumerate collection supported properties for ownPropertyNames**
- `Object.getOwnPropertyNames` returned no interceptor-backed properties. HTMLCollection, NamedNodeMap and DOMStringMap now register indexed/named enumerators implementing the spec's supported property names (including the spec's kebab→camel conversion for `data-*`).

**js: support definer/descriptor interceptors; fix HTMLCollection expandos**
- Adds definer and descriptor interceptor callbacks to the JS bridge (new `indexedFull`/`namedIndexedFull` builders; existing call sites untouched). HTMLCollection uses them so assignments and `Object.defineProperty` on supported properties fail, descriptors report the spec attributes (`[LegacyUnenumerableNamedProperties]`), and expandos on derived objects keep working.

## WPT coverage

| Test file | Before | After |
|---|---|---|
| /dom/collections/HTMLCollection-delete.html | 2/4 | 4/4 |
| /dom/collections/HTMLCollection-live-mutations.window.html | 4/5 | 5/5 |
| /dom/collections/HTMLCollection-supported-property-names.html | 2/6 | 6/6 |
| /dom/collections/namednodemap-supported-property-names.html | 0/3 | 3/3 |
| /dom/collections/domstringmap-supported-property-names.html | 0/5 | 5/5 |
| /dom/collections/HTMLCollection-own-props.html | 4/8 | 8/8 |
| /dom/collections/HTMLCollection-supported-property-indices.html | 4/7 | 7/7 |

`/dom/collections` ends fully green: 53/53.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2944, opened 2026-07-14
webapi: listener option validation, window-reflecting handlers, exception propagation

francisbouvier:
Stacked PR 3/22 (based on #2943). Event listener/handler plumbing fixes from the `/dom/events` suite.

## Commits

**webapi: throw TypeError for addEventListener({ signal: null })**
- The options dictionary declares `signal` as non-nullable, so an explicit `null` (or any non-AbortSignal) must throw a TypeError during dictionary conversion — even when the listener is null. The bridge collapsed missing and explicit-null into the same Zig `null`; the signal member is now received as a raw `js.Value` and validated before the null-callback early return.

**webapi: window-reflecting body/frameset event handlers**
- Per HTML, `onblur/onerror/onfocus/onload/onresize/onscroll` on body and frameset are aliases for the Window's handlers. Setting the IDL attribute or the content attribute (at parse time or runtime) now stores the handler on the Window, and non-callable assignments store null per `[LegacyTreatNonObjectAsNull]`.

**js: propagate pending JS exceptions from argument conversion**
- When argument conversion runs user code that throws (e.g. a throwing `toString`), the pending exception was replaced by a generic "TypeError: invalid argument". `error.JsException` now passes through the conversion catch so the original exception surfaces.

## WPT coverage

|---|---|---|
| /dom/events/AddEventListenerOptions-signal.any.html (+ .worker) | 9/11 | 11/11 |
| /dom/events/Body-FrameSet-Event-Handlers.html | 24/48 | 48/48 |
| /dom/events/Event-constructors.any.html (+ .worker) | 13/14 | 14/14 |

## Review follow-ups

- `dba25389c` — window-reflecting body/frameset handlers now resolve the **element's owner frame** instead of the caller's (review discussion): `iframeDoc.body.onblur = cb` set from the parent realm lands on the iframe's window. Applies to all six getter/setter pairs and the content-attribute Build hooks; cross-realm unit test added (`tests/frames/window_reflecting_handlers.html`), which fails without the fix.
- `0c0e3626f` — `HTMLCollection.getIndexes` takes the injected `*Frame` (same pattern as the getNames suggestions on #2943).

WPT re-verified after the follow-ups: Body-FrameSet-Event-Handlers.html 48/48, /dom/collections 53/53, AddEventListenerOptions-signal 11/11 (both variants).

### DIFF
--- a/src/browser/webapi/Window.zig
+++ b/src/browser/webapi/Window.zig
@@ -428,7 +428,7 @@ pub fn setOnScroll(self: *Window, setter: ?FunctionSetter) void {
 // event handlers of body and frameset elements are aliases for the Window's.
 // Returns the Window storage slot for the given content attribute name, or
 // null if the attribute isn't part of the set.
-pub fn windowReflectingHandler(self: *Window, name: lp.String) ?*?js.Function.Global {
+fn windowReflectingHandler(self: *Window, name: lp.String) ?*?js.Function.Global {
     if (name.eql(comptime .wrap("onblur"))) return &self._on_blur;
     if (name.eql(comptime .wrap("onerror"))) return &self._on_error;
     if (name.eql(comptime .wrap("onfocus"))) return &self._on_focus;
--- a/src/browser/webapi/Window.zig
+++ b/src/browser/webapi/Window.zig
@@ -82,6 +82,10 @@ _on_pageshow: ?js.Function.Global = null,
 _on_popstate: ?js.Function.Global = null,
 _on_hashchange: ?js.Function.Global = null,
 _on_error: ?js.Function.Global = null,
+_on_blur: ?js.Function.Global = null,
+_on_focus: ?js.Function.Global = null,
+_on_resize: ?js.Function.Global = null,
+_on_scroll: ?js.Function.Global = null,
 _on_message: ?js.Function.Global = null,
 _on_rejection_handled: ?js.Function.Global = null,
 _on_unhandled_rejection: ?js.Function.Global = null,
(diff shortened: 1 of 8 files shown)


src/browser/webapi/element/html/Body.zig
```diff
@@ -37,15 +38,50 @@ pub fn asNode(self: *Body) *Node {
...
 }
 
-/// Special-case: `body.onload` is actually an alias for `window.onload`.
-pub fn setOnLoad(_: *Body, callback: ?js.Function.Global, frame: *Frame) !void {
-    frame.window._on_load = callback;
+// Special-case: the "window-reflecting body element event handler set"
+// (blur, error, focus, load, resize, scroll) are aliases for the Window's
+// event handlers.
+pub fn getOnBlur(_: *Body, frame: *Frame) ?js.Function.Global {
+    return frame.window._on_blur;
```
karlseguin: So, we have a pretty persistent bug in the codebase. This should probably be:

```zig
pub fn getOnBlur(self: *Body, frame: *Frame) ?js.Function.Global {
   return self.asElement().ownerFrame(frame).window._on_blur;
}
```

The `frame` parameter is the frame that owns the executing context which might not be the frame for the element. Consider:

```js
let frame = document.createElement('iframe');
frame.contentDocument.querySelector('body').onblur = cb;
```

The way you have it, this would set the `onblur` event handler for the calling code,  but I assume the correct behavior is for it to set the event handler for iframe's window.

This is an issue with every getter / setter here.

src/browser/webapi/Window.zig
```diff
@@ -388,6 +392,68 @@ pub fn setOnError(self: *Window, setter: ?FunctionSetter) void {
...
+
+pub fn setOnScroll(self: *Window, setter: ?FunctionSetter) void {
+    self._on_scroll = getFunctionFromSetter(setter);
+}
+
+// The "window-reflecting body element event handler set" (HTML spec): these
+// event handlers of body and frameset elements are aliases for the Window's.
+// Returns the Window storage slot for the given content attribute name, or
+// null if the attribute isn't part of the set.
+pub fn windowReflectingHandler(self: *Window, name: lp.String) ?*?js.Function.Global {
```
karlseguin:
```suggestion
fn windowReflectingHandler(self: *Window, name: lp.String) ?*?js.Function.Global {
```

src/browser/webapi/element/html/Body.zig
francisbouvier: Good catch — fixed in dba25389c. All Body/FrameSet getters/setters now resolve the Window through `Node.ownerFrame` (element → ownerDocument → frame, caller-frame fallback for detached documents) via a shared `reflectedWindow` helper. The `Build` hooks (parse-complete / attributeChange / attributeRemove) had the same issue for the content-attribute path and now also compile the handler body in the owner frame's realm. Added a cross-realm unit test (`tests/frames/window_reflecting_handlers.html`) covering both paths — it fails without the fix.

PR #2946, opened 2026-07-14
webapi: event interface conformance (isTrusted, constructor defaults, handleEvent)

francisbouvier:
Stacked PR 5/22 (based on #2945). Event interface conformance: `[LegacyUnforgeable]` isTrusted, constructor dictionary defaults, and callback-interface listener invocation.

## Commits

**js: support [LegacyUnforgeable] accessors; make Event.isTrusted one**
- `isTrusted` must be an own, non-configurable accessor on every Event instance (getter shared per context), not a prototype accessor. `bridge.accessor` gains an `unforgeable` option installed on the instance template, and `attachClass` re-installs ancestors' unforgeable accessors on subclasses.

**webapi: fix UIEvent.view default, initCustomEvent guard, KeyboardEvent flags**
- `new UIEvent("x").view` must be null (window fallback kept for trusted events only); `initCustomEvent` was missing the while-dispatching guard; synthetic KeyboardEvents no longer force bubbles/cancelable/composed to true.

**webapi: accept arbitrary button/location values; legacy keyboard init members**
- `MouseEvent.button` is a WebIDL short and `KeyboardEvent.location` an unsigned long — arbitrary values are allowed, so they're stored as raw integers instead of enums (named constants remain). KeyboardEvent supports the legacy `charCode`/`keyCode` init members; `UIEvent.which` follows.

**webapi: report TypeError for non-callable handleEvent listeners**
- Per the callback-interface invocation steps, a non-callable `handleEvent` reports a TypeError to the global instead of being silently skipped; a throwing getter is reported like a throwing listener.

## WPT coverage

|---|---|---|
| /dom/events/Event-isTrusted.any.html (+ .worker) | 0/1 | 1/1 |
| /dom/events/Event-init-while-dispatching.html | 1/5 | 5/5 |
| /dom/events/Event-subclasses-constructors.html | 16/49 | 49/49 |
| /dom/events/EventListener-handleEvent.html | 4/6 | 6/6 |


karlseguin:
@francisbouvier I personally feel that Claude is too verbose with comments. In 4 places, it more or less explained what `[LegacyUnforgeable]` was..and, in my opinion, the only place this should have gone (bridge.unforgeable) had nothing.

Similarly, `// Legacy KeyboardEventInit values, reported as-is by synthetic events.` was duplicated at both the fields (makes sense) and when set in the constructor (odd).

Anyways, I adjusted the comments to my preference. (along with a slight refactor to how the snapshot applies unforgeable to children). But if you agree, it might be worth adding that hint to your claude for future cases.

francisbouvier:
> @francisbouvier I personally feel that Claude is too verbose with comments. In 4 places, it more or less explained what `[LegacyUnforgeable]` was..and, in my opinion, the only place this should have gone (bridge.unforgeable) had nothing.
> 
> Similarly, `// Legacy KeyboardEventInit values, reported as-is by synthetic events.` was duplicated at both the fields (makes sense) and when set in the constructor (odd).
> 
> Anyways, I adjusted the comments to my preference. (along with a slight refactor to how the snapshot applies unforgeable to children). But if you agree, it might be worth adding that hint to your claude for future cases.

Yes of course. I'm adding that in the instructions.

PR #2967, opened 2026-07-15
agent: split Terminal.zig into picker and prompt_assist

arrufat:
Terminal.zig had grown to ~1450 lines holding four subsystems: readline lifecycle, output printing, the isocline prompt assistance, and a choice picker. This splits it along those seams:

- **picker.zig** — the numbered/interactive choice picker. `settings.zig` (pre-REPL provider selection) no longer imports the isocline-configured Terminal. A tty that refuses raw mode now degrades to the line prompt instead of leaking termios errors.
- **prompt_assist.zig** — completion, ghost hints, and prompt highlighting. The `ps-*` style palette is one table that drives both `ic_style_def` registration and the token-kind mapping, so an unregistered-but-painted style is now a compile error (previously it silently rendered unstyled). Tab path completion and the ghost path hint share one `PathMatchIterator`.
- `all_names` + `closestCommand` move to SlashCommand.zig (the command registry); `columns()` moves into Spinner, its only caller, breaking the Spinner↔Terminal import cycle.

Terminal.zig keeps what its name says — readline, log sink, printing — at ~420 lines. Mostly code motion; each step is its own commit.


src/agent/picker.zig
```diff
@@ -0,0 +1,239 @@
...
+}
+
+/// Line-oriented fallback. Errors with NoChoice after 3 invalid attempts.
+fn promptNumberedChoiceLine(header: []const u8, items: []const [:0]const u8, default: ?usize) !usize {
+    var stdin_buf: [128]u8 = undefined;
+    var stdin = std.fs.File.stdin().reader(&stdin_buf);
+
+    var attempt: u8 = 0;
+    while (attempt < 3) : (attempt += 1) {
+        std.debug.print("{s}\n", .{header});
```
karlseguin, line 52: I see this is all existing code in Terminal.zig, but should we be using std.debug? I don't mind, but in most other places your using a stdout/stderr, usually through a Writer.

karlseguin (approved):

src/agent/picker.zig
arrufat: You're right, I will fix for consistency.

Issue #1736, opened 2026-03-07
Complete Chrome DevTools Accessibility domain support


mgifford:
Sorry about the AI generated issue. 

Problem

Accessibility tooling such as:

- axe-core
- Lighthouse
- Playwright accessibility snapshot
- Puppeteer accessibility APIs

depends on the Chrome DevTools Protocol (CDP) Accessibility domain to inspect the browser accessibility tree.

Many automated accessibility workflows assume the same APIs exposed by Chromium. Without them, these tools cannot analyze the accessibility tree or run accessibility checks that rely on accessibility tree semantics.

Providing compatibility with the CDP Accessibility domain would allow existing accessibility tooling to work with Lightpanda without modification.

Proposed improvement

Implement the Chrome DevTools Protocol Accessibility domain so that Lightpanda exposes an accessibility tree compatible with existing automation tools.

Relevant methods include:
- Accessibility.enable
- Accessibility.getFullAXTree
- Accessibility.getPartialAXTree
- Accessibility.getAXNode
- Accessibility.queryAXTree

Expected node properties should include fields typically returned by Chromium, such as:

- role
- name
- description
- value
- properties
- states
- ignored
- childIds
- backendDOMNodeId

Expected result

Accessibility inspection should work with standard browser automation tooling.

Example using Playwright:

`const snapshot = await page.accessibility.snapshot();
console.log(snapshot);

Example using Puppeteer:

const axTree = await client.send('Accessibility.getFullAXTree');
console.log(axTree);
`

These APIs are widely used by accessibility tooling and research tools to inspect accessibility trees.

Why this matters

Compatibility with the CDP Accessibility domain would allow Lightpanda to support existing accessibility testing ecosystems, including:

- axe-core integrations
- Lighthouse accessibility audits
automated accessibility scanning pipelines
- accessibility research and large-scale measurement projects

Reference

Chrome DevTools Protocol — Accessibility domain
https://chromedevtools.github.io/devtools-protocol/tot/Accessibility/

PR #2968, opened 2026-07-15
feat(cdp): implement Accessibility.getPartialAXTree

rohitsux:
Toward #1736.

## What

The Accessibility CDP domain implemented `getFullAXTree` and `queryAXTree` but not `getPartialAXTree`, so a client requesting the accessibility subtree rooted at a specific node received `UnknownMethod`. Accessibility tooling (Playwright/Puppeteer, axe-core, Lighthouse) uses `getPartialAXTree` for scoped a11y snapshots.

## Change

One method in `src/cdp/domains/accessibility.zig`, built from the two existing siblings: resolve the target node via `dom.getNode` (the node resolution `queryAXTree` uses — `nodeId` / `backendNodeId` / `objectId`), and emit its accessibility subtree via `bc.axnodeWriter` with no filter (the unfiltered shape `getFullAXTree` produces for the document root).

`fetchRelatives` is accepted for Chrome protocol compatibility; the current scope is the subtree rooted at the resolved node, matching `queryAXTree`. Emitting the ancestor chain is a natural follow-up.

Partial step toward the umbrella issue #1736; does not close it.

## Test

Two tests mirroring the existing `queryAXTree` coverage: `getPartialAXTree` with no node identifier returns `MissingParams`, and with an unknown `nodeId` returns `NodeNotFound`. `make test F="accessibility"` -> 4/4 pass; `zig fmt --check` clean.

### DIFF
--- a/src/cdp/domains/accessibility.zig
+++ b/src/cdp/domains/accessibility.zig
@@ -17,9 +17,14 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 const std = @import("std");
+const lp = @import("lightpanda");
+
 const id = @import("../id.zig");
 const CDP = @import("../CDP.zig");
+
 const dom = @import("dom.zig");
+
+const log = lp.log;
 
 pub fn processMessage(cmd: *CDP.Command) !void {
     const action = std.meta.stringToEnum(enum {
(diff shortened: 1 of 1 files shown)


karlseguin:
Thanks

PR #2969, opened 2026-07-15
webapi: add `Request#formData` and `Response#formData`

nikneym:
This PR adds `Request#formData` and `Response#formData` methods. also introduces `header_parser.zig` for reusable HTTP-related parsers, idea is to move away from one-off parsers created at each file to more generalized ones here.


karlseguin:
I'm not smart enough to review this. Maybe @krichprollsch can take a look.  Claude did find some issues though:

  Confirmed bugs (verified by running the built binary)

  1. Bare Zig errors leak to JS as generic Error, sometimes synchronously.
  Request.formData does self._body orelse return error.InvalidFormData and both methods do
  ... orelse return error.InvalidFormData when Content-Type is absent
  (Request.zig:311,314, Response.zig:471). I probed this live:

  - new Response().formData() → synchronously throws Error: InvalidFormData. Spec requires
  promise-returning methods to always reject, and with TypeError. The sibling methods
  (json rejects via rejectPromise) get this right; formData is the odd one out.
  - Request with urlencoded Content-Type and no body → throws Error: InvalidFormData. Per
  Fetch, a null body is an empty byte sequence: this must resolve with an empty FormData
  (Chrome does).

  2. The parser rejects bodies its own encoder produces. header_value_map marks HTAB
  (0x09) invalid, and parseHttpHeader skips only ' ' as leading OWS — but
  writeMultipartName escapes only CR/LF/". Verified live: fd.append('a\tb', 'v') encodes
  fine, then req.formData() rejects with TypeError. Chrome round-trips this. RFC 9110
  explicitly permits HTAB in field values and as OWS, so this is wrong even for the
  "general HTTP parser" ambition.

  3. parseDisposition hard-fails on any unquoted parameter. Any key=value without quotes →
  error.InvalidFormData for the entire body. That includes filename*=UTF-8''... (RFC
  5987), which real servers emit despite RFC 7578 discouraging it — and which, ironically,
  is the exact example string used in simd.zig's own parseHttpHeader test. Chrome skips
  unknown/malformed params; this parser torpedoes the whole parse. Response.formData()
  parses arbitrary server output, so lenience matters here.
  
  And claude's comment on the code:
  
- src/simd.zig is named for its implementation technique, not its domain — and half of
it (parseDisposition, ContentTypeIterator, HttpHeader) contains no SIMD at all. It's a
MIME/HTTP-parsing module. The codebase already has src/browser/Mime.zig as the
content-type home (to be fair: Mime.parse lowercases input and couldn't extract a
case-sensitive boundary as-is, so ContentTypeIterator isn't pure duplication — but the
codebase now has two content-type parsers in two unrelated locations).

- The generality is speculative. The PR description says the idea is to unify "one-off
parsers created at each file" — but no existing call-site was migrated; simd.zig's only
consumer is this feature. parseHttpHeader's error.Incomplete streaming contract is
designed for push-style incremental parsing that nothing uses (multipart always has the
full buffer). Earlier commits on the branch even deleted a dead parseHttpHeaders
function. This collides with the project's no-speculative-generality lean.

- The SIMD is optimizing a cold path. These parsers run when a page calls formData() on
an in-memory body — small part headers, parsed once. For that, the file maintains three
parallel strategies (vectors, SWAR, scalar tables), with inconsistent choices between
two nearly identical functions (matchHeaderKey: 16-byte vectors, "SWAR is not preferred
here"; matchHeaderValue: 32-byte vectors plus a SWAR path). Meanwhile
FormData.indexOfSpecial is behaviorally identical to its own fallback line —
std.mem.indexOfAnyPos(u8, slice, 0, "%+") — the entire function could be that one call.

- Micro-obscurities: broadcast derives bit width via @ctz(@as(T, 0)) instead of
@bitSizeOf; block_size = @sizeOf(usize) is commented "Block size of the CPU" (it's
register width); u16-bitcast compares for two-byte prefixes where mem.eql reads clearer.

- header_key_map is simultaneously stricter than needed (rejects nothing a browser
emits) and laxer than RFC token (accepts ", ,, ;-adjacent chars, and all bytes ≥ 0x80 in
header names) — asymmetric with the value map being too strict about tabs.

src/browser/webapi/net/FormData.zig
```diff
@@ -388,6 +430,255 @@ fn writeMultipartName(writer: *std.Io.Writer, name: []const u8) !void {
...
+        // Got a file.
+        if (parsed.filename) |filename| {
+            const blob = try Blob.initFromBytes(content, content_type, page);
+            errdefer blob.deinit(page);
+
+            const file = try blob._arena.create(File);
+            file.* = .{
+                ._proto = blob,
+                ._name = try blob._arena.dupe(u8, try decodeMultipartName(self._arena, filename)),
+                ._last_modified = std.Io.Clock.now(.real, lp.io).toMilliseconds(),
```
karlseguin: Just for the sake of standardizing, `@intCast(lp.datetime.milliTimestamp(.real))`

src/network/header_parser.zig
```diff
@@ -0,0 +1,539 @@
...
+inline fn isHeaderKeyByte(byte: u8) bool {
+    return switch (byte) {
+        0...' ', ':', 0x7f => false,
+        else => true,
+    };
+}
+
+/// Whether a byte is allowed in a header value; controls (excluding space)
+/// and DEL are excluded.
+inline fn isHeaderValueByte(byte: u8) bool {
```
karlseguin: Claude still insists HTAB should be allowed here.

src/network/header_parser.zig
```diff
@@ -0,0 +1,539 @@
...
+            8 => b * 0x01,
+            16 => b * 0x01_01,
+            32 => b * 0x01_01_01_01,
+            64 => b * 0x01_01_01_01_01_01_01_01,
+            else => @compileError("unexpected broadcast size"),
+        };
+    }
+}
+
+inline fn matchHeaderKey(cursor: *Cursor) void {
```
karlseguin: > Note that inline actually restricts what the compiler is allowed to do. This can harm binary size, compilation speed, and even runtime performance.

So unless we _have_ to inline (e.g. to force parameter comptime-ness) I think these should all be removed.

src/network/header_parser.zig
```diff
@@ -0,0 +1,539 @@
...
+    }
+
+    /// Peek the current and the next but don't advance.
+    /// SAFETY: This function doesn't check if out of bounds reachable.
+    pub inline fn peek2(cursor: *const Cursor, c0: u8, c1: u8) bool {
+        return cursor.asInteger(u16) == @as(u16, @bitCast([2]u8{ c0, c1 }));
+    }
+
+    /// Moves the cursor until no leading spaces there are.
+    pub inline fn skipSpaces(cursor: *Cursor) void {
```
karlseguin: Same as `isHeaderValueByte`, claude thinks tabs should be skipped.

PR #2970, opened 2026-07-15
Dispose Event for `NavigationHistoryEntry`

mookums:
This adds the `dispose` event for the `NavigationHistoryEntry`
https://developer.mozilla.org/en-US/docs/Web/API/NavigationHistoryEntry/dispose_event

### DIFF
--- a/src/browser/webapi/navigation/Navigation.zig
+++ b/src/browser/webapi/navigation/Navigation.zig
@@ -195,13 +195,13 @@ pub fn pushEntry(
 
     // truncates our history here.
     const retained_index = self._index + 1;
+    var disposed: []*NavigationHistoryEntry = &.{};
     if (self._entries.items.len > retained_index) {
-        const disposed = try frame.call_arena.dupe(
+        disposed = try frame.call_arena.dupe(
             *NavigationHistoryEntry,
             self._entries.items[retained_index..],
         );
         self._entries.shrinkRetainingCapacity(retained_index);
-        for (disposed) |d| try d.fireDispose(frame);
     }
 
     const index = self._entries.items.len;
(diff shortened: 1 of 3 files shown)


src/browser/webapi/navigation/NavigationHistoryEntry.zig
```diff
@@ -73,6 +80,25 @@ pub fn getState(self: *const NavigationHistoryEntry, frame: *Frame) !StateReturn
     return .undefined;
 }
 
+pub fn fireDispose(self: *NavigationHistoryEntry, frame: *Frame) !void {
+    const cec = self._on_dispose orelse return;
```
krichprollsch: You can have a registered event listener w/o on_dispose callback set. In this case, you won't dispatch the event to the listener...

You should use `if (!frame.hasDirectListeners(target, "dispose", self._on_dispose)) return;` instead

And `cec` looks like a strange choice for the var name 🤔

src/browser/webapi/navigation/Navigation.zig
```diff
@@ -56,10 +56,16 @@ fn asEventTarget(self: *Navigation) *EventTarget {
 
 pub fn onRemoveFrame(self: *Navigation) void {
     self._proto = undefined;
+    for (self._entries.items) |entry| {
+        entry._proto = undefined;
```
krichprollsch, line 65: You should also clear `_on_dispose` (and `on_currententrychange`) b/c the page which persisted the calbacks is removed.

src/browser/webapi/navigation/Navigation.zig
```diff
@@ -184,6 +190,9 @@ pub fn pushEntry(
 
     // truncates our history here.
     if (self._entries.items.len > self._index + 1) {
+        for (self._entries.items[self._index + 1 ..]) |disposed| {
+            try disposed.fireDispose(frame);
```
krichprollsch: There is a re-entrancy risk here: a listener can call a navigation. Claude suggest to duplicate dispose and cleanup entries before dispatching the event:

```zig
  if (self._entries.items.len > self._index + 1) {
      const disposed = try frame.call_arena.dupe(*NavigationHistoryEntry, self._entries.items[self._index + 1 ..]);
      self._entries.shrinkRetainingCapacity(self._index + 1);
      for (disposed) |d| try d.fireDispose(frame);
  }
```

src/browser/webapi/navigation/NavigationHistoryEntry.zig
```diff
@@ -73,6 +80,25 @@ pub fn getState(self: *const NavigationHistoryEntry, frame: *Frame) !StateReturn
...
+    try frame.dispatch(self.asEventTarget(), event, cec, .{ .context = "NavigationHistoryEntry" });
+}
+
+fn getOnDispose(self: *const NavigationHistoryEntry) ?js.Function.Global {
+    return self._on_dispose;
+}
+
+pub fn setOnDispose(self: *NavigationHistoryEntry, listener: ?js.Function) !void {
+    if (listener) |listen| {
+        self._on_dispose = try listen.persistWithThis(self);
```
krichprollsch: I think if `_on_dispose` is not null, you should release the previous value first.
similar to https://github.com/lightpanda-io/browser/blob/main/src/browser/webapi/net/WebSocket.zig#L638

src/browser/webapi/navigation/NavigationHistoryEntry.zig
mookums: A lot of this was adapted from another event handler that is in Navigation.

PR #2973, opened 2026-07-15
crash: increase HTMLCollection's RC from u8 to u32

karlseguin:


mookums (approved):
Seems to fix the crash for me too.

PR #2981, opened 2026-07-16
cli: page help output through $PAGER when stdout is a tty

arrufat:
## Why

`lightpanda help agent` prints ~225 lines (the agent section plus all common options) straight to the terminal, clobbering the screen. Help also went to **stderr** (via `std.debug.print`), so `lightpanda help | grep proxy` captured nothing at all.

## What

- Explicitly requested help (`help`, `help <cmd>`, `--help`) now goes to **stdout**, following the usual CLI convention.
- When stdout is an interactive terminal (and `TERM` isn't dumb/empty), the help text is piped through `$PAGER` (run via `/bin/sh -c`, like git) or `less -FIRX` as fallback. `-F` makes less exit immediately when the text fits one screen, so short help like the general `lightpanda help` looks exactly as before.
- Piped or redirected output stays plain — no pager involved, scripting against stdout works normally.
- Error-path usage (bad args) remains on **stderr**, unpaged, exit code 1 — unchanged.
- A pager that exits early or fails to start can't lose the help: writes to a dead pager surface as `error.BrokenPipe` (std already installs a noop SIGPIPE handler at startup), and a non-zero pager exit (e.g. `$PAGER` points at a missing command, sh exits 127) falls back to printing the help plainly.

## Behavior change note

Anything that captured help via `2>&1` or expected it on stderr will now find it on stdout. Error usage output is untouched.

### DIFF
--- a/src/Config.zig
+++ b/src/Config.zig
@@ -714,7 +714,7 @@ pub const HttpHeaders = struct {
     }
 };
 
-pub fn printUsageAndExit(self: *const Config, help_for: RunMode, success: bool) void {
+pub fn printUsageAndExit(self: *const Config, allocator: Allocator, help_for: RunMode, success: bool) !void {
     const exec_name = self.exec_name;
     const Help = @import("help.zon");
     const is_debug = builtin.mode == .Debug;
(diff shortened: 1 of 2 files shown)


src/Config.zig
```diff
@@ -714,44 +714,103 @@ pub const HttpHeaders = struct {
...
+    var child = std.process.Child.init(argv, allocator);
+    child.stdin_behavior = .Pipe;
+    child.stdout_behavior = .Inherit;
+    child.stderr_behavior = .Inherit;
+    child.spawn() catch return printPlain(text);
+
+    // The pager can exit before we finish writing (e.g. `q` in less, or
+    // `sh -c` failing to find $PAGER); without this the resulting SIGPIPE
+    // would kill us. The process exits right after, so no need to restore.
+    std.posix.sigaction(std.posix.SIG.PIPE, &.{
```
krichprollsch: Claude thinks this is redundant w/ stdlib
```
  src/Config.zig:787-794 installs SIG_IGN for SIGPIPE, but Zig already does this for you. std.start.maybeIgnoreSigpipe() runs at startup and installs a SIGPIPE handler unless std.options.keep_sigpipe is set — and src/main.zig:30 only sets
  crypto_always_getrandom, so the default (false) applies. A write to a dead pager already returns error.BrokenPipe rather than killing the process, which the catch {} at line 800 already handles. The premise in the comment ("without this the resulting
  SIGPIPE would kill us") isn't true for this program.

  It's also a small regression in intent. Zig's std deliberately uses a no-op handler rather than SIG_IGN, with this comment in start.zig:

  ▎ Set handler to a noop function instead of SIG.IGN to prevent leaking signal disposition to a child process.

  Ignored dispositions survive exec, so SIG_IGN is exactly what std goes out of its way to avoid. There's no live bug here — the sigaction lands after spawn(), so the pager doesn't inherit it, and the process exits shortly after — but the block costs
  three lines of subtle signal code and buys nothing.
```

arrufat: Thank you!

PR #2984, opened 2026-07-17
ci: pin github actions to commit shas

Totara-thib:
Hi, following up on the security email from June, Pierre said a PR was welcome so here it is.

This pins every action reference in the workflows and in the two composite actions to a full commit sha. The version each sha corresponds to is kept as a comment, so Renovate or Dependabot can keep tracking updates.

Why it matters: a tag like `@v6` is a movable pointer. Whoever controls the action, or anyone who compromises it, can re-point the tag and your next run executes their code with the job's token and secrets. The release path was the sensitive spot here, `ncipollo/release-action` runs with your release publishing token. Same pattern as the `tj-actions/changed-files` incident (CVE-2025-30066).

A few notes on specific pins:

- `dtolnay/rust-toolchain` is pinned to a master commit with the toolchain moved to an explicit `toolchain: stable` input, which is what its readme recommends for sha pinning.
- `contributor-assistant/github-action` is pinned too, but heads up: its upstream repo was archived in March, so no more security fixes are coming. Probably worth replacing at some point. I did not touch your CLA flow, that call is yours.
- `Swatinem/rust-cache` has a `v2` tag that currently points past its latest release, so it is pinned to the v2.9.1 release commit instead.

Every sha was resolved from the upstream repos and cross checked against their tags. No workflow logic changes, only the `uses:` lines.

### DIFF
--- a/.github/actions/install/action.yml
+++ b/.github/actions/install/action.yml
@@ -39,17 +39,19 @@ runs:
         sudo apt-get install -y wget xz-utils ca-certificates clang make git
 
     # Zig version used from the `minimum_zig_version` field in build.zig.zon
-    - uses: mlugg/setup-zig@v2
+    - uses: mlugg/setup-zig@d1434d08867e3ee9daa34448df10607b98908d29 # v2.2.1
 
     # Rust Toolchain for html5ever
-    - uses: dtolnay/rust-toolchain@stable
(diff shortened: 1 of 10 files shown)


Totara-thib:

krichprollsch:
thanks :pray:

PR #2985, opened 2026-07-17
ci: add plumber to watch workflow security

Totara-thib:
Companion to https://github.com/lightpanda-io/browser/pull/2984, merge that one first: the check added here flags the unpinned actions until the pins land, then it goes green.

This adds [Plumber](https://github.com/getplumber/plumber) to CI, the tool I used to find the unpinned actions in the first place. It scans the workflows on each push to main and on each PR, and fails when something regresses: an unpinned action, an archived dependency, a known CVE, that kind of thing.

- `plumber.yml`: pinned by sha, minimal permissions, findings go to the security tab as SARIF (skipped on PRs from forks, the report stays as an artifact there).
- `.plumber.yaml`: generated defaults scoped to this repo, the few controls that are off have a note in the file explaining why.
- `README.md`: one line, a live score badge next to your existing ones. Runs on main publish the score through OIDC, no secret. It reads UNKNOWN in gray until the first run, then shows the grade. Scores are public. Not interested? Drop the line and the `score-push` input, the rest works the same.

With the pins in, this runs green with a score of A. Set `soft-fail: true` if you prefer report only, without gating PRs.

To be fully transparent: I work on Plumber. If you do not want the tool in your CI, no hard feelings, the pinning PR is the one that matters and it stands on its own.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,6 @@ Not a Chromium fork. Not a WebKit patch. A new browser, written in Zig.
 [![Twitter Follow](https://img.shields.io/twitter/follow/lightpanda_io)](https://twitter.com/lightpanda_io)
 [![GitHub stars](https://img.shields.io/github/stars/lightpanda-io/browser)](https://github.com/lightpanda-io/browser)
 [![Discord](https://img.shields.io/discord/1391984864894521354?style=flat-square&label=discord)](https://discord.gg/K63XeymfB5)
-[![Plumber Score](https://score.getplumber.io/github.com/lightpanda-io/browser.svg)](https://score.getplumber.io/github.com/lightpanda-io/browser)
 
 </div>
 <div align="center">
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -14,6 +14,7 @@ Not a Chromium fork. Not a WebKit patch. A new browser, written in Zig.
 [![Twitter Follow](https://img.shields.io/twitter/follow/lightpanda_io)](https://twitter.com/lightpanda_io)
 [![GitHub stars](https://img.shields.io/github/stars/lightpanda-io/browser)](https://github.com/lightpanda-io/browser)
 [![Discord](https://img.shields.io/discord/1391984864894521354?style=flat-square&label=discord)](https://discord.gg/K63XeymfB5)
+[![Plumber Score](https://score.getplumber.io/github.com/lightpanda-io/browser.svg)](https://score.getplumber.io/github.com/lightpanda-io/browser)
```
```suggestion
```

I prefer keeping the less badges as possible. Moreover it seems the rendering size doesn't match others tags:

Image

krichprollsch (commented):
I'm OK to make a try in the CI, but at the condition to remove the badge.
If you agree, let's re-discuss later the possibility to mention Plumber somewhere in the readme if we are happy w/ the CI result.

README.md
Totara-thib: Thank for the report, we are currently working on the subject to make it compliant with other badges.

Would love to have feedback concerning the Plumber action workflow, keep me updated if you have any suggestions.

PR #2986, opened 2026-07-17
cache: Never cache the top-level root page

karlseguin:
The idea is that, for most cases, the top level page isn't revisited, so caching is just overhead + wasted space. Plus, if someone does revisit it, chances are they want a fresh copy.

Imagine someone scraping a typical ecommerce site. Sure, they want to cache the .js files which get re-used from product to product. But do they want to cache each individual product page?


karlseguin:
@krichprollsch https://github.com/lightpanda-io/demo/pull/209

karlseguin:
Seems like maybe a CI cache issue now. Gonna close and re-open to see if it magically works.

Issue #2987, opened 2026-07-17
serve: HTTP cache makes CDP page loads ~6x slower on cache hits (agent path unaffected)


arrufat:
With `--http-cache-dir`, the serve/CDP path gets dramatically *slower* — and cache **hits are slower than misses** — while the agent path on the same binary, same fixture, same cache code speeds up exactly as expected. This currently blocks any enable-cache-by-default decision: with the cache on, the agent path beats headless Chrome on every task we benchmark, but the serve path regresses.

## Repro (one command, fully offline, ~30 s)

https://github.com/lightpanda-io/agent-benchmarks/tree/main/pandascript-vs-cdp/repro-serve-cache

```bash
LIGHTPANDA=/path/to/lightpanda ./run.sh
```

Workload: 6 pages, each including the **same three** script assets served with `Cache-Control: public, max-age=3600` and a 30 ms artificial server delay. The walk is `goto` × 6 (wait: load), elapsed measured inside the walk (process startup excluded). Cache dir persists across the 3 trials of a cell, so trial 1 is a cold cache (misses + writes), trials 2–3 are warm (all hits).

## Observed (1.0.0-dev.8162+a59abc7f, linux x86_64)

| cell | trial 1 | trial 2 | trial 3 | |
|---|---:|---:|---:|---|
| agent nocache | 206 | 204 | 201 | |
| agent cache | 45 | 12 | 12 | warm hits ~17× faster |
| serve nocache | 210 | 212 | 210 | |
| serve cache | 1,059 | 1,226 | 1,228 | **~6× slower; warm hits slower than cold misses** |

## Notes

- Warm-cache trials being the *slowest* rules out cache I/O cost: the serve path pays ~170 ms per page **on hits**, on localhost. That's consistent with a per-response wait on something cache-served responses never deliver (network-event/timeout interaction in the CDP layer?).
- First seen as a live-site anomaly in the benchmark campaign: on asset-light HN, `--http-cache-dir` made puppeteer/playwright→serve **+5…+13% slower** while the agent path on the same site got −29…−31% faster (full matrix in [RETAIL-INVESTIGATION.md, Postscripts 1 & 3](https://github.com/lightpanda-io/agent-benchmarks/blob/main/pandascript-vs-cdp/RETAIL-INVESTIGATION.md)). The offline fixture amplifies the effect and removes the network as a variable.
[...]

PR #2989, opened 2026-07-18
perf: If a message is dispatched, don't poll

karlseguin:
Polling delays processing of messages. So don't poll after we know we've dispatched. Largely impacts CDP which is more likely to poll (for data on the CDP connection).

### DIFF
--- a/src/network/HttpClient.zig
+++ b/src/network/HttpClient.zig
@@ -576,16 +576,16 @@ pub fn _tick(self: *Client, timeout_ms: u32, mode: DrainMode) !void {
         return error.ClientDisconnected;
     }
 
-    self.dispatchCompleted(mode);
+    const dispatched = self.dispatchCompleted(mode);
 
     try self.startPending();
 
     const running = try self.handles.perform();
 
     const processed = try self.processMessages();
-    if (processed == false and self.dispatch_queue.first == null and self.ws_dispatch_queue.first == null) {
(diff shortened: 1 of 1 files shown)


arrufat:
Verified against the offline repro from #2987 ([repro-serve-cache](https://github.com/lightpanda-io/agent-benchmarks/tree/main/pandascript-vs-cdp/repro-serve-cache), 6 pages × 3 shared cacheable scripts, 30 ms server delay each; trial 1 = cold cache, 2–3 = warm):

| cell | main (`a59abc7f`) | this PR (`20cde7a6`) |
|---|---|---|
| serve nocache | 210, 212, 210 ms | 246, 243, 245 ms |
| serve cache | **1,059, 1,226, 1,228 ms** | **79, 23, 18 ms** |
| agent nocache | 206, 204, 201 ms | 237, 230, 221 ms |
| agent cache | 45, 12, 12 ms | 70, 11, 14 ms |

The serve-path regression is gone: warm cache hits go from ~1,200 ms (6× slower than no cache) to 18–23 ms (~13× faster). This matches the diagnosis in #2987 — cache-served responses are delivered via `dispatchCompleted` with no curl I/O behind them, so the old wait condition (which only consulted `processed`) parked the tick in the I/O poll after every cache hit.

Also reproduced identically on the `sqlite-cache` branch before this fix, so the fix covers both cache backends. Closes the blocker noted in #2987 for the cache-by-default discussion.

PR #2990, opened 2026-07-18
Add per-CDP control for Network.setCacheDisabled

staylor:
## Summary

- add `LP.configureCDP` with a `disableSetCacheDisabled` option
- scope the option to the current CDP connection and keep existing behavior by default
- allow browser-target CDP sessions to be created before a browser context exists
- acknowledge but ignore `Network.setCacheDisabled` while the option is enabled

## Usage

```js
const client = await browser.newBrowserCDPSession();
await client.send("LP.configureCDP", {
  disableSetCacheDisabled: true,
});
```

Sending the option as `false` restores the standard handling of subsequent `Network.setCacheDisabled` commands.

## Motivation

Automation libraries can send `Network.setCacheDisabled(true)` as a side effect of enabling request interception. That makes a configured HTTP cache ineffective even when the caller wants interception and caching to coexist.

The custom command provides an explicit, per-connection override without adding a library-specific CLI flag or changing the default cache policy.

## Validation

- added unit coverage for configuration toggling and both cache-disable paths
- added coverage for creating a browser-target session before a browser context
- `make test` reached the final local link step, which is blocked on this machine by a Nix/macOS framework search-path mismatch; CI provides the authoritative full run

### DIFF
--- a/src/cdp/CDP.zig
+++ b/src/cdp/CDP.zig
@@ -82,6 +82,7 @@ browser_context_id_gen: BrowserContextIdGen = .{},
 browser_session_id: ?[]const u8 = null,
 
 browser_context: ?BrowserContext,
+disable_set_cache_disabled: bool = false,
 
 // Re-used arena for processing a message. We're assuming that we're getting
 // 1 message at a time.
--- a/src/cdp/domains/lp.zig
+++ b/src/cdp/domains/lp.zig
@@ -44,6 +44,7 @@ pub fn processMessage(cmd: *CDP.Command) !void {
         scrollNode,
         waitForSelector,
         handleJavaScriptDialog,
+        configureCDP,
         configureLoading,
         version,
     }, cmd.input.action) orelse return error.UnknownMethod;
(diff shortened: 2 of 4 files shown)


krichprollsch:
@staylor thanks for the PR and the issue. I didn't know playwright disable cache when request interception are set.
I'm not sure with this new flag which is very specific to this playwright's behavior...

What if we add instead a custom `LP.configureCDP` with something like `disableSetCacheDisabled: true` you could call with:
```js
const client = await browser.newBrowserCDPSession();
await client.send('LP.configureCDP', {
  disableSetCacheDisabled: true,
});
```

staylor:
> What if we add instead a custom `LP.configureCDP`

@krichprollsch I'll take a pass at it

staylor:
@krichprollsch Implemented—thanks for the suggestion. The PR now adds the per-connection command:

```js
const client = await browser.newBrowserCDPSession();
await client.send("LP.configureCDP", {
  disableSetCacheDisabled: true,
});
```

The default `Network.setCacheDisabled` behavior is unchanged, and sending `false` restores normal handling for subsequent commands. I also updated browser-target sessions so this configuration can be applied before creating a browser context, and removed the interception-specific CLI flag.

krichprollsch (approved):

krichprollsch:
thanks!

PR #3002, opened 2026-07-20
CORS Implementation

### LINKED ISSUE #2015: Implement CORS mechanism
https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CORS

mookums:
This adds a CORS implementation as experimental feature for now.
Enable it with `--experimental-feature cors`

fix #2015


mookums:
Not ready yet, just want CI to start running on this branch.

src/browser/URL.zig
```diff
@@ -341,6 +341,11 @@ pub fn getOrigin(allocator: Allocator, raw: [:0]const u8) !?[]const u8 {
     return raw[0..authority_end];
 }
 
+pub fn isSameOrigin(url: [:0]const u8, origin: [:0]const u8) bool {
```
karlseguin: `Frame.isSameOrigin` should use this (there's a comment there that assumes Protocols are equal, but I'm not sure why that's true).

src/network/HttpClient.zig
```diff
@@ -568,6 +573,14 @@ pub fn newRequest(self: *Client, req: Request, owner: ?*Owner) anyerror!*Transfe
             owned.credentials = try arena.dupeZ(u8, c);
         }
 
+        if (req.authored_headers.len > 0) {
```
karlseguin: The callsite dupes these in call_arena, and then we dupe it again in transfer.arena. We should rework headers before landing this.`Request.headers` is a micro-optimization which put the headers in a curl-friendly list from the get go. It was always questionable, and if we're going to iterate + dupe + iterate + dupe, I think it would both be cleaner and more efficient to:
1 - A single header list
2 - The header has a authored: bool flag
3 - The caller sets the headers _after_ `newRequest` is called so that transfer.arena is available

src/network/HttpClient.zig
```diff
@@ -890,6 +913,26 @@ fn pipeline(self: *Client, transfer: *Transfer, from: SubmitFrom) !void {
...
 }
 
+fn addOriginHeader(transfer: *Transfer) !void {
+    const req = &transfer.req;
+    const is_cross_origin = !URL.isSameOrigin(req.cookie_origin, req.url);
+    const is_unsafe_method = switch (req.method) {
+        .GET, .HEAD => false,
+        else => true,
+    };
+    if (!is_cross_origin and !is_unsafe_method) {
```
karlseguin: There's two different types of double negative going on here. They can all be eliminated.

```zig
const is_same_origin = URL.isSameOrigin(req.cookie_origin, req.url);
const is_safe_method = switch (req.method) {
    .GET, .HEAD => true,
    else => false,
};

if (is_same_origin and is_safe_method) {
    return;
}
```

src/network/RobotsGate.zig
```diff
@@ -77,94 +73,80 @@ pub fn check(self: *RobotsGate, transfer: *Transfer) !Result {
...
-    // submit fails it fires error_callback — possibly synchronously, right
-    // here — which resolves the waiters (fail-open, may already have resumed
-    // `transfer`) and releases the arena. So there is nothing to unwind
-    // locally and the errdefers above must not run: swallow the error.
-    fetch_transfer.submit() catch {};
+    const res = try self.single_flight.enter(robots_url, transfer, .robots);
+    switch (res) {
+        .queued => {
+            // joined inflight fetch so release it.
+            client.arena_pool.release(arena);
```
karlseguin: I think there's a UAF bug in SingleFlight (1) related to this, BUT, as-is there's no reason to even acquire the arena in this case.

(1) Commented in SingleFlight.enter

src/network/SingleFlight.zig
```diff
@@ -0,0 +1,346 @@
...
+    while (it.next()) |entry| {
+        entry.value_ptr.deinit(self.allocator);
+    }
+    self.pending.deinit(self.allocator);
+}
+
+pub const EnterResult = enum { initial, queued };
+
+pub fn enter(self: *SingleFlight, key: []const u8, transfer: *Transfer, reason: Transfer.ParkedBy) !EnterResult {
+    const gop = try self.pending.getOrPut(self.allocator, key);
```
karlseguin, line 44: Pretty sure there's no guarantee that `key` has the necessary lifetime.

src/network/CorsGate.zig
```diff
@@ -0,0 +1,462 @@
...
+            "Access-Control-Request-Headers: {s}",
+            .{names.items},
+            0,
+        ));
+    }
+
+    try headers.add(try std.fmt.allocPrintSentinel(
+        arena,
+        "Origin: {s}",
+        .{transfer.req.cookie_origin},
```
karlseguin: I don't think this is right. `req.cookie_origin` is set as the full URL. That seems wrong here, and I assume if it's used as the request's origin, it's wrong in other places.

src/network/HttpClient.zig
```diff
@@ -872,15 +887,23 @@ fn pipeline(self: *Client, transfer: *Transfer, from: SubmitFrom) !void {
...
+                        },
+                        .blocked => {
+                            lp.metrics.robots_access.incr(.deny);
+                            return transfer.failAsync(error.RobotsBlocked);
+                        },
+                        .pending => return,
+                    }
+                }
+
+                switch (try self.cors.check(transfer)) {
```
karlseguin, line 1044: A request blocked on robots will resume the pipeline at network, skipping the CORS check.

src/network/CorsGate.zig
```diff
@@ -0,0 +1,462 @@
...
+    if (names.items.len > 0) {
+        try headers.add(try std.fmt.allocPrintSentinel(
+            arena,
+            "Access-Control-Request-Headers: {s}",
+            .{names.items},
+            0,
+        ));
+    }
+
+    try headers.add(try std.fmt.allocPrintSentinel(
```
karlseguin: This is a duplicate of what `addOriginHeader will add ?

src/network/CorsGate.zig
```diff
@@ -0,0 +1,462 @@
...
+                .req_method = transfer.req.method,
+                .req_headers = req_headers,
+                // TODO: add credentials mode to request.
+                .req_credentials = .omit,
+            };
+
+            log.debug(.cors, "sending cors preflight", .{ .url = transfer.req.url });
+
+            var headers = try client.newHeaders();
+            errdefer headers.deinit();
```
karlseguin: This `errdefer` needed to be scoped to `preflightHeaders`. `newRequest` takes responsible for the supplied headers. This was a necessary change to avoid ambiguity over double-free vs a leak. Possible that a header-rework would eliminate this sensitivity entirely.

src/network/HttpClient.zig
```diff
@@ -840,6 +853,8 @@ fn pipeline(self: *Client, transfer: *Transfer, from: SubmitFrom) !void {
                 try wba.signRequest(transfer.arena, &transfer.req.headers, authority);
             }
 
+            try addOriginHeader(transfer);
```
karlseguin: Neither Firefox nor Chrome seem to add this header on every request.

src/network/CorsGate.zig
```diff
@@ -0,0 +1,462 @@
...
+                    .wildcard => "*",
+                    .value => |v| v,
+                } else "(missing)",
+                .allow_credentials = @tagName(entry.credentials),
+                .allow_methods = allow_methods_str,
+                .allow_headers = allow_headers_str,
+            });
+        }
+
+        log.debug(.cors, "cors preflight allowed", .{ .key = self.key, .status = self.status });
```
karlseguin: It can reach here and not be allowed.

```zig
log.debug(.cors, "cors prefligh", .{ .key = self.key, .status = self.status, .allowed = allowed });
```

src/network/CorsGate.zig
```diff
@@ -0,0 +1,462 @@
...
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+const std = @import("std");
+const lp = @import("lightpanda");
+
+const http = @import("http.zig");
+const Request = @import("../browser/webapi/net/Request.zig");
+const Network = @import("Network.zig");
+const Robots = @import("Robots.zig");
```
karlseguin: unused

src/network/CorsGate.zig
```diff
@@ -0,0 +1,462 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+const std = @import("std");
+const lp = @import("lightpanda");
+
+const http = @import("http.zig");
+const Request = @import("../browser/webapi/net/Request.zig");
```
karlseguin: unused

src/network/CorsGate.zig
```diff
@@ -0,0 +1,462 @@
...
+const Transfer = @import("HttpClient.zig").Transfer;
+const ArenaPool = @import("../ArenaPool.zig");
+const URL = @import("../browser/URL.zig");
+
+const log = lp.log;
+const Allocator = std.mem.Allocator;
+
+pub const CorsGate = @This();
+
+network: *Network,
```
karlseguin: unused

src/network/CorsGate.zig
```diff
@@ -0,0 +1,462 @@
...
+    const mime = blk: {
+        const semi = std.mem.indexOfScalar(u8, ct, ';') orelse ct.len;
+        break :blk std.mem.trim(u8, ct[0..semi], &std.ascii.whitespace);
+    };
+
+    for (safelisted) |s| if (std.ascii.eqlIgnoreCase(mime, s)) return true;
+    return false;
+}
+
+pub fn needsPreflight(transfer: *Transfer) bool {
```
karlseguin: Doesn't need to be public (it isn't the only one)

src/network/CorsGate.zig
```diff
@@ -0,0 +1,462 @@
...
+const log = lp.log;
+const Allocator = std.mem.Allocator;
+
+pub const CorsGate = @This();
+
+network: *Network,
+allocator: Allocator,
+single_flight: SingleFlight,
+
+pub const Result = enum { allowed, blocked, pending };
```
karlseguin: Don't think anything returns/uses .blocked.

src/network/RobotsGate.zig
```diff
@@ -77,94 +73,80 @@ pub fn check(self: *RobotsGate, transfer: *Transfer) !Result {
...
+            return;
+        },
+        .initial => {
+            errdefer self.single_flight.abort(robots_url);
```
karlseguin: This parks it, but on failure it never gets unparked.

karlseguin (commented):
- I ran the demo tests, and there wasn't 1 OPTIONS request.
- The singleflight is nice, but it might have been a mistake to land it with this PR. It introduces bugs not related to CORS. I'd consider landing Singlelight first.
- Without caching CORS responses, this seems like it could hurt performance more than initially anticipated.

src/network/HttpClient.zig
karlseguin: Done in https://github.com/lightpanda-io/browser/pull/3118

I want to work on referrer policy, and I'll have the same issue.

mookums:
Redid a lot of the implementation of the CorsGate. I'm happier with this one. I purposefully omitted the `CorsStore` that I wrote, from my testing (mostly just running the WPT tests or the integration tests against it) i rarely if ever got a proper CORS cache hit with the store.

It is currently disabled by default behind an `--obey-cors` flag.

krichprollsch:
As far as I can see, chrome sends the pre-flight request each time... It doesn't use any cache.

karlseguin:
> As far as I can see, chrome sends the pre-flight request each time... It doesn't use any cache.

It does if the response includes a access-control-max-age header

src/help.zon
```diff
@@ -406,6 +390,9 @@
     \\  --obey-robots
     \\          Fetches and obeys robots.txt of the target page.
     \\          Defaults to false.
+    \\  --obey-cors
```
karlseguin: Should be above `--obey-robots`

karlseguin:
i started to review / test this, but I ran into a couple show-stoppers.

1. It panics on a preflight deny (trivial to fix, but makes it impossible to test)
2. It seems to block things it shouldn't? I expect to see 0 on virtually every website.

```
zig build run -- fetch --obey-cors "https://www.theverge.com/"  2>&1 | grep "CorsBlocked" | wc -l
24
```

Doesn't seem right.

mookums:
2 is fixed. It was blocking some scripts that are meant to be processed as `no-cors` by default, mostly analytics and tracking stuff.

w.r.t 1, I'm not sure how it would panic on a preflight deny? It just sets allowed as false and will resolve following the same path as preflight accepts and just resolve the requests as blocked.

mookums:
as a follow up, I ran it against the `/cors/preflight-failure.htm` WPT test  that does preflight failures and I got no panics

karlseguin:
use the wptrunner in the demo project to run /cors/.  It'll crash.

mookums:
I always run it with `--log-filter-scopes -all,+cors` so never came up for me.

src/network/CorsGate.zig
```diff
@@ -0,0 +1,470 @@
...
+        const arena = self.arena;
+        gate.flushPending(self.url, allowed);
+        arena.release();
+    }
+};
+
+fn fetchThenResume(self: *CorsGate, transfer: *Transfer) !void {
+    const url = transfer.req.url;
+
+    const result = try self.single_flight.enter(url, transfer, .cors);
```
karlseguin: SingleFlight is fine for robots.txt, those are always GET. But here the method matters...and so could the headers (technically true for robots.txt too, actually).

src/network/HttpClient.zig
```diff
@@ -874,9 +891,26 @@ fn pipeline(self: *Client, transfer: *Transfer, from: SubmitFrom) !void {
                 return transfer.failAsync(error.UrlBlocked);
             }
             if (try self.cacheLookup(transfer)) {
-                // response came from the cache, we're done
                 return;
             }
+
+            if (!isCrossOriginModeAllowed(transfer) and !transfer.req.internal) {
```
karlseguin: The CORS check should happen _before_ the cache.

src/network/HttpClient.zig
```diff
@@ -4169,6 +4255,7 @@ test "HttpClient: abort survives an error_callback that tears down the owner" {
                 .url = "http://example.com/",
                 .cookie_jar = null,
                 .cookie_origin = "",
+                .origin = "",
```
karlseguin: A bit more correct to pass null for all these tests. I don't think you'll ever get a "" origin in the code.

src/network/CorsGate.zig
```diff
@@ -0,0 +1,470 @@
...
+        }
+
+        // Access-Control-Allow-Headers
+        if (self.request_headers.len > 0) {
+            const allow_headers = acah orelse {
+                log.debug(.cors, "preflight blocked", .{ .url = self.url, .reason = "missing acah" });
+                return false;
+            };
+
+            const headers_wildcard = std.mem.eql(u8, allow_headers, "*") and !self.wants_credentials;
```
karlseguin, line 348: "The Authorization header doesn't accept wildcard and always needs to be listed explicitly." (1) which means this won't work for `*, Authorization` and it means `*` (standalone or in a list) still requires an explicit `Authorization`

(1) https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Access-Control-Allow-Headers#sect

src/network/HttpClient.zig
```diff
@@ -1596,6 +1636,16 @@ fn processOneMessage(self: *Client, msg: http.Handles.MultiMessage, transfer: *T
 
     try transfer.materializeResponse(msg.conn, .{});
 
+    // Validate the headers for the response with CORS.
+    if (transfer._cors_cross_origin) {
```
karlseguin: There are some references online to how this behaves with a 304, but I couldn't find anything official in the spec. There's at least a few mentions that a 304 doesn't have to (or maybe shouldn't) include CORS headers and they should come from the cached value.  `/cors/304.htm` is currently 0/4 which tells me we're doing it wrong.

src/browser/webapi/net/Fetch.zig
```diff
@@ -170,6 +184,17 @@ fn httpHeaderDoneCallback(transfer: *Transfer) !Transfer.HeaderResult {
     res._url = try arena.dupeZ(u8, transfer.req.url);
     res._is_redirected = transfer.redirectCount().? > 0;
 
+    // no-cors mode: regardless of what the server returned, JS only ever sees
+    // an opaque response — status 0, no headers, no body, url "".
+    if (self._no_cors) {
```
karlseguin: 1 - this should only happen with `--obey-cors`
2 - even with --obey-cors, it's only opaque on a cross origin request

src/network/CorsGate.zig
```diff
@@ -0,0 +1,470 @@
...
+            };
+
+            if (!std.mem.eql(u8, allow_credentials, "true")) {
+                log.debug(.cors, "preflight blocked", .{ .url = self.url, .reason = "credentials not allowed", .allow_credentials = acac });
+                return false;
+            }
+        }
+
+        // Access-Control-Allow-Methods
+        const allow_methods = acam orelse {
```
karlseguin: "GET, HEAD, and POST are always allowed, regardless of whether they are specified in this header," (1)

(1) https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Access-Control-Allow-Methods#method

src/network/HttpClient.zig
```diff
@@ -1698,17 +1748,36 @@ pub const Request = struct {
...
     cookie_jar: ?*CookieJar,
     cookie_origin: [:0]const u8,
+    origin: ?[]const u8,
     resource_type: ResourceType,
     redirect: RedirectMode = .follow,
     referrer_policy: ?referrer.Policy = null,
-    credentials: ?[:0]const u8 = null,
+    basic_auth_credentials: ?[:0]const u8 = null,
+    credentials_mode: CredentialsMode = .same_origin,
+    request_mode: RequestMode = .cors,
```
karlseguin: Having this defaulted means that things which shouldn't be blocked are. Better to leave it out and have every callsite have to think about what the correct value is.

karlseguin (commented):
A redirect to a different origin doesn't trigger a CORS check. It might be worth looking at how redirects are handled holistically, or maybe there's a simple / clean fix for CORS.

But, for now, I'm less worried about places where CORS doesn't block where it should, but rather places where CORS blocks when it shouldn't (since that can break existing use-cases). There's ~4 different ways that can  happen, all commented inline.

karlseguin:
No caching

src/network/HttpClient.zig
mookums: Yeah, the WPT tests also seem to want it to resolve the CORS status from the HTTP cache.

src/help.zon
```diff
@@ -348,6 +348,15 @@
     \\          contribute to computed styles (and therefore to visibility checks like
     \\          display, visibility, opacity, pointer-events).
     \\          Defaults to false, except in agent mode with an LLM, where it is on.
+    \\  --experimental-features <FEATURE>
+    \\          Enable an experimental, unstable feature. Can be passed multiple
+    \\          times, or as a single comma-separated list. Behavior may change
```
krichprollsch: comma separated is deprecated, don't advertise it.

krichprollsch:
Shouldn't `HttpClient.Rquest.getCookieString` check `credentials_mode` before writing cookies?

EDIT: it seems we also should check  `credentials_mode`  when we handle `Set-Cookie` from response (into `materializeResponse`)

src/browser/webapi/net/Fetch.zig
```diff
@@ -218,6 +243,10 @@ fn httpDataCallback(transfer: *Transfer, data: []const u8) !void {
         }
     }
 
+    if (self._no_cors) {
```
karlseguin: Shouldn't this be the same condition as the header?

```
if (self._no_cors and transfer.client.obey_cors and transfer._cors_cross_origin) {
```

src/browser/webapi/WorkerGlobalScope.zig
```diff
@@ -424,6 +424,9 @@ fn importScript(self: *WorkerGlobalScope, arena: Allocator, url: [:0]const u8) !
         .loader_id = self._loader_id,
         .cookie_jar = &session.cookie_jar,
         .cookie_origin = self.url,
+        .origin = self.origin,
+        .request_mode = .same_origin,
```
karlseguin: https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope/importScripts

"Unlike the initial classic module script, which must be same-origin with its document, this method can import cross-origin scripts unless blocked by a [Cross-Origin-Resource-Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cross-Origin-Resource-Policy), [Content Security Policy (CSP)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP), or some other security mechanism. Note that because classic scripts are fetched in no-cors mode, they can be fetched cross-origin even if the server does not set the appropriate CORS headers."

src/browser/webapi/net/Request.zig
```diff
@@ -60,6 +61,7 @@ pub const InitOpts = struct {
     credentials: Credentials = .@"same-origin",
     headers: ?Headers.InitOpts = null,
     method: ?[]const u8 = null,
+    mode: Mode = .cors,
```
krichprollsch: JS shouldn't be able to set `navigate` mode.
only `cors`, `same-origin` and `no-cors` are available.
https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#making_cross-origin_requests

mookums:
Latest batch of fixes and correctness issues gets `/cors` wpt passing up to 52% (423/813)

src/browser/ScriptManager.zig
```diff
@@ -97,6 +97,38 @@ pub fn tailHook(base: *ScriptManagerBase) void {
...
+        if (std.ascii.eqlIgnoreCase(value, "use-credentials")) {
+            break :blk .use_credentials;
+        }
+
+        // Empty-value and invalid-value defaults are both Anonymous.
+        break :blk .anonymous;
+    };
+
+    return switch (mode) {
+        .no_cors => .{ .request_mode = .no_cors, .credentials_mode = .same_origin },
```
krichprollsch: it seems it should be `.credentials_mode = .include`

src/browser/frame/resource_load.zig
```diff
@@ -79,6 +79,9 @@ pub fn image(frame: *Frame, img: *Element.Html.Image, src: []const u8) !void {
         .ctx = load,
         .url = resolved,
         .method = .GET,
+        .origin = frame.origin,
+        .request_mode = .no_cors,
+        .credentials_mode = .same_origin,
```
krichprollsch: it should be `.credentials_mode = .include`

krichprollsch (commented):
Now we have `Request.credentials_mode` I think (and Claude thinks too) we should drop the `Request.cookie` bool and use only  `Request.credentials_mode` to define how we should treat cookies.
In the current version, we don't set `cookie` for ScriptManager request's and it is set to `true` by default...

We should always  use the owner's cookie jar, but use it depending the credentials_mode in to `getCookieString`, `fulfillRedirect`, `materializeResponse` and `configureConn` (to be apply correclty on each transfer's hop)

mookums:
Rebased again against main. Up to 457/813 `/cors/` tests making it 56.2%.

mookums:
More changes to get more WPT tests passing, up to 491/813 -> 60.4%

src/browser/webapi/net/Fetch.zig
krichprollsch: We could avoid this reservation for opaque response.

krichprollsch (approved):

PR #3004, opened 2026-07-20
feat(cdp): add Network.setBlockedURLs and --block-urls request blocking

mvanhorn:
## What

Add request-time URL blocking from two entry points that share one blocklist:

- `--block-urls <LIST>` — comma-separated `*`-wildcard patterns, matched case-insensitively against the full URL, on `serve` / `fetch` / `mcp`, next to the existing `--block-cidrs`.
- `Network.setBlockedURLs` — the CDP method, matching Chrome's semantics.

## Why

Lightpanda's `Network` CDP domain implemented no request-blocking method (`src/cdp/domains/network.zig` dispatched `enable`, `disable`, `setCacheDisabled`, header/cookie methods, and `getResponseBody`, nothing else), and the `fetch` CLI had no URL blocklist. So a Puppeteer/Playwright script that blocks ad/tracker requests against Chrome via `page.setRequestInterception` / `page.route(..., r => r.abort())` — backed by CDP `Network.setBlockedURLs` — silently lost that behavior when pointed at Lightpanda. For a high-volume scraper those dropped requests are the point: they cut connections, latency, and bandwidth an agent never reads (Browserless markets the same request/ad blocking as a scraping speedup).

## Change

- New `src/network/UrlBlocklist.zig`, modeled on `IpFilter.zig`: compiles and owns a set of wildcard patterns once, and answers `isBlocked(url)` with an allocation-free two-pointer glob match.
- `HttpClient` holds an optional `UrlBlocklist` built from `config.blockedUrlPatterns()` at init. Enforcement sits at the existing request-start gate that already handles robots and IP/CIDR blocking, and also on the redirect target, so a blocked URL never opens a socket and fails via the same async path. Internal transfers (e.g. `robots.txt`) are exempt, matching the robots / IP-filter exemption. `setBlockedUrls()` replaces the live list at runtime for the CDP path (owning its own copies).
- A blocked request fails with `error.UrlBlocked`, which maps through `Notification.RequestFail.blocked_reason` to `Network.loadingFailed` with `blockedReason: "inspector"` — the reason Chrome reports for `setBlockedURLs` hits — so CDP clients see the expected event.

## Demo

Simulated demo (illustrative, not a live capture): `--block-urls "*doubleclick*,*analytics*,*://*/*.png"` drops the matching ad/tracker/image requests while `index.html` and the non-matching `pixel.gif` still load.

Simulated demo of --block-urls request blocking

## Test

- `src/network/UrlBlocklist.zig`: wildcard-matcher unit tests — case-insensitivity, backtracking, consecutive/trailing `*`, empty and `*` patterns, and pattern ownership (mutating the source after `init` leaves matches intact). These run standalone under the leak-detecting test allocator and pass.
- `src/network/HttpClient.zig`: `setBlockedUrls` owns/replaces/clears, and the internal-transfer exemption.
- `src/cdp/domains/network.zig`: a blocked request emits `Network.loadingFailed` with `blockedReason: "inspector"`, for both the initial request and a redirect target.
- `zig fmt --check` clean; `zig ast-check` clean on all touched files. The full `make test` needs the V8 toolchain build.

### DIFF
--- a/src/cdp/domains/network.zig
+++ b/src/cdp/domains/network.zig
@@ -100,11 +100,11 @@ fn setCacheDisabled(cmd: *CDP.Command) !void {
 
 fn setBlockedURLs(cmd: *CDP.Command) !void {
     const params = (try cmd.params(struct {
-        urls: []const []const u8,
+        urlPatterns: []const HttpClient.BlockPattern,
     })) orelse return error.InvalidParams;
 
     const bc = cmd.browser_context orelse return error.BrowserContextNotLoaded;
-    try bc.cdp.browser.http_client.setBlockedUrls(params.urls);
+    try bc.cdp.browser.http_client.setBlockedUrlPatterns(params.urlPatterns);
     return cmd.sendResult(null, .{});
 }
 
@@ -984,7 +984,9 @@ test "cdp.Network: setBlockedURLs blocks requests with inspector reason" {
     try ctx.processMessage(.{
         .id = 2,
         .method = "Network.setBlockedURLs",
-        .params = .{ .urls = &[_][]const u8{"*://blocked.test/*"} },
+        .params = .{ .urlPatterns = &[_]HttpClient.BlockPattern{
+            .{ .urlPattern = "*://blocked.test/*", .block = true },
+        } },
     });
     try ctx.expectSentResult(null, .{ .id = 2 });
 
--- a/src/cdp/domains/network.zig
+++ b/src/cdp/domains/network.zig
@@ -28,8 +28,9 @@ const Mime = @import("../../browser/Mime.zig");
 const Notification = @import("../../Notification.zig");
 const timestamp = @import("../../datetime.zig").timestamp;
 
-const Headers = @import("../../network/HttpClient.zig").Headers;
-const Transfer = @import("../../network/HttpClient.zig").Transfer;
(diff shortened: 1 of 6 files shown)


src/cdp/domains/network.zig
```diff
@@ -95,6 +98,16 @@ fn setCacheDisabled(cmd: *CDP.Command) !void {
     return cmd.sendResult(null, .{});
 }
 
+fn setBlockedURLs(cmd: *CDP.Command) !void {
+    const params = (try cmd.params(struct {
+        urls: []const []const u8,
```
krichprollsch: It seems `urls` is deprecated in favor or `urlPatterns` which is similar but you can configure block/allow per pattern. I think it's better to first implement the new way here.
https://chromedevtools.github.io/devtools-protocol/tot/Network/#method-setBlockedURLs

mvanhorn:
Reworked onto urlPatterns in b573eb1: each pattern carries its block/allow flag (first match wins), the blocklist owns the compiled patterns plus flags, and the legacy urls path stays for back-compat. The three standalone UrlBlocklist tests pass; the full Zig 0.15.2 build runs in CI since my local toolchain is 0.16.

krichprollsch:
merged via #3040 
Thanks!

PR #3005, opened 2026-07-21
perf, v8: reduce gc hint frequency

karlseguin:
Reduce v8 GC hint frequency from every 1s to every 5s.

I also experimented with changing the .critical -> .moderate on Session.deinit, but that caused our e2e memory regression to trip (+ ~1.2MB). So staying with this safe change for now.

V8 does have callbacks for when a GC starts/stops, and we could probe memory usage to do this more surgically. Users can also tweak thisvia the dangerous/unsupported `--v8-flags-unsafe` flag.


krichprollsch:
I don't notice an important change w/ the magic8ball test unfortunately :/
even with a `moderate` call...

krichprollsch (approved):

PR #3007, opened 2026-07-21
zig: Zig 0.16

karlseguin:
Built against https://github.com/lightpanda-io/zig-v8-fork/tree/zig-0.16 but it doesn't require a new v8 build.

Built against https://github.com/lightpanda-io/boringssl-zig/tree/zig-0.16 since the current fork we point to isn't updated.

A global std.Io instance, lp.io. Way easier this way and requires 0 changes to our libcurl integration / event loop.

Network code uses a new layer that does what Zig 0.15's posix package used to do. Again, quicker migration that way. But, as long as we have the global IO, and given the half-baked nature of networking in std.Io 0.16, this just makes sense. Things can be migrated as needed.

The std.time.* -> std.Io.Timestamp/Clock/Duration resulted in _a lot_ of changes. ArrayList = .{} -> ArrayList -> .empty also resulted in a lot of changes, but that's obviously superficial. As is the trimLeft/trimRight -> trimStart/trimEnd rename.

Locking adopt the `Uncancelable` variants, e.g. mutex.lockUncancelable() to preserve the error-free signature (and, because cancellation would be something we'd have to put more thought into).

std.json.ObjectMap is now unmanaged, so the allocator had to be passed along. However, there's still a deprecated managed variant of MemoryPool, so I switched to it (we can do a small follow up PR to move to the unmanaged after).

I tried use_llvm = false, but it locks my computer, consuming RAM until MacOS gives me a popup I've never seen before, begging me to start killing processes.

Agent and the networking stuff saw the most significant changes.

### DIFF
--- a/.github/actions/install/action.yml
+++ b/.github/actions/install/action.yml
@@ -41,6 +41,14 @@ runs:
     # Zig version used from the `minimum_zig_version` field in build.zig.zon
     - uses: mlugg/setup-zig@d1434d08867e3ee9daa34448df10607b98908d29 # v2.2.1
 
+    # Zig 0.16 writes fetched .zip archives to <global cache>/tmp but no longer
+    # creates that directory, so zip dependencies (sqlite amalgamation) fail
+    # with "failed to create temporary zip file: FileNotFound" on the fresh
+    # cache dir setup-zig points ZIG_GLOBAL_CACHE_DIR at.
+    - name: Zig 0.16 zip-fetch workaround
+      shell: bash
+      run: mkdir -p "${ZIG_GLOBAL_CACHE_DIR:-$HOME/.cache/zig}/tmp"
+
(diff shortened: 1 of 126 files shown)


karlseguin:
We did talk about waiting for 0.17. But at the time, 0.17 looked like a minor release to bump the llvm version. Now it's looking bigger, with breaking changes (especially to the build system).

nikneym:
> A global std.Io instance, lp.io. Way easier this way and requires 0 changes to our libcurl integration / event loop.

Haven't looked at changes yet, though I'm curious if a custom `std.Io` implementation for needs of LP is planned.

arrufat:
You can use this, too:
https://github.com/lightpanda-io/zenai/tree/zig-0.16

krichprollsch (approved):

PR #3010, opened 2026-07-21
Interrupt nested script waits during teardown

### LINKED ISSUE #3009: CDP page/context teardown still hangs after #2647 with mixed deferred and async scripts
## Summary

`page.close()` and `BrowserContext.close()` can still stall for more than five seconds under concurrent CDP load, despite the fix for #2646 being present.

This is a self-contained localhost repro derived from a production storefront's *sanitized resource shape*. It does not request any external host or include storefront HTML/content.

## Environment

- Lightpanda: `1.0.0-dev.8180+299eb2c5`
- Source commit: `299eb2c5` (production runs the equivalent `1.0.0-nightly.8180+299eb2c5` Linux build)
- Playwright Core: `1.59.1`
- Six persistent CDP clients / `--cdp-max-connections 6`
- Flags: `--disable-subframes --disable-workers --http_proxy <local forwarding proxy>`

## Reproducer

Script: https://gist.github.com/staylor/900a3f825f110e9904e3cfa2191a6637

It starts all local services itself:

- an HTTP fixture;
- an unauthenticated localhost forwarding proxy;
- Lightpanda with the production-relevant flags; and
- six persistent Playwright CDP clients.

The fixture has no external dependencies. Each document contains:

- five never-completing `async` scripts;
- five never-completing `defer` scripts;
- five completing `async` scripts;
- eighteen completing `defer` scripts;
- nine completing parser-blocking scripts; and
- twelve completing stylesheets.

Every client repeatedly creates a context/page, installs a `context.route(... continue())` handler and init script, navigates with a deliberately short 250ms `domcontentloaded` budget, then closes the page and context in parallel. Each close is independently measured against a 5.1s wall-clock budget.

```sh
pnpm add playwright-core
LIGHTPANDA_BIN=/path/to/lightpanda node lightpanda-close-timeout-repro-standalone.mjs
```

## Observed

On the stated build, one 6-client x 50-iteration run yields:

```text
{
  contexts: 300,
  navigationTimeouts: 300,
  pageCloseTimeouts: 22,
  contextCloseTimeouts: 22
}
```

The navigation timeout is intentional: the deferred scripts never finish. The defect is that teardown then intermittently exceeds the full 5.1s close budget for both page and context.

Controls did not produce teardown timeouts:

- five hanging async scripts plus five hanging deferred scripts, without the completing-script mix: 0 / 300;
- holding only the async subset of the captured shape: 0 / 300;
- holding only the deferred subset: 0 / 300.

The failure needs the mixed script scheduling/load pattern, rather than merely a hanging request.

## Expected

A page or context teardown should cancel outstanding script fetches and return promptly. It should not retain the CDP command for more than five seconds after the client has abandoned navigation.

## Production connection

Production sees the same paired `page.close()` / `context.close()` five-second timeout warnings. Its worker contains the existing slot-retirement and process-recycle containment, but that bounds the impact rather than resolving it.

The build is 1,553 commits after the #2647 merge, so this is not a report that the original #2646 patch is absent. It appears to be a remaining teardown edge case with mixed deferred/async script loading.

staylor:
## Summary

- expose the existing queued-teardown check through `HttpClient.hasPendingTeardown()`
- reject new synchronous requests when teardown is already queued
- interrupt `waitForPreload()` and `waitForImport()` instead of continuing their nested synchronous polling loops
- add leak-checked regression tests for both wait paths

## Root cause

`syncRequest()` already interrupted an active blocking fetch when a close or disconnect message reached the CDP inbox. The script manager could then enter its own `waitForPreload()` or `waitForImport()` loop, where repeated `tickSync(200)` calls did not check that inbox. Because sync-wait mode cannot dispatch teardown while parser frames are on the stack, the close command remained queued until the outstanding HTTP operations timed out.

The new early checks unwind those nested waits to the next safe inbox drain.

Fixes #3009.

## Validation

- `make test`: 1087/1087 passed
- standalone reproducer from #3009, six CDP clients × 50 contexts:
  - current `main`: 13 page-close and 13 context-close timeouts in 300 contexts
  - patched run 1: 0 / 300 and 0 / 300
  - patched run 2: 0 / 300 and 0 / 300
  - patched run 3: 0 / 300 and 0 / 300

### DIFF
--- a/src/browser/ScriptManager.zig
+++ b/src/browser/ScriptManager.zig
@@ -171,7 +171,6 @@ fn waitForPreload(self: *ScriptManager, url: [:0]const u8) ?*Script {
         const entry = self.preloaded_scripts.getPtr(url) orelse return null;
         switch (entry.state) {
             .loading => {
-                if (client.hasPendingTeardown()) return null;
                 _ = client.tickSync(200) catch return null;
                 continue;
             },
--- a/src/browser/ScriptManager.zig
+++ b/src/browser/ScriptManager.zig
@@ -171,6 +171,7 @@ fn waitForPreload(self: *ScriptManager, url: [:0]const u8) ?*Script {
         const entry = self.preloaded_scripts.getPtr(url) orelse return null;
         switch (entry.state) {
             .loading => {
+                if (client.hasPendingTeardown()) return null;
                 _ = client.tickSync(200) catch return null;
                 continue;
             },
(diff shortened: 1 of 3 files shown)


karlseguin:
Thanks.

PR #3019, opened 2026-07-22
tools: add Brave search + /searchEngine metacommand

arrufat:
Add Brave as a second API-backed web search engine: the search tool's auto mode tries Brave then Tavily (keyed on BRAVE_API_KEY/TAVILY_API_KEY) before the DuckDuckGo scrape. A new /searchEngine REPL metacommand pins the engine explicitly (auto|tavily|brave|duckduckgo) — an explicit engine surfaces its failures instead of silently degrading — and is persisted to .lp-agent.zon like the other REPL settings.

Bumps zenai to pick up the zenai.search.brave client.

### DIFF
--- a/src/browser/tools.zig
+++ b/src/browser/tools.zig
@@ -22,6 +22,7 @@ const zenai = @import("zenai");
 
 const log = lp.log;
 const tavily = zenai.search.tavily;
+const brave = zenai.search.brave;
 
 const DOMNode = @import("webapi/Node.zig");
 const CDPNode = @import("../cdp/Node.zig");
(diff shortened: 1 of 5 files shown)


src/browser/tools.zig
```diff
@@ -1016,6 +1056,45 @@ fn formatTavilyMarkdown(arena: std.mem.Allocator, resp: tavily.types.SearchRespo
...
+}
+
+fn formatBraveMarkdown(arena: std.mem.Allocator, resp: brave.types.SearchResponse) ![]const u8 {
+    var aw: std.Io.Writer.Allocating = .init(arena);
+    const w = &aw.writer;
+    const results: []const brave.types.Result = if (resp.web) |web| web.results else &.{};
+    for (results, 0..) |r, i| {
+        try w.print("{d}. **{s}** — {s}\n   {s}\n\n", .{ i + 1, r.title, r.url, r.description });
+    }
+    if (results.len == 0) {
```
karlseguin: It works, and this comment is minor, but this implementation is a bit weird. Why no check `results.len == 0` upfront?

```zig
fn formatBraveMarkdown(...) ![]const u8 {
  const results: []const brave.types.Result = if (resp.web) |web| web.results else &.{};
  if (results.len == 0) {
    return "No results.";
  }
  // ....
}
```

(and, since I'm here, this function could take a *Writer instead of an Allocator)

src/browser/tools.zig
```diff
@@ -1016,6 +1056,45 @@ fn formatTavilyMarkdown(arena: std.mem.Allocator, resp: tavily.types.SearchRespo
...
+
+    return formatBraveMarkdown(arena, response.value);
+}
+
+fn formatBraveMarkdown(arena: std.mem.Allocator, resp: brave.types.SearchResponse) ![]const u8 {
+    var aw: std.Io.Writer.Allocating = .init(arena);
+    const w = &aw.writer;
+    const results: []const brave.types.Result = if (resp.web) |web| web.results else &.{};
+    for (results, 0..) |r, i| {
+        try w.print("{d}. **{s}** — {s}\n   {s}\n\n", .{ i + 1, r.title, r.url, r.description });
```
karlseguin: Can description have values that message with this format?

src/browser/tools.zig
```diff
@@ -1016,6 +1056,45 @@ fn formatTavilyMarkdown(arena: std.mem.Allocator, resp: tavily.types.SearchRespo
...
+) ![]const u8 {
+    var client: brave.Client = .init(lp.io, arena, api_key, .{});
+    defer client.deinit();
+
+    // text_decorations=false: no <strong> markup in model-read snippets.
+    var response = client.search(query, .{ .count = 10, .text_decorations = false }) catch |err| {
+        if (client.last_error_status) |status| {
+            log.warn(.browser, "brave non-2xx", .{
+                .status = status,
+                .body = client.last_error_body orelse "",
```
karlseguin:
```suggestion
                .body = client.last_error_body,
```

It'll handle the optional..unless you have a reason to want null -> ""

karlseguin (approved):
just a few small nits inline

src/browser/tools.zig
arrufat: Yes, but the consumer is an LLM. Fixed anyway, thanks!

PR #3020, opened 2026-07-22
cdp: preserve node IDs across child frame navigation

### LINKED ISSUE #2994: CDP: backendNodeIds of nodes replaced by framework re-render become UnknownNode — puppeteer handle adoption races hydration
On pages where a framework re-renders DOM shortly after load (observed on eu.gymshark.com — React hydration replaces the product-card subtree), puppeteer's element-handle adoption fails with:

```
ProtocolError: Protocol error (DOM.resolveNode): UnknownNode
    at IsolatedWorld.adoptBackendNode (puppeteer-core/lib/esm/puppeteer/cdp/IsolatedWorld.js:107)
    at CdpElementHandle.$$ / $$eval
```

Sequence: `page.$$eval("[class*='product-card_card-wrapper']", ...)` queries the cards, puppeteer adopts each result via `DOM.resolveNode(backendNodeId)` — and if the hydration re-render replaces those nodes between the query and the adoption, lightpanda answers `UnknownNode`. **Headless Chrome never fails this** (0 failures across all benchmark rotations on the same page): it keeps detached nodes resolvable by backendNodeId (they're detached, not forgotten) so the adoption succeeds and the driver reads the stale-but-alive nodes. Playwright on lightpanda is also clean — it doesn't do this adopt dance.

This is distinct from the mid-query *document replacement* case (#2887, refuted — Chrome hard-fails identically there). Here the document survives; only a subtree is re-rendered, and Chrome degrades gracefully where we throw.

## Repro

```bash
cd agent-benchmarks/pandascript-vs-cdp   # npm ci; needs puppeteer-core
lightpanda serve &
for i in $(seq 10); do node scripts/puppeteer/retail.js >/dev/null || echo "FAILED at $i"; sleep 3; done
```

Fails within a few iterations. Highly timing-sensitive: on a slower CPU profile (EPP `balance_performance`) it hit ~10% of runs; with EPP/platform profile `performance` it fails 30–70% of runs (faster loads land the query right in the re-render window) — it aborted a benchmark session today, which is how it got promoted from footnote to issue.

Suggested direction: keep replaced/detached nodes resolvable by backendNodeId for some grace period (or until their id-map entry is evicted), rather than dropping registry entries at replacement time — matching Chrome's observable behavior for `DOM.resolveNode`.

Benchmark context and raw failure records: agent-benchmarks `results/v2-retail-warm`, `results/v3-retail-*` (`pandascript-vs-cdp/RETAIL-INVESTIGATION.md` companion).

raj921:
Fixes #2994

`frameNavigate` reset the entire browser context node registry for every frame navigation. An iframe navigation could therefore invalidate a main-document `backendNodeId` between Puppeteer's node lookup and `DOM.resolveNode`, which then returned `UnknownNode`.

Reset the registry only when the main frame navigates. The regression test checks that child navigation keeps the node ID and main-frame navigation still clears it.

Checks:

- `make test F="cdp.frame: child navigation preserves node registry"`
- `make test`

### DIFF
--- a/src/cdp/domains/page.zig
+++ b/src/cdp/domains/page.zig
@@ -455,12 +455,14 @@ pub fn frameNavigate(bc: *CDP.BrowserContext, event: *const Notification.FrameNa
     // things, but no session.
     const session_id = bc.session_id orelse return;
 
-    // is_pending_root means this navigation is in flight against a pending
-    // Page while the OLD page is still alive and addressable. Don't blow
-    // away the node_registry — the OLD page's nodes are still referenced
-    // by client-held objectIds. The reset moves to frameRemove (commit).
+    // Child-frame navigations must not invalidate node IDs owned by other
+    // frames. A pending root keeps its old page addressable until frameRemove;
+    // other root navigations can clear the registry immediately.
     if (!event.is_pending_root) {
-        bc.reset();
+        const root_frame = bc.mainFrame() orelse return error.FrameNotLoaded;
+        if (event.frame_id == root_frame._frame_id) {
+            bc.reset();
+        }
     }
 
     const frame_id = &id.toFrameId(event.frame_id);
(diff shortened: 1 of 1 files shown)


raj921:

krichprollsch (approved):

PR #3026, opened 2026-07-22
browser: inline finalizer callbacks

### LINKED ISSUE #3025: Finalizable objects avoid one dedicated arena per cleanup callback
## Summary

Every finalizable Zig object exposed to V8 currently allocates a dedicated `ArenaPool` arena solely to hold one `FinalizerCallback`. JavaScript-heavy pages can create hundreds of thousands or millions of these objects, amplifying native allocator churn and page-lifetime RSS even though the callback has a fixed size and its address does not escape through the V8 weak-callback parameter.

In a current-nightly Linux reproduction, constructing 100,000 `Request` objects produced 194,834 arena misses: approximately one arena for the object and one for its finalizer callback. Real Shopify pages under four concurrent CDP connections produced 1.5–6.8 million arena misses and 0.5–1.1 GiB RSS before context teardown returned the process near baseline.

## Current behavior

`Page.finalizer_callbacks` is an `AutoHashMapUnmanaged(usize, *js.FinalizerCallback)`. `Local.createFinalizerCallback` acquires a tiny arena, creates one callback in it, and `FinalizerCallback.deinit` or the final V8 weak callback releases that arena.

The stable V8 weak-callback parameter is `FinalizerCallback.Identity`, allocated from `Browser.fc_identity_pool`. It stores the page and finalizer IDs and looks the callback up in `Page.finalizer_callbacks`; it does not retain a pointer to the `FinalizerCallback`. This means the callback itself does not appear to require a dedicated stable allocation.

## Proposed outcome

Store `FinalizerCallback` values directly in `Page.finalizer_callbacks` and remove the callback-owned arena. Weak callbacks continue to use their existing stable `FinalizerCallback.Identity`, while the page-lifetime map owns callback metadata.

This is an incremental mitigation, not a complete solution for native memory visibility. A separate follow-up should report ArenaPool-backed memory to V8 as external allocated memory and enforce an optional connection-scoped memory budget.

## Acceptance criteria

- Creating one finalizable Zig object no longer acquires an ArenaPool arena solely for its `FinalizerCallback` metadata.
- V8 weak finalization still releases the underlying Zig reference exactly once after the last identity is collected.
- Page teardown still releases every remaining underlying Zig reference exactly once and leaves late weak callbacks safe.
- Existing leak-detection and browser tests pass.
- A regression test covers callback-map growth or movement so no callback pointer is assumed stable across map insertions.

staylor:
## Summary

Store `FinalizerCallback` metadata directly in `Page.finalizer_callbacks` instead of acquiring a dedicated tiny arena for every finalizable Zig object.

The stable V8 weak-callback parameter remains `FinalizerCallback.Identity` from `Browser.fc_identity_pool`. Callback code looks up the map value with `getPtr`, so it does not retain a pointer across map growth.

Add a regression case that keeps 1,024 `Request` wrappers live simultaneously, forcing the callback map to grow before page teardown.

Closes #3025.

## Testing

- `zig build -Dprebuilt_v8_path=.lp-cache/prebuilt-v8/libc_v8_14.9.207.35_macos_aarch64.a test` (1058/1058 passed)
- `git diff --check`

### DIFF
--- a/src/browser/Page.zig
+++ b/src/browser/Page.zig
@@ -86,7 +86,7 @@ identity: js.Identity = .{},
 // Zig instance ptr. The backing FinalizerCallback.Identity structs come from
 // Browser.fc_identity_pool so they outlive the Page (and the Session) for v8
 // weak-callback safety.
-finalizer_callbacks: std.AutoHashMapUnmanaged(usize, *js.FinalizerCallback) = .empty,
+finalizer_callbacks: std.AutoHashMapUnmanaged(usize, js.FinalizerCallback) = .empty,
 
 // Persisted v8 handles owned by this Page. Handles that outlive the Page are
 // reset on teardown; handles that can be released early are dropped
@@ -205,7 +205,7 @@ pub fn deinit(self: *Page) void {
     {
         var it = self.finalizer_callbacks.valueIterator();
         while (it.next()) |fc| {
-            fc.*.deinit(self);
+            fc.deinit(self);
         }
         self.finalizer_callbacks = .empty;
     }
--- a/src/browser/js/Local.zig
+++ b/src/browser/js/Local.zig
@@ -339,9 +339,9 @@ pub fn mapZigInstanceToJs(self: *const Local, js_obj_handle: ?*const v8.Object, 
                     // so that we can cleanup on Page teardown if v8 doesn't finalize.
                     errdefer _ = page.finalizer_callbacks.remove(finalizer_ptr_id);
                     finalizer.acquire_ref(finalizer_ptr_id);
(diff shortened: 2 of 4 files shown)


karlseguin:
This is nice, thanks. I want to believe I had a good reason to do it via an allocator in the first place, but I'm afraid to look into the history and find out I was just wrong from the get-go.

We recently added prometheus metrics, and the first thing that stood out was our ArenaPool's `.tiny` miss rate. First thing I was planning on looking at this morning, but I bet this solves the mystery.

krichprollsch:
The version's upgrade improved the tiny miss arena :+1: 

image

Issue #190, opened 2024-02-29
web API URL


krichprollsch:
https://developer.mozilla.org/en-US/docs/Web/API/URL

PR #3028, opened 2026-07-22
browser: report native arena memory to V8

### LINKED ISSUE #3027: Report browser-owned arena memory to V8
## Summary

Lightpanda allocates most browser wrapper and page-lifetime state in Zig `ArenaPool` arenas, outside V8's heap. V8 cannot see this memory, so pages can accumulate large native working sets before V8 decides to collect weak wrappers. Periodic unconditional memory-pressure notifications mitigate the symptom but do not make collection respond to the amount of native memory retained by a specific CDP connection.

Add per-`Browser` native-memory accounting and report signed changes to that browser's V8 isolate with `v8::Isolate::AdjustAmountOfExternalAllocatedMemory`.

## Design

- Extend `zig-v8-fork` with a C binding for `AdjustAmountOfExternalAllocatedMemory`.
- Give each `Browser` a `NativeMemoryAccount` containing atomic active-byte and pending-delta counters.
- Attribute backing allocations for arenas checked out through a `Session` to its owning browser. Pooled idle capacity remains process-scoped and is not charged to a browser.
- Allocation and release paths only update atomics. They must never call V8 because arena activity can occur outside the isolate thread.
- Flush pending signed deltas at browser runner/isolate boundaries and before isolate teardown.
- Preserve exact accounting when an arena is reused, reset, resized, or returned to the global pool.

## Acceptance criteria

- Browser-owned arena backing allocations increase that browser's active native-byte count; frees and pool returns decrease it without underflow.
- Reusing retained arena capacity transfers its existing bytes to the new browser owner and returning it to the pool removes those bytes from that owner.
- `AdjustAmountOfExternalAllocatedMemory` is only invoked from the browser/isolate execution thread, never directly by allocator or network callbacks.
- Signed deltas are flushed to zero after runner boundaries and before isolate teardown.
- Two browser accounts using the shared global pool remain isolated from each other.
- Unit tests cover allocation, reset/release, retained-capacity reuse, and cross-browser ownership transfer.
- The complete browser test suite and leak checks pass.

This is the native-memory visibility follow-up to #3025. Connection-scoped hard limits and bounded CDP response-body capture can be layered on this accounting separately.

staylor:
## Summary

Track backing allocations for browser-owned `ArenaPool` arenas and report signed changes to the owning V8 isolate with `AdjustAmountOfExternalAllocatedMemory`.

Each `Browser` owns an atomic `NativeMemoryAccount`. Arena allocation, resize, remap, free, reuse, reset, and pool-return paths update only atomics, so network or allocator threads never call V8. Pending deltas are flushed from runner ticks, V8-to-Zig callback returns, and session teardown on the isolate-owning thread.

Retained capacity is transferred to an account when a pooled arena is checked out and removed when it returns to the process-wide pool. Session, page, context, origin, worker, and isolated-world arenas are attributed to their browser. The process-wide `browser_native_memory_bytes` gauge exposes the currently attributed total.

Tests cover signed delta accumulation, backing allocation and reset accounting, retained-capacity transfer between two browser accounts, V8 reporting, and zero active bytes at browser teardown.

Closes #3027.

## Dependencies

This PR is now rebased onto current `main`, so the merged #3026 commit is no longer part of its diff.

`build.zig.zon` now points to merged upstream commit `acde80cb05841e60a8396273adebb49b708c83ac` in `lightpanda-io/zig-v8-fork`, which contains the Zig 0.16 migration from zig-v8-fork#189 and the external-memory binding from zig-v8-fork#190. The temporary dependency on my fork has been removed.

The remaining readiness blocker is a published zig-v8 prebuilt archive containing the new binding. Browser CI and release builds consume those archives, and the latest published archive predates #190. Once that archive is released, this PR can update to its release tag and be marked ready.

## Testing

- Full Zig 0.16 debug suite: 1073/1073 passed
- Zig 0.16 `ReleaseFast` targeted V8-reporting test: 1/1 passed
- `git diff --check`

Until the new upstream archive is published, local validation used the existing macOS V8 archive with a temporary uncommitted C++ shim for `v8__Isolate__AdjustAmountOfExternalAllocatedMemory`. No generated archive, shim, or local SDK-path adjustment is committed.

### DIFF
--- a/src/ArenaPool.zig
+++ b/src/ArenaPool.zig
@@ -23,6 +23,7 @@ const builtin = @import("builtin");
 const log = lp.log;
 const Allocator = std.mem.Allocator;
 const ArenaAllocator = std.heap.ArenaAllocator;
+const NativeMemoryAccount = @import("browser/NativeMemoryAccount.zig");
 
 const ArenaPool = @This();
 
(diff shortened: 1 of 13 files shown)


karlseguin:
It'll be a couple more days before I can look at this. Sorry.

karlseguin:
@staylor this captures memory that isn't pinned by V8, was that intentional?  

I'm thinking of a more targeted approach, a distinct getArena function use for arenas which are finalized (e.g. XHR, ...).  I agree that a prometheus metric for all allocation is still useful.

I'd like do work on this, but I wanted to first find out if if the current "report everything" was important.

staylor:
> @staylor this captures memory that isn't pinned by V8, was that intentional?
> 
> I'm thinking of a more targeted approach, a distinct getArena function use for arenas which are finalized (e.g. XHR, ...). I agree that a prometheus metric for all allocation is still useful.
> 
> I'd like do work on this, but I wanted to first find out if if the current "report everything" was important.

Yes, "report everything" was intentional in that I wanted V8 to react to the total native working set attributed to a browser connection. It wasn’t important that every byte be reported to V8 specifically.

Your distinction is better. I rechecked the V8 contract: `AdjustAmountOfExternalAllocatedMemory` is for external memory "kept alive by JavaScript objects." Long-lived browser/page/session/context arenas aren’t reclaimable by a V8 collection, so including them can trigger collections that cannot release that memory.

I’d separate the two concerns:

- keep total browser-owned arena bytes in the Prometheus metric;
- report to V8 only arenas whose native lifetime is released by a V8 weak finalizer, through a distinct `getArena` path as you suggest.

That still covers the motivating case: finalizable wrappers such as `Request`, `XHR`, and `Blob` were accumulating native arenas until V8 collected their weak wrappers. So the broad V8 reporting isn’t important—please go ahead with the targeted approach.

karlseguin:
https://github.com/lightpanda-io/browser/pull/3086 is now merged. Thanks for the idea. 3086 is both more targeted and invasive. It limits the memory that v8 gets notified about to memory that a GC would actually free (which is both a function of what that memory is used for and whether a GC could actually free it (i.e. isn't being used by Zig itself)).  It's more invasive because virtually every arena: Allocator was turned into a custom type: arena: *lp.Arena.

PR #3030, opened 2026-07-23
webapi: add live SVG collections

mh0pe:
## Summary

- add live point, transform, and string list APIs with real element consumers
- make mutations transactional and animated views read-only
- share validated transform parsing with `DOMMatrix`
- update allocation writers and scratch lifetimes for Zig 0.16

## Maintainer feedback resolved

- preserve Karl's optimization and worker-safety commit (`e4b6b329`)
- attach the live `SVGTransform.matrix` to its owner so every matrix mutator writes through `applyState`
- detach externally removed/replaced items and keep `animVal` matrices read-only
- make point, transform, and string-list updates failure-atomic

## Stack dependency

This is unit 3 of 7. Its logical predecessor is #3034. The corrected branch tip is `39199380`. GitHub targets `main`, so the displayed diff remains cumulative until earlier units merge.

## Validation

- this branch independently passes `make test F="WebApi: Svg"`
- corrected top of stack passes 1,085 of 1,085 repository tests
- production `make build` and the WPT-enabled ReleaseFast build pass
- current WPT: `SVGMatrix-tentative.html` 2/2 and single-parameter translate 1/1
- current WPT `[SameObject]` check has no failure for `SVGTransform.matrix`

## Hosted checks

This head was rewritten after the earlier fork CI runs, so those stale run links are intentionally omitted. Upstream workflows may remain `action_required` until a maintainer approves running fork code; that is an approval gate, not a local test failure.


karlseguin:
@mh0pe I pushed a commit to your branch that (a) does some optimization and (b) removes Frame access from a WebAPI that's also accessible from a Worker (DOMPoint) which is not safe.

Maybe you'd like to review and maybe there are some things you can apply to subsequent PRs. Worth noting that the field -> lookup change is the same we did in the previous commit.

One thing that I didn't change is that `SVGTransform.matrix` returns the live `DOMMatrix`, which can be set directly, bypassing `applyState`. I'm not sure if this is an oversight or something you deemed acceptable.  I think the solution would be something like the attachement on DOMPointReadOnly.

mh0pe:
@karlseguin Thanks — I preserved your `e4b6b329` optimization/worker-safety commit and fixed the live-matrix concern you identified. Every mutating `DOMMatrix` path on `SVGTransform.matrix` now routes through the attached transform and writes back through `applyState`; removed/replaced items detach, and `animVal` remains read-only.

I also adversarially checked the downstream stack and fixed the same failure/identity class in string-list updates, resource transform cache keys, `SVGAnimatedNumber` WebIDL scalar reflection, and invalid-value default recovery. The corrected heads are now live on #3030, #3033, #3031, #3029, and #3032. Each tip passes its SVG suite independently; the final tip passes 1,085/1,085 repository tests, production/WPT builds, and the applicable focused WPT cases.

Could you please self-assign those five PRs? My fork contributor token cannot set upstream assignees, and I have not sent a reviewer invitation.

PR #3012, opened 2026-07-22
webapi: derive SVG prototype chains

mh0pe:
## Summary

- replace three fixed-depth SVG factory helpers with one compile-time `_proto` walker
- preserve SVG tag casing and namespace while initializing arbitrary graphics and geometry chains
- add hierarchy coverage for parser-created nodes and `createElementNS`

This reauthors the factory-chain subset of #2157 on top of the SVG hierarchy merged in #2917 instead of replaying the obsolete implementation.

## Atomic stack

1. #3012 — prototype-chain foundation
2. https://github.com/lightpanda-io/browser/pull/3034 — live scalar values
3. https://github.com/lightpanda-io/browser/pull/3030 — live collections
4. https://github.com/lightpanda-io/browser/pull/3033 — geometry and paths
5. https://github.com/lightpanda-io/browser/pull/3031 — structural elements
6. https://github.com/lightpanda-io/browser/pull/3029 — resource elements
7. https://github.com/lightpanda-io/browser/pull/3032 — dependency-free text DOM

Dependent units are ready for review and remain stacked on their preceding branch history so every owning commit stays explicit.

## Validation

- rebased onto current `main` at `6f82bdcd`
- `zig fmt` passed
- full release suite passed: 1,090 of 1,090 tests
- full debug/TSAN suite passed: 1,090 of 1,090 tests
- final-stack CI evidence: https://github.com/mh0pe/browser/actions/runs/29922276068

The workflow aggregate is red only because the fork cannot pull the private `ghcr.io/lightpanda-io/perf-fmt` container; that job executed no repository code.

PR #3034, opened 2026-07-23
webapi: add live SVG scalar values

mh0pe:
## Summary

- add live `SVGLength`, `SVGAngle`, `SVGTransform`, and `SVGPreserveAspectRatio` implementations
- expose stable `baseVal`/`animVal` wrappers and root SVG scalar factories
- resolve viewport- and font-relative units with bounded traversal, correct error semantics, and frame-owned wrapper identity
- incorporate all maintainer review feedback, including local-arena serialization and shared conversion paths

## Stack dependency

This is unit 2 of 7. Its predecessor, #3012, is merged. The atomic owning commit is `c40a2cca`.

## Validation

- `zig fmt --check` passed
- local top-of-stack suite passed: 1,065 of 1,065 tests
- fork Zig CI passed: https://github.com/mh0pe/browser/actions/runs/30022136916
- fork plumber CI passed: https://github.com/mh0pe/browser/actions/runs/30022138751
- upstream CLA Assistant passed

Fork e2e functional jobs passed with zero failures; only `cdp-and-hyperfine-bench` remains queued for private `lpd-bench-hetzner` runner capacity.

## Upstream workflow gate

Three upstream GitHub Actions suites are `action_required` pending maintainer approval to run fork code. This is an approval gate, not a test failure.


mh0pe:
@karlseguin the whole merge train should be good to go!

src/browser/webapi/svg/Length.zig
```diff
@@ -0,0 +1,389 @@
...
+        return;
+    };
+    self._value = parsed.value;
+    self._unit = parsed.unit;
+}
+
+fn writeBack(self: *Length, frame: *Frame) !void {
+    const element = self._element orelse return;
+    const value = try self.serialize(frame);
+    try element.setAttributeSafe(self._attr_name, String.wrap(value), frame);
```
karlseguin:
```suggestion
    try element.setAttributeSafe(self._attr_name, .wrap(value), frame);
```

Minor. I wouldn't have commented on _just_ this but since I'm asking for other changes...

src/browser/webapi/svg/Angle.zig
```diff
@@ -0,0 +1,238 @@
...
+}
+
+fn writeBack(self: *Angle, frame: *Frame) !void {
+    const element = self._element orelse return;
+    const value = try self.serialize(frame);
+    try element.setAttributeSafe(self._attr_name, String.wrap(value), frame);
+}
+
+fn serialize(self: *const Angle, frame: *Frame) ![]const u8 {
+    return std.fmt.allocPrint(frame.call_arena, "{d}{s}", .{ self._value, unitSuffix(self._unit) });
```
karlseguin:
```suggestion
    return std.fmt.allocPrint(frame.local_arena, "{d}{s}", .{ self._value, unitSuffix(self._unit) });
```

Recent addition. Shorter lived. Safe for values that need to exist until exactly the next time control is returned to v8 (either by returning from a callback, or invoking v8 directly (e.g. executing a function)). Of course, you don't necessarily know how your caller is going to use the value..so the `call_arena` _is_ safer, but it has its own limited (and hard to reason about) lifetime. In this case, the value is always immediately going to a setAttribute will will clone it anyways.

src/browser/webapi/svg/Angle.zig
```diff
@@ -0,0 +1,238 @@
...
+pub fn getValueInSpecifiedUnits(self: *Angle) f64 {
+    self.syncFromAttribute();
+    return self._value;
+}
+
+pub fn setValueInSpecifiedUnits(self: *Angle, value: f64, frame: *Frame) !void {
+    try self.ensureWritable();
+    try ensureFinite(value);
+    self.syncFromAttribute();
+    if (self._unit == .unknown) self._unit = .unspecified;
```
karlseguin: We aren't super consistent in the codebase, but since I'll likely review all the PRs, I have bad eyesight and appreciate blocks, at least for if statements. Zig itself makes it difficult to be consistent here, since `orelse statement` and `catch statement` are more often than not a single line.

I'll only mention it here. If you feel strongly that it's better as-is, that's fine.

src/browser/webapi/svg/Angle.zig
```diff
@@ -0,0 +1,238 @@
...
+const Frame = @import("../../Frame.zig");
+const Element = @import("../Element.zig");
+
+const String = lp.String;
+const Angle = @This();
+
+_value: f64 = 0,
+_unit: Unit = .unspecified,
+_element: ?*Element = null,
+_attr_name: String = String.empty,
```
karlseguin:
```suggestion
_attr_name: String = .empty,
```

src/browser/webapi/svg/Length.zig
```diff
@@ -0,0 +1,389 @@
...
+}
+
+// The style engine currently exposes inline declarations but does not compute
+// stylesheet font inheritance. Resolve the sources it can represent exactly,
+// then fall back to CSS's initial medium size (16px).
+fn resolvedFontSize(element: ?*Element, frame: *Frame) f64 {
+    const current = element orelse return 16;
+    const parent = current.parentElement();
+
+    if (frame._style_manager.inlineStyleValue(current, comptime String.wrap("font-size"))) |raw| {
```
karlseguin:
```suggestion
    if (frame._style_manager.inlineStyleValue(current, comptime .wrap("font-size"))) |raw| {
```

src/browser/webapi/svg/Length.zig
```diff
@@ -0,0 +1,389 @@
...
+        if (parseFontSize(raw, parent, frame)) |size| return size;
+    }
+    if (current.getAttributeSafe(comptime String.wrap("font-size"))) |raw| {
```
karlseguin:
```suggestion
    if (current.getAttributeSafe(comptime .wrap("font-size"))) |raw| {
```

src/browser/webapi/svg/PreserveAspectRatio.zig
```diff
@@ -0,0 +1,151 @@
...
+    const alignment = alignValue(parts.next() orelse return .{}) orelse return .{};
+    const meet_or_slice = meetOrSliceValue(parts.next() orelse "meet") orelse return .{};
+    if (parts.next() != null) return .{};
+    return .{ .alignment = alignment, .meet_or_slice = meet_or_slice };
+}
+
+fn write(self: *PreserveAspectRatio, value: Value, frame: *Frame) !void {
+    const alignment = alignName(value.alignment) orelse return error.TypeError;
+    const meet_or_slice = meetOrSliceName(value.meet_or_slice) orelse return error.TypeError;
+    const serialized = try std.fmt.allocPrint(frame.call_arena, "{s} {s}", .{ alignment, meet_or_slice });
```
karlseguin:
```suggestion
    const serialized = try std.fmt.allocPrint(frame.local_arena, "{s} {s}", .{ alignment, meet_or_slice });
```

src/browser/webapi/svg/Length.zig
```diff
@@ -0,0 +1,389 @@
...
+}
+
+fn writeBack(self: *Length, frame: *Frame) !void {
+    const element = self._element orelse return;
+    const value = try self.serialize(frame);
+    try element.setAttributeSafe(self._attr_name, String.wrap(value), frame);
+}
+
+fn serialize(self: *const Length, frame: *Frame) ![]const u8 {
+    return std.fmt.allocPrint(frame.call_arena, "{d}{s}", .{ self._value, unitSuffix(self._unit) });
```
karlseguin:
```suggestion
    return std.fmt.allocPrint(frame.local_arena, "{d}{s}", .{ self._value, unitSuffix(self._unit) });
```

src/browser/webapi/element/svg/Svg.zig
```diff
@@ -67,6 +77,46 @@ pub const JsApi = struct {
 
     pub const getElementById = bridge.function(Svg.getElementById, .{});
 
+    pub const x = bridge.accessor(_getX, null, .{});
+    fn _getX(self: *Svg, frame: *Frame) !*AnimatedLength {
```
karlseguin: I was going to complain that these should be methods on SVG. We try to limit JsApi wrappers for v8/js-specific binding (e.g. returning null if an index is < 0 else a value or undefined..very JS-specific). But I see that it's based on my own violation of that pattern...If you want to fix it, please go ahead.

src/browser/webapi/svg/Length.zig
```diff
@@ -0,0 +1,389 @@
...
+        .vertical => ancestorViewportDimension(element, .vertical, frame),
+        .unspecified => blk: {
+            const width = ancestorViewportDimension(element, .horizontal, frame);
+            const height = ancestorViewportDimension(element, .vertical, frame);
+            break :blk @sqrt((width * width + height * height) / 2.0);
+        },
+    };
+}
+
+fn ancestorViewportDimension(element: *Element, direction: Direction, frame: *Frame) f64 {
```
karlseguin, line 204: Add a depth limit? 32?

src/browser/webapi/svg/Length.zig
```diff
@@ -0,0 +1,389 @@
...
+}
+
+fn fontSize(self: *const Length, frame: *Frame) f64 {
+    return resolvedFontSize(self._element, frame);
+}
+
+// The style engine currently exposes inline declarations but does not compute
+// stylesheet font inheritance. Resolve the sources it can represent exactly,
+// then fall back to CSS's initial medium size (16px).
+fn resolvedFontSize(element: ?*Element, frame: *Frame) f64 {
```
karlseguin, line 264: Add a depth limit?

src/browser/webapi/svg/Length.zig
```diff
@@ -0,0 +1,389 @@
...
+    const element = self._element orelse return;
+    const value = try self.serialize(frame);
+    try element.setAttributeSafe(self._attr_name, String.wrap(value), frame);
+}
+
+fn serialize(self: *const Length, frame: *Frame) ![]const u8 {
+    return std.fmt.allocPrint(frame.call_arena, "{d}{s}", .{ self._value, unitSuffix(self._unit) });
+}
+
+fn unitToUserUnits(self: *const Length, unit: Unit, frame: *Frame) f64 {
```
karlseguin, line 175: It would be nice to see if `unitToUserUnits`, `resolveParsedLength` and `parseFontSize` could be deduped.  I think they could handle their own special cases, and then rely on a shared function for the rest...

src/browser/webapi/element/svg/Svg.zig
```diff
@@ -28,11 +28,21 @@ const DOMRect = @import("../../DOMRect.zig");
...
+const SvgAngle = @import("../../svg/Angle.zig");
+const SvgTransform = @import("../../svg/Transform.zig");
+const AnimatedLength = @import("../../svg/AnimatedLength.zig");
+const AnimatedPreserveAspectRatio = @import("../../svg/AnimatedPreserveAspectRatio.zig");
 
 const Graphics = @import("Graphics.zig");
 
 const Svg = @This();
 _proto: *Graphics,
+_x: ?*AnimatedLength = null,
```
karlseguin: I'm curious if you considered a frame lookup for these? Element uses this a lot. It's probably less important here, since we expect far fewer SVG elements than Nodes/Elements, but it would be more consistent.

Your animated length would have:
```zig
pub const Kind = enum {
    x, y, width, height, rx, ry, cx, cy, r, x1, y1, x2, y2,

    fn attributeName(self: Kind) String {
        return switch (self) {
            .x => comptime .wrap("x"),
            // ...
        };
    }

    fn direction(self: Kind) Length.Direction {
        return switch (self) {
            .x, .width, .rx, .cx, .x1, .x2 => .horizontal,
            .y, .height, .ry, .cy, .y1, .y2 => .vertical,
            .r => .unspecified,
        };
    }
};

pub const Key = struct { element: *Element, kind: Kind };

pub fn getOrCreate(element: *Element, kind: Kind, frame: *Frame) !*AnimatedLength {
    const gop = try frame._svg_animated_lengths.getOrPut(frame.arena, .{ .element = element, .kind = kind });
    if (!gop.found_existing) {
        gop.value_ptr.* = try create(element, kind.attributeName(), kind.direction(), frame);
    }
    return gop.value_ptr.*;
}
```

With every getter becoming something like:

```
return AnimatedLength.getOrCreate(self.asElement(), .x, frame)
```

src/browser/webapi/svg/Length.zig
mh0pe: Implemented in c40a2cca: the wrapped value is passed directly to setAttributeSafe.

src/browser/webapi/svg/Angle.zig
mh0pe: Implemented in c40a2cca: Angle serialization now uses frame.local_arena.

src/browser/webapi/svg/Angle.zig
mh0pe: Implemented in c40a2cca: the affected conditionals now use blocks.

src/browser/webapi/svg/Angle.zig
mh0pe: Implemented in c40a2cca: the attribute name initializes with .empty.

src/browser/webapi/svg/Length.zig
mh0pe: Implemented in c40a2cca: the font-size lookup uses a comptime-wrapped key.

src/browser/webapi/svg/Length.zig
mh0pe: Implemented in c40a2cca: the attribute lookup uses a comptime-wrapped key.

src/browser/webapi/svg/PreserveAspectRatio.zig
mh0pe: Implemented in c40a2cca: preserveAspectRatio serialization uses frame.local_arena.

src/browser/webapi/svg/Length.zig
mh0pe: Implemented in c40a2cca: Length serialization now uses frame.local_arena.

src/browser/webapi/element/svg/Svg.zig
mh0pe: Implemented in c40a2cca: these helpers are methods on SVG rather than JsApi wrappers.

src/browser/webapi/svg/Length.zig
mh0pe: Implemented in c40a2cca: viewport traversal has a 32-level recursion budget and a deep-nesting regression test.

src/browser/webapi/svg/Length.zig
mh0pe: Implemented in c40a2cca: font-size inheritance has a 32-level recursion budget and a deep-nesting regression test.

src/browser/webapi/svg/Length.zig
mh0pe: Implemented in c40a2cca: shared absolute/relative unit conversion is centralized and the special cases delegate to it.

src/browser/webapi/element/svg/Svg.zig
mh0pe: Implemented in c40a2cca: animated lengths and preserveAspectRatio values use transactional frame-owned lookup caches keyed by element and kind.

mh0pe:
@karlseguin final follow-up: all 13 review threads on this PR have now been answered and resolved. The current head is ready and GitHub reports it mergeable; the local suite passes 1,065/1,065 and all six fork Zig workflows are green. The successor PRs are exact one-commit units in this merge order: #3034 -> #3030 -> #3033 -> #3031 -> #3029 -> #3032. The remaining upstream workflow suites are action_required pending maintainer approval of fork code. Thanks again for the review.

PR #3037, opened 2026-07-23
cdp: complete Network events and include worker requests

Ppsoft1991:
## Summary

- add required CDP fields to Network events and their request, response, and timing payloads
- emit Network.requestWillBeSent for WorkerGlobalScope requests that are not present in the document frame tree
- add a regression fixture covering both the Worker script and a Fetch issued from inside the Worker

## Problem

Strict CDP clients discard Lightpanda Network messages when protocol-required fields are absent. Worker requests had a separate issue: their WorkerGlobalScope frame IDs do not resolve through the document frame tree, so requestWillBeSent returned early while later response events were still emitted.

## Testing

- targeted CDP Network tests: 9/9 passed
- full Lightpanda test suite with Zig 0.16.0: 1062/1062 passed
- formatting checked for every tracked Zig source
- external CDP probe captured the document, an eval-issued Fetch, the Worker script, and a Fetch issued inside the Worker

### DIFF
--- a/src/cdp/domains/network.zig
+++ b/src/cdp/domains/network.zig
@@ -328,6 +328,7 @@ pub fn httpRequestFail(bc: *CDP.BrowserContext, msg: *const Notification.Request
     // We're missing a bunch of fields, but, for now, this seems like enough
     try bc.cdp.sendEvent("Network.loadingFailed", .{
         .requestId = &id.toRequestId(msg.transfer),
+        .timestamp = timestamp(.monotonic),
         // Seems to be what chrome answers with. I assume it depends on the type of error?
         .type = "Ping",
         .errorText = msg.err,
@@ -382,6 +383,8 @@ pub fn httpResponseHeaderDone(arena: Allocator, bc: *CDP.BrowserContext, msg: *c
         .frameId = &id.toFrameId(req.frame_id),
         .requestId = &id.toRequestId(transfer),
         .loaderId = &id.toLoaderId(req.loader_id),
+        .timestamp = timestamp(.monotonic),
+        .type = req.resource_type.string(),
         .response = ResponseWriter.init(arena, msg.transfer),
         .hasExtraInfo = false, // TODO change after adding Network.responseReceivedExtraInfo
     }, .{ .session_id = session_id });
@@ -393,6 +396,7 @@ pub fn httpRequestDone(bc: *CDP.BrowserContext, msg: *const Notification.Request
     const session_id = bc.session_id orelse return;
     try bc.cdp.sendEvent("Network.loadingFinished", .{
         .requestId = &id.toRequestId(msg.transfer),
+        .timestamp = timestamp(.monotonic),
         .encodedDataLength = msg.content_length,
     }, .{ .session_id = session_id });
 }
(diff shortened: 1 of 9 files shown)


src/cdp/domains/network.zig
```diff
@@ -430,6 +443,16 @@ pub const RequestWriter = struct {
             try jws.write(request.body != null);
         }
 
+        {
+            try jws.objectField("initialPriority");
+            try jws.write("High");
+        }
+
+        {
```
karlseguin, line 459: I don't think this is accurate. Unless you have a strong need for this right now, I'd rather leave it out until we can generate a correct value

src/cdp/domains/network.zig
```diff
@@ -499,6 +522,17 @@ const ResponseWriter = struct {
             try jws.write(transfer._from_cache);
         }
 
+        {
+            try jws.objectField("connectionReused");
+            try jws.write(false);
```
karlseguin: This is a field we _could_ accurately provide. Do you think it's important? I can do it in a follow up commit.

src/cdp/domains/network.zig
```diff
@@ -499,6 +522,17 @@ const ResponseWriter = struct {
...
+            try jws.objectField("connectionId");
+            try jws.write(0);
```
karlseguin: This is a field we _could_ accurately provide. Do you think it's important? I can do it in a follow up commit.

karlseguin:
I just rebased, didn't touch anything else.

src/cdp/domains/network.zig
Ppsoft1991: I agree that hardcoding strict-origin-when-cross-origin is inaccurate. However, chromiumoxide treats referrerPolicy as required and drops the entire Network.requestWillBeSent event when it is missing. Could we expose the policy Lightpanda actually applies instead of omitting the field?

src/cdp/domains/network.zig
Ppsoft1991: This field is important for us because chromiumoxide treats connectionReused as required and drops the entire Network.responseReceived event when it is missing. An accurately generated value would be preferable.

src/cdp/domains/network.zig
Ppsoft1991: This field is also important for us because chromiumoxide treats connectionId as required and drops the entire Network.responseReceived event when it is missing. An accurately generated value would be preferable.

Ppsoft1991: Would CURLINFO_CONN_ID be an appropriate source here? libcurl documents it as the ID of the last connection used by the handle, unique within the connection cache shared by the multi handle, which seems to match CDP’s physical connection ID semantics closely.

karlseguin: I pushed a commit that did just this.

src/cdp/domains/network.zig
karlseguin: I set it to `unsafe-url`.

karlseguin:
@Ppsoft1991 are you able to test the last version of this PR on your use case?

Ppsoft1991:
> @Ppsoft1991 are you able to test the last version of this PR on your use case?

I tested the latest version at 5acfd382:
SSO navigation completed successfully, including cross-origin redirects.
Network events for Documents, Scripts, XHR, eval-triggered Fetch, Worker scripts, and Worker-triggered Fetch were captured successfully.
The strict CDP client deserialized all Network events without missing-field errors.
Network tests passed 10/10; the full test suite passed 1074/1074.
The latest version works as expected. Thank you.

PR #3039, opened 2026-07-23
cli: fix help pager broken by lp.io's failing allocator

arrufat:
Since the Zig 0.16 migration, `lightpanda help <command>` no longer pages through `$PAGER`/`less`.

`lp.io` is backed by `std.Io.Threaded.init_single_threaded`, which sets `.allocator = .failing`; `spawnPosix` allocates the child's argv/env through it, so `std.process.spawn(lp.io, ...)` always returns `OutOfMemory` and `printPaged` silently fell back to plain output. (The doc comment on `Threaded.init` claiming a failing allocator only affects `Io.async`/`Io.concurrent` is stale.)

Fix: spawn the pager through a local `Threaded` instance with a real allocator and the real environ — the single-threaded instance's environ is empty, which would also break PATH resolution of the bare `less` fallback.

Verified under a pty: `PAGER="wc -l" lightpanda help serve` prints the line count, and the no-`PAGER` case enters `less`.

### DIFF
--- a/src/Config.zig
+++ b/src/Config.zig
@@ -922,22 +922,27 @@ fn printPaged(allocator: Allocator, text: []const u8) void {
     var environ_map = lp.environMap(allocator) catch return printPlain(text);
     defer environ_map.deinit();
 
-    var child = std.process.spawn(lp.io, .{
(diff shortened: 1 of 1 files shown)


karlseguin:
LGTM

PR #3051, opened 2026-07-24
net: accept Blob and File values in FormData.append/set

### LINKED ISSUE #3050: net: FormData.append(name, File|Blob) stringifies the value — file bytes never reach the wire
### Summary

`FormData.append(name, value)` accepts only a string on the Zig side, so passing a `File` or `Blob` from JS coerces it through `toString()`. The entry stores the literal text `[object File]` / `[object Blob]` and the file's bytes never reach the wire.

A `FormData` carrying a 200 KB `File` produces a **209-byte** request body:

```
Content-Type: multipart/form-data; boundary=----cecb5564245dd5d65950

------cecb5564245dd5d65950
Content-Disposition: form-data; name="field"

plain-string
------cecb5564245dd5d65950
Content-Disposition: form-data; name="doc"

[object File]
------cecb5564245dd5d65950--
```

Three things are wrong in that part: the bytes are gone, there is no `filename=` parameter, and there is no per-part `Content-Type`.

The request still succeeds with a 200, so nothing surfaces the loss — the server simply stores a 13-byte "file".

### Reproducer

Dependency-free: `python3` (stdlib only) plus a `lightpanda` binary. No Node, no npm.

<details>
<summary><code>repro.html</code></summary>

```html
<!doctype html>
<meta charset="utf-8" />
<title>FormData.append with a File value</title>
<body>
  <script>
    var payload = "x".repeat(200 * 1024);
    var fd = new FormData();
    fd.append("field", "plain-string");
    fd.append("doc", new File([payload], "big.txt", { type: "text/plain" }));

    fetch("/upload", { method: "POST", body: fd });
  </script>
</body>
```

</details>

<details>
<summary><code>server.py</code> — serves the fixture, records the POST body</summary>

```python
#!/usr/bin/env python3
import http.server, socketserver, sys, pathlib

PORT = int(sys.argv[1])
HERE = pathlib.Path(__file__).parent
PAYLOAD_SIZE = 200 * 1024

    def log_message(self, *args): pass

    def do_GET(self):
        body = (HERE / "repro.html").read_bytes()
        self.send_response(200)
        self.send_header("Content-Type", "text/html")
        self.end_headers()
        self.wfile.write(body)

    def do_POST(self):
        declared = int(self.headers.get("Content-Length", 0))
        body = self.rfile.read(declared)
        (HERE / "result.txt").write_text(
            "content_type=%s\ndeclared=%d\nreceived=%d\npayload_expected=%d\nbody_head=%s\n"
            % (self.headers.get("Content-Type", ""), declared, len(body),
               PAYLOAD_SIZE, body[:300].decode("utf-8", "replace")))
        self.send_response(200)
        self.send_header("Content-Length", "2")
        self.end_headers()
        self.wfile.write(b"ok")

socketserver.TCPServer.allow_reuse_address = True
    httpd.serve_forever()
```

</details>

<details>
<summary><code>repro.sh</code> — exits 1 today, 0 once fixed</summary>

```bash
#!/usr/bin/env bash
set -uo pipefail

HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
LIGHTPANDA="${LIGHTPANDA_BIN:-lightpanda}"
PORT="${PORT:-8731}"
PAYLOAD=$((200 * 1024))

rm -f "$HERE/result.txt"
python3 "$HERE/server.py" "$PORT" &
SERVER_PID=$!
trap 'kill $SERVER_PID 2>/dev/null' EXIT

for _ in $(seq 1 50); do
  curl -fsS "http://127.0.0.1:$PORT/" -o /dev/null 2>/dev/null && break
  sleep 0.1
done

"$LIGHTPANDA" fetch "http://127.0.0.1:$PORT/" >/dev/null 2>&1
sleep 0.5

[ -f "$HERE/result.txt" ] || { echo "INCONCLUSIVE: no POST was sent."; exit 2; }
cat "$HERE/result.txt"

RECEIVED=$(grep '^received=' "$HERE/result.txt" | cut -d= -f2)
if [ "$RECEIVED" -lt "$PAYLOAD" ]; then
  echo "FAIL: body is ${RECEIVED} bytes; the ${PAYLOAD}-byte File value was dropped."
  exit 1
fi
grep -q 'filename=' "$HERE/result.txt" || { echo "FAIL: no filename= parameter."; exit 1; }
echo "PASS"
```

</details>

Run it:

```
LIGHTPANDA_BIN=/path/to/lightpanda bash repro.sh
```

Observed on nightly `1.0.0-nightly.8285+de85a51d` (macOS aarch64):

```
FAIL: multipart body is 209 bytes; the 204800-byte File value was dropped.
```

### Broken flow

```mermaid
sequenceDiagram
    participant JS as Page script
    participant FD as FormData
    participant MW as multipart writer
    participant SRV as Server

    JS->>FD: append a 200KB File value
    Note over FD: js.Bridge coerces the argument<br/>to a string via toString
    FD->>FD: entry value becomes the text object File
    JS->>FD: fetch with the FormData as body
    FD->>MW: write multipart
    MW->>SRV: part carries the literal text object File
    Note over SRV: 209-byte body<br/>no filename parameter<br/>no per-part Content-Type
```

### Expected flow

```mermaid
sequenceDiagram
    participant JS as Page script
    participant FD as FormData
    participant MW as multipart writer
    participant SRV as Server

    JS->>FD: append a 200KB File value
    Note over FD: value is a File,<br/>so create a file entry
    FD->>FD: entry value becomes a File pointer
    JS->>FD: fetch with the FormData as body
    FD->>MW: write multipart
    MW->>SRV: part carries filename big.txt<br/>Content-Type text/plain<br/>and 200KB of bytes
    Note over SRV: about 205KB body
```

### Where it comes from

`src/browser/webapi/net/FormData.zig`:

```zig
pub fn append(self: *FormData, name: []const u8, value: []const u8) !void {
    try self._entries.append(self._arena, .{
        .name = try String.init(self._arena, name, .{}),
        .value = .{ .string = try String.init(self._arena, value, .{}) },
    });
}
```

`value` is `[]const u8`, and `JsApi` exposes it directly with `bridge.function(FormData.append, .{})`, so `js.Bridge` maps any JS argument to a string. `set` has the same signature and calls straight through to `append`.

The storage layer is already there — only the binding is string-typed:

```zig
const Value = union(enum) {
    no_file: void,
    file: *File,
    string: String,
    ...
};
```

`Entry.Value.file` exists, `deleteByName` already does `file.releaseRef(exec.page)`, `init` already does `file.acquireRef()`, and the multipart writer already emits `filename=` and a per-part `Content-Type` for `.file` entries — that is how `<input type=file>` submissions work today (#2635, #2654). The JS `append`/`set` path is the one entry point that cannot construct a `.file` entry.

`body_init.zig` shows the accepted shape for taking a union from JS:

```zig
pub const BodyInit = union(enum) {
    blob: *Blob,
    ...
    bytes: []const u8, // must be last, js.Bridge will map anything to a string
};
```

### Expected behavior

Per the [XHR Standard's "create an entry" algorithm](https://xhr.spec.whatwg.org/#create-an-entry):

> 2. If *value* is a `Blob` object and not a `File` object, then set *value* to a new `File` object, representing the same bytes, whose `name` attribute value is `"blob"`.
> 3. If *value* is (now) a `File` object and *filename* is given, then set *value* to a new `File` object, representing the same bytes, whose `name` attribute value is *filename*.

And per the [HTML Standard's multipart/form-data encoding algorithm](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#multipart/form-data-encoding-algorithm), a file entry contributes a `filename` parameter and a `Content-Type` header for the part.

So `append(name, value, filename?)` should accept `File`/`Blob` and build a `.file` entry, with `set` following the same path.

### Impact

Every upload whose `FormData` is assembled in JS rather than by a native form submission: drag-and-drop uploaders, cropped-avatar widgets that hand over a canvas `Blob`, chunked/direct-to-storage upload flows, and anything doing `fd.append("file", blob)`. Both `fetch` and `XMLHttpRequest` are affected since they share the encoder. Native `<input type=file>` form submission is unaffected and works correctly.

navidemad:
Closes #3050

### Problem

The JS-facing `append()` / `set()` took `value: []const u8`, so `js.Bridge` coerced any object argument through `toString()`. A `Blob` or `File` was stored as the literal text `[object Blob]` / `[object File]` and its bytes never reached the wire.

A `FormData` carrying a 200 KB `File` produced a 209-byte body — no `filename` parameter, no per-part `Content-Type`, no payload:

```
------cecb5564245dd5d65950
Content-Disposition: form-data; name="doc"

[object File]
------cecb5564245dd5d65950--
```

The request still returned 200, so nothing surfaced the loss.

### What changed

Only the binding was string-typed. `Entry.Value` already carries a `file: *File` variant with refcounting, and the multipart writer already emits `filename=` and a per-part `Content-Type` for it — that is how `<input type=file>` submissions work today. The JS entry point was the one place that could not construct a file entry.

```mermaid
flowchart TD
    subgraph after["after"]
        A2["append(name, value, filename?)"] --> B2{"EntryValue tag"}
        B2 -->|bytes| C2["Entry.Value.string"]
        B2 -->|blob| D2["File named blob<br/>or the given filename"]
        B2 -->|file| E2{"filename given?"}
        E2 -->|no| F2["acquireRef, reuse the File"]
        E2 -->|yes| G2["new File over the same bytes"]
        D2 --> H2["Entry.Value.file"]
        F2 --> H2
        G2 --> H2
        C2 --> I2["multipart writer"]
        H2 --> I2
        I2 --> J2["name=doc; filename=big.txt<br/>Content-Type: text/plain<br/>200KB of bytes"]
    end

    subgraph before["before"]
        A1["append(name, value)"] --> B1["value: []const u8"]
        B1 --> C1["js.Bridge toString"]
        C1 --> D1["Entry.Value.string<br/>object File"]
        D1 --> E1["multipart writer"]
        E1 --> F1["name=doc<br/>object File"]
    end
```

`value` is now a tagged union, following the shape `body_init.zig` already uses for `BodyInit` — `bytes` stays last so `js.Bridge` only falls back to a string once the JsApi classes have been tried:

```zig
pub const EntryValue = union(enum) {
    file: *File,
    blob: *Blob,
    bytes: []const u8,
};
```

`jsAppend` implements the [entry-creation steps](https://xhr.spec.whatwg.org/#create-an-entry): a `Blob` that is not a `File` becomes a `File` named `"blob"`, and an explicit `filename` produces a new `File` over the same bytes rather than renaming the caller's object. `jsSet` routes through the same path after deleting existing entries.

`fileFrom` builds that `File` on `Blob.initFromBytes`, mirroring `File.init`'s Blob-and-File-sharing-one-arena layout, and returns it at refcount 1 so the existing `deleteByName` / `deinit` release path stays balanced.

The internal `append(name, value)` helper keeps its string signature. It backs form collection and the existing unit tests, neither of which goes through JS.

### Tests

Added to `src/browser/tests/net/form_data.html` (covered by the existing `WebApi: FormData` runner): a `Blob` value reads back as `blob`, a `File` keeps its own name, an explicit filename overrides both, `set()` takes the same path, and a string value is still a string.

Toggled off against `main`'s `FormData.zig` the fixture fails; restored it passes. Full suite: 1073/1073.

Verified end-to-end with a CDP-free reproducer (`python3` stdlib + the binary, in the issue). Against nightly `1.0.0-nightly.8285+de85a51d` it exits 1:

```
FAIL: multipart body is 209 bytes; the 204800-byte File value was dropped.
```

Against this branch it exits 0:

```
Content-Disposition: form-data; name="doc"; filename="big.txt"
Content-Type: text/plain

xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx...

PASS: the File value was multipart-encoded with its bytes.
```

declared and received both 205042 bytes.

### DIFF
--- a/src/network/layer/CacheLayer.zig
+++ b/src/network/layer/CacheLayer.zig
@@ -63,7 +63,9 @@ fn request(ptr: *anyopaque, client: *Client, req: Request) anyerror!void {
         .timestamp = std.time.timestamp(),
         .request_headers = req_header_list.items,
     })) |cached| {
-        return serveFromCache(req, &cached);
+        try serveFromCache(req, &cached);
+        client.deinitRequest(req);
+        return;
     }
 
     const cache_ctx = try arena.create(CacheContext);


karlseguin:
I accidentally closed this..I'll manually run the CI and merge it.

PR #3056, opened 2026-07-25
fix(net): buffer closed ReadableStream bodies for fetch POST

### LINKED ISSUE #3052: net: fetch with a ReadableStream request body sends an empty body and resolves successfully
### Summary

`fetch(url, { method: "POST", body: <ReadableStream>, duplex: "half" })` sends the request with `Content-Length: 0` and an empty body. The promise resolves with a 200 and nothing reports the loss, so the caller believes the upload succeeded.

Observed on nightly `1.0.0-nightly.8285+de85a51d` (macOS aarch64), for a 21-byte payload:

```
content_type=application/x-www-form-urlencoded
declared=0
received=0
payload_expected=21
```

### Reproducer

Dependency-free: `python3` (stdlib only) plus a `lightpanda` binary. No Node, no npm.

<details>
<summary><code>repro.html</code></summary>

```html
<!doctype html>
<meta charset="utf-8" />
<title>fetch with a ReadableStream request body</title>
<body>
  <script>
    var payload = "name=Streamed&count=9";
    var stream = new ReadableStream({
      start: function (c) {
        c.enqueue(new TextEncoder().encode(payload));
        c.close();
      },
    });

    fetch("/upload", {
      method: "POST",
      body: stream,
      duplex: "half",
      headers: { "Content-Type": "application/x-www-form-urlencoded" },
    });
  </script>
</body>
```

</details>

<details>
<summary><code>server.py</code> — serves the fixture, records what arrives</summary>

```python
#!/usr/bin/env python3
import http.server, socketserver, sys, pathlib

PORT = int(sys.argv[1])
HERE = pathlib.Path(__file__).parent
PAYLOAD_SIZE = 21

    def log_message(self, *args): pass

    def do_GET(self):
        body = (HERE / "repro.html").read_bytes()
        self.send_response(200)
        self.send_header("Content-Type", "text/html")
        self.end_headers()
        self.wfile.write(body)

    def do_POST(self):
        declared = int(self.headers.get("Content-Length", 0))
        body = self.rfile.read(declared)
        (HERE / "result.txt").write_text(
            "content_type=%s\ndeclared=%d\nreceived=%d\npayload_expected=%d\n"
            % (self.headers.get("Content-Type", ""), declared, len(body), PAYLOAD_SIZE))
        self.send_response(200)
        self.send_header("Content-Length", "2")
        self.end_headers()
        self.wfile.write(b"ok")

socketserver.TCPServer.allow_reuse_address = True
    httpd.serve_forever()
```

</details>

<details>
<summary><code>repro.sh</code> — exits 1 today; passes on either candidate fix</summary>

```bash
#!/usr/bin/env bash
set -uo pipefail

HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
LIGHTPANDA="${LIGHTPANDA_BIN:-lightpanda}"
PORT="${PORT:-8732}"
PAYLOAD=21

rm -f "$HERE/result.txt"
python3 "$HERE/server.py" "$PORT" &
SERVER_PID=$!
trap 'kill $SERVER_PID 2>/dev/null' EXIT

for _ in $(seq 1 50); do
  curl -fsS "http://127.0.0.1:$PORT/" -o /dev/null 2>/dev/null && break
  sleep 0.1
done

"$LIGHTPANDA" fetch "http://127.0.0.1:$PORT/" >/dev/null 2>&1
sleep 0.5

# A rejected stream body sends nothing at all, which is an acceptable outcome.
[ -f "$HERE/result.txt" ] || { echo "PASS: no request was sent."; exit 0; }

cat "$HERE/result.txt"
RECEIVED=$(grep '^received=' "$HERE/result.txt" | cut -d= -f2)
if [ "$RECEIVED" -lt "$PAYLOAD" ]; then
  echo "FAIL: sent carrying ${RECEIVED} of ${PAYLOAD} bytes."
  exit 1
fi
echo "PASS"
```

</details>

Run it:

```
LIGHTPANDA_BIN=/path/to/lightpanda bash repro.sh
```

### Broken flow

```mermaid
sequenceDiagram
    participant JS as Page script
    participant BI as body_init extract
    participant HTTP as HttpClient
    participant SRV as Server

    JS->>BI: fetch POST with a ReadableStream body
    Note over BI: the stream arm returns<br/>bytes empty and content_type null
    BI->>HTTP: body is zero bytes
    HTTP->>SRV: POST with Content-Length 0
    SRV-->>JS: 200 OK
    Note over JS: promise resolves,<br/>payload silently lost
```

### Expected flow

```mermaid
sequenceDiagram
    participant JS as Page script
    participant BI as body_init extract
    participant HTTP as HttpClient
    participant SRV as Server

    JS->>BI: fetch POST with a ReadableStream body
    alt stream is drained
        BI->>HTTP: body is the 21 collected bytes
        HTTP->>SRV: POST with Content-Length 21
        SRV-->>JS: 200 OK
    else streaming is unsupported here
        BI-->>JS: reject with a TypeError
        Note over SRV: no request is sent
    end
```

### Where it comes from

`src/browser/webapi/net/body_init.zig`:

```zig
.stream => {
    // Callers that support streaming bodies (Response) special-case
    // the `.stream` arm before calling extract; the Request/XHR
    // paths that reach here have no place to store a stream, so
    // ... stream carries no default Content-Type.
    return .{ .bytes = "", .content_type = null };
},
```

The comment shows this is a deliberate shortcut rather than an oversight, so the issue is not that the arm exists — it is that the caller gets no signal. `Request.zig` calls `b.extract(arena)` and uses `extracted.bytes` verbatim, so the empty slice flows straight to the wire.

`Response.zig` already handles the mirror-image situation the other way, and rejects instead of returning nothing:

```zig
.stream => return local.rejectPromise(.{ .type_error = "Cannot read text from stream body" }),
```

### Expected behavior

Per the [Fetch Standard's body extraction](https://fetch.spec.whatwg.org/#concept-bodyinit-extract), a `ReadableStream` body sets the body's stream to that object, and the [`duplex` member](https://fetch.spec.whatwg.org/#dom-requestinit-duplex) exists precisely so a request may carry a streaming body. Chrome transmits the bytes. Silently transmitting none is the one outcome the spec does not allow.

Three ways to close the gap, in increasing order of work:

1. **Reject.** Throw a `TypeError` from the Request/XHR path when the body is a stream, matching what `Response` already does on the read side. Smallest change, and it converts silent data loss into an obvious error.
2. **Buffer.** Drain the stream to completion during extraction, then send it as ordinary bytes with a real `Content-Length`. Covers what almost every real caller wants, at the cost of losing the streaming property.
3. **Stream.** Feed the reader into the HTTP client with chunked transfer encoding. Fully spec-correct and the largest change.

Happy to write the patch — which of these would you want? I did not want to guess, since 1 and 2 differ in observable behavior.

### Impact

Narrower than most fetch bugs, since it needs a `ReadableStream` handed to `body` explicitly: streaming/chunked uploaders, and code that pipes one response's `.body` into a new request. The severity comes from the failure mode rather than the frequency — the request succeeds, so nothing upstream or downstream notices the payload is gone.

arimu1:
## Summary

- Fixes https://github.com/lightpanda-io/browser/issues/3052 where `fetch()` with a `ReadableStream` request body silently sent `Content-Length: 0` and an empty payload.
- Adds `ReadableStream.collectBodyBytes` to drain a **closed** stream's internal queue synchronously; open, locked, or errored streams reject with `TypeError` instead of sending nothing.
- JS `controller.enqueue` stores chunks as `.js_value` via `enqueueValue`; `collectBodyBytes` now converts TypedArray/ArrayBuffer/string `.js_value` entries (matching `Response.StreamConsumer._onReadFulfilled`) before buffering.
- `BodyInit.extract` `.stream` arm uses that helper instead of returning an empty byte slice.
- HTML fixtures: `fetch_readablestream_body` (multi-chunk success), `fetch_readablestream_body_reject` (open + locked streams → TypeError), plus `/xhr/echo_post_body` test-server endpoint.

## Approach

Implements option **2 (buffer)** from the issue for the common case (stream closed synchronously in `start()` with enqueued chunks), and option **1 (reject)** for streams that cannot be drained synchronously. This matches the issue's expected behavior diagram and avoids silent data loss.

## Test plan

- [x] `zig fmt --check` on changed `.zig` files — clean
- [ ] `TEST_FILTER="fetch_readablestream_body" make test` — **not run locally**: V8 `gclient sync` failed (`_bad_scm` conflicts; also `No space left on device` during llvm artifact extract). Maintainer CI should exercise the HTML fixtures.
- [ ] Full `make test` — blocked by same V8 cache/disk issue locally

## CLA

I will sign the CLA via cla-assistant-lite when prompted on this PR (first contribution to this repo).

Fixes #3052

### DIFF
--- a/src/browser/tests/net/fetch.html
+++ b/src/browser/tests/net/fetch.html
@@ -352,9 +352,11 @@
   {
     const state = await testing.async();
     const payload = "name=Streamed&count=9";
+    const enc = new TextEncoder();
     const stream = new ReadableStream({
       start(controller) {
-        controller.enqueue(new TextEncoder().encode(payload));
+        controller.enqueue(enc.encode("name=Streamed&"));
+        controller.enqueue(enc.encode("count=9"));
         controller.close();
       },
     });
@@ -371,6 +373,44 @@
     await state.done(() => {
       testing.expectEqual(200, response.status);
       testing.expectEqual(payload, text);
+    });
+  }
+</script>
+
+<script id=fetch_readablestream_body_reject type=module>
+  {
+    const state = await testing.async();
+    const url = "http://127.0.0.1:9582/xhr/echo_post_body";
+    const opts = { method: "POST", duplex: "half" };
+
+    let openRejected = false;
+    try {
+      const open = new ReadableStream({
+        start(c) { c.enqueue(new TextEncoder().encode("x")); },
+      });
+      await fetch(url, { ...opts, body: open });
+    } catch (e) {
+      openRejected = e instanceof TypeError;
+    }
+
+    let lockedRejected = false;
+    try {
+      const locked = new ReadableStream({
+        start(c) {
+          c.enqueue(new TextEncoder().encode("x"));
+          c.close();
+        },
+      });
+      locked.getReader();
+      await fetch(url, { ...opts, body: locked });
+    } catch (e) {
+      lockedRejected = e instanceof TypeError;
+    }
+
+    state.resolve();
+    await state.done(() => {
+      testing.expectEqual(true, openRejected);
+      testing.expectEqual(true, lockedRejected);
     });
   }
 </script>
--- a/src/browser/tests/net/fetch.html
+++ b/src/browser/tests/net/fetch.html
@@ -348,6 +348,33 @@
   }
 </script>
 
+<script id=fetch_readablestream_body type=module>
+  {
+    const state = await testing.async();
+    const payload = "name=Streamed&count=9";
+    const stream = new ReadableStream({
+      start(controller) {
+        controller.enqueue(new TextEncoder().encode(payload));
+        controller.close();
+      },
+    });
+
+    const response = await fetch("http://127.0.0.1:9582/xhr/echo_post_body", {
+      method: "POST",
+      body: stream,
+      duplex: "half",
+      headers: { "Content-Type": "application/x-www-form-urlencoded" },
+    });
+    const text = await response.text();
+    state.resolve();
+
+    await state.done(() => {
+      testing.expectEqual(200, response.status);
+      testing.expectEqual(payload, text);
+    });
+  }
+</script>
+
 <script id=fetch_blob_url_in_worker type=module>
   {
     const state = await testing.async();
--- a/src/browser/tests/net/fetch.html
+++ b/src/browser/tests/net/fetch.html
@@ -361,7 +361,7 @@
       },
     });
 
-    const response = await fetch("http://127.0.0.1:9582/xhr/echo_post_body", {
+    const response = await fetch("http://127.0.0.1:9582/echo_body", {
       method: "POST",
       body: stream,
       duplex: "half",
@@ -380,7 +380,7 @@
 <script id=fetch_readablestream_body_reject type=module>
   {
     const state = await testing.async();
-    const url = "http://127.0.0.1:9582/xhr/echo_post_body";
+    const url = "http://127.0.0.1:9582/echo_body";
     const opts = { method: "POST", duplex: "half" };
 
     let openRejected = false;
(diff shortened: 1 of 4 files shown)


arimu1:

arimu1:
recheck

karlseguin:
Thanks.

PR #3075, opened 2026-07-28
agent: support subscription auth via Codex

arrufat:


src/agent/auth/auth.zig
```diff
@@ -0,0 +1,275 @@
...
+    defer arena.deinit();
+    const a = arena.allocator();
+    const path = try storePath(a, dir);
+    const data = std.Io.Dir.cwd().readFileAlloc(lp.io, path, a, .limited(64 * 1024)) catch return null;
+    const parsed = std.json.parseFromSliceLeaky(std.json.ArrayHashMap(StoredToken), a, data, .{ .ignore_unknown_fields = true }) catch return null;
+    const t = parsed.map.get(id) orelse return null;
+    return try TokenSet.dup(allocator, t.access, t.refresh, t.expires_at_ms, t.account_id);
+}
+
+fn storeSaveAt(dir: []const u8, id: []const u8, tokens: TokenSet) !void {
```
karlseguin: There appears to be a single caller, and it has an arena, why not pass that instead of creating another one on the `page_allocator`? (Same for the other functions)

src/agent/auth/auth.zig
```diff
@@ -0,0 +1,275 @@
...
+        .refresh = tokens.refresh_token,
+        .expires_at_ms = tokens.expires_at_ms,
+        .account_id = tokens.account_id,
+    });
+
+    var buf: std.Io.Writer.Allocating = .init(a);
+    try std.json.Stringify.value(map, .{}, &buf.writer);
+    try std.Io.Dir.cwd().writeFile(lp.io, .{ .sub_path = path, .data = buf.written() });
+    // Secrets file: owner read/write only.
+    _ = std.c.chmod(path.ptr, 0o600);
```
karlseguin: This can fail. If you create the file upfront, can you set the permission and then write? (You could then also Stringify directly to the File's writer writer)

src/agent/auth/auth.zig
```diff
@@ -0,0 +1,275 @@
...
+pub fn sessionFor(allocator: std.mem.Allocator, provider: Config.AiProvider) !?Session {
+    const desc = descriptorFor(provider) orelse return null;
+    const tokens = (try storeLoad(allocator, desc.storeKey())) orelse return null;
+    return .{ .allocator = allocator, .descriptor = desc, .tokens = tokens };
+}
+
+/// Run the interactive login and persist the result, returning a live session.
+pub fn login(allocator: std.mem.Allocator, desc: *const Descriptor, interrupt: ?*zenai.http.Interrupt) !Session {
+    const tokens = try desc.loginFn(allocator, interrupt);
+    storeSave(desc.storeKey(), tokens) catch {};
```
karlseguin: Could be useful to know these are failing, no?

src/agent/auth/codex.zig
```diff
@@ -0,0 +1,269 @@
...
+    if (code_res.status != .ok) return error.DeviceCodeRequestFailed;
+    const dc = try std.json.parseFromSliceLeaky(DeviceCode, a, code_res.body, .{ .ignore_unknown_fields = true });
+    const interval_ms: u64 = @as(u64, @intCast(std.fmt.parseInt(u32, dc.interval, 10) catch 5)) * std.time.ms_per_s;
+
+    std.debug.print(
+        "\nTo authorize Lightpanda with your ChatGPT subscription:\n  1. Open {s}\n  2. Enter the code: {s}\n\nWaiting for authorization...\n",
+        .{ verify_url, dc.user_code },
+    );
+
+    const poll_body = try std.fmt.allocPrint(a, "{{\"device_auth_id\":\"{s}\",\"user_code\":\"{s}\"}}", .{ dc.device_auth_id, dc.user_code });
```
karlseguin: Gluing our hard-coded client_id into a json string is cool. While I get that, in this case,  doing it with the values which come from a 3rd party API call is safe it's still a little iffy.

src/agent/Agent.zig
```diff
@@ -1868,18 +1919,22 @@ pub fn listModels(allocator: std.mem.Allocator, opts: Config.Agent) !void {
...
-        return err;
-    };
+    // A subscription provider can't list models via the provider API; use the
+    // free models.dev catalog.
+    const ids: []const []const u8 = if (auth.descriptorFor(resolved.credential.provider)) |desc|
+        models_dev.modelIds(arena.allocator(), desc.models_dev_id, auth.dataDir(arena.allocator()))
+    else
+        zenai.provider.listChatModelIds(lp.io, allocator, arena.allocator(), resolved.credential.provider, resolved.credential.keySlice(), .{ .base_url = opts.base_url, .environ = lp.environ() }) catch |err| {
+            if (resolved.credential.provider == .vertex and !settings.vertexProjectMode()) {
+                std.debug.print("Vertex express mode cannot list models (the endpoint requires OAuth); set GOOGLE_CLOUD_PROJECT for project mode.\n", .{});
```
karlseguin, line 1964: Is there a pattern between what uses std.debug.print and what uses `self.terminal` ?

src/agent/auth/models_dev.zig
```diff
@@ -0,0 +1,124 @@
...
+
+fn writeCache(app_dir: []const u8, provider_id: []const u8, ids: []const []const u8) !void {
+    var arena: std.heap.ArenaAllocator = .init(std.heap.page_allocator);
+    defer arena.deinit();
+    const a = arena.allocator();
+    const path = try cachePath(a, app_dir, provider_id);
+
+    var buf: std.Io.Writer.Allocating = .init(a);
+    try std.json.Stringify.value(Cache{ .fetched_ms = auth.nowMs(), .ids = ids }, .{}, &buf.writer);
+    try std.Io.Dir.cwd().writeFile(lp.io, .{ .sub_path = path, .data = buf.written() });
```
karlseguin: For anything that writes, maybe it's worth doing a write-to-temp + rename dance to avoid a partial write that (a) deletes valid entries and (b) leaves an invalid file.

src/agent/auth/models_dev.zig
```diff
@@ -0,0 +1,124 @@
...
+}
+
+/// Model-id keys of `catalog[provider_id].models`, filtered to tool-call-capable
+/// models (the agent needs tools; this also drops embedding/image entries).
+/// `ignore_unknown_fields` skips the rest of the metadata, so no Value tree is
+/// built for the multi-MB payload. The parse runs in a scoped arena; the ids
+/// are duped into `arena`.
+fn parseProviderModels(arena: std.mem.Allocator, catalog: []const u8, provider_id: []const u8) ![]const []const u8 {
+    var parse_arena: std.heap.ArenaAllocator = .init(arena);
```
karlseguin: I'm not sure nesting arenas is ever useful. Should assume anything `parse_arena` allocates won't be freed on deinit. I believe here its just a bit of `parseFromSliceLeaky` overheard, which is probably fine..but even then, I'd just use `arena` directly to be honest about the memory accumulation of this function.

src/agent/auth/auth.zig
```diff
@@ -0,0 +1,275 @@
...
+    return .{ .allocator = allocator, .descriptor = desc, .tokens = tokens };
+}
+
+/// Is a usable (present, not hard-expired) stored token available for `provider`?
+/// Lets the picker offer the subscription without an API-key env var.
+pub fn subscriptionAvailable(provider: Config.AiProvider) bool {
+    const desc = descriptorFor(provider) orelse return false;
+    const tokens = (storeLoad(std.heap.page_allocator, desc.storeKey()) catch return false) orelse return false;
+    defer tokens.deinit(std.heap.page_allocator);
+    return tokens.expires_at_ms > nowMs();
```
karlseguin: So if it's expired but has a refresh_token, it isn't considered available?

src/agent/auth/codex.zig
```diff
@@ -0,0 +1,269 @@
...
+        try lp.URL.percentEncodeSegment(arena, code_verifier, .component),
+    });
+}
+
+// --- Device-code flow ---
+
+const DeviceCode = struct {
+    device_auth_id: []const u8,
+    user_code: []const u8,
+    interval: []const u8 = "5",
```
karlseguin: I tested the openai endpoint, and it _does_ return a string, but the spec does say a number. I *think* if you use an integer type here, std.json will parse it for you (but, as-is, I don't think it would convert a number to a string).

src/agent/auth/codex.zig
```diff
@@ -0,0 +1,269 @@
...
+            const slice_ms = @min(remaining_ms, cancel_slice_ms);
+            lp.io.sleep(.fromMilliseconds(@intCast(slice_ms)), .awake) catch {};
+            remaining_ms -= slice_ms;
+        }
+        const res = try post(a, interrupt, device_token_url, "application/json", poll_body);
+        switch (res.status) {
+            .ok => break try std.json.parseFromSliceLeaky(DeviceToken, a, res.body, .{ .ignore_unknown_fields = true }),
+            // Still pending — the user hasn't finished authorizing.
+            .forbidden, .not_found => continue,
+            else => return error.DeviceAuthFailed,
```
karlseguin: Might be interesting to log these failures. Claude is worried about ignoring 429 (slow down) and opening a new connection every N seconds (rather than re-using the client). I think that can be put off until it's proven necessary, but without logging, any failure case is pretty opaque.

src/agent/auth/models_dev.zig
```diff
@@ -0,0 +1,124 @@
...
+
+/// Chat-model ids for `provider_id` from models.dev, allocated in `arena`. Reads
+/// a fresh on-disk cache when present, else fetches and refreshes it. Returns an
+/// empty slice on any failure (offline, parse error, no cache) so callers
+/// degrade to accepting a typed-in model name.
+pub fn modelIds(arena: std.mem.Allocator, provider_id: []const u8, app_dir: ?[]const u8) []const []const u8 {
+    if (app_dir) |dir| {
+        if (readCache(arena, dir, provider_id)) |ids| return ids;
+    }
+    const catalog = fetch(arena) catch return &.{};
```
karlseguin: If this fails, but readCache has expired results. Should it use the expired results?

src/agent/Agent.zig
arrufat: yes, the `self.terminal` is for user-facing messages when there's an Agent terminal (spinner, coloring, etc).

The raw `std.debug.print` is used when there's no terminal (`listModels`, etc).
After looking deeply into this, I realized I missed the model/effort banner, which should be in terminal. Let me fix that.

arrufat:
Thank you so much for the detailed code review!

arrufat:
I've added a way to logout/relogin with

```
❯ /provider codex

Already logged in with your ChatGPT subscription. Pick:
  > keep — use the stored login (default)
    login — run the login again (e.g. to switch accounts)
    logout — forget the stored login
Use Up/Down then Enter. Esc cancels.
```

karlseguin (approved):

PR #3090, opened 2026-07-30
Rework `Connection.upgrade` with `header_parser`

nikneym:


src/cdp/Connection.zig
```diff
@@ -418,59 +419,96 @@ fn pushCdp(self: *Connection, bytes: []const u8) !bool {
...
+            required_headers.sec_websocket_key = true;
         }
-
-        const next = index + 2;
-        buf = buf[next..];
     }
 
-    if (required_headers != 15) {
+    // Check if we've received all related headers.
+    const satisfied = @as(u8, @bitCast(required_headers)) == @as(u8, @bitCast(RequiredHeaders{
```
karlseguin: Whatever readability was gained by having a struct was lost here.  Dropping the bitcast and checking if all fields are true would be better.

src/network/header_parser.zig
```diff
@@ -231,6 +231,66 @@ pub const Header = struct {
...
+        return error.Incomplete;
+    }
+    // Make sure we got HTTP/1.1.
+    if (cursor.asInteger(u64) != @as(u64, @bitCast(@as([]const u8, "HTTP/1.1")[0..8].*))) {
+        return error.Invalid;
+    }
+    cursor.advance(8);
+
+    // Trailing (CR)LF.
+    switch (cursor.char()) {
```
karlseguin, line 427: This switch appears more or less 3 times now (2 added in this PR). It would be better if a header parser wouldn't require callers to do this.

src/cdp/Connection.zig
```diff
@@ -418,59 +419,96 @@ fn pushCdp(self: *Connection, bytes: []const u8) !bool {
 }
 
 pub fn upgrade(self: *Connection, request: []u8) !void {
-    // our caller already confirmed that we have a trailing \r\n\r\n
-    const request_line_end = std.mem.indexOfScalar(u8, request, '\r') orelse unreachable;
-    const request_line = request[0..request_line_end];
-
-    if (!std.ascii.endsWithIgnoreCase(request_line, "http/1.1")) {
+    var cursor = header_parser.Cursor{
```
karlseguin: `Cursor.init(request);`

FormData has the same.

src/network/header_parser.zig
```diff
@@ -231,6 +231,66 @@ pub const Header = struct {
     }
 };
 
+/// Validates WebSocket initialization requests.
+/// Currently does not validate paths.
+pub fn validateWebSocketRequestLine(cursor: *Cursor) !void {
+    // GET / HTTP/1.1\n
+    const min_request_len = 0xf;
```
karlseguin:
```suggestion
    const min_request_len = 15;
```

src/network/header_parser.zig
```diff
@@ -231,6 +231,66 @@ pub const Header = struct {
...
+    const path_start = cursor.current();
+    // Find the first space.
+    while (cursor.end - cursor.current() > 0 and cursor.char() != ' ') : (cursor.advance(1)) {}
+    const path_end = cursor.current();
+    // 0 length path.
+    if (path_start == path_end) {
+        return error.Invalid;
+    }
+    const path = path_start[0 .. path_end - path_start];
+    _ = path;
```
karlseguin: If it isn't needed..why slice it?

src/cdp/Connection.zig
```diff
@@ -418,59 +419,96 @@ fn pushCdp(self: *Connection, bytes: []const u8) !bool {
...
+        }
 
-        const name = std.mem.trim(u8, toLower(buf[0..separator]), &std.ascii.whitespace);
-        const value = std.mem.trim(u8, buf[(separator + 1)..index], &std.ascii.whitespace);
+        // A malformed header maps to a 400 in processHttpRequest.
+        header.parse(&cursor) catch {
+            return error.InvalidRequest;
+        };
+        const key = header.key;
+        const value = header.value;
```
karlseguin, line 460: from what I can tell, `header.value` isn't trimmed, so this won't handle cases that it should.

src/network/header_parser.zig
```diff
@@ -231,6 +231,66 @@ pub const Header = struct {
...
+
+    // Skip past the delimiting space(s); the scan above guarantees we're on
+    // a space or at the end, and recipients may parse on whitespace
+    // boundaries (RFC 9112 §3).
+    while (cursor.end - cursor.current() > 0 and cursor.char() == ' ') : (cursor.advance(1)) {}
```
karlseguin: Is this whole line just a way to skip spaces? There's already a `cursor.skipSpaces() (which seems more spec compliant than this).

Is the first condition just a way to check the end? There's already a `reachedEnd()`

karlseguin (commented):
Seems to make things more complicated. Not sure why it isn't something like

```
var header = try HeaderParser.init(request);
if (header.method != "GET") {....}
while (try header.next()) |kv| {
    kv.name, kv.value
}
```

Don't see why a caller needs to know about string.ptr, or trimming values, or line termination.

src/cdp/Connection.zig
nikneym: It's actually trimmed from the beginning (after `:`), which should be right for the case.

karlseguin: I ran it to double check and your own test confirms it: the expected value should be  "value" not "value \t".

```zig
try testing.expectString("value \t", header.value);
```

PR #3096, opened 2026-07-31
c-api: expose liblightpanda.so and C header for embedding

arrufat:
A C ABI (include/lightpanda.h + src/c_api.zig) over the browser tool surface: lp_init/lp_shutdown, lp_fetch, sessions with lp_call/pump/ cancel, lp_tools_json. Built as a shared library only; the version script keeps everything but lp_* internal so the bundled OpenSSL/curl/ sqlite cannot collide with a host's own.

Embedders get a hidden 'embed' Config mode (not parseable from the CLI) with telemetry defaulting off; crash reports honor the same opt-out. ToolSession in lightpanda.zig owns the browser/session/ registry lifecycle the C API drives.

Build: C deps are always PIC (like boringssl's force_pic) and the pinned zig-v8-fork always builds V8 library-safe, so 'zig build lib' needs no flags. It only refuses -Dprebuilt_v8_path: today's published archives are exe-only (local-exec TLS, malloc shim); the guard goes away once a fork release ships library-safe archives. make lib-shared / lib-test / lib-shared-example drive it.


src/c_api.zig
```diff
@@ -0,0 +1,648 @@
...
+var app_state: AppState = .uninitialized;
+
+/// Initialize the library and return the process-wide browser handle.
+/// Callable exactly once per process; after `lp_shutdown` it fails with
+/// `misuse` forever (V8 cannot be re-initialized).
+pub export fn lp_init(opts_: ?*const InitOpts, out_: ?**BrowserHandle) Status {
+    const out = out_ orelse return .misuse;
+    if (app_state != .uninitialized) return .misuse;
+
+    const handle = createBrowser(opts_) catch |err| return errStatus(err);
```
karlseguin, line 149: If this fails, `app_state` is still `.uninitialized` but v8 might have already been internally poisoned. The `catch` block should change it.

src/c_api.zig
```diff
@@ -0,0 +1,648 @@
...
+
+// Mirrored in include/lightpanda.h (lp_options), which documents the
+// sentinel values. Zero-initialized means defaults everywhere.
+const InitOpts = extern struct {
+    user_agent: ?[*:0]const u8,
+    http_proxy: ?[*:0]const u8,
+    http_cache_dir: ?[*:0]const u8,
+    http_timeout_ms: u32,
+    watchdog_ms: i32,
+    enable_telemetry: bool,
```
karlseguin: We need to check with @francisbouvier if default-off (no telemetry) is ok for embedded use.

src/c_api.zig
```diff
@@ -0,0 +1,648 @@
...
+    // ToolSession.init left the isolate entered; park it so sessions can
+    // share the thread.
+    entry.ts.exitIsolate();
+    return entry;
+}
+
+/// Close a session created by `lp_session_new`, freeing its isolate, page
+/// and memory. The handle is invalid afterwards.
+pub export fn lp_session_close(entry_: ?*SessionHandle) void {
+    const entry = entry_ orelse return;
```
karlseguin, line 377: Should check `if (app_state != .live) return .misuse;` ?

src/c_api.zig
```diff
@@ -0,0 +1,648 @@
...
+    const s = aw.written();
+    return s[0 .. s.len - 1 :0];
+}
+
+/// Pump the session's background work (timers, in-flight fetches, resumed
+/// JS) once. Returns the number of milliseconds the caller may sleep before
+/// pumping again. Only needed when idling between calls; every tool call
+/// already waits for its own completion.
+pub export fn lp_session_pump(entry_: ?*SessionHandle) u32 {
+    const entry = entry_ orelse return 0;
```
karlseguin, line 450: Should check `if (app_state != .live) return .misuse;` ?

src/c_api.zig
```diff
@@ -0,0 +1,648 @@
...
+/// blocking waits: once it returns true, the in-flight call fails with
+/// `LP_ERR_CANCELLED`. The probe is invoked on the session's thread, but it
+/// may read state set by another thread (e.g. an atomic flag flipped by a
+/// signal handler).
+pub export fn lp_session_set_cancel_hook(
+    entry_: ?*SessionHandle,
+    cb: ?*const fn (?*anyopaque) callconv(.c) bool,
+    ctx: ?*anyopaque,
+) void {
+    const entry = entry_ orelse return;
```
karlseguin, line 467: Should check `if (app_state != .live) return .misuse;` ?

karlseguin (commented):
I'm building on MacOS now. Will update if there's an issue.

Three things (+ the telemetry-off decision that needs to be OKd)
1 - Consuming a C API from a non-C language is often tedious because of the null-terminated strings. I always prefer APIs that support both the ability to pass a null-terminated string OR a ptr + len
2 - All the error codes are opaque. You have an arena, so an lp_last_error that takes a SessionHandle woudl be nice
3 - Would be nice if the CI could build this, build an example, and verify its output.

src/c_api.zig
arrufat: Right, I made it behave like the rest of the browser

arrufat:
@karlseguin regarding the example, I will add it once we provide pre-built v8 binaries with the latest changes.

karlseguin (approved):

arrufat:
Thank you, maybe we should wait for @krichprollsch in case he has some ideas on what the C API should look like.
Ah, and it needs to be coordinated with a new release of v8, and the telemetry server, too.

arrufat:
Note from the ToolSession follow-up (#3312): a reuse pass flagged that `writeToolsJson` here and `Tool.jsonStringify` in mcp/protocol.zig duplicate the tools/list serializer. What they actually share is the three raw-JSON lines for `inputSchema`; the rest differs on purpose (MCP emits `title` and the annotations, plus the save/session_* extras). So the question is whether `lp_tools_json` should carry `title` and annotations too. If yes, one shared struct falls out naturally; if no, the two writers should stay separate. Leaving that to this PR rather than a dedup change.

PR #3116, opened 2026-08-03
Initial implementation for easy list parsing and AdBlocker

nikneym:
This PR adds an easy list parser and a small ad blocker that works over trie data structure to block and allow domains.


karlseguin:
Missing help.zon entry?

src/network/HttpClient.zig
```diff
@@ -35,6 +35,7 @@ const Cache = @import("cache/Cache.zig");
 const RobotsGate = @import("RobotsGate.zig");
 const UrlBlocklist = @import("UrlBlocklist.zig");
 pub const BlockPattern = UrlBlocklist.Pattern;
+const AdBlocker = @import("../adblock/AdBlocker.zig");
```
karlseguin: Can we keep imports grouped together? Also, maybe src/adblock -> src/network/adblock/  is better since it's only used from network/* files.

src/network/HttpClient.zig
```diff
@@ -208,6 +211,20 @@ pub fn init(self: *Client, allocator: Allocator, network: *Network, cdp: ?*CDP)
     }
     errdefer if (url_blocklist) |*blocklist| blocklist.deinit();
 
+    var adblocker: ?AdBlocker = null;
+    errdefer if (adblocker) |*blocker| blocker.deinit();
+    if (network.config.adblockLists()) |initial_paths| {
```
karlseguin: If this list is readonly, it would be wonderful to parse it once and share it for all HttpClients (aka, Browsers). Less memory, faster, better UX (you can return an error on startup).

src/network/HttpClient.zig
```diff
@@ -387,6 +406,38 @@ fn isUrlBlocked(self: *const Client, url: []const u8, internal: bool) bool {
...
+        path,
+        allocator,
+        .limited(64 * 1024 * 1024),
+        .of(u8),
+        null,
+    );
+    defer allocator.free(text);
+
+    var reader: std.Io.Reader = .fixed(text);
+    const stats = try blocker.parse(&reader);
```
karlseguin: I presume that if blocker.parse speaks `*std.Io.Reader` then we can create a Reader from the File, rather than reading the whole thing into memory?

src/adblock/adblock.zig
```diff
@@ -0,0 +1,302 @@
...
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+const std = @import("std");
+
+pub const AdBlocker = @import("AdBlocker.zig");
+pub const NetworkFilter = @import("NetworkFilter.zig");
+pub const domain = @import("domain.zig");
+
+pub const LineClass = enum(u3) {
```
karlseguin: All four of these are only used within this file. No reason to be `pub`. Can you double check every other public declaration ?

karlseguin: I read more...The naming of this file, and that everything was public to me said "this is where adblocking constants, and shared structures are stored. But really, I think this is the `Parser` (Parser.zig) and there's 2 public functions: `init` and `next`.

src/adblock/adblock.zig
```diff
@@ -0,0 +1,302 @@
...
+                    stripped = stripped[3..];
+                }
+            }
+            const line = std.mem.trim(u8, stripped, &std.ascii.whitespace);
+            self.stats.lines += 1;
+
+            switch (LineClass.fromLine(line)) {
+                .empty => {},
+                .comment => {
+                    self.stats.comments += 1;
```
karlseguin: Is this being tracked for a reason?

src/adblock/adblock.zig
```diff
@@ -0,0 +1,302 @@
...
+    invalid: usize = 0,
+    /// Hosts-file noise ("127.0.0.1 localhost").
+    ignored: usize = 0,
+};
+
+pub const Parser = struct {
+    reader: *std.Io.Reader,
+    stats: ParseStats = .{},
+    title: ?[]const u8 = null,
+    expires: ?[]const u8 = null,
```
karlseguin: As far as I can tell, none of these meta fields are used. Do we plan on using them?

src/adblock/adblock.zig
```diff
@@ -0,0 +1,302 @@
...
+            &self.homepage
+        else if (std.ascii.eqlIgnoreCase(key, "version"))
+            &self.version
+        else
+            return;
+        if (slot.* == null) slot.* = try arena.dupe(u8, value);
+    }
+};
+
+pub const List = struct {
```
karlseguin: This is only used in tests, for this file? I'd prefer to have it under the test (or at least under the testing declaration...or a comment).

src/adblock/NetworkFilter.zig
```diff
@@ -0,0 +1,979 @@
...
+    out.pattern = try arena.dupe(u8, self.pattern);
+    out.hostname = try arena.dupe(u8, self.hostname);
+    out.domains = try self.domains.dupe(arena);
+    return out;
+}
+
+/// uBO allows a trailing "  # comment" on lines containing whitespace.
+fn stripInlineComment(line: []const u8) []const u8 {
+    var i: usize = 1;
+    while (i < line.len) : (i += 1) {
```
karlseguin: Is this an `std.mem.indexOfScalar` ?

src/adblock/adblock.zig
```diff
@@ -0,0 +1,302 @@
...
+            .title = parser.title,
+            .expires = parser.expires,
+            .homepage = parser.homepage,
+            .version = parser.version,
+            .stats = parser.stats,
+        };
+    }
+};
+
+const testing = std.testing;
```
karlseguin: most of the code uses `testing.zig`, I don't mind if you want to use std.testing...just want to make sure it's intentional.

src/adblock/CosmeticFilter.zig
```diff
@@ -0,0 +1,134 @@
+// Copyright (C) 2023-2026 Lightpanda (Selecy SAS)
+//
+// Francis Bouvier <francis@lightpanda.io>
```
karlseguin: This file is unused

src/adblock/AdBlocker.zig
```diff
@@ -0,0 +1,223 @@
...
+const adblock = @import("adblock.zig");
+const HostnameTrie = @import("HostnameTrie.zig");
+const NetworkFilter = @import("NetworkFilter.zig");
+
+const AdBlocker = @This();
+
+arena: std.heap.ArenaAllocator,
+/// Network filters the tries cannot express: patterns, type/party/domain
+/// constraints, badfilter directives.
+filters: std.ArrayList(NetworkFilter),
```
karlseguin: this is never used, but it dupes a [potentially] huge amount of data

karlseguin (changes requested):

src/adblock/adblock.zig
nikneym: It was essentially for testing and metrics, which, release builds don't require. Removing it as a result.

src/adblock/adblock.zig
nikneym: They were added for future parsing requirements, which we can defer for now.

src/network/Network.zig
```diff
@@ -311,6 +322,41 @@ pub fn deinit(self: *Network) void {
     globalDeinit();
 }
 
+/// Read buffer for filter lists, and therefore the longest line we can see.
+/// Real-world rules are well under 1KB; the parser skips anything longer.
+const ADBLOCK_LINE_MAX = 8 * 1024;
```
karlseguin: Look at the sibling WebAuth code which uses a `WebBotAuth.fromConfig`. None of this is stuff Network should know about.

src/network/HttpClient.zig
```diff
@@ -224,6 +230,7 @@ pub fn init(self: *Client, allocator: Allocator, network: *Network, cdp: ?*CDP)
         .obey_robots = network.config.obeyRobots(),
         .robots = .{ .allocator = allocator, .network = network },
         .url_blocklist = url_blocklist,
+        .adblocker = if (network.adblocker) |*blocker| blocker else null,
```
karlseguin: only used in `isHostAdblocked`, could access it through network and remove the field, e.g.:

```zig
const blocker = self.adblocker orelse return false;
```

src/network/HttpClient.zig
```diff
@@ -185,6 +187,10 @@ obey_robots: bool,
 
 robots: RobotsGate,
 url_blocklist: ?UrlBlocklist,
+/// Hostname dictionaries built from `--adblock-lists`. Owned by Network,
```
karlseguin: Comment is mostly a duplicate of the field in network, more places for it to become stale. See next comment, but removing this field, solves it.

src/network/HttpClient.zig
```diff
@@ -846,6 +865,10 @@ fn pipeline(self: *Client, transfer: *Transfer, from: SubmitFrom) !void {
                 log.warn(.http, "blocked url", .{ .url = transfer.req.url });
                 return transfer.failAsync(error.UrlBlocked);
             }
+            if (self.isHostAdblocked(transfer.req.url, transfer.req.internal)) {
```
karlseguin: I would expect that whenever we call `self.isUrlBlocked` we would also call `self.isHostAdblocked`, but that isn't the case. Is that right?

If they should both always be called in the same place, maybe isUrlBlocked can just check both lists?

src/network/adblock/AdBlocker.zig
```diff
@@ -0,0 +1,202 @@
...
+    const scratch = scratch_instance.allocator();
+
+    var parser: Parser = .init(reader);
+    while (try parser.next(scratch)) |filter| {
+        // Nothing survives the iteration but the trie entry, so the
+        // scratch arena is recycled rather than grown per filter.
+        defer _ = scratch_instance.reset(.retain_capacity);
+
+        // Filters the tries cannot express (patterns, type/party/domain
+        // constraints, badfilter directives) are dropped: deciding those
```
karlseguin: Because we don't support badfilter,  we'll block URLs that we shouldn't?  e.g. uBO unbreak.txt has 143 badFilters.

I asked claude to see if there were other cases where we might block a URL that we shouldn't and it said any @@ exception with a domain doesn't get added and gave this specific example from easylist:

  easylist.txt:17462  ||adsafeprotected.com^
  easylist.txt:76776  @@||adsafeprotected.com/iasPET.$script,domain=independent.co.uk|reuters.com|wjs.com

which should not block `adsafeprotected.com` on reuters.com, but would.

src/network/HttpClient.zig
karlseguin: PLUS....`error.UrlBlocked` is handled specifically by CDP for the fail reason, so these AdblockBlocked behave differently (and more wrongly) for CDP clients.

And the logging between the two isn't consistent. I prefer your info (or even debug level).

src/network/Network.zig
```diff
@@ -311,6 +322,41 @@ pub fn deinit(self: *Network) void {
...
+
+/// Streams `path` into `blocker`. The list is consumed line by line, so a
+/// 100MB list never needs 100MB of memory.
+fn loadAdblockList(blocker: *AdBlocker, path: []const u8, buf: []u8) !void {
+    const file = try std.Io.Dir.cwd().openFile(lp.io, path, .{});
+    defer file.close(lp.io);
+
+    var file_reader = file.reader(lp.io, buf);
+    try blocker.parse(&file_reader.interface);
+    log.info(.app, "adblock list loaded", .{ .path = path });
```
karlseguin: The first thing I tried was a file with a single rule: `||lightpanda.io` and when I ran, it said:
```
INFO  app : adblock list loaded . . . . . . . . . . . . . . .  [+0ms]
    path = list.txt
```

But it didn't actually block. I get why we drop non-anchored rules, and I checked a couple list and non-anchored rules are a very rare...and NOT blocking a site we should block is a lot better than blocking a site we shouldn't. But it might be surprising and hard to debug what is and isn't being blocked for people. a loaded = # and skip = # log line might help indicate that "hey, not all the rules were loaded, heads up"

src/network/adblock/NetworkFilter.zig
```diff
@@ -0,0 +1,970 @@
...
+const domain = @import("domain.zig");
+
+const NetworkFilter = @This();
+
+kind: PatternKind,
+/// Normalized (lowercased) match pattern. Empty for pure-hostname filters.
+/// For hostname-anchored filters with a path part, this is the remainder
+/// after `hostname` (starting with '/', '^' or '*'). For regex filters, the
+/// body between the slashes, stored raw.
+pattern: []const u8 = "",
```
karlseguin: From what I can tell, nothing ever uses this.

src/network/adblock/domain.zig
```diff
@@ -0,0 +1,163 @@
...
+const std = @import("std");
+
+pub const Domain = struct {
+    /// Lowercase hostname. For entities, stored without the trailing ".*".
+    value: []const u8,
+    entity: bool = false,
+};
+
+pub const List = struct {
+    included: []const Domain = &.{},
```
karlseguin: From what I can tell, nothing ever uses included or excluded, other than isEmpty() (which is used). But if we only need a boolean to track emptiness, then that's all we should have.

karlseguin:
@nikneym What state is it in with respect to over and under blocking?

nikneym:
Its biased toward under-blocking; we drop some rules (some require request engine and most are cosmetic).

Tested against [EasyList + EasyPrivacy](https://github.com/easylist/easylist); 87,061 blocking hostnames, 1,362 suppressions, 4 allows applied. 20,738 network rules dropped.

|                                     | EasyList       | EasyPrivacy    |
| ----------------------------------- | -------------- | -------------- |
| network rule lines                  | 53,324         | 55,841         |
| enforced as hostname blocks         | 44,130 (82.8%) | 42,931 (76.9%) |
| suppression entries                 | 541            | 821            |
| whole-hostname exceptions (allowed) | 0              | 4              |
| network rules dropped               | 8,653 (16.2%)  | 12,085 (21.6%) |

karlseguin (approved):
Looks like, in the current state, we don't block things we shouldn't, but still allow some domains we should be blocking. This seems like a safer first step.

PR #3122, opened 2026-08-03
Fix Playwright auxiliary CDP session interception

staylor:
## Motivation

Browser automation clients sometimes create an auxiliary CDP session for a page so they can use low-level domains such as `Fetch` without interfering with the session Playwright already uses to manage that page. For this to work, the auxiliary session must have its own identity, receive the events for domains enabled through it, and observe the same request-scoped override semantics as Chromium.

Without those guarantees, adding request interception through `browserContext.newCDPSession(page)` can destabilize the existing page session or change the headers sent to a redirect destination. This makes an otherwise supported Playwright integration fail in ways that are difficult for callers to diagnose.

## Problem

Playwright implements `browserContext.newCDPSession(page)` by attaching a browser-target session and then calling `Target.attachToTarget` through it. Lightpanda reused the page's auto-attached session ID instead of creating a distinct child session. Playwright then replaced its existing page-session mapping, and an in-flight protocol response triggered an internal assertion.

There were two related session/Fetch mismatches:

- `Fetch.enable` state and `Fetch.requestPaused` / `Fetch.authRequired` events were tied to the primary page session rather than the session that enabled Fetch.
- `Fetch.continueRequest` header overrides survived redirects. Chromium limits those overrides to the current request hop and pauses the redirected request separately, allowing the client to decide its headers independently.

## Change

- Create stable, unique auxiliary target sessions and route attach/detach events through their parent session.
- Treat auxiliary sessions as valid CDP command sessions and clean them up with their target.
- Route `Fetch.requestPaused` and `Fetch.authRequired` to the session that enabled Fetch.
- Restore `Fetch.continueRequest` header overrides after one network hop and pause each redirected request independently.

## Verification

- `make test`: 1128/1128 passed.
- `zig fmt --check ./*.zig ./**/*.zig`.
- Added a committed Playwright integration test and dedicated `playwright-cdp-session` CI job. The test starts two local origins and verifies that:
  - `browserContext.newCDPSession(page)` completes without replacing the page session.
  - `Fetch.requestPaused` fires for both the initial request and redirected request.
  - A header override applied to the initial request is absent from the redirected request.
  - Navigation resolves with HTTP 200.
- The committed integration test passes locally against the patched Lightpanda binary.


karlseguin:
Hi @staylor this will get merged tomorrow along with https://github.com/lightpanda-io/browser/pull/3142  . Your integration tests was also split out into https://github.com/lightpanda-io/demo/pull/221 and used as the basis for an equivalent puppeteer and chromedp test.

Issue #3138, opened 2026-08-04
agent/mcp goto always waits for the load event, no waitUntil option


arrufat:
`goto` in the agent/MCP/PandaScript tool layer always waits with `until = .load` (`default_nav_wait` in `src/browser/tools.zig:1934`). There is no way to opt down. `fetch` mode already exposes `--wait-until` and `--wait-selector`, and `Runner.waitForFrame` accepts any `Config.WaitUntil`, so this is only missing at the tool layer.

Since #3085 this is expensive on ad-heavy pages: dynamically injected ad scripts now load (correctly), they chain (each one's onload injects the next), and every link injected before load delays the load event. The page content is usually ready much earlier.

Measured on an apnews article (ReleaseFast e14353398, `--disable-subframes`, fresh cache; extracted markdown is equivalent in both modes):

| wait | wall time (3 runs) |
| :--- | ---: |
| `load` | 3.1-5.3 s |
| `--wait-selector ".RichTextStoryBody p"` | 1.3-1.9 s |

Proposal:

1. Add an optional `waitUntil` param to the goto tool (`load` \| `domcontentloaded` \| `networkalmostidle` \| `networkidle`). Default stays `load`.
2. Have the agent and `/save` record `goto {waitUntil: "domcontentloaded"}` + `waitForSelector` when the next recorded action after a navigation is an extract or waitForSelector anyway.

Puppeteer/Playwright expose the same option on `page.goto`, so recorded scripts stay comparable across stacks.

Benchmark context: `v7`/`v7py` datasets in [agent-benchmarks/pandascript-vs-cdp](https://github.com/lightpanda-io/agent-benchmarks/tree/main/pandascript-vs-cdp). Related (navigation in the same tool layer, different concern): #2176.

I can implement this.

arrufat:
Sibling issue for the subframes-default half: #3139

arrufat:
Implemented in #3140: `goto` takes `waitUntil` (`load` | `domcontentloaded` | `networkalmostidle` | `networkidle` | `done`, default `load`) from the tool layer, REPL and scripts, and the recorder rewrites `goto` to `domcontentloaded` when the next recorded step is `waitForSelector` / `waitForScript` / `waitForState`.

PR #3140, opened 2026-08-04
goto: optional waitUntil + dcl-gated selector waits

arrufat:
Implements #3138.

- `goto` accepts an optional `waitUntil` (`load` | `domcontentloaded` | `networkalmostidle` | `networkidle` | `done`, default `load`). Works from the tool layer (agent/MCP), PandaScript object form (`await page.goto({ url, waitUntil })`), and the REPL (`/goto <url> waitUntil=domcontentloaded`). The schema enum comes from `Config.tagJsonArray` and the description steers away from `done`.
- `waitForSelector`/`waitForScript` gate their poll on `domcontentloaded` instead of `load` (polling needs a parsed document, nothing more). Without this the pre-wait re-waited the exact tail a dcl navigation skips.
- Wait-budget guard from review: a wait entered before `load` defaults to nav-timeout + 5s instead of 5s, so a selector that only appears post-load keeps the wall time it had when goto waited for `load` itself. Applies to the tool layer and to CDP's `LP.waitForSelector` (same underlying action). Explicit timeouts unchanged.
- The recorder rewinds the just-emitted goto line to the `domcontentloaded` variant when the next recorded call waits for readiness on its own (`Tool.waitsForReadiness`, which also covers `waitForScript`). Explicit `waitUntil` is never rewritten; goto followed by a bare `extract` keeps `load` (dcl could miss late-rendered content the session actually saw).

Measured on an ad-heavy apnews article (ReleaseFast at the current head, `--disable-subframes`, fresh cache, identical extracted output):

| variant | wall |
| :--- | ---: |
| `goto` (load) + waitForSelector | 5.2-7.0 s |
| `goto {waitUntil: "domcontentloaded"}` + waitForSelector | 0.9-1.3 s |

1129/1129 tests, `zig fmt --check` clean, main merged in.

### DIFF
--- a/src/browser/actions.zig
+++ b/src/browser/actions.zig
@@ -276,7 +276,10 @@ fn remainingMs(timeout_ms: u32, timer: std.Io.Timestamp) u32 {
 pub fn waitForSelector(selector: [:0]const u8, timeout_ms: u32, frame_id: u32, session: *Session) !*DOMNode {
     const timer: std.Io.Timestamp = .now(lp.io, .boot);
     var runner = session.runner(.{});
-    try runner.waitForFrame(frame_id, timeout_ms, .{ .until = .load });
+    // Only parsing has to be done before polling — gating on `.load` would
+    // re-wait for the late-script tail a `waitUntil: domcontentloaded`
+    // navigation deliberately skipped (#3138).
+    try runner.waitForFrame(frame_id, timeout_ms, .{ .until = .domcontentloaded });
 
     const el = try runner.waitForSelector(frame_id, selector, remainingMs(timeout_ms, timer));
     return el.asNode();
@@ -285,7 +288,7 @@ pub fn waitForSelector(selector: [:0]const u8, timeout_ms: u32, frame_id: u32, s
 pub fn waitForScript(script: [:0]const u8, timeout_ms: u32, frame_id: u32, session: *Session) !void {
     const timer: std.Io.Timestamp = .now(lp.io, .boot);
     var runner = session.runner(.{});
-    try runner.waitForFrame(frame_id, timeout_ms, .{ .until = .load });
+    try runner.waitForFrame(frame_id, timeout_ms, .{ .until = .domcontentloaded });
 
     return runner.waitForScript(frame_id, script, remainingMs(timeout_ms, timer));
 }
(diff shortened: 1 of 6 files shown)


src/browser/tools.zig
```diff
@@ -329,7 +339,10 @@ pub const Tool = enum {
                     \\  "type": "object",
                     \\  "properties": {
                     \\    "url": { "type": "string", "description": "The URL to navigate to, must be a valid URL." },
-                    \\    "timeout": { "type": "integer", "description": "Optional timeout in milliseconds. Defaults to 10000." }
+                    \\    "timeout": { "type": "integer", "description": "Optional timeout in milliseconds. Defaults to 10000." },
+                    \\    "waitUntil": { "type": "string", "enum":
+                ++ lp.Config.tagJsonArray(lp.Config.WaitUntil) ++
```
karlseguin, line 344: This exposes 'done', the slowest possible option. Fine to expose it, just worried people will see "done", and pick it because it sounds like what they want.

src/browser/actions.zig
```diff
@@ -276,7 +276,10 @@ fn remainingMs(timeout_ms: u32, timer: std.Io.Timestamp) u32 {
 pub fn waitForSelector(selector: [:0]const u8, timeout_ms: u32, frame_id: u32, session: *Session) !*DOMNode {
     const timer: std.Io.Timestamp = .now(lp.io, .boot);
     var runner = session.runner(.{});
-    try runner.waitForFrame(frame_id, timeout_ms, .{ .until = .load });
+    // Polling needs a parsed document, nothing more. Gating on `.load` would
+    // re-wait the late-script tail a `waitUntil: domcontentloaded` navigation
+    // deliberately skipped.
+    try runner.waitForFrame(frame_id, timeout_ms, .{ .until = .domcontentloaded });
```
karlseguin, line 282: I don't know how to handle it (or even if it has to be handled), but previously, you'd have called `waitForSelector` later in the page's lifecycle, e.g. it would be at 'load'. Now you're only at dcl. If the selector is only available after load, this has a much greater chance of timing out. Or, put differently, if the selector appears after load, you used to have 10+5 seconds to find it. Now you only have time-to-dcl + 5.

src/browser/actions.zig
```diff
@@ -276,7 +276,10 @@ fn remainingMs(timeout_ms: u32, timer: std.Io.Timestamp) u32 {
 pub fn waitForSelector(selector: [:0]const u8, timeout_ms: u32, frame_id: u32, session: *Session) !*DOMNode {
     const timer: std.Io.Timestamp = .now(lp.io, .boot);
     var runner = session.runner(.{});
-    try runner.waitForFrame(frame_id, timeout_ms, .{ .until = .load });
+    // Polling needs a parsed document, nothing more. Gating on `.load` would
```
karlseguin, line 279: Just want to flag that this change also changes CDP's `LP.waitForSelector`

arrufat:
Thank you :)

karlseguin (approved):

PR #3157, opened 2026-08-06
Entry Limit for HTTP Cache

mookums:
This adds a `http_cache_entry_limit` flag that allows you to limit the maximum number of entries that persist in the HTTP Cache. It currently uses `stored_at` times as a heuristic to determine what gets evicted when full.

### DIFF
--- a/src/Config.zig
+++ b/src/Config.zig
@@ -191,7 +191,7 @@ const CommonOptions = .{
     .{ .name = "log_filter_scopes", .type = log.FilterRule, .multiple = true, .validator = logFilterScopesValidator },
     .{ .name = "user_agent_suffix", .type = ?[]const u8 },
     .{ .name = "http_cache_dir", .type = ?[]const u8 },
-    .{ .name = "http_cache_entry_limit", .type = ?u32 },
+    .{ .name = "http_cache_entry_limit", .type = ?u32, .default = 1000 },
     .{ .name = "web_bot_auth_key_file", .type = ?[]const u8 },
     .{ .name = "web_bot_auth_keyid", .type = ?[]const u8 },
     .{ .name = "web_bot_auth_domain", .type = ?[]const u8 },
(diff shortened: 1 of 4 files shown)


src/Config.zig
```diff
@@ -191,6 +191,7 @@ const CommonOptions = .{
     .{ .name = "log_filter_scopes", .type = log.FilterRule, .multiple = true, .validator = logFilterScopesValidator },
     .{ .name = "user_agent_suffix", .type = ?[]const u8 },
     .{ .name = "http_cache_dir", .type = ?[]const u8 },
+    .{ .name = "http_cache_entry_limit", .type = ?u32 },
```
```suggestion
    .{ .name = "http_cache_entry_limit", .type = u32, .default = 1000 },
```

Then you can remove the `orlese` in `httpCacheEntryLimit`..just groups things together a little better.

src/network/cache/SqliteCache.zig
```diff
@@ -125,13 +127,25 @@ pub fn init(allocator: std.mem.Allocator, path: SqliteCachePath) !SqliteCache {
...
 
 pub fn deinit(self: *SqliteCache) void {
     self.pool.deinit(self.allocator);
 }
 
+fn evictOverflow(self: *SqliteCache, conn: Conn) !void {
+    const limit = self.entry_limit;
+    if (limit == 0) return;
+
+    try conn.exec(
```
karlseguin, line 141: This is relatively expensive, on every put. It would be nice to find a solution to that but nothing immediately comes to mind. We can at least optimize the query. As-is,we're doing table scans (with a large body in there):

```
QUERY PLAN
|--SCAN cache
`--LIST SUBQUERY 1
   |--SCAN cache USING INDEX cache_stored_at
   `--CREATE BLOOM FILTER
```

But, written as:

```sql
delete from cache
where rowid in (
    select rowid from cache order by stored_at desc limit -1 offset $1
)
```

It avoids the scan and uses a covering index:

```sql
QUERY PLAN
|--SEARCH cache USING INTEGER PRIMARY KEY (rowid=?)
`--LIST SUBQUERY 1
   |--SCAN cache USING COVERING INDEX cache_stored_at
   `--CREATE BLOOM FILTER
```

PR #3158, opened 2026-08-06
Single Flight

mookums:
This adds a `SingleFlight` component that can be reused across future gates, such as the CORS one.

### DIFF
--- a/src/network/RobotsGate.zig
+++ b/src/network/RobotsGate.zig
@@ -39,7 +39,6 @@ const Allocator = std.mem.Allocator;
 const RobotsGate = @This();
 
 network: *Network,
-allocator: Allocator,
 single_flight: SingleFlight,
 
 pub const Result = enum { allowed, blocked, pending };
--- a/src/network/RobotsGate.zig
+++ b/src/network/RobotsGate.zig
@@ -31,6 +31,7 @@ const http = @import("http.zig");
 const Robots = @import("Robots.zig");
 const Network = @import("Network.zig");
 const Transfer = @import("HttpClient.zig").Transfer;
+const SingleFlight = @import("SingleFlight.zig");
 
 const log = lp.log;
 const Allocator = std.mem.Allocator;
@@ -39,16 +40,12 @@ const RobotsGate = @This();
 
 network: *Network,
 allocator: Allocator,
-pending: std.StringHashMapUnmanaged(std.ArrayList(*Transfer)) = .empty,
+single_flight: SingleFlight,
 
 pub const Result = enum { allowed, blocked, pending };
 
 pub fn deinit(self: *RobotsGate) void {
-    var it = self.pending.iterator();
-    while (it.next()) |entry| {
-        entry.value_ptr.deinit(self.allocator);
-    }
-    self.pending.deinit(self.allocator);
+    self.single_flight.deinit();
 }
 
 pub fn check(self: *RobotsGate, transfer: *Transfer) !Result {
@@ -77,25 +74,10 @@ pub fn check(self: *RobotsGate, transfer: *Transfer) !Result {
 // stays: the in-flight fetch owns it (the key lives on the fetch's context
 // arena) and still resolves the remaining waiters.
 pub fn remove(self: *RobotsGate, transfer: *Transfer) void {
-    var it = self.pending.valueIterator();
-    while (it.next()) |waiting| {
-        for (waiting.items, 0..) |t, i| {
-            if (t == transfer) {
-                _ = waiting.swapRemove(i);
-                return;
-            }
-        }
-    }
+    self.single_flight.remove(transfer);
 }
 
 fn fetchThenResume(self: *RobotsGate, robots_url: [:0]const u8, transfer: *Transfer) !void {
-    if (self.pending.getPtr(robots_url)) |waiting| {
-        // A fetch for this robots.txt is already in flight, queue behind it.
-        try waiting.append(self.allocator, transfer);
-        transfer.park(.robots);
-        return;
-    }
-
     const client = transfer.client;
 
     // The context, the response buffer and the pending-map key live on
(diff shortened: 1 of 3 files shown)


src/network/RobotsGate.zig
```diff
@@ -107,6 +89,15 @@ fn fetchThenResume(self: *RobotsGate, robots_url: [:0]const u8, transfer: *Trans
     errdefer arena.release();
 
     const owned_url = try arena.dupeZ(u8, robots_url);
+    const result = try self.single_flight.enter(owned_url, transfer, .robots);
```
karlseguin: Any failure after this ends up caller `arena.release()`  but the single_flight is still referencing allocations made from the arena.

src/network/RobotsGate.zig
```diff
@@ -107,6 +89,15 @@ fn fetchThenResume(self: *RobotsGate, robots_url: [:0]const u8, transfer: *Trans
...
+    if (result == .queued) {
+        // Someone else already owns an in-flight fetch for this key
+        // (under its own, earlier-allocated owned_url). This arena was
+        // never handed to a fetch, so release it ourselves.
+        arena.release();
```
karlseguin: Should be able to avoid getting the arena and duping a value in this case.

src/network/RobotsGate.zig
```diff
@@ -161,11 +142,11 @@ fn fetchThenResume(self: *RobotsGate, robots_url: [:0]const u8, transfer: *Trans
 // each judged against its own path. No store entry (fetch failed, or a 200
 // whose body never got parsed) fails open.
 fn flushPending(self: *RobotsGate, robots_url: []const u8) void {
-    var queued = self.pending.fetchRemove(robots_url) orelse return;
-    defer queued.value.deinit(self.allocator);
+    var queued = self.single_flight.take(robots_url) orelse return;
+    defer queued.deinit(self.allocator);
```
karlseguin: the list is allocated via the SingleFlight's `self.allocator` and here it's freed using the RobotGate's `self.allocator`. Possibly the same allocator in the end, but doesn't seem safe to depend on that.

src/network/SingleFlight.zig
```diff
@@ -0,0 +1,352 @@
...
+
+    waiting.* = .empty;
+    try waiting.append(self.allocator, transfer);
+    errdefer waiting.deinit(self.allocator);
+    transfer.park(reason);
+
+    return .initial;
+}
+
+pub fn abort(self: *SingleFlight, key: []const u8) void {
```
karlseguin: not called by anyone?

src/network/SingleFlight.zig
```diff
@@ -0,0 +1,352 @@
...
+    };
+    return t;
+}
+
+test "SingleFlight: enter returns initial for the first waiter" {
+    var pool = ArenaPool.init(testing.allocator, .{});
+    defer pool.deinit();
+
+    var client: HttpClient = undefined;
+    // Only transfers.remove/pending_queue.remove/etc. touched by deinit
```
karlseguin: All tests pass if client is set to undefined? The comment seems incorrect?  Also, client is set to undefined twice. test_browser.http_client might be the safest and most correct option and removes 3 lines per test.

src/network/RobotsGate.zig
```diff
@@ -77,61 +74,37 @@ pub fn check(self: *RobotsGate, transfer: *Transfer) !Result {
...
 
-    const owned_url = try arena.dupeZ(u8, robots_url);
     const robots_ctx = try arena.create(RobotsContext);
     robots_ctx.* = .{
         .gate = self,
         .buffer = .empty,
         .arena = arena,
         .arena_pool = client.arena_pool,
-        .robots_url = owned_url,
+        .robots_url = robots_url,
```
karlseguin: The unnecessary dupe is fixed, but now this can dangle. It needs to be duped by  arena (but not unnecessarily when it's already enqueued in singleflight, which was the state in the previous commit.

src/network/SingleFlight.zig
```diff
@@ -0,0 +1,310 @@
...
+    }
+
+    const owned_key = try self.allocator.dupe(u8, key);
+    errdefer self.allocator.free(owned_key);
+
+    var waiting: std.ArrayList(*Transfer) = .empty;
+    errdefer waiting.deinit(self.allocator);
+    try waiting.append(self.allocator, transfer);
+
+    try self.pending.put(self.allocator, owned_key, waiting);
```
karlseguin: You could use putNoClobber here, but the way we typically handle this is:

```zig
const gop = self.pending.getOrPut(key);
if (gop.found_existing) {
  // .. existing code
}
const owned_key = try self.allocator.dupe(u8, key);
errdefer self.allocator.free(owned_key);
gop.key_ptr.* = owned_key;
// ....
```

So you get 1 lookup in all cases.

src/network/RobotsGate.zig
```diff
@@ -39,16 +40,12 @@ const RobotsGate = @This();
...
+single_flight: SingleFlight,
 
 pub const Result = enum { allowed, blocked, pending };
 
 pub fn deinit(self: *RobotsGate) void {
-    var it = self.pending.iterator();
-    while (it.next()) |entry| {
-        entry.value_ptr.deinit(self.allocator);
-    }
-    self.pending.deinit(self.allocator);
```
karlseguin, line 51: nothing uses self.allocator now, it's a dead field.

src/network/SingleFlight.zig
```diff
@@ -0,0 +1,310 @@
...
+    try testing.expectEqual(Transfer.State{ .parked = .robots }, t1.state);
+    try testing.expectEqual(Transfer.State{ .parked = .robots }, t2.state);
+    try testing.expectEqual(Transfer.State{ .parked = .robots }, t3.state);
+}
+
+test "SingleFlight: different keys get independent entries" {
+    var pool = ArenaPool.init(testing.allocator, .{});
+    defer pool.deinit();
+
+    var client = testing.test_browser.http_client;
```
karlseguin:
```suggestion
    var client = &testing.test_browser.http_client;
```

(in all tests)

src/network/SingleFlight.zig
```diff
@@ -0,0 +1,313 @@
...
+pub const EnterResult = enum { initial, queued };
+
+pub fn enter(self: *SingleFlight, key: []const u8, transfer: *Transfer, reason: Transfer.ParkedBy) !EnterResult {
+    const gop = try self.pending.getOrPut(self.allocator, key);
+    if (gop.found_existing) {
+        const waiting = gop.value_ptr;
+        try waiting.append(self.allocator, transfer);
+        transfer.park(reason);
+        return .queued;
+    }
```
krichprollsch, line 50: We should remove the insertion into `self.pending` in case of a following error.
```
errdefer _ = self.pending.remove(key);
```

src/network/SingleFlight.zig
krichprollsch: The [suggestion](https://github.com/lightpanda-io/browser/pull/3158#discussion_r3773636624) still applies

krichprollsch (approved):

PR #3164, opened 2026-08-10
mask url fields from crash report

krichprollsch:


src/crash_handler.zig
```diff
@@ -101,6 +101,11 @@ fn report(reason: []const u8, begin_addr: usize, args: anytype) !void {
     const body = blk: {
         var writer: std.Io.Writer = .fixed(body_buffer[0..8191]); // reserve 1 space
         inline for (@typeInfo(@TypeOf(args)).@"struct".fields) |f| {
+            // remove url value from the crash report.
+            if (comptime std.mem.eql(u8, f.name, "url")) {
```
karlseguin, line 105: what about if a value has an http://  or https://.... ?

krichprollsch: it's possible, but I think a comptime check on the field name is good enough.  I'm not sure a runtime check on each value worth it. but it's safer on the other hand...

PR #3167, opened 2026-08-11
dev: add -Ddev_fast flag

karlseguin:
Only works in Debug mode. Only works on x86_64 Linux.

sets use_llvm = false and forces glibc 2.43 (1), causes v8 to be dynamically linked (https://github.com/lightpanda-io/zig-v8-fork/pull/197).

For me, zig build -Ddev_fast is about 6x faster (~60 seconds -> ~10 seconds)

(1) https://codeberg.org/ziglang/zig/issues/31272

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -36,12 +36,39 @@ const Build = blk: {
 };
 
 pub fn build(b: *Build) !void {
-    const target = b.standardTargetOptions(.{});
     const optimize = b.standardOptimizeOption(.{});
+
+    // The three settings below only work as a set, so they get one knob rather
+    // than three defaults: the self-hosted backend needs the shared V8 (it
+    // cannot apply the CREL relocations in the archive) and needs zig's own CRT
+    // (a system crt1.o with SFrame unwind data has relocations its linker does
+    // not handle). Debug-only, opt-in, and a good deal faster to rebuild.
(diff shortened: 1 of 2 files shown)


build.zig
```diff
@@ -36,12 +36,39 @@ const Build = blk: {
...
 pub fn build(b: *Build) !void {
-    const target = b.standardTargetOptions(.{});
     const optimize = b.standardOptimizeOption(.{});
 
+    // The three settings below only work as a set, so they get one knob rather
+    // than three defaults: the self-hosted backend needs the shared V8 (it
+    // cannot apply the CREL relocations in the archive) and needs zig's own CRT
+    // (a system crt1.o with SFrame unwind data has relocations its linker does
+    // not handle). Debug-only, opt-in, and a good deal faster to rebuild.
+    const dev_fast = b.option(bool, "dev_fast", "Linux debug builds: shared V8 + self-hosted backend. Implies -Dshared_v8, -Duse_llvm=false and a bundled-CRT target") orelse false;
```
krichprollsch, line 46: I tried w/ and w/o `-Duse_llvm=false` and the big win is using the zig backend on my side: 9s vs 60s 
🎉

krichprollsch (approved):
So nice, thanks. We can merge as it, but I think the only thing I would like to have is a way to download a pre-built `.so` v8 for linux similar to what we have with the `.a`

karlseguin:
I'll look at the prebuilt download after.

PR #3168, opened 2026-08-11
webapi: improve events from shadowroot

karlseguin:
Correctly encapsulate events that happen within a shadowroot to stop from leaking out. This involves calculating the related_target relative to the potential node receiving the event.

Also Event.composedPath now re-uses the same path-building logic as EventManager (previously, the same-ish algo in both places, that were out of sync).

Some smaller changes:
- input.list getter (showed up in the shadow dom wpt tests)
- activeElement getter added to ShadowDOM
- document activeElement is ShadowDOM-aware and won't cross


krichprollsch (approved):
I think ( 😅 ) it looks good to me 👍

PR #3170, opened 2026-08-11
cdp: fall back to loopback when /json/version bind host is a wildcard

### LINKED ISSUE #1922: [BUG] CDP WebSocketDebuggerUrl returns 0.0.0.0 causing remote CDP clients to fail (ECONNREFUSED)
**Bug Report**

When starting Lightpanda using Docker with `--host 0.0.0.0`, the `/json/version` endpoint returns `webSocketDebuggerUrl` as `ws://0.0.0.0:9222/`. If another machine (such as 192.168.0.127) attempts to connect using Playwright MCP to a Lightpanda instance running on 192.168.0.128, Playwright MCP will try to connect to `ws://0.0.0.0:9222/`.

Since `0.0.0.0` is not a valid address for clients to connect to (it represents INADDR_ANY for binding sockets), the client receives ECONNREFUSED and cannot establish a CDP connection.

**Steps to Reproduce:**
1. On machine 192.168.0.128, run Lightpanda Docker:
    ```
    docker run --name lightpanda -p 9222:9222 lightpanda/browser:nightly /bin/lightpanda serve --host 0.0.0.0 --port 9222
    ```
2. On machine 192.168.0.127 (or any other machine), run Playwright MCP:
    ```
    npx @playwright/mcp@latest --cdp-endpoint http://192.168.0.128:9222
    ```
3. Observe the error:
    ```
    Error: browserType.connectOverCDP: WebSocket error: connect ECONNREFUSED 0.0.0.0:9222
    Call log:
     - <ws preparing> retrieving websocket url from http://192.168.0.128:9222
     - <ws connecting> ws://0.0.0.0:9222/
     - <ws error> ws://0.0.0.0:9222/ error connect ECONNREFUSED 0.0.0.0:9222
     - <ws connect error> ws://0.0.0.0:9222/ connect ECONNREFUSED 0.0.0.0:9222
     - <ws disconnected> ws://0.0.0.0:9222/ code=1006 reason=
    ```

**Expected Behavior:**
- The `webSocketDebuggerUrl` should return a client-accessible address (e.g. `ws://192.168.0.128:9222/` or `ws://localhost:9222/`) instead of `0.0.0.0`.
- This allows remote Playwright MCP clients to connect successfully.

**Suggested Fix:**
- When generating the `webSocketDebuggerUrl`, detect if the host is `0.0.0.0`; return the resolved public IP or allow configuration.
- Alternatively, provide a CLI or environment variable to override the returned WebSocket endpoint for `/json/version`.

**Root Cause:**
In [`src/Server.zig`](https://github.com/lightpanda-io/browser/blob/edd0c5c83f716e365a8bf8b77bac0bd357b2861d/src/Server.zig#L488-L492), the `buildJSONVersionResponse` function directly formats the bound `address` into the `webSocketDebuggerUrl`:
```zig
const body_format = "{{\"webSocketDebuggerUrl\": \"ws://{f}/\"}}";
```
When `--host 0.0.0.0` is used, `0.0.0.0` is placed verbatim in the URL, which is not a routable address for remote clients.

**Environment:**
- Lightpanda browser (Docker nightly release)
- Playwright MCP client
- Docker host: 192.168.0.128
- Client: 192.168.0.127

guyua9:
Closes #1922.

When `serve` is started with `--host 0.0.0.0` (or `::`) and no explicit `--advertise-host`, `/json/version` was returning `webSocketDebuggerUrl: ws://0.0.0.0:9222/`. That URL is unroutable by any client (Chromedp, Playwright MCP, etc.) and surfaces as `ECONNREFUSED` on the very next dial. This is exactly the failure mode reported in #1922 when the official Docker image (`Dockerfile` `--host 0.0.0.0 --port 9222`) is used from another host.

The existing `--advertise-host` flag added in #1955 was a workaround, not a fix: the default behavior is still unsafe and operators have to know to set it.

This PR makes the default safe:

- `Config.advertiseHost()` substitutes `127.0.0.1` for wildcard binds when no `--advertise-host` is provided. Same-host clients (the Docker host loopback via the published port, or the WSL localhost bridge) now get a connectable URL out of the box.
- Operators serving CDP on a dedicated interface still pin the URL explicitly via `--advertise-host <reachable>`; the wildcard fallback only kicks in when nothing is set.
- `Server.buildJSONVersionResponse` keeps emitting a guidance log when the bind is a wildcard, so multi-host setups are still surfaced.

Unit tests cover the four relevant branches:
- `--host 0.0.0.0` and `--host ::` fall back to `127.0.0.1`
- explicit `--advertise-host` overrides the bind host
- concrete bind host (`127.0.0.1`) is preserved unchanged

The existing `server: buildJSONVersionResponse` and `server: get /json/version` tests are untouched because the test harness binds to `127.0.0.1:9583`, which is not a wildcard.

Validation notes:
- I edited the upstream `main` snapshot (`upstream/main` HEAD `6312d43d97`); the local `/Users/ming/code/lightpanda-io/browser` main is 314 commits behind, so I used a worktree at `upstream/main` to keep the diff base correct.
- I do not have Zig 0.15+ / `v8` toolchain installed locally, so I have not run `zig build test`. The change is two line-flip-class edits + three test cases; upstream CI on the PR should validate.
- Branch: `guyua9:fix/advertise-host-wildcard`
- Head SHA: `8be30d56f8032c12520d7a0cef115928fda37c22`
- Account: `guyua9` (registered in `github-accounts.json`, browserProfile `hm11`)

### DIFF
--- a/src/Config.zig
+++ b/src/Config.zig
@@ -677,7 +677,7 @@ pub fn advertiseHost(self: *const Config) []const u8 {
 // For remote hosts, users can still pin the URL with --advertise-host.
 // See issue #1922.
 fn advertiseHostFallback(host: []const u8) []const u8 {
-    if (std.mem.eql(u8, host, "0.0.0.0") or std.mem.eql(u8, host, "::")) {
+    if (isHostWildcard(host)) {
         return "127.0.0.1";
     }
     return host;
@@ -689,9 +689,13 @@ fn advertiseHostFallback(host: []const u8) []const u8 {
 // resolvable, but the caller still benefits from emitting a guidance log.
 pub fn bindIsWildcard(self: *const Config) bool {
     return switch (self.mode) {
-        .serve => |opts| std.mem.eql(u8, opts.host, "0.0.0.0") or std.mem.eql(u8, opts.host, "::"),
+        .serve => |opts| isHostWildcard(opts.host),
         else => false,
     };
+}
+
+fn isHostWildcard(host: []const u8) bool {
+    return std.mem.eql(u8, host, "0.0.0.0") or std.mem.eql(u8, host, "::");
 }
 
 pub fn webBotAuth(self: *const Config) ?WebBotAuthConfig {
--- a/src/Config.zig
+++ b/src/Config.zig
@@ -663,9 +663,34 @@ pub fn port(self: *const Config) u16 {
 
 pub fn advertiseHost(self: *const Config) []const u8 {
     return switch (self.mode) {
-        .serve => |opts| opts.advertise_host orelse opts.host,
+        .serve => |opts| opts.advertise_host orelse advertiseHostFallback(opts.host),
         .mcp => "127.0.0.1",
         else => unreachable,
+    };
+}
+
+// Wildcard bind addresses (0.0.0.0, ::) are not routable for clients
+// resolving /json/version. Fall back to a loopback address so the
+// advertised webSocketDebuggerUrl is at least connectable from the same
+// host (covers the official Docker image, which exposes 9222 via the
+// container's published port, and the WSL localhost bridge).
+// For remote hosts, users can still pin the URL with --advertise-host.
+// See issue #1922.
+fn advertiseHostFallback(host: []const u8) []const u8 {
+    if (std.mem.eql(u8, host, "0.0.0.0") or std.mem.eql(u8, host, "::")) {
+        return "127.0.0.1";
+    }
+    return host;
+}
+
+// True when serve is binding a wildcard address (e.g. Docker --host
+// 0.0.0.0) without an explicit --advertise-host. /json/version replaces
+// the wildcard with 127.0.0.1 in advertiseHostFallback so the URL stays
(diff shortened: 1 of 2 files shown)


src/Config.zig
```diff
@@ -663,12 +663,37 @@ pub fn port(self: *const Config) u16 {
...
 
+// Wildcard bind addresses (0.0.0.0, ::) are not routable for clients
+// resolving /json/version. Fall back to a loopback address so the
+// advertised webSocketDebuggerUrl is at least connectable from the same
+// host (covers the official Docker image, which exposes 9222 via the
+// container's published port, and the WSL localhost bridge).
+// For remote hosts, users can still pin the URL with --advertise-host.
+// See issue #1922.
+fn advertiseHostFallback(host: []const u8) []const u8 {
+    if (std.mem.eql(u8, host, "0.0.0.0") or std.mem.eql(u8, host, "::")) {
```
krichprollsch: What about using `self.bindIsWildcard()` here also?

PR #3171, opened 2026-08-11
`Element`: alias `matches` with `webkitMatchesSelector`

nikneym:
Adds legacy alias of matches(), kept for compatibility per the DOM spec.


karlseguin:
Firefox exposes it too.

PR #3173, opened 2026-08-11
`Origin` and (additional) `Host` header validation for CDP conns

nikneym:
This PR adds `Origin` header validation to `Connection.upgrade`; where we validate CDP connections.
1. Any WebSocket upgrade request carrying an Origin header is rejected with 403 Origin not allowed.
2. the Host header must be an IP literal (for example 127.0.0.1:9222). Any host name, including localhost, is rejected with 403 Host not allowed.

### DIFF
--- a/src/cdp/Connection.zig
+++ b/src/cdp/Connection.zig
@@ -246,6 +246,8 @@ fn processHttpRequest(self: *Connection) !HttpResult {
     return self.handleHttpRequest(request) catch |err| {
         switch (err) {
             error.NotFound => self.sendHttpError(404, "Not found"),
+            error.ForbiddenOrigin => self.sendHttpError(403, "Origin not allowed"),
+            error.ForbiddenHost => self.sendHttpError(403, "Host not allowed"),
             error.InvalidRequest => self.sendHttpError(400, "Invalid request"),
             error.InvalidProtocol => self.sendHttpError(400, "Invalid HTTP protocol"),
             error.MissingHeaders => self.sendHttpError(400, "Missing required header"),
@@ -434,8 +436,6 @@ fn pushCdp(self: *Connection, bytes: []const u8) !bool {
 }
 
 pub fn upgrade(self: *Connection, request: []u8) !void {
-    // We need to extract the `Sec-WebSocket-Key` value.
-    var sec_websocket_key: []const u8 = "";
     // We need to make sure that we got all the necessary headers + values;
     // a bit per required header.
     const FOUND_UPGRADE: u8 = 1 << 0; // Upgrade: websocket
@@ -453,6 +453,8 @@ pub fn upgrade(self: *Connection, request: []u8) !void {
     }
 
     var found_headers: u8 = 0;
+    // We need to extract the `Sec-WebSocket-Key` value.
+    var sec_websocket_key: []const u8 = "";
 
     // A malformed header maps to a 400 in processHttpRequest.
     while (header_iterator.next() catch return error.InvalidRequest) |header| {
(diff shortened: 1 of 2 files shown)


karlseguin:
Is the point here strictly to prevent _our_ browser from connecting to its own CDP instance? AND/OR is it to stop someone running lightpanda AND Firefox, and having Firefox connect to the lightpanda instance?

The implementation is based on the fact that we always send an Origin header, but that isn't true. The implementation also fails to protect the HTTP endpoints, e.g /json/version.

The one header we always send is `Sec-Ch-Ua`, so if we're just trying to protect against the first  case, it seems like checking this header is both simpler and more complete. Arguably using the header this way might make the header a little too "load bearing".  There's also the possibility of blocking connections to the listen port, but that could unintentionally block valid external requests.

krichprollsch:
> Is the point here strictly to prevent our browser from connecting to its own CDP instance? AND/OR is it to stop someone running lightpanda AND Firefox, and having Firefox connect to the lightpanda instance?

Originally the fix is to block the 2nd case, ie. an external browser connecting to LP instance.
I guess when LP will implement CORS, it will be blocked the same way.

All clients don't send origin header at all.

Here is the chrome's rules according to Claude:

```
Host check for all HTTP endpoints

  Rule:
  - No Host header → allowed.
  - Host is an IP address (like 127.0.0.1, 192.168.1.5) → allowed.
  - Host is a local name (localhost, or a name ending in .localhost) → allowed.
  - Anything else (a real domain name like evil.com) → rejected with HTTP 500.

Origin check for WS connection

Rule:
  - No Origin header at all → allowed. This is the key point. Tools like Puppeteer, Playwright, or a raw socket client do not send Origin, so they always pass. The check only bites  real browsers, because a browser always adds Origin and a malicious page cannot remove or fake it.
  - Origin header present → it must be an exact string match against the allow-list, or the allow-list must contain *. Otherwise → rejected with HTTP 403.
  ```
  
  Isn't simpler to just refuse any connection including an origin?

**Comment by krichprollsch** [comment]
The `--remote-allow-origins` chrome's option allows connection from browsers to the websocket. It is useful when you want to control the CDP using a web ui devtool (maybe also allowing my chrome's devtool to connect to another remote browser). But I don't think we really need it for Lightpanda.

**Comment by karlseguin** [comment]
> Isn't simpler to just refuse any connection including an origin?

Yes. Question is whether `/metrics`, `/json/version` should also be blocked, because Firefox won't send Origin for these (I like being able to access them from Firefox for development).

**Comment by krichprollsch** [comment]
Re-entrant call will be blocked by CORS mechanism.
So I don't think we need to protect using origin check.

But it seems checking if the `host` is IP will protect us against DNS rebinding. So maybe it worth it.

**Comment by nikneym** [comment]
> Is the point here strictly to prevent our browser from connecting to its own CDP instance? AND/OR is it to stop someone running lightpanda AND Firefox, and having Firefox connect to the lightpanda instance?

Idea is to prevent both.

> The implementation is based on the fact that we always send an Origin header, but that isn't true.

It's optional, actually. We rely on the fact that browsers always send it and automation tools (like Puppeteer) do not.

> The implementation also fails to protect the HTTP endpoints, e.g /json/version.

That's the domain of CORS, what's implemented here is more for Cross Site WebSocket Hijacking (CSWH).

> The one header we always send is Sec-Ch-Ua, so if we're just trying to protect against the first case, it seems like checking this header is both simpler and more complete.

I think that'd also be good to have, though not all browsers seem to send it.

**Comment by nikneym** [comment]
> But it seems checking if the host is IP will protect us against DNS rebinding. So maybe it worth it.

The `isHostAllowed` function should cover it.

**Comment by karlseguin** [comment]
@nikneym what about @krichprollsch suggesting to just block any request with an Origin header?

**Comment by nikneym** [comment]
Yes, that's the approach we've settled on.

File: src/cdp/Connection.zig [hunk 1]
```diff
@@ -481,11 +485,48 @@ pub fn upgrade(self: *Connection, request: []u8) !void {
...
     }
 
+    // Only a browser sends `Origin`, and a browser has no business driving
+    // CDP: whatever page sent this is cross-origin to us by definition,
+    // including one served from loopback itself. Scripted clients (Puppeteer,
+    // Playwright, chromedp, ...) never send it. Rejecting here rather than
+    // inside the loop is what lets the duplicate check above run at all.
+    if (found_headers & FOUND_ORIGIN == FOUND_ORIGIN) {
+        log.warn(.cdp, "rejected websocket origin", .{});
+        return error.ForbiddenOrigin;
```
krichprollsch: What about returning earlier, immediately when you find the first origin header?

src/cdp/Connection.zig
```diff
@@ -526,6 +567,29 @@ pub fn upgrade(self: *Connection, request: []u8) !void {
     return self.send(response);
 }
 
+// The Host header of a request that reached a loopback listener has to be an
+// IP literal: nothing resolved it, so nothing could have pointed it at us.
+fn isHostAllowed(host: []const u8) bool {
```
krichprollsch: What about using [parseLiteral]() which handle port and both type of IPs instead?

krichprollsch:
Can you also squash the commits please?

krichprollsch (commented):
Can you revert the changes on `FOUND_ALL` and remove `FOUND_HOST`?
We don't need to verify the host is unique. So this change is useless.

krichprollsch (approved):

PR #3187, opened 2026-08-13
add --http-header option to send headers on every HTTP requests

krichprollsch:
Add `--http-header` option to configure default headers for all http requests.
Websocket is not covered.

`User-Agent` and `Sec-Ch-Ua` are refused.

Additional headers added by the browser like `Referer`, `Cache-Control`, `Accept` will replace the configured headers.

```$ ./zig-out/bin/lightpanda fetch --http-header "Accept-Language: fr-FR,fr;q=0.9" --dump markdown https://httpbin.io/headers
INFO  frame : navigate . . . . . . . . . . . . . . . . . . . . [+0ms]
      url = https://httpbin.io/headers
      method = GET
      reason = address_bar
      body = false
      type = root

{
  "headers": {
    "Accept": [
      "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8"
    ],
    "Accept-Encoding": [
      "deflate, gzip, br"
    ],
    "Accept-Language": [
      "fr-FR,fr;q=0.9"
    ],
    "Host": [
      "httpbin.io"
    ],
    "Sec-Ch-Ua": [
      "\"Lightpanda\";v=\"1\""
    ],
    "User-Agent": [
      "Lightpanda/1.0"
    ]
  }
}
```

### DIFF
--- a/src/Config.zig
+++ b/src/Config.zig
@@ -24,6 +24,7 @@ const cli = @import("cli.zig");
 const dump = @import("browser/dump.zig");
 
 const WebBotAuthConfig = @import("network/WebBotAuth.zig").Config;
+const HttpHeader = @import("network/http.zig").Header;
 
 const log = lp.log;
 const crypto = @import("sys/libcrypto.zig");
@@ -83,6 +84,29 @@ fn logLevelValidator(_: Allocator, args: *std.process.Args.Iterator, target: *?l
         log.fatal(.app, "invalid option choice", .{ .arg = "--log-level", .value = str });
         return error.InvalidArgument;
     };
+}
+
+fn httpHeaderValidator(allocator: Allocator, args: *std.process.Args.Iterator, list: *std.ArrayList(HttpHeader)) !void {
+    const str = args.next() orelse return error.MissingArgument;
+    const header = HttpHeader.parse(str) orelse {
+        log.fatal(.app, "invalid option value", .{ .arg = "--http-header", .value = str, .hint = "expected \"Name: Value\"" });
+        return error.InvalidArgument;
+    };
+
+    if (std.ascii.eqlIgnoreCase(header.name, "User-Agent")) {
+        log.fatal(.app, "invalid option value", .{ .arg = "--http-header", .value = str, .hint = "Use --user-agent instead" });
+        return error.InvalidArgument;
+    }
+
+    if (std.ascii.eqlIgnoreCase(header.name, "Sec-Ch-Ua")) {
+        log.fatal(.app, "invalid option value", .{ .arg = "--http-header", .value = str, .hint = "Sec-Ch-Ua is not oevrridable" });
+        return error.InvalidArgument;
+    }
+
+    try list.append(allocator, .{
+        .name = try allocator.dupe(u8, header.name),
+        .value = try allocator.dupe(u8, header.value),
+    });
 }
 
 const Cert = struct {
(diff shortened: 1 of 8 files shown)


src/Config.zig
```diff
@@ -85,6 +86,23 @@ fn logLevelValidator(_: Allocator, args: *std.process.Args.Iterator, target: *?l
     };
 }
 
+fn httpHeaderValidator(allocator: Allocator, args: *std.process.Args.Iterator, list: *std.ArrayList(HttpHeader)) !void {
```
karlseguin, line 90: Test `--http-header $'X-A: b\r\nUser-Agent: Mozilla/5.0'`  and it resulted in 2 User-Agent's being added

src/Config.zig
```diff
@@ -85,6 +86,23 @@ fn logLevelValidator(_: Allocator, args: *std.process.Args.Iterator, target: *?l
...
+    const str = args.next() orelse return error.MissingArgument;
+    const header = HttpHeader.parse(str) orelse {
```
karlseguin, line 92: CLI.zig dupes string values, so I think it should be duped here.

src/help.zon
```diff
@@ -352,6 +352,9 @@
     \\          Time in ms to establish an HTTP connection before timing out. 0 means
     \\          never.
     \\          Defaults to 0.
+    \\  --http-header <HEADER>
+    \\          Extra default value header added to every outgoing HTTP request,
```
karlseguin:
```suggestion
    \\          Extra default value header added to \*every\* outgoing HTTP request,
```

Wondering if it's worth mentioning redirects explicitly.

karlseguin (commented):
I find the behavior a bit ambiguous. Maybe you had a specific use case it mind. I was thinking  addHeader should become private. So that only "setHeader" is exposed, meaning every layer strictly becomes "override". Then you can document the chain.

The only wrinkle I see to that is that webapi/net/Headers.zig uses `addHeader` now since JS _could_ add the same header multiple times. But, by spec we're supposed to combine those into a single line (which we don't do now). So if pre-merged in Header.zig then (a) setHeader would be right and (b) it would be per-spec.

src/Config.zig
krichprollsch: Do you think I should specifically check for EOL in the CLI arg, or if I should change http.Header parser? :thinking:

karlseguin: The only thing using Header.parse right now is Frame and its only using it for a "controlled" header (the comment about it coming from the CLI seems wrong). Almost possible to remove the function, except it's useful as-is so we can dupe that value. Long story short, changing the parse function is safe / minimal today, and might save us in the future.

krichprollsch:
> I find the behavior a bit ambiguous. Maybe you had a specific use case it mind.

My use case is to add a default HTTP header into the cloud platform to identify requests from it.
I currently set a UA suffix, but user's can easily change it via CDP.
I know they will be able to still change any header through CDP, but it less likely. 

Curl offers a `--header` option to set extra http header and even replace any of the curl's values.
So the behavior here is different. 

Maybe the curl's way is better: the CLI header are forced and replace any browser's change (internal or via CDP).

krichprollsch:
Ok I will rebase this PR on #3200
#3200 fix missing checks for headers + implement the `Transfer.addHeader` private suggestion.

Then here I will use it to introduce a new RequestHeader source making any cli defined header non-overridable.

karlseguin (approved):
We had Mime.isHttpToken for the header value, and I created Mime.isHttpHeaderValue from the existing net/Headers.zig validation. All places (including Config.zig) now use these 2 functions to validate headers.

src/help.zon
```diff
@@ -354,6 +354,13 @@
     \\          Time in ms to establish an HTTP connection before timing out. 0 means
     \\          never.
     \\          Defaults to 0.
+    \\  --http-header <HEADER>
+    \\          Extra custom header added to every outgoing HTTP request, including redirections,
+    \\          in "Name: Value" form. Can be passed multiple times. The first name is used.
```
karlseguin: The last one is used

src/help.zon
```diff
@@ -354,6 +354,13 @@
...
+    \\          Note that if you add a custom header that has the same name as one of the
+    \\          internal ones Lightpanda would use, your set header is used instead of the
```
karlseguin: It also overwrites script-set headers, which could be anything. I think that's worth pointing out. If it causes issues, we might need a 2nd argument which sets a lower-priority header.

krichprollsch:
I updated the help

PR #3189, opened 2026-08-13
webapi: don't fire error event when a worker script fetch is aborted

krichprollsch:
terminate() can cancel a still-inflight script fetch. That's a cancellation, not a failure: a terminated worker must fire no further events, and Chrome is silent here. We logged an error and fired an ErrorEvent — observable by Cloudflare challenge scripts, which create a blob-URL worker and terminate it in the same tick.

On error.Abort, skip the log and the error event; same log skip in SharedWorkerGlobalScope.

Previously we had
```
$ lightpanda fetch https://bed-vyne.com/
$time=1786629337502 $scope=browser $level=error $msg="worker fetch error" url=blob:https://bed-vyne.com/dafb8b4c-d872-47fe-9c31-5c61a3e96753 err=Abort
```

### DIFF
--- a/src/browser/webapi/Worker.zig
+++ b/src/browser/webapi/Worker.zig
@@ -274,10 +274,17 @@ fn httpErrorCallback(ctx: *anyopaque, err: anyerror) void {
     self._http_transfer = null;
     self.releaseScriptArena();
 
-    log.err(.browser, "worker fetch error", .{
-        .url = self._url,
-        .err = err,
-    });
+    // error.Abort is not a load failure: terminate() (or worker teardown)
+    // cancelled a still-inflight script fetch. Per spec a terminated worker
+    // fires no further events, so no error event either — Chrome is silent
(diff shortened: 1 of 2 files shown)


src/browser/webapi/Worker.zig
```diff
@@ -286,7 +293,9 @@ fn httpErrorCallback(ctx: *anyopaque, err: anyerror) void {
     self._script_loaded = true;
     self._worker_scope.drainPendingMessages();
 
-    self.fireErrorEvent(@errorName(err), null);
+    if (!is_abort) {
+        self.fireErrorEvent(@errorName(err), null);
```
karlseguin, line 297: This isn't right. `error.Abort` is used in other cases. Worker.httpHeaderCallback returns an .abort status on a non-200 which eventually translates into an `error.Abort`..these should fire the event.

The least disruptive option is to track an `_aborted: bool` flag, and use that rather than comparing to error.Abort which is used for _any_ abort.

Issue #3163, opened 2026-08-09
Request() rejects an open ReadableStream, breaking Fetch capability detection


turicas:
A public Brazilian election-results page fails before it can load a candidate profile in Lightpanda. The page is an Angular application. While opening a candidate route, it loads one extra JavaScript file. That file runs a small feature check: it asks whether the browser accepts a `ReadableStream` as the body of a `Request`. Firefox accepts the request object, so the application continues. Lightpanda throws `TypeError` instead and Angular stops loading that part of the page.

This is reproducible without the TSE website (see **Minimal independent reproduction**). It appears to be caused by Lightpanda requiring a stream to be closed when `Request` is *created*. Web browsers accept an open stream at that point; the stream may produce data later.

## Environment

- Lightpanda: `1.0.0-nightly.8580+8691568f`
- OS: Linux x86-64
- Affected URL: https://divulgacandcontas.tse.jus.br/divulga/#/candidato/SUDESTE/ES/2045202024/80002363211/2024/56553

## Reproduction

```sh
lightpanda fetch \
  --log-level debug \
  --wait-ms 8000 \
  --dump semantic_tree_text \
  'https://divulgacandcontas.tse.jus.br/divulga/#/candidato/SUDESTE/ES/2045202024/80002363211/2024/56553'
```

<details>
<summary> Output with `TypeError` </summary>
```
$time=1786290795155 $scope=telemetry $level=info $msg="telemetry status" disabled=false
$time=1786290795155 $scope=app $level=debug $msg="startup" mode=fetch dump_mode=semantic_tree_text url=https://divulgacandcontas.tse.jus.br/divulga/#/candidato/SUDESTE/ES/2045202024/80002363211/2024/56553 snapshot=true
$time=1786290795161 $scope=frame $level=info $msg="navigate" url=https://divulgacandcontas.tse.jus.br/divulga/#/candidato/SUDESTE/ES/2045202024/80002363211/2024/56553 method=GET reason=address_bar body=false type=root
$time=1786290795286 $scope=http $level=debug $msg="script fetch start" req=GET https://divulgacandcontas.tse.jus.br/divulga/runtime.113bd67c35d7d457.js
$time=1786290795286 $scope=browser $level=info $msg="executing script" src=https://divulgacandcontas.tse.jus.br/divulga/runtime.113bd67c35d7d457.js kind=module
[...]

PR #3190, opened 2026-08-13
Request Duplex (Readable Stream) Support

mookums:
This adds support for `Request` taking a `ReadableStream`.

https://developer.mozilla.org/en-US/docs/Web/API/Request/duplex

### DIFF
--- a/src/browser/webapi/net/Fetch.zig
+++ b/src/browser/webapi/net/Fetch.zig
@@ -68,6 +68,14 @@ pub fn init(input: Input, options: ?InitOpts, exec: *const Execution) !js.Promis
         }
     }
 
+    const body_bytes: []const u8 = switch (request._body) {
+        .stream => |stream| stream.collectBodyBytes(request._arena.allocator()) catch {
+            resolver.rejectError("fetch init error", .{ .type_error = "Failed to read body stream" });
+            return resolver.promise();
+        },
+        else => request.bodyBytes(),
+    };
+
     const response = try Response.initPending(exec);
     errdefer response.deinit(exec.page);
 
@@ -91,7 +99,7 @@ pub fn init(input: Input, options: ?InitOpts, exec: *const Execution) !js.Promis
         .ctx = fetch,
         .url = request._url,
         .method = request._method,
-        .body = request.bodyBytes(),
+        .body = body_bytes,
         .resource_type = .fetch,
         .cookies = switch (request._credentials) {
             .omit => false,
--- a/src/browser/webapi/net/Fetch.zig
+++ b/src/browser/webapi/net/Fetch.zig
@@ -91,7 +91,7 @@ pub fn init(input: Input, options: ?InitOpts, exec: *const Execution) !js.Promis
         .ctx = fetch,
         .url = request._url,
         .method = request._method,
-        .body = request._body,
+        .body = request.bodyBytes(),
         .resource_type = .fetch,
         .cookies = switch (request._credentials) {
             .omit => false,
--- a/src/browser/webapi/net/Request.zig
+++ b/src/browser/webapi/net/Request.zig
@@ -32,6 +32,7 @@ const Headers = @import("Headers.zig");
 const FormData = @import("FormData.zig");
 const body_init = @import("body_init.zig");
 const BodyInit = body_init.BodyInit;
+const ReadableStream = @import("../streams/ReadableStream.zig");
 
 const Execution = js.Execution;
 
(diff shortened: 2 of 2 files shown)


krichprollsch:
@mookums the PR is in draft, is it ready for review?

mookums:
Was mostly finished but needed some updating on some tests.

krichprollsch:
@mookums are you still working on this one?

mookums:
Yeah, done now. It supports taking in a ReadableStream for the Request body, consuming it before use and throwing an error if it isn't ready. It prevents the error on #3163 that is present on main.

src/browser/webapi/net/Request.zig
```diff
@@ -231,32 +263,44 @@ pub fn getHeaders(self: *Request, exec: *const Execution) !*Headers {
...
         return false;
     }
     return self._body_used;
 }
 
 // Marks a present body consumed; returns a rejected promise if it already was.
 fn consume(self: *Request, local: *const js.Local) ?js.Promise {
-    if (self._body == null) {
-        return null;
-    }
```
krichprollsch, line 244: Shouldn't we keep this early return by testing the case of `body == .none`?

src/browser/webapi/net/Request.zig
```diff
@@ -344,7 +388,7 @@ pub fn formData(self: *Request, exec: *const Execution) !js.Promise {
 }
 
 pub fn clone(self: *const Request, exec: *const Execution) !*Request {
-    const arena = try exec.getPinnedArena(if (self._body) |b| b.len else 512, "Request.clone");
+    const arena = try exec.getPinnedArena(512, "Request.clone");
```
krichprollsch, line 391: Keep the hint when body is bytes.
```
  const arena = try exec.getPinnedArena(switch (self._body) {
      .bytes => |b| b.len,
      .none, .stream => 512,
  }, "Request.clone");
```

src/browser/webapi/net/Fetch.zig
```diff
@@ -68,6 +68,14 @@ pub fn init(input: Input, options: ?InitOpts, exec: *const Execution) !js.Promis
         }
     }
 
+    const body_bytes: []const u8 = switch (request._body) {
+        .stream => |stream| stream.collectBodyBytes(request._arena.allocator()) catch {
```
krichprollsch, line 72: You drain the stream, but you let the request in an invalid state.
you should replace the request's body with the collected bytes.
as you do in `Request.consume`

src/browser/webapi/net/Request.zig
```diff
@@ -41,14 +42,32 @@ _rc: lp.RC = .{},
...
+
+    fn isPresent(self: Body) bool {
+        return self != .none;
+    }
+
+    fn clone(self: Body, allocator: std.mem.Allocator) !Body {
+        switch (self) {
+            .none => return .none,
+            .bytes => |b| return .{ .bytes = try allocator.dupe(u8, b) },
+            .stream => |s| return .{ .stream = s },
```
krichprollsch, line 66: Not sure it's safe to share the stream. If one collect it, the other won't be updated 🤔

src/browser/webapi/net/Request.zig
```diff
@@ -122,23 +141,36 @@ pub fn init(input: Input, opts_: ?InitOpts, exec: *const Execution) !*Request {
...
-        // before this one.
-        .request => |r| if (r._body) |b| try arena.dupe(u8, b) else null,
+        .url => .none,
+        // Dupe: the source Request owns its body bytes/stream and may be
+        // finalized before this one.
+        .request => |r| switch (r._body) {
+            .none => .none,
+            .bytes => |b| .{ .bytes = try arena.dupe(u8, b) },
+            .stream => |s| blk: {
+                break :blk .{ .stream = s };
```
krichprollsch, line 171: same about sharing the stream

krichprollsch (commented):
One suggestion to resolve the sharing stream issue could be:

```zig
  // Resolve a stream body to bytes on first need, so every copy (clone(), or
  // `new Request(other)`) owns its own bytes. Null means no body.
  fn resolveBody(self: *Request) !?[]const u8 {
      switch (self._body) {
          .none => return null,
          .bytes => |b| return b,
          .stream => |s| {
              const collected = s.collectBodyBytes(self._arena.allocator()) catch return error.TypeError;
              self._body = .{ .bytes = collected };
              return collected;
          },
      }
  }
```

Then:

```zig
  pub fn clone(self: *Request, exec: *const Execution) !*Request {
      // Fetch: clone() throws if the body is already disturbed.
      if (self._body_used) return error.TypeError;

      const body = try self.resolveBody();
      const arena = try exec.getPinnedArena(if (body) |b| b.len else 512, "Request.clone");
      errdefer arena.release();
      ...
      ._body = if (body) |b| .{ .bytes = try arena.dupe(u8, b) } else .none,
```

and

```
  .request => |r| blk: {
      if (r._body_used) return error.TypeError;
      break :blk if (try r.resolveBody()) |b| .{ .bytes = try arena.dupe(u8, b) } else .none;
  },
```


Or we need Tee for request clone

PR #3191, opened 2026-08-13
`URL.isSameOrigin`

mookums:
This was originally in the CORS PR (#3002) but is isolated and simple enough to be an independent PR.

### DIFF
--- a/src/browser/Frame.zig
+++ b/src/browser/Frame.zig
@@ -638,15 +638,7 @@ pub fn getPinnedArena(self: *Frame, size_or_bucket: anytype, debug: []const u8) 
 
 pub fn isSameOrigin(self: *const Frame, url: [:0]const u8) bool {
     const current_origin = self.origin orelse return false;
-
-    // fastpath
-    if (!std.mem.startsWith(u8, url, current_origin)) {
-        return false;
-    }
-
-    // Starting here, at least protocols are equals.
-    // Compare hosts (domain:port) strictly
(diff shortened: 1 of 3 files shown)


karlseguin:
WorkerGlobalScope still has its own isSameOrigin (which now behaves a bit different than Frame's) ,so it should be updated too I think.

karlseguin:
Frame.zig has a couple commented out cases  (// FIXME ....). Would be nice to (a) see if this fixes them and (b) fix them if they should be.

src/browser/Frame.zig
```diff
@@ -638,15 +638,7 @@ pub fn getPinnedArena(self: *Frame, size_or_bucket: anytype, debug: []const u8)
...
-
-    // fastpath
-    if (!std.mem.startsWith(u8, url, current_origin)) {
-        return false;
-    }
-
-    // Starting here, at least protocols are equals.
-    // Compare hosts (domain:port) strictly
-    return std.mem.eql(u8, URL.getHost(url), URL.getHost(current_origin));
+    return URL.isSameOrigin(current_origin, url);
```
karlseguin: The parameters are `url, origin`, but here we pass `origin, url`. There are correctness things we should be doing which would cause this to break.

mookums:
Using it in `WorkerGlobalScope` now and fixed some port handling stuff so that those two FIXME tests pass in `Frame.zig`

PR #3192, opened 2026-08-13
Internal Requests shouldn't be interceptable

mookums:
Requests flagged as `internal` should not be intercepted. This currently just applies to Robots but will apply to CORS in the future as well.


karlseguin:
Did you test how Chrome behaves?

PR #3196, opened 2026-08-14
chore: Move cache maintenance to startup and browser.deinit

karlseguin:
--http-cache-entry-limit is currently implemented as a hard-limit via a check on every put. For a large --http-cache-entry-limit, this can result in a non- trivial delay on every http get.

This commit calls cache.maintenance() on Browser.deinit and on startup. maintenance:
1 - purges stale entries
2 - enforces the limit

With concurrent requests, the number of entries can easily exceed the limit by hundreds of entries (but this is temporary).

### DIFF
--- a/src/browser/Browser.zig
+++ b/src/browser/Browser.zig
@@ -154,6 +154,9 @@ pub fn deinit(self: *Browser) void {
     // fire — only now is it safe to free the pool backing their parameters.
     self.fc_identity_pool.deinit(allocator);
     self.page_pool.deinit(allocator);
+    if (self.http_client.cache) |cache| {
+        cache.maintenance(lp.datetime.timestamp(.real));
+    }
     self.http_client.deinit();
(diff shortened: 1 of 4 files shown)


mookums (approved):
I think this is fine as long as we're cool with a soft-limit.

mookums:
Might be out of scope for this PR but since we now have this maintenance path, it might be worth running `pragma optimize` during the shutdown pass. (https://sqlite.org/lang_analyze.html)

karlseguin:
> Might be out of scope for this PR but since we now have this maintenance path, it might be worth running `pragma optimize` during the shutdown pass. (https://sqlite.org/lang_analyze.html)

Claude seems to think that, as-is, it adds no value. The few queries we have have no ambiguity about how they execute. Also, the new purge SQL has a test that asserts the `cache_expiry` is being used.

PR #3201, opened 2026-08-14
`robots.txt` status handling

mookums:
This makes the `robots.txt` more spec compliant. On `400...499`, the `robots.txt` is considered unavailable and we are allowed to freely crawl the page. On `500...599`, the `robots.txt` is considered unreachable and we are completely disallowed.

### DIFF
--- a/src/network/RobotsGate.zig
+++ b/src/network/RobotsGate.zig
@@ -222,8 +222,16 @@ const RobotsContext = struct {
                     try network.robot_store.putAllowed(robots_url);
                 }
             },
+            // Unauthorized/Forbidden: treat as fully disallowed since we can't verify permissions.
+            401, 403 => {
+                log.debug(.http, "robots.txt access denied", .{
+                    .url = robots_url,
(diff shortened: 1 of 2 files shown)


src/network/RobotsGate.zig
```diff
@@ -201,26 +206,41 @@ const RobotsContext = struct {
...
+                    try network.robot_store.putAllowed(robots_url);
                 }
             },
-            404 => {
-                log.debug(.http, "robots not found", .{ .url = robots_url });
-                try network.robot_store.putAbsent(robots_url);
+            // RFC9309: Unavailable (400-499) means that we may access any resources on the server.
+            400...499 => {
+                log.debug(.http, "robots.txt unavailable", .{ .url = robots_url });
+                try network.robot_store.putAllowed(robots_url);
```
krichprollsch, line 236: I think it makes more sense for `401` and `403` to disallow.

krichprollsch (approved):

PR #3207, opened 2026-08-15
build: simplify build.zig

arrufat:
Cleanup pass over build.zig, no functional change to the build graph except where noted.

- Deduplicate the three executable blocks into an addExe helper. This also fixes the snapshot-creator and skills executables silently missing the sanitize_c/sanitize_thread flags the browser executable gets.
- Deduplicate the identical module setup for zlib, brotli, nghttp2 and curl into cLibModule.
- Report every prebuilt V8 cache miss with the probed path. A miss on the static archive path used to fall back to the 10 minute source build without a message, and the dev_fast message claimed a source build was starting when the source path may reuse a previous build. Also warn when the action.yml defaults stop parsing instead of silently disabling discovery.
- Drop vestigial !void on linkV8/linkCurl/linkHtml5Ever, flatten the lightpanda_module block, move linkSqlite next to the other link calls.
- Use b.fmt, b.pathJoin and std.mem.concat instead of hand-rolled equivalents; misc small forms (if x true else false, single-arm switch) and comment trims.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -24,15 +24,13 @@ const min_zig_version = std.SemanticVersion.parse(@import("build.zig.zon").minim
 
 const Build = blk: {
     if (builtin.zig_version.order(min_zig_version) == .lt) {
-        const message = std.fmt.comptimePrint(
+        @compileError(std.fmt.comptimePrint(
             \\Zig version is too old:
             \\  current Zig version: {f}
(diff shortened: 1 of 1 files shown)


karlseguin:
LGTM  (tested on linux and mac)

PR #3209, opened 2026-08-16
WebDriver BiDi

### LINKED ISSUE #576: selenium java support ?
is there any way to use this library on selenium java ?
if you give clue how to implement.
We can implement it on java for help

### LINKED ISSUE #1861: Add webdriver Bidi support
[Webdriver Bidi](https://www.w3.org/TR/webdriver2/) is a protocol based on websocket to remote control a browser. Similar to CDP.

* [Chrome article about webdriver bidi](https://developer.chrome.com/blog/webdriver-bidi/)
* [Specs](https://www.w3.org/TR/webdriver-bidi/)

⚠️ [Webdriver Bidi](https://www.w3.org/TR/webdriver-bidi/) is different from [Webdriver "classic"](https://www.w3.org/TR/webdriver2/), the previous HTTP only specs. This issue is about webdriver Bidi only. 

WPT runner is based to this protocol, so implementing it would allow us to run WPT the same as other browser

Wedriver bidi could enables Selenium support. Bidi is *not the default* form Selenium, it uses webdriver classic. But we can enable its support.

Playwright has a partial [webdriver bidi support](https://firefox-dev.tools/playwright-bidi-dashboard/) https://github.com/microsoft/playwright/issues/32577

Puppeteer has also a partial webdriver bidi support: https://pptr.dev/webdriver-bidi

karlseguin:
Adds support for connecting to websocket server using WebSocket BiDi. The driver to use is determined by the path, CDP and BiDi connections can be established side-by-side. Depends on https://github.com/lightpanda-io/zig-v8-fork/pull/200

The main change:
1 - Most things outside of CDP that were CDP-aware (e.g. HttpClient's `cdp_link_active: bool` field) are now Driver-aware. (`s/cdp/driver`)
2 - Driver is a struct that wraps a CDP or BiDi implementation
3 - CDP's NodeRegistry wasn't CDP-specific and was used in agent/mcp, so it's been moved to src/NodeRegistry.zig
4 - Reorg a little, src/Server.zig -> src/server/server.zig,  src/cdp -> src/server/cdp (and thus src/server/bidi)
5 - The BiDi implementation is basic, but should cover common cases (see the new demo cases)
6 - CDP metrics `cdp_*` renamed `serve_*` with a driver (cdp/bidi) label

bidi/input.zig was almost all done by claude.

100 iteration magic8ball
Firefox:
total duration (ms) 16113
avg run duration (ms) 159
min run duration (ms) 141
max run duration (ms) 792

lightpanda:
total duration (ms) 651
avg run duration (ms) 6
min run duration (ms) 5
max run duration (ms) 14

(about 1.75x faster than cdp for me).

Fix #576
Fix #1861


build.zig.zon
```diff
@@ -5,8 +5,8 @@
     .minimum_zig_version = "0.16.0",
     .dependencies = .{
         .v8 = .{
-            .url = "https://github.com/lightpanda-io/zig-v8-fork/archive/57747955a59f09147b4dc1142152f0d960b2c7e1.tar.gz",
-            .hash = "v8-0.0.0-xddH6x4GAwB6psyI4sllqMbK7cLkz-Bo3xC9WDheGksn",
+            .url = "https://github.com/lightpanda-io/zig-v8-fork/archive/d8843529988ff10acddbb1d86ef3747207e381de.tar.gz",
```
navidemad: The zig-v8-fork dep is bumped to `d8843529` (one commit past `v0.5.3`), which adds `v8__Value__IsMap/IsSet/IsWeakMap/IsWeakSet/IsRegExp/IsProxy/IsGeneratorObject/StrictEquals/ToBigInt` and `v8__Date__ValueOf` (used by `js/Value.zig` and `Local.newBigInt`), but the prebuilt tag in `.github/actions/install/action.yml` is still `v0.5.3` and no newer release exists. Any prebuilt build (`make build`, `make test`, CI) fails at link with 10 undefined symbols (`_v8__Value__IsMap`, `_v8__Date__ValueOf`, ...). Needs a zig-v8-fork release containing this commit and a tag bump before merge.

src/server/bidi/input.zig
```diff
@@ -0,0 +1,1090 @@
...
+                };
+            }
+
+            if (wait > 0) {
+                return frame.js.scheduler.add(self, resumeFromScheduler, wait, .{ .name = "bidi.performActions", .finalizer = cancelled });
+            }
+        }
+
+        defer self.deinit();
+        try bidi.sendResult(self.command_id, struct {}{});
```
navidemad, line 300: `Pending.deinit` is a bare `self.arena.release()`, and every terminating branch of `run()` registers `defer self.deinit()` *before* the `try bidi.sendResult`/`sendError`. If the send fails (peer closed, `Connection.send` returns `error.Closed`/EPIPE, or `sendJSON` OOM), the arena is released twice: on the synchronous path `performActions` still holds `errdefer arena.release()` (line 58); on the scheduler path `resumeFromScheduler` catches the error and calls `self.deinit()` again, dereferencing `self` (which lives inside that arena) after `ArenaPool.release` reset it. Debug hits `@panic("ArenaPool: double-free detected")`; release builds corrupt the pool free list. Same for the dispatch OOM branch at 265-268 and the `arena.release(); return cmd.sendError(...)` at 63-64. Suggest a `released` guard or dropping the errdefer/second deinit once `run()` owns the arena.

src/server/bidi/input.zig
```diff
@@ -0,0 +1,1090 @@
...
+        0xE05A => k.n("ArrowRight", "Numpad6"),
+        0xE05B => k.n("ArrowDown", "Numpad2"),
+        0xE05C => k.n("Insert", "Numpad0"),
+        0xE05D => k.n("Delete", "NumpadDecimal"),
+        else => k.k("Unidentified", ""),
+    };
+}
+
+// The `code` of a printable character on a US layout.
+fn asciiCode(cp: u21) []const u8 {
```
navidemad, line 754: `asciiCode` returns `&[_]u8{ 'K','e','y', std.ascii.toUpper(c) }` (and the `'A'...'Z'` / `'0'...'9'` arms) with a *runtime* `c`, so the array is a stack temporary and the returned slice dangles. `keyInfo` stores it in `KeyInfo.code`, and between that call (lines 514/523) and the read in `dispatchKey` (line 593 → `KeyboardEvent.initTrusted` dupes it) the caller runs `setModifier`, `pressed.append`, `utf8Encode` etc., which reuse the same stack region. Reproduced standalone: `code` reads as `0xAA...` in Debug/ReleaseSafe and random bytes in ReleaseFast. Tests pass only because none asserts `event.code`. Fix: comptime tables (`"KeyA".."KeyZ"`, `"Digit0".."Digit9"`) or a caller-owned buffer.

src/server/Server.zig
```diff
@@ -190,91 +188,103 @@ fn handleConnection(self: *Server, socket: posix.socket_t) void {
     defer _ = self.active_threads.fetchSub(1, .monotonic);
     defer _ = std.c.close(socket);
 
+    const route = Handshake.run(self.app, socket, self.json_version_response) orelse return;
```
navidemad: Regression vs `main`: the old `handleConnection` appended the cdp to `self.cdps` *before* `conn.handshake()`, and `Server.shutdown()` called `conn.shutdown()` on every entry, so a socket blocked in the handshake `poll` woke immediately on SIGINT. Now `Handshake.run` runs before `serve()` calls `self.track(driver)` (line 257), so pre-upgrade sockets are invisible to `shutdown()`, and `Handshake.run`'s loop (`Handshake.zig:59-83`) polls with a fresh 5 s budget per received byte and never checks `app.shutdown()`. `Server.deinit()` then spins in `while (active_threads > 0)` for up to 5 s per idle pre-upgrade connection and unboundedly for a client trickling one byte every <5 s. Either track the socket before the handshake (as before) or give `Handshake.run` a total deadline + `app.shutdown()` check.

src/server/bidi/input.zig
```diff
@@ -0,0 +1,1090 @@
...
+    };
+
+    // dispatch pops the released key/button off `pressed` itself
+    const state = &bidi.input_state;
+    for (state.sources.items) |*source| {
+        switch (source.kind) {
+            .key => while (source.key.pressed.getLastOrNull()) |cp| {
+                try dispatch(bidi, frame, source, &.{ .key_up = cp });
+            },
+            .pointer => while (source.pointer.pressed.getLastOrNull()) |button| {
```
navidemad, line 109: `releaseActions` uses `try dispatch(...)` for the synthesized keyup/mouseup, so a `DispatchError` (`JsException`/`CompilationError` from a throwing keyup/mouseup/click listener — see `EventManagerBase.DispatchError`) escapes the handler. `BiDi.onMessage` only maps `UnknownCommand`/`InvalidParams`; anything else is `return err` and `HttpClient.drainInbox` just logs "client dispatch", so the client never receives a response for this command id, and `state.reset(...)` at line 107 is skipped so the not-yet-popped keys/buttons persist into the next `performActions`. `Pending.run` handles the same errors via `errorCode()/errorMessage()` → `sendError`; `releaseActions` should do the same.

src/server/bidi/remote_value.zig
```diff
@@ -0,0 +1,756 @@
...
+        TypeError,
+        JsException,
+        ExecutionTerminated,
+        MethodNotFound,
+        DeadFunctionHandle,
+    };
+
+    fn remote(self: *Serializer, value: js.Value, depth: u32, own: bool) Error!Remote {
+        const handle: ?[]const u8 = if (own and value.isObject()) blk: {
+            const id = try self.handles.add(value);
```
navidemad, line 344: With `resultOwnership: "root"`, `handles.add(value)` persists the global and inserts it into `Handles.map` *before* `self.body(value, depth)` runs, and there is no `errdefer self.handles.release(id)`. `body()` can genuinely fail with a JS exception (`properties()` → `Object.get` returns `error.JsException` for a throwing accessor / Proxy `get` trap; `nameIterator` → `TypeError`; the `toISOString` fallback in `date()`). `script.zig:serialize` then replies `"unknown error" / "cannot serialize result"` and the client never learns the id, so the global lives (keeping the object alive) until `resetRealm` → `releaseAll()`. Add the `errdefer` or serialize the body first.

src/server/bidi/remote_value.zig
```diff
@@ -0,0 +1,756 @@
...
+        }
+        if (value.isNull()) {
+            return .null;
+        }
+        if (value.isBigInt()) {
+            return .{ .bigint = try value.toStringSliceWithAlloc(self.arena) };
+        }
+        if (value.isObject() == false) {
+            // A symbol. BiDi has no type for one.
+            return .{ .other = "object" };
```
navidemad, line 382: A JS Symbol lands here (`Value.isSymbol` exists but there is no branch for it) and is emitted as `{"type":"object"}` with no `value`. The spec has `script.SymbolRemoteValue = { type: "symbol", handle?, internalId? }`, so a conformant client mis-types the result; and because `isObject()` is false a symbol never gets a handle even with `resultOwnership: "root"`. One-liner: `.other = "symbol"` (and consider allowing a handle for it).

src/network/HttpClient.zig
```diff
@@ -1322,6 +1325,9 @@ fn isSyncWaitInterrupt(msg: *Inbox.Message) bool {
         .close, .disconnect => true,
         .ping => false,
         .cdp => |c| isTeardownMethod(c.input.method),
+        // Frames aren't parsed on the Network thread for BiDi, so we
+        // can't spot a teardown command without re-parsing here.
+        .bidi => false,
```
navidemad, line 1390: `allowDuringSyncWait`/`isSyncWaitInterrupt` both hard-code `.bidi => false`, so `hasPendingTeardown()` is always false for BiDi. During a blocking sync wait (`syncRequest` — every blocking `<script src>` during parsing, or sync XHR — against a slow host) a CDP `Target.closeTarget`/`Page.close` interrupts the wait, but BiDi `browsingContext.close` / `browser.close` / `session.end` sit undispatched (they only push `.close` from *inside* their handlers) until the transfer finishes or `http_timeout` fires. Only closing the WebSocket itself interrupts. This also puts CDP method names in the generic HTTP client; parsing the `{id, method}` envelope in `Connection.pushBiDi` (like `pushCdp`) and letting the driver answer `isTeardown(method)` would fix both drivers at the right layer.

build.zig.zon
karlseguin: It's updated now. We tend to wait to see if we can group more things together.

src/server/Server.zig
karlseguin: Done, but hoping to improve this further in a future commit.

src/network/HttpClient.zig
karlseguin: Known limitation for this first cut.

Issue #3156, opened 2026-08-06
Lightpanda clients are currently attacking Haskell infrastructure


Xe:
The Haskell infrastructure team is currently battling a flood of Lightpanda clients scraping every link on their GitLab server, which has `robots.txt` configured to block them. Why is this not working as expected and why is respecting robots.txt disabled by default?

karlseguin:
Hi. Sorry for the problems this has caused.

```
curl "https://gitlab.haskell.org/robots.txt" -H "User-Agent: Lightpanda/1.0"
```

doesn't return a valid robots.txt, it's returning an anubis bot checker (basically a normal HTML page). We try to parse, fail, and default to allow (which I do think is reasonable).. When I tested locally with your actual robots.txt AND `--obey-robots`, it was properly rejected (at least on the few paths that I tried).

@francisbouvier might have more insight, but at least one reason --obey-robots isn't on by default is that robots support is relatively new and we didn't want to make a breaking change for existing users. I know that isn't a great answer, and we've been talking about changing a few other defaults (unrelated)..so the best I can do is promise to bring it up.

MagicRB:
Bit tangential, but i find it very concerning and dishonest that lightpanda does not respect robots.txt. if anything there should be no way for it to not do that. I personally ban all known scrapers using robots.txt and run iocaine on my domain to fend off any that do appear. I would very much appreciate, as one human to another, if my website wasn't blasted by LLM tooling and my hard work wasn't fed into an LLM without my permission.

Thank you

AndreasPK:
> curl "https://gitlab.haskell.org/robots.txt" -H "User-Agent: Lightpanda/1.0"

This is because we were forced to blacklist the user agent due to excessive load from requests using the lightpanda User-Agent.

MangoIV:
> We try to parse, fail, and default to allow (which I do think is reasonable)..

Even if the client *didn't* have access to it until yesterday which it did (see above comment by Andreas), I still want to contest this. 

To conclude from the absence of an explicit denial to do so that it is fine to just wreak havoc on someones infrastructure is a completely absurd understanding of consent to me.

You don't usually have to put a big sign on your car that says "you are not welcome to break the windows and break off the mirrors" in order for you to expect to not get your car destroyed.

MangoIV:
[...]

PR #3212, opened 2026-08-17
feat: obey robots.txt by default

### LINKED ISSUE #3208: Lightpanda clients abusing Gentoo Linux infrastructure - lightpanda ignoring robots.txt completely
Hi,

Similar to #3156 - Lightpanda UA traffic sent to Gentoo Linux's systems is extremely aggressive.
Traffic to gitweb.gentoo.org is one of the worst examples - more than 1M unique IPs/day, many on very expensive queries selecting part of Git history or diffs between random ranges.

Lightpanda entirely ignores the robots.txt
```
$ curl "https://gitweb.gentoo.org/robots.txt" -H "User-Agent: Lightpanda/1.0" 
User-agent: *
Disallow: /
```

If the lightpanda users are finding that much value from Gentoo; they need to contribute back to sustain the infrastructure, or will not be possible to sustain them. 
https://www.gentoo.org/donate/

We've implemented short-term CDN protections, but these clients need to be much better behaved.

pasmud:
Fixes #3208 and #3156: Lightpanda ignored `robots.txt` unless `--obey-robots` was passed, letting crawls run against sites that explicitly disallow them (Gentoo, Haskell infra were hit hard).

The robots.txt gate was already fully implemented and correct; the only problem was the default. This change flips the option so robots.txt is respected out of the box in `fetch`, `serve`, `mcp`, and `agent` modes.

## Changes

- `src/Config.zig`: `obey_robots` now defaults to `true`. The flag becomes `--no-obey-robots` to opt out; `--obey-robots` is still accepted as an explicit opt-in (backward compatible).
- `src/help.zon`: updated `--no-obey-robots` help text and default.
- `README.md`: examples no longer pass `--obey-robots`; documented the new default and the opt-out flag.
- `src/cdp/domains/lp.zig`: comment updated to reflect the new default.

## Verification

- `zig fmt --check` passes on all changed files.
- A full `zig build test` could not run in this environment: the build fetches V8 and other dependencies from GitHub, which the sandbox network proxy rejects (HTTP fetch fails with `HttpConnectionClosing`/`TemporaryNameServerFailure`). The change is confined to the declarative CLI option table plus docs; the CLI framework's bool semantics (presence flips the field opposite its default; variants without a default flip to true) were verified against `src/cli.zig`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,8 +102,11 @@ docker run -d --name lightpanda -p 127.0.0.1:9222:9222 lightpanda/browser:nightl
 ### Dump a URL
 
 ```console
-./lightpanda fetch --obey-robots --dump html --log-format pretty  --log-level info https://demo-browser.lightpanda.io/campfire-commerce/
+./lightpanda fetch --dump html --log-format pretty  --log-level info https://demo-browser.lightpanda.io/campfire-commerce/
 ```
+
+Lightpanda fetches and obeys each site's `robots.txt` by default. Pass
+`--no-obey-robots` to opt out.
 
 You can use `--dump markdown` to convert directly into markdown.
 `--wait-until`, `--wait-ms`, `--wait-selector` and `--wait-script` are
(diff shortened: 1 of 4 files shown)


pasmud:

Xe (approved):

jappeace (approved):
what a good idea!

t4ccer (approved):
Thanks for the fix!

robbat2 (approved):

PR #3217, opened 2026-08-18
fix: pop drive-letter lookalike segment for non-file schemes (#2794)

### LINKED ISSUE #2794: url-constructor: '..' against a non-special base must pop the last path segment

pasmud:
Fixes #2794: `new URL('..', 'abc://x/y/z/C:/')` must resolve to `abc://x/y/z/` but rust-url refuses to pop the trailing `C:` segment.

## Root cause

rust-url 2.5.8's `last_slash_can_be_removed` applies the Windows-drive-letter exemption to **every** scheme when shortening a path with `..`, but per the WHATWG URL Standard that exemption only applies to `file:`. For any other scheme a `C:`/`C|` segment is ordinary and must be popped. The upstream fix (servo/rust-url#1138) is open but unmerged, so this adds an in-repo workaround.

## Change

- `src/html5ever/url.rs`: new `fix_drive_letter_join` helper, called from all four join/resolve FFI entry points (`url_parse_with_base`, `url_join`, `url_resolve_without_encoding`, `url_resolve_with_encoding`). When the base path ends with a drive-letter lookalike segment (`/X:/` or `/X|/`) on a non-file scheme and the relative input starts with `..`, it pops that segment from the joined URL.
- `src/browser/URL.zig`: Zig test for the resolve path covering the WPT case plus the http variant, the `C|` form, control segments, the no-trailing-slash case, and the `file:` guard.

## Verification

- `cargo test --lib` in `src/html5ever`: 4/4 pass (join cases, file guard, parse-with-base path, lookalike matcher).
- `zig fmt --check src/browser/URL.zig`: clean.
- Full `zig build test` could not run in this sandbox: the build fetches V8 + dependencies from GitHub, which the sandbox proxy/DNS rejects (`HttpConnectionClosing`/`TemporaryNameServerFailure`). The behavior was verified end-to-end via the Rust crate tests plus a raw probe of rust-url (`abc://x/y/z/C:/ + ..` returns the buggy `abc://x/y/z/C:/` before the fix, `abc://x/y/z/` after).

### DIFF
--- a/src/browser/URL.zig
+++ b/src/browser/URL.zig
@@ -1013,6 +1013,58 @@ test "URL: resolve validates ASCII punycode (xn--) labels" {
     try testing.expectError(error.TypeError, resolve(testing.arena_allocator, "https://example.com/", "https://xn--a.pt/x", .{}));
 }
 
+test "URL: resolve pops drive-letter lookalike segment for non-file schemes (#2794)" {
+    const Case = struct {
+        base: [:0]const u8,
+        path: [:0]const u8,
+        expected: [:0]const u8,
+    };
+
+    const cases = [_]Case{
+        // A "C:" segment is only a Windows drive letter for file: URLs; for any
+        // other scheme ".." must pop it as an ordinary segment.
+        .{
+            .base = "abc://x/y/z/C:/",
+            .path = "..",
+            .expected = "abc://x/y/z/",
+        },
+        // Special (but non-file) scheme hits the same path.
+        .{
+            .base = "http://x/y/z/C:/",
+            .path = "..",
+            .expected = "http://x/y/z/",
+        },
+        // The "C|" (pipe) form is affected too.
+        .{
+            .base = "abc://x/y/z/C|/",
+            .path = "..",
+            .expected = "abc://x/y/z/",
+        },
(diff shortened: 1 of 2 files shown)


krichprollsch:
Thanks for your contribution.
I saw there is also an open PR on rust-url to fix the issue: https://github.com/servo/rust-url/pull/1138
What about waiting for the merge?

nikneym:
Hi & thank you for your contribution. We were actually hoping if https://github.com/servo/rust-url/pull/1138 be merged so I'd like to keep this open for a little more and take action if it won't be fixed in the upstream.

nikneym (approved):

PR #3222, opened 2026-08-18
webapi: Re-parse element.style when style attribute changes

karlseguin:
If we have a materialized element.style (stored in Frame._element_styles) and the style attribute changes (e.g. via setAttribute('style', '....'), then the materialized CSSStyleDeclaration has to be updated (cleared and the style re-parsed)

### DIFF
--- a/src/browser/Frame.zig
+++ b/src/browser/Frame.zig
@@ -2964,6 +2964,8 @@ pub fn attributeChange(self: *Frame, element: *Element, name: String, value: Str
     } else if (name.eql(comptime .wrap("popover"))) {
         const old = if (old_value) |o| o.str() else null;
         popover.attributeChanged(element, old, value.str(), self);
+    } else if (name.eql(comptime .wrap("style"))) {
+        self.styleAttributeChanged(element, value.str());
     }
 }
 
@@ -2985,7 +2987,16 @@ pub fn attributeRemove(self: *Frame, element: *Element, name: String, old_value:
         }
     } else if (name.eql(comptime .wrap("popover"))) {
         popover.attributeChanged(element, old_value.str(), null, self);
+    } else if (name.eql(comptime .wrap("style"))) {
+        self.styleAttributeChanged(element, null);
     }
+}
+
+fn styleAttributeChanged(self: *Frame, element: *Element, value: ?[]const u8) void {
+    const style = self._element_styles.get(element) orelse return;
+    style.asCSSStyleDeclaration().styleAttributeChanged(value, self) catch |err| {
+        log.err(.frame, "style attribute reparse", .{ .err = err, .type = self._type, .url = self.url });
+    };
 }
 
 pub fn signalSlotChange(self: *Frame, slot: *Element.Html.Slot) void {
--- a/src/browser/webapi/css/CSSStyleDeclaration.zig
+++ b/src/browser/webapi/css/CSSStyleDeclaration.zig
@@ -34,6 +34,7 @@ const CSSStyleDeclaration = @This();
 _element: ?*Element = null,
 _properties: std.DoublyLinkedList = .{},
 _is_computed: bool = false,
+_syncing: bool = false,
 
 // Parse the element's existing style attribute into _properties so that
 // subsequent JS reads and writes see all CSS properties, not just newly
@@ -43,11 +44,14 @@ pub fn parseInlineStyle(self: *CSSStyleDeclaration, frame: *Frame) !void {
         return;
     }
     const el = self._element orelse return;
-    if (el.getAttributeSafe(comptime .wrap("style"))) |attr_value| {
-        var it = CssParser.parseDeclarationsList(attr_value);
-        while (it.next()) |declaration| {
-            try self.applyParsedDeclaration(declaration, frame);
-        }
+    const attr_value = el.getAttributeSafe(comptime .wrap("style")) orelse return;
+    try self.applyDeclarations(attr_value, frame);
+}
+
+fn applyDeclarations(self: *CSSStyleDeclaration, text: []const u8, frame: *Frame) !void {
+    var it = CssParser.parseDeclarationsList(text);
+    while (it.next()) |declaration| {
+        try self.applyParsedDeclaration(declaration, frame);
     }
 }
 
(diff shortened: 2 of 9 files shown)


krichprollsch (approved):

src/browser/Frame.zig
```diff
@@ -2985,9 +2987,18 @@ pub fn attributeRemove(self: *Frame, element: *Element, name: String, old_value:
         }
     } else if (name.eql(comptime .wrap("popover"))) {
         popover.attributeChanged(element, old_value.str(), null, self);
+    } else if (name.eql(comptime .wrap("style"))) {
+        self.styleAttributeChanged(element, null);
     }
 }
 
+fn styleAttributeChanged(self: *Frame, element: *Element, value: ?[]const u8) void {
+    const style = self._element_styles.get(element) orelse return;
```
navidemad, line 2996: The resync only reaches the *owner* frame's `_element_styles`, but `Element.getOrCreateStyle` (Element.zig:973) stores the declaration in the *caller's* `frame._element_styles`. `Attribute._put`/`_delete` dispatch `attributeChange` on `element.ownerFrame(frame)`, so for an element in another same-origin realm the lookup here misses and `.style` stays stale.

Reproduced with a probe under `tests/frames/` (iframe from `support/cross_realm_focus.html`):

```js
const b = idoc.body;
b.style.color = 'red';                 // declaration lands in the parent's map
b.setAttribute('style', 'color: blue'); // owner = iframe frame -> map miss
b.style.color                          // still 'red' (expected 'blue')
```

The next parent-side `b.style.x = ...` then re-serializes the stale list and overwrites the iframe's attribute value. Scroll accessors already resolve `ownerFrame` for exactly this reason (Element.zig:1459 comment); `getOrCreateStyle`/`getStyle` and `StyleManager.getInlineStyleProperty` could do the same so all three agree on one map.

src/browser/webapi/css/CSSStyleDeclaration.zig
```diff
@@ -213,9 +217,36 @@ fn removePropertyImpl(self: *CSSStyleDeclaration, property_name: []const u8, fra
 fn syncStyleAttribute(self: *CSSStyleDeclaration, frame: *Frame) !void {
     const element = self._element orelse return;
     const css_text = try self.getCssText(frame);
+    self._syncing = true;
+    defer self._syncing = false;
     try element.setAttributeSafe(comptime .wrap("style"), .wrap(css_text), frame);
 }
 
+// The element's style attribute changed (null: removed)
+pub fn styleAttributeChanged(self: *CSSStyleDeclaration, text: ?[]const u8, frame: *Frame) !void {
```
navidemad, line 227: Minor: by the time `attributeChange`/`attributeRemove` fire, the entry is already updated (`_put` calls `e.setValue` first) or already removed (`_delete`), so the `text` parameter carries nothing the element doesn't have. Re-reading the attribute through `parseInlineStyle` drops the `?[]const u8` plumbing (its `orelse return` covers the removed case), and the two Frame.zig call sites collapse to `self.styleAttributeChanged(element)` (suggestions posted there too).

```suggestion
// The element's style attribute changed or was removed: re-read it
pub fn styleAttributeChanged(self: *CSSStyleDeclaration, frame: *Frame) !void {
    if (self._syncing) {
        // this was us making the change internally, automatically in-sync
        return;
    }
    // ok, this change was from the outside, e.g. via setAttribute('style', ....)
    // we need to get this declaration back in sync
    self.clearProperties(frame);
    try self.parseInlineStyle(frame);
}
```

src/browser/Frame.zig
```diff
@@ -2964,6 +2964,8 @@ pub fn attributeChange(self: *Frame, element: *Element, name: String, value: Str
     } else if (name.eql(comptime .wrap("popover"))) {
         const old = if (old_value) |o| o.str() else null;
         popover.attributeChanged(element, old, value.str(), self);
+    } else if (name.eql(comptime .wrap("style"))) {
+        self.styleAttributeChanged(element, value.str());
```
navidemad, line 2968: Goes with the `styleAttributeChanged` suggestion in CSSStyleDeclaration.zig (attribute entry is already updated here, no need to pass the value).

```suggestion
        self.styleAttributeChanged(element);
```

src/browser/Frame.zig
```diff
@@ -2985,9 +2987,18 @@ pub fn attributeRemove(self: *Frame, element: *Element, name: String, old_value:
         }
     } else if (name.eql(comptime .wrap("popover"))) {
         popover.attributeChanged(element, old_value.str(), null, self);
+    } else if (name.eql(comptime .wrap("style"))) {
+        self.styleAttributeChanged(element, null);
```
navidemad, line 2991: Same as above (entry already removed when this fires).

```suggestion
        self.styleAttributeChanged(element);
```

src/browser/Frame.zig
```diff
@@ -2985,9 +2987,18 @@ pub fn attributeRemove(self: *Frame, element: *Element, name: String, old_value:
...
     }
 }
 
+fn styleAttributeChanged(self: *Frame, element: *Element, value: ?[]const u8) void {
+    const style = self._element_styles.get(element) orelse return;
+    style.asCSSStyleDeclaration().styleAttributeChanged(value, self) catch |err| {
+        log.err(.frame, "style attribute reparse", .{ .err = err, .type = self._type, .url = self.url });
+    };
+}
```
navidemad, line 3000: Same as above: the declaration re-reads the attribute itself.

```suggestion
fn styleAttributeChanged(self: *Frame, element: *Element) void {
    const style = self._element_styles.get(element) orelse return;
    style.asCSSStyleDeclaration().styleAttributeChanged(self) catch |err| {
        log.err(.frame, "style attribute reparse", .{ .err = err, .type = self._type, .url = self.url });
    };
}
```

src/browser/webapi/css/CSSStyleDeclaration.zig
karlseguin: I prefer to pass in the value from the attributeChange. Prevents a lookup and, imo, is more explicit.

src/browser/Frame.zig
karlseguin: Yes, that's an issue. I've updated callsites of the StyleManager to get the element's ownerFrame.

Issue #3179, opened 2026-08-12
Anchor clicks fire handlers but don't run the default action (no navigation)


katie-lpd:
Click events dispatch and bubble, and handlers attached to an `<a>` run, but the default action never does. Clicking a link leaves `location.href` unchanged regardless of how the click is produced. Navigation itself works: `Page.navigate` and a JS `location.href` assignment both load the target. The missing piece is the activation step between a click on `<a href>` and the navigation it should start.
 
## Environment
 
- Lightpanda nightly `1.0.0-nightly.8589+6312d43d`, Linux x86_64
- `lightpanda serve --host 127.0.0.1 --port 9222`
- Reproduced over CDP with plain `Runtime.evaluate` and `Input.dispatchMouseEvent`, so any CDP client hits it. Surfaced while validating Lightpanda under the Browser Use harness / Hermes `browser_exec`.
 
## Repro
 
On `https://example.com`, attach a click listener to the only link, then click it three ways:
 
1. `element.click()`
2. a synthesized `dispatchEvent(new MouseEvent("click", {bubbles:true, cancelable:true}))`
3. CDP `Input.dispatchMouseEvent` press + release at the anchor box center
 
After each: the listener flag is set (`1`) but `location.href` is still `https://example.com/`. No handler calls `preventDefault()`.
 
## Expected
 
`location.href` becomes the link target after any of these clicks.
 
## Verified working (scope narrowing)
 
- `Input.dispatchMouseEvent` hit-tests correctly and fires handlers (confirmed on a sized `position:fixed` button and on the anchor).
- `document.elementFromPoint` returns the anchor at its box center.
- `js("location.href=...")` navigates; `Page.navigate` navigates.
 
So the input pipeline up to handler dispatch is fine. This is isolated to default-action handling on activation. Only anchors were tested; form submission via a submit-button click may share the path and is worth checking together.
 
## Why it matters
 
Code-mode agent harnesses follow links by clicking them. On Lightpanda the click lands and nothing happens, so the agent has to discover the `location.href` workaround by trial and error. Fixing activation lets the default harness workflow run unmodified.'

krichprollsch:
I can't reproduce the issue. Did you wait until the navigation is done? If you check href immediately after the click, you could be in a state where the navigation is still in progress.

PR #3223, opened 2026-08-18
test: lock in anchor click default navigation (#3179)

pasmud:
Adds a regression test for #3179: clicking an `<a href>` must run the default action (navigation) even when the anchor has a click handler that does not call `preventDefault()`, and a handler that calls `preventDefault()` must suppress it.

## Investigation

The reporter's symptom (handlers run but `location.href` unchanged) was against nightly `8589`; the anchor activation path has since been fully implemented (`findClickActivationTarget` + `handleClick` + `followLink` in `src/browser/frame/user_input.zig`, with the spec-conformant click activation target selection from July). The existing `cdp.frame: anchor click sends Referer` test already proves `element.click()` on an anchor navigates. So the core defect is fixed on current main; what was missing was coverage for the reporter's exact scenario (an anchor *with* a click handler).

## Change

- `src/cdp/domains/page.zig`: new test `cdp.frame: anchor click runs the default navigation unless preventDefault (#3179)` that clicks an anchor with a non-preventing handler and asserts navigation commits, then clicks an anchor whose handler calls `preventDefault()` and asserts navigation is suppressed. Driven via `element.click()`, the same path CDP `Input.dispatchMouseEvent` release uses.
- `src/testing.zig`: `activation.html` fixture with the two anchors.

## Verification

- `zig fmt --check` clean on both changed files.
- `zig ast-check` clean on both changed files.
- Full `zig build test` could not run in this sandbox: the build fetches V8 and dependencies from GitHub, which the sandbox proxy/DNS rejects (`HttpConnectionClosing` / `TemporaryNameServerFailure`). The change adds no production code, only a fixture and a regression test modeled on the existing `#3179`-adjacent referer test.

### DIFF
--- a/src/cdp/domains/page.zig
+++ b/src/cdp/domains/page.zig
@@ -1687,6 +1687,80 @@ test "cdp.frame: anchor click sends Referer matching the originating page" {
     }
 }
 
+test "cdp.frame: anchor click runs the default navigation unless preventDefault (#3179)" {
+    // Regression test for #3179. Clicking an <a href> must run the default
+    // action (navigation) even when the anchor has a click handler that does
+    // NOT call preventDefault. A handler that calls preventDefault() must
+    // suppress the navigation. The two behaviors are driven by element.click(),
+    // the same path a CDP Input.dispatchMouseEvent release uses.
+    var ctx = try testing.context();
+    defer ctx.deinit();
+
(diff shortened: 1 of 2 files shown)


karlseguin:
I think it's better in our demo/integration test project:

https://github.com/lightpanda-io/demo/pull/228

PR #3230, opened 2026-08-19
Introduce `--load-resources`, supporting `image` parameter

nikneym:
This PR adds `--load-resources` CLI argument. It can take `image` parameter to actually request images over network for their existence, verified by HTTP response. In order to reuse connections on HTTP/1.1, we read the response body if it includes `Content-Length` header with value that's <= `HEADERS_ONLY_DRAIN_MAX` (16 KiB), which consumes more bandwidth but performs better than opening a socket, doing TLS handshake etc. steps.

### DIFF
--- a/src/Config.zig
+++ b/src/Config.zig
@@ -211,6 +211,26 @@ fn caPathValidator(
     }
 }
 
+// Sub-resources the browser actually requests. A comma-separated list so
+// more can be opted into individually as they land; only `image` fetches
+// today.
+pub const LoadResources = packed struct(u1) {
+    image: bool = false,
+};
+
+fn loadResourcesValidator(_: Allocator, args: *std.process.Args.Iterator, target: *LoadResources) !void {
+    const resources = args.next() orelse return error.MissingArgument;
+
+    var it = std.mem.splitScalar(u8, resources, ',');
+    while (it.next()) |resource| {
+        inline for (@typeInfo(LoadResources).@"struct".fields) |field| {
+            if (std.mem.eql(u8, field.name, resource)) {
+                @field(target, field.name) = true;
+            }
+        }
+    }
+}
+
 /// Common CLI args.
 const CommonOptions = .{
     .{ .name = "obey_robots", .type = bool },
@@ -245,6 +265,12 @@ const CommonOptions = .{
     .{ .name = "disable_subframes", .type = bool },
     .{ .name = "disable_workers", .type = bool },
     .{ .name = "enable_external_stylesheets", .type = bool },
+    .{
+        .name = "load_resources",
+        .type = LoadResources,
+        .default = LoadResources{},
+        .validator = loadResourcesValidator,
+    },
     .{ .name = "v8_flags_unsafe", .type = ?[]const u8 },
     .{ .name = "v8_max_heap_mb", .type = ?u32 },
     .{ .name = "watchdog_ms", .type = ?u32 },
@@ -522,6 +548,13 @@ pub fn watchdogMs(self: *const Config) ?u32 {
 pub fn enableExternalStylesheets(self: *const Config) bool {
     return switch (self.mode) {
         inline .serve, .fetch, .mcp, .agent => |opts| opts.enable_external_stylesheets,
+        else => unreachable,
+    };
+}
+
+pub fn loadResources(self: *const Config) LoadResources {
+    return switch (self.mode) {
+        inline .serve, .fetch, .mcp, .agent => |opts| opts.load_resources,
         else => unreachable,
     };
 }
--- a/src/Config.zig
+++ b/src/Config.zig
@@ -211,25 +211,9 @@ fn caPathValidator(
     }
 }
 
-// Sub-resources the browser actually requests. A comma-separated list so
-// more can be opted into individually as they land; only `image` fetches
-// today.
 pub const LoadResources = packed struct(u1) {
     image: bool = false,
 };
-
-fn loadResourcesValidator(_: Allocator, args: *std.process.Args.Iterator, target: *LoadResources) !void {
-    const resources = args.next() orelse return error.MissingArgument;
-
-    var it = std.mem.splitScalar(u8, resources, ',');
-    while (it.next()) |resource| {
-        inline for (@typeInfo(LoadResources).@"struct".fields) |field| {
-            if (std.mem.eql(u8, field.name, resource)) {
-                @field(target, field.name) = true;
-            }
-        }
-    }
-}
 
 /// Common CLI args.
 const CommonOptions = .{
@@ -265,12 +249,7 @@ const CommonOptions = .{
     .{ .name = "disable_subframes", .type = bool },
     .{ .name = "disable_workers", .type = bool },
     .{ .name = "enable_external_stylesheets", .type = bool },
-    .{
-        .name = "load_resources",
-        .type = LoadResources,
-        .default = LoadResources{},
-        .validator = loadResourcesValidator,
-    },
+    .{ .name = "load_resources", .type = LoadResources, .default = LoadResources{} },
     .{ .name = "v8_flags_unsafe", .type = ?[]const u8 },
     .{ .name = "v8_max_heap_mb", .type = ?u32 },
     .{ .name = "watchdog_ms", .type = ?u32 },
--- a/src/browser/Frame.zig
+++ b/src/browser/Frame.zig
@@ -2384,6 +2384,155 @@ pub fn loadExternalStylesheet(self: *Frame, link: *Element.Html.Link, href: []co
     try self.fireElementEvent(element, comptime .wrap("load"));
 }
 
+// Asynchronous, unlike `loadExternalStylesheet`: nothing in the document
+// depends on the result, so there's no reason to block the parser on it.
+// The request does take a `_pending_loads` slot, so the window load event
+// waits for it the way the HTML spec says it should.
+pub fn loadImage(self: *Frame, image: *Element.Html.Image, src: []const u8) !void {
+    const session = self._session;
+    // Fragment-parsed images (innerHTML, DOMParser, ...) may never be
+    // attached, and may belong to another Document. Same call as
+    // `loadExternalStylesheet` makes. They still get the synthetic load the
+    // no-fetch path would have given them, so the two modes agree.
+    if (self._parse_mode == .fragment) {
+        return self.queueLoad(Factory.protoOf(image));
+    }
+
+    const arena = try session.getArena(.small, "Frame.loadImage");
+    defer arena.release();
+
+    const resolved = URL.resolve(arena.allocator(), self.base(), src, .{ .encoding = self.charset }) catch |err| {
+        // An unresolvable src is a load failure the same way a 404 is, and
+        // unlike a real fetch we can settle it without a request. No pending
+        // load was taken out yet, so queue rather than decrementing.
+        log.info(.http, "image resolve", .{ .err = err, .src = src });
+        image._complete = true;
+        return self.queueElementEvent(Factory.protoOf(image), .@"error");
+    };
+
+    // an in-flight image is a fixed-size record we can hand back the moment it
+    // settles, and a page can create a lot of them.
+    const load = try self._factory.create(ImageLoad{
+        .frame = self,
+        .image = image,
+        .generation = image._generation,
+    });
+    errdefer self._factory.destroy(load);
+
+    // New load event always sets `_complete` back to false.
+    image._complete = false;
+    self._pending_loads += 1;
+    errdefer {
+        image._complete = true;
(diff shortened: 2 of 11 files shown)


src/browser/webapi/element/html/Image.zig
```diff
@@ -1,16 +1,27 @@
 const lp = @import("lightpanda");
 const std = @import("std");
+const log = lp.log;
```
karlseguin: Should be lower, grouped with the other non-imports:

```
const log = lp.log;
const String = lp.String;
```

src/help.zon
```diff
@@ -346,6 +346,19 @@
     \\          contribute to computed styles (and therefore to visibility checks like
     \\          display, visibility, opacity, pointer-events).
     \\          Defaults to false, except in agent mode with an LLM, where it is on.
+    \\  --load-resources <LIST>
+    \\          Comma-separated list of sub-resources to actually request.
```
karlseguin: https://github.com/lightpanda-io/browser/pull/3221 removed all mention of comma-separated lists in favor of specifying the parameter multiple times.

src/Config.zig
```diff
@@ -211,6 +211,26 @@ fn caPathValidator(
...
+// more can be opted into individually as they land; only `image` fetches
+// today.
+pub const LoadResources = packed struct(u1) {
+    image: bool = false,
+};
+
+fn loadResourcesValidator(_: Allocator, args: *std.process.Args.Iterator, target: *LoadResources) !void {
+    const resources = args.next() orelse return error.MissingArgument;
+
+    var it = std.mem.splitScalar(u8, resources, ',');
```
karlseguin: In addition to moving away from *documenting* comma-separated values, I think we should just not support it here. Also, this doesn't seem to behave consistently with existing flags:

```
$ ... --strip-mode invalid
FATAL app : invalid option choice . . . . . . . . . . . . . .  [+0ms]
      arg = --strip-mode
      value = invalid
```

vs

```
$ ... --load-resources invalid
(no error)
```

src/cdp/domains/lp.zig
```diff
@@ -20,6 +20,7 @@ const std = @import("std");
 const lp = @import("lightpanda");
 
 const CDP = @import("../CDP.zig");
+const Config = @import("../../Config.zig");
```
karlseguin: Not used.

src/browser/Frame.zig
```diff
@@ -2376,6 +2376,154 @@ pub fn loadExternalStylesheet(self: *Frame, link: *Element.Html.Link, href: []co
...
+        if (!superseded) {
+            image._complete = true;
+        }
+
+        if (!superseded and !frame.isGoingAway()) {
+            const name: String = switch (kind) {
+                .load => comptime .wrap("load"),
+                .@"error" => comptime .wrap("error"),
+            };
+            frame.fireElementEvent(image.asElement(), name) catch |err| {
```
karlseguin: What happens if the callback triggers a navigation?  This needs to be considered from two paths:
1 - the transfer.submit() synchronously calls it, so we're inside a Build.created / html5ever parsing
2 - the transfer fails asynchronously

src/browser/Frame.zig
```diff
@@ -2376,6 +2376,154 @@ pub fn loadExternalStylesheet(self: *Frame, link: *Element.Html.Link, href: []co
...
+    const arena = try session.getArena(.small, "Frame.loadImage");
+    defer arena.release();
+
+    const resolved = URL.resolve(arena.allocator(), self.base(), src, .{ .encoding = self.charset }) catch |err| {
+        // An unresolvable src is a load failure the same way a 404 is, and
+        // unlike a real fetch we can settle it without a request. No pending
+        // load was taken out yet, so queue rather than decrementing.
+        log.info(.http, "image resolve", .{ .err = err, .src = src });
+        image._complete = true;
+        return self.queueElementEvent(Factory.protoOf(image), .@"error");
```
karlseguin: Generation isn't captured, so if the src is updated, it'll still get the error delivery. Small window, but it is inconsistent with how `settle` works.

src/browser/Frame.zig
```diff
@@ -2376,6 +2376,154 @@ pub fn loadExternalStylesheet(self: *Frame, link: *Element.Html.Link, href: []co
     try self.fireElementEvent(element, comptime .wrap("load"));
 }
 
+// Asynchronous, unlike `loadExternalStylesheet`: nothing in the document
+// depends on the result, so there's no reason to block the parser on it.
+// The request does take a `_pending_loads` slot, so the window load event
+// waits for it the way the HTML spec says it should.
+pub fn loadImage(self: *Frame, image: *Element.Html.Image, src: []const u8) !void {
```
karlseguin: Would like to see all of this image code moved out of the busy Frame.zig and into a `src/browser/frame/resource_load.zig` (alongside preload.zig, obsevers.zig, ...)

src/browser/webapi/element/html/Image.zig
```diff
@@ -94,7 +111,28 @@ pub fn imageAddedCallback(self: *Image, frame: *Frame) !void {
...
+    // If image loading not desired, we just do fake "load" event.
+    if (frame._session.load_resources.image == false) {
+        return frame.queueLoad(Factory.protoOf(self));
+    }
+
+    // A fetch still in flight for the previous src is stale as of right now,
+    // and this is the only place that knows it. Wrapping is fine: colliding
+    // needs a request to still be in flight 2^32 src assignments later, which
+    // outlives any transfer timeout by several orders of magnitude.
+    self._generation +%= 1;
```
karlseguin: This needs to happen before the src.len == 0 check.

src/browser/webapi/element/html/Image.zig
```diff
@@ -140,9 +178,22 @@ pub const Build = struct {
         const self = node.as(Image);
         return self.imageAddedCallback(frame);
     }
+
+    // `img.src = ...` routes through `setSrc`, but `setAttribute("src", ...)`
+    // only reaches us here, and both have to (re)do fetch.
+    pub fn attributeChange(element: *Element, name: String, _: String, frame: *Frame) !void {
```
karlseguin, line 158: Do we need an `attributeRemove` ?

src/network/HttpClient.zig
```diff
@@ -89,6 +89,12 @@ in_use: std.DoublyLinkedList = .{},
 // Queue for request that are waiting an available connection (aka, easy)
 pending_queue: std.DoublyLinkedList = .{},
 
+// Same, for resources nothing on the page is blocked on. Only drained once
+// `pending_queue` is empty, so a document full of images can't hold every
```
karlseguin: `pending_low_queue` appears less useful than the comment indicates. It only triggers once our connection pool is exhausted. A document full of images **can** hold every connection while a script waits behind them. It's also harder to reason about how 2 queues interact with libcurl's own per-host connection limit. Finally, `pending_low_queue` always gives way to to a "normal" request, but it holds up the page "load" event (which not all normal requests do)...so page "load" is being artificially delayed.

src/network/HttpClient.zig
```diff
@@ -1033,6 +1064,13 @@ fn cacheStore(self: *Client, transfer: *Transfer) void {
     // entry a second time on any early return below.
     transfer._cache_intent = .none;
 
+    // A headers_only transfer never read the body. Storing it would put an
+    // empty entry under the real cache key and every later full fetch of
+    // that URL (an XHR, a `full` image fetch) would hit it and get nothing.
+    if (transfer.req.headers_only) {
```
karlseguin: I get the idea behind this, but...

One forum post where someone's avatar shows up 5 times will result in loading that image 5 times.

Or navigating from one page to another on the same site, will keep reloading the same (potentially) cacheable images over and over.

src/browser/webapi/element/html/Image.zig
```diff
@@ -1,16 +1,27 @@
...
+
 const Image = @This();
 
 pub const Proto = HtmlElement;
-_pad: bool = false,
+
+// Bumped on every src assignment. An in-flight fetch carries the value it
+// was issued under; when they no longer match, a newer src has taken over
+// and the older response must not fire events on this element.
+_generation: u32 = 0,
```
karlseguin, line 17: Did you consider aborting the transfer? XHR does that. Not saying it's better or would necessarily work, just curious if you considered it. (it might solve some of the edge cases around error handling).

src/network/HttpClient.zig
```diff
@@ -2935,6 +3018,29 @@ pub const Transfer = struct {
                 return @intCast(chunk_len);
             }
 
+            if (transfer.req.headers_only) {
+                const drainable = if (transfer.getContentLength()) |cl|
```
karlseguin: I like this, but it needs to be refined.

https://m.media-amazon.com/images/G/01/perc/star-empty-v2.svg

Is a 524 byte image and it shows up on _every_ amazon product page, with no content-length. Do we know exactly what libcurl does with the connection on CURLE_WRITE_ERROR? Does it kill the keepalive even for HTTP 2?  If it doesn't, could we just always abort in that case? If it does (or for HTTP/1.1), maybe we should read up to HEADERS_ONLY_DRAIN_MAX in the hopes that it fits?

karlseguin (commented):
The comments are verbose (looks like Claude to me). For example, in `Image.zig` `_generation` is documented twice; which is twice too many (the generation pattern is established in the codebase, if we think it's worth documenting, we should add a design document).  Also don't need an explanation that `+%= 1` is safe because we'd need 2^32 src sets.

They aren't just unnecessary, they risk being wrong. The non-existing `--fetch-images` flag is referenced multiple times.

testing.zig says the route is for `element/html/image_fetch.html`. What if someone else use it? The 2nd paragraph repeats the comment for HEADERS_ONLY_DRAIN_MAX. You could remove it and:

```zig
const body = try arena_allocator.alloc(u8, HttpClient.Request.HEADERS_ONLY_DRAIN_MAX + 1);
```

`HEADERS_ONLY_DRAIN_MAX` is public (but has no reason to be). So if you do make it private: 

```zig
// HttpClient.Request.HEADERS_ONLY_DRAIN_MAX + 1);
const body = try arena_allocator.alloc(u8, 16 * 1024 + 1);
```


`Session.load_resources` documents the setters. IDE / search can figure that out and not risk becoming stale. It also more or less summarizes the entire feature. (I do see `load_external_stylesheets` above does the same thing.)

karlseguin:
I think this is an interesting observation:

./lightpanda fetch "https://lightpanda.io/"
vs
./lightpanda fetch "https://lightpanda.io/" --load-resources image

For me, loads quite differently. The delay introduced by the "load" event waiting for the images, allows XHR/fetch requests to a number of resources (e.g. https://lightpanda.io/jobs.txt?_rsc=19zvn) to complete.

Don't think it's good or bad, but it is a bit surprising. We should remember that the impact of --load-resources image (or any other 'load'-delaying resource) can be larger than just images/css/etc.

src/browser/webapi/element/html/Image.zig
nikneym: That's also doable, I was unsure if `Image` should ever know about `ImageLoad`. With `_generation`, its `ImageLoad` that knows which `Image` it belongs. Though I'm experimenting it now.

src/browser/Frame.zig
nikneym: `settle` runs `pendingLoadCompleted` unconditionally which affects async fail and JS execution. I've also made `settle` not to dispatch anything which addresses the first case.

src/network/HttpClient.zig
nikneym: > Does it kill the keepalive even for HTTP 2?

For HTTP/2, it just kills the specific stream since its multiplexed. HTTP/1.1 results in aborting the data transfer.

> maybe we should read up to HEADERS_ONLY_DRAIN_MAX in the hopes that it fits?

I'm not sure on this, but I see the idea. Maybe we should not care about `Content-Length` and always read up to threshold.

src/browser/webapi/element/html/Image.zig
nikneym: Yes, we match the spec that way.

src/browser/Frame.zig
karlseguin: that js could location.href = "blah" or re-enter the parser which may or may not cause issues. Should add tests for these known pitfalls (or try to avoid it entirely)

src/network/HttpClient.zig
karlseguin: If you're sure about HTTP/2, seems like you could call use https://curl.se/libcurl/c/CURLINFO_HTTP_VERSION.html to always abort in that case.

karlseguin:
Would like to merge https://github.com/lightpanda-io/browser/pull/3248 first, since it fixes a UAF in this PR.

PR #3231, opened 2026-08-19
Screenshot

karlseguin:
Add new `--dump png` option. Replace `Page.captureScreenshot` placeholder with actual screenshot.

This generates a PNG of, essentially, the markdown representation of the page. Heavily Claude driven, but also isolated. 

src/html5ever/ -> src/rust/  and while we still build a single artifact, they rust dependencies are better organized. Should be adaptable to generating a PDF. There's at least 1 ugly thing: a mutex-protected singleton renderer to avoid having to cache/create a bunch of state/cache per worker. 

```
zig build run -- fetch "https://lightpanda.io/" --dump png > screenshot.png
```

screenshot


src/rust/render/lib.rs
```diff
@@ -0,0 +1,690 @@
...
+        let style = gr.style();
+        let mut paint = Paint::default();
+        paint.anti_alias = true;
+        paint.set_color_rgba8(style.brush.0, style.brush.1, style.brush.2, 255);
+
+        let baseline = oy + gr.baseline();
+        let mut x = ox + gr.offset();
+        let skew = synth
+            .skew()
+            .map(|deg| (-deg.to_radians()).tan())
```
arrufat: The synthesized oblique leans left (backslant) instead of right. The transform below composes this with a y-flip: `Transform::from_row(scale, 0.0, skew * scale, -scale, gx, gy)` maps a y-up outline point to `x' = scale·x + skew·scale·y_up + gx`, so for glyph tops (`y_up > 0`) the negative skew shifts them left.

```suggestion
        let skew = synth
            .skew()
            .map(|deg| deg.to_radians().tan())
            .unwrap_or(0.0);
```

Easy to spot on any page with `<i>`/`<em>` — e.g. img alt text, which renders italic.

src/rust/render/lib.rs
```diff
@@ -0,0 +1,690 @@
...
+
+impl Renderer {
+    fn new() -> Self {
+        let mut collection = Collection::new(CollectionOptions {
+            shared: false,
+            system_fonts: false,
+        });
+        let sans = Self::register(&mut collection, include_bytes!("fonts/DejaVuSans.ttf"));
+        Self::register(&mut collection, include_bytes!("fonts/DejaVuSans-Bold.ttf"));
+        let mono = Self::register(&mut collection, include_bytes!("fonts/DejaVuSansMono.ttf"));
```
arrufat, line 310: Worth bundling DejaVuSansMono-Bold too (2.37, ~330KB)? Every heading is bold, so any `<code>` inside an `<h1>`–`<h6>` — the usual shape of docs pages — goes through the synthesized embolden today, and the stroked outline reads blobby next to the real face. With it registered, fontique matches weight 700 for monospace on its own, and the `synth.embolden()` stroke at L617–644 becomes dead code.

(The suggestion only covers the registration; the .ttf itself needs adding under `fonts/`.)

```suggestion
        let mono = Self::register(&mut collection, include_bytes!("fonts/DejaVuSansMono.ttf"));
        Self::register(&mut collection, include_bytes!("fonts/DejaVuSansMono-Bold.ttf"));
```

PR #3240, opened 2026-08-20
XHR Sync Request

mookums:
This adds support for synchronous XHR requests. It also improves coverage on the `/xhr/` tests. This will eventually be needed by certain WPT tests to pass on #3002

### DIFF
--- a/src/browser/webapi/net/XMLHttpRequest.zig
+++ b/src/browser/webapi/net/XMLHttpRequest.zig
@@ -347,6 +347,7 @@ pub fn send(self: *XMLHttpRequest, body_: ?BodyInit, exec_: *const Execution) !v
         return error.NetworkError;
     };
     defer resp.deinit();
+    defer self.releaseSelfRef();
 
     self._response_status = resp.status;
     self._response_url = self._url;
--- a/src/browser/webapi/net/XMLHttpRequest.zig
+++ b/src/browser/webapi/net/XMLHttpRequest.zig
@@ -434,6 +434,11 @@ pub fn getResponseType(self: *const XMLHttpRequest) ResponseType {
 }
 
 pub fn setResponseType(self: *XMLHttpRequest, value: []const u8, exec: *const Execution) !void {
+    const rt: ResponseType = if (value.len == 0)
+        .default
+    else
+        std.meta.stringToEnum(ResponseType, value) orelse return;
+
     if (self._ready_state == .loading or self._ready_state == .done) {
         return error.InvalidStateError;
     }
@@ -447,10 +452,8 @@ pub fn setResponseType(self: *XMLHttpRequest, value: []const u8, exec: *const Ex
         return;
     }
 
-    if (std.meta.stringToEnum(ResponseType, value)) |rt| {
-        if (rt != .default) {
-            self._response_type = rt;
-        }
+    if (rt != .default) {
+        self._response_type = rt;
     }
 }
 
--- a/src/browser/webapi/net/XMLHttpRequest.zig
+++ b/src/browser/webapi/net/XMLHttpRequest.zig
@@ -187,10 +187,9 @@ pub fn getTimeout(self: *const XMLHttpRequest) u32 {
     return self._timeout;
 }
 
-pub fn setTimeout(self: *XMLHttpRequest, value: u32) !void {
+pub fn setTimeout(self: *XMLHttpRequest, value: u32, exec: *const Execution) !void {
     // https://xhr.spec.whatwg.org/#the-timeout-attribute
-    // Throw if the request is sync OR if it is already sent.
-    if (!self._async and self._ready_state != .unsent) {
+    if (!self._async and exec.js.global == .frame) {
         return error.InvalidAccessError;
     }
 
@@ -254,6 +253,12 @@ pub fn send(self: *XMLHttpRequest, body_: ?BodyInit, exec_: *const Execution) !v
     }
     if (self._ready_state != .opened or self._send_flag) {
         return error.InvalidStateError;
+    }
+
+    if (!self._async and exec_.js.global == .frame and
+        (self._timeout != 0 or self._response_type != .default))
+    {
+        return error.InvalidAccessError;
     }
 
     if (body_) |b| {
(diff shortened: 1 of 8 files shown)


src/browser/webapi/net/XMLHttpRequest.zig
```diff
@@ -303,20 +318,59 @@ pub fn send(self: *XMLHttpRequest, body_: ?BodyInit, exec_: *const Execution) !v
...
+            return err;
+        };
+        return;
+    }
+
+    var resp = transfer.submitSync() catch |err| {
         self._send_flag = false;
-        return err;
+        self.handleError(err);
+        self.releaseSelfRef();
```
krichprollsch, line 352: Claude thinks we shouldn't callback error here:

```
Verified by running it: a sync request to a closed port gives send() returns normally, onerror fires, status = 0, readyState = 4.

  The spec says the opposite for sync requests. The "request error steps" must throw a NetworkError DOMException and must not fire the error event. See https://xhr.spec.whatwg.org/#request-error-steps step 4: "If xhr's synchronous flag is set, then throw exception."

  The PR description says this is for WPT tests on #3002. The WPT xhr/send-network-error-sync-events.sub.html tests exactly this. The fix is in the catch block at src/browser/webapi/net/XMLHttpRequest.zig:339 — set state to done, release the self ref, then return error.NetworkError instead of calling handleError.
```

src/browser/webapi/net/XMLHttpRequest.zig
```diff
@@ -303,20 +318,59 @@ pub fn send(self: *XMLHttpRequest, body_: ?BodyInit, exec_: *const Execution) !v
...
+    try self._proto.dispatch(.load_end, .{ .total = loaded, .loaded = loaded, .length_computable = true }, exec);
+
+    log.info(.http, "request complete", .{
+        .source = "xhr",
+        .url = self._url,
+        .status = self._response_status,
+        .len = self._response_data.items.len,
+    });
+
+    self.releaseSelfRef();
```
krichprollsch: shouldn't we release in case of earlier error?

src/browser/webapi/net/XMLHttpRequest.zig
```diff
@@ -303,20 +318,59 @@ pub fn send(self: *XMLHttpRequest, body_: ?BodyInit, exec_: *const Execution) !v
...
     };
+    defer resp.deinit();
+
+    self._response_status = resp.status;
+    self._response_url = self._url;
+    self._response_len = resp.body.items.len;
+
+    // TODO: Headers.
+    self._response_headers = .empty;
+    self._response_mime = null;
```
krichprollsch: Clause thinks we should delete these lines
```
  Two problems:
  - It abandons the capacity that open() kept with clearRetainingCapacity() (line 219). The list is arena-backed, so it is not
    a hard leak, but the memory is retained by the pinned arena for the life of the XHR object. Repeated sync requests grow it
    for nothing.
  - It is dead code anyway: send() requires _ready_state == .opened, so open() has always just cleared _response_headers and
    set _response_mime = null. Delete both lines and keep the TODO comment.
```

src/browser/webapi/net/XMLHttpRequest.zig
krichprollsch, line 443: Claude says we should throw error for sync req.

```
  The spec's open() step throws InvalidAccessError when async is false and either timeout is not 0 or responseType is not "".
  The PR only checks timeout (XMLHttpRequest.zig:225). The responseType setter must also throw when the synchronous flag is set
  (in a Window). Today this works with no error:

  req.open('GET', url, false);
  req.responseType = 'document';  // spec: InvalidAccessError
  req.send();

  WPT xhr/responsetype.any.js covers this.
```

krichprollsch, line 230: move method and url resolve before setting `self._async = async_ orelse true;` to avoid letting the request in a bad state for future reuse.

src/browser/webapi/net/XMLHttpRequest.zig
```diff
@@ -185,13 +187,19 @@ pub fn getTimeout(self: *const XMLHttpRequest) u32 {
     return self._timeout;
 }
 
-pub fn setTimeout(self: *XMLHttpRequest, value: u32) void {
+pub fn setTimeout(self: *XMLHttpRequest, value: u32) !void {
+    // https://xhr.spec.whatwg.org/#the-timeout-attribute
+    // Throw if the request is sync OR if it is already sent.
+    if (!self._async and self._ready_state != .unsent) {
```
krichprollsch: ar you sure about the `OR if it is already sent.` part?

src/browser/webapi/net/XMLHttpRequest.zig
```diff
@@ -303,20 +323,60 @@ pub fn send(self: *XMLHttpRequest, body_: ?BodyInit, exec_: *const Execution) !v
...
+    try self._proto.dispatch(.load_end, .{ .total = loaded, .loaded = loaded, .length_computable = true }, exec);
+
+    log.info(.http, "request complete", .{
+        .source = "xhr",
+        .url = self._url,
+        .status = self._response_status,
+        .len = self._response_data.items.len,
+    });
+
+    self.releaseSelfRef();
```
krichprollsch: we don't need to release here anymore, thanks to the `defer`.

src/browser/webapi/net/XMLHttpRequest.zig
```diff
@@ -303,20 +323,60 @@ pub fn send(self: *XMLHttpRequest, body_: ?BodyInit, exec_: *const Execution) !v
...
+        return;
+    }
+
+    var resp = transfer.submitSync() catch |err| {
+        log.err(.http, "sync request failed", .{
+            .source = "xhr",
+            .url = self._url,
+            .err = err,
+        });
         self._send_flag = false;
```
krichprollsch, line 351: I seems you should also change direclty  the ready_state to `done` in case of error (w/o triggering an event)
```
self._ready_state = .done;
```

krichprollsch (approved):

PR #3245, opened 2026-08-21
`AdBlocker`: request engine

nikneym:
On-going effort; here's the recent benchmark made against EasyList + EasyPrivacy on this branch:

# Adblock engine: EasyList + EasyPrivacy corpus report

Lists downloaded fresh from easylist.to the same day (135,908 lines total). Measured with
the per-request engine (tries + token-bucket indexes) after the adversarial-review fixes.

## Coverage: what loads

|                              | EasyList           | EasyPrivacy         | Combined            |
| ---------------------------- | ------------------ | ------------------- | ------------------- |
| File lines                   | 79,233             | 56,675              | 135,908             |
| Comments / blank             | 13,920             | 771                 | 14,691              |
| **Rules**                    | **65,313**         | **55,904**          | **121,217**         |
| Cosmetic rules (not network) | 10,852             | 32                  | 10,884              |
| **Network rules**            | **54,461**         | **55,872**          | **110,333**         |
| **Loaded**                   | **54,266 (99.6%)** | **55,858 (99.97%)** | **110,124 (99.8%)** |
| → in hostname tries          | 45,264             | 42,955              | 88,219              |
| → engine-indexed             | 9,002              | 12,903              | 21,905              |

## The 209 network rules we don't apply, by reason

- **31 `/regex/` filters** (24 EL + 7 EP) — the only rules uBO enforces as
  blocks that we genuinely can't: there's still no regex engine. ~0.03% of the
  corpus.
- **168 cosmetic-realm exceptions** (`$generichide`/`$elemhide`/
  `$specifichide`) — they modify cosmetic filtering, which we don't have yet,
  so there's nothing for them to except.
- **6 modifier-only rules** (`$removeparam`-class) — uBO applies these as URL
  rewrites, never blocks; dropping them is the conservative, correct choice.
- **3 malformed rules** (EasyPrivacy) — uBO rejects these too.
- **1 unsupported option** (the `@@…$redirect-rule` google-analytics line —
  correctly treated as excepting nothing, so `google-analytics.com` stays
  blocked).

The ~10,884 "cosmetic" drops are element-hiding rules (`##`, `#@#` — including
944 with spaces in their CSS selectors that take the hosts-file parsing path);
no network engine applies those, so they're out of scope rather than failures.
Notably, **zero hostnames land in the `suppressed` trie** across the whole
corpus — every exception either evaluates per-request or provably unblocks
nothing.

## Correctness on the loaded rules

- A 26-assertion doubleclick battery against uBO ground truth passes with the
  review fixes in: gpt.js blocked except on the listed carve-out sites, the
  `/ssai/` exception honored, wrong-type/wrong-site/wrong-path variants all
  correct, and the apex `doubleclick.net` undecided rather than suppressed.
- End-to-end: `lightpanda fetch` on a page loading the real ad stack blocks
  `gpt.js`, `td.`/`googleads.`/`fls.doubleclick.net` and Google Analytics
  before they reach the network.

## Bottom line

**99.8% of the network rules in EasyList+EasyPrivacy load and enforce; the
only real gap is 31 regex filters.**

### DIFF
--- a/src/network/HttpClient.zig
+++ b/src/network/HttpClient.zig
@@ -4264,6 +4264,9 @@ fn testIsUrlBlocked(client: *const Client, opts: TestRequest) bool {
         .req = .{
             .method = .GET,
             .url = opts.url,
+            .origin = null,
+            .credentials_mode = .omit,
+            .request_mode = .no_cors,
             .resource_type = opts.resource_type,
             .internal = opts.internal,
             .shutdown_callback = noopShutdown,
--- a/src/network/HttpClient.zig
+++ b/src/network/HttpClient.zig
@@ -404,16 +404,7 @@ fn isUrlBlocked(self: *const Client, transfer: *const Transfer) bool {
     if (self.url_blocklist) |*blocklist| {
         if (blocklist.isBlocked(req.url)) return true;
     }
-    return self.isAdblocked(transfer);
-}
-
-fn isAdblocked(self: *const Client, transfer: *const Transfer) bool {
-    const blocker = if (self.network.adblocker) |*b| b else return false;
-    var buffers: AdBlocker.Request.Buffers = undefined;
-    const target = AdBlocker.Request.fromHttp(transfer, &buffers) orelse return false;
-    const verdict = blocker.match(target);
-    lp.metrics.adblock_verdicts.incr(verdict);
-    return verdict == .blocked;
+    return if (self.network.adblocker) |*blocker| blocker.isBlocked(transfer) else false;
 }
 
 fn adblockSourceUrl(transfer: *const Transfer) ?[]const u8 {
--- a/src/network/HttpClient.zig
+++ b/src/network/HttpClient.zig
@@ -410,11 +410,7 @@ fn isUrlBlocked(self: *const Client, transfer: *const Transfer) bool {
 fn isAdblocked(self: *const Client, transfer: *const Transfer) bool {
     const blocker = if (self.network.adblocker) |*b| b else return false;
     var buffers: AdBlocker.Request.Buffers = undefined;
-    const target = AdBlocker.Request.fromHttp(
-        &transfer.req,
-        adblockSourceUrl(transfer),
-        &buffers,
-    ) orelse return false;
+    const target = AdBlocker.Request.fromHttp(transfer, &buffers) orelse return false;
     const verdict = blocker.match(target);
     lp.metrics.adblock_verdicts.incr(verdict);
     return verdict == .blocked;
@@ -1904,11 +1900,6 @@ pub const Request = struct {
     // Whether this request should (possibly) be throttled based on the RateLimiter
     throttle: bool = false,
 
-    // Whether a `.document` request is loading a nested frame rather than the
-    // top-level page. The reason why is filter lists separate the two ($document vs
-    // $subdocument), and only ever block the latter.
-    is_subframe: bool = false,
-
     // Set by syncRequest; only used to label the http_requests metric.
     sync: bool = false,
 
@@ -2109,7 +2100,9 @@ pub const Owner = struct {
 
     // The parent frame's Owner; for a worker, its creating frame's. Outlives
     // this Owner: child frames are torn down before their parent, a worker
-    // before its frame.
+    // before its frame. A `.document` request's owner is the frame it
+    // navigates, so a parent here is what makes that load a nested frame's;
+    // the adblocker tells $document from $subdocument by it.
     parent: ?*const Owner,
 
     // Copied onto every request made through this owner, see Request.
@@ -4251,7 +4244,6 @@ const TestRequest = struct {
     /// The page embedding `document`, when the test wants a deeper chain.
     parent_document: [:0]const u8 = "",
     resource_type: Request.ResourceType = .document,
-    is_subframe: bool = false,
     internal: bool = false,
 };
 
@@ -4282,7 +4274,6 @@ fn testIsUrlBlocked(client: *const Client, opts: TestRequest) bool {
             .method = .GET,
             .url = opts.url,
             .resource_type = opts.resource_type,
-            .is_subframe = opts.is_subframe,
             .internal = opts.internal,
             .shutdown_callback = noopShutdown,
         },
@@ -4331,10 +4322,11 @@ test "HttpClient: adblock verdicts apply per request" {
         .resource_type = .xhr,
     }));
 
-    // A `.document` request is $subdocument only inside a nested frame.
+    // A `.document` request is $subdocument only inside a nested frame,
+    // which its owner chain tells: the navigated frame has a parent.
     try testing.expect(testIsUrlBlocked(&client, .{
         .url = "https://framed.example.com/",
-        .is_subframe = true,
+        .document = "https://news.com/",
     }));
     try testing.expect(!testIsUrlBlocked(&client, .{ .url = "https://framed.example.com/" }));
 
@@ -4352,16 +4344,13 @@ test "HttpClient: adblock verdicts apply per request" {
     }));
 
     // A top-level navigation is its own context: the page it was clicked on
-    // does not make it third-party...
-    try testing.expect(!testIsUrlBlocked(&client, .{
-        .url = "https://partied.example.com/",
-        .document = "https://news.com/",
-    }));
+    // is not in its owner chain (it only travels as cookie_origin, which the
+    // adblocker never reads), so nothing makes it third-party...
+    try testing.expect(!testIsUrlBlocked(&client, .{ .url = "https://partied.example.com/" }));
     // ...but a subframe loading the same URL keeps its document's context.
     try testing.expect(testIsUrlBlocked(&client, .{
         .url = "https://partied.example.com/",
         .document = "https://news.com/",
-        .is_subframe = true,
     }));
 
     // The context is the issuing frame's document even under a cross-site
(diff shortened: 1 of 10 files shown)


src/network/adblock/Engine.zig
```diff
@@ -0,0 +1,319 @@
...
+    kind: NetworkFilter.ResourceTypes,
+    third_party: bool,
+
+    /// `url` must already be lowercased and fragment-free; `source_hostname`
+    /// may be empty when there is no document context.
+    pub fn init(
+        url: []const u8,
+        source_hostname: []const u8,
+        kind: NetworkFilter.ResourceTypes,
+    ) Request {
```
krichprollsch, line 104: What about having a constructor from an `HttpClient.Request` direclty?

krichprollsch: You could move `HttpCLient.adblockResourceType` and `normalizeForAdblock` directly here then.I think it's ok to depends on `HttpClient.Request` here. `HttpClient` is very big, so code we can move from here to adblock is a win IMO.

krichprollsch:
It seems `||coinimp.com` `||coinimp.com^` and `||coinimp.com|` are all the same in ublock: hostname is `coinimp.com` or a subdomain

krichprollsch:
`@@…$important` seems to be invalid in ublock. we should return `error.InvalidOption` in this case.

src/network/adblock/Engine.zig
```diff
@@ -0,0 +1,319 @@
...
+
+/// Tokens considered per filter. Long query-string patterns can exceed this;
+/// taking the first few costs a wider bucket, never a wrong answer.
+const TOKENS_MAX = 32;
+
+/// The first filter that matches, or null.
+pub fn match(self: *const Engine, request: Request) ?*const NetworkFilter {
+    if (self.filters.len != 0) {
+        var it: Tokens = .{ .text = request.url.text };
+        while (it.next()) |token| {
```
krichprollsch, line 192: We could tokenize once instead of having to do it for each engines.

src/network/HttpClient.zig
```diff
@@ -391,23 +391,59 @@ fn clearUrlBlocklist(self: *Client) void {
...
-    return blocker.matchHostname(hostname) == .blocked;
+
+    var url_buf: [ADBLOCK_URL_MAX]u8 = undefined;
+    const url = normalizeForAdblock(req.url, &url_buf) orelse return false;
+
+    var source_buf: [253]u8 = undefined;
+    const top_level = req.resource_type == .document and !req.is_subframe;
+    const source_host = if (top_level) "" else URL.getOriginHostname(req.cookie_origin);
+    const source = if (source_host.len == 0 or source_host.len > source_buf.len)
+        ""
```
krichprollsch: Claude suggests to return `false` directly here. WDYT?

nikneym: Making a change for document hostnames longer than 253 bytes to skip adblocking entirely.

src/browser/Frame.zig
```diff
@@ -856,6 +856,8 @@ pub fn navigate(self: *Frame, request_url: [:0]const u8, opts: NavigateOpts) !vo
         .throttle = self.parent == null,
         .cookie_origin = opts.initiator_url,
         .resource_type = .document,
+        .is_subframe = self.parent != null,
+        .notification = self._session.notification,
```
karlseguin: Why? the point of https://github.com/lightpanda-io/browser/pull/3364 was to remove this burden from every caller.

src/network/adblock/AdBlocker.zig
```diff
@@ -123,93 +162,354 @@ pub fn parse(self: *AdBlocker, reader: *Io.Reader) !void {
...
+                var buf: [253]u8 = undefined;
+                if (exceptionHostname(dropped.line, &buf)) |hostname| {
+                    try self.addToTrie(self.suppressed, hostname);
+                }
+                continue;
+            },
+            .filter => |filter| filter,
+        };
+
+        if (!isSupported(&filter)) {
```
karlseguin, line 204: This makes the `rules_skipped` log message look worse than it is due to skipping cosmetic rules.

src/network/adblock/AdBlocker.zig
```diff
@@ -25,25 +25,45 @@ const Allocator = std.mem.Allocator;
...
+const Engine = @import("Engine.zig");
 const HostnameTrie = @import("HostnameTrie.zig");
 const NetworkFilter = @import("NetworkFilter.zig");
 
 const log = lp.log;
 
 const AdBlocker = @This();
 
+pub const Request = Engine.Request;
+pub const ResourceTypes = NetworkFilter.ResourceTypes;
```
karlseguin: No reason to be public.

src/network/adblock/Engine.zig
```diff
@@ -0,0 +1,406 @@
...
+
+    return .{
+        .filters = filters,
+        .buckets = buckets,
+        .fallback = try arena.dupe(u32, fallback.items),
+    };
+}
+
+/// Runs of alphanumerics. Everything else ('.', '/', '-', '%', '_', '?') is a boundary.
+pub const Tokens = struct {
```
karlseguin: No reason to be public.

src/network/HttpClient.zig
```diff
@@ -407,23 +407,34 @@ fn clearUrlBlocklist(self: *Client) void {
...
+fn isAdblocked(self: *const Client, transfer: *const Transfer) bool {
     const blocker = if (self.network.adblocker) |*b| b else return false;
-    const host = URL.getHostname(url);
-    if (host.len == 0 or host.len > 253) return false;
-    // The trie expects normalized (lowercase) hostnames; URLs aren't
-    // guaranteed to arrive that way.
-    var buf: [253]u8 = undefined;
-    const hostname = std.ascii.lowerString(&buf, host);
-    return blocker.matchHostname(hostname) == .blocked;
+    var buffers: AdBlocker.Request.Buffers = undefined;
```
karlseguin: All of this are details of AdBlocker that HttpClient doesn't need to know.

```zig
  fn isUrlBlocked(self: *const Client, transfer: *const Transfer) bool {
      if (req.internal) return false;
      if (self.url_blocklist) |*b| if (b.isBlocked(req.url)) return true;
      return if (self.network.adblocker) |*b| b.isBlocked(transfer) else false;
  }
```

src/network/adblock/AdBlocker.zig
```diff
@@ -123,93 +162,354 @@ pub fn parse(self: *AdBlocker, reader: *Io.Reader) !void {
...
+}
+
 pub const Verdict = enum { none, allowed, blocked };
 
-/// `.none` means no hostname-wide verdict applies. Filters that need more
-/// than a hostname match are not represented here at all.
-pub fn matchHostname(self: *const AdBlocker, hostname: []const u8) Verdict {
+/// The verdict for one request.
+/// `$important` blocks outrank exceptions, which outrank plain blocks; same as uBO.
+pub fn match(self: *const AdBlocker, request: Request) Verdict {
```
karlseguin: `Request` is pretty big...should be  passed as a `const *Request` everywhere

src/network/adblock/Engine.zig
```diff
@@ -0,0 +1,406 @@
...
+            .url = parsed,
+            .source_hostname = source,
+            .kind = kind,
+            .third_party = domain.isThirdParty(parsed.hostname(), source),
+            .tokens_buf = undefined,
+            .tokens_len = 0,
+        };
+        var it: Tokens = .{ .text = url };
+        while (it.next()) |token| {
+            if (request.tokens_len == request.tokens_buf.len) break;
```
karlseguin: Right now, AdBlocker.match does:

```zig
    if (self.trie.matches(self.suppressed, hostname) != null) return .none;

    if (self.trie.matches(self.blocked_important, hostname) != null) return .blocked;
    if (self.blocking_important.match(request) != null) return .blocked;

    if (self.exceptions.match(request) != null) return .allowed;
    if (self.trie.matches(self.allowed, hostname) != null) return .allowed;

    if (self.trie.matches(self.blocked, hostname) != null) return .blocked;
    if (self.blocking.match(request) != null) return .blocked;
```

There are cases where you won't even need the tokens that you've already split / hashed. AND, you can rewrite that to be even more efficient:

```zig
  if (self.trie.matches(self.suppressed, hostname) != null) return .none;

  if (self.trie.matches(self.blocked_important, hostname) != null) return .blocked;
  if (self.blocking_important.match(request) != null) return .blocked;
  
  if (self.trie.matches(self.blocked, hostname) != null or self.blocking.match(request) != null) {
      if (self.trie.matches(self.allowed, hostname) != null) return .allowed;
      if (self.exceptions.match(request) != null) return .allowed;
      return .blocked;
  }
  return .none;
```

There's no need to check exceptions unless something is blocked (.allowed vs .none is just a metrics difference) AND there are fewer cases where you need to tokenize/hash anything.

With that in mind, don't pre-tokenize. Tokenize on demand. Whether or not it's worth caching in the tokenizer, I don't know.  But it isn't worth limiting it to 128 tokens. Since the tokenizer becomes streaming, you can cache the first 128 values, and then stream until the end of the URL.

src/browser/Frame.zig
nikneym: Right, this was before that being merged I believe.

src/network/adblock/Engine.zig
nikneym: > Since the tokenizer becomes streaming, you can cache the first 128 values, and then stream until the end of the URL.

I actually thought the same, Claude also recommended, but unsure how many cases would benefit from it. Currently overflow results to loss of tokens; comparing it with stream-the-rest approach.

karlseguin:
LGTM

PR #3246, opened 2026-08-21
chore: Remove Server component from Network

karlseguin:
Network is less cohesive than https://github.com/lightpanda-io/browser/pull/3242 would indicate. It has two distinct and _completely_ separate responsibilities.

1 - It acts as the base for each HttpClient, providing a shared connection pool
    (for http and ws) and access to the process wide Cache, RobotStore,
    WebBothAuth, certificates, ...

2 - It accepts, polls and reads from CDP connection

There is zero relationship between these, and it's a particularly bad place for this duality to exist because both parts are, in their own way, the main multi-threaded junction in the system.

This commit is purely mechanical it:

1 - Keeps network as the base for each HttpClient. 
2 - Extract the CDP interaction into the existing Server.zig 3 - Gives mcp's HttpServer its own accept loop

### DIFF
--- a/src/App.zig
+++ b/src/App.zig
@@ -81,10 +81,6 @@ pub fn init(allocator: Allocator, config: *const Config) !*App {
     return app;
 }
 
-pub fn shutdown(self: *const App) bool {
-    return self.network.shutdown.load(.acquire);
-}
-
 pub fn deinit(self: *App) void {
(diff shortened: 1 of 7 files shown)


krichprollsch (approved):
That's a great simplification 👍

PR #3252, opened 2026-08-23
search: add Keenable engine with a keyless rung above the DDG scrape

ilya-bogin-keenable:
Adds `keenable` to the `search` tool's engines. Depends on the client in lightpanda-io/zenai#9.

What changes:
- `keenable` joins `api_engines` (`KEENABLE_API_KEY`, tried after brave/tavily/exa in `.auto`) and the `SearchEngine` enum, so `/searchEngine keenable` and the persisted setting pick it up automatically.
- **Keyless rung**: Keenable also answers with no key (public endpoint, rate-limited per client IP), so `.auto` now tries it as the last rung before the DuckDuckGo scrape — structured `{title, url, snippet}` results instead of the degraded SERP scrape, still zero-credential. Skipped when `KEENABLE_API_KEY` is set (the keyed loop already covered it). If you'd rather keep the default fallback chain untouched, this rung is one hunk and can be dropped — the keyed engine stands alone.
- `/searchEngine keenable` with no key uses the keyless endpoint instead of erroring; the REPL prints an info line instead of the missing-key warning.
- Engine table entries gain `init_options`, which lets the Keenable client send `X-Keenable-Title: lightpanda` (required by the keyless endpoint, pure attribution otherwise).

Note on the zon pin: it points at the zenai PR branch so CI can build; happy to repoint once lightpanda-io/zenai#9 lands.

Verified:
- `make test`: 1238/1238 pass (includes new `formatKeenableMarkdown` tests on the real response shape); `zig fmt --check` clean.
- Live e2e over MCP stdio with an empty environment (no search keys at all): `tools/call search` returned 10 structured results through the keyless endpoint.

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -36,8 +36,8 @@
             .hash = "sqlite3-3.53.2-DMxLWuAOAAA_Px0arJOIOaP4AKEu5prbsQgPMA35W1zz",
         },
         .zenai = .{
-            .url = "git+https://github.com/lightpanda-io/zenai.git#5c208f61c5b6b0bc46ceaf2670d0d267b1eef87e",
-            .hash = "zenai-0.0.0-iOY_VKQ9BgC0_zD9nj_EedEkNioEJnGqoKsHWcq27D74",
+            .url = "git+https://github.com/keenableai/zenai.git#d657a7a5d7c080201ac249b4a1e2cb95981f1bbd",
+            .hash = "zenai-0.0.0-iOY_VBRZBgDgVyBTKY2QfeTorEsE3evpqIBbzrlBONvu",
         },
         .isocline = .{
             .url = "git+https://github.com/arrufat/isocline?ref=lightpanda#832a9fe25f5f4458fcc47b5acc7c21db669c2f47",
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -36,7 +36,7 @@
             .hash = "sqlite3-3.53.2-DMxLWuAOAAA_Px0arJOIOaP4AKEu5prbsQgPMA35W1zz",
         },
         .zenai = .{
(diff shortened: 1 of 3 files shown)


ilya-bogin-keenable:

ilya-bogin-keenable:
recheck

ilya-bogin-keenable:
recheck

arrufat:
zenai#9 is merged (`lightpanda-io/zenai@4aa2ccc`), so this PR can drop the fork pin. I prepared the follow-up commit, but since this branch lives in an org-owned fork, GitHub's "allow edits by maintainers" doesn't apply — so I've opened it as a PR against your branch instead: **keenableai/browser#1**. Merging it there folds it into this PR.

What it contains:

- **`build.zig.zon`** repinned to `lightpanda-io/zenai#4aa2ccc` — the merge commit of zenai#9. Same tree as your `8f89838`, so the hash is unchanged.
- **Keyless rung guard relaxed**: the rung now also runs after a failed keyed attempt, which is what the search tool description and `/searchEngine` help already say ("or on API failure"); previously a revoked `KEENABLE_API_KEY` silently degraded every `.auto` search to the DDG scrape.
- **`retry_policy = .disabled` on the keenable entry**: the `.auto` cascade is the retry mechanism, and honoring the public endpoint's `Retry-After: 60` (zenai's default policy sleeps up to 60s × 3 retries) would stall the search tool short of the DDG fallback that exists for exactly that case.
- Two stale doc comments (`SearchEngine`, `searchExplicit`) updated for the keyless rung.

Full test suite passes (1238/1238), `zig fmt --check` clean. Happy to adjust if you'd rather reconcile the guard the other way (keyless only when the key is unset) — but then the tool description and help prose need rewording instead.

arrufat:
Thanks for the PR :)

PR #3256, opened 2026-08-24
mcp: enforce HTTP conditions

krichprollsch:
Apply same conditions than CDP for MCP HTTP conns:
* refuse requests including Origin header
* accept only ip as host
* ensure content-type is json for POST

### DIFF
--- a/src/mcp/HttpServer.zig
+++ b/src/mcp/HttpServer.zig
@@ -403,7 +403,7 @@ fn serve(self: *HttpServer, out: *std.Io.Writer, arena: std.mem.Allocator, reque
     // the head's string memory.
     const session_id = checkHeaders(arena, request) catch |err| switch (err) {
         error.ForbiddenOrigin => return request.respond("Origin not allowed\n", .{ .status = .forbidden, .keep_alive = false }),
-        error.ForbiddenHost => return request.respond("Host not allowed, connect to the MCP endpoint by IP address\n", .{ .status = .forbidden, .keep_alive = false }),
+        error.ForbiddenHost => return request.respond("Host not allowed, connect to the MCP endpoint by IP address or localhost\n", .{ .status = .forbidden, .keep_alive = false }),
         else => return err,
     };
     const keep_alive = request.head.keep_alive;
@@ -457,6 +457,11 @@ fn checkHeaders(arena: std.mem.Allocator, request: *std.http.Server.Request) !?[
             return error.ForbiddenOrigin;
         } else if (std.ascii.eqlIgnoreCase(key, "host")) {
             const is_allowed = blk: {
+                // allow literal localhost
+                if (std.mem.startsWith(u8, value, "localhost:")) {
+                    break :blk true;
+                }
+
                 _ = std.Io.net.IpAddress.parseLiteral(value) catch break :blk false;
                 break :blk true;
             };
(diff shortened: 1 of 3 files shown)


src/mcp/HttpServer.zig
```diff
@@ -395,9 +395,18 @@ fn serve(self: *HttpServer, out: *std.Io.Writer, arena: std.mem.Allocator, reque
...
+    }
+
+    // Read the headers and keep_alive before the body reader invalidates
+    // the head's string memory.
+    const session_id = checkHeaders(arena, request) catch |err| switch (err) {
+        error.ForbiddenOrigin, error.ForbiddenHost => {
+            return request.respond("", .{ .status = .forbidden, .keep_alive = false });
+        },
+        else => return err,
+    };
```
arrufat, line 408: Every documented way of reaching this endpoint now gets a bodyless 403: `parseLiteral("localhost:8000")` fails, and the "connect by IP address" hint only goes to the server log, never to the client. The docs' own curl example (`docs/src/content/usage/mcp.mdx`, "Calling the HTTP endpoint") uses `http://localhost:8000/mcp`, and `README.md` / `help.zon` still say "POST JSON-RPC to `http://host:9223/mcp`".

CDP already sends a reason (`Connection.zig:249-250`); `request.respond` takes the body as its first argument, so mirroring that is free:

```suggestion
    const session_id = checkHeaders(arena, request) catch |err| switch (err) {
        error.ForbiddenOrigin => return request.respond("Origin not allowed\n", .{ .status = .forbidden, .keep_alive = false }),
        error.ForbiddenHost => return request.respond("Host not allowed, connect to the MCP endpoint by IP address\n", .{ .status = .forbidden, .keep_alive = false }),
        else => return err,
    };
```

The docs curl example and README/help text should switch to `127.0.0.1` in the same change (or a follow-up), otherwise the first thing users copy-paste breaks.

src/mcp/HttpServer.zig
```diff
@@ -395,9 +395,18 @@ fn serve(self: *HttpServer, out: *std.Io.Writer, arena: std.mem.Allocator, reque
         return request.respond("", .{ .status = .expectation_failed, .keep_alive = false });
     }
 
-    // Read the session header and keep_alive before the body reader
-    // invalidates the head's string memory.
-    const session_id = try sessionHeader(arena, request);
+    if (method == .POST and !isJsonContentType(request.head.content_type)) {
```
arrufat, line 398: Every `.keep_alive = false` response (this 415, the 403 below, and the pre-existing 405/413/417) leaves the POST body unread in the socket: std's `discardBody` only drains when the *server* wants keep-alive, otherwise it just sets `reader.state = .closing` and relies on the caller to close ("the connection will be severed after the response is sent").

But `handleConn` loops on the *client's* flag — `if (!request.head.keep_alive) return;` — which is `true` for any HTTP/1.1 request, so it calls `receiveHead()` again and feeds the JSON body to the head parser. That blocks the worker thread in `fillMore()` until the peer closes or 16KB arrive (`HttpHeadersOversize`). It was already reachable via 405/413, but this PR makes it the routine outcome for a real client POST with a wrong `Content-Type` or a `localhost` Host.

Fix is in `handleConn` (outside this diff, so no suggestion): honour std's closing state.

```zig
        self.serve(&out.writer, arena.allocator(), &request) catch return;
        if (!request.head.keep_alive or http_server.reader.state == .closing) return;
```

arrufat: Fix for this in #3268 (independent of this PR, merges in either order).

krichprollsch:
> Every documented way of reaching this endpoint now gets a bodyless 403: parseLiteral("localhost:8000")**

We could authorize explicitly `localhost` WDYT?

arrufat:
> > Every documented way of reaching this endpoint now gets a bodyless 403: parseLiteral("localhost:8000")**
> 
> We could authorize explicitly `localhost` WDYT?

yes

karlseguin:
I rebased to leverage the smaller memory footprint of `main` so the CI would pass. Pretend I wasn't here.

PR #3260, opened 2026-08-24
stability: drop async module completions after termination

staylor:
## Summary

Do not let an async `import()` source completion re-enter V8 after the watchdog or another caller has requested termination. A successful late source is released without compilation; an error completion is ignored because the browser is already tearing down.

## Root cause

The sticky `terminate_requested` gate added around JS functions, scripts, and microtask checkpoints does not cover `dynamicModuleSourceCallback`. If an async module fetch completes while the CDP tick is unwinding a terminated page, this callback opens a V8 scope and can compile/evaluate the module or settle its resolver before `CDP.tick` observes the pending termination and closes the connection.

A deterministic fixture demonstrates the ordering on current `main`:

```
watchdog stall
MutObserver.deliverRecords err=ExecutionTerminated
frame.deliverMutations err=ExecutionTerminated
console.log ASYNC_MODULE_EVALUATED
closing connection reason="pending terminate"
```

That violates the termination invariant directly. It also matches a production crash where pending async module transfers completed after the mutation-delivery termination and V8 then failed its module-evaluation status check.

With this patch, the same fixture retains the watchdog/termination/close sequence but never evaluates the late module.

## Test

The regression test leaves the sticky termination request set while clearing only V8's consumed termination state, then delivers a late async-module error. Before the fix, the callback re-enters V8 and rejects the resolver (`expected .pending, found .rejected`). With the fix, it leaves the resolver untouched.

- `zig build ... test -freference-trace` — 1237/1237
- CDP fixture: 5/5 termination cycles; zero post-termination module evaluations (10/10 before the fix)

### DIFF
--- a/src/browser/Frame.zig
+++ b/src/browser/Frame.zig
@@ -1809,12 +1809,17 @@ fn frameErrorCallback(ctx: *anyopaque, err: anyerror) void {
         return;
     };
 }
+
 pub fn isGoingAway(self: *const Frame) bool {
+    return self.js.env.terminatePending() or self.hasQueuedNavigation();
+}
+
+fn hasQueuedNavigation(self: *const Frame) bool {
     if (self._queued_navigation != null) {
         return true;
     }
     const parent = self.parent orelse return false;
-    return parent.isGoingAway();
+    return parent.hasQueuedNavigation();
 }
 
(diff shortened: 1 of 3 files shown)


karlseguin:
Thanks. I added the check in Frame.isGoingAway which closed this issue in a few more cases.

PR #3262, opened 2026-08-24
Screenshot with a pure-Zig renderer

arrufat:
Alternative to #3231: the same screenshot feature, with the rasterizer in Zig instead of Rust. Based on `main` directly so the two can be compared as whole PRs. The DOM walk, CDP wiring, tests and PNG contract are Karl's from #3231 (he is the author of the feature commit; I'm co-author) — only the stage behind the block/span model is different, and the `src/rust` workspace reorganization is not needed, so html5ever stays where `main` has it.

```
zig build run -- fetch "https://lightpanda.io/" --dump png > screenshot.png
```

### What's different

Instead of #3231's `src/rust/render` (parley + fontique + skrifa + tiny-skia + png, ~1.2k lines of Rust plus their dependency tree), the renderer is `src/browser/screenshot/raster.zig` and `gpos.zig` (~1.5k lines of Zig), built on:

- **z2d** for glyph outlines, antialiased fills and PNG-free drawing, consumed from `arrufat/z2d` branch `lightpanda` (upstream + one line exporting glyph metrics; upstream accepts no public PRs, so that line is a permanent carry).
- **ICU inside V8's archive** for bidi (`ubidi_*`), Arabic shaping (`u_shapeArabic`) and UAX #14 line breaking (`ubrk_*`).
- our own GPOS reader for mark attachment (harakat, niqqud, combining Latin), a glyph atlas with quarter-pixel phases, a direct blit, and a streamed PNG encoder through the zlib the binary already links.
- a fourth DejaVu face (Sans Mono Bold, +332KB) instead of stroke-synthesized bold code.

Not needed without the crate: the extern struct mirrors, the panic-catching return-code protocol, the mutex-protected singleton renderer (state is arena-owned per call and covered by the leak-detecting test runner), the Cargo workspace split, and #3231's streaming `Base64Writer` — the compressed PNG is buffered once on the capture's arena and encoded with `std.base64` (a few hundred KB for a text page, against a raster that is 17MB+).

### Fidelity

Line-for-line with #3231 on everything tried: Latin pages (lightpanda.io 4756 vs 4763px tall, same wraps), a style sampler (identical height), Arabic and Hebrew including mixed direction and mirrored brackets, vocalized Arabic with stacked marks, decomposed Latin accents, CJK wrapping. On the like-for-like sampler pages the two renderers differ by a mean of 1.7/255 per pixel, all antialiasing edges. Two things found on the way are posted on #3231 as suggestions: the synthesized italic leans left, and bold mono is worth a real face.

### Performance

Per-phase timers in both renderers, both built in release mode (`cargo --profile release` with LTO for #3231's crate, ReleaseFast for this branch), saved copy of news.ycombinator.com (1920×2233), same machine, performance governor, best of three.

| phase | Rust (release) | Zig (release) |
|---|---:|---:|
| layout (wrap, bidi, marks) | 0.8ms | 1.8ms |
| destination surface | 1.9ms | 3.7ms |
| draw | 14.2ms | 3.0ms |
| PNG encode | 41.6ms | 14.5ms |
| **total** | **~58ms** | **~23ms** |
| ar.wikipedia.org, 1920×16384, total | 405ms | 163ms |

parley's layout and tiny-skia's surface setup are faster than ours; the raster and PNG stages are not. In the browser's own dev builds the two are close (~77ms vs ~68ms), since the Rust crate's dependencies are already optimized there. Peak RSS on a capped-height capture: 276MB (the encode streams scanline-in / IDAT-out). Pixel parity held through the optimization commits (≤3/255 on blended edges).

One finding that applies to `main` regardless: the vendored C libraries compile at the main module's optimize mode, so dev builds run zlib at -O0 (12.9MB deflated in ~150ms vs 13ms at -O2). This branch compiles zlib and z2d optimized in dev builds, the way the Rust workspace already treats its dependencies; brotli and sqlite may deserve the same look.

### Trade-offs, honestly

**For the Rust crate (#3231):** parley + swash are a full OpenType text stack from Linebender — real shaping for every script a font covers (GPOS marks, cursive attachment, LookupFlag filtering, Indic if a font ever provides it), with tiny-skia's rasterizer behind it. Nothing to carry, nothing to explain to a future maintainer beyond one FFI function.

**For this branch:** no FFI boundary, no singleton, five fewer crates, and ~2.5× faster end to end in release (draw and PNG; parley wins layout and surface setup). Shaping is ICU's presentation-form shaper plus our GPOS mark attachment, which covers Arabic, Hebrew and Latin completely with the bundled fonts but is not a general OpenType shaper (no LookupFlag mark-class filtering, no cursive attachment — DejaVu uses neither). z2d is one maintainer's project (responsive, Ghostty-adopted) that takes no PRs, so the carry lives on our fork and leans on an internal module; version bumps need re-validation.

**Operational, and the one real blocker today:** the ICU symbols are in the static V8 archive, but the shared `libc_v8.so` that `dev_fast` builds link hides them behind the fork's version script. lightpanda-io/zig-v8-fork#201 adds the five export lines; until a prebuilt ships with it, this branch builds and tests with `ZIGFLAGS=-Dprebuilt_v8_path=.lp-cache/prebuilt-v8/v0.5.3/libc_v8_14.9.207.35_linux_x86_64.a`. Release builds are unaffected.

### Not in either PR

Indic, Thai and Syriac render as tofu on both — DejaVu Sans has no glyphs for them. Italic is a synthesized shear on both.

### Commits

- screenshot: `--dump png` and `Page.captureScreenshot` with a Zig rasterizer (author Karl, co-author me: the DOM walk, CDP wiring and tests are #3231's)
- build: compile vendored zlib optimized in dev builds

Tests: 1237/1237 (13 in the screenshot filter, including a regression test for stacked marks on a non-first line). Output is byte-identical to the previous seven-commit history of this branch.


src/browser/screenshot/raster.zig
```diff
@@ -0,0 +1,1149 @@
...
+        var window: Window = .{ .buf = dst.buf, .stride = pw_u, .x = 0, .y = 0, .w = pw_u, .h = ph_u };
+        if (opts.clip) |clip| {
+            const cx: u64 = @intFromFloat(std.math.clamp(@floor(clip.x * scale), 0, @as(f64, @floatFromInt(pw - 1))));
+            const cy: u64 = @intFromFloat(std.math.clamp(@floor(clip.y * scale), 0, @as(f64, @floatFromInt(ph - 1))));
+            window = .{
+                .buf = dst.buf,
+                .stride = pw_u,
+                .x = cx,
+                .y = cy,
+                .w = @min(@as(u64, @intFromFloat(@max(@ceil(clip.width * scale), 1.0))), pw_u - cx),
```
karlseguin, line 506: "Page.captureScreenshot {clip:{x:0,y:0,width:1e30,height:10,scale:1}} → integer part of  floating point value out of bounds at raster.zig:506 — .w = @min(@intFromFloat(@max(@ceil(clip.width * scale), 1.0)), pw_u - cx) casts before the @min. Thread panic, whole server SIGABRTs."

karlseguin:
https://github.com/lightpanda-io/browser/pull/3231 was largely Claude-driven. I don't have a strong opinion about this. I don't really understand this code, but it's isolated, so I'm fine with it. @arrufat has more knowledge about this than me, so I'm happy for you to decide.

I don't want to go back and forth on Claude outputs, but mine thought the original is more robust. It said Vocalized Hebrew is broken in this version.

arrufat:
I still think we should merge #3231.
I just wanted to see how far I could take the Zog version and make a PR to leave a trace.

PR #3265, opened 2026-08-25
search: default to Keenable's keyless endpoint, drop the DuckDuckGo scrape

arrufat:
## What changes

**DuckDuckGo is removed.** The `html.duckduckgo.com` endpoint the search tool scraped as its last resort is disallowed by its robots.txt. Gone: the `duckduckgo` engine, the scrape via `performGoto`, and the tool's dependency on a browser session (`execSearch(arena, args)` now).

**Keenable's keyless endpoint is the default search backend.** `.auto` is one walk over the engine table in preference order — Brave, Tavily, Exa, each only when its key is set — with Keenable last: keyed if `KEENABLE_API_KEY` is set, its public endpoint (rate-limited per client IP) otherwise. Search always has a zero-config rung (a comptime assert pins that some engine is keyless), and a configured key is never shadowed by a keyless retry (the #3252 rung fired the public endpoint even after a keyed attempt failed, which masked a bad key). If every rung fails the tool returns `web search failed: <err>`.

## Timeout

With the libcurl scrape gone, every search goes through zenai's `std.http.Client`, which had no timeout: a stalled response would block the MCP server's shared browser thread for every session. lightpanda-io/zenai#10 (merged, pinned here) adds a per-attempt watchdog; the tool's existing `timeout` argument drives it (default 10 s, `default_search_timeout_ms`). Known gap, inherited from std: the connect phase itself is still unbounded (`ConnectTcpOptions.timeout` is unimplemented in Zig 0.16).

HTTP-level retries are off for every search path: in `.auto` the next rung is the retry, in explicit mode the model is, and either way the `timeout` argument then bounds the whole call rather than one of up to four attempts plus `Retry-After` sleeps (60 s each on the public endpoint).

## Cleanups from the #3252 review

- `keyless` derived from the client's `api_key` type (`?[]const u8`) instead of a hand-set table flag that could disagree with it.
- One `engineKey(engine)` resolver (on `lp.environ().getPosix`) for the set / keyless / missing decision, used by the cascade, the explicit path, and the REPL's `searchKeyStatus` — previously three spellings.
- `engineIndex` no longer depends on `SearchEngine` declaration order; the comptime assert block is gone.
- The cascade sentence in the tool description and `/searchEngine` help is one shared constant with the engine order and env-var list generated from the table.
- `searchExplicit` flattened; `writeResultItem` falls back to the URL when a provider returns an empty title (was `N. **** — url`).

## Compatibility note

A `.lp-agent.zon` that persisted `search_engine = .duckduckgo` no longer parses; the whole remembered file is ignored until re-saved. Judged acceptable for a dev tool rather than adding a migration.

## Tests

`make test`: 1241/1241. New: `writeResultItem uses the URL as title when the title is empty`.

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -36,8 +36,8 @@
             .hash = "sqlite3-3.53.2-DMxLWuAOAAA_Px0arJOIOaP4AKEu5prbsQgPMA35W1zz",
         },
         .zenai = .{
-            .url = "git+https://github.com/lightpanda-io/zenai.git#4aa2cccdaa765938fd96e4a9e74a0765362e9a17",
-            .hash = "zenai-0.0.0-iOY_VApeBgDxgim4YNZ8UQd3XkAyUGOIrgAMx2pD6zxs",
+            .url = "git+https://github.com/lightpanda-io/zenai.git#a745bb1b654e150955f7f4d84ddd1e35ed8beb9e",
+            .hash = "zenai-0.0.0-iOY_VI6EBgAryRPO5dH1uS_knQguW5hqcL8EfpnBKBFK",
         },
         .isocline = .{
             .url = "git+https://github.com/arrufat/isocline?ref=lightpanda#832a9fe25f5f4458fcc47b5acc7c21db669c2f47",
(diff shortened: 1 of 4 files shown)


karlseguin:
I tried it, seems to work :)  Code is fine. LGTM

PR #3267, opened 2026-08-25
http: strip Authorization on cross-origin redirects

krichprollsch:


src/network/HttpClient.zig
```diff
@@ -2827,6 +2813,19 @@ pub const Transfer = struct {
...
         };
 
+        // Fetch "HTTP-redirect fetch": when the redirect target is not
+        // same-origin with the current URL, the Authorization header and the
+        // basic-auth credentials must not follow the request to the new
+        // origin. Independent of CORS: applies to every request kind.
+        if (!URL.isSameOrigin(url, req.url)) {
+            transfer.removeHeader("Authorization");
+            // Proxy credentials are tied to the proxy, not the target origin.
+            const proxy_auth = if (transfer._auth_challenge) |c| c.source == .proxy else false;
```
karlseguin: `transfer._auth_challenge` is set to null before this in detectAuthChallenge for any redirect code.  Seems like an existing issue where the proxy credentials are erased prior to a redirect. This proxy_auth check doesnt make it worse...I had to look over the code...I don't understand the flow. If we only have a single `_auth_challenge`, it's impossible to have both one for a proxy and then one for the backend.

PR LGTM, but I'm not sure about the existing auth code in more complex scenarios (e.g. redirects, or proxy + origin auth).

krichprollsch: I removed the auth_challenge cleanup part and updated the comment. Thanks.

I don't think we really need to handle both target and proxy auth challenge at the same time. I agree that could be blocking, but in general now we start proxy w/ the CONNECT  and make the target's request through the opened connection.

karlseguin (approved):

PR #3269, opened 2026-08-25
markdown: skip hidden elements

arrufat:
## Summary

`markdown` rendered `display:none`, `[hidden]`, `aria-hidden="true"`, closed `<details>` content and closed `<dialog>`s: `isVisibleElement` only looked at the tag. Hidden nav drawers, tab panels, cookie modals and SEO blocks all landed in the output of the `markdown` tool, `--dump markdown` and `mcp://page/markdown`.

- `isVisibleElement` now also checks the element's own computed `display:none` (`StyleManager.hasDisplayNone`, which already covers inline style, author rules, `[hidden]`, closed `<details>`) and `aria-hidden="true"`.
- Own state only, and the dump root is exempt, so `markdown {selector}` on a hidden subtree still renders it while hidden descendants are filtered.
- `analyzeContent` / `isStandaloneAnchor` use the same predicate, so an anchor whose only content is hidden falls back to its `aria-label`/`title`. `screenshot.zig` updated for the signature.
- Adds the `dialog:not([open]) { display: none }` UA rule to `StyleManager`, which was missing (closed dialogs were visible to `checkVisibility`/`getComputedStyle`/`tree` too).

## Verification

New tests: inline style, `[hidden]`, `aria-hidden` (true/TRUE/false), closed details/dialog, stylesheet rule (via `dump.html`), anchor label fallback, scoped hidden root; plus a `check_visibility.html` case for `dialog.show()`/`close()`. Full suite passes (1260/1260).

Noticed but out of scope: keyword values are compared case-sensitively in `StyleManager` (`style="DISPLAY: NONE"` is missed), separate PR.

### DIFF
--- a/src/SemanticTree.zig
+++ b/src/SemanticTree.zig
@@ -131,7 +131,7 @@ fn walk(
         if (tag == .datalist or tag == .option or tag == .optgroup) return;
 
         // Check visibility using the engine's checkVisibility which handles CSS display: none
-        if (!el.checkVisibilityCached(ctx.visibility_cache, self.frame)) {
+        if (!el.checkVisibilityCached(ctx.visibility_cache, self.frame, .scan)) {
             return;
         }
 
--- a/src/browser/StyleManager.zig
+++ b/src/browser/StyleManager.zig
@@ -644,6 +644,9 @@ fn matchesUaDisplayNoneRule(el: *Element) bool {
         }
     }
 
+    // dialog:not([open]) { display: none }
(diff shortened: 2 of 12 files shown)


karlseguin:
@arrufat on a large dom, this ends up ballooning `Frame._element_styles`. So I made the StyleManager take a comptime `.materialize` (existing before) or `.scan` (only parse the style, don't cache it) and changed all the one-time dumps to use `.scan`

PR #3271, opened 2026-08-25
mem: On linux CI builds, use an orderfile

karlseguin:
CI build, for e2e-test and nightly now build with: -Dorderfile/lightpanda.ld

This file informs the build on how to organize the code in the binary, grouping hot code together so that we have to load less of the binary into memory.

lightpanda.ld will drift: we'll refactor our code, add new features, update dependencies, update Zig, ... So it has to be re-generated. But we can do that automatically in the CI (say, before the nightly build). That's for a follow up PR.

This does not currently cover V8. V8 is being build with `-no-unique-section-names`, so we don't get names that we can correctly organize. The real win comes from doing this in V8, since a lot of V8 is cold. This PR can land as-is, a zig-v8-fork PR will remove that flag, and then we can have a follow up PR with an lightpanda.ld that includes the v8 symbols.

This is opt-in (via the -Dorderfile flag) because it adds ~20 seconds of linking time.


karlseguin:
PR
VmHWM: 26380

Main
VmHWM: 27912

PR #3276, opened 2026-08-25
mcp: add title and annotations to tools/list

arrufat:
`tools/list` entries carried only `name`/`description`/`inputSchema`. Clients use `title` for display and `annotations.readOnlyHint` to skip approval prompts on reads.

- `protocol.Tool` gains `title` and `annotations` (`ToolAnnotations` with the spec defaults: not read-only, destructive, not idempotent, open world).
- Browser tools take `title` from the existing `summary`; `annotations` come from an exhaustive switch in `mcp/tools.zig`, so a new tool must classify itself. Read-only (cannot change page or session state): `nodeDetails`, `findElement`, `getUrl`, `getCookies`, `getEnv`, `extract`, `session_list`; `consoleLogs` is read-only but not idempotent since it drains its buffer. Navigation (`goto`, `search`, and the readers when given a `url`) writes cookies/storage and waits let page scripts run, so they are not read-only, only non-destructive. `evaluate`/`click`/`press` keep the destructive default; `fill`/`selectOption`/`setChecked`/`hover`/`scroll` are non-destructive.
- The four extra tools get titles and annotations as well; `save` (overwrites a file) and `session_close` are destructive.

Tests: field serialization, and a `tools/list` check of titles/hints for a sample of tools. Full suite passes.

### DIFF
--- a/src/mcp/tools.zig
+++ b/src/mcp/tools.zig
@@ -29,12 +29,16 @@ const browser_tool_list = blk: {
 
 const read_only: protocol.ToolAnnotations = .{ .readOnlyHint = true, .destructiveHint = false, .idempotentHint = true, .openWorldHint = false };
 
-/// Exhaustive so a new tool must classify itself. Navigation is an HTTP GET,
-/// so tools that only navigate and read stay read-only; they are open-world.
+/// Exhaustive so a new tool must classify itself. Read-only means the call
+/// cannot change page or session state: navigation writes cookies and
+/// storage, and waiting lets page scripts run, so neither qualifies.
 fn annotations(tool: BrowserTool) protocol.ToolAnnotations {
     return switch (tool) {
-        .nodeDetails, .findElement, .consoleLogs, .getUrl, .getCookies, .getEnv, .extract, .waitForSelector, .waitForScript, .waitForState => read_only,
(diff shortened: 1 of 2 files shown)


src/mcp/tools.zig
```diff
@@ -18,13 +18,29 @@ const browser_tool_list = blk: {
...
         };
     }
     break :blk tools;
 };
 
+const read_only: protocol.ToolAnnotations = .{ .readOnlyHint = true, .destructiveHint = false, .idempotentHint = true, .openWorldHint = false };
+
+/// Exhaustive so a new tool must classify itself. Navigation is an HTTP GET,
+/// so tools that only navigate and read stay read-only; they are open-world.
+fn annotations(tool: BrowserTool) protocol.ToolAnnotations {
```
karlseguin, line 35: I don't know what part is supposed to be "readonly", but `waitForSelector`, `waitForScript`, `waitForState` and `navigate` can all "write" to the system...they can cause cookies to be written, or local storage...and they can cause the state of the system to change. Doesn't seem "readonly" to me.

Also , `consoleLog` "Returns all messages since last call and clears the buffer"...so it should be `.idempotentHint = false`

arrufat: Agreed on both: navigation and waits can mutate cookies/storage/page state, so they're now non-destructive but not read-only; consoleLogs drain their buffer, so idempotentHint is false. Read-only is down to the pure current-state reads (nodeDetails, findElement, getUrl, getCookies, getEnv, extract, session_list).

karlseguin (approved):

PR #3278, opened 2026-08-25
Make subframe defaults explicit by mode

### LINKED ISSUE #3139: Default subframes off in fetch/agent/mcp, log a fidelity manifest at startup
Since #3085 the engine loads dynamically injected script chains. That's correct, and it makes ad iframes much more expensive: one apnews hub-page load went from 16 script fetches / 1 frame navigation to 183 / 106. News task numbers (4 page loads, load waits, ReleaseFast e14353398):

| | pre-#3085 | post-#3085 | post-#3085 + `--disable-subframes` |
| :--- | ---: | ---: | ---: |
| wall time | ~3-4 s | ~13-14 s | ~9 s |
| peak PSS | ~120 MB | ~255 MB | ~237 MB |

`--disable-subframes` went from about -9% to about -45% on this task (the frames it skips used to not load at all). For fetch/agent/mcp, where the output is data and none of it comes from ad iframes, that's a lot to leave behind a flag most users won't find.

Tested and not viable as alternatives: raising `--http-max-host-open`/`--http-max-concurrent` (6/40 to 24/120) changed nothing (the chain is latency-bound, not pool-bound), and `--block-urls` with the big ad hosts recovered little (too many independent chains).

Proposal (came out of an internal discussion):

1. Flip the default to subframes-off in `fetch`, `agent`, `mcp`. Add `--enable-subframes`. Keep accepting `--disable-subframes` as a no-op so existing invocations don't break.
2. `serve` keeps subframes on: CDP consumers expect `frameAttached` and frame locators to work (precedent for per-mode defaults: `--enable-external-stylesheets`). Leaving this as an open question in case someone prefers one default everywhere.
3. Log a one-line fidelity manifest at startup so the trade is visible, e.g.
   `fidelity: subframes=off images=never external-stylesheets=off workers=on`
   (first-line answer when someone's iframe-dependent page breaks)

Benchmark evidence: `v7`/`v7py` datasets in [agent-benchmarks/pandascript-vs-cdp](https://github.com/lightpanda-io/agent-benchmarks/tree/main/pandascript-vs-cdp). Sibling issue for the wait-semantics half: #3138

fuyua9:
## Summary

- Make subframe loading defaults explicit by mode: serve keeps subframes enabled, while fetch, agent, and mcp default to disabled.
- Add a compatibility `--enable-subframes` flag for data-oriented modes and document the transition.
- Emit the effective mode and subframe state in the startup fidelity log.

Fixes #3139

## Tests

- `zig fmt --check src/Config.zig src/main.zig` passed.
- `zig ast-check src/Config.zig` passed.
- `zig ast-check src/main.zig` passed.
- `make test F='Config: subframe'` could not reach compilation because dependency downloads from `build.zig.zon` failed with `HttpConnectionClosing`.

The focused regression test is included in `src/Config.zig`, but could not execute until the dependency fetch succeeds.

## Real Behavior Proof

- Behavior or issue addressed: Mode-specific subframe defaults and an explicit startup fidelity signal for issue #3139.
- Real environment tested: Local macOS checkout at commit `d52229b8e263bda75bb6fb66183e024ea2a95d17`.
- Exact steps or command run after this patch: `zig fmt --check src/Config.zig src/main.zig`; `zig ast-check src/Config.zig`; `zig ast-check src/main.zig`.
- Evidence after fix: Formatting and AST checks passed; the focused Config regression test is present but the build did not reach compilation because dependency fetching failed.
- Observed result after fix: The committed configuration maps serve to subframes-on and fetch/agent/mcp to subframes-off, with `--enable-subframes` restoring data-mode loading; startup logs expose the effective mode and state.
- What was not tested: Full `make test` and runtime startup execution remain unverified because the dependency fetch failed with `HttpConnectionClosing`.

### DIFF
--- a/src/Config.zig
+++ b/src/Config.zig
@@ -241,6 +241,7 @@ const CommonOptions = .{
     .{ .name = "cookie", .type = ?[]const u8 },
     .{ .name = "cookie_jar", .type = ?[]const u8 },
     .{ .name = "disable_subframes", .type = bool },
+    .{ .name = "enable_subframes", .type = bool },
     .{ .name = "disable_workers", .type = bool },
     .{ .name = "enable_external_stylesheets", .type = bool },
     .{ .name = "v8_flags_unsafe", .type = ?[]const u8 },
@@ -495,7 +496,8 @@ pub fn obeyRobots(self: *const Config) bool {
 
 pub fn disableSubframes(self: *const Config) bool {
     return switch (self.mode) {
-        inline .serve, .fetch, .mcp, .agent => |opts| opts.disable_subframes,
+        .serve => |opts| opts.disable_subframes,
+        inline .fetch, .mcp, .agent => |opts| !opts.enable_subframes,
         else => unreachable,
     };
 }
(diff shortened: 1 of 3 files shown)


karlseguin:
@krichprollsch what would you think of:
1 - deprecating  `--disable-subframes`
2 - deprecating `--disable-workers`
3 - deprecating `--enable-external-stylesheets`

And then having them all under the new `--load-resources` (https://github.com/lightpanda-io/browser/pull/3230), e.g. `--load-resources iframe --load-resources stylesheet --load-resources worker --load-resources image ` 

AND having them all **off** by default ?  On startup, if any of them are off, we can `log.warn(...)`.

krichprollsch:
So we will break only subframes which is on by default isn't it?

It seems possible.

karlseguin:
> So we will break only subframes which is on by default isn't it?
> 
> It seems possible.

And workers, they're on by default too.

krichprollsch:
The original idea was to keep on by default what browsers have enable by default too.
But if it doesn't make sense for agent (which is the more browser~ish use case), maybe it's better to disable them by default everywhere.

We have to make sure it's still possible to enable these feature via CLI of course but also via CDP (and Bidi?)

WDYT @arrufat?

arrufat:
I agree with @karlseguin: fold everything under --load-resources (iframe, worker, stylesheet, and image) and default them all off in every mode, serve included.

karlseguin:
superseded by https://github.com/lightpanda-io/browser/pull/3305

PR #3279, opened 2026-08-25
interactive: skip hidden elements

arrufat:
`walkInteractive` filtered only on tag and `pointer-events: none`, so `interactiveElements` and `findElement` listed controls inside `display:none` / `[hidden]` subtrees (collapsed menus, closed dialogs, inactive tabs). `tree` already filters with `checkVisibilityCached`, so the two disagreed about the same page.

Same cached check now runs per element, and a hidden subtree is skipped whole via `skipChildren`.

Test: inline `display:none`, `[hidden]`, and both under a hidden ancestor are dropped; the visible sibling stays. Full suite passes.

### DIFF
--- a/src/browser/interactive.zig
+++ b/src/browser/interactive.zig
@@ -179,25 +179,36 @@ fn walkInteractive(
     const listener_targets = try buildListenerTargetMap(frame, arena);
 
     var css_cache: Element.PointerEventsCache = .empty;
-    var visibility_cache: Element.VisibilityCache = .empty;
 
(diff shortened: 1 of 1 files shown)


karlseguin:
@arrufat I rebased AND..

If we skip all children of an invisible element, than any element we do visit has to have a visible parent. Therefore we only have to check if an element itself is visible. (The root element still has the full `checkVisibilityCached` check though). I made that change, so feel free to merge if yo agree.

PR #3284, opened 2026-08-26
XHR Blob Response Type

mookums:
This adds support for `Blob` to be a responseType of an XHR response.

### DIFF
--- a/src/browser/webapi/net/XMLHttpRequest.zig
+++ b/src/browser/webapi/net/XMLHttpRequest.zig
@@ -31,6 +31,7 @@ const Frame = @import("../../Frame.zig");
 const Node = @import("../Node.zig");
 const Event = @import("../Event.zig");
 const EventTarget = @import("../EventTarget.zig");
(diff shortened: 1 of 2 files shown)


src/browser/webapi/net/XMLHttpRequest.zig
```diff
@@ -460,6 +463,12 @@ pub fn getResponse(self: *XMLHttpRequest, exec: *const Execution) !?Response {
             }
         },
         .arraybuffer => .{ .arraybuffer = .{ .values = data } },
+        .blob => blk: {
+            const mime = self._override_mime orelse self._response_mime;
+            const content_type = if (mime) |m| m.contentTypeString() else "";
+            const blob = try Blob.initFromBytes(data, content_type, exec);
+            break :blk .{ .blob = blob };
```
karlseguin, line 550: Blob is reference counted. We also pin it in `self._response`. If the v8 finalizer fires on the blob we hand to it, and `response` is read again, it's a UAF.

PR #3291, opened 2026-08-26
links: dedup by href, skip hidden, fall back to alt/aria-label/title

arrufat:
`links` returned every `a[href]` verbatim: hidden nav entries, one row per duplicate href, and `text: null` for image/icon links although `markdown` already falls back to `alt`. Nav-heavy pages produced thousands of rows for an agent looking for one link.

- One entry per resolved href; the first occurrence is kept and its text upgraded from a later duplicate if it had none.
- Hidden anchors are skipped, with the same cached visibility check `tree` and `interactiveElements` use.
- Text falls back to `aria-label` → `title` → first descendant `img[alt]`, matching markdown's anchor rules.
- The tool gains an optional `limit` (document order).

Tests in `links.zig` for each rule and an MCP-level test with a small fixture. Full suite passes.

### DIFF
--- a/src/browser/links.zig
+++ b/src/browser/links.zig
@@ -18,12 +18,11 @@
 
 const std = @import("std");
 
+const AXNode = @import("../cdp/AXNode.zig");
 const Element = @import("webapi/Element.zig");
 const Node = @import("webapi/Node.zig");
 const Frame = @import("Frame.zig");
 const Selector = @import("webapi/selector/Selector.zig");
-const TreeWalker = @import("webapi/TreeWalker.zig");
-const interactive = @import("interactive.zig");
 const log = @import("../lightpanda.zig").log;
 
 const Allocator = std.mem.Allocator;
@@ -59,18 +58,18 @@ pub fn registerNodes(links: []Link, registry: anytype) !void {
 }
 
 /// Collect the visible links (anchor tags with an href) under `root`, one
-/// entry per resolved href.
+/// entry per resolved href. `text` is the accessible name, the same one the
+/// semantic tree reports for the node.
 pub fn collectLinks(arena: Allocator, root: *Node, frame: *Frame) ![]Link {
-    var links: std.ArrayList(Link) = .empty;
-    var by_href: std.StringHashMapUnmanaged(usize) = .empty;
+    var links: std.StringArrayHashMapUnmanaged(Link) = .empty;
     var visibility_cache: Element.VisibilityCache = .empty;
 
     if (Selector.querySelectorAll(root, "a[href]", frame)) |list| {
         defer list.deinit(frame._page);
 
         for (list._nodes) |node| {
-            const el = node.is(Element) orelse continue;
             const anchor = node.is(Element.Html.Anchor) orelse continue;
+            const el = anchor.asElement();
             if (!el.checkVisibilityCached(&visibility_cache, frame, .scan)) continue;
 
             const href = anchor.getHref(frame) catch |err| {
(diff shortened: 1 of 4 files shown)


src/browser/links.zig
```diff
@@ -57,24 +58,36 @@ pub fn registerNodes(links: []Link, registry: anytype) !void {
...
             const anchor = node.is(Element.Html.Anchor) orelse continue;
+            if (!el.checkVisibilityCached(&visibility_cache, frame, .scan)) continue;
+
             const href = anchor.getHref(frame) catch |err| {
                 log.err(.app, "resolve href failed", .{ .err = err });
                 continue;
             };
             if (href.len == 0) continue;
 
+            const text = linkText(el, arena);
```
karlseguin: You don't need to do this if you get a hit with a text.

src/browser/links.zig
```diff
@@ -57,24 +58,36 @@ pub fn registerNodes(links: []Link, registry: anytype) !void {
     }
 }
 
-/// Collect all links (anchor tags with an href) under `root`.
+/// Collect the visible links (anchor tags with an href) under `root`, one
+/// entry per resolved href.
 pub fn collectLinks(arena: Allocator, root: *Node, frame: *Frame) ![]Link {
     var links: std.ArrayList(Link) = .empty;
+    var by_href: std.StringHashMapUnmanaged(usize) = .empty;
```
karlseguin: could combine `by_href` and `links` in a single `array_hash_map.String`

src/browser/links.zig
```diff
@@ -86,6 +99,22 @@ pub fn collectLinks(arena: Allocator, root: *Node, frame: *Frame) ![]Link {
     return links.items;
 }
 
+// Same fallbacks markdown uses for an anchor without text.
+fn linkText(el: *Element, arena: Allocator) ?[]const u8 {
+    if (interactive.getTextContent(el.asNode(), arena) catch null) |text| return text;
+    if (el.getAttributeSafe(comptime .wrap("aria-label"))) |label| return label;
+    if (el.getAttributeSafe(comptime .wrap("title"))) |title| return title;
+
+    var tw: TreeWalker.FullExcludeSelf = .init(el.asNode(), .{});
```
karlseguin: `TreeWalker.FullExcludeSelf.Element` will only yield elements.

src/browser/links.zig
```diff
@@ -57,24 +58,36 @@ pub fn registerNodes(links: []Link, registry: anytype) !void {
...
 pub fn collectLinks(arena: Allocator, root: *Node, frame: *Frame) ![]Link {
     var links: std.ArrayList(Link) = .empty;
+    var by_href: std.StringHashMapUnmanaged(usize) = .empty;
+    var visibility_cache: Element.VisibilityCache = .empty;
 
     if (Selector.querySelectorAll(root, "a[href]", frame)) |list| {
         defer list.deinit(frame._page);
 
         for (list._nodes) |node| {
+            const el = node.is(Element) orelse continue;
```
karlseguin: probably cheaper to to call .asElement() on `anchor`

src/browser/links.zig
```diff
@@ -86,6 +99,22 @@ pub fn collectLinks(arena: Allocator, root: *Node, frame: *Frame) ![]Link {
     return links.items;
 }
 
+// Same fallbacks markdown uses for an anchor without text.
+fn linkText(el: *Element, arena: Allocator) ?[]const u8 {
```
karlseguin: markdown has `getAnchorLabel`, interactive has `getAccessibleName`.  Is there no single consistent text for a link?

src/browser/links.zig
arrufat: Done in 5870ce1, resolved after the lookup and only when the existing entry has no text.

src/browser/links.zig
arrufat: Done, `StringArrayHashMapUnmanaged(Link)` + `values()`.

src/browser/links.zig
arrufat: Done.

src/browser/links.zig
arrufat: Gone with `linkText`, no more TreeWalker here.

src/browser/links.zig
arrufat: There is: `AXNode.getName`, which `tree`/`findElement` use. `links` now uses it too. `markdown` differs on purpose (it renders the children, only needs a fallback label). `interactive.getAccessibleName` is the real outlier (title before text, no aria-labelledby/label), I would move it to `AXNode.getName` in a follow-up since it changes `interactiveElements` output.

karlseguin (approved):

PR #3292, opened 2026-08-26
fetch: add --selector and --max-bytes, honor --strip-mode for markdown

arrufat:
`--strip-mode` was passed to the html dump only; `--dump markdown` got a literal `.{}`, so the flag silently did nothing. `fetch` also had no way to scope a dump to one element or cap its size, although the `markdown`/`html` tools have both, and one failing URL in a multi-URL run aborted the whole output.

- `--dump-selector <css>`: dump the first matching element instead of the document, in every dump mode (html via `dump.deep`, markdown, png, semantic tree).
- `--dump-max-bytes <n>`: caps html and markdown with the same UTF-8-safe truncation and `[truncated]` marker the tools use. `dump.Opts` now takes `max_bytes` like `markdown.Opts`, so `root`/`deep` cap themselves and the html tool and fetch no longer wrap the writer by hand. Refused up front for png/semantic tree.
- `markdown` honors `--strip-mode` through `dump.shouldStripElement`; only `ui` changes its output since scripts, styles and hidden elements are never rendered.
- Per-page failures: a wait (`Timeout`), navigation (`CouldntConnect`, …) or dump (`SelectorNotFound`) failure is recorded for that URL, every page is still written, and the `--json` envelope carries it under `"error"` (`null` otherwise). The process then exits 1. The `--wait-ms` budget is recomputed per page instead of once before the loop. A non-JSON single-URL timeout now prints the page before exiting 1.
- `--fail-on-http-error`: any status >= 400 exits 22 (curl's code) after the dump is written.

Verified end to end against a local fixture: selector scoping in every mode and with `--json`; `--strip-mode ui`; both caps; two URLs where only one matches the selector (complete JSON, `"error":"SelectorNotFound"` on the second, exit 1); two URLs with an unmet `--wait-selector --wait-ms 800` finishing in ~1 s total; a dead host next to a good URL; 404 with and without the flag. Unit tests for `strip.ui`, `dump` capping, and the envelope. Full suite passes.

### DIFF
--- a/src/browser/markdown.zig
+++ b/src/browser/markdown.zig
@@ -28,9 +28,13 @@ const Slot = @import("webapi/element/html/Slot.zig");
 
 const isAllWhitespace = @import("../string.zig").isAllWhitespace;
 const LimitedWriter = @import("../LimitedWriter.zig");
+const Strip = @import("dump.zig").Opts.Strip;
 
 pub const Opts = struct {
     max_bytes: ?u32 = null,
+    // Only `ui` (images) applies: scripts, styles and hidden elements are
+    // never rendered.
+    strip: Strip = .{},
 };
 
 const truncation_marker = LimitedWriter.truncation_marker;
@@ -148,6 +152,7 @@ const Context = struct {
     writer: *std.Io.Writer,
     frame: *Frame,
     root: *Node,
+    strip: Strip,
 
     // When there's a slot-attribute, we skip rendering, unless this flag has
     // bet set to true.
@@ -318,6 +323,7 @@ const Context = struct {
                 return;
             },
             .img => {
+                if (self.strip.ui) return;
                 try self.writer.writeAll("![");
                 if (el.getAttributeSafe(comptime .wrap("alt"))) |alt| {
                     try self.escape(alt);
@@ -524,6 +530,7 @@ pub fn dump(node: *Node, opts: Opts, writer: *std.Io.Writer, frame: *Frame) !voi
             .writer = &lw.writer,
             .frame = frame,
             .root = node,
+            .strip = opts.strip,
         };
         ctx.render(node) catch |err| switch (err) {
             error.WriteFailed => {
@@ -543,6 +550,7 @@ pub fn dump(node: *Node, opts: Opts, writer: *std.Io.Writer, frame: *Frame) !voi
         .writer = writer,
         .frame = frame,
         .root = node,
+        .strip = opts.strip,
     };
     try ctx.render(node);
     if (!ctx.state.last_char_was_newline) {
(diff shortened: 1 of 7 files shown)


src/help.zon
```diff
@@ -84,12 +84,21 @@
     \\          prints one object; multiple URLs print {{"results": [ ... ]}} with
     \\          one object per URL. When used with --dump <MODE> the dumped
     \\          content is wrapped within each object.
+    \\  --max-bytes <INT>
```
karlseguin:
```suggestion
    \\  --dump-max-bytes <INT>
```

src/help.zon
```diff
@@ -84,12 +84,21 @@
...
+    \\          Soft cap on the dumped html or markdown, in bytes. Output is cut
+    \\          at a UTF-8 boundary and a '[truncated]' marker is appended.
+    \\          Defaults to no cap.
     \\  --metrics
     \\          Write metrics (Prometheus text format) to stdout on exit, after
     \\          any --dump output.
     \\          Defaults to false.
+    \\  --selector <QUERY>
```
karlseguin:
```suggestion
    \\  --dump-duselector <QUERY>
```

src/browser/markdown.zig
```diff
@@ -318,6 +323,7 @@ const Context = struct {
                 return;
             },
             .img => {
+                if (self.strip.ui) return;
```
karlseguin: Could we expose and re-use dump's `shouldStripElement` ?

src/main.zig
```diff
@@ -163,13 +163,20 @@ fn run(allocator: Allocator, main_arena: Allocator, proc_args: std.process.Args)
                 });
             }
 
+            if (opts.max_bytes != null and opts.dump != .html and opts.dump != .markdown) {
```
karlseguin: --max-byte 10  without a --dump is an error, but --selector '.hi' without a --dump is ok.

src/lightpanda.zig
```diff
@@ -372,20 +379,38 @@ pub fn fetch(app: *App, browser: *Browser, urls: []const [:0]const u8, opts: Fet
...
+            const result = if (opts.selector == null)
+                dump.root(frame.window._document, opts.dump, &lw.writer, frame)
+            else
+                dump.deep(root, opts.dump, &lw.writer, frame);
+            result catch |err| {
+                if (!lw.truncated) return err;
+                try writer.writeAll(LimitedWriter.truncation_marker);
+            };
+        },
+        .markdown => try markdown.dump(root, .{ .max_bytes = opts.max_bytes, .strip = opts.dump.strip }, writer, frame),
```
karlseguin: markdown is max_bytes aware, should dump be too?

karlseguin (commented):
Works:
```
zig build run -- fetch --selector "#theme" https://www.openmymind.net/ --json --dump
```

errors with a partial dump:
```
zig build run -- fetch --selector "#theme" https://www.openmymind.net/ https://www.example.com --json --dump
```

src/help.zon
arrufat: Applied; Config/main renamed to match in 1fb7a1b98.

src/help.zon
arrufat: Applied as `--dump-selector` (typo in the suggestion); Config/main renamed in 1fb7a1b98.

src/browser/markdown.zig
arrufat: Done in e46964e26: `shouldStripElement` is pub, takes `Strip`, markdown calls it.

src/main.zig
arrufat: Fixed: both now require `--dump`.

src/lightpanda.zig
arrufat: Done in 56241e26b: `dump.Opts.max_bytes`, `root`/`deep` cap themselves; html tool and fetch no longer wrap by hand.

arrufat:
Two-URL case fixed in 6811cc822: failures are per page, the second URL gets `"error":"SelectorNotFound"` in complete JSON and the process exits 1.

PR #3293, opened 2026-08-26
compute site-for-cookies from ancestor chain of a `Frame`

nikneym:
The site-for-cookies were computed from the immediate parent `Frame`, which allow sending a cookie that's `SameSite=Strict` from 2 levels deep under. Directly from RFC6265bis, this PR essentially implements (except for step 4, we skip host-less ancestors):

   Given a Document (document), the following algorithm returns its
   "site for cookies":

   1.  Let top-document be the active document in document's navigable's
       top-level traversable.

   2.  Let top-origin be the origin of top-document's URI if top-
       document's sandboxed origin browsing context flag is set, and
       top-document's origin otherwise.

   3.  Let documents be a list consisting of the active documents of
       document's inclusive ancestor navigables.

   4.  For each item in documents:

       1.  Let origin be the origin of item's URI if item's sandboxed
           origin browsing context flag is set, and item's origin
           otherwise.

       2.  If origin is not same-site with top-origin, return an origin
           set to an opaque origin.

   5.  Return top-origin.


karlseguin (commented):
The same logic should be applied to document.getCookies and Workers then?

src/browser/webapi/storage/Cookie.zig
```diff
@@ -575,11 +584,13 @@ pub const Jar = struct {
         request_time: ?u64 = null,
         is_navigation: bool = true,
         prefix: ?[]const u8 = null,
-        origin_url: ?[:0]const u8 = null,
+        // null means there is no initiating document (a browser-initiated
```
karlseguin: but browser-initiated requests _do_ pass a `cookie_origin` . This comment seems wrong then? So now I'm curious when it is/should be null?

src/browser/webapi/storage/Cookie.zig
```diff
@@ -641,15 +652,20 @@ fn areCookiesEqual(a: *const Cookie, b: *const Cookie) bool {
     return true;
 }
 
-fn areSameSite(origin_url_: ?[:0]const u8, target_host: []const u8) !bool {
-    const origin_url = origin_url_ orelse return true;
-    const origin_host = URL.getHostname(origin_url);
+pub fn areSameSite(maybe_origin_url: ?SiteForCookies, target_host: []const u8) bool {
+    // No initiating document (browser-initiated request).
```
karlseguin: Repeats a (I think wrong) comment

src/browser/webapi/storage/Cookie.zig
```diff
@@ -1017,6 +1033,49 @@ test "Jar: forRequest SameSite=Strict on cross-site navigation" {
...
+    const now = lp.datetime.timestamp(.real);
+    const victim_url: [:0]const u8 = "https://victim.example/";
+    try jar.add(try Cookie.parse(testing.allocator, victim_url, "strict=1; Path=/; SameSite=Strict"), now, true);
+    try jar.add(try Cookie.parse(testing.allocator, victim_url, "lax=2; Path=/; SameSite=Lax"), now, true);
+    try jar.add(try Cookie.parse(testing.allocator, victim_url, "none=3; Path=/; SameSite=None; Secure"), now, true);
+
+    // .none is the site-for-cookies of a frame whose ancestor chain contains
+    // a cross-site document. Even though the target here is the cookies' own
+    // site, the request is cross-site: Strict is withheld. (Lax still rides
+    // navigations; whether a subframe load should count as one is #240.)
```
karlseguin: Took me a bit to find the #240 reference. I don't think a bare hashtag to a different project is useful.

karlseguin (commented):
Can be merged once the comments are corrected.

PR #3311, opened 2026-08-27
interactive: take roles and names from AXNode

arrufat:
`findElement` and `interactiveElements` computed roles from their own tag table and names from a short attribute chain that ended in text content, so a control the tree showed as `textbox "Email address"` (via `<label for>`) could not be found by that name — the exact workflow the MCP guidance recommends.

- Roles and names now come from `AXNode` (the same code `tree` uses); the tag table and `getAccessibleName` are gone. An element with no accessibility role (interactive only by listener or `tabindex`) keeps its text content as its name, so clickable divs stay findable.
- `AXNode.getName` takes a `LabelByForIndex` (built lazily on the first `<label for>` lookup), so a walk resolves labels with one document scan instead of one per control; `interactiveElements` and the semantic tree walk share one across the walk, single-node callers declare one. The unindexed `findLabelByFor` scan is gone — a fresh index costs the same single walk.

Verified over MCP: `findElement {role: textbox, name: "Email address"}` and `{name: remember}` (wrapping label) both resolve. Tests cover label-for, wrapping label, aria-labelledby, placeholder, and the no-role fallback. Full suite passes.

### DIFF
--- a/src/browser/interactive.zig
+++ b/src/browser/interactive.zig
@@ -222,8 +222,10 @@ fn walkInteractive(
             if (!std.ascii.eqlIgnoreCase(r, rf)) continue;
         }
 
-        // Names walk labels and text; filter on role first.
-        const name = (try axn.getName(frame, arena, &label_index)) orelse try getTextContent(node, arena);
+        // Names walk labels and text; filter on role first. Role-less elements
+        // (listener/tabindex only) sit outside AccName; their text is the only handle.
+        const name = try axn.getName(frame, arena, &label_index) orelse
+            if (role == null) try getTextContent(node, arena) else null;
         if (filter.name) |nf| {
             const n = name orelse continue;
             if (std.ascii.indexOfIgnoreCase(n, nf) == null) continue;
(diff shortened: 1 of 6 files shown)


src/browser/interactive.zig
```diff
@@ -212,15 +215,15 @@ fn walkInteractive(
...
         if (filter.role) |rf| {
             const r = role orelse continue;
             if (!std.ascii.eqlIgnoreCase(r, rf)) continue;
         }
 
-        // Resolve accessible name only after the cheap role filter passes,
-        // since getAccessibleName walks the element's text subtree.
-        const name = try getAccessibleName(el, arena);
+        // Names walk labels and text; filter on role first.
+        const name = (try axn.getName(frame, arena, &label_index)) orelse try getTextContent(node, arena);
```
karlseguin: Only following along a little, but...if the goal here is to unify shouldn't this either not orelse or the orelse should be in axn.getName?

karlseguin (approved):

src/browser/interactive.zig
arrufat: Good catch. The bare `orelse` was leaking text-content names onto roled elements the tree leaves unnamed; it's now scoped to role-less listener/tabindex elements, where the text is the only handle.

PR #3312, opened 2026-08-27
mcp: fold Server.Session onto lp.ToolSession, bracket every isolate use

arrufat:
Follow-up deferred from #3096: the c-api branch defines a `ToolSession` inline in lightpanda.zig for the Browser/Session/Notification/node Registry quartet, and mcp/Server.zig carries its own copy of the same init/teardown sequence. Landing the shared type on main first, so the c-api branch only drops its inline copy when it merges (same init/deinit/enterIsolate/exitIsolate API).

- `src/ToolSession.zig`: the quartet with its ordering-sensitive init/teardown and isolate enter/exit, re-exported as `lp.ToolSession`.
- mcp/Server.zig: no more `Session` wrapper. `sessions` maps straight to `*lp.ToolSession`, the id is only the map key.
- Every isolate use is bracketed now, stdio included. Before, stdio kept its isolate permanently entered and only HTTP parked between requests, so a handler touching page JS without `enterIsolate` passed the stdio-mode tests (56 of 57) and only failed under `--port`. `park_isolates` survives as `multi_session`, which is what it actually gated: the session tools need a transport that routes by `Mcp-Session-Id`.
- Three tests that reach into the page directly (`registry.register`, `frame.js`) bracket the isolate; `register` creates V8 handles.
- A direct `ToolSession` test: two sessions on one thread, entered only around their own use, separate JS globals, balanced teardown. That is what c_api.zig relies on without Server in the loop.

Not in this PR: Agent.zig, see #3313 (stacked on this one).

Verified: full suite (1282/1282); `lightpanda mcp` over stdio with goto, evaluate, a refused session_new, clean exit.

### DIFF
--- a/src/ToolSession.zig
+++ b/src/ToolSession.zig
@@ -1,0 +1,70 @@
+// Copyright (C) 2023-2025  Lightpanda (Selecy SAS)
+//
+// Francis Bouvier <francis@lightpanda.io>
+// Pierre Tachoire <pierre@lightpanda.io>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+//! An isolated browsing context driven through `lp.tools`: its own Browser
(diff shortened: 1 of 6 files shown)


src/ToolSession.zig
```diff
@@ -0,0 +1,116 @@
...
+    try loadBlank(&a);
+    a.exitIsolate();
+
+    var b: ToolSession = undefined;
+    try b.init(testing.test_app);
+    try loadBlank(&b);
+    b.exitIsolate();
+
+    a.enterIsolate();
+    try std.testing.expectEqualStrings("a", (try tools.evalScript(arena, a.session, &a.registry, "globalThis.tag = 'a'")).text);
```
karlseguin: Unless you don't like it, `testing.expectEqual` (from `testing.zig`) handles strings "correctly". So you can use that and then drop the std import since you only use it for this.

karlseguin (approved):
needs rebase

src/ToolSession.zig
arrufat: Right!

PR #3315, opened 2026-08-27
css: reuse storage for updated property values

staylor:
## Summary

Store mutable CSS property values in the page factory's reusable slab allocator instead of the frame-lifetime arena.

Long values are explicitly duplicated rather than interned, so replacing or removing a property can safely return the owned buffer to the slab. A regression test verifies that storage from an earlier value is reused by a later update.

## Production context

This was found while investigating unexpected browser-process exits in a production rendering workload. Replaying mutation-heavy storefront behavior locally showed native RSS increasing for the page lifetime. `malloc_history -callTree` attributed the dominant allocation growth to repeated `CSSStyleDeclaration.setPropertyImpl` calls from inline transform updates.

This PR fixes that independently useful retention problem. It does not claim to explain every production exit; the timer-rearmed MutationObserver loop uncovered by the same investigation is handled separately in #3316.

## Why

`CSSStyleDeclaration.setPropertyImpl` allocated every updated value from `frame.arena`. That arena is monotonic for the frame lifetime, so repeated writes to an existing property retained every previous value even though only the latest value was reachable.

Mutable values now use owned buffers from the page factory's reusable slab allocator. This also avoids calling `String.deinit` on potentially interned long strings.

## Test

- `make test` — 1312/1312
- Regression test verifies property-value storage is reused after replacement

### DIFF
--- a/src/browser/Factory.zig
+++ b/src/browser/Factory.zig
@@ -56,6 +56,10 @@ pub fn init(arena: Allocator) Factory {
     };
 }
 
+pub fn storageAllocator(self: *Factory) Allocator {
+    return self._slab.allocator();
+}
+
 // this is a root object
 pub fn eventTarget(self: *Factory, child: anytype) !*@TypeOf(child) {
     return self.eventTargetWithAllocator(self._slab.allocator(), child);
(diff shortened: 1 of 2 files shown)


karlseguin:
Thanks. I think we'll be able to do this in at least a couple other places. I'll take a look.

PR #3317, opened 2026-08-27
agent: draw a REPL /screenshot in terminals that speak kitty graphics

arrufat:
Stacked on #3301.

A `/screenshot` without a path had nowhere to go in the REPL: the image went to the model when one was attached and was refused otherwise. This probes the terminal once and, when it speaks the [kitty graphics protocol](https://sw.kovidgoyal.net/kitty/graphics-protocol/) (kitty, ghostty, WezTerm, Konsole), draws the PNG below the outcome line.

**How**

- `src/agent/kitty.zig` — the probe is a graphics query followed by Device Attributes; every terminal answers DA, which bounds the read so nothing leaks into isocline's input, and multiplexers that don't pass graphics through (tmux) answer DA themselves and read as unsupported. The PNG the tool already produces is sent as-is (`f=100`), chunked at 4096, with `c`/`r` cell counts so the terminal scales it into the window's width and a screenful of rows.
- The tool is only asked for an inline image when the terminal or a model can take it, so an unsupported terminal keeps the tool's own `pass \`path\`` refusal; when only the model took it, the outcome line says so. The PNG is rendered once and shared between the terminal and the conversation.
- `src/agent/tty.zig` — the raw-mode and `TIOCGWINSZ` helpers the picker and spinner each carried, so the probe doesn't grow a third copy.

**Not covered**: sixel-only terminals (foot, xterm, mlterm) — a follow-up.

**Tested** with a pty harness that plays the terminal (answers the probe like kitty or like a plain terminal, sets the window size in cells and pixels): kitty → one probe, `● PNG, 1280x1080`, transmit `c=100 r=43` in 4 chunks with the PNG signature intact, prompt returns clean; plain → red `● pass \`path\`: this client can't display images`. Unit tests cover reply parsing, cell fitting, and chunking.

### DIFF
--- a/src/agent/Agent.zig
+++ b/src/agent/Agent.zig
@@ -696,11 +696,12 @@ fn runRepl(self: *Agent) void {
                 self.terminal.beginTool(tc.name(), slash_split.?.rest);
                 const result = self.runCommand(aa, cmd);
                 self.terminal.endTool();
-                self.printCommandResult(cmd, result);
+                const image = renderImage(aa, result);
+                self.printCommandResult(cmd, result, image);
                 if (!result.is_error) {
                     self.recordSaveCommand(navigationGoto(aa, tc.tool, tc.args) orelse cmd);
                 }
-                self.recordSlashToolCall(command_text, tc.name(), tc.args, result) catch |err| {
+                self.recordSlashToolCall(command_text, tc.name(), tc.args, result, image) catch |err| {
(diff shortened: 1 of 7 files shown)


karlseguin:
I potentially just have a bad setup (scaling, large font, ...) but is this what it's supposed to look like?
screenshot

arrufat:
No, that's not how I see it, but anyway, this was just a fun experiment, never intended to merge it. This one is better: https://github.com/lightpanda-io/browser/pull/3331

PR #3320, opened 2026-08-28
`application/octet-stream` in Mime for Blob

mookums:
This adds the `application/octet-stream` to the Mime.zig, allowing us to get rid of the TODO in `xhr.html` for Blob response type.


karlseguin:
I was going to dupe the raw type in the XHR object (using its arena). Even for "known" types, it's currently lossy, "text/html; charset=utf-8" -> "text/html"

PR #3323, opened 2026-08-28
cdp: fill all DOM.getBoxModel quads

arrufat:
`DOM.getBoxModel` returned zeroed `padding`/`border`/`margin` quads. Playwright reads `border` for `boundingBox()`, so every element came back as `{0,0,0,0}`. Puppeteer is unaffected: it evaluates `getBoundingClientRect` in page.

Lightpanda's layout has no padding, border or margin, so all four boxes are the content box.

### DIFF
--- a/src/cdp/domains/dom.zig
+++ b/src/cdp/domains/dom.zig
@@ -528,15 +528,16 @@ fn getBoxModel(cmd: *CDP.Command) !void {
     // TODO implement for document or text
     const element = node.dom.is(DOMNode.Element) orelse return error.NodeIsNotAnElement;
 
+    // Without padding, border or margin in the layout, the four boxes coincide.
+    // Drivers read `border` for boundingBox(), so it must not be zero.
     const rect = element.boundingClientRectValues(frame);
     const quad = rectToQuad(rect);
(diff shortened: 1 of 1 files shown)


karlseguin:
This...doesn't seem right? ELI5

arrufat:
Chrome's getBoxModel returns four rectangles: content, padding, border and margin. Each one wraps the previous one. For an element with no padding, border or margin, all four are the same rectangle.

Our layout never computes padding, border or margin, so that is always our case. We were filling content and leaving the other three as all zeros. Zeros do not mean "no border", they mean "a 0x0 box at the top left corner".

Playwright's boundingBox() reads the border quad, so every element looked like it was 0x0 at (0,0). That broke clicks, scrollIntoView and screenshot clips. This change fills all four quads with the one rectangle we have.

Puppeteer is not affected, it uses getBoundingClientRect in page. I will fix the description, it wrongly says both are affected.


Is my understanding of quads correct? If there's something else that looks off, let me know.

karlseguin:
Just didn't think it worked like that...don't understand why. But I tested it in chrome and you're right.

PR #3327, opened 2026-08-28
browser: open target=_blank links as popups and let tools follow them

arrufat:
Clicking a `target="_blank"` link was logged as `not_implemented` and dropped. It now opens a popup `Frame` via `openPopup`, the same top-level context `window.open` creates (#2237 listed this as the missing piece). The opener is withheld unless `rel=opener`, per spec.

Popups aren't visible to the tool layer (agent/MCP act on the root frame), so `finalizeAction` snapshots the popup count before the action and, when one appears, waits for it to load and makes it the session's current frame (`Session._followed_popup`). The result says `Opened a new window; tools now act on it.` Once the popup is gone (closed, or the root navigates), tools fall back to the root. The PandaScript `Runtime` routes through explicit page handles (`_tool_frame_override`), which takes precedence, so scripts are unaffected.

Not covered: popups still don't surface as CDP targets, and `<a download>` is still dropped.

Tests: `frames/target.html` (`_blank` leaves the opener in place) and an MCP test (click → follow → `getUrl`). Full suite: 1328/1328.

### DIFF
--- a/src/browser/frame/user_input.zig
+++ b/src/browser/frame/user_input.zig
@@ -553,7 +553,15 @@ fn followLink(frame: *Frame, target: *Node, element: *Element, href: []const u8,
             break :blk target.ownerFrame(frame);
         }
         break :blk frame.resolveTargetFrame(target_name) orelse {
-            log.warn(.not_implemented, "target", .{ .type = frame._type, .url = frame.url, .target = target_name });
+            // _blank: a new top-level context, as window.open creates. The
+            // opener is withheld unless rel=opener.
+            const owner = target.ownerFrame(frame);
+            try element.focus(frame);
+            _ = try owner.openPopup(.{
+                .url = href,
+                .name = "",
+                .opener = if (hasRelToken(element, "opener")) owner.window else null,
+            });
             return;
         };
     };
@@ -563,6 +571,17 @@ fn followLink(frame: *Frame, target: *Node, element: *Element, href: []const u8,
         .reason = .script,
         .kind = .{ .push = null },
     }, .{ .anchor = target_frame });
+}
+
+fn hasRelToken(element: *Element, token: []const u8) bool {
+    const rel = element.getAttributeSafe(comptime .wrap("rel")) orelse return false;
+    var it = std.mem.tokenizeAny(u8, rel, &std.ascii.whitespace);
+    while (it.next()) |t| {
+        if (std.ascii.eqlIgnoreCase(t, token)) {
+            return true;
+        }
+    }
+    return false;
 }
 
 pub fn triggerKeyboard(frame: *Frame, keyboard_event: *KeyboardEvent) !void {
--- a/src/browser/frame/user_input.zig
+++ b/src/browser/frame/user_input.zig
@@ -552,17 +552,13 @@ fn followLink(frame: *Frame, target: *Node, element: *Element, href: []const u8,
         if (target_name.len == 0) {
             break :blk target.ownerFrame(frame);
         }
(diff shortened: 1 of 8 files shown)


src/browser/frame/user_input.zig
```diff
@@ -553,7 +553,15 @@ fn followLink(frame: *Frame, target: *Node, element: *Element, href: []const u8,
             break :blk target.ownerFrame(frame);
         }
         break :blk frame.resolveTargetFrame(target_name) orelse {
```
karlseguin: I don't like how "load-bearing" the null is now. Easy to imagine some future change to `resolveTargetFrame` return `null` without knowing how it's being interpreted. I think we should change the signature to return a tagged union, and then each callsite gets to be more explicit.

As an example, you might want to apply this same change to a <form target="_blank"> to make it consistent - that code still orelse not implemented.

src/browser/frame/user_input.zig
```diff
@@ -553,7 +553,15 @@ fn followLink(frame: *Frame, target: *Node, element: *Element, href: []const u8,
...
-            log.warn(.not_implemented, "target", .{ .type = frame._type, .url = frame.url, .target = target_name });
+            // _blank: a new top-level context, as window.open creates. The
+            // opener is withheld unless rel=opener.
+            const owner = target.ownerFrame(frame);
+            try element.focus(frame);
+            _ = try owner.openPopup(.{
+                .url = href,
+                .name = "",
```
karlseguin:
```suggestion
                .name = target_name,
```

src/browser/tools.zig
```diff
@@ -1778,23 +1778,52 @@ fn formatActionResult(
...
+    return .{ .frame = frame, .popups = if (frame) |f| f._page.popups.items.len else 0 };
+}
+
 /// Finish a state-changing action: drain any queued navigation triggered by
 /// the action, then tag `body` with the resulting page URL and title so the
 /// caller (LLM, MCP client) can see whether the action triggered navigation.
-fn finalizeAction(arena: std.mem.Allocator, session: *lp.Session, registry: *CDPNode.Registry, body: []const u8) ToolError![]const u8 {
-    const before = session.currentFrame();
+fn finalizeAction(arena: std.mem.Allocator, session: *lp.Session, registry: *CDPNode.Registry, scope: ActionScope, body: []const u8) ToolError![]const u8 {
+    const before = scope.frame;
```
karlseguin, line 1800: `before.?._frame_id` can now represent a popup, and almost nothing is popup aware. So `awaitQueuedNavigation` would return an `error.FrameNotFound`

src/browser/frame/user_input.zig
arrufat: Done in 95e8dfa40: it returns a tagged union (`frame` / `blank`) and each callsite switches on it. `<form target=_blank>` now takes the popup path too; the popup is opened only once the submission goes ahead, so validation or `preventDefault` can't leave a stray window.

src/browser/frame/user_input.zig
arrufat: Folded into `Frame.openBlankTarget`, shared with the form path. `.blank` only comes from `_blank`, which `openPopup` already maps to an empty name.

src/browser/tools.zig
arrufat: Confirmed, any navigation from inside a followed popup returned `NavigationFailed`. `awaitQueuedNavigation` now takes the `Frame` and waits on its Page root, read before processing since a synthetic root navigation frees the Page in place. Covered by the MCP test: `evaluate(location.href = ...)` inside the popup, then `getUrl`.

PR #3328, opened 2026-08-28
Add failing Polymer components tests

nikneym:
These tests are currently failing. Elements inside `<template>` tags must not have browsing context since they should be stored in a distinct `DocumentFragment`; since they're inactive, they should not be upgraded by scripts, load events, styles etc.


nikneym:
Cherry-picked to #3587, closing.

PR #3329, opened 2026-08-28
Better server

karlseguin:
Significant rework of the CDP/BiDi server. There are two main changes:

1 - poll replaced with EPoll/KQueue 
2 - make http serving a first class citizen

The change from poll -> epoll/kqueue isn't performance driven, it's just about
tighter code. Both epoll and kqueue let you associate arbitrary data with a
socket, so we don't need to keep arrays in sync in order to associate a socket
with a CDP by index. They both provide some event/notification mechanism, which
is cleaner than the pipe required by poll.

The poll -> epoll/kqueue change could almost have been mechanical. Making HTTP
a first class citizen is the more significant of the two changes

In `main`, a new connection always spawns a thread and, until does its own
little read loop until the connection is upgraded. This is not efficient, it
uses up a connection slot, and it's inconsistent with the final WebSocket
connection which _is_ polled off the main loop. Using up a slot means that
keepalive isn't possible, else HTTP connections would quickly use up all
available slots/threads.

This commit parses and serves HTTP requests on the main thread (safe
because none of the processing is blocking). The approach is better streamlined
for HTTP requests which never upgrade (/metrics, WebDriver) without causing
any performance overhead for those that do. It simplifies some things (e.g. an
"http" socket or a "websocket" socket is monitored and read in a similar manner
(on the main loop)). It makes other things more complicated; the flow is no
longer accept -> spawn -> upgrade -> websocket loop. It's loop -> accept -> loop
-> process -> (http | ws).

This is built ontop of the BiDi branch because (a) WebDriver is what needs
better  HTTP support and (b) some of the more mechanical changes already exist
in that branch (e.g. src/cdp/, src/server.zig -> src/server/*)


src/server/Link.zig
```diff
@@ -0,0 +1,388 @@
...
+    var pushed = false;
+    var remaining = budget;
+    while (remaining > 0) {
+        const dst = self.reader.readBuf();
+        if (dst.len == 0) {
+            // a partial message already fills the buffer
+            return error.TooLarge;
+        }
+        const want = dst[0..@min(dst.len, remaining)];
+        const n = posix.read(self.socket, want) catch |err| switch (err) {
```
krichprollsch, line 188: Claude thinks we can have troubles due to the blocking mode switch of the socket in `send`.
It suggests to use `posix.recv(self.socket, want, posix.MSG.DONTWAIT)`

src/server/http.zig
```diff
@@ -0,0 +1,927 @@
...
+
+// Best effort, connection is being closed. A partial write ends up as a partial
+// write: no pending, no retry.
+fn writeError(conn: *Connection, err: anyerror) void {
+    const response: []const u8 = switch (err) {
+        error.ConnectionClosed, error.ConnectionResetByPeer, error.BrokenPipe => return,
+        error.InvalidHeader, error.InvalidHTTPMethod, error.BodyNotSupported => invalid_request_response,
+        error.RequestTooLarge => request_too_large_response,
+        else => {
+            log.warn(.serve, "serve error", .{ .err = err });
```
krichprollsch, line 546: Shouldn't we return a `500` in this case?

krichprollsch:
I get a crash on my test:
```
Lightpanda has crashed. Please report the issue:
https://github.com/lightpanda-io/browser/issues
or let us know on discord: https://discord.gg/g24PtgD6

reason: attempt to use null value
OS: linux
mode: Debug
version: 1.0.0-dev.8994+ebad491fd
source: global
panicked during a panic. Aborting.
run
└─ run exe lightpanda failure
error: process terminated with signal ABRT
failed command: ./.zig-cache/o/62e8c1a5e8e7236a87af96615ce57131/lightpanda

Build Summary: 31/33 steps succeeded (1 failed)
run transitive failure
└─ run exe lightpanda failure

.zig-cache/o/552d4e306d7293bacfec2178eeaee1bc/build /usr/local/zig-0.16.0/zig /usr/local/zig-0.16.0/lib /home/pierre/wrk/browser-review .zig-cache /home/pierre/.cache/zig --seed 0x69735f8c -Z672f52b44756592d run
```


here is the script https://gist.github.com/krichprollsch/c9fed136c2bdb467a01563df9452526c
I run it with

```
LP_WS=ws://127.0.0.1:9222  CONCURRENCY=1 ITERATIONS=100 node lightpanda-repro.mjs
```

karlseguin:
> I get a crash on my test:
> 
> ```
> version: 1.0.0-dev.8994+ebad491fd
> ```

This isn't the latest version of the branch. The bug was in main and fixed in https://github.com/lightpanda-io/browser/pull/3337 and was rebased into the branch at some point.

krichprollsch (approved):

PR #3332, opened 2026-08-28
http: adjust rate limit depending the time to first byte

krichprollsch:
Compute the time to first byte received to adjust host's rate limit. The more the server takes time to send the first byte, the more we rate limit it.

Note: only requests send after a response is received are impacted. Requests already queued stay untouched.

follow-up for #3266

### DIFF
--- a/src/help.zon
+++ b/src/help.zon
@@ -409,7 +409,9 @@
     \\          Defaults to 1.
     \\  --http-nav-delay <INT>
     \\          Minimum time in ms between two top-level navigations to the same
-    \\          host (see --http-nav-burst). Disable by setting to 0.
+    \\          host (see --http-nav-burst). The spacing grows with the number
+    \\          of navigations sent to the host, and comes back down when the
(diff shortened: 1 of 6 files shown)


karlseguin:
There's a fundamental problem in the original RateLimiter which I now see:  we don't check on redirect. (CORS suffers because of this too). Here the observe() is thrown away because the host was never reserved.

Also, a site that quickly errors (429, 500) _reduces_ the delay. A site that timesout or fails to connect, has no impact. More generically, I'd say that a site's speed isn't a great measure of load (certainly not compared to status code / errors). A site that is slow (but not necessarily more burdened) gets a greater delay than a site that's fast. 

I, because of my location, get a greater delay to a European site than you do (essentially a doubling of my network latency)

I setup a puppeteer script to hit `https://httpbin.org/delay/1` 3 times. With `--http-nav-delay 100` on `main`, it take 4.5 seconds. On `rate-limit-ttfb` it takes 8 seconds.

PR #3333, opened 2026-08-28
markdown: separate flex and grid items in the dump

arrufat:
On openmymind.net every post is `<a class="row">Title<b>Aug 04 2026</b></a>` with `main > a { display: flex }` and `b { margin-left: auto }`, so the dump read `[Title**Aug 04 2026**](…)`. Children of a flex/grid container are separate boxes however inline their tags are, so the dumper should not run them together.

- `StyleManager` keeps the display kind (`none | flex | grid | other`) in the same single-pass cascade that resolved `display: none`, exposed as `display(el, access)`; `hasDisplayNone` is now `display() == .none`. The cascade returns a small `Resolved` instead of a bool so `isElementHidden` and `display` share it.
- `markdown.zig` gets visibility and container kind from that one call per element; `renderChildren(parent, boxed)` puts a space between rendered items of a flex/grid parent (not before blocks, which break the line anyway), skips the whitespace between items like layout does, and trims text items.

### DIFF
--- a/src/browser/markdown.zig
+++ b/src/browser/markdown.zig
@@ -171,13 +171,29 @@ const Context = struct {
         }
     }
 
+    fn getRenderDisplay(self: *Context, el: *Element, force_slot: bool) ?StyleManager.Display {
+        const display = visibleDisplay(el, self.frame) orelse {
+            if (el.asNode() == self.root) {
+                return StyleManager.Display.other;
+            } else {
+                return null;
+            }
+        };
+        if (dump_html.shouldStripElement(el, self.strip, self.frame)) return null;
+        if (!force_slot and el.getAttributeSafe(comptime .wrap("slot")) != null) return null;
+        return display;
+    }
+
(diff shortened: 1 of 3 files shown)


src/browser/markdown.zig
```diff
@@ -187,19 +194,44 @@ const Context = struct {
...
+            if (!boxed) {
+                try self.render(child);
+                continue;
+            }
+            if (child.is(Element)) |el| {
+                if (self.skips(el, self.force_slot)) continue;
+                if (separate and !el.getTag().isBlock() and !self.state.last_char_was_newline) {
+                    try self.writer.writeByte(' ');
+                }
+                try self.renderElement(el);
```
karlseguin: renderElement re-calculates things that skips (4 lines up) already did, e.g. `renderElement` calls `shouldStripElement` which skips already asserted was false. I realize renderElement can be called from a different place, but these checks aren't free, so doing them 2x per element isn't ideal.

arrufat: Fixed, thank you!

PR #3334, opened 2026-08-28
apply a stronger rate limit on cgit websites

krichprollsch:
Detect `<meta name="generator" content="cgit...` to enforce a strongest specific rate limit.

Following #3266 and #3332


karlseguin:
Is this meant as a temporary solution? Adjusting the limit based on feedback from the host (something like https://github.com/lightpanda-io/browser/pull/3332) seems like a better approach.  Hard-coding generators seems like a path we should avoid. If WordPress operators ask us to special-case their generator, are we going to do it?

PR #3335, opened 2026-08-28
Apply `SameSite=Lax` exception only on safe methods

nikneym:
On top of #3293.

### DIFF
--- a/src/browser/webapi/storage/Cookie.zig
+++ b/src/browser/webapi/storage/Cookie.zig
@@ -43,12 +43,34 @@ expires: ?f64,
 secure: bool = false,
 http_only: bool = false,
 same_site: SameSite = .none,
+// True when Set-Cookie carried no SameSite attribute: the cookie is Lax
+// through the "Default" enforcement mode (RFC 6265bis 5.6.7.1), which
+// makes it eligible for "Lax-allowing-unsafe", see appliesTo.
+same_site_default: bool = false,
+// Seconds since the epoch. Set by Jar.add, and inherited from the cookie it
+// replaces, so a site re-setting a cookie on every response can't keep it
+// "recently created" forever.
+creation_time: u64 = 0,
 
 pub const SameSite = enum {
     strict,
     lax,
     none,
 };
+
+// How the request carrying the Cookie header reaches its target. Only a
+// cross-site request cares; it's what unlocks SameSite=Lax cookies.
+pub const RequestKind = enum(u2) {
+    // A sub-resource (image, script, fetch...), a sub-frame's document, or
+    // a non-HTTP retrieval (document.cookie).
+    subresource,
+    // A top-level navigation using a "safe" method (GET, HEAD, OPTIONS).
+    navigation,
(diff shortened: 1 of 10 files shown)


src/browser/webapi/storage/Cookie.zig
```diff
@@ -581,7 +581,9 @@ pub const Jar = struct {
     pub const LookupOpts = struct {
         is_http: bool,
         request_time: ?u64 = null,
-        is_navigation: bool = true,
+        // True only for a top-level navigation using a safe HTTP method.
```
karlseguin, line 584: I don't think this comment is correct, what about iframes?

karlseguin (commented):
Any thoughts on https://re-sync.nl/en/blog/samesite-lax-cookie-bypass/ ?

(With the official documentation about it: https://www.chromium.org/updates/same-site/faq/#q-what-is-the-lax-post-mitigation)

nikneym:
> Any thoughts on [re-sync.nl/en/blog/samesite-lax-cookie-bypass](https://re-sync.nl/en/blog/samesite-lax-cookie-bypass/) ?

Interesting find & article... Turns out there's also a section that touch this in RFC6265bis:

5.6.7.2.  "Lax-Allowing-Unsafe" enforcement

   As discussed in Section 8.8.6, compatibility concerns may necessitate
   the use of a "Lax-allowing-unsafe" enforcement mode that allows
   cookies to be sent with a cross-site HTTP request if and only if it
   is a top-level request, regardless of request method.  That is, the
   "Lax-allowing-unsafe" enforcement mode waives the requirement for the
   HTTP request's method to be "safe" in the SameSite enforcement step
   of the retrieval algorithm in Section 5.8.3.  (All cookies,
   regardless of SameSite enforcement mode, may be set for top-level
   navigations, regardless of HTTP request method, as specified in
   Section 5.7.)


   "Lax-allowing-unsafe" is not a distinct value of the SameSite
   attribute.  Rather, user agents MAY apply "Lax-allowing-unsafe"
   enforcement only to cookies that did not explicitly specify a
   SameSite attribute (i.e., those whose same-site-flag was set to
   "Default" by default).  To limit the scope of this compatibility
   mode, user agents which apply "Lax-allowing-unsafe" enforcement
   SHOULD restrict the enforcement to cookies which were created
   recently.  Deployment experience has shown a cookie age of 2 minutes
   or less to be a reasonable limit.

nikneym:
This currently conflicts w/ https://github.com/lightpanda-io/browser/pull/3448; since both touch `SameSite=Lax` movement, I'm working on a change that cover both.

nikneym:
This now also includes a solution for what #3448 trying to fix.

karlseguin:
LGTM, but I think this is worth tests in ../demo/ (or as an html test in this project if that works). There's 1 unit test for the Cookie.Jar, but the feature requires all the layers to do their part, which isn't tested anywhere.

nikneym:
Agreed, will add this through `demo`.

PR #3343, opened 2026-08-29
js: reject module evaluation before instantiation

staylor:
## Why

Browser processes have been observed terminating in V8 with this fatal API check while loading ES modules:

```text
Check failed: module_status == kLinked ||
              module_status == kEvaluatingAsync ||
              module_status == kEvaluated
```

This is a process-level abort, not a recoverable JavaScript exception. In a server handling multiple CDP connections, one raced module evaluation therefore disconnects every unrelated context on that process.

The earlier termination guard in #3260 prevents late async module completions from entering V8 during teardown. The remaining signature can occur without a pending termination, so the lower-level `Module.evaluate` wrapper also needs to enforce V8's documented state precondition.

## Root cause

The module cache publishes a compiled module before dependency instantiation has completed. During re-entrant/asynchronous module loading, another path can observe that entry while V8 still reports it as uninstantiated or instantiating. Calling `v8::Module::Evaluate` in either state triggers the fatal check above.

## Fix

Reject evaluation with `InvalidModuleStatus` before entering V8 when the module is uninstantiated or currently instantiating. Existing script and dynamic-import paths already convert evaluation errors into a failed script or rejected import, keeping the browser process and other contexts alive.

The regression test covers both a newly compiled module and re-entrant evaluation from inside its dependency resolver. Without the guard, the test aborts in `Module::Evaluate` with `Expected instantiated module`; with it, both calls return the typed error and subsequent JavaScript still runs.

## Tests

- `make test` — 1341/1341

### DIFF
--- a/src/browser/js/Context.zig
+++ b/src/browser/js/Context.zig
@@ -406,7 +406,10 @@ pub fn module(self: *Context, comptime want_result: bool, local: *const js.Local
 }
 
 fn evaluateModule(self: *Context, comptime want_result: bool, mod: js.Module, url: []const u8, cacheable: bool) !(if (want_result) ModuleEntry else void) {
-    const evaluated = mod.evaluate() catch {
+    const evaluated = mod.evaluate() catch |err| {
+        if (err == error.InvalidModuleStatus) {
+            return err;
+        }
         if (comptime lp.IS_DEBUG) {
             std.debug.assert(mod.getStatus() == .kErrored);
         }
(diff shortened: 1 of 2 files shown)


karlseguin:
Do you have a specific site / case that reproduces this? As far as I can tell, the unit test forces a state that shouldn't be reachable through normal execution.

I don't see any harm in the change, so we'll merge it and hope it helps. In debug, it continues to panic. In release, it'll log a error-level message with a bit of details that might help pin the issue down.

PR #3345, opened 2026-08-30
telemetry: track if --obey-robots is used

karlseguin:
Re-purposes the "proxy" integer as a CLI bitmask flag. Requires a server update.


krichprollsch (approved):
Server is ready

PR #3346, opened 2026-08-30
CI: nightly orderfile generation

karlseguin:


karlseguin:
@krichprollsch No clue if this will work. It runs 2 hours before the nightly. I needs to push 2 files to main, so:

We need to install the DISTRIBUTION_APP on lightpanda-io/browser with `Contents: Read and write` (this is currently installed on browser-docker and homebrew-browser, so whatever we did there, we'd need here). AND on main's protection rule, the app needs to be added to the bypass.

If you prefer we could have it open a PR every week or so.

krichprollsch:
I think I would prefer the CI creates a new PR, I dislike allowing it to commit directly in main.

PR #3352, opened 2026-08-30
cdp: serve /json target list and create the target for flat-protocol clients

arrufat:
## What

- `/json` and `/json/list` return a one-element target list instead of `[]`; new `/json/new` (GET+PUT), `/json/activate/*` and `/json/close/*` endpoints. The listed target is synthetic (built at server init, before any browser context exists); `/json/new?url` and `/json/close` are accepted no-ops that log `not_implemented`.
- Flat-protocol clients (chrome-remote-interface, pychrome) list `/json`, connect to the `webSocketDebuggerUrl`, and send `Page.*`/`Runtime.*` directly with no `Target` handshake. Previously every such command failed with `BrowserContextNotLoaded`. Page-scoped handlers now obtain their context through `Command.requireBrowserContext`, which for a sessionless command creates the browser context, target and attached session — reusing `Target.createTarget`'s setup, extracted into `setupTarget`. Attaching is required because page events (including the `Page.navigate` response) are only emitted to an attached session.
- Teardown methods (`X.disable`, `Page.close`) on a connection with no context succeed as no-ops via `Command.teardownBrowserContext` instead of creating a page. Handlers that resolve ids minted by an earlier command keep returning `BrowserContextNotLoaded`, since a fresh context can never satisfy them.
- Routing in `Handshake.handleHttpRequest` ignores the query string and one trailing slash (as Chrome does) and doesn't wait on a PUT body. HTTP responses share `Handshake.buildResponse`.

## Why

Target-listing drivers could not connect at all. Session-based drivers (Puppeteer, Playwright, chromedp) are unaffected: their commands either carry a session id or go through the `Target`/`Browser` domains, neither of which creates a page implicitly. Handlers that never needed a context (`Page.enable`, `Page.getFrameTree`) are unchanged.

Three tests that asserted `BrowserContextNotLoaded` for sessionless page commands were updated to the new behavior.


arrufat:
Converting this to a draft because it is really a "do we want this surface?" question rather than a bug fix, and I would rather have your call on it. @karlseguin @krichprollsch what do you think?

**Demand:** honestly, none on record. No issue mentions pychrome or chrome-remote-interface, and nobody has reported the `BrowserContextNotLoaded` failure a flat client hits today. The beneficiaries would be low-level CDP clients that list `/json` and drive a page directly without the `Target` handshake — chrome-remote-interface, pychrome, `curl`+`websocat` scripting. Real, but niche next to Puppeteer/Playwright.

**Cost:** the `/json` listing is trivial. The implicit target creation is not free: a new code path in `dispatchCommand` that creates browser state on its own, an `isPageDomain` list to maintain, sessionless `Page.enable` now creating the target instead of returning the startup stub, and three tests rewritten.

**Both or neither:** the two halves are coupled. Shipping only the `/json` list would be worse than today — clients would go from failing fast ("no targets") to connecting and then failing with an opaque `BrowserContextNotLoaded`.

If no, I will close this and leave `/json` returning `[]` as the honest fail-fast; the branch stays around for when a real user shows up. If yes, it is ready for review as is.

karlseguin:
At the very least, I'd like to merge https://github.com/lightpanda-io/browser/pull/3209 and https://github.com/lightpanda-io/browser/pull/3329 first. They've been a bit of a pain to keep rebased.

src/server/cdp/CDP.zig
```diff
@@ -1376,6 +1376,27 @@ pub const Command = struct {
...
+    pub fn requireBrowserContext(self: *Command) !*BrowserContext {
+        if (self.browser_context) |bc| return bc;
+        if (self.input.session_id != null) return error.BrowserContextNotLoaded;
+        try @import("domains/target.zig").createImplicitTarget(self);
+        return self.browser_context.?;
+    }
+
+    // Teardown (X.disable, Page.close): with no context there is nothing to
+    // act on — sends the ok result and returns null.
+    pub fn teardownBrowserContext(self: *Command) !?*BrowserContext {
```
karlseguin: The function name isn't great. `getBrowserContextForDisable` ??

src/server/http.zig
```diff
@@ -639,12 +655,31 @@ fn upgradeCDP(server: *Server, conn: *Connection, req: *Connection.Request) !Ser
...
+    return serveHTTPResponse(server, conn, req, .{ .static = server.http_responses.new });
+}
+
+// GET /json/activate/{id} (cdp)
+fn serveJSONActivate(server: *Server, conn: *Connection, req: *Connection.Request) !Served {
+    return serveHTTPResponse(server, conn, req, .{ .static = target_activated_response });
+}
+
+// GET /json/close/{id} (cdp)
+fn serveJSONClose(server: *Server, conn: *Connection, req: *Connection.Request) !Served {
```
karlseguin: This fires on `/json/close`  and with a successful "Target is closing" response. Chrome 404s.

karlseguin:
I'm uneasy about any CDP changes.   I like the `try cmd.requireBrowserContext();` as a convenience but unsure about the implicit creation. Claude thinks this change will conflict with our single-browser context logic. It recommends having `createBrowserContext` and `createTarget` adopt and untouched (how do you tell if it's been touched?) implicit browser context.

But, at most of the new APIs are wrong / noop. Does that really improve anything? We advertize id: 1, but you can't attach to it, the url and title are always "about:blank" and ""   /json/new, /json/close act like they do something, but they don't. I don't really get it.

arrufat:
Closing: no demand on record, and the /json surface is mostly a fiction (unattachable id, no-op new/close) held up by implicit target creation in dispatch. If a flat-protocol client ever shows up, we should do it properly with /json reporting real per-driver targets. Keeping the branch for reference.

Issue #3359, opened 2026-09-01
performance.getEntriesByType('resource') returns empty, breaking CDP networkidle detection


angri450:
## Summary

`performance.getEntriesByType('resource')` returns an empty/zero list, breaking CDP-based automation that uses resource-count polling to detect network idle.

## Context

We're building a CDP client (C#) that renders pages through Lightpanda's CDP server (`lightpanda serve`). Our renderer waits for network idle by polling `performance.getEntriesByType('resource').length` until the count stabilizes across two consecutive polls (a common pattern, also used by Playwright's `waitUntil: 'networkidle'` heuristic).

## Observed Behavior

With Lightpanda CDP (browser-level WS, `Target.createTarget` + `attachToTarget`):

1. `Page.navigate` fires `Page.loadEventFired` normally
2. `Runtime.evaluate` of `performance.getEntriesByType('resource').length` returns `0` immediately — even after the page has loaded many resources (verified on https://www.mdpi.com/1420-3049/20/12/19754, a heavy SPA with 600KB+ DOM)
3. Because the resource count never changes (always 0), the networkidle poll declares "idle" instantly and we read the DOM too early — getting only the skeleton (`<title>&nbsp;</title>`, empty body) instead of the fully rendered content

When we instead use a fixed delay (e.g. 10s) before reading the DOM, Lightpanda renders the same page completely (636KB DOM, correct title). So the page rendering itself works — the issue is specifically that the resource-performance API does not report entries.

## Expected

`performance.getEntriesByType('resource').length` should return the number of loaded resources (scripts, stylesheets, images, XHR/fetch), matching Chromium behavior. This is a spec'd API (Resource Timing API, https://w3c.github.io/resource-timing/).

## Reproduction

```
lightpanda serve --host 127.0.0.1 --port 9224
```

Connect CDP browser-level WS at `ws://127.0.0.1:9224/`:

1. `Target.createTarget` (url: about:blank)
2. `Target.attachToTarget` (flatten: true) → get sessionId
3. `Page.enable` (sessionId)
4. `Page.navigate` (url: https://www.mdpi.com/1420-3049/20/12/19754, sessionId)
5. Wait for `Page.loadEventFired`
6. `Runtime.evaluate` → `performance.getEntriesByType('resource').length` → returns 0
7. Wait 10s, evaluate again → still 0

[...]

PR #3367, opened 2026-09-01
fix: synthesize Resource Timing entries for loaded sub-resources (issue #3359)

angri450:
## Problem

`performance.getEntriesByType('resource')` returns an empty array, breaking CDP networkidle detection: puppeteer-style clients poll `resource.length` for stability and never see it move, so they wait forever.

## Root cause

Lightpanda tracks sub-resource loading only via `_pending_loads` (main document + iframes). Regular sub-resources (scripts, stylesheets, images) never increment it, and with `load_resources` defaults (image/iframe/stylesheet off) images are not fetched at all. External scripts additionally take a separate prescan-preload path that bypasses `pendingLoadCompleted` entirely.

## Fix

Record a synthetic resource entry (count only, no URL/duration) at every sub-resource load completion:

- `Frame.pendingLoadCompleted` else branch -> iframes, images (when `load_resources.image` enabled)
- `Frame.loadExternalStylesheet` -> external stylesheets
- `ScriptManager.addFromElement` sync fetch -> sync scripts
- `ScriptManager.PreloadedScript.doneCallback` -> prescan-preloaded scripts
- `ScriptManagerBase.Script.doneCallback` -> async scripts

`getEntriesByType('resource')` synthesizes `_resource_count` entries. No real URL/duration captured; the count is what networkidle polling needs.

## Verification

- 3-script page -> `resource.length = 3`
- Polling every 500ms -> count stable after load (networkidle semantics)
- cnki.net -> `resource.length = 24` (was 0), document complete

Tested with CDP: createTarget -> attachToTarget -> Page.navigate -> Runtime.evaluate.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,10 +1,5 @@
-/.zig-cache/
-/.lp-cache/
-/.lp-history
-/.lp-history.js
-/.lp-agent.zon
-/zig-pkg/
-zig-out
-lightpanda.id
-/src/rust/target/
-src/snapshot.bin
+.local-deps/
+.lp-cache/
+.zig-cache/
+zig-out/
+zig-pkg/
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -4,45 +4,17 @@
     .fingerprint = 0xda130f3af836cea0, // Changing this has security and trust implications.
     .minimum_zig_version = "0.16.0",
     .dependencies = .{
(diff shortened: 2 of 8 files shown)


karlseguin:
Thank you for the PR, went with this: https://github.com/lightpanda-io/browser/pull/3391

PR #3371, opened 2026-09-01
webapi: document.execCommand and its query/designMode companions

arrufat:
`document.execCommand`, `queryCommandSupported`/`Enabled`/`State`/`Indeterm`, `queryCommandValue` and `designMode` were absent, so the dominant copy-to-clipboard path and its feature probe both threw `TypeError`.

There's no editing pipeline or clipboard, so every command reports unsupported (`false` / `""`, `designMode` stays `"off"`), which keeps callers on their feature-detection fallbacks.

### DIFF
--- a/src/browser/tests/document/document.html
+++ b/src/browser/tests/document/document.html
@@ -443,3 +443,20 @@
     testing.expectEqual(1, acss.length);
   }
 </script>
+
+<script id=execCommand>
+  {
+    // No editing pipeline or clipboard: every command is unsupported, so
+    // callers stay on their feature-detection fallbacks instead of throwing.
+    testing.expectEqual(false, document.queryCommandSupported('copy'));
+    testing.expectEqual(false, document.queryCommandEnabled('copy'));
+    testing.expectEqual(false, document.queryCommandState('bold'));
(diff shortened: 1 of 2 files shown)


karlseguin:
Dummy implementations can do more harm than good. In some cases they unblock sites. In others, they make the error more obscure. I've been avoiding implementing any part of this API (as a dummy or for real) until we had good reason to. Is there a specific site / library this is  helping to unblock?

arrufat:
No specific site. It came from an API-coverage scan, not a repro. Checking the libraries I had in mind: copy-to-clipboard and clipboard.js both wrap execCommand in try/catch and fall back, so the absence already degrades cleanly, and editors that rely on it need real editing semantics a stub can't give. Agreed, it's better left out until there's a concrete case. Closing.

arrufat:
No specific site. This came from an API coverage scan, not a repro. copy-to-clipboard and clipboard.js both wrap execCommand in try/catch and fall back, so the absence already degrades cleanly, and editors that rely on it need real editing semantics a stub cannot give. Agreed it is better left out until there is a concrete case. Closing.

PR #3377, opened 2026-09-01
webapi: img.decode()

arrufat:
`img.decode()` was missing, so Next.js `<Image>` and React lazy-image paths threw `TypeError`. Nothing is decoded here, so it resolves with `undefined` right away. `naturalWidth`/`naturalHeight` stay 0 on purpose (`Image.zig` notes non-zero values would misrepresent the browser).


arrufat:
For context: this is a dummy, nothing is decoded. The concrete caller is Next.js Image, which calls img.decode() unguarded and only catches the rejection, so a missing method throws synchronously. If you would rather not add it without decoding, say so and I will close.

PR #3379, opened 2026-09-01
webapi: input valueAsNumber, valueAsDate, stepUp/stepDown, showPicker

arrufat:
`HTMLInputElement` lacked `valueAsNumber`, `valueAsDate` and `stepUp`/`stepDown`, so react-hook-form's `valueAsNumber` and date pickers read `undefined`/`NaN`.

- `valueAsNumber` get/set for number/range/date/month/week/time/datetime-local per HTML "value as number" (ms since epoch, months for `month`, ms since midnight for `time`), built on the existing date/time validators; other types give NaN / throw `InvalidStateError`.
- `valueAsDate` get/set for date/month/week/time (`null` elsewhere, `TypeError` on non-Date, `InvalidStateError` on unsupported types).
- `stepUp`/`stepDown` with the HTML step base and ladder snapping, clamped to min/max; `step="any"` throws.
- The same conversion now backs `rangeUnderflow`/`rangeOverflow` for the date types, which only number/range had.
- Review follow-up: `tooLong`/`tooShort` only for user-edited values (input and textarea), empty `pattern` is a pattern, three-digit time fractions. The test files pass in Firefox and Chrome.

### DIFF
--- a/src/browser/tests/element/html/input-validity.html
+++ b/src/browser/tests/element/html/input-validity.html
@@ -279,3 +279,24 @@
   testing.expectEqual(false, el.validity.patternMismatch);
 }
 </script>
+
+<input id="date-low" type="date" min="2024-01-01" value="2023-12-31">
+<input id="date-ok" type="date" min="2024-01-01" max="2024-12-31" value="2024-06-15">
+<input id="time-high" type="time" max="17:00" value="17:30">
+<input id="month-low" type="month" min="2024-02" value="2024-01">
+<input id="week-ok" type="week" min="2024-W01" value="2024-W10">
+
+<script id="date_time_ranges">
+  {
+    testing.expectEqual(true, $('#date-low').validity.rangeUnderflow);
+    testing.expectEqual(false, $('#date-low').validity.rangeOverflow);
+    testing.expectEqual(false, $('#date-ok').validity.rangeUnderflow);
+    testing.expectEqual(false, $('#date-ok').validity.rangeOverflow);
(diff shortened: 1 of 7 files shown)


karlseguin:
`input-value-as.html` has a few errors when opened in Firefox. The existing `input-validity.html` also has existing errors plus an error in the new case `date_time_ranges`.

We have plenty of tests which deviate from browsers, but I feel like neither of these should. If you can figure out / fix what's wrong with `input-value-as.html` and the new  `date_time_ranges` case that would be great (if you could also fix whatever is wrong with the existing  `input-validity.html` , that would be even better).

arrufat:
Ran all three files in Firefox and Chrome and fixed everything that differed, on our side where we were wrong:

- stepUp/stepDown now use the step base (min, else the value attribute) and snap off ladder values to the next rung the way browsers do; clamping lands on the last rung inside min/max.
- time strings keep three fraction digits.
- an empty pattern attribute is a pattern (matches only the empty string), which is what the pre-existing pattern case was hitting.
- tooLong/tooShort only for values last changed by a user edit (input and textarea), with a CDP Input.insertText test for the user path. That was the pre-existing input-validity failure; textarea-validity had the same one.
- showPicker dropped: browsers throw NotAllowedError without a gesture.
- validationMessage wording is browser specific, so those assertions check presence only; month/week assertions are skipped where the browser falls back to text (Firefox).

All three pass in Firefox 154, Chrome 152 and our runner.

PR #3381, opened 2026-09-01
webapi: canvas measureText and the drawing methods chart libraries probe

arrufat:
`CanvasRenderingContext2D` (and the offscreen one) lacked `measureText`, and `roundRect`, `ellipse`, `setLineDash`/`getLineDash`, `createLinearGradient`/`RadialGradient`/`ConicGradient`/`createPattern`, `isPointInPath`/`isPointInStroke` and the shadow/filter/spacing properties, so Chart.js, D3 and ApexCharts threw at init.

- `measureText` returns a `TextMetrics` from the deterministic model already used for SVG text (`text_measure.zig`): width from the per-glyph advances, ascent/descent 0.8em/0.2em. `font` is now Zig-backed so its size feeds the measurement (invalid strings ignored, per spec).
- `CanvasGradient` (validates stop offsets) and `CanvasPattern` (validates repetition) exist so they can be assigned; nothing is painted.
- `setLineDash` keeps spec semantics (negatives ignored, odd lists doubled); the rest are no-ops / plain properties.
- Shared state lives in `canvas/context2d.zig` so both contexts stay in step.

Full suite: 1383/1383.


arrufat:
Note on the no-op parts: measureText and font are real (text_measure.zig widths). roundRect and createLinearGradient/createRadialGradient are called unguarded by Chart.js 4 at draw time, which is what motivated this. ellipse, isPointInPath/isPointInStroke and the shadow/filter properties are not backed by a specific caller I checked; happy to drop them if you prefer to keep this to what has a known consumer.

PR #3383, opened 2026-09-01
stability: disconnect runaway intersection observers

arrufat:
## The problem

Without layout we report every attached element as fully visible ([`IntersectionObserver.zig`](https://github.com/lightpanda-io/browser/blob/main/src/browser/webapi/IntersectionObserver.zig) `calculateIntersection`):

```zig
// For a headless browser without real layout, we treat all elements as fully visible.
const intersection_ratio: f64 = if (has_parent) 1.0 else 0.0;
```

That is the right default — the alternative uses `calculateDocumentPosition`'s faux position (preceding-node-count × 5px), which against a 768px viewport would call everything past ~150 nodes invisible and break lazy loading across the board.

Infinite scroll turns it into a loop that never terminates. The page observes a sentinel, we report it intersecting the moment it is attached, the callback loads the next page and observes a *fresh* sentinel, and there is nothing to stop it. `_tracked` already prevents a static element from re-firing, so only this observe-a-new-element pattern gets through, and the chain is async (fetch → render → observe), so the existing depth-style guards would not see it either.

## What it does in production

On a live storefront (`eu.gymshark.com`), roughly **one navigation in seven**: the collection page paginated itself to `?page=10` over **68 delivery sessions** and ~94 requests, exhausted the timer table (`error.TooManyTimeout`), threw a framework client-side exception, and then hung until the watchdog killed it:

```
$scope=frame $level=error $msg="frame.deliverMutations" err=JsException url=.../mens?page=9
$scope=app   $level=error $msg="watchdog stall" stalled_ms=30633
Error: navigation timed out
```

Reproduced offline, deterministically, with a 22-line page that appends a new sentinel per intersection: **50 of 50 pages loaded in a single `goto`**, no scrolling. A real browser loads one, because the replacement sentinel is below the fold.

## The fix

Bound a runaway by how long it keeps delivering, not by how many times, and disconnect the frame's intersection observers past the bound, mirroring `disconnectRunawayMutationObservers` (#105ba20b0).

A delivery count cannot separate a runaway from a busy but healthy page. Uncapped delivery sessions per navigation:

| page | sessions across runs | first to last delivery |
| :---- | :---- | :---- |
| news.ycombinator.com | 0 | none |
| vercel.com, /templates | 0 | none |
| apnews.com hub (ad-heavy) | 5 | under 1 s |
| airbnb.com | 25, 28, 37, 37, 48 | 1.9 to 4.4 s, then quiet |
| airbnb.com/s/Paris/homes | 32, 43, 47, 47, 48 | 1.5 to 2.7 s, then quiet |
| **storefront runaway** | **122 at 4 s, 215 at 7 s** | **never quiet, ~30/s** |

Airbnb legitimately needs up to 48 sessions and the runaway had already exhausted the timer table by 68, with the same URL varying almost 2x run to run. Duration does separate them: every healthy burst went quiet within 4.5 s of its first delivery; the runaway never did.

So deliveries closer than 2 s apart form a burst, and a burst that runs past 10 s disconnects. The clock starts at the beginning of the current burst rather than at the first delivery ever, so lazy loading triggered minutes into a serve or agent session by a scroll or a click is not penalized by page age. The largest gap inside a healthy airbnb burst was 772 ms; the runaway paused once for 1.1 s on a network stall.

A chain that re-observes synchronously never leaves the microtask checkpoint, so the clock alone would let it spin for the full 10 s. A per-burst ceiling of 1024 deliveries exists for that case only; at 20x airbnb's peak it is not a tuning knob.

Logged once per burst rather than per attempt: the framework re-mounts its lazy loader afterwards, and each new observer trips the path again.

## Verification

| check | result |
| :---- | :---- |
| airbnb.com, airbnb.com/s/Paris/homes, 2 loads each | guard never fires |
| HN, apnews, vercel | guard never fires |
| offline sync fixture | disconnected at 1025 deliveries within the first tick |
| offline deferred fixture, burst backdated past 10 s | disconnected on the next delivery |
| offline deferred fixture, quiet gap backdated past 2 s | new burst, observers kept |
| `zig fmt --check ./*.zig ./**/*.zig` | clean |

## Limitations

This bounds the damage; it does not fix the visibility model.

- In fetch mode the storefront runaway is finite: it paginates to page 17 in 4 to 8 s and ends on its own before the 10 s limit, so the guard does not fire and all 17 pages load. The serve-mode hang, which ran 30 s until the watchdog, is cut at 10 s.
- A page that legitimately keeps delivering for more than 10 s without a 2 s pause loses its observers, and the disconnect is as blunt here as it is for mutation observers.
- The 10 s and 2 s constants are not configurable yet.

The follow-up I would want, and did not attempt because it is a bigger design call than a safety valve: report intersections against a real viewport rather than treating every attached element as visible, so `scroll` becomes the thing that legitimately advances pagination.

### DIFF
--- a/src/browser/frame/observers.zig
+++ b/src/browser/frame/observers.zig
@@ -341,14 +341,11 @@ fn disconnectRunawayIntersectionObservers(frame: *Frame) void {
         log.err(.frame, "frame.IntersectionRunaway", .{ .type = frame._type, .url = frame.url });
     }
 
-    var i = frame._intersection.observers.items.len;
-    while (i > 0) {
-        i -= 1;
-        if (i >= frame._intersection.observers.items.len) {
-            continue;
-        }
-        frame._intersection.observers.items[i].disconnect(frame);
+    for (frame._intersection.observers.items) |observer| {
+        observer.reset(frame._page);
+        observer.releaseRef(frame._page);
     }
+    frame._intersection.observers.clearRetainingCapacity();
 }
 
 pub fn deliverIntersections(frame: *Frame) void {
--- a/src/browser/frame/observers.zig
+++ b/src/browser/frame/observers.zig
@@ -58,9 +58,12 @@ pub const Intersection = struct {
     check_scheduled: bool = false,
     delivery_scheduled: bool = false,
 
-    // Delivery sessions run against this document. Reset per navigation, since
-    // Frame.init re-initializes the whole struct.
-    deliveries: u32 = 0,
+    // Deliveries closer together than INTERSECTION_QUIET_MS form one burst.
+    // Frame.init re-initializes the whole struct, so a navigation starts fresh.
+    burst_start_ms: u64 = 0,
+    last_delivery_ms: u64 = 0,
+    burst_deliveries: u32 = 0,
+    runaway: bool = false,
 };
 
 // ResizeObserver bookkeeping for a frame.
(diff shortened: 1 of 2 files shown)


arrufat:
Follow-up on the constant, since "why 32?" is the obvious review question.

**What real pages need.** Instrumented delivery-session counts per navigation, before the guard existed:

| page | sessions |
| :---- | ----: |
| news.ycombinator.com | 0 |
| apnews.com hub (ad-heavy) | 5 |
| storefront collection, normal load | 5 |
| storefront collection, partial pagination | 7 |
| storefront collection, runaway | 68 |

So 32 is ~4.5x the worst legitimate case observed and well under the runaway.

**I also measured 8, and it is not better.** Built and tested it against the same storefront through `lightpanda serve` + puppeteer:

- No false positives: zero trips across 12 loads of Hacker News, the apnews hub and healthy storefront pages.
- Tests pass, including the runaway test in this PR.
- Performance: two experiments disagreed. A paired, interleaved A/B (8 rotations, Chrome anchor) put limit 8 **slower** — 7,211 ms vs 5,623 ms, slower in 6 of 8 rotations. A smaller sequential check (3 runs each) put it faster. That site's per-run variance spans roughly -56% to +66%, so the difference between 8 and 32 is not resolvable there.

There is a plausible mechanism for tightening to hurt, which is really a limitation of the approach rather than of the number: the guard disconnects **all** of the frame's intersection observers, so firing earlier also stops lazy loading the page legitimately still needs. That is the argument for the follow-up already noted in the description — throttling, or ignoring only newly observed targets, instead of disconnecting the frame's observers outright.

**One more data point in favour of merging something here:** on that storefront today, the guard trips on essentially every `lightpanda serve` load (6 of 6 runs across both limits). The runaway is not an edge case on sites that paginate this way.

Happy to change the constant or the response if you would rather; I did not want to tune it further against a site this noisy.

karlseguin:
Last time I looked, airbnb and vercel can produce a lot, but not in an endless loop, it eventually settles down. So, I'm uneasy about the limit. And I'm uneasy about no limit.

I don't have a great solution. I'm thinking of a time-based limit....disconnect T seconds after the first one where T is 10 or configurable.

arrufat:
Fair, and it turns out to be measurable. I instrumented uncapped delivery sessions per navigation:

| page | sessions across runs | first to last delivery |
| :--- | :--- | :--- |
| airbnb.com | 25, 28, 37, 37, 48 | 1.9 to 4.4 s, then quiet |
| airbnb.com/s/Paris/homes | 32, 43, 47, 47, 48 | 1.5 to 2.7 s, then quiet |
| vercel.com, /templates | 0 | none |
| storefront runaway | 122 at 4 s, 215 at 7 s | never quiet, about 30/s |

So you're right: 32 fired on 3 of 4 airbnb loads, and my earlier "no false positives" did not cover airbnb. A count can't be tuned out of it either. Airbnb legitimately reaches 48 and the runaway had already exhausted the timer table by 68, with the same URL varying almost 2x run to run.

Duration does separate them. Every airbnb burst went quiet within 4.5 s of its first delivery; the runaway never did. I've switched the guard to your time-based idea with one adjustment (6b24f1e86): the clock is measured from the start of the current burst, not from the first delivery ever. Deliveries within 2 s of each other belong to the same burst, and a burst that runs past 10 s disconnects. Measured from the first delivery, a page in serve or agent mode that lives for minutes would lose lazy loading at the 10 s mark regardless of activity, which is a worse false-positive surface than the count. Largest gap inside a healthy airbnb burst was 772 ms; the runaway paused once for 1.1 s on a network stall, so 2 s has margin on both sides.

Two things to know:

- A chain that re-observes synchronously never leaves the microtask checkpoint, so the clock alone would let it spin for 10 s of CPU. There is a per-burst ceiling of 1024 deliveries for that case only. At 20x airbnb's peak it isn't a limit in your sense.
- The storefront runaway is actually finite in fetch mode. It ends around page 17 in 4 to 8 s, under the 10 s limit, so the guard does not fire there and the page loads all 17 pages. The serve-mode hang, which ran 30 s until the watchdog, gets cut at 10 s instead of the 4.4 s the count cap gave. That is the price of not clipping airbnb, and the real fix stays the viewport model.

T is a constant for now. Happy to plumb it through Config if you want it configurable, or to pick a different T; 10 s has about 2x margin over the slowest airbnb burst.

PR #3391, opened 2026-09-02
webapi: resource-timing

### LINKED ISSUE #3359: performance.getEntriesByType('resource') returns empty, breaking CDP networkidle detection

karlseguin:
Adds resource timing, e.g. `performance.getEntriesByType("resource")`.

Based off the https://github.com/lightpanda-io/browser/pull/3389 branch.

The `resource-timing` WPT category is currently at 4.6%, and this is a first step at improving it. It also hopefully fixes https://github.com/lightpanda-io/browser/issues/3359

This is more complicated than I thought because there's a "Timing-Allow-Origin" header that a server can include which hides some of the data if the request doesn't come from the listed origin. And that, of course, interacts with redirects.

(The DOMException change is seemingly random, but it came up in one of the WPT cases I was looking at).

### DIFF
--- a/src/browser/webapi/Performance.zig
+++ b/src/browser/webapi/Performance.zig
@@ -21,6 +21,7 @@ const lp = @import("lightpanda");
 
 const js = @import("../js/js.zig");
 const Factory = @import("../Factory.zig");
+const Scheduler = @import("../js/Scheduler.zig");
 
 const EventCounts = @import("EventCounts.zig");
 const PerformanceObserver = @import("PerformanceObserver.zig");
@@ -51,9 +52,11 @@ _navigation: PerformanceNavigation = .{},
 _event_counts: EventCounts = .{},
 _resource_buffer_size: u32 = DEFAULT_RESOURCE_BUFFER_SIZE,
 
-// Captured when the first observer registers. Entries created outside of
-// JS (resource timing, from the HTTP client) need it to notify observers.
-_exec: ?*const Execution = null,
+// The owner's task scheduler, set by the owner once its JS context exists
+// (Frame builds the Window, and thus this, before the context). Never taken
+// from a caller's Execution: an isolated-world or cross-frame caller's
+// context can be destroyed while we live on.
+_scheduler: *Scheduler = undefined,
 
 // PerformanceObserver infrastructure. Lives here (rather than on the owning
 // Frame/WorkerGlobalScope) so that both contexts get observers for free.
@@ -112,7 +115,7 @@ pub fn mark(
     const start_time = opts.startTime orelse self.now();
     const m = try Mark.init(name, opts.detail, start_time, exec);
     try self.insertOrdered(&self._entries, m._proto);
-    try self.notifyObservers(m._proto, exec);
+    try self.notifyObservers(m._proto);
     return m;
 }
 
@@ -163,7 +166,7 @@ pub fn measure(
                 exec,
             );
             try self.insertOrdered(&self._entries, m._proto);
-            try self.notifyObservers(m._proto, exec);
+            try self.notifyObservers(m._proto);
             return m;
         },
         .start_mark => |start_mark| {
@@ -187,14 +190,14 @@ pub fn measure(
                 exec,
             );
             try self.insertOrdered(&self._entries, m._proto);
-            try self.notifyObservers(m._proto, exec);
+            try self.notifyObservers(m._proto);
             return m;
         },
     };
 
     const m = try Measure.init(name, null, 0.0, self.now(), null, exec);
     try self.insertOrdered(&self._entries, m._proto);
-    try self.notifyObservers(m._proto, exec);
+    try self.notifyObservers(m._proto);
     return m;
 }
 
@@ -260,9 +263,10 @@ pub const ResourceInfo = struct {
 };
 
 pub fn addResource(self: *Performance, info: ResourceInfo) !void {
-    if (self._resources.items.len >= self._resource_buffer_size) {
-        // TODO: Perfomance should be an EventTarget and it should fire
-        // a resourcetimingbufferfull event
+    // TODO: Perfomance should be an EventTarget and it should fire
+    // a resourcetimingbufferfull event
+    const buffer_full = self._resources.items.len >= self._resource_buffer_size;
+    if (buffer_full and !self.hasObserverFor(.resource)) {
         return;
     }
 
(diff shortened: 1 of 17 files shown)


src/browser/webapi/Performance.zig
```diff
@@ -20,24 +20,40 @@ const std = @import("std");
...
+// polled more)
+_resources: std.ArrayList(*Entry) = .empty,
 _timing: PerformanceTiming = .{},
 _navigation: PerformanceNavigation = .{},
 _event_counts: EventCounts = .{},
+_resource_buffer_size: u32 = DEFAULT_RESOURCE_BUFFER_SIZE,
+
+// Captured when the first observer registers. Entries created outside of
+// JS (resource timing, from the HTTP client) need it to notify observers.
+_exec: ?*const Execution = null,
```
krichprollsch: Opencode (yes 🙄 ) points a risk of UAF here. The lifetime of the performance can be longer than the captured execution context.

src/browser/webapi/Performance.zig
```diff
@@ -205,46 +223,255 @@ pub fn clearMeasures(self: *Performance, measure_name: ?[]const u8) void {
...
+    content_type: []const u8, // unsafe to reference past the function call
+    content_encoding: []const u8, // unsafe to reference past the function call
+    from_cache: bool,
+    // Every response in the chain was same-origin or carried a matching
+    // Timing-Allow-Origin.
+    timing_allow: bool,
+};
+
+pub fn addResource(self: *Performance, info: ResourceInfo) !void {
+    if (self._resources.items.len >= self._resource_buffer_size) {
```
krichprollsch: opencode thinks we should notify observers here before early return.

src/browser/webapi/PerformanceObserver.zig
```diff
@@ -73,7 +73,6 @@ const ObserveOptions = struct {
     type: ?[]const u8 = null,
 };
 
-/// TODO: Support `buffered` option.
 pub fn observe(self: *PerformanceObserver, maybe_options: ?ObserveOptions, exec: *const Execution) !void {
```
krichprollsch: We shouldn't accept `observe({ entryTypes: ["resource"], buffered: true })` combination.

src/network/HttpClient.zig
```diff
@@ -2669,6 +2683,183 @@ pub const Transfer = struct {
...
+        target.performance.addResource(.{
+            .name = t.url,
+            .initiator = switch (self.req.resource_type) {
+                // A document fetch only makes the timeline as a child frame.
+                .document => "iframe",
+                .xhr => "xmlhttprequest",
+                .script => "script",
+                .fetch => "fetch",
+                .stylesheet => "link",
+                .eventsource, .worker => "other",
```
krichprollsch, line 2830: Opencode thinks worker should be mapped to script instead 🤔

src/browser/webapi/PerformanceObserver.zig
karlseguin: It's true that the spec says we shouldn't. But I tested in both Firefox and Chrome and they both just ignore the buffered value when entryTypes is provided (Chrome logs a warning).

src/network/HttpClient.zig
karlseguin: I tested this in our worker.html test with firefox, and I believe we have it correct. a "classic" worker appears as "other", but a module worker appears as "script". For a module worker, the `resource_type` would be "script", so it would show up as "script".

src/browser/webapi/Performance.zig
karlseguin: Yes, good catch. The Frame/Worker push their scheduler in now.

src/browser/webapi/Performance.zig
karlseguin: Yes. Added. Worth noting the follow up is to make Performance inherit EventTarget to have the correct complete handling here (`resourcetimingbufferfull`)

krichprollsch (approved):

PR #3397, opened 2026-09-03
CDP: create an context per isolated world per frame

### LINKED ISSUE #3347: CDP: `Page.createIsolatedWorld(childFrameId)` evaluates the main document
## Summary

When `Page.createIsolatedWorld` is called with the ID of a loaded cross-origin child frame, Lightpanda returns an execution-context ID that evaluates against the **main document**, not the requested child document.

The behavior reproduces on both the latest stable release (`0.3.7`) and the current nightly (`1.0.0-nightly.9021+27831c20b`). Chromium evaluates the correct child document with the same CDP sequence.

The reproduction uses two local HTTP fixture servers on different ports. It makes no requests to third-party websites.

## Environment

| Component | Version tested |
|---|---|
| Lightpanda stable | `0.3.7`, tag commit [`5b0835f5a17fbb6fd45b56e091a14dded2e887b0`](https://github.com/lightpanda-io/browser/commit/5b0835f5a17fbb6fd45b56e091a14dded2e887b0) |
| Stable Docker image | `linux/amd64`, `lightpanda/browser@sha256:eae2c7f8ff39a7a24540d21c7561ce219fd14b1bf2b92403930c6d140f2d1337` |
| Lightpanda nightly | `1.0.0-nightly.9021+27831c20b`, commit [`27831c20bf50bfd3220a89643e85e192df737ca2`](https://github.com/lightpanda-io/browser/commit/27831c20bf50bfd3220a89643e85e192df737ca2) |
| Nightly Docker image | Created `2026-08-30T02:50:37Z`, `linux/amd64`, `lightpanda/browser@sha256:d22c9cc1df1e6f081d90deb0fee96938014262f49406c547cda772982229103a` |
| Chromium control | Playwright `1.62.0` browser: Chrome for Testing `151.0.7922.34`, macOS arm64 binary SHA-256 `7687bff7cb2db075f250e6d5848bbc8838cac3802ac3952a899c574f8eccab45` |
| CDP client | `github.com/chromedp/chromedp v0.14.2` |
| Go used for the run | `go1.27.0 darwin/arm64` |
| Container runtime | Docker Desktop engine `28.3.2` |

Nightly disables iframe resource loading by default, so its command below includes `--load-resources iframe`. Without that flag, the missing child frame is expected and is not the bug described here.

## Reproduction

The complete self-contained Go reproducer, dependency manifest, and checksum file are available in this public Gist:

**https://gist.github.com/viktor-shcherb/cd1175056d7ac4d755e29d4a21ab6576**

```bash
git clone https://gist.github.com/cd1175056d7ac4d755e29d4a21ab6576.git lightpanda-iframe-repro
cd lightpanda-iframe-repro
go mod verify
```

The fixture creates:

- a parent document titled `Parent jobs`;
- a cross-origin iframe initially titled `Jobs page one`; and
- a second iframe document titled `Jobs page two`.

The relevant CDP sequence is:

```go
frameTree, err := page.GetFrameTree().Do(ctx)
if err != nil {
    return err
}

mainFrameID := frameTree.Frame.ID
childFrameID := frameTree.ChildFrames[0].Frame.ID

mainContextID, err := page.CreateIsolatedWorld(mainFrameID).
    WithWorldName("repro-main-utility").
    Do(ctx)
if err != nil {
    return err
}

childContextID, err := page.CreateIsolatedWorld(childFrameID).
    WithWorldName("repro-child-utility").
    Do(ctx)
if err != nil {
    return err
}

childTitle, exception, err := runtime.Evaluate("document.title").
    WithContextID(childContextID).
    WithReturnByValue(true).
    Do(ctx)
```

The program evaluates the child title, changes the iframe from `/frame-one` to `/frame-two`, obtains the navigated child frame, creates a new isolated world for it, and evaluates its title again. It also evaluates through the original main-frame context to verify that child navigation did not retarget or invalidate the main world.

### Latest stable Lightpanda

```bash
docker run --rm --detach \
  --platform linux/amd64 \
  --name lightpanda-iframe-repro \
  --publish 127.0.0.1:9222:9222 \
  lightpanda/browser@sha256:eae2c7f8ff39a7a24540d21c7561ce219fd14b1bf2b92403930c6d140f2d1337

go run . \
  -backend lightpanda-0.3.7 \
  -cdp ws://127.0.0.1:9222 \
  -parent-listen 0.0.0.0:18080 \
  -child-listen 0.0.0.0:18081 \
  -parent-url http://host.docker.internal:18080/ \
  -child-origin http://host.docker.internal:18081
```

### Current nightly

```bash
docker run --rm --detach \
  --platform linux/amd64 \
  --name lightpanda-iframe-repro \
  --publish 127.0.0.1:9222:9222 \
  lightpanda/browser@sha256:d22c9cc1df1e6f081d90deb0fee96938014262f49406c547cda772982229103a \
  /bin/lightpanda serve \
  --host 0.0.0.0 \
  --port 9222 \
  --load-resources iframe

go run . \
  -backend lightpanda-nightly-27831c20b \
  -cdp ws://127.0.0.1:9222 \
  -parent-listen 0.0.0.0:18080 \
  -child-listen 0.0.0.0:18081 \
  -parent-url http://host.docker.internal:18080/ \
  -child-origin http://host.docker.internal:18081
```

### Chromium control

```bash
"$CHROMIUM_BIN" \
  --headless \
  --no-sandbox \
  --remote-debugging-address=127.0.0.1 \
  --remote-debugging-port=9223 \
  --user-data-dir="$CHROMIUM_PROFILE" \
  about:blank

CHROMIUM_WS="$(curl --silent http://127.0.0.1:9223/json/version | jq -r .webSocketDebuggerUrl)"

go run . \
  -backend chromium-151.0.7922.34 \
  -cdp "$CHROMIUM_WS" \
  -parent-listen 127.0.0.1:18080 \
  -child-listen 127.0.0.1:18081 \
  -parent-url http://127.0.0.1:18080/ \
  -child-origin http://127.0.0.1:18081
```

## Expected behavior

- The main-frame world evaluates `Parent jobs`.
- The initial child-frame world evaluates `Jobs page one`.
- After child navigation, a newly created child-frame world evaluates `Jobs page two`.
- Child navigation does not invalidate or retarget the saved main-frame world.

Chromium produces:

```json
{
  "backend": "chromium-151.0.7922.34",
  "title_before_child_navigate": "Parent jobs",
  "title_after_child_navigate": "Parent jobs",
  "child_title_before_navigate": "Jobs page one",
  "child_title_after_navigate": "Jobs page two"
}
```

## Actual behavior

Both tested Lightpanda builds return the parent title from execution contexts requested for the child frame.

Lightpanda `0.3.7`:

```json
{
  "backend": "lightpanda-0.3.7",
  "title_before_child_navigate": "Parent jobs",
  "title_after_child_navigate": "Parent jobs",
  "child_title_before_navigate": "Parent jobs",
  "child_title_after_navigate": "Parent jobs"
}
```

Nightly `27831c20b`, with iframe loading enabled, produces the same result:

```json
{
  "backend": "lightpanda-nightly-27831c20b",
  "title_before_child_navigate": "Parent jobs",
  "title_after_child_navigate": "Parent jobs",
  "child_title_before_navigate": "Parent jobs",
  "child_title_after_navigate": "Parent jobs"
}
```

The saved main-frame execution context remains valid. The incorrect behavior is specifically that the child-frame contexts resolve to the main document.

## Downstream usage context: Jobseek

[Jobseek](https://github.com/colophon-group/jobseek) is an open-source crawler that monitors company career pages. It is migrating its crawler runtime from Python + Playwright/Chromium to Go + self-hosted Lightpanda. Browser capabilities that Lightpanda cannot execute remain assigned to the isolated Chromium compatibility backend.

This child-frame functionality is needed by Jobseek's rendered DOM monitor:

- The [`hexagon-robotics-careers` board configuration](https://github.com/colophon-group/jobseek/blob/795e76621ff6afeec3cbe30dbce3505fbefe4b02/apps/crawler/data/boards.csv#L2880) identifies an Onlyfy iframe and an in-frame pagination selector.
- The [`_execute_repeat` implementation](https://github.com/colophon-group/jobseek/blob/795e76621ff6afeec3cbe30dbce3505fbefe4b02/apps/crawler/src/shared/browser.py#L1159-L1228) evaluates link counts, dispatches the pagination click, and extracts resulting job links inside the child frame.
- The extracted iframe links are injected into the main document so the ordinary DOM link-discovery pipeline can process them.

The concrete board embeds an Onlyfy job list and requires repeated clicks on its in-frame pagination control:

```json
{
  "page": "https://robotics.hexagon.com/careers/",
  "frame": "iframe[src*=\"onlyfy\"]",
  "action": "repeat-click",
  "selector": "a.infinite-next"
}
```

The automation must count links, click `a.infinite-next`, and extract the resulting links inside the child frame. When the child execution context points to the parent document, the selector is evaluated in the wrong DOM and the interaction cannot be completed. Disabling subframes is not a workaround because the required content exists only inside the iframe.

Until child-frame isolated worlds target the correct document, this configuration must continue using Chromium rather than Lightpanda. That leaves a concrete Chromium-only capability in Jobseek's migration instead of allowing the shared browser executor to route the board to Lightpanda.

## Suggested acceptance test

Add a CDP integration fixture with one main frame and one cross-origin iframe and assert that:

1. `Page.createIsolatedWorld` returns independently usable contexts for the main and child frames;
2. each context evaluates the correct `document.title`;
3. navigating only the child allows a new child world to evaluate the new child document;
4. the original main world still evaluates the main document; and
5. the behavior works through direct CDP and drivers that use isolated utility worlds.

## Potentially related

- [#2400 — Child iframe navigation invalidates main frame's executionContextId for CDP drivers](https://github.com/lightpanda-io/browser/issues/2400). That issue describes a shared per-`BrowserContext` isolated-world/context-group design. The main context remains stable in this reproduction, but the child-world aliasing may share the same underlying cause.
- [#2401 — optional subframe disabling](https://github.com/lightpanda-io/browser/pull/2401). Disabling iframe loading avoids iframe context handling, but it cannot support automation where the target content is inside the iframe.

karlseguin:
Previously, we created a single context per isolated world for the main frame. We now create an isolate world per frame.

Fixes: https://github.com/lightpanda-io/browser/issues/3347 & https://github.com/lightpanda-io/browser/issues/2400

### DIFF
--- a/src/server/cdp/CDP.zig
+++ b/src/server/cdp/CDP.zig
@@ -640,17 +640,26 @@ pub const BrowserContext = struct {
         self.set_child_nodes_sent.clearRetainingCapacity();
     }
 
-    pub fn createIsolatedWorld(self: *BrowserContext, world_name: []const u8, grant_universal_access: bool) !*IsolatedWorld {
-        // The name is the world's identity (matching Chrome). Clients re-issue
-        // this call after every navigation; appending a duplicate each time
-        // would grow the per-page context count without bound.
+    pub const GetOrPutIsolatedWorld = struct {
+        world: *IsolatedWorld,
+        found_existing: bool,
+    };
+
+    pub fn findIsolatedWorld(self: *const BrowserContext, world_name: []const u8) ?*IsolatedWorld {
         for (self.isolated_worlds.items) |world| {
             if (std.mem.eql(u8, world.name, world_name)) {
-                if (world.grant_universal_access != grant_universal_access) {
-                    log.warn(.cdp, "isolated world mismatch", .{ .name = world_name, .gua = grant_universal_access });
-                }
                 return world;
             }
+        }
+        return null;
+    }
+
+    pub fn createIsolatedWorld(self: *BrowserContext, world_name: []const u8, grant_universal_access: bool) !GetOrPutIsolatedWorld {
+        if (self.findIsolatedWorld(world_name)) |world| {
+            if (world.grant_universal_access != grant_universal_access) {
+                log.warn(.cdp, "isolated world mismatch", .{ .name = world_name, .gua = grant_universal_access });
+            }
+            return .{ .world = world, .found_existing = true };
         }
 
         const browser = &self.cdp.browser;
@@ -667,7 +676,19 @@ pub const BrowserContext = struct {
 
         try self.isolated_worlds.append(self.arena, world);
 
-        return world;
+        return .{ .world = world, .found_existing = false };
+    }
+
+    // only called when we fail to fully create a world (e.g. errdefer in
+    // Page.createIsolatedWorld).
+    pub fn removeIsolatedWorld(self: *BrowserContext, world: *IsolatedWorld) void {
+        for (self.isolated_worlds.items, 0..) |w, i| {
+            if (w == world) {
+                _ = self.isolated_worlds.swapRemove(i);
+                world.deinit();
+                return;
+            }
+        }
     }
 
     pub fn nodeWriter(self: *BrowserContext, root: *const NodeRegistry.Node, opts: Node.Writer.Opts) Node.Writer {
@@ -1165,12 +1186,16 @@ pub const BrowserContext = struct {
 const ScriptOnNewDocument = struct {
     identifier: u32,
     source: []const u8,
+    // Page.addScriptToEvaluateOnNewDocument's worldName. null means the main
+    // world. A named world is seeded into every frame (see IsolatedWorld).
+    world_name: ?[]const u8,
 };
 
 /// An isolated world is identified by its name and has one V8::Context per
-// it was requested it. Generally the client needs to resolve a node into the
-// isolated world to be able to work with it.
-///
+/// frame it has been seeded into. A world enters a frame on an explicit
+/// trigger: Page.createIsolatedWorld or, or a preload script which is seeded
+/// into every frame. Once seeded, the frame's context is rebuilt on every
+/// navigation with no further client involvement.
 /// Frame ids are stable across a child frame's re-navigation (the Frame is
 /// torn down and re-initialized in place), so a per-frame context is removed
 /// on frame_destroyed and created again on the frame's next frame_navigated.
@@ -1180,6 +1205,9 @@ pub const IsolatedWorld = struct {
     name: []const u8,
     grant_universal_access: bool,
     contexts: std.ArrayList(FrameContext) = .empty,
+
+    // Frames this world has been seeded into, by frame id.
+    seeded_frames: std.ArrayList(u32) = .empty,
 
     // Identity tracking for this isolated world (separate from main world).
     // Shared by all of the world's frame contexts, like the main world shares
@@ -1199,6 +1227,17 @@ pub const IsolatedWorld = struct {
     pub fn deinit(self: *IsolatedWorld) void {
         self.removeAllContexts();
         self.arena.release();
+    }
+
+    pub fn seed(self: *IsolatedWorld, frame_id: u32) !void {
+        if (self.isSeeded(frame_id)) {
+            return;
+        }
+        return self.seeded_frames.append(self.arena.allocator(), frame_id);
+    }
+
+    pub fn isSeeded(self: *const IsolatedWorld, frame_id: u32) bool {
+        return std.mem.indexOfScalar(u32, self.seeded_frames.items, frame_id) != null;
     }
 
     // Keyed by Frame, not frame id: a retired root Page keeps its frame id
--- a/src/server/cdp/CDP.zig
+++ b/src/server/cdp/CDP.zig
@@ -1271,6 +1271,7 @@ pub const IsolatedWorld = struct {
             .debug_name = "IsolatedContext",
         });
         errdefer browser.env.destroyContext(ctx);
+        try ctx.setOrigin(frame.origin);
 
         try self.contexts.append(self.arena.allocator(), .{
             .frame = frame,
--- a/src/server/cdp/CDP.zig
+++ b/src/server/cdp/CDP.zig
@@ -568,6 +568,7 @@ pub const BrowserContext = struct {
         try notification.register(.frame_navigated_within_document, self, onFrameNavigatedWithinDocument);
         try notification.register(.frame_navigate_failed, self, onFrameNavigateFailed);
         try notification.register(.frame_child_frame_created, self, onFrameChildFrameCreated);
+        try notification.register(.frame_destroyed, self, onFrameDestroyed);
         try notification.register(.frame_dom_content_loaded, self, onFrameDOMContentLoaded);
         try notification.register(.frame_loaded, self, onFrameLoaded);
         try notification.register(.javascript_dialog_opening, self, onJavascriptDialogOpening);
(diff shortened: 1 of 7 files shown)


krichprollsch:
Does it impact #2400?

karlseguin:
> Does it impact #2400?

Yes, with the last commit, I believe both issues will be fixed. Worth pointing out that this _will_ increase memory when iframes are loaded (for drivers that use this, which Puppeteer and Playwright do).

src/server/cdp/domains/page.zig
```diff
@@ -158,13 +158,26 @@ fn addScriptToEvaluateOnNewDocument(cmd: *CDP.Command) !void {
...
 
+    // A worldName registers the world itself:
+    var world_name: ?[]const u8 = null;
+    if (params.worldName) |name| {
+        if (name.len > 0) {
+            const gop = try bc.createIsolatedWorld(name, true);
+            // Borrowed from the world's arena, which outlives the script: a
+            // world is only ever removed on the createIsolatedWorld command's
+            // own error path, before any script can name it.
+            world_name = gop.world.name;
```
krichprollsch: We should copy the string into the bc's arena, the isolated world could be destroyed before the script.
Or we must cleanup scripts when we destroy the isolated world.

src/server/cdp/CDP.zig
```diff
@@ -1168,66 +1186,131 @@ pub const BrowserContext = struct {
...
+    contexts: std.ArrayList(FrameContext) = .empty,
+
+    // Frames this world has been seeded into, by frame id.
+    seeded_frames: std.ArrayList(u32) = .empty,
 
     // Identity tracking for this isolated world (separate from main world).
-    // This ensures CDP inspector contexts don't share v8::Globals with main world.
+    // Shared by all of the world's frame contexts, like the main world shares
+    // Page.identity across frames, and reset with them on root teardown.
     identity: js.Identity = .{},
```
krichprollsch, line 1215: Isn't it a problem having multiple frame context sharing the same identity map?

src/server/cdp/CDP.zig
```diff
@@ -1168,66 +1186,131 @@ pub const BrowserContext = struct {
...
-            self.context = null;
+    pub fn seed(self: *IsolatedWorld, frame_id: u32) !void {
+        if (self.isSeeded(frame_id)) {
+            return;
         }
-        // I don't think it's possible to have any identity without a context,
-        // but there's no harm in being safe.
-        self.identity.deinit();
-        self.identity = .{};
+        return self.seeded_frames.append(self.arena.allocator(), frame_id);
```
krichprollsch, line 1236: shouldn't we remove the id from the `seeded_frames` on removeContext? but maybe we don't really care.

src/server/cdp/domains/dom.zig
```diff
@@ -396,6 +381,51 @@ fn resolveNode(cmd: *CDP.Command) !void {
     } }, .{});
 }
 
+// The frame owning the node's document. Synthetic documents (DOMParser,
+// DOMImplementation) have no frame and fall back to the root.
+fn nodeFrame(dom_node: *DOMNode, root: *Frame) *Frame {
+    const document = if (dom_node._type == .document)
+        dom_node.subtype(DOMNode.Document)
+    else
+        dom_node.ownerDocument(root) orelse return root;
```
krichprollsch, line 390: opencode thinks we have an issue w/ detached node from a frame which will fallback to root.

src/server/cdp/CDP.zig
karlseguin: The entire point of `seeded_frames` is to track the frame_id to identity which frames an isolated world should be applied to. On navigate, we fire removeContext, and then re-create the frame with the same frame_id..it's `seeded_frames` we use to make sure the new frame (with the old id) automatically gets its isolated world.

src/server/cdp/domains/dom.zig
karlseguin: That's true, but it's generally consistent with the rest of the codebase (e.g. `Node.ownerFrame` more or less does the same). Fixing this is a bigger change than this PR.

src/server/cdp/CDP.zig
karlseguin: No, this is why the identity map is stored on the page. Two frames in the same origin get the same v8::Object instance.

karlseguin: But that did make me realize that we don't give the context an origin, so it keeps its opaque origin and never gets the correct SecurityToken. That's fixed now.

krichprollsch (approved):

PR #3398, opened 2026-09-03
chore: Move ownership of the Inbox from HttpClient to WebSocker Driver

karlseguin:
Currently, the HttpClient owns the inbox and its borrowed by the Link. This is a bit backwards, but it also means that we can't eagerly create a Link: the Link needs the inbox, so it needs the HttpClient, which is created by the Browser (which creates an Isolate).

Remember, the Inbox is one of the few things shared between the main thread and the worker, so either end can own it and the other can borrow it.

This switches the ownership so that the HttpClient now borrows the Inbox from the Server's side of the Link (the WebSocket).

The main goal of this change is to prepare for more advanced HTTP WebDriver flows. The more we can create _without_ a Browser, the fewer edge cases we have to deal with (Browser because it's expensive and has to be created on the Worker thread due to how V8::Isolate works).

### DIFF
--- a/src/server/Server.zig
+++ b/src/server/Server.zig
@@ -585,6 +585,7 @@ fn releaseWorker(self: *Server, ws: *WebSocket, notify: *std.Io.Event) void {
 // or upgradeConnection when there never was a worker.
 fn releaseWebSocket(self: *Server, ws: *WebSocket) void {
     self.websockets.remove(&ws.node);
+    ws.driver = null;
     ws.inbox.deinit();
     lp.metrics.serve_active_connections.decr(ws.protocol);
     self.websocket_pool.release(ws);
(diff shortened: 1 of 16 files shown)


src/server/Server.zig
```diff
@@ -582,6 +585,7 @@ fn releaseWorker(self: *Server, ws: *WebSocket, notify: *std.Io.Event) void {
 // or upgradeConnection when there never was a worker.
 fn releaseWebSocket(self: *Server, ws: *WebSocket) void {
     self.websockets.remove(&ws.node);
```
krichprollsch, line 587: What about setting `ws.driver = null;` here?

krichprollsch (approved):

PR #3401, opened 2026-09-04
webdriver: HTTP WebDriver session management

karlseguin:
This is a small step towards WebDriver supports (non-bidi). It allows creating and deleting a BiDi "Session" (e.g. a worker). It also allows attaching a BiDi driver to an HTTP-created BiDi session (the typical selenium startup flow).

This change unblocks the most basic setup/teardown of Selenium, so it still isn't enough to actually use a Selenium script as-is. But it's significant because it models a worker (thread) that isn't tied to a WebSocket, something we haven't had before.

A consequence of a pure HTTP Session is that we don't have a clear cleanup signal. There is no "the socket is disconnected". There's a new HTTP reaper which kills HTTP Sessions after --http-session-timeout. It's expected that drivers properly DELETE /session/:id. I imagine we're going to run into --cdp-max-connections limits and need to tweak this code. BUT, this entire flow is only enabled with --protocol webdriver, so it won't impact exiting CDP users.


src/server/Server.zig
```diff
@@ -1100,6 +1175,175 @@ const KQueue = struct {
...
+        switch (self.protocol) {
+            .cdp => {
+                // only WebDriver has HTTP sessions
+                const socket = switch (origin) {
+                    .socket => |socket| socket,
+                    .session => return error.NoSocket,
+                };
+                const cdp = try allocator.create(CDP);
+                defer allocator.destroy(cdp);
+                try cdp.init(server.app, socket, &self.inbox);
```
krichprollsch, line 1291: It seems there is a small room here were we don't close the socket in case of error. same in the bidi side.

src/server/http.zig
```diff
@@ -635,74 +631,101 @@ fn gateOpen(server: *const Server, gate: Route.Gate) bool {
...
+    };
+
+    const url: ?[]const u8 = blk: {
+        if (is_requesting_websocket_url) {
+            break :blk try std.fmt.allocPrint(req.arena, "{s}{s}", .{ server.bidi_session_url, &session_id });
+        }
+        break :blk null;
+    };
 
     return serveWebDriver(server, conn, req, "200 OK", .{
```
krichprollsch, line 734: Shouldn't we close the worker in case of error here?

src/server/http.zig
```diff
@@ -713,32 +736,49 @@ fn newSession(server: *Server, conn: *Connection, req: *Connection.Request) !Ser
...
-        if (always.webSocketUrl == true) {
-            return true;
-        }
+// GET /session/ID  (webdriver (upgrade to bidi))
+fn upgradeSession(server: *Server, conn: *Connection, req: *Connection.Request) !Served {
+    const worker = server.findSession(req.session_id.?) orelse {
+        return serveNotFound(server, conn, req);
+    };
+
+    if (worker.link != null) {
```
krichprollsch, line 755: It seems on client's disconnect and immediate reconnect, we can hit this case until the link is released.
Maybe the link could include a state to detect a pending release, then we could return a 429?

But it sounds like a nice to have.

src/server/Server.zig
```diff
@@ -626,16 +736,49 @@ fn beginShutdown(self: *Server) void {
...
+            link.shutdown();
+        }
+    }
+}
+
+// HTTP sessions nobody has talked to for --session-timeout.
+fn reapSessions(self: *Server, now: u64) void {
+    while (self.idle_sessions.first) |node| {
+        const worker: *Worker = @fieldParentPtr("idle_node", node);
+        if (worker.deadline.? > now) {
```
krichprollsch, line 759: If we set `--http-session-timeout 0` the deadline will always return immediately.

src/server/bidi/BiDi.zig
```diff
@@ -128,11 +167,47 @@ pub fn deinit(self: *BiDi) void {
...
+    const worker = switch (self.mode) {
+        .bidi_only => return true,
+        .http => |worker| worker,
+    };
+    self.releaseLink(worker);
+    return false;
+}
+
+fn releaseLink(self: *BiDi, worker: *Server.Worker) void {
+    const l = self.link orelse return;
```
krichprollsch: Should we assert the link is not null instead? not sure at all.

src/server/Server.zig
karlseguin: I tightened socket ownership throughout, CDP.init/BiDi.init/Link.init now have a clearer / more explicit ownership path.

src/server/http.zig
karlseguin: Done.

src/server/bidi/BiDi.zig
karlseguin: added a debug-only assertion.

src/server/Server.zig
karlseguin: 0 now means disable the reaper.

PR #3407, opened 2026-09-04
Disarm watchdog before protocol teardown

staylor:
## Context

The server teardown order leaves a race:

```text
Worker.releaseConnection
Env.cancelTerminate
CDP/BiDi.deinit -> Browser.closeSession
Browser.deinit -> Watchdog.unregister
```

The watchdog remains registered after termination is canceled and while protocol teardown disposes V8 contexts. A concurrent scan can therefore request another termination during cleanup. CDP and BiDi both close the session before `Browser.deinit`, so moving unregistration within `Browser.deinit` alone does not close the window.

## Change

Add `Browser.prepareForTeardown`, which unregisters the watchdog and then clears termination. `Watchdog.unregister` synchronizes with an in-progress scan, so no watchdog request can land after the cancel. Server workers call it after releasing the connection and before CDP/BiDi teardown. `Browser.deinit` also calls it as an idempotent fallback for other owners.

## Validation

- `TEST_FILTER="WebApi: MutationObserver terminated page tears down" make test`
- `TEST_FILTER="cdp: run sends a close frame on pending terminate" make test`
- `zig build`
- `git diff --check`

### DIFF
--- a/src/browser/Browser.zig
+++ b/src/browser/Browser.zig
@@ -142,14 +142,13 @@ pub fn init(self: *Browser, app: *App, opts: InitOpts) !void {
 pub fn deinit(self: *Browser) void {
     const allocator = self.allocator;
 
+    self.prepareForTeardown();
+
     self.closeSession();
 
     lp.metrics.js_heap_physical_bytes.add(-@as(i64, @intCast(self.last_reported_js_bytes)));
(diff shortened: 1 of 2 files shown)


karlseguin:
Thanks

PR #3408, opened 2026-09-04
Internal Requests are non interceptable

mookums:
This allows for internal requests to skip intercepting of CDP. This currently affects robots and CORS requests.

### DIFF
--- a/src/network/HttpClient.zig
+++ b/src/network/HttpClient.zig
@@ -1000,7 +1000,7 @@ fn pipeline(self: *Client, transfer: *Transfer, from: SubmitFrom) !void {
                 try wba.signRequest(transfer, authority);
             }
 
-            if (self.serve_mode) {
+            if (self.serve_mode and !transfer.req.internal) {
                 transfer._notify_cdp = true;
                 transfer.notify(.http_request_start, &.{
                     .transfer = transfer,


karlseguin:
If we did this, it should be in the CDP layer...a different driver might have different behaviors. Having said that, I don't see why. I tested with Chrome and saw CORS requests in my RI.

krichprollsch:
Are CORS requests interceptable? I was thinking they were viewable via network but not interceptable, but I'm maybe wrong

karlseguin:
Yes, I tested it. In `demo/public/campfire-commerce/script.js` change the `fetch` to:

```js
const response = await fetch('https://demo-browser.lightpanda.io/campfire-commerce/json/reviews.json', {headers: {'RI': '1'}});
```

Then `RUNS=1 node puppeteer/cdp.js`  and it'll fail (you might need to extend the deadline past 100ms, else it'll fail because of a timeout).

Then change the cdp.js to add RI and provide an alternative CORS response to make it pass:

```js
    const page = await context.newPage();       // <-- existing
    await page.setRequestInterception(true);
    page.on('request', req => {
      if (req.method() == 'OPTIONS') {
          console.log(req.url());
          return req.respond({
            ok: true,
            status: 200,
            headers: {'Access-Control-Allow-Origin': '*', 'access-control-allow-headers': 'RI'}
          });
      }
      req.continue();
    });
```

What would be useful would be e2e tests in demo, especially since CORS interacts distinctly with caching and redirects.

PR #3427, opened 2026-09-06
stability: stop re-observed infinite-scroll sentinels

arrufat:
The timer cap raise was split out and merged as #3445; this PR is now only the observer change.

## Problem

Without layout every attached element intersects, so a load-more sentinel that the page unobserves, loads a batch for and observes again fires on every observe. `eu.gymshark.com`'s listing paginates to page 17 that way: 250-350 product cards where a browser renders 70, 10x the timers, hundreds of Algolia requests, and a CDP `Runtime.callFunctionOn` issued right after `load` waits 1.6-10 s behind the churn (62 ms on Chrome). The burst-duration guard from #3383 only trips after 10 s.

## Fix

Track per frame how many times each target was reported intersecting; past 3 in one navigation treat it as below the fold for every observer (`observers.INTERSECTION_TARGET_REPORT_LIMIT`). Three leaves room for a lazy loader, a prefetcher and an analytics observer on the same element, and lets a page load its second and third batch.

## Result

Storefront collection page: 130 cards and 4 search requests, deterministic (was 250-350 and 250-1000). Cache-matched A/B of the puppeteer retail flow, 6 interleaved rounds: evaluate after load 27-96 ms (was 34 ms-3.2 s), whole flow 4.5 s median (was 7.8 s).

Offline busy-page shapes (timer chains, microtask chains, a long sync task, 2000 timers due at once, a script chain) behave identically on lightpanda and Chrome, so the serve loop is not involved.

### DIFF
--- a/src/browser/webapi/IntersectionObserver.zig
+++ b/src/browser/webapi/IntersectionObserver.zig
@@ -270,8 +270,7 @@ fn checkIntersection(self: *IntersectionObserver, target: *Element, frame: *Fram
     const reported = try frame._intersection.reported.getOrPut(frame.arena, target);
     if (!reported.found_existing) {
         reported.value_ptr.* = 0;
-    }
-    if (reported.value_ptr.* >= Frame.observers.INTERSECTION_TARGET_REPORT_LIMIT) {
+    } else if (reported.value_ptr.* >= Frame.observers.INTERSECTION_TARGET_REPORT_LIMIT) {
         if (!frame._intersection.sentinel_logged) {
             frame._intersection.sentinel_logged = true;
             log.warn(.frame, "IntsctObserver.sentinel", .{ .url = frame.url });
--- a/src/browser/webapi/IntersectionObserver.zig
+++ b/src/browser/webapi/IntersectionObserver.zig
@@ -266,6 +266,21 @@ fn checkIntersection(self: *IntersectionObserver, target: *Element, frame: *Fram
         return;
     }
 
+    // See observers.INTERSECTION_TARGET_REPORT_LIMIT.
+    const reported = try frame._intersection.reported.getOrPut(frame.arena, target);
+    if (!reported.found_existing) {
+        reported.value_ptr.* = 0;
+    }
+    if (reported.value_ptr.* >= Frame.observers.INTERSECTION_TARGET_REPORT_LIMIT) {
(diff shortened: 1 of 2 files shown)


arrufat:
## Known limitation: same-node sentinels stop after three batches

The per-target report cap makes results deterministic and keeps the page responsive. It also means a page that reuses one sentinel node (unobserve, fetch, observe again) stops paginating after three batches for the rest of the navigation.

Scrolling does not help. `window.scrollTo` and `scrollBy` dispatch a `scroll` event but never touch `frame._intersection`. Without layout the sentinel is already "in view", so the usual scrape loop (scroll to the bottom, wait for new items, repeat) has no effect.

Before this PR the same page loaded whatever it could until the 10 s burst guard tripped. That was neither complete nor repeatable: 250 to 350 cards out of 17 pages. So this PR trades a random partial result for a deterministic one. It does not remove a working "load everything" path. Sites that mount a fresh sentinel node per batch are not affected, since a new node starts at zero.

Workarounds for now on same-node sites: open the paginated URL directly (`?page=N`), click a "load more" button when the site has one, or call the site's loader from `evaluate`.

### Proposed follow-up: scrolling re-arms capped sentinels

Reset the per-element counter (or allow one more report) when the page or a tool scrolls down. This matches what a real browser does: each scroll can bring the sentinel back into view and load one more batch. The initial load stays capped at three, and a scraper gets exactly as many batches as it scrolls for. This seems better than making the limit configurable, since the caller should not have to guess the batch count up front. Happy to do that in a separate PR on top of this one.

arrufat:
Follow-up with the scroll re-arm: #3437 (stacked on this branch).

src/browser/webapi/IntersectionObserver.zig
```diff
@@ -266,6 +266,21 @@ fn checkIntersection(self: *IntersectionObserver, target: *Element, frame: *Fram
         return;
     }
 
+    // See observers.INTERSECTION_TARGET_REPORT_LIMIT.
+    const reported = try frame._intersection.reported.getOrPut(frame.arena, target);
+    if (!reported.found_existing) {
+        reported.value_ptr.* = 0;
+    }
+    if (reported.value_ptr.* >= Frame.observers.INTERSECTION_TARGET_REPORT_LIMIT) {
```
karlseguin: `else if`

PR #3428, opened 2026-09-07
render: add "shell" and "clutter" option to --strip-mode

karlseguin:
Some additional API changes:
- Add strip-mode support to pdf/png generation.
- Add LP.dump which provides greater content gathering capability to CDP, exposing most `fetch` dump-related parameters (e.g. format, strip, selector, ...)

The new "--strip-mode shell" is designed to try to remove non-content elements such as the header and footer.  The purpose of this mode is to be relatively safe. It'll fallback to NOT stripping if too much non-link-text is removed with the general idea that stripping less is better than stripping too much.

The new "--strip-mode clutter" is more aggressive and based on readability.js. `clutter.zig` is all Claude.  This will fallback to `shell` only in specific cases (the most notable being that it didn't cut anything OR it cut too much based on the char length). 

Both `shell` and `clutter` require a pre-walk through the document to score things `clutter` also requires a map of nodes-to-prune. So they have a real performance/memory cost.


karlseguin:
No strip-mode:
all

`--strip-mode shell` , in this case, very similar to no strip mode, but the header is removed.
shell

`--strip-mode clutter`
clutter

src/browser/clutter.zig
```diff
@@ -0,0 +1,932 @@
...
+        var current = selected;
+        while (current != self.root) {
+            const parent = current.parentNode() orelse blk: {
+                const shadow = current.is(Node.ShadowRoot) orelse return;
+                break :blk shadow.getHost().asNode();
+            };
+            var it = self.tree.children(parent, false);
+            while (it.next()) |child| {
+                if (child.node == current) continue;
+                try self.pruned.put(self.arena, child.node, {});
```
arrufat: The prune set is built from `RenderTree.children`, which only yields nodes that render: `<script>`, `<style>`, metadata, `<svg>`, `[hidden]`, aria-hidden and unslotted siblings never enter it. Markdown never sees those anyway, but `dump._deep` walks the raw DOM and only skips nodes *in* the set. Clutter forces `shell` + `invisible` but not `js`/`css`, and `hasAuthorDisplayNone` excludes UA rules, so none of them are caught by a flag either.

Repro with `fetch --dump html --strip-mode clutter` on:

```html
<body>
<script>window.__DATA__ = {...}</script>
<div hidden>cookie banner</div>
<main><article>...600 chars...</article></main>
</body>
```

The HTML dump keeps the script and the hidden banner beside the article; the markdown dump of the same page drops them. Walking raw siblings makes the prune set mean the same thing to both renderers:

```suggestion
            var child = parent.firstChild();
            while (child) |c| : (child = c.nextSibling()) {
                if (c == current) continue;
                try self.pruned.put(self.arena, c, {});
```

Note this also prunes `<head>` (a sibling of `<body>` on the path), so the HTML dump loses `<title>`. If that matters, skip the `head` tag here or start the walk at `<body>`. The clutter tests all assert through `markdown.dump`, so an HTML-format case would be worth adding.

src/browser/dump.zig
```diff
@@ -29,15 +30,21 @@ pub const Opts = struct {
...
+
     /// Soft cap: output is cut at a UTF-8 boundary and a truncation marker
     /// appended.
     max_bytes: ?u32 = null,
 
-    pub const Strip = packed struct(u4) {
+    // Nodes to remove (clutter remove, from RenderTree.resolve)
+    pruned: ?*const RenderTree.PruneSet = null,
+
+    pub const Strip = packed struct(u6) {
```
arrufat, line 41: Is it intended that `--strip-mode full` now also means `shell` + `clutter`? The parser derives `full` from `maxInt` of the backing integer (cli.zig:621), so widening to `u6` changed its meaning: on an article page the HTML dump now collapses to the readability-selected subtree, and on a short page it logs `strip clutter fallback` and returns a shell-stripped page. The `full` line was dropped from help.zon, but the value is still accepted, cli.zig:102 still documents it as "sets every field", and the docs (`guides/markdown-axtree.mdx:96`, `reference/cli/fetch.mdx`) still recommend it as "remove all UI, scripts and styles".

If not intended, `full` could stay pinned to `js,ui,css,invisible`; if intended, the help line and docs should say so.

src/browser/clutter.zig
```diff
@@ -0,0 +1,932 @@
...
+        return buf;
+    }
+
+    /// Text under `node` once the prune set is applied.
+    fn selectedText(self: *const Pass, node: *Node) usize {
+        var n: usize = 0;
+        var it = self.tree.children(node, false);
+        while (it.next()) |child| {
+            switch (child.what) {
+                .text => |text| n += measureText(text).len,
```
arrufat: `withSiblings` puts the parent's loose text nodes into `pruned` (they won't render), but the gate only checks `pruned` for element children, and `tree.state.pruned` is null during the pass, so that text still counts toward `char_threshold`.

Repro: `<div id=wrap>[600 chars of bare text]<div class=body><p>120c</p><p>120c</p><p>120c</p></div><p>100c</p></div>`. Top candidate is `div.body`, `withSiblings` keeps the trailing `<p>` and returns `#wrap` with the 600 loose chars pruned. `selectedText(#wrap)` reports 1059, attempt 0 is accepted, and the rendered markdown is 460 chars, under the gate that should have sent it down the ladder.

```suggestion
                .text => |text| {
                    if (self.pruned.contains(child.node)) continue;
                    n += measureText(text).len;
                },
```

src/browser/clutter.zig
```diff
@@ -0,0 +1,932 @@
...
+                    .anchor => {
+                        inner.in_link = true;
+                        const href = el.getAttributeSafe(comptime .wrap("href")) orelse "";
+                        inner.hash_link = href.len > 0 and href[0] == '#';
+                    },
+                    .h1, .h2, .h3, .h4, .h5, .h6 => inner.in_heading = true,
+                    .ul, .ol => inner.in_list = true,
+                    .code, .pre => inner.in_code = true,
+                    .table => inner.in_table = true,
+                    .figure => inner.in_figure = true,
```
arrufat, line 298: `Stats.table_head` is only defaulted and OR-merged in `add`, never set to true, so `isDataTable`'s `st.th > 0 or st.table_head` is really just `st.th > 0`. A table whose header row uses `<td>` inside `<thead>` gets no data-table signal and can fall into the `.mix` rule.

Repro: a 3-row pricing table (`<thead><tr><td>Plan</td><td>Price</td></tr></thead>`, rows like `<td>Basic $10</td><td><a href=/buy>Buy now</a></td>`) inside an article. Link density ≈ 0.36 > 0.2 with weight 0, so the whole table is pruned (`rule = mix`); the identical table with `<th>` cells is kept. readability's `_markDataTables` marks it as data via `thead`.

Setting the flag on the descendants readability checks:

```suggestion
                    .figure => inner.in_figure = true,
                    // readability's _markDataTables descendants.
                    .caption, .col, .colgroup, .tfoot, .thead => st.table_head = true,
```

(`table_head` may deserve a broader name now that it covers more than `<thead>`.)

src/browser/dump.zig
```diff
@@ -412,6 +422,76 @@ pub fn shouldStripElement(el: *Node.Element, strip: Opts.Strip, frame: *Frame) b
...
+    while (it.next()) |token| {
+        for (shell_tokens) |shell_token| {
+            if (std.ascii.eqlIgnoreCase(token, shell_token)) return true;
+        }
+    }
+    return false;
+}
+fn hasSectioningAncestor(el: *Node.Element) bool {
+    var node = el.asNode().parentNode();
+    while (node) |n| : (node = n.parentNode()) {
```
arrufat: A ShadowRoot's `parentNode` is null, so a `<header>` in a shadow tree hosted under `<main>` is judged page-level and stripped, while its light-DOM sibling `<header>` is kept. Same hop `pruneBeside` does in clutter.zig:

```suggestion
    while (node) |n| : (node = n.parentNode() orelse blk: {
        // A shadow tree renders in place of its host.
        const shadow = n.is(Node.ShadowRoot) orelse break :blk null;
        break :blk shadow.getHost().asNode();
    }) {
```

Also: the class/id token rule applies to any tag, so `<tr class="header">` in a body-level table loses the header row under `shell`. Limit it to container tags?

src/browser/tools.zig
```diff
@@ -401,7 +401,7 @@ pub const Tool = enum {
...
+                    \\    "strip": { "type": "object", "description": "Optional. Omit element groups from the output: `js` (script, noscript, script preloads), `css` (style, stylesheet links), `ui` (css plus img, picture, video, audio, svg, canvas, iframe), `invisible` (elements an author rule or inline style sets to display:none), `shell` (nav, aside, dialog, page-level header/footer and the matching landmark roles; skipped when that would drop most of the text). {\"js\":true,\"css\":true} keeps a page dump small.", "properties": { "js": { "type": "boolean" }, "css": { "type": "boolean" }, "ui": { "type": "boolean" }, "invisible": { "type": "boolean" }, "shell": { "type": "boolean" } } },
```
arrufat: `clutter` is missing from this description and from `properties` (same on the screenshot schema below), yet `Params` parses `lp.dump.Opts.Strip`, so it is accepted but undiscoverable by a schema-driven client. protocol.json's `LP.dump` does list it.

src/browser/tools.zig
```diff
@@ -1314,7 +1315,7 @@ fn execMarkdown(arena: std.mem.Allocator, session: *lp.Session, registry: *NodeR
     const node = try resolveScope(session, registry, page, args.selector, args.backendNodeId);
 
     var aw: std.Io.Writer.Allocating = .init(arena);
-    lp.markdown.dump(node, .{ .max_bytes = args.maxBytes }, &aw.writer, page) catch return ToolError.InternalError;
+    lp.markdown.dump(.{ .root = node }, .{ .max_bytes = args.maxBytes }, &aw.writer, page) catch return ToolError.InternalError;
```
arrufat: The markdown tool has no `strip` and never calls `RenderTree.resolve`, while `parseValue` uses `ignore_unknown_fields = true`: a `markdown {strip:{shell:true}}` call is silently rendered unstripped and the inert arg gets recorded in the script. Since `shell`/`clutter` now change markdown output (help.zon's "strip does not affect markdown" line went away), should the primary reading tool take `strip` too?

src/browser/clutter.zig
```diff
@@ -0,0 +1,932 @@
...
+            },
+            else => if (self.tree.classify(self.root, .{})) |child| {
+                try self.walkChild(child, .{}, &root_stats);
+            },
+        }
+        self.total = root_stats.text;
+
+        const top = self.topCandidate() orelse return null;
+        const candidate = self.refine(top);
+        log.debug(.browser, "clutter candidate", .{ .top = describe(top.asNode()), .score = self.score(top), .refined = describe(candidate.asNode()) });
```
arrufat: `describe()` returns a slice of one threadlocal buffer and is called twice in this tuple, so `.top` prints (a torn copy of) the refined name. Splitting the line fixes it: on a page where `article` refines to `main` the log now shows `top = article`, `el = main` instead of `top = main.#e.#`.

```suggestion
        log.debug(.browser, "clutter candidate", .{ .top = describe(top.asNode()), .score = self.score(top) });
        log.debug(.browser, "clutter refined", .{ .el = describe(candidate.asNode()) });
```

src/browser/clutter.zig
```diff
@@ -0,0 +1,932 @@
...
+const char_threshold = 500;
+
+/// readability's retry ladder: each step drops one heuristic.
+const Flags = struct {
+    unlikelys: bool,
+    weight_classes: bool,
+    clean: bool,
+};
+
+const attempts = [_]Flags{
```
arrufat, line 54: Attempts 3 and 4 differ only in `clean`, which only `shouldPrune` reads, so attempt 4 repeats attempt 3's whole stats walk before diverging. Not a one-liner to dedupe (needs a pre-clean snapshot of `pruned`), just noting it.

src/browser/dump.zig
```diff
@@ -110,6 +117,9 @@ pub fn deep(node: *Node, opts: Opts, writer: *std.Io.Writer, frame: *Frame) erro
 fn _deep(node: *Node, opts: Opts, comptime force_slot: bool, writer: *std.Io.Writer, frame: *Frame) error{WriteFailed}!void {
```
arrufat, line 125: The resolve → copy `strip`/`pruned` into `Opts` → Document-or-node split is written three times (here, `tools.zig` execHtml, `lightpanda.zig` dumpContent) and the copies already differ in shape. markdown/screenshot/pdf take `State` whole; a helper lets HTML do the same:

```suggestion
/// The HTML of a resolved dump: the whole document when it is the root,
/// else the node's outerHTML. `opts.strip`/`opts.pruned` come from `state`.
pub fn render(state: RenderTree.State, opts: Opts, writer: *std.Io.Writer, frame: *Frame) !void {
    var o = opts;
    o.strip = state.strip;
    o.pruned = state.pruned;
    if (state.root.is(Node.Document)) |doc| return root(doc, o, writer, frame);
    return deep(state.root, o, writer, frame);
}

fn _deep(node: *Node, opts: Opts, comptime force_slot: bool, writer: *std.Io.Writer, frame: *Frame) error{WriteFailed}!void {
```

src/server/cdp/domains/lp.zig
```diff
@@ -158,6 +161,66 @@ fn getSemanticTree(cmd: anytype) !void {
...
+            var aw: std.Io.Writer.Allocating = .init(cmd.arena);
+            defer aw.deinit();
+            const opts: lp.dump.Opts = .{ .strip = state.strip, .pruned = state.pruned, .max_bytes = params.maxBytes };
+            if (params.format == .markdown) {
+                try markdown.dump(state, .{ .max_bytes = params.maxBytes }, &aw.writer, frame);
+            } else if (root.is(DOMNode.Document)) |doc| {
+                try lp.dump.root(doc, opts, &aw.writer, frame);
+            } else {
+                try lp.dump.deep(root, opts, &aw.writer, frame);
+            }
```
arrufat: With `dump.render` (see dump.zig):

```suggestion

    switch (params.format) {
        .html, .markdown => {
            var aw: std.Io.Writer.Allocating = .init(cmd.arena);
            defer aw.deinit();
            if (params.format == .markdown) {
                try markdown.dump(state, .{ .max_bytes = params.maxBytes }, &aw.writer, frame);
            } else {
                try lp.dump.render(state, .{ .max_bytes = params.maxBytes }, &aw.writer, frame);
            }
```

src/browser/tools.zig
```diff
@@ -1338,13 +1339,19 @@ fn execHtml(arena: std.mem.Allocator, session: *lp.Session, registry: *NodeRegis
...
     var aw: std.Io.Writer.Allocating = .init(arena);
-    if (args.selector == null and args.backendNodeId == null) {
-        lp.dump.root(page.document, opts, &aw.writer, page) catch return ToolError.InternalError;
+    if (state.root.is(DOMNode.Document)) |document| {
+        lp.dump.root(document, opts, &aw.writer, page) catch return ToolError.InternalError;
     } else {
-        const node = (try resolveTarget(session, registry, args.selector, args.backendNodeId)).node;
-        lp.dump.deep(node, opts, &aw.writer, page) catch return ToolError.InternalError;
+        lp.dump.deep(state.root, opts, &aw.writer, page) catch return ToolError.InternalError;
     }
```
arrufat: With `dump.render`, and `resolveScope` (15 lines up, used by markdown/screenshot) instead of the inline `whole` + `resolveTarget`:

```suggestion
    const target = try resolveScope(session, registry, page, args.selector, args.backendNodeId);
    const state = lp.RenderTree.resolve(arena, target, args.strip, page) catch return ToolError.OutOfMemory;
    var aw: std.Io.Writer.Allocating = .init(arena);
    lp.dump.render(state, .{ .max_bytes = args.maxBytes }, &aw.writer, page) catch return ToolError.InternalError;
```

src/lightpanda.zig
```diff
@@ -473,29 +478,30 @@ fn dumpRoot(frame: *Frame, selector: ?[]const u8) !*Node {
...
+    dump_opts.pruned = state.pruned;
     switch (mode) {
         .html => if (opts.selector == null)
-            try dump.root(frame.window._document, opts.dump, writer, frame)
+            try dump.root(frame.window._document, dump_opts, writer, frame)
         else
-            try dump.deep(root, opts.dump, writer, frame),
-        .markdown => try markdown.dump(root, .{ .max_bytes = opts.dump.max_bytes, .strip = opts.dump.strip }, writer, frame),
+            try dump.deep(state.root, dump_opts, writer, frame),
+        .markdown => try markdown.dump(state, .{ .max_bytes = dump_opts.max_bytes }, writer, frame),
```
arrufat: With `dump.render` (`dumpRoot` already returns the document when there is no selector, so `state.root` covers both cases):

```suggestion
    switch (mode) {
        .html => try dump.render(state, opts.dump, writer, frame),
        .markdown => try markdown.dump(state, .{ .max_bytes = opts.dump.max_bytes }, writer, frame),
```

src/browser/dump.zig
karlseguin: The intent is to deprecate --strip-mode full, which is why I removed it from the help. It makes these types of changes unnecessarily difficult to roll out - you either include these in --full, which might break some cases, or you don't in which case `--full` becomes not full.

Maybe I should just remove support for `--full` support from cli.zig?

PR #3440, opened 2026-09-08
fix(css): preserve custom properties in computed styles

### LINKED ISSUE #3199: css: getComputedStyle drops CSS custom properties from every source (stylesheet, inline, inherited)
## Summary

`getComputedStyle().getPropertyValue('--x')` returns the empty string for CSS custom properties declared in any source: stylesheet rules, inline `style=` declarations, and inherited values. After #2836, known CSS properties declared inline resolve correctly, but custom properties are excluded even on that path.

## Reproduction

```html
<!DOCTYPE html>
<html>
<head>
  <style>
    :root { --brand-color: #336699; }
    .card { --card-pad: 16px; }
  </style>
</head>
<body>
  <div class="card" id="card" style="--inline-var: 4px"></div>
  <script>
    const root = getComputedStyle(document.documentElement);
    const card = getComputedStyle(document.getElementById('card'));
    console.log('ROOT_VAR=[' + root.getPropertyValue('--brand-color') + ']');
    console.log('CLASS_VAR=[' + card.getPropertyValue('--card-pad') + ']');
    console.log('INLINE_VAR=[' + card.getPropertyValue('--inline-var') + ']');
    console.log('INHERITED_VAR=[' + card.getPropertyValue('--brand-color') + ']');
  </script>
</body>
</html>
```

Observed (nightly `1.0.0-nightly.8643+fb7123cf`, macOS aarch64, `lightpanda fetch <url> --log_level debug`):

```
ROOT_VAR=[]
CLASS_VAR=[]
INLINE_VAR=[]
INHERITED_VAR=[]
```

Chrome returns `#336699`, `16px`, `4px`, `#336699`.

## Why it matters

Design-system-driven SPAs commonly build their theme at boot by reading design tokens through `getComputedStyle` (MUI themes built from CSS tokens, Tailwind v4 CSS-first tokens, most in-house design systems). When every `--var` resolves to `""`, the theme factory throws, React unmounts the tree, and the page renders empty - even though the JS itself runs fine in Lightpanda. Resolving custom properties would unblock this whole class of apps.

## Proposed scope (following the direction in #2733)

#2733 deliberately left stylesheet-cascade resolution out of #2836, with a suggestion of making computation opt-in. Custom properties are a much smaller, self-contained slice of that problem:

- Per css-variables-1 §3.1 the computed value of a custom property is the declared token stream - no unit resolution, no layout, none of the resolved-value complexity that kept the full cascade out of scope.
- Resolution order: element inline style, then matching author rules by specificity/order, then inherit up the parent chain to `:root`.
- Could sit behind an opt-in consistent with `--enable-external-stylesheets` if the default-path cost is a concern.
- `var()` substitution inside returned longhand values could be a separate follow-up.

Happy to work on a PR for this if the direction sounds right - from an initial read of `StyleManager`/`CSSStyleProperties`, the rule parsing and selector matching this needs are already in place.

Tiancheng-Xu:
## Summary

Closes #3199.

- Preserve `--*` declarations from stylesheet rules in the StyleManager rule cache.
- Resolve custom properties from matching stylesheet rules, inline style, and inherited ancestors through `getComputedStyle()`.
- Tokenize double-hyphen custom property names as CSS identifiers.
- Keep the change scoped to custom property lookup; `var()` expansion and general computed-value resolution remain unchanged.

## Tests

- `TEST_VERBOSE=true TEST_FILTER="#element.html" make test`
- `TEST_VERBOSE=true make test F="WebApi: Element"`
- `TEST_VERBOSE=true make test F="WebApi: CSSStyleSheet"`
- `zig fmt --check src/browser/StyleManager.zig src/browser/webapi/css/CSSStyleDeclaration.zig src/browser/css/Tokenizer.zig`
- `git diff --check`

The repository-wide formatter also reports a pre-existing formatting difference in `zig-pkg/boringssl.../build.zig`, which is outside this change.

### DIFF
--- a/src/browser/StyleManager.zig
+++ b/src/browser/StyleManager.zig
@@ -49,6 +49,11 @@ const Tag = Element.Tag;
 const Input = Element.Html.Input;
 const RuleList = std.MultiArrayList(VisibilityRule);
 
+const CustomProperty = struct {
+    name: []const u8,
+    value: []const u8,
+};
+
 frame: *Frame,
 
 arena: *lp.Arena,
@@ -454,11 +459,24 @@ fn addRawRule(self: *StyleManager, build_arena: Allocator, selector_text: []cons
     if (selector_text.len == 0) return;
 
     var props = VisibilityProperties{};
+    var custom_properties: std.ArrayList(CustomProperty) = .empty;
     var it = CssParser.parseDeclarationsList(block_text);
     while (it.next()) |decl| {
         const name = decl.name;
         const val = decl.value;
-        if (std.ascii.eqlIgnoreCase(name, "display")) {
+        if (std.mem.startsWith(u8, name, "--")) {
+            var replaced = false;
+            for (custom_properties.items) |*custom| {
+                if (std.mem.eql(u8, custom.name, name)) {
+                    custom.value = val;
+                    replaced = true;
+                    break;
+                }
+            }
+            if (!replaced) {
+                try custom_properties.append(self.arena.allocator(), .{ .name = name, .value = val });
+            }
+        } else if (std.ascii.eqlIgnoreCase(name, "display")) {
             props.display = Display.parse(val);
         } else if (std.ascii.eqlIgnoreCase(name, "visibility")) {
             props.visibility_hidden = std.ascii.eqlIgnoreCase(val, "hidden") or std.ascii.eqlIgnoreCase(val, "collapse");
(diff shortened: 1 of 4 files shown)


karlseguin:
Thanks for the inspiration. I decided to take a different approach (https://github.com/lightpanda-io/browser/pull/3461). My main concern was with the overhead for what is essentially a very rare feature. Especially on a site like Tailwind where there could be thousands of these, most (if not all) never being needed.

PR #3448, opened 2026-09-08
Change how Lax allowance computed (more RFC 2625bis compliance)

nikneym:
Specifically to distinguish cross-site iframe navigation from top-level navigation, this PR reworks how `SameSite=Lax` moved. Since we're not checking if its a navigation alone now, the field for it is also renamed to `lax_allowed`.


nikneym:
Closing since fixed at #3335.

PR #3451, opened 2026-09-08
cli: suggest the closest flag or command on a typo

arrufat:
An unknown `--flag` now logs the nearest accepted spelling within two edits as `did_you_mean`, painted green next to the red typo in the pretty log format; logfmt stays plain:

```
FATAL app : unknown argument
      mode = fetch
      arg = --wait-mss
      did_you_mean = --wait-ms
```

The same hint covers two more cases:

- A bare first argument within two edits of a command name (`lightpanda agend`) is rejected instead of being fetched as a URL. Anything with a dot, a scheme or a leading `--` is parsed exactly as before.
- A misspelt value for an enum-typed flag (`--wait-until networkidl`) names the nearest tag. `--dump`'s peeking validator treats a word within two edits of a format name as that mistake rather than as the URL, so `--dump htmx` no longer turns into a second URL.

The Levenshtein helper moves from `SlashCommand` into `string.zig` so the agent REPL and the CLI share it, with the table widened to fit the longest flag name. `log.red`/`log.green` ride the existing `logFmt`/`format` hooks, so the `Value` model is untouched.

### DIFF
--- a/src/Config.zig
+++ b/src/Config.zig
@@ -21,6 +21,7 @@ const zenai = @import("zenai");
 const lp = @import("lightpanda");
 
 const cli = @import("cli.zig");
+const string = @import("string.zig");
 const dump = @import("browser/dump.zig");
 const Mime = @import("browser/Mime.zig");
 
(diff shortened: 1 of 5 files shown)


karlseguin:
fancies

PR #3455, opened 2026-09-08
Fix `Notification` use-after-free

mookums:
This should clear out the `*Notification` on owner-less Transfers when we de-init the Notification that they are referencing.


src/server/cdp/CDP.zig
```diff
@@ -615,6 +615,18 @@ pub const BrowserContext = struct {
         self.node_registry.deinit();
         self.node_search_list.deinit();
         self.set_child_nodes_sent.deinit(self.cdp.allocator);
+
+        // CorsGate/RobotsGate fetches are ownerless, so closeSession's owner-based teardown never reaches them.
```
karlseguin: The same bug exists in WebDriver, which makes me think this would be better in `Session.deinit`

PR #3463, opened 2026-09-09
`AdBlocker`: run /regex/ filters with PCRE2

arrufat:
Filter lists carry rules written as JavaScript regex literals (24 in EasyList, 165 in uBO's badware list, ~250 across the six common lists). They parsed but were dropped as unsupported. This runs them with PCRE2.

## Why PCRE2

The lists' regex syntax is JavaScript's: `\/` and `\-` identity escapes, `(?:` groups, lookahead, the odd backreference, and uBO compiles them with the `i` flag unless `$match-case`. What was needed: that syntax read as-is, a compiled pattern one `AdBlocker` can share across every HTTP client thread, a bound on backtracking (lists are third-party content), and Zig 0.16.

PCRE2 meets all of it. 10.48 ships a `build.zig` for 0.16, so it is wired like sqlite3 (`dep.artifact("pcre2-8")` + translate-c), no JIT. It compiles every one of the 252 regex rules across the six lists. Cost: ~0.8 MB on the release binary (+0.45%).

### Alternatives considered

Each was run against the same corpus (the 24 EasyList regexes, then all 252) and diffed against Python's `re` as a reference.

- **[zig-utils/zig-regex](https://github.com/zig-utils/zig-regex)** — v0.2.1 pins `minimum_zig_version` to 0.17-dev; v0.2.0 is the 0.16 release. It rejects the JavaScript identity escapes `\/` and `\-` (`InvalidEscapeSequence`), which fails 23 of EasyList's 24 rules outright; after rewriting `\/` it agreed with the reference on all 768 pattern×URL pairs, so the engine is sound but not JavaScript-compatible. Lookaround and backreferences route it to a backtracking engine that mutates shared state through `@constCast`, which rules out sharing one compiled pattern across threads. ~9 µs per match with heap allocations on every call.
- **[mnemnion/mvzr](https://github.com/mnemnion/mvzr)** — the closest pure-Zig fit: 0.16, zero-allocation, matching takes `*const` so it is shareable, ~0.8 µs per match, and it agreed with the reference on every pattern it compiled. It does not parse `(?:` groups, lookaround or backreferences, and has no case-insensitive mode: 3 of EasyList's 24 rules fail at its default op budget and 1 still fails at `SizedRegex(1024, 64)`; across the six lists 66 of 252 fail by default and 57 with the larger budget. Case-insensitivity would need a pattern-rewriting layer of our own (lowercasing literals without touching `\D`, `\W`, `\S`, `\B`), and there is no backtracking limit, which its author flags for untrusted patterns. It would be the pick if a C dependency were unacceptable.
- **V8's `RegExp`** — already in the binary, but tied to an isolate; the adblocker runs on the HTTP client path from several threads, none of which can enter one.
- **A hand-written matcher** for the subset EasyList uses today — would have been ~300 lines, but the uBO lists already use constructs beyond that subset, and every gap is a silent false negative.

## What changes

- `Regex.zig` wraps compile/match/free. `Regex.Context` routes PCRE2's allocations through the blocker's allocator so a compiled pattern nobody freed fails a test like any other leak, and sets match/depth limits so a broken pattern costs a false negative rather than a stalled request. Unknown alphanumeric escapes are literals, as in non-unicode `RegExp`.
- As in uBO, a regex tests the raw URL with the case-insensitive flag unless `$match-case`; `pattern.Url` now carries the raw text next to the lowercased one and `Request.init` does the lowercasing itself.
- Regex filters are indexed by the literals every match carries, a port of uBO's `tokenizableStrFromRegex` (`RegexShape` in `Engine.zig`): the pattern is flattened into a plain pattern where literal token characters stay and anything uncertain becomes `*`, then read by the same bounded-token loop. Anything the flattening does not follow yields no token, which is never wrong. One departure from uBO: positive lookaround is not inlined, so a token never comes from text the regex does not consume.
- `Regex.matches` runs on a stack-first allocator: PCRE2 wants a match data block and 20KB of backtracking frames per call, which no longer touches the heap in the common case.
- Patterns PCRE2 rejects are counted in `rules_skipped` and logged at debug level with PCRE2's message and offset.

## Numbers

| List | before | after |
|---|---|---|
| EasyList, its 24 regex rules alone | 0 loaded / 24 skipped | 24 loaded / 0 skipped |
| EasyList | 27 skipped | 3 skipped |
| The 79 regex rules the parser accepts across six lists | all in the fallback bucket | 72 bucketed, 7 fallback |

PCRE2 rejects none of the 252 regex rules across those lists; the remaining skips are option-level (`$strict1p`, `$removeparam`, ...).

### DIFF
--- a/src/network/adblock/Engine.zig
+++ b/src/network/adblock/Engine.zig
@@ -66,6 +66,9 @@ pub fn deinit(self: *Engine, allocator: Allocator) void {
 /// request and shared by all three engines.
 pub const Request = struct {
     url: pattern.Url,
+    /// The URL as requested, fragment stripped but case kept: what a regex
+    /// filter reads, since `$match-case` is only meaningful on the original.
+    raw: []const u8,
     /// The hostname of the document the request belongs to. Falls back to the
     /// request's own hostname, which is what uBO does for top-level loads.
     source_hostname: []const u8,
@@ -95,9 +98,11 @@ pub const Request = struct {
         source: [SOURCE_MAX]u8,
     };
 
-    /// `url` must already be lowercased and fragment-free; `source_hostname`
-    /// may be empty when there is no document context.
+    /// `raw` and `url` are the same fragment-free URL, the second one
+    /// lowercased; `source_hostname` may be empty when there is no document
+    /// context.
     pub fn init(
+        raw: []const u8,
         url: []const u8,
         source_hostname: []const u8,
         kind: NetworkFilter.ResourceTypes,
@@ -106,6 +111,7 @@ pub const Request = struct {
         const source = if (source_hostname.len == 0) parsed.hostname() else source_hostname;
         var request: Request = .{
             .url = parsed,
+            .raw = raw,
             .source_hostname = source,
             .kind = kind,
             .third_party = domain.isThirdParty(parsed.hostname(), source),
@@ -125,7 +131,10 @@ pub const Request = struct {
 
     pub fn fromHttp(transfer: *const HttpClient.Transfer, buffers: *Buffers) ?Request {
         const req = &transfer.req;
-        const url = normalizeUrl(req.url, &buffers.url) orelse return null;
+        // No request URL carries a fragment onto the wire.
+        const fragment = std.mem.indexOfScalar(u8, req.url, '#') orelse req.url.len;
+        const raw = req.url[0..fragment];
+        const url = normalizeUrl(raw, &buffers.url) orelse return null;
 
         var owner: ?*const HttpClient.Owner = transfer.owner;
         var subframe = false;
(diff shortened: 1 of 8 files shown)


src/network/adblock/Engine.zig
```diff
@@ -340,22 +348,304 @@ fn collectTokens(filter: *const NetworkFilter, buf: []u32) []u32 {
...
+        const start = self.n;
+        var branches: usize = 1;
+        var branch_start = start;
+        var first = false;
+        var last = false;
+        while (true) {
+            try self.sequence(nested);
+            const branch = self.out[branch_start..self.n];
+            first = first or startsTokenish(branch);
+            last = last or endsTokenish(branch);
```
nikneym, line 419: There's a chance that `not_token_optional` can appear sequentially, considering this expression:
```
/\/ads(\/?x|\/y)/$script
```

```
https://example.com/adsx -> expected `.blocked`, received `.none`
```

producing `[0x02, 0x02, 'x']` right at `\/?x`.

src/network/adblock/Regex.zig
```diff
@@ -0,0 +1,225 @@
...
+        pattern.ptr,
+        pattern.len,
+        options,
+        &err_code,
+        &err_offset,
+        context.compile_context,
+    ) orelse {
+        // Compile errors are positive codes; 21 is the one for a failed
+        // allocation, and that is ours, not the pattern's.
+        if (err_code == 21) return error.OutOfMemory;
```
nikneym: I believe this should check if `err_code` is  `121`.
https://github.com/PCRE2Project/pcre2/blob/aac57f978e38fb4a04899d623b68e0fbb5bcaf6c/src/pcre2.h.in#L246

src/network/adblock/Engine.zig
arrufat: Good catch, thanks. The alternation read the first marker of a branch as a definite non-token, and an optional stretch there can be absent. `startsTokenish`/`endsTokenish` now look through optional non-token markers, so this rule gets no token and rides the fallback bucket. Added your case as a test at both the token and the verdict level.

src/network/adblock/Regex.zig
arrufat: Right, and it has a public name. Now `pcre2.PCRE2_ERROR_HEAP_FAILED`, no magic number.

nikneym:
LGTM

nikneym (approved):

PR #3465, opened 2026-09-09
Fix `history.state` and `Navigation.currentEntry` crash

mookums:
When you create a CDP target, you can have an empty Navigation list. This hits the assert in `Navigation.currentEntry` and `history.state` and causes a crash. This fixes it by pushing an Entry to Navigation when a Target is created.

### DIFF
--- a/src/server/cdp/domains/target.zig
+++ b/src/server/cdp/domains/target.zig
@@ -221,13 +221,18 @@ fn createTarget(cmd: *CDP.Command) !void {
         try doAttachtoTarget(cmd, target_id);
     }
 
-    if (!std.mem.eql(u8, "about:blank", params.url)) {
-        const encoded_url = try URL.resolveNavigation(frame.call_arena, params.url, .{});
-        try frame.navigate(
-            encoded_url,
-            .{ .reason = .address_bar, .kind = .{ .push = null } },
-        );
-    }
+    // https://html.spec.whatwg.org/multipage/document-sequences.html,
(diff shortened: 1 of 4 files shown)


src/server/cdp/domains/target.zig
```diff
@@ -221,6 +221,17 @@ fn createTarget(cmd: *CDP.Command) !void {
         try doAttachtoTarget(cmd, target_id);
     }
 
+    // https://html.spec.whatwg.org/multipage/document-sequences.html --
+    // Creating a new browsing context always produces an initial about:blank
+    // Document with its own session history entry, even before any real
+    // navigation happens.
+    _ = try frame._session.navigation.pushEntry(
```
karlseguin: Three things.

1. Should this be moved to `createPage` because it seems like this would be the same issue for any one that calls createPage. Seems like the same crash exists in all places where this is called

2. Claude thinks this default entry should be replaced by the first real navigation

3. `currentEntry` CAN return null: "Thrown if the Navigation.currentEntryis null, i.e., there is no current history entry. This could occur for example if the current page is about:blank." "https://developer.mozilla.org/en-US/docs/Web/API/Navigation/updateCurrentEntry#exceptions"

So I'm not sure if this should be set on all new pages and replaced, or just returning null.

karlseguin: WPT's `navigation-api/navigation-activation/activation-initial-about-blank.html` suggest it's a set + replace scenario.

PR #3466, opened 2026-09-09
cli: add --locale and --timezone, derive language signals from one value

arrufat:
`navigator.language` was hard-coded to `en-US` and `Accept-Language` was a constant, while Intl and Date followed the host environment. On a German host a page saw `navigator.language === "en-US"` next to German number formatting, and the same page rendered differently across machines.

One configured tag now drives `navigator.language(s)`, the `Accept-Language` header and ICU's default locale, the way Chrome's `--lang` does.

- `--locale <BCP47>`, default `en-US`. The header is Chrome-shaped (`de-DE,de;q=0.9,en;q=0.8`) and `navigator.languages` lists its tags in order.
- `--timezone <IANA>`. Absent, the host zone is kept.
- Both are applied by setting `LC_ALL` and `TZ` before V8 initializes ICU, which reads them lazily. No fork change.
- CDP `Emulation.setUserAgentOverride.acceptLanguage`, what Playwright sends for its `locale` option, overrides header and `navigator.languages` for the browser context's lifetime.

Behaviour change: Intl is now `en-US` on every host by default instead of following `LANG`, so it matches `navigator.language` and tests no longer depend on the host locale.

`Emulation.setLocaleOverride` / `setTimezoneOverride` stay no-ops; changing ICU's defaults at runtime needs zig-v8-fork bindings.

### DIFF
--- a/src/Config.zig
+++ b/src/Config.zig
@@ -276,6 +276,8 @@ const CommonOptions = .{
     .{ .name = "web_bot_auth_keyid", .type = ?[]const u8 },
     .{ .name = "web_bot_auth_domain", .type = ?[]const u8 },
     .{ .name = "user_agent", .type = ?[]const u8, .validator = userAgentValidator },
+    .{ .name = "locale", .type = []const u8, .default = HttpHeaders.default_locale, .validator = localeValidator },
+    .{ .name = "timezone", .type = ?[]const u8, .validator = timezoneValidator },
     .{ .name = "block_private_networks", .type = bool },
     .{ .name = "block_cidrs", .type = ?[]const u8, .validator = accumulateValidator },
     .{ .name = "block_urls", .type = ?[]const u8, .validator = accumulateValidator },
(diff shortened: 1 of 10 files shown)


src/Config.zig
```diff
@@ -1007,6 +1064,25 @@ pub const HttpHeaders = struct {
         if (self.user_agent.ptr != user_agent_base.ptr) {
             allocator.free(self.user_agent);
         }
+        self.accept_language.deinit(allocator);
+    }
+
+    /// Chrome's shape: the tag, then its language alone, then English as a
+    /// last resort, with descending q values. `buf` must hold the longest
+    /// output for a tag that passed validateLocale (35 + 24 bytes).
+    pub fn acceptLanguageFor(buf: *[64]u8, tag: []const u8) []const u8 {
```
```suggestion
    fn acceptLanguageFor(buf: *[64]u8, tag: []const u8) []const u8 {
```

PR #3474, opened 2026-09-09
Probe own visibility inside walks that prune hidden subtrees

arrufat:
Stacked on #3469 (which stacks on #3468 and #3467).

Two walks probed the full ancestor chain per node although they never enter a hidden subtree, so below their root only the element's own display can change the answer:

- **`elementFromPoint`** (`Document.zig`): the preorder walk probed the chain for every node, O(N × depth) memo lookups per call. The hidden verdict now travels down the stack. Hidden subtrees are still counted towards the preorder index so hit positions keep agreeing with `getBoundingClientRect`.
- **AX tree writer** (`AXNode.zig`): every child was probed three times — once to prune in `writeNodeChildren`, once for `ignored` in `writeNode`, once more by the parent's `childIds` — and each probe re-read `aria-hidden` / `hidden` / `inert` and walked the chain. The prune and `childIds` now use an own-only probe (`isHiddenSelf`), and `writeNode` receives the verdict instead of recomputing it. The walk root and the query walk (`emitMatch`, which visits hidden subtrees by spec) keep the chain probe, as does the label-promotion target, which is not on the walk.

`StyleManager.isHiddenSelf(el, options)` is the new own-only probe with the same options as `isHidden`.

## Numbers

ReleaseFast builds of #3469 and this branch on the static theverge fixture (2.7k elements, depth 23).

| | #3469 | this PR |
|---|---|---|
| 50 × `document.elementFromPoint` spread down the page | 18.5 ms | 5.5 ms |
| `Accessibility.getFullAXTree` over CDP (2,762 nodes, warm) | 7.3 ms | 7.2 ms |

`elementFromPoint` is the payoff. The AX change is within noise: with the memo, a chain probe is a handful of hash hits and the writer's time goes to names and JSON. It stays in the PR because it removes redundant work and the verdict plumbing is what lets `writeNode` and `childIds` agree by construction, but it is not a measurable speed-up on its own.

Not included: `classifyInteractivity`'s `pointer-events` chain probe. `SemanticTree.walk` calls it without pruning on pointer-events, so an own-only probe there would change which descendants of a `pointer-events: none` element count as interactive.

One behaviour change: `getPartialAXTree` on a node inside a hidden subtree now emits that root with an empty `childIds` and no descendants, matching Chromium. Everything else is covered by the existing AX and `elementFromPoint` fixtures.


src/server/cdp/AXNode.zig
```diff
@@ -84,7 +84,8 @@ pub const Writer = struct {
             try self.walkQuery(self.root.dom, false, w);
         } else {
             const root = AXNode.fromNode(self.root.dom);
-            if (try self.writeNode(self.root.id, root, false, w)) {
+            const root_hidden = if (self.root.dom.is(DOMNode.Element)) |el| isHidden(el, self.frame, .{}) else false;
+            if (try self.writeNode(self.root.id, root, false, root_hidden, w)) {
```
karlseguin, line 88: `writeNode` has:

```zig
const write_children = axn.ignoreChildren() == false;
```

which both controls how `writeNode` behaves and it what returns. But it ignores this new `root_hidden` flag. Should that line be changed to:

```zig
const write_children = axn.ignoreChildren() == false and hidden == false;
```

?

arrufat: Good catch, fixed in 91cdb52 and added a test for a hidden root.

Note this only affects `getPartialAXTree` on a node inside a hidden subtree, since `hidden` is only ever true for the walk root. It is a small behaviour change vs main (the hidden root used to list its children); I've updated the PR description accordingly.

PR #3479, opened 2026-09-10
cdp: wheel events scroll the viewport unless over a scroll container

arrufat:
`Input.dispatchMouseEvent` with `type: mouseWheel` (and the BiDi wheel action, which shares `triggerMouseWheel`) only ever moved the scroll position of the element under the cursor, and returned early when no element was there. `window.scrollY` reads a separate field that `Window.scrollTo` maintains, so a wheel never moved the page and never fired the document's `scroll` event. Every CDP driver's scroll-by-wheel API was a no-op: Playwright `mouse.wheel`, Puppeteer `mouse.wheel`, chromedp `Input.dispatchMouseEvent`, Selenium BiDi wheel actions, browser-use `scroll()`, Stagehand v3 `mouse.wheel`.

Now the nearest ancestor whose `overflow` is `auto` or `scroll` is scrolled as before (the `#scrollbox` case), and anything else scrolls the window through `Window.scrollBy`, which already schedules the trusted `scroll` and `scrollend` events. A wheel over empty space targets the root element instead of returning early.

Adds `cdp.input` tests for the scroll-container case, viewport scrolling with accumulation and a negative delta, and a wheel over empty space.

Evidence from the integrations matrix (https://github.com/lightpanda-io/integrations): with this branch, the `scroll` cell goes from ❌ to ✅ for Playwright, Puppeteer, chromedp, Selenium (BiDi), Browser Use CLI and Stagehand v3, with no other cell changing.


src/browser/frame/user_input.zig
```diff
@@ -301,15 +303,48 @@ pub fn triggerMouseWheel(frame: *Frame, x: f64, y: f64, delta_x: f64, delta_y: f
...
+        }
+        if (isScrollContainer(el, frame)) {
+            return el;
+        }
+    }
+    return null;
+}
+
+fn isScrollContainer(el: *Element, frame: *Frame) bool {
+    const style = frame.window.getComputedStyle(el, null, frame) catch return false;
```
karlseguin: This will always return a `CSSStyleProperties` that has an empty `_properties`.  But it'll cache the value inside of the Frame's `_element_computed_styles` and have extra ceremony to eventually just call `style_manager.inlineStyleValue`. I think you just just call `style_manager.inlineStyleValue` directly here.

src/browser/frame/user_input.zig
```diff
@@ -301,15 +303,48 @@ pub fn triggerMouseWheel(frame: *Frame, x: f64, y: f64, delta_x: f64, delta_y: f
...
-    const new_left: i32 = @as(i32, @intCast(target.getScrollLeft(frame))) +| deltaToScroll(delta_x);
-    const new_top: i32 = @as(i32, @intCast(target.getScrollTop(frame))) +| deltaToScroll(delta_y);
-    try target.setScrollLeft(new_left, frame);
-    try target.setScrollTop(new_top, frame);
+    const dx = deltaToScroll(delta_x);
+    const dy = deltaToScroll(delta_y);
+    if (scrollContainerOf(target, frame)) |container| {
+        const new_left: i32 = @as(i32, @intCast(container.getScrollLeft(frame))) +| dx;
+        const new_top: i32 = @as(i32, @intCast(container.getScrollTop(frame))) +| dy;
+        try container.setScrollLeft(new_left, frame);
```
karlseguin: This is an issue in `main` also, but `setScrollTop` fires events (via the `scheduleScrollEvents`) and it'll have a different bubbles flag in most cases...and then we fire `scroll` a few lines down immediately.

src/browser/frame/user_input.zig
```diff
@@ -301,15 +303,48 @@ pub fn triggerMouseWheel(frame: *Frame, x: f64, y: f64, delta_x: f64, delta_y: f
...
-    // Apply the scroll and fire a trusted scroll event, mirroring WebDriver wheel.
+    // Scroll the nearest scroll container under the cursor, else the viewport.
     // CDP deltas are untrusted, so guard NaN and saturate the addition.
-    const new_left: i32 = @as(i32, @intCast(target.getScrollLeft(frame))) +| deltaToScroll(delta_x);
-    const new_top: i32 = @as(i32, @intCast(target.getScrollTop(frame))) +| deltaToScroll(delta_y);
-    try target.setScrollLeft(new_left, frame);
-    try target.setScrollTop(new_top, frame);
+    const dx = deltaToScroll(delta_x);
+    const dy = deltaToScroll(delta_y);
+    if (scrollContainerOf(target, frame)) |container| {
```
karlseguin: This should be axis-aware. If you're scrolling the x and find a overflow-y: auto, then it isn't scrollable.

karlseguin:
In addition to the inline comments, github says it can be merged as-is, but it can't. I tried, compiler errors when applied to `main`.

arrufat:
Merged `main` in. The compile error was `Window.scrollBy` going private in the declarations sweep (2e6999f); it is `pub` again since the wheel path now calls it. While in there, both `scrollBy` implementations use a saturating add, so an oversized delta from CDP or a script no longer overflows the i32 position (the wheel path used to clamp before writing, and now relies on `scrollBy`).

src/browser/frame/user_input.zig
arrufat: Done. It now reads `style_manager.inlineStyleValue` directly: the axis longhand (`overflow-x`/`overflow-y`) first, then the matching half of the `overflow` shorthand. The comment on `isScrollContainer` notes that only inline overflow is resolved, since that is all computed styles see today.

src/browser/frame/user_input.zig
arrufat: Done. The container path goes through `Element.scrollBy` and the viewport path through `Window.scrollBy`, and the inline dispatch is gone, so only the scheduled `scroll`/`scrollend` fire, with the bubbles flag those tasks already pick. The scroll-container test waits for the scheduled event with `runner.waitForScript` instead of asserting it synchronously.

src/browser/frame/user_input.zig
arrufat: Done. Each axis resolves its own container: a delta along an axis whose overflow on the box is not `auto`/`scroll` falls through to the viewport. Added a case to the container test with `overflow: hidden scroll` and a diagonal wheel, where the y delta lands on the box and the x delta on the window.

PR #3483, opened 2026-09-10
Cors Store

mookums:
This provides a CorsStore that acts as a preflight cache.

This makes the `preflight-cache.htm` and `cors-preflight-cache.any.html` pass, increasing our coverage to 495/813 (60.9%).


src/network/CorsStore.zig
```diff
@@ -0,0 +1,348 @@
...
+    const entry = self.map.get(key) orelse return null;
+
+    if (entry.expires_at <= lp.datetime.milliTimestamp(.real)) {
+        const kv = self.map.fetchRemove(key).?;
+        kv.key.deinit(self.allocator);
+        kv.value.deinit(self.allocator);
+        return null;
+    }
+
+    return entry;
```
krichprollsch, line 169: You return the shared entry which can be free while reading by a concurrent put.

src/network/CorsStore.zig
```diff
@@ -0,0 +1,348 @@
...
+    entry: Entry,
+    method: http.Method,
+    wants_credentials: bool,
+    authored_headers: []const []const u8,
+) bool {
+    if (wants_credentials and !entry.credentials) {
+        return false;
+    }
+
+    if (!entry.methods_wildcard and !entry.methods.contains(method)) {
```
krichprollsch, line 210: Claude suggests "to skip the method check when isSafelistedMethod(method)."

src/network/CorsStore.zig
```diff
@@ -0,0 +1,348 @@
...
+    pub fn hash(_: KeyContext, key: Key) u64 {
+        var hasher = std.hash.Wyhash.init(0);
+        hasher.update(key.origin);
+        hasher.update(&.{0});
+        hasher.update(key.target);
+        return hasher.final();
+    }
+
+    pub fn eql(_: KeyContext, a: Key, b: Key) bool {
+        return std.ascii.eqlIgnoreCase(a.origin, b.origin) and std.ascii.eqlIgnoreCase(a.target, b.target);
```
krichprollsch, line 53: it seems `hash` is key sensitive but `eql` is not...

krichprollsch: You should use case sensitive comparison here.

src/network/CorsStore.zig
```diff
@@ -0,0 +1,348 @@
...
+            }
+            try list.append(allocator, try allocator.dupe(u8, s));
+        }
+
+        return list.toOwnedSlice(allocator);
+    }
+
+    fn merge(self: Entry, allocator: std.mem.Allocator, new: Entry) !Entry {
+        return .{
+            .credentials = self.credentials or new.credentials,
```
krichprollsch, line 97: Claude thinks there is a problem here: we have conflict between req. w/ and w/o credentials.
It suggest to store two entries in one, something like this to keep them separated:

```
const Entry = struct {
    plain: ?Grant = null, // preflight made without credentials
    creds: ?Grant = null, // preflight made with credentials
};
```

krichprollsch (commented):
We never really cleanup the store isn't it? only get remove out of date cache entries.
So the risk is the cache grows forever. Maybe we should have a leaky bucket and cleaning the expired then older value on `put`.

Also Claude points another issue:

> put and get use different origins
>
> put uses ctx.origin = transfer.req.origin orelse "null" (CorsGate.zig:546). get uses transfer.effectiveOrigin(), which is "null" after a tainting redirect.
> 
> This comes from an older bug: the preflight sends req.origin instead of the effective origin. But with the cache, a preflight done for a tainted request is now stored under the real origin and reused by untainted requests. Using effectiveOrigin() in fetchThenResume would fix both.

karlseguin:
> We never really cleanup the store isn't it?

RobotStore is the same

src/network/CorsGate.zig
krichprollsch, line 544: Should'nt we use `transfer.effectiveOrigin()` here instead?

src/network/CorsStore.zig
```diff
@@ -0,0 +1,384 @@
...
+        return list.toOwnedSlice(allocator);
+    }
+
+    fn merge(self: Entry, allocator: std.mem.Allocator, new: Entry) !Entry {
+        return .{
+            .methods_wildcard = self.methods_wildcard or new.methods_wildcard,
+            .methods = self.methods.unionWith(new.methods),
+            .headers_wildcard = self.headers_wildcard or new.headers_wildcard,
+            .headers = try unionHeaders(allocator, self.headers, new.headers),
+            .expires_at = @max(self.expires_at, new.expires_at),
```
krichprollsch, line 106: Isn't using the minimal expiration more accurate here?
Since we don't store each expiration date, I think it's safer than using max.

src/network/CorsStore.zig
```diff
@@ -0,0 +1,384 @@
...
+    defer self.mutex.unlock(lp.io);
+
+    const gop = try self.map.getOrPut(self.allocator, key);
+    if (!gop.found_existing) {
+        errdefer _ = self.map.remove(key);
+        gop.key_ptr.* = try key.dupe(self.allocator);
+        gop.value_ptr.* = try entry.dupe(self.allocator);
+        return;
+    }
+
```
krichprollsch, line 194: You should check if the value is expired to clean it up before merging.

src/network/CorsGate.zig
```diff
@@ -212,6 +217,27 @@ pub fn check(self: *CorsGate, transfer: *Transfer) !Result {
         return .allowed;
     }
 
+    const wants_credentials = req.credentials_mode == .include;
+
+    if (try self.network.cors_store.get(.{
+        .origin = origin,
+        .target = req.url,
+        .credentials = wants_credentials,
```
krichprollsch, line 225: It seems non-creds requests are covered by creds ones.
So if non-creds cache is missing, we can fallback to the creds one.

Morover Claude suggests to merge `get` and `covers` , it avoids entry duplication then.

`CorsStore`

```zig
// unchanged logic, just no longer public
fn matches(entry: Entry, method: http.Method, authored_headers: []const []const u8) bool {
    if (!isSafelistedMethod(method) and !entry.methods_wildcard and !entry.methods.contains(method)) {
        return false;
    }
    for (authored_headers) |name| {
        const is_authorization = std.ascii.eqlIgnoreCase(name, "authorization");
        if (entry.headers_wildcard and !is_authorization) continue;
        var found = false;
        for (entry.headers) |allowed| {
            if (std.ascii.eqlIgnoreCase(allowed, name)) { found = true; break; }
        }
        if (!found) return false;
    }
    return true;
}

pub fn covers(self: *CorsStore, key: Key, method: http.Method, authored_headers: []const []const u8) bool {
    self.mutex.lockUncancelable(lp.io);
    defer self.mutex.unlock(lp.io);

    const entry = self.map.get(key) orelse return false;

    if (entry.expires_at <= lp.datetime.milliTimestamp(.real)) {
        const kv = self.map.fetchRemove(key).?;
        kv.key.deinit(self.allocator);
        kv.value.deinit(self.allocator);
        return false;
    }

    return matches(entry, method, authored_headers);
}
```

`CorsGet`

```zig
  const wants_credentials = req.credentials_mode == .include;
  const authored = try collectAuthoredHeaders(transfer, transfer.arena.allocator());

  var covered = store.covers(.{ .origin = origin, .target = req.url, .credentials = true }, req.method, authored.items);
  if (!covered and !wants_credentials) {
      covered = store.covers(.{ .origin = origin, .target = req.url, .credentials = false }, req.method, authored.items);
  }
```

result
```diff
@@ -0,0 +1 @@
+/nix/store/9487wkgckgjqypvf0p0v62sjrf635281-docker-image-wpt-server.tar.gz
```
krichprollsch: Should be removed isn't it?

lp-net-cache/cache.db
krichprollsch, line 1: Should be removed?

src/network/CorsGate.zig
```diff
@@ -355,6 +396,55 @@ const CorsPreflightContext = struct {
...
+                while (it.next()) |raw| {
+                    const token = std.mem.trim(u8, raw, &std.ascii.whitespace);
+                    if (std.meta.stringToEnum(http.Method, token)) |m| methods.insert(m);
+                }
+            }
+        }
+
+        const headers_wildcard = acah != null and std.mem.eql(u8, acah.?, "*") and !self.wants_credentials;
+        var owned_headers: []const []const u8 = &.{};
+        if (!headers_wildcard and self.request_headers.len > 0) {
```
krichprollsch, line 428: Maybe you could parse `acah` to add directly all the returned headers in case the server responds directly with a list.  But it's a nice to have.

src/network/CorsGate.zig
```diff
@@ -355,6 +396,55 @@ const CorsPreflightContext = struct {
         return true;
     }
 
+    fn cacheGrant(self: *CorsPreflightContext, acam: ?[]const u8, acah: ?[]const u8, acma: ?[]const u8) !void {
+        if (self.url.len == 0) return;
+
+        const max_age_s: u64 = blk: {
+            const v = acma orelse break :blk 5;
+            if (v.len == 0) break :blk 5;
+            break :blk std.fmt.parseUnsigned(u64, v, 10) catch return;
```
krichprollsch, line 405: unparsable max age should also `break :blk 5` instead of returning.

src/network/CorsGate.zig
krichprollsch: Should'nt we use `transfer.effectiveOrigin()` here instead?

krichprollsch, line 628: Should'nt we use `transfer.effectiveOrigin()` here instead?

src/network/CorsStore.zig
```diff
@@ -0,0 +1,384 @@
...
+    const old = gop.value_ptr.*;
+    const merged = old.merge(self.allocator, entry) catch |err| {
+        return err;
```
krichprollsch, line 197: what about `try old.merge(...` instead?

src/network/CorsGate.zig
mookums: If we change the return to `break :blk 5`, we end up failing `/cors/preflight-cache.htm`, specifically with

```
 Fail    "age = -1, should not be cached"
                "|assert_equals: did preflight expected \"1\" but got \"0\""
```

krichprollsch (approved):
I'm good with this version, @karlseguin if you want to take another look.

src/help.zon
```diff
@@ -505,6 +505,9 @@
     \\  --robot-store-entry-limit <INT>
     \\          Maximum number of entries kept in the RobotStore. 0 means no limit.
     \\          Defaults to 1000.
+    \\  --cors-store-entry-limit <INT>
```
karlseguin: The commands are alphabetically sorted (though I see `--robot-store-entry-limit` is also wrong)

src/network/CorsStore.zig
```diff
@@ -0,0 +1,427 @@
...
+            }
+        }
+        if (!found) return false;
+    }
+    return true;
+}
+
+/// Whether a cached grant for `key` covers this method/headers combination.
+/// A miss (expired or absent entry) is treated as not covered.
+pub fn covers(
```
karlseguin: This doesn't need to be public and it's only called from the tests. We normally put test helpers with the tests..but can the `coversRequest` not be used in tests?

src/network/CorsGate.zig
```diff
@@ -355,6 +400,45 @@ const CorsPreflightContext = struct {
...
+            }
+        }
+
+        const headers_wildcard = acah != null and std.mem.eql(u8, acah.?, "*") and !self.wants_credentials;
+
+        try self.gate.network.cors_store.put(
+            .{ .origin = self.origin, .target = self.url, .credentials = self.wants_credentials },
+            .{
+                .methods_wildcard = methods_wildcard,
+                .methods = methods,
```
karlseguin: So we'll cache the [valid] ACAM values from the *response*, but only the ACAH values from the *request*? Is there a reason for that?

src/network/CorsGate.zig
```diff
@@ -223,6 +250,21 @@ pub fn check(self: *CorsGate, transfer: *Transfer) !Result {
     return .pending;
 }
 
+fn collectAuthoredHeaders(transfer: *Transfer, allocator: std.mem.Allocator) !std.ArrayList([]const u8) {
```
karlseguin: `check` calls this method, and then `check` calls `fetchThenResume` which repeats this exact step.  `fetchThenResume` should take `authored.items`

PR #3484, opened 2026-09-10
XHR changes for CORS WPT tests

mookums:
1. XHR `getResponse` is able to actually return things if type is default or text even if it isn't done. This feature is used by the `/cors/status-preflight.htm` test.
2. XHR `parseMethod`should be able to parse and return `HEAD`. This feature is used by the `/cors/status-preflight-htm`.

This helps us pass 12/16 (previously 0/16) of the `/cors/status-preflight.htm` tests. We fail three because of detectAuthChallenge (I think) and one before our `http.Method` is a closed set and it tries to pass `CHICKEN` as the method.


src/browser/webapi/net/XMLHttpRequest.zig
```diff
@@ -518,7 +518,10 @@ pub fn getResponseURL(self: *XMLHttpRequest) []const u8 {
 }
 
 pub fn getResponse(self: *XMLHttpRequest, exec: *const Execution) !?Response {
-    if (self._ready_state != .done) {
+    // https://xhr.spec.whatwg.org/#the-response-attribute
+    // If the response type is default or text, we don't need it to be done to return things.
```
karlseguin: I don't think that's correct.

The response is cached in self._response and, just a few lines down, that's what's returned if it isn't null. So, if someone call `getResponse()` in `readystatechange` callback, they'll lock the initial (empty) value.

src/browser/webapi/net/XMLHttpRequest.zig
```diff
@@ -518,7 +518,10 @@ pub fn getResponseURL(self: *XMLHttpRequest) []const u8 {
...
+    const needs_done = self._response_type != .default and self._response_type != .text;
```
karlseguin: The responseType can be changed in some states. But not `.done`, so the cached `_response` was always correct for the final response_type. Now, we can end up with a stale cached response, if the type is changed after.

src/browser/webapi/net/XMLHttpRequest.zig
```diff
@@ -786,21 +789,14 @@ fn stateChanged(self: *XMLHttpRequest, state: ReadyState, exec: *const Execution
...
-        return .PUT;
-    }
-    if (std.ascii.eqlIgnoreCase(method, "propfind")) {
-        return .PROPFIND;
-    }
+    if (std.ascii.eqlIgnoreCase(method, "get")) return .GET;
+    if (std.ascii.eqlIgnoreCase(method, "put")) return .PUT;
+    if (std.ascii.eqlIgnoreCase(method, "post")) return .POST;
+    if (std.ascii.eqlIgnoreCase(method, "delete")) return .DELETE;
+    if (std.ascii.eqlIgnoreCase(method, "head")) return .HEAD;
```
karlseguin, line 805: A unit test would help here, and this is probably the WPT issue you're saying: HttpClient needs to be updated, it currently does:

```zig
        try conn.setMethod(req.method);
        if (req.body) |b| {
            try conn.setBody(b);
        } else {
            try conn.setGetMode();
        }
```

Why 2? setMethod just sets the physical text "GET" or  "HEAD" in the request. But the other 2 lines are what changes curl's actual behavior. You need a 3rd branch that sets `CURLOPT_NOBODY`  (`setGetmode()` does that, but it does MORE than that too).

mookums: I made this change but I think the core of that WPT issue is that our Method is a closed enum so we get `InvalidMethod` for any Method that isn't within our closed set so "CHICKEN" won't work unless we add a variant that carries a slice representing what got passed in.

PR #3485, opened 2026-09-10
Fetch Referrer Policy + CorsGate referrer checking

mookums:
This allows us to have per Request referrer overrides, this is something used by `/fetch/api/cors/cors-preflight-referrer.any.html`. This fixes the entire suite getting us from 0/12 -> 12/12.


src/browser/webapi/net/Request.zig
```diff
@@ -162,6 +173,24 @@ pub fn init(input: Input, opts_: ?InitOpts, exec: *const Execution) !*Request {
         .request => |r| r._signal,
     };
 
+    const referrer_value: ReferrerValue = if (opts.referrer) |r| blk: {
+        if (r.len == 0) break :blk .none;
+        if (std.mem.eql(u8, r, "about:client")) break :blk .client;
+        break :blk .{ .url = try URL.resolve(arena.allocator(), exec.base(), r, .{ .encoding = exec.charset.* }) };
+    } else switch (input) {
+        .url => .client,
+        .request => |r| r._referrer,
```
krichprollsch, line 182: If the value is an url, you don't own the string.

src/browser/webapi/net/Request.zig
```diff
@@ -342,6 +373,8 @@ pub fn clone(self: *const Request, exec: *const Execution) !*Request {
         ._mode = self._mode,
         ._body = if (self._body) |b| try arena.dupe(u8, b) else null,
         ._signal = self._signal,
+        ._referrer = self._referrer,
```
krichprollsch, line 376: If the value is an url, you share the string between 2 requests.

src/browser/webapi/net/Fetch.zig
```diff
@@ -133,7 +135,28 @@ pub fn init(input: Input, options: ?InitOpts, exec: *const Execution) !js.Promis
...
+                    if (try referrer.compute(transfer.arena.allocator(), policy, source, transfer.req.url)) |ref| {
+                        try transfer.setHeader("Referer", ref, .{});
+                    }
+                    transfer.req.referrer_policy = policy;
+                } else {
+                    try exec.headersForRequest(transfer);
+                }
+            },
+            .url => |source| {
+                const policy = request._referrer_policy orelse referrer.Policy.default;
```
krichprollsch, line 153: I seems you must fallback to the frame's referrer policy at some point, here and potentially in the client case. You should check against the specs.

src/browser/webapi/net/Request.zig
```diff
@@ -162,6 +173,24 @@ pub fn init(input: Input, opts_: ?InitOpts, exec: *const Execution) !*Request {
         .request => |r| r._signal,
     };
 
+    const referrer_value: ReferrerValue = if (opts.referrer) |r| blk: {
+        if (r.len == 0) break :blk .none;
+        if (std.mem.eql(u8, r, "about:client")) break :blk .client;
+        break :blk .{ .url = try URL.resolve(arena.allocator(), exec.base(), r, .{ .encoding = exec.charset.* }) };
```
krichprollsch, line 179: It seems you must check the referrer is the same origin.

src/browser/webapi/net/Fetch.zig
krichprollsch: Here is the claude's suggestion (`exec.referrerPolicy` must be added)

```zig
  const source: ?[:0]const u8 = switch (request._referrer) {
      .none => null,
      .client => exec.referrerSource(),
      .url => |u| u,
  };
  if (source) |s| {
      const policy = request._referrer_policy orelse exec.referrerPolicy();
      if (try referrer.compute(transfer.arena.allocator(), policy, s, transfer.req.url)) |ref| {
          try transfer.setHeader("Referer", ref, .{});
          transfer.req.referrer_policy = policy;
      }
  }
```

src/browser/webapi/net/Request.zig
```diff
@@ -162,6 +173,24 @@ pub fn init(input: Input, opts_: ?InitOpts, exec: *const Execution) !*Request {
...
+    } else switch (input) {
+        .url => .client,
+        .request => |r| r._referrer,
+    };
+
+    // Per spec, an unrecognized policy string is ignored
+    // referrer.parse already returns null for that case.
+    const referrer_policy: ?referrer.Policy = if (opts.referrerPolicy) |rp|
+        referrer.parse(rp)
```
krichprollsch, line 188: you must return a `TypeError` in case of invalid policy.

krichprollsch (commented):
referrer and referre policy are exposed as RO properties by the request, maybe you can add that too.

src/browser/webapi/net/Request.zig
mookums: This is wrong I think. Results in a regression of the WPT tests down to 10/12.

```
Fail 10/12      "/fetch/api/cors/cors-preflight-referrer.any.html"
        Pass    "Referrer policy: no-referrer and referrer: default"
        Pass    "Referrer policy: no-referrer and referrer: 'myreferrer'"
        Fail    "Referrer policy: \"\" and referrer: default"
                "|promise_test: Unhandled rejection with value: object \"TypeError: Failed to construct Request\""
        Fail    "Referrer policy: \"\" and referrer: 'myreferrer'"
                "|promise_test: Unhandled rejection with value: object \"TypeError: Failed to construct Request\""
        Pass    "Referrer policy: no-referrer-when-downgrade and referrer: default"
        Pass    "Referrer policy: no-referrer-when-downgrade and referrer: 'myreferrer'"
        Pass    "Referrer policy: origin and referrer: default"
        Pass    "Referrer policy: origin and referrer: 'myreferrer'"
        Pass    "Referrer policy: origin-when-cross-origin and referrer: default"
        Pass    "Referrer policy: origin-when-cross-origin and referrer: 'myreferrer'"
        Pass    "Referrer policy: unsafe-url and referrer: default"
        Pass    "Referrer policy: unsafe-url and referrer: 'myreferrer'"
Fail 0/1        "/fetch/api/cors/cors-preflight-referrer.any.worker.html"
        Fail    "no test suite completion"
                "|The test never reaches the completion callback."
```

src/browser/webapi/net/Fetch.zig
```diff
@@ -141,7 +143,20 @@ pub fn init(input: Input, options: ?InitOpts, exec: *const Execution) !js.Promis
         if (request._headers) |h| {
             try h.populateRequestHeaders(transfer);
         }
-        try exec.headersForRequest(transfer);
```
krichprollsch, line 144: Not calling to `headersForRequest` could generate divergences in headers between fetch and other requests. For now it's ok b/c we only set `referer`,  but later we could add more headers...
Idk what is the good way to handle that: add a way to override headers into `headersForRequest`?

Or at least add a comment into `Frame.headersForRequest` to indicates fetch doesn't call it.

src/network/CorsGate.zig
```diff
@@ -486,6 +486,11 @@ fn fetchThenResume(self: *CorsGate, transfer: *Transfer) !void {
     const owned_key = try arena.dupe(u8, key);
     const owned_origin = try arena.dupe(u8, origin);
 
+    const owned_referer: ?[]const u8 = if (transfer.findRequestHeader("referer")) |r|
+        try arena.dupe(u8, r)
```
krichprollsch, line 490: No need to dupe the referer here, `setHeader` will dupe it already.

src/browser/webapi/net/Request.zig
```diff
@@ -173,6 +183,29 @@ pub fn init(input: Input, opts_: ?InitOpts, exec: *const Execution) !*Request {
         .request => |r| r._signal,
     };
 
+    const referrer_value: ReferrerValue = if (opts.referrer) |r| blk: {
+        if (r.len == 0) break :blk .none;
+        if (std.mem.eql(u8, r, "about:client")) break :blk .client;
+        const resolved = try URL.resolve(arena.allocator(), exec.base(), r, .{ .encoding = exec.charset.* });
+        if (!exec.isSameOrigin(resolved)) break :blk .client;
+        break :blk .{ .url = resolved };
+    } else switch (input) {
```
krichprollsch, line 192: The specs says you must set `.client` if `_opts` is not null,


```suggestion
   } else if (opts_ != null) .client else switch (input) {
```

> If init [is not empty](https://infra.spec.whatwg.org/#map-is-empty), then:
> Set request’s [referrer](https://fetch.spec.whatwg.org/#concept-request-referrer) to "client".
> Set request’s [referrer policy](https://fetch.spec.whatwg.org/#concept-request-referrer-policy) to the empty string.
https://fetch.spec.whatwg.org/#dom-request

src/browser/webapi/net/Request.zig
```diff
@@ -173,6 +183,29 @@ pub fn init(input: Input, opts_: ?InitOpts, exec: *const Execution) !*Request {
...
+            .url => |u| .{ .url = try arena.dupeZ(u8, u) },
+            else => r._referrer,
+        },
+    };
+
+    // Per spec, an unrecognized policy string is ignored
+    // referrer.parse already returns null for that case.
+    const referrer_policy: ?referrer.Policy = if (opts.referrerPolicy) |rp|
+        referrer.parse(rp)
+    else switch (input) {
```
krichprollsch, line 204:
```suggestion
    else if (opts_ != null)
        null
    else switch (input) {
```

src/browser/webapi/net/Fetch.zig
mookums: I put it back in but above where we handle setting the Referer and RefererPolicy. This way, we just overwrite the Headers if we meet the conditions.

krichprollsch:
I think I'm OK with the code, but 2 WPT tests you targeted don't work:
```
$ ./wptrunner /fetch/api/cors/cors-preflight-referrer.any.html
2026/09/16 15:28:09 INFO test suite length=37266
Fail 10/12      "/fetch/api/cors/cors-preflight-referrer.any.html"
        Fail    "Referrer policy: no-referrer and referrer: default"
                "|assert_equals: Preflight's referrer is correct expected \"\" but got \"http://web-platform.test:8000/\""
        Fail    "Referrer policy: no-referrer and referrer: 'myreferrer'"
                "|assert_equals: Preflight's referrer is correct expected \"\" but got \"http://web-platform.test:8000/\""
```

mookums:
Changed this so that headersForRequest takes an option for referer. The issue is that if we always do it, we end up setting fields even when we shouldn't.

krichprollsch (approved):

PR #3492, opened 2026-09-11
`cookies`: accept Netscape cookie format in `--cookie` argument

nikneym:


src/cookies.zig
```diff
@@ -150,6 +163,206 @@ fn parseJsonSameSite(value: ?[]const u8) Cookie.SameSite {
...
+                try allocator.dupe(u8, bare_domain);
+
+            // Bake a cookie.
+            const cookie = Cookie{
+                .arena = cookie_arena,
+                .name = name,
+                .value = value,
+                .domain = domain,
+                .path = _path,
+                .expires = fields.expires_at,
```
krichprollsch: you can't take `0` as it b/c we will discard it immediately. `0` in netscape means session's duration.
So I guess you should set to null value instead.

krichprollsch (approved):

PR #3500, opened 2026-09-13
Check native stack headroom through V8 before re-entry

staylor:
## Status — native Linux validated; x86_64 CI/TSan awaiting approval

https://github.com/lightpanda-io/zig-v8-fork/pull/206 is merged and included in **v0.5.6**. This branch now pins upstream release commit `d3d7b41677a0015fdfa55a8b1caa4f214de6d209` with its recalculated Zig package hash. The install action, V8 snapshot-cache action, and Dockerfile all select `v0.5.6`.

`make download-v8` now supplies matching official archives; locally augmented archives are no longer needed. macOS and native Linux aarch64 integration checks pass. This PR is ready for review so the existing native Linux x86_64 release and debug/TSan jobs can run, but **those jobs still require maintainer approval**: [zig-test run](https://github.com/lightpanda-io/browser/actions/runs/35046131112). They are not yet validation passes.

## Why

Deep JS → native → JS recursion can leave too little stack for another V8 call or exception reporting. Reconstructing V8's limit in Zig can disagree with the engine's platform handling, sanitizer defaults, and runtime overrides.

Ask V8 directly through `HasStackHeadroom`, a non-allocating wrapper around its own `StackLimitCheck`. This is separate from the TryCatch-copy fix in #3497 (regression coverage in #3498).

## Changes

- Reserve headroom before function/constructor calls, script compilation/execution, module work, and exception formatting.
- Defer both microtask queues when headroom is insufficient.
- Count refused listener calls without invoking another error handler. This is containment, not a claim of identical error-reporting behavior.
- Add deterministic refusal/recovery tests, including resumed observer delivery.
- Use the upstream v0.5.6 source and archives instead of the companion fork commit and v0.5.5 archive configuration, including Docker builds. V8 itself remains `14.9.207.35`.

## Current validation

Using the **unmodified official v0.5.6 macOS arm64 archive** (validated on main at `6d87fbf0e`, then rebased onto `9cc702f4a`, whose additional changes only regenerate orderfiles):

- `make test`: **1530/1530**, Debug with leak checks.
- Headroom unit tests under `ReleaseFast`: **4/4**.
- ReleaseFast application build succeeds.
- CDP recursion/recovery checks pass with 1/2/4 MiB stack limits, including space, newline, underscore and single-dash flag forms. MutationObserver delivery resumes and the CDP connection stays alive in all five cases.
- CI formatting command and `git diff --check` pass.

### Native Linux aarch64 — current head `04de841f3`

The official Linux aarch64 release archive finished uploading after the initial release check. Using it unmodified on native aarch64/glibc (not QEMU):

- `zig build -j1 -Ddev_fast=false test`: **1530/1530**, Debug with leak checks.
- Debug browser application build succeeds.
- All five CDP recursion/recovery cases pass: 1/2/4 MiB stack limits and alternate flag forms. MutationObserver delivery resumes, a subsequent `page.evaluate` succeeds, and the connection remains usable.

### Remaining x86_64 / sanitizer checks

- Both official x86_64 release archives export the new symbol. Earlier execution of a minimal probe under QEMU failed internally before yielding results; that attempt is not counted as a pass.
- The v0.5.6 x86_64 debug archive is now published. Its release workflow builds with `is_tsan=true` and `v8_enable_sandbox=true`, matching the browser's existing `zig-test-debug` job.
- The native x86_64 release and debug/TSan jobs are waiting on the workflow approval linked above. **Wait for those checks before merging.**
- No ASan run is claimed. ASan is a separate follow-up: the current `-Dasan=true` option requests an instrumented V8 source build, not complete Zig/Rust instrumentation.

The merged wrapper retains the companion's behavior; its C++ reserve parameter is now `uintptr_t`, matching the C header's pointer-width `usize` ABI on the supported targets.


staylor:
Native Linux/aarch64 validation of 04de841f3 now passes with the official v0.5.6 archive: 1530/1530 tests and all five CDP stack-limit/recursion recovery cases. Marked ready to trigger x86_64 CI. Could a maintainer approve https://github.com/lightpanda-io/browser/actions/runs/35046131112 so the release and debug/TSan jobs can run? The matching debug/TSan archive has now been published. Those checks are still pending, not passed.

karlseguin:
On Linux x86_64 I couldn't reproduce this. The new test works fine in `main` (Debug and ReleaseFast). The new gate makes it stop a recursion a bit sooner, but that's it. Have you hit a case where this is an issue?

I _was_ able to make it crash by having too much recursion on the Zig side (dumping a huge tree), but that crashes the same on main as it does here. It's something I'm keen to fix, but I think that's a separate issue.

PR #3507, opened 2026-09-14
Share pointer/mouse click dispatch across click paths

R4m1r0qu41:
## Summary

Follows up on the reviewer's "longer term" note on #3431:

> Longer term this is now the third copy of the same sequence (`WebDriver.click`, `user_input.triggerMousePress/Release`, and this one), and they already disagree on the points above. Not asking for it in this PR, but a shared `pub fn` in `frame/user_input.zig` that all three call would be the real fix, and would also give Puppeteer/Playwright clicks `pointerdown`, which they currently lack.

- Adds shared `dispatchPointerPress`/`dispatchPointerRelease`/`dispatchClickAsPointer` primitives in `frame/user_input.zig`, now used by `actions.click`, `WebDriver.click` (testdriver's simple click), and CDP's `triggerMousePress`/`triggerMouseRelease` (what Puppeteer/Playwright drive).
- CDP mouse input (Puppeteer/Playwright) now gets `pointerdown`/`pointerup` for the first time, closing the gap the reviewer's comment named.
- `WebDriver.click` gains suppress/focus handling and a `PointerEvent` click it never had before (it used to dispatch a fixed five-event sequence unconditionally).
- Multi-button chords on the CDP path (a second mouse button pressed while the first is still held) are now tracked correctly: `pointerdown`/`pointerup` fire only at the aggregate mask's 0/nonzero transitions, matching the [Pointer Events spec](https://www.w3.org/TR/pointerevents3/#chorded-button-interactions) and real Chrome, instead of firing a duplicate `pointerdown` per button and corrupting pointerdown-cancellation state across the gesture.
- `performPointerSource` (testdriver's `action_sequence`, a fourth, more complex copy) and each call site's own error-handling philosophy are deliberately left untouched — see the PR log for the full list of what was and wasn't unified, and why.

## Review history

This branch went through five independently-verified local rounds before being opened:

1. Initial implementation, reviewed by an independent verifier before the first commit (caught 3 real bugs: a dropped click-count, lost modifier propagation on `WebDriver.click`, and a suppression-flag leak across gestures — all fixed).
2. A Grok review pass cleaned up comments that narrated the refactor's history instead of stating the invariants (one suggested defensive fix was checked against the code and rejected with a written reason).
3. A Codex audit reproduced a real defect in the new CDP pointer-chord handling (a chorded press was treated as an unrelated gesture, corrupting cancellation state) — fixed and verified against the audit's own harness and real Chrome.
4. Rebased onto `main` as it moved.
5. A second Codex audit found a narrower regression in the chord fix itself (a chorded press skipped its `mousedown` focus default action) — fixed and independently re-verified, including against real Chrome, before this PR was opened.

Full design notes, every finding, and exact verification commands/results are in this session's working log (happy to share on request).

## Test plan

- [x] `zig build test`: 1524/1524
- [x] `zig build test -Dwpt_extensions`: 1524/1524
- [x] `zig fmt --check` clean
- [x] New regression tests added for: the full pointer/mouse event sequence on the CDP path, cancelled-pointerdown suppression across split `mousePressed`/`mouseReleased` CDP messages, a multi-button chord (no duplicate pointerdown/pointerup, correct suppression), and chorded-press focus — each confirmed to fail against the pre-fix code before being trusted
- [x] Independent audit harness (dependency-free Node script, driving the built binary over real CDP/MCP) cross-checks the same contracts against real Chrome

https://claude.ai/code/session_01XLXnBHBxQNskg2MAke3Lhv


arrufat (commented):
Thanks for taking on the "longer term" note from #3431 — this is the real fix, and the shape is right: one pair of press/release primitives, CDP/Puppeteer finally getting `pointerdown`/`pointerup`, and the chord tracking on `Page`.

I drove the fixture through real Chrome over CDP with the exact `Input.dispatchMouseEvent` messages the new tests use. **The chord pointer tracking matches Chrome exactly** — press-left → press-right → release-right → release-left gives `pointerdown:1, pointermove:3, pointermove:1, pointerup:0` in both. That was the hard part and it's correct. The split-message suppression via `input_pressed_buttons`/`input_mousedown_suppressed` mirrors `input_hover_target` nicely, and the parametrized chord-focus test pins exactly the regression the audits caught.

One thing worth addressing, and two notes.

### Should fix — CDP `mousedown` drops the click count

The new CDP test asserts `mousedown:0:1:0` (detail 0). Real Chrome, given the same message, fires `mousedown:0:1:1` (detail 1):

```
Chrome  #btn press+release:  pointerdown:0:1:0  mousedown:0:1:1  pointerup:0:0:0  mouseup:0:0:1  click:0:0:1
```

The MCP path (`actions.click`) already sends detail 1 and matches Chrome; only the CDP `triggerMousePress` path sends 0, so the two paths disagree and the CDP one is the odd one out. It's pre-existing (that function never took `clickCount`), but the PR refactors this exact code and now locks the deviation in with a test and a comment calling it intentional. `Input.dispatchMouseEvent` already carries `clickCount` — threading it into `triggerMousePress` → `mouse_detail` would match Chrome and align the paths. At minimum I'd not assert `:0` as the expected contract.

### Note — the chord's *activation* events aren't Chrome-faithful

Same chord in Chrome fires `contextmenu` on the right **press** (`buttons:3`) and `auxclick` on the right release, with **no** trailing left `click`:

```
Chrome  chord:  pointerdown:1  pointermove:3  contextmenu:3  pointermove:1  auxclick:1  pointerup:0
this PR chord:  pointerdown:1  pointermove:3  pointermove:1  contextmenu:1  pointerup:0  click:0
```

These live in the pre-existing release switch, not your new pointer code, so you didn't introduce them — but the chord test now codifies the lightpanda-specific order as the contract. A comment marking it a known deviation, or scoping the assertion to the pointer events this PR actually owns, would keep it honest.

### Nit — `dispatchClickAsPointer` hardcodes `buttons: 0`

A primary click that fires while another button is still held (release primary mid-chord) should carry the still-held mask; this reports 0. Extreme edge case, just noting it.

Everything else looks good — `WebDriver.click` gaining focus and a `PointerEvent` click is the right call, and leaving `performPointerSource` and the per-call-site error philosophies alone was the right scope.

arrufat (changes requested):

R4m1r0qu41:
Thanks for the thorough review — pushed a follow-up (f44539d) addressing all three points.

**Should fix (CDP mousedown detail)**: `triggerMousePress` now takes `click_count` and threads it into `mouse_detail` for both the first-press and chorded-press paths, matching the MCP path. Re-verified against live Chrome (headless, raw `Input.dispatchMouseEvent` over CDP) — `mousedown:0:1:1` for `clickCount:1`, matching your trace exactly.

While verifying that fix I caught a regression it introduced: the fallback for `click_count <= 0` forced detail to `1`, but Chrome (and Firefox) actually preserve `0` for an omitted/zero `clickCount` — and don't fire `click` at all in that case. Fixed the fallback to preserve `0` instead. Added a parametrized test over clickCount 0/1/2 for mousedown's detail.

**Nit (`dispatchClickAsPointer` hardcoded `buttons: 0`)**: we did tackle this edge case in this same round — `triggerMouseRelease` now passes `remaining` (the still-held mask) through instead of hardcoding 0. Confirmed against live Firefox via WebDriver BiDi: pressing left+right, releasing left first (right still held) fires `click:2` in real Firefox, matching the fix.

**Note (chord activation-event ordering)**: left alone, as you scoped it — but while I had Firefox wired up I checked it there too, and it's not Chrome-specific: Firefox also fires `contextmenu` on the right *press* (not release) and also fires `auxclick` on the right release, same as your Chrome trace. Added a comment on the chord test flagging this as a known, now cross-browser-confirmed deviation, in case it's worth a follow-up PR.

One more pre-existing item, found while verifying the clickCount fix: `triggerMouseRelease` has the identical zero/omitted-clickCount fallback bug on the release side (forces detail to 1 instead of preserving 0). Didn't touch it — it's pre-existing and out of scope here, but flagging it since it's the same class of bug you caught.

All new/changed behavior was re-tested against real Chrome and Firefox, not just the fixture assertions. `cdp.input` 20/20, `bidi.input` 5/5, `zig fmt --check` clean.

arrufat (approved):
Great follow-up — all three points addressed, and I re-verified the two behavior changes against live Chrome (raw `Input.dispatchMouseEvent` over CDP), matching your traces:

| check | Chrome | this PR |
| --- | --- | --- |
| `mousePressed` clickCount 0/1/2 → `mousedown.detail` | 0 / 1 / 2 | 0 / 1 / 2 |
| primary release mid-chord → `click.buttons` | 2 | 2 |

Preserving `0` for an omitted/zero clickCount rather than forcing `1` is the right call, and Chrome agrees. Nice catch on the regression during your own verification.

On the release-side item you flagged: confirmed real against Chrome — for clickCount 0 it fires `mouseup:0:0:0` and **no** `click`, so the press side now matches Chrome while the release side still forces detail 1 and fires `click`. Agreed it's pre-existing and out of scope here; worth a small follow-up since the release side also needs to skip the `click`/activation when clickCount is 0, not just preserve the detail. The chord activation-event ordering (contextmenu-on-press, auxclick, cross-browser-confirmed) is the natural companion for that same follow-up.

Approving. Thanks for the careful, browser-checked work on this one.

PR #3510, opened 2026-09-14
internal: store document in every node

karlseguin:
Follow up to https://github.com/lightpanda-io/browser/pull/3501.

To recap: many apis use the injected-frame, when, for cross-frame operations, can be wrong. We've been moving to using the Node's Document's Frame on a case- by-case basis. The goal with this series of PRs is to more holistically fix it.

The main goal of this commit is to make Node.getDocument fast. Previously, Node.nodeDocument (renamed to Node.getDocument) and Node.ownerFrame had to walk up the _parent tree to find the document. Now there's a DocumentRegistry on the Browser, and every Node has a `_document: u32` index. Thus, Node -> Document is O(1). This makes it so we can have correctness without a performance cost.

The DocumentRegistry is held on the Browser because of shared CDP nodes. A follow up should be able to move this to the Page and change the index to u16.

Adding a DocumentRegistry to Browser, and a document: u32 to Node is easy. The reason this PR is so big is because every `node_factory` and many Factory methods need to be aware of all of this. You can't create an HTMLDivElement without the document it belongs to. I incorporated more expected future changes into the node_factory/factory mechanism so that  [hopefully] the next PRs don't have to touch any of this code (e.g. I removed Frame as a parameter and added a *Page to every Document so that documents have access to factory/arena/ etc..because a `*Document` might not have a `*Frame`, but it will always have a `*Page`.

Give every document an index (u32), and store that index in Node.


src/browser/webapi/Node.zig
```diff
@@ -708,58 +706,36 @@ pub fn ownerDocument(self: *const Node, frame: *const Frame) ?*Document {
...
 
-fn ownerDocumentIncludingSelf(self: *const Node, frame: *const Frame) ?*Document {
+// The spec's "node document": every node has one, a document's is itself.
+// `frame` only reaches the registry; any frame will do.
+pub fn getDocument(self: *const Node, frame: *const Frame) *Document {
     if (self._type == .document) {
         return self.subtype(Document);
     }
-    return self.ownerDocument(frame);
+    return self.ownerDocument(frame).?;
```
krichprollsch: Shouldn't we put an assert here?

krichprollsch (approved):

src/browser/webapi/Node.zig
karlseguin: No, a `Node` *always* has a `Document`. `ownerDocument` is the WebApi and it returns `null` only when the `node == Document`. The `if` check directly above checked that.

karlseguin: I shouldn't be so confident.... I added an `lp.assert`. Cheap to add, gives us a much better error if I'm wrong.

PR #3522, opened 2026-09-15
`Regex`: share the PCRE2 wrapper; `findElement` matches names by `/regex/`

arrufat:
PR #3463 linked PCRE2 for `/regex/` adblock filters. This lifts the wrapper out of the adblock directory so other code that runs someone else's pattern can use it, and adds the first such consumer.

## `Regex` becomes shared

- `src/network/adblock/Regex.zig` → `src/Regex.zig`, re-exported as `lp.Regex`.
- `compile` takes an `Options` struct (`case_insensitive`, `unicode`) and an optional `Diagnostic` that carries PCRE2's message and offset. `unicode` is `PCRE2_UTF | PCRE2_MATCH_INVALID_UTF` with UCP left off, so `\b`/`\w` stay ASCII as in JavaScript and an invalid byte in a DOM string fails to match rather than erroring.
- `App` owns the one `Regex.Context`; the adblocker borrows it instead of creating its own. The adblock log line for a rejected pattern now reads from the diagnostic and is otherwise unchanged.

## `findElement` accepts `name: "/regex/"`

`name` is a substring, which cannot say "starts with", "exactly this" or "either of these". A name written as `/.../` is now a JavaScript regex literal (unanchored; flags `i`, `m`, `s`, `u` accepted, any other flag is a tool error; case-insensitive even without `i`, `(?-i)` opts out). The spelling is the one the adblock parser already reads as a regex literal and the one Playwright and uBO use, so it stays one field with no extra rule. Since matching is unanchored and `/` is an ordinary pattern character, an accessible name that really is written as `/foo/` still matches its own text.

An invalid pattern comes back as a tool error with PCRE2's text so the model can fix it rather than retry blind:

```
findElement: invalid name regex '(': missing closing parenthesis at offset 1
```

The compile happens per call and is freed in the same function; a pattern cache is not warranted at tool-call frequency.

## Not in this PR

- A `:has-text(/re/)` scope for `markdown`/`extract`: that needs a selector-engine hook, not a tool field.
- The docs site's `findElement` row should mention the `/.../` form of `name`.


src/browser/tools.zig
```diff
@@ -2078,14 +2073,36 @@ fn execFindElement(arena: std.mem.Allocator, session: *lp.Session, registry: *No
...
     lp.interactive.registerNodes(matched, registry) catch
         return ToolError.InternalError;
-    return renderJson(arena, matched);
+    return .{ .text = try renderJson(arena, matched) };
+}
+
+/// The body of a `/.../` literal, the spelling adblock lists use for a regex
+/// too. Unanchored, so a name that really is written that way still matches.
+fn regexBody(text: []const u8) ?[]const u8 {
+    if (text.len > 2 and text[0] == '/' and text[text.len - 1] == '/') return text[1 .. text.len - 1];
```
karlseguin: I know the help says to use `/.../` with `(?-i) ` as a prefix...but we don't think it'll ever try to use a modifier, e.g. `/spice/i` ?

src/network/adblock/AdBlocker.zig
```diff
@@ -225,9 +222,15 @@ pub fn parse(self: *AdBlocker, reader: *Io.Reader) !void {
 
         if (filter.kind == .regex) {
             const body = filter.pattern[1 .. filter.pattern.len - 1];
-            const compiled = Regex.compile(self.regex_context, body, !filter.match_case) catch |err| switch (err) {
+            var diag: Regex.Diagnostic = .{};
+            const compiled = Regex.compile(self.regex_context, body, .{ .case_insensitive = !filter.match_case }, &diag) catch |err| switch (err) {
```
karlseguin: why not as a method? `self.regex_context.compile(body, .{...});`

arrufat: Done, `compile` is a method on `Context` now (35dbda368).

src/browser/tools.zig
arrufat: Fair, it will. The literal now takes flags ( 5710f1d68 ): `i` and `u` are already how names match, `s` and `m` map to their PCRE2 options, and any other letter is a tool error that names it (`/x/g` → unsupported regex flag 'g'). A non-letter after the closing slash makes the whole thing plain text again, so a name like `/foo/ bar` still substring-matches.

Kept case-insensitive by default so `/Spice/` and `/spice/i` both match; `(?-i)` stays the opt-out, since JavaScript has no flag for "case-sensitive".

PR #3523, opened 2026-09-15
Resolve scroll containers through the style cascade and unify wheel dispatch

arrufat:
Follow-ups to #3515, one commit each.

## Stylesheet-declared scroll containers

`Element.scrollContainer` read the inline `style=` attribute only, so a scroller declared in a stylesheet was invisible to the scroll tool and to wheel scrolling, which fell through to the viewport.

`StyleManager` now tracks `overflow-x` and `overflow-y` alongside display, visibility, opacity and pointer-events, and exposes `overflowAxes(el)` as an own-element probe. The `overflow` shorthand is expanded into its longhands in declaration order, in both the attribute scan and the materialized style object, so a shorthand and its longhands keep the precedence of the source text. `overlay` counts as auto, as in Chrome. The two unused `Props` bits hold the flags, so the per-element memo does not grow.

The `Visibility*` types are renamed `Tracked*` in a separate commit, since the cascade no longer tracks only visibility.

## One wheel path

CDP/BiDi wheel and WebDriver wheel each had their own definition of what a wheel does. `user_input.wheel` now owns the sequence: `wheel`, Blink's legacy `mousewheel`, then the scroll unless either was canceled, each non-cancelable when every listener on its path is passive. The passivity query moves onto `EventManager`. Callers supply the target and deltas.

Behavior change for CDP: a page's document-level `wheel` listener can no longer cancel scrolling unless it opts out of the default-passive behavior, and `mousewheel` listeners now fire. Both match Chrome.

## Explicit scroll target and result

`Element.scrollContainers` walks once for every requested axis and returns a `ScrollTarget` per axis, viewport or container, plus the nearest. The tool's fallback to the given node is a visible switch arm. `actions.ScrollResult` names what scrolled as a tagged union, so the tool and `LP.scrollNode` format it without touching the request.

## Tests

- Stylesheet scroller in `mcp_actions.html`, covered by the CDP wheel test and the MCP scroll test.
- `StyleManager` unit tests cover shorthand and longhand precedence, `!important`, `overlay`, and memo invalidation.
- New CDP test for wheel cancelability following listener passivity and for `mousewheel`.


src/browser/webapi/Element.zig
```diff
@@ -1603,41 +1603,77 @@ pub fn setScrollLeft(self: *Element, value: i32, frame: *Frame) !void {
...
 
-/// Nearest ancestor-or-self that is a scroll container along any of `axes`.
-/// null once the chain reaches html/body: those scroll the viewport.
-pub fn scrollContainer(self: *Element, axes: ScrollAxes, frame: *Frame) ?*Element {
-    if (!axes.x and !axes.y) return null;
-    const owner = self.ownerFrame(frame) orelse return null;
-    const style_manager = &owner._style_manager;
+/// What a scroll along one axis lands on. html and body scroll the viewport,
+/// and so does a detached element.
+pub const ScrollTarget = union(enum) {
```
karlseguin: Don't think it needs to be `pub`

src/browser/webapi/Element.zig
```diff
@@ -1603,41 +1603,77 @@ pub fn setScrollLeft(self: *Element, value: i32, frame: *Frame) !void {
...
+    pub fn scrollBy(self: ScrollTarget, left: i32, top: i32, frame: *Frame) !void {
+        const opts: ScrollToOpts = .{ .opts = .{ .left = left, .top = top } };
+        return switch (self) {
+            .container => |el| el.scrollBy(opts, null, frame),
+            .viewport => frame.window.scrollBy(opts, null, frame),
+        };
+    }
+};
+
+pub const ScrollTargets = struct {
```
karlseguin: Don't think it needs to be `pub`

src/browser/StyleManager.zig
```diff
@@ -742,14 +752,16 @@ fn ownProps(self: *StyleManager, el: *Element) Props {
...
+const property_fields = std.meta.fieldNames(TrackedProperties);
 
 // INLINE_PRIORITY can't be beaten, so allInline ends the rule scan early.
 const Priorities = struct {
     display: u64 = 0,
     visibility_hidden: u64 = 0,
     opacity_zero: u64 = 0,
     pointer_events_none: u64 = 0,
+    overflow_x_scrolls: u64 = 0,
+    overflow_y_scrolls: u64 = 0,
```
karlseguin, line 763: Do you think we should remove the `allInline` check? Seemed unlikely before, not seems like it'll never fire. Cheap, but an extra branch people have to understand / reason about.

src/browser/StyleManager.zig
```diff
@@ -1004,19 +1016,19 @@ fn getBucketKey(compound: Selector.Compound) ?BucketKey {
...
-    var props: VisibilityProperties = .{};
-    const decl = style.asCSSStyleDeclaration();
-    for (property_names) |name| {
-        if (decl.findProperty(.wrap(name))) |property| {
-            props.apply(name, property._value.str());
-        }
+/// Extracts the tracked properties from a style declaration. The object holds
+/// one entry per name in first-declared order, so folding it in order gives a
+/// shorthand and its longhands the same precedence as the source text.
+fn extractTrackedProperties(style: *CSSStyleProperties) TrackedProperties {
```
karlseguin, line 1013: right now, if you `setPropertyImpl("overflow", "hidden"); it doesn't expand it into the overflow-y and overflow-x properties. So, if you have:

```html
style="overflow: hidden; overflow-y: scroll"
```

and then call `setProperty("overflow", "hidden")`, internally nothing changes...we overwrite the existing "overflow" entry.

The solution seems to be to expand the property on set and remove the shorthand. But, maybe you don't think it's worth it.

src/browser/EventManager.zig
```diff
@@ -149,6 +149,21 @@ pub fn hasDirectListeners(self: *EventManager, target: *EventTarget, typ: []cons
     return self.base.hasDirectListeners(target, typ, handler);
 }
 
+/// Whether any listener for `typ` on the path from `node` up to the window
+/// is non-passive. The UA dispatches scroll-blocking events as
+/// non-cancelable when none is: it already knows preventDefault can't be
+/// called.
+pub fn hasNonPassiveListener(self: *EventManager, node: *Node, comptime typ: []const u8) bool {
```
karlseguin: This is taken from WebDriver, but I don't think it's right (and now it's potentially exposed to CDP/BiDi). `buildEventPath` is the real path the event will take.

Claude suggesting putting the following, in dispatchNode, after we have the `const path = ...` (around line 313).

```zig
if (event._cancelable_unless_passive) {
    event._cancelable = self.anyNonPassive(path_buffer[0..path_len], event);
}
```

So we'd need an `_cancelable_unless_passive: bool = false` on Event which this would set.

src/browser/frame/user_input.zig
```diff
@@ -287,42 +287,49 @@ pub fn triggerMouseWheel(frame: *Frame, x: f64, y: f64, delta_x: f64, delta_y: f
...
+        }
+    }
+    if (canceled) {
         return;
     }
 
-    // CDP deltas are untrusted, so guard NaN and saturate the addition.
-    try wheelScroll(target, deltaToScroll(delta_x), deltaToScroll(delta_y), frame);
+    // Deltas come from the wire, so guard NaN and saturate the addition.
+    try wheelScroll(target, deltaToScroll(delta_x), deltaToScroll(delta_y), owner);
```
karlseguin, line 316: This is now the only caller of `wheelScroll`, so that function doesn't have to be `pub`.

src/browser/frame/user_input.zig
```diff
@@ -287,42 +287,49 @@ pub fn triggerMouseWheel(frame: *Frame, x: f64, y: f64, delta_x: f64, delta_y: f
...
+/// still fires, then the scroll unless either was canceled. Each event is
+/// non-cancelable when every listener on its path is passive.
+pub fn wheel(frame: *Frame, target: *Element, x: f64, y: f64, delta_x: f64, delta_y: f64) !void {
+    // Listeners live in the event manager of the element's own frame, which
+    // is not the caller's when the element belongs to an iframe's document.
+    const owner = target.ownerFrame(frame) orelse return;
+    const event_manager = &owner._event_manager;
+
+    var canceled = false;
+    inline for (.{ "wheel", "mousewheel" }) |typ| {
```
karlseguin: If you put this in chrome:

```js
 var div = document.createElement('div');
  div.style = 'border: 1px solid red; width: 200px; height: 3000px; overflow-y: scroll';
  document.documentElement.appendChild(div);
  const w = e => console.log('wheel', e.type);
  div.addEventListener('wheel', w);
  div.addEventListener('mousewheel', e => console.log('mousewheel', e.type));
```

and then scroll in the red rectangle, you'll only get a "wheel" event.  Only if there's no "wheel" event will "mousewheel" fire..you can quickly test by remove the wheel listener: `div.removeEventListener('wheel', w);`

src/browser/webapi/Element.zig
arrufat: Done, it is private now.

src/browser/webapi/Element.zig
arrufat: Done, same here.

src/browser/StyleManager.zig
arrufat: Agreed, removed it.

src/browser/StyleManager.zig
arrufat: Went with expanding it on set. overflow is the one shorthand the cascade folds, so the declaration now stores it as overflow-x and overflow-y. Set and parse expand it, remove drops both, and reading or cssText recombine them when both are set with the same priority.

src/browser/EventManager.zig
arrufat: Done as you suggested. The flag lives on Event and dispatchNode resolves it against the path it just built. Inline handlers count as non-passive.

src/browser/frame/user_input.zig
arrufat: Done, it is private now.

src/browser/frame/user_input.zig
arrufat: You are right. Blink retypes the same event per target instead of firing a second one. The event manager now does that for trusted events: a target with a live wheel listener runs those, otherwise its mousewheel listeners see the event as mousewheel. One dispatch covers both names, and the same lookup decides what counts as non-passive on the path. The CDP test covers a target with both listeners, the fallback, and a mousewheel listener canceling the scroll.

Issue #3453, opened 2026-09-08
Network.setCookie sameSite: opaque 'InvalidEnumTag' error, rejects the CDP-spec values that getCookies outputs


Raknaos:
## Problem

`Network.setCookie` (and `setCookies`) rejects `sameSite` values with an **opaque, unhelpful error**, and the accepted casing is undocumented and differs from Chrome's. When a caller passes a value Chrome would send, the failure message is useless for debugging which field or value caused it.

## Reproduction (Lightpanda 1.0, CDP over 127.0.0.1:9222)

| `sameSite` value sent | Result |
|---|---|
| `"Lax"` | ✅ `{"success": true}` |
| `"None"` | ✅ works |
| `"Strict"` | ✅ works |
| `"lax"` (lowercase) | ❌ `{"code": -31998, "message": "InvalidEnumTag"}` |
| `"no_restriction"` (the value Chrome DevTools Protocol itself reports for `SameSite=None` cookies) | ❌ `{"code": -31998, "message": "InvalidEnumTag"}` |

The two failures are byte-for-byte identical:

```
{'code': -31998, 'message': 'InvalidEnumTag'}
```

They give **no indication of which parameter is wrong**, no list of accepted values, and are the same error for a wrong-case value and for a value that comes from the *documented CDP enum itself* (`CookieSameSite` in the Chrome DevTools Protocol declares `Strict | Lax | NoSameSite | none | lax | strict` — lowercase variants included).

## Why it matters

Any tool bridging **real browser cookies into Lightpanda** reads them with `Network.getCookies`, which reports `sameSite` using Chrome's casing convention (e.g. `"no_restriction"`). Feeding those cookies straight back via `setCookie` fails on every single one unless the caller already knows to map `no_restriction → None`, `lax → Lax`, `strict → Strict`. We hit this building https://github.com/Raknaos/lightpanda-session-bridge and had to add a normalizer (see `cookie_for_cdp()` in `relay/server.py`).

## Requested

1. **Accept the CDP-spec values** (`no_restriction`, case-insensitive `Lax`/`Strict`/`None`) — matching what `Network.getCookies` outputs, so a get→set round-trip always works.
2. **Improve the error**: `-31998 InvalidEnumTag` → include the field name and accepted values, e.g. `"Invalid value 'no_restriction' for 'sameSite'. Accepted: Strict | Lax | None"`.

Even #2 alone would have saved us hours of debugging.

PR #3529, opened 2026-09-15
`cdp`: report back invalid same-site

nikneym:
This PR satisfies the second request of #3453, `parseSameSite` accepts the lenient spellings and an unknown value returns `error.InvalidSameSite`, which the dispatcher reports as -31998 "InvalidSameSite".


karlseguin:
The second request of #3453 is to improve the error message. I'm not sure how that's satisfied. Could reasonably assume that this would be a nice improvement throughout the CDP parser..though I'm not sure how practical that is.

I tested against chrome, and passing either "unspecified" or "no_restriction" resulted in no sameSite setting, but with this PR, both cases set it to "none".  Also, setting garbage is chrome behaves the same, whereas we error (I'm not against this divergence though). I found other existing inconsistencies against chrome with this API, so it would be nice to clean it all up.

There's already a `parseJsonSameSite` in `cookies.zig` which (a) could be moved to `Cookie.zig` and updated to return an optional, because, from what I'm seeing, that's what unspecified/no_restriction/garbage means...it doesn't mean "None", it means no value. `Cookie.zig` seems to handle this correctly by simply skipping the property if it doesn't parse.

karlseguin:
Did you compare this behavior to Chrome? From what I can tell, it now rejects an invalid value, but (as I said in my previous comment), Chrome allows it and leave it as not-set. You also have `no_restriction` => .none but I was pretty sure I saw Chrome treat that as  not-set. 

The more interesting change, as suggested in the issue, would be to improve our error messages. But that might be a tall order given the state of Zig's JSON parsing.

Last time I actually compared it with Chrome. This time, I asked Claude to review and this is what it gave me:
```
I compared the diff with Chrome's MakeCookieFromProtocolValues in
  content/browser/devtools/protocol/network_handler.cc:482.

  What Chrome does: it sets css = UNSPECIFIED, then compares sameSite exactly against "Lax", "Strict" and
  "None". Anything else, including a missing value, stays unspecified. That means Lax by default, and
  getCookies leaves sameSite out. Chrome never rejects the value and never treats a missing one as None.

  What the PR still gets wrong:

  1. A missing sameSite is still None. CdpCookie.sameSite defaults to "None" (storage.zig:130). The most
     common case, a driver that just doesn't send the field, still gives a cross-site cookie. That's the
     real sameSite bug here, and the PR doesn't touch it.
  2. It's now inconsistent with itself. Sending "unspecified" gives Lax, but leaving the field out gives
     None. Before, both paths at least agreed.
  3. "unspecified" isn't real unspecified. It sets .lax but leaves same_site_default = false. The Set-Cookie
     path sets same_site_default when no SameSite is given (Cookie.zig:244), which turns on the
     Lax-allowing-unsafe window (Cookie.zig:442). A CDP cookie with no SameSite loses that window, so the
     "cookie set just before the IdP POST-back" login flow still behaves differently than it does in Chrome.
  4. SameSite=None without Secure is still accepted. Set-Cookie rejects it (Cookie.zig:167). Chrome's CDP
     path rejects it too, because building the canonical cookie fails. The CDP path here stores it as-is.
     This was already broken before the PR, but it's arguably the second real sameSite bug, and it isn't
     fixed.
  5. setCookies applies part of the batch. It returns on the first bad value after the earlier cookies are
     already in the jar. In Chrome nothing fails, so this situation never comes up there.
  6. A cookie-file behavior change nobody mentions. parseJsonSameSite in cookies.zig now turns a missing or
     unknown value into Lax instead of None. That's closer to correct, but it isn't in the PR description,
     and it now disagrees with the CDP path (point 1).

  On the lenient spellings: the issue's premise is wrong. CDP's CookieSameSite is just Strict | Lax | None.
  getCookies emits exactly those, both in Chrome and in our code, so the get→set round-trip already worked.
  no_restriction is the chrome.cookies extension API's spelling, and the PR's own test comment says so.
  Accepting it is harmless, but it isn't a fix. It's also not what Chrome does: there, "no_restriction" and
  "strict" both end up as Lax-by-default, not None and Strict. Rejecting unknown values outright is also
  stricter than Chrome.

  Bottom line: the PR only makes the spelling lenient. A real fix would:
  - Default a missing value to unspecified, meaning .lax with same_site_default = true, the same as
    Set-Cookie. Or match Chrome exactly by treating any unrecognised value that way too.
  - Reject None without Secure, the same way Set-Cookie does.
  - Validate the whole setCookies batch before adding anything.
  
  I haven't run the PR's tests; this is from reading the diff.
  ```

PR #3530, opened 2026-09-15
Stop load events on a document superseded by navigation

staylor:
## Why

When a script navigates away while its document is still loading, Lightpanda continues firing `DOMContentLoaded` and `load` on the superseded document. Clients waiting on those signals are told the old page is ready just as it is being replaced. In Chromium, that document still moves `readyState` to `complete`, but its `DOMContentLoaded` and `load` never fire, and a navigation-aware client resolves on the destination.

This reproduces on [Pia Mia Handmade Shoes](https://pia-mia-handmade-shoes.myshopify.com/products/basic-micaela-espadrille?_fd=0): an inline locale redirect calls `location.assign()` during parsing. Playwright's `page.goto(..., {waitUntil: 'domcontentloaded'})` resolves on the doomed document, and a following `page.evaluate` fails with `Execution context was destroyed`. A minimal loopback page with `<script>location.assign('/final')</script>` shows the same divergence.

## Changes

- Add `Document._load_aborted`, set when a cross-document navigation is scheduled and when a root navigation is initiated directly. It survives queue consumption and a discarded or failed replacement, matching Chromium, which does not revive the old document's load events if the destination fails.
- `documentIsLoaded`/`documentIsComplete` still dispatch `readystatechange` for the superseded document, but guard `DOMContentLoaded` and the document-completion sequence (queued element events, `window.load`, `pageshow` and meta-refresh). The check walks parent documents so a subframe of a navigating page also stops. This is not a blanket ban on independently queued element events; Chromium can still deliver those after navigation begins.
- Extract `Frame.cancelQueuedNavigation`, used by `stopLoading` and now `document.open()`. Once parsing has finished, `document.open()` can cancel a queued navigation; the rewritten document does not inherit the aborted load events.
- Preserve HTML's active-parser-was-aborted flag separately from load state. Navigation can change `readyState` to `complete` while the original parser is still on the stack. In that case, `document.open()`, `write()` and `writeln()` must remain no-ops, even after navigation is cancelled. This prevents a second parser from starting and crashing in `ScriptManagerBase.staticScriptsDone` when the original parser completes. Both queued and direct root navigation use `Frame.abortDocumentLoad` to set the flags.

Fragment navigations and `history.replaceState` are unaffected: they never schedule a cross-document navigation.

## Tests

- Event sequences for six trigger points, checked against Chromium: inline script, `interactive` readystatechange, `DOMContentLoaded`, `complete` readystatechange, fragment change, `history.replaceState`.
- Pending and discarded replacement pages do not resume the old document's load events.
- `document.open()` after a post-parse `location.assign()` cancels the navigation without firing `load` on the rewritten document.
- A `readystatechange` handler that renavigates and throws is handled safely.
- A real HTML response with inline `location.assign(); document.open(); document.write(); document.close()` navigates without restarting the parser. This regression test aborts the previous PR head with `SIGABRT` before the follow-up fix.
- Real `interactive` and `DOMContentLoaded` callbacks can still cancel navigation and rewrite the document once parsing has finished.
- Cancelling an inline navigation does not revive its aborted parser, including later calls to `open`, `write` and `writeln` after the original parse has unwound.

## Validation

Latest follow-up (`39dde6db7`):

- `make test`: **1531/1531** on macOS arm64, Debug with leak checking.
- CI formatting command and `git diff --check` pass; macOS Debug build succeeds.
- Chromium comparisons confirm matching behavior for inline navigation followed by `open/write/close`, and for cancelled inline navigation followed by later writes.

Initial implementation (`1a9557e41`):

- `make test`: **1528/1528** on macOS arm64 and on native Linux aarch64 (glibc 2.41, stock prebuilt V8).
- macOS ReleaseFast build validated.
- Loopback comparison against Chromium 152 across ten scenarios (the six trigger points above plus a delayed-iframe destination, a failed destination, `window.stop()`, and post-parse `document.open()` during navigation): the superseded document's event sequence matches in all ten.
- Storefront replay, 3 products: `goto` resolves on the localized destination and an extraction `evaluate` succeeds on each, where it previously failed with a destroyed context.

The parser follow-up has not been revalidated on Linux or in ReleaseFast.

Pre-existing and out of scope: a failed destination leaves the previous document displayed rather than an error page; `window.stop()` racing a synchronous navigation; and a post-parse `document.open()/close()` leaving `readyState` at `loading` rather than Chromium's `complete`.


karlseguin:
closed by a merged PR that was referencing the wrong issue, re-opened.

PR #3534, opened 2026-09-15
Implement WebAssembly.compileStreaming and instantiateStreaming

### LINKED ISSUE #385: handle WASM execution
Relates with #382

We want the browser be able to execute WASM scripts.

Example how modules have been handled: #345

arrufat:
Closes #385.

Both `WebAssembly.compileStreaming` and `WebAssembly.instantiateStreaming` were `undefined`, so every page that loads its wasm with `WebAssembly.instantiateStreaming(fetch('x.wasm'))` died on a `TypeError` before rendering anything. Two examples that now load: https://www.nicbarker.com/clay (the whole DOM is laid out by wasm) and https://arrufat.github.io/zignal/ (Zig autodoc, unpacks a 20 MB sources tarball).

```
lightpanda fetch https://www.nicbarker.com/clay --dump markdown --wait-until networkidle
lightpanda fetch https://arrufat.github.io/zignal/ --dump markdown --wait-until networkidle
```

(`--wait-until load` still shows the pre-wasm shell on these sites, as in Chrome: the fetches start after the load event.)

## Why they were missing

V8 installs the two functions in `WasmJs::Install` only when the isolate has a `WasmStreamingCallback`, and it does so at context genesis. So the callback has to be registered on the `SnapshotCreator` isolate as well as the runtime one; the pointer itself is not serialized.

## What this does

- `src/browser/js/WasmStreaming.zig`: the embedder half, modeled on Node's `wasm_web_api`. It resolves the `Response | Promise<Response>` argument, checks the `application/wasm` MIME essence, the status and the body usability, sets the source URL, drains the body through `Response.bytes` and feeds V8. Failures `Abort` the stream with a `TypeError`, or with the source promise's own rejection reason. The body is handed to V8 in one piece; feeding the `ReadableStream` chunk by chunk is a possible follow-up.
- `Env.zig` / `Snapshot.zig` register the callback on both isolates.
- `Context.zig` tracks in-flight streams and cancels them on teardown.
- `TestHTTPServer.zig` serves `.wasm` as `application/wasm` and `.bin` as `application/octet-stream`.
- `src/browser/tests/wasm/streaming.html` with a hand-assembled `add.wasm` (`add(i32, i32)`): instantiate, compile, a `Response` argument with MIME parameters, and rejections for a non-Response, wrong MIME, non-ok status, used body, rejected source and invalid module. It asserts error types, not messages, so it holds in real browsers too.

## Depends on

lightpanda-io/zig-v8-fork#208 (merged, released as v0.5.6) for the `v8__WasmStreaming__*` bindings. `main` already pins that release, so this branch builds against the published prebuilt archives.


src/browser/js/WasmStreaming.zig
```diff
@@ -0,0 +1,152 @@
...
+// Body accessors report a TypeError through the context's error_message,
+// the same way the bridge surfaces it to JS.
+fn abortZigError(self: *WasmStreaming, local: *const js.Local, err: anyerror) void {
+    switch (err) {
+        error.TypeError => {
+            const env = local.ctx.env;
+            self.abortTypeError(local, env.error_message orelse "Response body is not usable");
+            env.error_message = null;
+        },
+        else => {
```
karlseguin, line 130: I think you want to catch 3 magic errors here: `TryCatchRethrow`, `JsException` and `ExecutionTerminated` (same as Caller's `handleError`). I don't think any of them should re-enter v8. So, for all three, just `self.release()`

arrufat: Good catch, done in 6209893. All three now just `self.release()`, which drops our shared_ptr and de-registers the stream without touching V8, so the pending exception (or the termination) survives.

Worth noting these are reachable and not just defensive: a ReadableStream-backed body means `Response.bytes` runs the stream's JS `pull`/`read`, so a throwing or killed source lands right here.

PR #3542, opened 2026-09-16
cdp: window.devicePixelRatio from viewport scale and Input char text insertion

arrufat:
## Summary

This PR addresses two compatibility deviations when driving Lightpanda over CDP:

1. **`window.devicePixelRatio` now reflects `Emulation.setDeviceMetricsOverride`**:
   - `Emulation.setDeviceMetricsOverride` already stores `deviceScaleFactor` into `Viewport.scale`, but `window.devicePixelRatio` was a static constant `1`.
   - Turned `devicePixelRatio` into an accessor mirroring `innerWidth`/`innerHeight`:
     - **Getter**: reads `frame.page.getViewport().scale` (defaults to `1.0`).
     - **Setter**: preserves `[Replaceable]` semantics by delegating to `replaceGlobalProperty(value, "devicePixelRatio")`, so scripts assigning to `devicePixelRatio` keep working.
   - Ensures consistency for device emulation, `srcset 2x`, and resolution-dependent logic.

2. **`Input.dispatchKeyEvent` `type: "char"` text insertion**:
   - `Input.dispatchKeyEvent` with `type: "char"` previously fired a `keypress` event but did not insert text into the active element.
   - Now inserts `params.text` into the focused element via `Frame.user_input.insertText(frame, text)` when not default-prevented by a `keypress` listener (`!keyboard_event.asEvent().getDefaultPrevented()`).
   - If no `text` is provided (e.g. key `Enter`), the value remains untouched.

3. **`Input.dispatchKeyEvent` text-only `keyDown` fallback**:
   - For `keyDown` events where a client sends `text` without specifying `key`, falls back to using `text` as the key string, ensuring the printable character is recognized and inserted into the active element.

## Testing

- Added CDP test coverage in `src/server/cdp/domains/emulation.zig` verifying `devicePixelRatio` updates to `2` under `deviceScaleFactor: 2`, preserves across zero-overrides, and resets to `1` on `clearDeviceMetricsOverride`.
- Added CDP test coverage in `src/server/cdp/domains/input.zig` testing:
  - `char` with `text: "h"` fires `keypress` and inserts `"h"`.
  - `preventDefault()` on `keypress` blocks text insertion.
  - `char` with key `Enter` and no `text` leaves the value untouched.
  - `keyDown` with only `text: "z"` falls back to `key_str`, fires `keypress`, and inserts `"z"`.
- `make test`: 1547/1547 tests pass with no leaks.
- `zig fmt --check ./*.zig ./**/*.zig`: Clean.


src/server/cdp/domains/input.zig
```diff
@@ -61,22 +62,30 @@ fn dispatchKeyEvent(cmd: *CDP.Command) !void {
...
     }, .{
-        .key = params.key,
+        .key = key_str,
         .code = params.code,
         .altKey = params.modifiers & 1 == 1,
         .ctrlKey = params.modifiers & 2 == 2,
         .metaKey = params.modifiers & 4 == 4,
         .shiftKey = params.modifiers & 8 == 8,
     }, frame);
     try Frame.user_input.triggerKeyboard(frame, keyboard_event);
```
karlseguin: `EventManager.dispatch` frees the event if it isn't actually dispatched. It's meant as a convenience for the common case where we don't use the event after dispatching. The downside is that when we _do_ use it after dispatching, as we do here, we need an acquireRef/releaseRef.

See the end of `Frame.submitForm`:

```zig
        submit_event.acquireRef();
        defer _ = submit_event.releaseRef(self.page);

        try self._event_manager.dispatch(form_element.asEventTarget(), submit_event);
        // If the submit event was prevented, don't submit the form
        if (submit_event._prevent_default) {  <-- need to acquire/release because we read here.
            return;
        }
```

karlseguin (commented):
I wrote a new test case: https://github.com/lightpanda-io/demo/pull/251

It passes in chrome, but both things being tested fail on this PR.

1- I believe it's conflicting with the (possibly wrong) keydown special handling we have in EventManager.dispatchNode

2 - The new char handling needs to call `allowEdit` to trigger the beforeinput and textinput events, which can set prevent_default.

arrufat:
Thanks, the demo test made this clear. I went with the keydown split you hinted at.

- A keydown now only runs its own default action: Tab, caret moves, Backspace, Enter activation. Text is typed by a separate char step, `user_input.typeChar`, which fires keypress and then goes through `allowEdit`, so both keypress and beforeinput can veto it.
- A keyDown that carries `text` runs the char step inline (Puppeteer, Playwright). A text-less keyDown followed by a `char` message runs it on the `char` (chromedp). Each character is typed once either way. A cancelled text-less keyDown drops the next `char`, as Chrome does.
- BiDi, WebDriver and the MCP press action all go through the same `pressKey`, deriving the text from the key, so their behavior is unchanged.
- On the acquireRef point: the CDP handler no longer reads the event after dispatch. `pressKey` holds the ref for the keypress it builds from the keydown.
- `Input.insertText` now fires beforeinput too.

Both cases from your demo test are covered by unit tests in `cdp/domains/input.zig`.

karlseguin:
I updated the demo branch with more cases which pass in Chrome but fail in the PR (I didn't test it, but I think these cases would pass in `main`). There's a double-submit going on.

arrufat:
Found it, thanks for the extra cases.

Enter on a button was activated on the keydown, so the click came before the keypress. Then the char step, which submits for an `<input>` on Enter, submitted again for button-type inputs. That was the double submit.

Now a button clicks on the keypress Enter produces, after the keypress fires, and the form submits once. Only a link still follows Enter on the keydown, which matches Chrome. The MCP press action had its own implicit submission on top of the shared path, so I removed it.

I ran the updated demo branch against this, and the same six cases are in `cdp/domains/input.zig`.

PR #3546, opened 2026-09-17
Robot Limit Entries

mookums:
This adds an EvictionQueue which is generic over a Key type. This is basically a min heap that allows us to keep track and quickly return the oldest entry that we currently have. There are also some changes to RobotStore (such as returning an owned Entry) and checking the Robot entry under the lock.


src/network/EvictionQueue.zig
```diff
@@ -0,0 +1,99 @@
...
+        fn olderFirst(_: void, a: Item, b: Item) std.math.Order {
+            return std.math.order(a.seq, b.seq);
+        }
+
+        const Heap = std.PriorityQueue(Item, void, olderFirst);
+
+        allocator: Allocator,
+        capacity: usize,
+        next_seq: u64 = 0,
+        heap: Heap = .empty,
```
karlseguin: The priority just increments, i.e. this is a FIFO. I think `std.Deque` may be a better choice.

src/network/EvictionQueue.zig
```diff
@@ -0,0 +1,99 @@
...
+
+        pub fn init(allocator: Allocator, capacity: usize) Self {
+            return .{ .allocator = allocator, .capacity = capacity };
+        }
+
+        pub fn deinit(self: *Self) void {
+            self.heap.deinit(self.allocator);
+        }
+
+        pub fn count(self: *const Self) usize {
```
karlseguin: only used within this file for tests.

src/Config.zig
```diff
@@ -245,14 +245,14 @@ pub const LoadResources = packed struct(u4) {
...
-pub const ExperimentalFeatures = packed struct(u2) {
+const ExperimentalFeatures = packed struct(u1) {
     cors: bool = false,
-    serviceworker: bool = false,
 };
 
 /// Common CLI args.
 const CommonOptions = .{
     .{ .name = "obey_robots", .type = bool },
+    .{ .name = "robot_store_entry_limit", .type = ?u32, .default = 1000 },
```
karlseguin: Got a crash when I set this to 0 and then tried to fetch with  `--obey-robots`.

src/network/Robots.zig
```diff
@@ -100,10 +99,19 @@ pub const RobotStore = struct {
...
+    evictions: EvictionQueue([]const u8),
     mutex: std.Io.Mutex = .init,
 
-    pub fn init(allocator: std.mem.Allocator) RobotStore {
-        return .{ .allocator = allocator, .map = .empty };
+    pub fn init(allocator: std.mem.Allocator, capacity: u32) RobotStore {
+        return .initCapacity(allocator, capacity);
+    }
+
+    pub fn initCapacity(allocator: std.mem.Allocator, capacity: usize) RobotStore {
```
karlseguin: The only caller of this is `init`. I don't understand why this is split out. Also, from the std lib, I expect `initCapacity` to reserve the capacity. This doesn't.

src/Config.zig
mookums: Latest forces a minimum of 1. I don't mind this but if we want to have a fallback to unbounded on 0 (like the http cache), I can do that too.

src/network/RobotsGate.zig
```diff
@@ -141,17 +134,11 @@ fn flushPending(self: *RobotsGate, robots_url: []const u8) void {
...
         transfer.unpark();
 
-        const allowed = if (robot_entry) |entry| switch (entry) {
-            .allowed => true,
-            .disallowed => false,
-            .present => |robots| robots.isAllowed(URL.getPathname(transfer.req.url)),
-        } else true;
-
-        if (!allowed) {
+        const decision = self.network.robot_store.checkPath(robots_url, URL.getPathname(transfer.req.url));
```
karlseguin: With a small cache, `robots_url` could be evicted by the time you check, so it would be null and thus a blocked URL wouldn't actually block. It's unlikely, but the fix is pretty simple since the caller knows what the action is.

karlseguin (commented):
The Robots store (Robots.zig) has a `url => RobotsEntry` `map`.  It now gains a `url => LruCache.Entry` `evictions`.  The flow is:

1. when we have spare capacity
put into `map` and put into `evictions`

2. when we have no spare capacity
put into `map` and put into `evictions`, get oldest (from linked list), delete from `evictions`, delete from `map`.

`map` and `evictions` are kept in sync, so why have both?

Also, Zig's HashMap doesn't rehash on deletes (1). Once full, it won't be long before the map is full of tombstones and every lookup becomes a scan over the internal slots. An ArrayHashMap is an alternative, and if you switch form LRU to clock, you no longer need the linked list.

mookums:
Using a ClockCache now backed by the ArrayHashMaps. I also have it so we no longer check the store for the flushPending but use the resolved outcome from `doneCallback`.

karlseguin:
That's good, but I meant that `map: RobotsMap`  should become `map: Cache(RobotsEntry)`  it isn't the key that should be generic (it could be, but I don't see why we'd ever have anything but a `[]const u8`,  it's the _value_). Then the lookup in Cache is  `std.array_hash_map.String(Entry(V))`, where `Entry` is:

```zig
fn Entry(comptime V: type) type {
  return struct {
     value: V,
     // any other cache-specific fields you ned
  }
}
```

The cache _is_ the map.

src/network/RobotsGate.zig
```diff
@@ -200,67 +194,63 @@ const RobotsContext = struct {
...
-        // instead — resolve() must run exactly once.
-        self.resolve();
     }
 
     fn errorCallback(ctx_ptr: *anyopaque, err: anyerror) void {
         const self: *RobotsContext = @ptrCast(@alignCast(ctx_ptr));
 
         log.warn(.http, "robots fetch failed", .{ .err = err });
-        self.resolve();
+        self.settle(.{ .decision = .allowed });
```
karlseguin: Is this right? I think main would allow this transfer, and any waiting for it in the single_flight, but it wouldn't *store* anything. So a subsequent entry would try again.

src/network/RobotsGate.zig
```diff
@@ -134,24 +126,27 @@ fn fetchThenResume(self: *RobotsGate, robots_url: [:0]const u8, transfer: *Trans
...
+const Outcome = union(enum) {
+    decision: Robots.RobotStore.Decision,
+    robots: Robots.Robots,
+};
+
 // The robots.txt fetch resolved: hand every waiter back to the pipeline,
 // each judged against its own path. No store entry (fetch failed, or a 200
 // whose body never got parsed) fails open.
-fn flushPending(self: *RobotsGate, robots_url: []const u8) void {
+fn flushPending(self: *RobotsGate, robots_url: []const u8, outcome: Outcome) void {
```
karlseguin, line 137: This goes to my other comment about `errorCallback` storing `.allowed`, the comment says "No store entry" - but there is no "No store entry" now. I think the comment is right and the code should follow it.

src/network/Robots.zig
```diff
@@ -139,69 +145,64 @@ pub const RobotStore = struct {
...
-        if (try self.putKey(url)) |value_ptr| {
-            // first time seeing this url, store the value
-            value_ptr.* = .{ .present = robots };
-            return;
-        }
+        if (try self.insert(url, .{ .present = robots })) return;
 
         // cannot overwrite an existing value, if it was `present`, we'd have
         // to free the value but it might be being used.
-
```
karlseguin, line 150: comment about this is no longer relevant...nothing holds a map value outside of the mutex anymore (the old `get` is gone).

src/Config.zig
```diff
@@ -233,6 +233,28 @@ fn caPathValidator(
...
+    const value = std.fmt.parseInt(u32, str, 10) catch {
+        log.fatal(.app, "invalid option value", .{
+            .arg = "--robot-store-entry-limit",
+            .value = str,
+            .hint = "must be a positive integer",
+        });
+        return error.InvalidArgument;
+    };
+
+    if (value == 0) {
```
karlseguin: I think _every_ other numeric option, 0 == no limit  (e.g. `--http-cache-entry-limit`)

PR #3547, opened 2026-09-17
`Navigation.reload`

mookums:
Not sure when reload got unwired from Navigation but this just readds it. It gets us passing a few more WPT tests related to the basic functioning of `reload`.


karlseguin:
`/navigation-api/precommit-handler/precommitHandler-redirect-throws.html`  seems to endlessly loop on this branch, but not main (it reports a timeout and notrun, whereas main doesn't).

src/browser/webapi/navigation/Navigation.zig
```diff
@@ -480,18 +538,11 @@ fn updateCurrentEntry(self: *Navigation, options: UpdateCurrentEntryOptions, fra
...
-            frame,
-        )).asEvent();
-        try self.dispatch(cec, event, frame);
-    }
+    self.fireCurrentEntryChangeEvent(previous, null, frame);
 }
 
-pub fn dispatch(self: *Navigation, func: js.Function.Global, event: *Event, frame: *Frame) !void {
-    return frame._event_manager.dispatchDirect(
+pub fn dispatch(self: *Navigation, func: ?js.Function.Global, event: *Event, frame: *Frame) !void {
```
karlseguin, line 540: As far as I can tell, this doesn't have to be public.

src/browser/webapi/navigation/Navigation.zig
```diff
@@ -292,19 +306,77 @@ pub fn replaceEntry(
...
+        .wrap("navigatesuccess"),
+        null,
+        frame.page,
+    ) catch |err| {
+        log.warn(.event, "Navigation.fireNavigateSuccess", .{ .err = err });
+        return;
+    };
+
+    self.dispatch(self._on_navigatesuccess, event, frame) catch |err| {
+        log.warn(.event, "Navigation.fireNavigateSuccess dispatch", .{ .err = err });
```
karlseguin: Are you sure that's the right thing to do for (example) `error.ExecutionTerminated`?  From what I can tell, every caller of these APIs is a JS API, and `Caller` knows how best to handle errors so you're almost always better of bubbling it up. (same in 3 places).

mookums:
We don't really support the precommit handlers right now, I believe they're a part of NavigateEvent (and they aren't baseline available either). 

That test that times out is supposed to try to navigate/redirect within a NaviagetEvent precommitHandler and throw a `InvalidStateError` error. Since we don't support that with this PR, it results in it just timing out instead of the previous error which was `Unhandled rejection with value: object \"TypeError: navigation.reload is not a function`

PR #3548, opened 2026-09-17
webapi: img srcset support

karlseguin:
The image src can now come from a srcset, including the source of a picture. The (simple) Parser is all Claude.

There's some similarity here with the recently changed Select<->Option (1) code in that, changes to a child (source/option) need impact their parent ( picture/select).

(1) https://github.com/lightpanda-io/browser/pull/3502


krichprollsch (commented):
Claude thinks the code `picture.innerHTML = '<source srcset="a.png">'` could result in a `_selected_hash` for b.png while `currentSrc` says a.png

PR #3549, opened 2026-09-17
internal: Always correct SemanticTree context

karlseguin:
Applies the frame-ownership pass to SemanticTree, copying what we did for StyleManager (1). SemanticTree doesn't visit iframes, so the frame of the root is the frame/frame._style_manager we need to target for all visited nodes.

Like #3536, it's up to the callers to (a) get the correct frame and (b) decide what to do on a frameless-node.

(1) https://github.com/lightpanda-io/browser/pull/3536


arrufat:
`LP.getInteractiveElements` in the same file has the same frame mismatch for a `nodeId` root inside a child frame: it still passes `bc.mainFrame()` to `collectInteractiveElements`.

The style checks are fine, since `isVisible`, `hasPointerEventsNone` and the display check in `walkInteractive` all resolve `ownerFrame` per element. Two things do go wrong though:

- `buildListenerTargetMap` reads the main frame's event manager and inline handler table. Listeners registered by a script inside the iframe live in the child frame's event manager, so a listener-only element in the iframe is reported as non-interactive.
- `href` is resolved against `frame.base()`, so relative links inside the iframe resolve against the parent's URL.

Pre-existing and out of scope for this PR, so fine as a follow-up. The fix is the same as here: resolve `root.ownerFrame(frame)` before the walk and pass that.

arrufat (approved):

PR #3555, opened 2026-09-17
ci: dispatch the lightpanda-js npm build on version tags

arrufat:
Adds an `update-js-package` job to the release workflow, a clone of `update-python-package` (#3390, #3528): on version-shaped tags (never `nightly`), it dispatches the `npm.yml` workflow on `lightpanda-io/lightpanda-js` with the distribution app token, the same permission the python, homebrew and docker jobs use.

That workflow downloads this release's four binaries, packs the `lightpanda` package and its four `@lightpanda/<platform>` optional dependencies, tests them on native runners plus Deno and Bun, publishes through npm trusted publishing once a maintainer approves the `npm` environment there, and then records the matching release on that repo itself. Nothing reaches npm without the approval.

The version-shape check and the "already released" guard are the same as the python job, so re-running after a success does not re-dispatch, and it runs after `build-linux` + `build-macos` so all four binaries are on the release first.

Prerequisites, both outside this PR:
- the `lightpanda-browser-distribution` app needs access to `lightpanda-js`, as it was granted for `lightpanda-python`;
- the five npm packages need a manual first publish, since trusted publishing is configured per existing package on npmjs.com.


arrufat:
@krichprollsch could an org owner add `lightpanda-js` to the `lightpanda-browser-distribution` app? Org settings, GitHub Apps, Configure on the app, then add the repo under "Only select repositories". Same setup as `lightpanda-python`, no permission changes. It is a prerequisite for this PR: without it the token step fails on the first version tag.

PR #3556, opened 2026-09-17
cdp: announce the committed main world at commit, name the destroyed one

arrufat:
## Problem

On a committed root navigation, Lightpanda cleared the V8 inspector context group at response-header time (`frame_remove` → `resetContextGroup`) but only announced the new main world later, inside `frameNavigated`, after the body had downloaded and parsed. Between those two points there is no announced execution context.

A CDP driver that evaluates in that window gets `-32000 Execution context was destroyed`. It also emitted a second, redundant `Runtime.executionContextsCleared` and never named the specific outgoing context as destroyed, so a driver tracking context ids by `Runtime.executionContextDestroyed` (Stagehand v3's `ExecutionContextRegistry`, for one) never learned the old id was gone.

Chrome, for a same-target cross-document navigation, names the outgoing context destroyed and creates the new one back to back, with no unnamed gap.

## Change

- `frame_remove` now emits `Runtime.executionContextDestroyed{executionContextId}` for the outgoing default context before it resets the group.
- The committed replacement's main world is announced at `frame_created`, right after the group is cleared, so an evaluate that arrives while the body is still in flight finds a context.
- `frameNavigated` no longer clears or re-announces the main world for that committed case (an in-place root navigation — the first navigation of a pristine `about:blank`, which keeps its Frame — still clears and re-announces, exactly once).

Adds a `cdp.frame` test asserting the order `executionContextDestroyed → executionContextsCleared → Page.frameNavigated → executionContextCreated` with no evaluate-able gap. The existing frame/page/runtime tests are unchanged.

## Scope

Verified against the integrations matrix (https://github.com/lightpanda-io/integrations): this flips Stagehand v3's `fill` from ❌ to ✅ and does not regress Playwright, Puppeteer or chromedp. It is a partial step: `select`, `press` and `upload` still fail because the driver's in-flight `evaluate` targets the outgoing context id, and a pending `awaitPromise` evaluate blocks the loop so the navigation is deferred. Closing those needs deferring the old context's teardown until in-flight evaluations resolve (or pumping navigation during `awaitPromise`), tracked separately.


karlseguin:
I'm always weary of changes here. From past experience, changes like this can fix one driver but break another.  I ran our puppeteer/click.js in main/branch/chrome and, from what I can tell, `main` is closer to chrome than this branch. The relevant parts:

```
MAIN:
EVT     Runtime.executionContextsCleared {} [SID-1]
RES  52 Page.navigate {"frameId": "FID-0000000001", "loaderId": "LID-0000000003"} [SID-1]
EVT     Page.lifecycleEvent init [SID-1]
EVT     Runtime.executionContextsCleared {} [SID-1]
EVT     Page.frameNavigated url=http://127.0.0.1:1234/ loader=LID-0000000003 [SID-1]
EVT     Runtime.executionContextCreated id=7 isDefault=True type=default name='' [SID-1]
EVT     Runtime.executionContextCreated id=8 isDefault=False type=isolated 

CHROME:
RES  74 Page.navigate {"frameId": "AEA73ACFBC05E064B618427084DCAB8B", "loaderId": "64D0DA99A [822D72]
EVT     Runtime.executionContextsCleared {} [822D72]
EVT     Page.lifecycleEvent init [822D72]
EVT     Runtime.executionContextsCleared {} [822D72]
EVT     Page.frameNavigated url=http://127.0.0.1:1234/ loader=64D0DA99A52C5AC3EA3CF986C83E2730 [822D72]
EVT     Target.targetInfoChanged {"targetInfo": {"targetId": "AEA73ACFBC05E064B618427084DCAB8 []
EVT     Runtime.executionContextCreated id=7 isDefault=True type=default name='' [822D72]
EVT     Runtime.executionContextCreated id=8 isDefault=False type=isolated 

BRANCH:
EVT     Runtime.executionContextDestroyed id=5 [SID-1]
EVT     Runtime.executionContextsCleared {} [SID-1]
EVT     Runtime.executionContextCreated id=7 isDefault=True type=default name='' [SID-1]
RES  52 Page.navigate {"frameId": "FID-0000000001", "loaderId": "LID-0000000003"} [SID-1]
EVT     Page.lifecycleEvent init [SID-1]
EVT     Page.frameNavigated url=http://127.0.0.1:1234/ loader=LID-0000000003 [SID-1]
EVT     Runtime.executionContextCreated id=8 isDefault=False type=isolated 
```

Chrome/Main never send `executionContextDestroyed` and they both send a redundant `executionContextsCleared`. Finally, they both agree that `executionContextCreated` comes after the `navigate`, not before.

I then threw Claude at this, and it thinks this is a timing bug in Stagehand. I found https://github.com/browserbase/stagehand/issues/1778 which is I think is tangentially. Claude thinks this issue was fixed by retrying for a "-32000 Cannot find context with specified id" and suggested that the fix for our issue is to also retry on a "-32000 Execution context was destroyed". But I haven't dug into it.

arrufat:
You're right, and I should have checked the later navigations before claiming Chrome names the destroyed context. Closing this.

I re-ran `puppeteer/click.js` against Chrome 153 with full protocol logging. Across all seven navigations, including the anchor clicks from a loaded document, Chrome never sends `Runtime.executionContextDestroyed`, sends `executionContextsCleared` twice, and sends `executionContextCreated` after both the `Page.navigate` response and `Page.frameNavigated`. Main matches that; this branch doesn't.

The Stagehand claim doesn't hold either. Running the integrations Stagehand suite against binaries from `main` and this branch gives identical results, and `fill` flips between pass and fail on *both* depending on a few ms of latency (it passes on both when a logging proxy sits in between). The failures are the timing bug you suspected: Stagehand's `waitForSelector` is one long-lived `Runtime.evaluate` with `awaitPromise`, sent a millisecond after the click that submits the form and scoped to the old context id. If the commit lands after it, the commit kills the pending evaluate with `Execution context was destroyed`, which Stagehand doesn't retry (it only retries `Cannot find context with specified id`). Chrome has the same race: with a 100 ms delay on the form endpoint, Chrome fails `fill`, `press` and `upload` with `-32000 Inspected target navigated or closed`. That's the fix to send upstream (retry a wait, never a click, on that error class), same family as browserbase/stagehand#1778.

The announce gap on `main` (cleared at headers, created after the body is parsed) is real but I have no driver failure it causes: Stagehand's registry already wipes on `executionContextsCleared` and waits for the next created event, and with a 1 s body delay `main` and this branch fail identically. Without a motivating failure, matching Chrome's event shape wins.

PR #3558, opened 2026-09-17
cdp: DOM.getFrameOwner returns the owner <iframe>, nodes carry frameId

arrufat:
`DOM.getFrameOwner` answered with the child frame's document instead of the element hosting it. Clients that map an `<iframe>` to its frame compare the `backendNodeId` it returns with the one they resolved in the parent (Stagehand's `frameLocator`/`deepLocator`, Playwright's `contentFrame`), so the lookup never matched and they could not descend into frames.

- Return the owner element, as Chrome does, and error for the main frame.
- The node writer emits `frameId` on frame-owner elements and on the document element (the `<html>` node, as Chrome does), the other half of that pairing.

Against the integrations fixture (`iframe.html` hosting `frame.html`), before: the iframe element is backendNodeId 14, `getFrameOwner` returns 16 (a `#document`), the iframe node has no `frameId`. After: `getFrameOwner` returns 14 and the node carries `frameId: FID-0000000002`. Stagehand 3.7.3 `page.frameLocator('#child').locator('#child-text').textContent()` and `page.deepLocator('#child >> #child-text')` both resolve; they failed with "Unable to obtain a content frame" before. Playwright and Puppeteer `iframe`, `click`, `goto`, `eval` still pass.

First two sub-items of "Iframe DOM descent" in the integrations gaps list.


src/server/cdp/Node.zig
```diff
@@ -217,9 +219,27 @@ pub const Writer = struct {
...
+                        try w.objectField("frameId");
+                        try w.write(&id.toFrameId(child._frame_id));
+                    }
+                }
+            }
         } else {
             try w.objectField("localName");
             try w.write("");
+
+            if (dom_node._type == .document) {
```
karlseguin: Claude thinks this should go on the child of the document and gave me this from Chrome which does it on the node whose parent is the document

```cpp
  // inspector_dom_agent.cc:2436, inside the `if (auto* element = DynamicTo<Element>(node))` block
  if (node->parentNode() && node->parentNode()->IsDocumentNode()) {
    LocalFrame* frame = node->GetDocument().GetFrame();
    if (frame) value->setFrameId(IdentifiersFactory::FrameId(frame));
  }
```

So you'd end up with:

```zig
  if (dom_node.is(DOMNode.Element)) |element| {
      // ... attributes, localName ...

      if (element.is(IFrame)) |iframe| {
          if (iframe.getContentDocument()) |document| {
              if (document._frame) |child| {
                  try w.objectField("frameId");
                  try w.write(&id.toFrameId(child._frame_id));
              }
          }
      } else if (dom_node._parent) |dom_parent| {
          if (dom_parent._type == .document) {
              if (dom_parent.subtype(DOMNode.Document)._frame) |frame| {
                  try w.objectField("frameId");
                  try w.write(&id.toFrameId(frame._frame_id));
              }
          }
      }
  } else {
      try w.objectField("localName");
      try w.write("");
  }
```

but I haven't tested it through actual CDP.

arrufat: You're right. I checked Chromium main: the two setFrameId calls are both inside the Element block, one for frame owners and one for elements whose parent is the document. The document node never gets one. Applied your version in e8b02db, the test now checks the child frame's <html> element instead of its #document.

PR #3567, opened 2026-09-18
add `navigationsuccess` and `navigationerror`

mookums:
This adds the setters and getters for `navigationsuccess` and `navigationerror`. It also fires `navigationsuccess` right after resolving the finished promise.


src/browser/webapi/navigation/Navigation.zig
```diff
@@ -293,6 +302,48 @@ pub fn replaceEntry(
...
+            return;
+        };
+
+        self.dispatch(ons, event, frame) catch |err| {
+            log.warn(.event, "Navigation.fireNavigateSuccess dispatch", .{ .err = err });
+        };
+    }
+}
+
+fn fireNavigateError(self: *Navigation, reason: js.Value, frame: *Frame) void {
```
karlseguin, line 322: Dead code?

src/browser/webapi/navigation/Navigation.zig
```diff
@@ -293,6 +302,48 @@ pub fn replaceEntry(
     return entry;
 }
 
+fn fireNavigateSuccess(self: *Navigation, frame: *Frame) void {
```
karlseguin, line 305: What if there's an addEventListener? There's a frame.hasDirectListeners for this case, which will look at both the attribute (_on_navigateerror) and the attached listeners.

karlseguin (approved):

PR #3569, opened 2026-09-18
feat(input): support single-contact touch events (Web APIs, CDP, WebDriver)

R4m1r0qu41:
## Scope

Lightpanda's [Senior Software Engineer, System Programming](https://lightpanda.io/jobs/senior-software-engineer_system-programming) role scopes the work as implementing Web APIs and the Chrome DevTools Protocol in Zig, and improving the browser's quality and stability as part of the engine team. This PR is a contribution against that scope: a spec-compliant touch input path across the Web API surface, CDP, and WebDriver.

## Gap

`TouchEvent` existed but was a stub — `src/browser/webapi/event/TouchEvent.zig` returned empty lists for `touches`, `targetTouches`, and `changedTouches` unconditionally ("There is no touch input source: the touch lists are always empty."). There was no `Touch` or `TouchList` type, no `Input.dispatchTouchEvent` handler in the CDP `input` domain (only `dispatchKeyEvent`, `dispatchMouseEvent`, `insertText` existed), and WebDriver's `actionSequence` had no touch-specific dispatch path. Any test or automation relying on real touch events — `touchstart`/`touchmove`/`touchend`/`touchcancel` with populated `Touch` data — had nothing to hook into.

## Fix

- **Web APIs**: added `Touch.zig` and `TouchList.zig` (single-touch scope: at most one contact, identified by the client-chosen id, default `0`), and wired `TouchEvent` to hold a real `Touch` value and populate `touches`/`targetTouches`/`changedTouches` correctly (`touchend`/`touchcancel` drop the contact from `touches`/`targetTouches` but keep it in `changedTouches`, per spec). `Touch` exposes `identifier`, `target`, `clientX/Y`, `pageX/Y` and `screenX/Y` (aliased to client coordinates, since the engine has no scroll offset), and `radiusX/Y`, `rotationAngle` and `force` at Chrome's synthetic-point defaults (1, 1, 0, 1). `touches !== targetTouches` on read, matching `[SameObject]` semantics in real browsers.
- **CDP**: `src/server/cdp/domains/input.zig` gained `dispatchTouchEvent`, validating touch-point counts per event type and rejecting anything outside the single-contact model:
  - `touchStart`/`touchMove` take exactly one point; `touchEnd` takes zero or one (Playwright sends an empty list, Puppeteer sends the released point, and Chrome dispatches at that position); `touchCancel` must be empty, as in Chrome, and a populated one is rejected before the contact changes.
  - The contact keeps the id from `touchStart` (Puppeteer counts up from 1). Later points must carry the same id or get `InvalidParams`, and `Touch.identifier` reports it.
  - A second concurrent `touchStart`, or a `touchMove`/`touchEnd`/`touchCancel` with no active contact, is `InvalidParams`.
  - A `touchStart` that misses every element falls back to the document element, so the contact is still recorded and the following end is valid.
  - The client's `modifiers` are carried onto the touch events.

  `Page.input_touch_contact` (`src/browser/Page.zig`) tracks the active contact so `touchmove`/`touchend`/`touchcancel` resolve against the element hit-tested at `touchstart`, instead of re-hit-testing or defaulting to `(0, 0)`. Touch dispatch goes through a `TouchType` enum.
- **WebDriver**: `actionSequence`'s touch pointer source now dispatches through the same `dispatchTouchEventOn` helper as CDP, and forwards its modifier state. Fixed along the way:
  - `touchmove`/`touchend` stay pinned to the `touchstart` target (`Touch.target` is fixed at `touchstart` per spec; the code was previously re-resolving it via `elementFromPoint` at the current pointer position).
  - Pointer events for a touch source stay on the `pointerdown` element while pressed (implicit capture), and report `pointerType: "touch"` (previously hard-coded to `"mouse"`).
  - An unmatched or duplicate `pointerUp` (no preceding `pointerDown` in that action source) is a no-op for every pointer type instead of fabricating `pointerup`/`touchend`/`mouseup`/`click`/`dblclick`.
- **Retargeting**: `EventManager`'s shadow-DOM retargeting now covers `Touch.target` the same way it already covers `relatedTarget`, so a touch crossing a shadow boundary retargets to the host and resets correctly.
- **Spec details**: `touchcancel` is never cancelable; `touchstart`/`touchmove` follow the same passive-listener-dependent cancelable rule as `wheel`.

## Known gaps (not in this PR)

Tracked in #3584:

- **PointerEvents and compatibility mouse events for CDP touch.** CDP-driven touch dispatches TouchEvents only. Chrome also emits `pointerdown`/`pointerup` and the compatibility mouse/click events for the same tap, so a Puppeteer `page.tap()` fires touch listeners but never clicks. This needs a design for the CDP touch-to-PointerEvent path and is intended as a separate PR.
- **`Touch` and `TouchEvent` constructors.** The JS constructors ignore the `touches`/`targetTouches`/`changedTouches` init members, and `Touch` has no constructor, so `new Touch(...)` throws and feature-detection libraries will conclude touch is unsupported. Follow-up.

## Test plan

- [x] `ZIGFLAGS='-Dwpt_extensions' make test`: 1589 passed, 5 skipped, no regressions, no leaks (the custom runner fails a test that allocates without freeing)
- [x] `zig fmt --check ./*.zig ./**/*.zig`: clean
- [x] New coverage: CDP `dispatchTouchEvent` (population, modifiers, lift semantics with empty and populated `touchEnd`, populated `touchCancel` rejection, client id tracking and mismatch rejection, miss-fallback, validation/state-machine rejections, shadow retargeting, cancelable rules, bounded-arena reads on repeated property access), WebDriver `actionSequence` touch (population, target pinning and pointer capture across move/end, unmatched/duplicate release no-ops for touch and mouse)
- [x] Real-client checks: the same Puppeteer and Playwright scenarios against Lightpanda and Chrome (taps, `TouchHandle` start/move/end, populated and empty `touchEnd`/`touchCancel`, target retention across movement, recovery after a tap outside the page); all contract checks pass on Lightpanda


src/server/cdp/domains/input.zig
```diff
@@ -153,6 +156,71 @@ fn dispatchMouseEvent(cmd: *CDP.Command) !void {
...
+            touchStart,
+            touchMove,
+            touchEnd,
+            touchCancel,
+        };
+    }) catch return error.InvalidParams) orelse return error.InvalidParams;
+
+    switch (params.type) {
+        .touchStart, .touchMove => if (params.touchPoints.len != 1) return error.InvalidParams,
+        .touchEnd, .touchCancel => if (params.touchPoints.len != 0) return error.InvalidParams,
```
arrufat, line 187: Puppeteer always sends the released point in `touchPoints` on `touchEnd` (puppeteer-core `cdp/Input.js`, `CdpTouchHandle.end()`), so `page.tap()` and `touchscreen.tap()` fail here with InvalidParams. Playwright sends an empty list. Chrome accepts both because the touchPoints semantics are state based. touchEnd and touchCancel should accept 0 or 1 points.

The contact also stays active after the rejection, so the next `touchStart` fails too. The test at the bottom of this file cements the rejecting behavior and should be flipped.

src/server/cdp/domains/input.zig
```diff
@@ -153,6 +156,71 @@ fn dispatchMouseEvent(cmd: *CDP.Command) !void {
...
+    }
+    for (params.touchPoints) |point| {
+        if (point.id != 0) return error.InvalidParams;
```
arrufat, line 190: Puppeteer's id generator returns `++id` starting from 0, so its first contact has id 1 and it increments per tap. With this check no Puppeteer touch command ever gets past validation.

Suggestion: store the client's id on touchStart and require the same id on touchMove and touchEnd, instead of hard-coding 0. `Touch.identifier` can then report the client id.

src/browser/frame/user_input.zig
```diff
@@ -504,6 +505,80 @@ fn deltaToScroll(d: f64) i32 {
...
+
+/// CDP path: hit-test on touchstart, then keep that target for move/end/cancel.
+pub fn triggerTouch(frame: *Frame, typ: []const u8, x: f64, y: f64) !void {
+    const page = frame.page;
+    const is_start = std.mem.eql(u8, typ, "touchstart");
+    const target = if (!is_start)
+        (if (page.input_touch_contact) |c| c.target else null)
+    else
+        null;
+    const resolved = target orelse (try frame.window._document.elementFromPoint(x, y, frame)) orelse return;
```
arrufat, line 557: When `elementFromPoint` misses on touchstart (faux layout returns null for any y past the last element), this returns silently. CDP already answered with success, but no contact is recorded, so the client's following touchMove or touchEnd is rejected with InvalidParams.

Example: Playwright `page.touchscreen.tap(400, 3000)` on a short page throws. Chrome completes the sequence silently. Either record the contact with a fallback target (documentElement, like the WebDriver pointerMove fallback) or make touchEnd and touchCancel a no-op when there is no active contact.

src/browser/webapi/WebDriver.zig
```diff
@@ -301,11 +301,19 @@ fn performPointerSource(source: js.Object, frame: *Frame) !void {
...
+                // No matching pointerDown in this source (a bare pointerUp,
+                // or a second pointerUp after the first already consumed the
+                // contact): nothing is pressed to release. Dispatching
+                // pointerup/touchend here would fabricate events for a
+                // gesture that never started.
+                continue;
+            }
             pressed = false;
             pressed_mask = 0;
             dispatchPointer(el, "pointerup", button, 0, frame);
```
arrufat, line 314: For a touch source, touchmove and touchend are now pinned to `down_target`, but pointermove (line 265) and pointerup here still go to the hover element `el`. Pointer Events use implicit capture for touch, so both streams should target the pointerdown element.

With pointerMove(a), pointerDown, pointerMove(b), pointerUp, a handler listening for pointerup on `a` never sees the release. Use `down_target` in both places when `is_touch and pressed`.

src/browser/webapi/WebDriver.zig
```diff
@@ -301,11 +301,19 @@ fn performPointerSource(source: js.Object, frame: *Frame) !void {
         } else if (action_type.eql(comptime .wrap("pointerUp"))) {
             const el = target orelse continue;
             const button = readI32(action, "button", 0);
+            if (is_touch and down_target == null) {
```
arrufat, line 304: This guard is limited to `is_touch`, but the WebDriver rule it implements (releasing a button that is not pressed is a no-op) applies to every pointer type. The mouse path still fabricates pointerup, mouseup, click and dblclick for a bare or duplicate pointerUp, and increments `click_count` into a dblclick.

Moving `if (down_target == null) continue;` above the `is_touch` branch gives both sources the same behavior from one place.

src/browser/webapi/event/Touch.zig
```diff
@@ -0,0 +1,77 @@
...
+    pub const Meta = struct {
+        pub const name = "Touch";
+        pub const prototype_chain = bridge.prototypeChain();
+        pub var class_id: bridge.ClassId = undefined;
+    };
+
+    pub const identifier = bridge.accessor(Touch.getIdentifier, null, .{});
+    pub const target = bridge.accessor(Touch.getTarget, null, .{});
+    pub const clientX = bridge.accessor(Touch.getClientX, null, .{});
+    pub const clientY = bridge.accessor(Touch.getClientY, null, .{});
```
arrufat, line 76: Touch only exposes identifier, target, clientX and clientY. pageX, pageY, screenX, screenY, radiusX, radiusY, rotationAngle and force are undefined. Most real swipe and slider handlers read pageX and pageY, so `e.touches[0].pageX - startX` is NaN and the gesture never fires, with no error to point at.

With no scroll offset in this engine, pageX/pageY/screenX/screenY can alias clientX/clientY. radius, rotationAngle and force can default to 1, 0 and 1 like Chrome does for a synthetic point.

src/browser/frame/user_input.zig
```diff
@@ -504,6 +505,80 @@ fn deltaToScroll(d: f64) i32 {
...
 }
 
+/// The CDP-tracked touch contact: single-touch scope, identifier 0.
+pub const TouchContact = struct {
+    target: *Element,
+    x: f64,
+    y: f64,
+};
+
+fn isTouchLift(typ: []const u8) bool {
```
arrufat, line 515: The touch type is a runtime string that gets re-classified with three `std.mem.eql` calls per dispatch (here, the touchcancel check below, and the touchstart check in `triggerTouch`). A typo like `"touchEnd"` compiles and silently produces an active, cancelable event with the wrong name.

An enum `{ touchstart, touchmove, touchend, touchcancel }` with derived `active` and `cancelable` would remove the string compares. The CDP switch already has an enum to map from.

src/server/cdp/domains/input.zig
```diff
@@ -1009,6 +1077,463 @@ test "cdp.input: chorded mousedown focuses its target unless pointerdown or mous
...
+
+    try ctx.processMessage(.{
+        .id = 2,
+        .method = "Input.dispatchTouchEvent",
+        .params = .{ .type = "touchEnd", .touchPoints = &.{.{ .x = rect_x, .y = rect_y }} },
+    });
+    try ctx.expectSentError(-31998, "InvalidParams", .{ .id = 2 });
+}
+
+// Regression coverage for the 2026-09-17 touch-input audit's four reproduced
```
arrufat, line 1407: Neither `reviews/2026-09-17-touch-input-audit.md` nor `reviews/touch-audit-probes.zig` exists in this branch or on main, so the reference leads nowhere. A one-line why per test, or nothing, is better.

Also, these four tests are labelled `cdp.input:` but call `dispatchTouchEventOn` directly, bypassing CDP.

src/browser/webapi/event/TouchEvent.zig
```diff
@@ -50,6 +64,13 @@ pub const Options = Event.inheritOptions(
...
 );
 
+pub const TouchInit = struct {
+    identifier: i32 = 0,
+    target: *Element,
+    clientX: f64,
+    clientY: f64,
+};
+
 pub fn init(typ: []const u8, _opts: ?Options, frame: *Frame) !*TouchEvent {
```
arrufat, line 74: The JS constructor still ignores the `touches`, `targetTouches` and `changedTouches` init members, and Touch has no constructor. So `new Touch({...})` throws and `new TouchEvent('touchstart', {touches: [t]})` has `touches.length === 0`, even though real Touch and TouchList types now exist.

Libraries that feature-detect with `try { new Touch(...) }` conclude touch is unsupported while CDP-driven ontouchstart handlers fire fine. Fine as a follow-up, but worth noting.

src/browser/webapi/event/TouchEvent.zig
```diff
@@ -22,12 +22,21 @@ const js = @import("../../js/js.zig");
...
 const UIEvent = @import("UIEvent.zig");
+const Touch = @import("Touch.zig");
+const TouchList = @import("TouchList.zig");
 
 const String = lp.String;
 
 // https://w3c.github.io/touch-events/#touchevent-interface
-// There is no touch input source: the touch lists are always empty.
+//
+// Single-touch: at most one contact. touchend/touchcancel drop it from
```
arrufat, line 35: Nit: this header, and similar blocks in Touch.zig (26-35), TouchList.zig (26-31), `touchList` (133-137), `initTrustedWithTouch` (82-86) and input.zig (159-165), describe the structure of the code rather than a non-obvious why. Once the single-touch model changes (for the id issue above, for example) each of them has to be rewritten or it becomes wrong. Keeping only the why would make that cheaper.

arrufat (commented):
Review of the touch input changes. The first two comments on input.zig block every Puppeteer touch interaction, the rest are should-fix and cleanup.

R4m1r0qu41:
Thanks for the thorough review. Everything below is on the branch, and I re-ran the tests and real clients against it.

**Your comments (reviewed at `3ba849b7d`)**

- **1 (`touchEnd` with the released point):** `touchEnd` now accepts zero or one point. Puppeteer's released point lifts the contact, and Playwright's empty list still works. A following `touchStart` succeeds. More than one point is still `InvalidParams`.
- **2 (Puppeteer ids start at 1):** the contact now stores the id from `touchStart`. Later points must carry the same id, or they get `InvalidParams` and the contact is unchanged. `Touch.identifier` reports it, and an omitted id defaults to 0.
- **3 (`touchStart` that misses every element):** it falls back to the document element, the same as WebDriver's `pointerMove`. The contact is no longer dropped, so the following end is valid.
- **4 (touch pointer capture):** pressed `pointermove` and the `pointerup` now stay on the `pointerdown` element. The test asserts `pointermove` and `pointerup` on `a` and none on `b` while pressed.
- **5 (unmatched `pointerUp`):** the no-op rule now applies to mouse sources too. There are tests for a bare `pointerUp` and for a duplicate `pointerUp`. Both dispatch no extra `click` or `dblclick`.
- **6 (`Touch` fields):** added `pageX/pageY/screenX/screenY/radiusX/radiusY/rotationAngle/force`.
- **7 (enum):** touch dispatch goes through a `TouchType` enum.
- **8:** I removed the dead reference and added a why-comment. The four direct-dispatch tests keep their `cdp.input:` names; I can rename them if you'd like.
- **9 (`Touch`/`TouchEvent` constructors):** left as a follow-up, as you suggested.
- **10 (structure comments):** trimmed the comments in `TouchEvent.zig`, `Touch.zig`, `TouchList.zig` and the CDP touch header in `input.zig` to the non-obvious "why" only. I also switched the comments above the touch functions to `///`, per your note on the click-dispatch PR. Comments only, no code changes.

**Also on the branch since your review**

- **Populated `touchCancel` is rejected.** The protocol says `TouchEnd`/`TouchCancel` carry no touch points, and Chrome rejects a populated cancel while accepting Puppeteer's populated `touchEnd`. The rejection happens before the active contact changes. A test asserts the contact stays active and that a later empty cancel releases it.
- **Modifier keys are preserved.** CDP touch events now carry the client's `modifiers`, and WebDriver touch forwards its modifier state. Omitted modifiers reset to false.
- **WebDriver touch pointer events report `pointerType: "touch"`.** They were hard-coded to `"mouse"`, which was pre-existing. Mouse sources still report `"mouse"`.
- I also removed two working-note files that had been committed to the branch. Net diff is `src/` only.

**Verification**

- `ZIGFLAGS='-Dwpt_extensions' make test`: 1589 passed, 5 skipped. `zig fmt --check` and `make build-dev` pass.
- To go beyond the unit tests, I also ran a set of evals: the same Puppeteer and Playwright scenarios against Lightpanda and Chrome. They cover taps, `TouchHandle` start/move/end, populated and empty `touchEnd`/`touchCancel`, target retention across movement, and recovery after a tap outside the page. All contract checks pass on Lightpanda.

**Known gaps, not addressed in this PR**

- CDP-driven touch dispatches TouchEvents only. Chrome also emits PointerEvents (`pointerdown`/`up`) and compatibility mouse/click events for the same tap. Adding them means designing the CDP touch-to-PointerEvent path, so I'd like to do that as a separate PR.
- Playwright's `page.tap('#a')` times out. I checked it against the branch base (without this branch) and it times out the same way, and `page.click` stalls too, so it's pre-existing. It stalls at Playwright's "visible, enabled and stable" check, before any input is sent.

src/browser/frame/user_input.zig
```diff
@@ -504,6 +505,113 @@ fn deltaToScroll(d: f64) i32 {
...
+}
+
+/// CDP TouchEnd/TouchCancel. Playwright sends an empty touchPoints list, so
+/// there's nowhere to read a release position from but the stored contact.
+/// Puppeteer sends the point being released, and Chrome dispatches at that
+/// wire position rather than the last-seen one, so `point` (when given)
+/// wins over the stored coordinates. Target and identifier always come from
+/// the stored contact: the lift still fires on the touchstart element, not
+/// a re-hit-test, and CDP has already checked the id matches.
+pub const TouchPoint = struct { x: f64, y: f64 };
```
arrufat: The `///` conversion left this doc comment on the wrong declaration. It describes `triggerTouchLift`, but `pub const TouchPoint` now sits between the comment and the function, and Zig attaches a doc comment to whatever declaration follows it. So this text documents the coordinate struct.

Moving `pub const TouchPoint` above the comment block fixes it.

src/server/cdp/domains/input.zig
```diff
@@ -153,6 +156,108 @@ fn dispatchMouseEvent(cmd: *CDP.Command) !void {
...
+}
+
+/// CDP's id is a wire f64; Touch.identifier is the DOM's i32. Reject anything
+/// that isn't a plain, in-range integer rather than truncating silently
+/// (@intFromFloat is illegal behavior on NaN/out-of-range in Zig, and this
+/// value comes straight off the wire).
+fn touchPointId(id: f64) ?i32 {
+    if (!std.math.isFinite(id) or id != @round(id)) return null;
+    if (id < std.math.minInt(i32) or id > std.math.maxInt(i32)) return null;
+    return @intFromFloat(id);
```
arrufat: Nit, your call: `deltaToScroll` in `user_input.zig` lets `@trunc` coerce straight to its `i32` return type instead of going through `@intFromFloat`. Same thing works here with `@round` if you want to match it.

arrufat (commented):
Thanks, this all checks out. I read each fix against the code and ran the suite locally: 1588/1588 on `make test` and 1594/1594 with `-Dwpt_extensions`, plus the six WebDriver touch tests by name. `zig fmt --check` is clean.

The two blocking ones are properly fixed. touchEnd taking zero or one point and the client-chosen id both match what Puppeteer sends and what Chrome validates. Keeping the `cdp.input:` names on the four direct-dispatch tests is fine now that the comment says why they bypass CDP. No need to rename them.

One small thing from the last commit, inline. Leaving the constructors to a follow-up works for me.

On the known gaps: agreed that the PointerEvent and compat mouse path is its own PR. Could you note it in the PR description or open a tracking issue? Without those events a Puppeteer `page.tap()` fires touch listeners but never clicks, and that is easy to mistake for tap being done.

src/browser/frame/user_input.zig
R4m1r0qu41: Good catch, sorry about that. Moved `pub const TouchPoint` above the doc block so the comment attaches to `triggerTouchLift`. I also found the same mistake on `TouchType` (the single-contact note was sitting on it) and moved that onto `TouchContact`. Fixed in 5450238cc.

src/server/cdp/domains/input.zig
R4m1r0qu41: Took it. `touchPointId` now returns `@round(id)` directly, matching `deltaToScroll`. Also reworded the comment above it so it no longer names `@intFromFloat`. Fixed in 5450238cc.

arrufat:
I pushed two small commits onto the branch instead of asking for another round, since both are mechanical. They are independent, so drop either one if you disagree.

**843340522** replaces `touchPointId` with an `i32` field on the params struct. The JSON parser already rejects a fractional or out of range number for an integer field, so the hand written version was doing the parser's job. It also had to order its NaN, rounding and range checks correctly ahead of a conversion that is illegal behaviour on either, which is a hazard worth removing rather than just lines. Same error on the same inputs, including the `0.5` case your test covers.

Three smaller ones rode along in that commit:

- `pressed` in `performPointerSource` moved in lockstep with `down_target` and was read once, where it guarded an `orelse continue` that could never be taken but read like a real failure mode.
- `Touch`'s four constant attributes now go through `bridge.property`, the idiom `Screen` and `Document` already use.
- `TouchList.indexedGet` and `getIndexes` go through `item()` and `length()` instead of repeating the index guard and the length expression, like `FileList` and `NodeList`.
- The CDP modifier bitmask is decoded in one place now, shared with `dispatchKeyEvent`, which spelled the same decode differently.

**8fdf07766** trims comments that restate the statement below them, list call sites, or repeat something said a few lines up. One duplicate paragraph above the touchEnd test went too, since it described an earlier revision. The Chrome and Puppeteer rationale, the spec reason `Touch.target` is an `EventTarget`, and the ordering constraint on clearing the contact before a fallible dispatch all stayed.

One gap worth naming before merge: nobody has run Puppeteer or Playwright against the id change, and that is the one that alters the parse path. A real client tap would be worth more than the unit coverage here.

R4m1r0qu41:
Thanks, I'm keeping both commits (`843340522`, `8fdf07766`).

I ran them at `8fdf07766`:

- `zig fmt --check`, `make build-dev` and `ZIGFLAGS='-Dwpt_extensions' make test` pass (1589 passed, 5 skipped).
- I ran the real-client suite (Puppeteer and Playwright against this build and local Chrome) to cover the id parse path you flagged. The touch scenarios that send ids pass on Lightpanda: `touchStart`/`touchMove`/`touchEnd` through `TouchHandle`, populated and empty `touchEnd`, empty `touchCancel`, target retention, and recovery after a tap outside the page. Populated `touchCancel` is rejected on both Lightpanda and Chrome, as intended.
- The only failures are the two gaps already in the description: missing PointerEvents and compat mouse/click for CDP touch (#3584), and Playwright `page.tap('#a')` timing out, which predates this branch.
- The suite has no fractional or out-of-range id scenario. The `0.5` case is covered by the unit test.

arrufat:
I ran this branch against Chrome 153 over CDP, with both Puppeteer and Playwright, and recorded what Chrome does as `puppeteer/touch.js` in lightpanda-io/demo (branch `touch-chrome-check`). One thing came out of it worth fixing here, pushed as b2ed099f8.

`Input.dispatchTouchEvent` accepted `radiusX`, `radiusY`, `rotationAngle` and `force` and then dropped them, so `Touch` always reported a 1x1 circle at full force. Puppeteer sends 0.5 for both radii and the force on every tap, and Chrome hands those straight back to the page, so anything reading pressure got a plausible looking wrong number with nothing to point at. The four now travel from the wire to the `Touch` with Chrome's defaults, including `radiusY` following `radiusX` when omitted rather than falling back to 1. They ride on a `TouchPoint` struct next to the coordinates and the id, which also takes `dispatchTouchEventOn` and `triggerTouch` from seven arguments to five.

Three differences I deliberately did not change, though I did reword the comments that read as if Chrome shares the rules:

- Chrome accepts a `touchEnd` carrying more than one point, releasing the contact it knows about and ignoring the rest.
- An id it has not seen opens a second contact whatever `type` says, since it reads `touchPoints` as the set of contacts currently down.
- A fractional id is not rejected. It truncates, so 0.5 arrives as 0, and a value past the range wraps to INT_MIN and hands the page garbage.

Refusing all three is the honest answer while only one contact exists, and on the last one we are better than Chrome. They are our rules rather than Chrome's, which is now what the comments say.

What the recorded test still flags is all #3584: a tap here is `touchstart, touchend`, where Chrome fires eleven events ending in a `click`, plus the enter and leave pairs.

One finding that is not about touch, for whoever wants it. The fixture div measures 5x5 at the origin here against 200x120 in place in Chrome, so a raw coordinate tap lands on `body` instead of the div, and Playwright's `page.tap` times out in its actionability check before sending anything. `touchscreen.tap` works. That is why the recorded test drives taps through `page.tap`, which resolves the element itself.

R4m1r0qu41:
Rechecked `b2ed099f8` (fmt, `make build-dev`, `make test` 1590/1590, and the Puppeteer/Playwright suite; nothing new fails). One correctness point on the comments and test, not a merge blocker.

`radiusY: ?f64 = null, // defaults to radiusX, as in Chrome` (and `wire.radiusY orelse wire.radiusX`) doesn't match Chrome 153.0.8010.52. Raw CDP: `radiusX: 3` alone reports `radiusX=3, radiusY=1`, and `radiusY: 4` alone reports `1, 4`. The protocol also documents `radiusY (default: 1.0)` independently of `radiusX`. For the shape the unit test uses (populated `touchEnd` with `radiusX: 3, rotationAngle: 45`), Chrome gives `[3, 1, 45, 1]`, while the test expects `[3, 3, 45, 1]`. The suite passes because the test asserts our behavior.

An explicit `0` on either radius is kept, and the other still defaults to 1 (an independent second probe confirmed this).

No real client hits this, since Puppeteer sends both radii. Two options: default `radiusY` to 1 and update the test to `[3, 1, 45, 1]` (ideally with a `radiusY`-only case too), or keep the fallback and reword the comment and test as our rule, like your other three. Your call.

PR #3570, opened 2026-09-18
orderfile: regenerate the profile from build.zig, tools in Zig (experiment)

arrufat:
An experiment, opened for discussion: close it, or cherry-pick the parts that seem worth it. Each commit stands on its own.

## What

`zig build orderfile` (or `make orderfile`) regenerates `orderfile/lightpanda.ld` and `v8.txt` from the build system instead of `orderfile/tools/regen.sh`, and the python profile tools become Zig.

- **dc62d402a** `zig build orderfile`: builds the unordered exe, runs the bench (`orderfile/profile.sh`, still shell: root for `fault_around_bytes`, the demo server, node), generates the script from the link inputs the build graph knows, rebuilds with it to check it links, writes the two files back and the profile into `zig-out/orderfile/`. The nightly action calls it. Isocline is built as a static library like the other C deps so its object is a graph artifact.
- **115d2bd09** python -> Zig: `gen_order.py`, `hotlist.py` and `pagemap.py` become `gen_order.zig` and `hotlist.zig` on an `elf.zig` shared with `mark_hot_sections.zig`. No python3 or binutils in the job.
- **7d43c875b** a profile regenerated with it (same as what the nightly would produce; drop it if the nightly lands first).
- **72725aebb** cleanup: `elf.zig` on `std.elf`, inputs mmapped, `std.sort` bounds, build.zig back to main's shape with `sectionize` in `orderfile/orderfile.zig`.

## Assessment

The step is an ergonomics win, not an engineering one. It replaces grepping the `ld.lld` line for the link inputs with the build graph, at the cost of a second full compile for the link check (regen.sh replayed the captured link line in seconds) and two assumptions about Zig's cache layout: the exe's object sits at `dirname(exe)/lightpanda_zcu.o`, and the libc++/libunwind/compiler_rt archives are found by globbing `~/.cache/zig/o/*/`. A missing object fails loudly; missing archives would degrade silently (`nomap` grows), which the generator should probably error on.

The port is justified by one finding rather than by speed: parsing `nm` output by whitespace dropped every Zig symbol whose name contains a space (generic instantiations) and every sizeless symbol, about 70 symbols per profile. Reading the symbol table directly fixes that and consolidates the ELF parsing that `mark_hot_sections.zig` already had. It is about four times the lines of the python.

The regenerated profile is functionally equivalent to the python's output (byte-identical script body on the same inputs), so the e2e VmHWM gate should not move.

An `addObject` + link split of the exe was tried first to get the object as a graph artifact and dropped: Zig relocatable-links every library of the module graph into the object.

## Trying it

```
sudo -v && make orderfile
```

Needs root, `../demo` with `npm install` done, node and go. About 15 minutes: two compiles and 100 bench runs.


karlseguin:
I'll take the space fix (https://github.com/lightpanda-io/browser/pull/3593).

To me, this is code that's maintained by an Claude. Whether it's in Zig (more lines of code) or in Python (less lines of code), doesn't change anything for me. The current version gets its link output from `--verbose-link` which seems like something worth keeping (and something I assume the zig version could use just as easily). Would also want "which the generator should probably error on" implemented.

Issue #1839, opened 2026-03-15
[Bug] CDP session management causes assertion error in Playwright integration


ai-nurmamat:
## Summary
Lightpanda Browser crashes with assertion error when Playwright sends certain CDP commands for session management. This makes Lightpanda incompatible with Playwright's default behavior.

## Steps to Reproduce
1. Start Lightpanda: `./lightpanda serve --port 9222`
2. Run a Playwright script that creates multiple contexts or uses session management

```javascript
const { chromium } = require('playwright');
const browser = await chromium.launch({
  channel: 'chrome',
  launchType: 'remote',
  wsEndpoint: 'ws://127.0.0.1:9222'
});
const context = await browser.newContext();
const page = await context.newPage();
await page.goto('https://example.com');
```

## Expected Behavior
Lightpanda should handle CDP session management commands gracefully without crashing.

## Actual Behavior
The browser crashes with an assertion error in CRSession._onMessage:

```
Assertion failed: invalid session id
```

## Root Cause Analysis
The issue appears to be in how Lightpanda handles CDP session lifecycle commands. When Playwright sends multiple rapid session-related CDP commands:

1. `Target.createBrowserContext` - creates a new browser context
2. `Target.createTarget` - creates a new target within that context
3. `Target.attachToTarget` - attaches to the target with a session ID
4. Various Page/Runtime commands with the session ID
5. `Target.detachFromTarget` - detaches the session

The assertion fails when Lightpanda tries to route messages to a session that has been detached, or when the session ID becomes invalid.

**Code Location**: Likely in `src/cdp/session.zig` or `src/cdp/domains/target.zig` where session management happens.

## Proposed Solution
1. Add proper session state management - don't assert on invalid session IDs, instead return a proper CDP error response
2. Track session lifecycle more carefully - when a session is detached, either:
   - Return a proper error like `{ "error": { "code": -32602, "message": "Invalid session ID" }}`
   - Or automatically clean up and route messages to the default browser context

Example fix in session routing:
```zig
fn routeMessage(msg: *const CDPMessage) !void {
    const sessionId = msg.sessionId;
    if (sessionId == null) {
        // Route to default context
        return routeToDefaultContext(msg);
    }
    
    const session = sessions.get(sessionId);
    if (session == null) {
        // Don't crash - return proper error instead
        return sendErrorResponse(msg.id, -32602, "Invalid session ID");
    }
    
    return session.send(msg);
}
```

## Impact
- Severity: High
[...]

Issue #1838, opened 2026-03-15
CDP: Playwright crashes with assertion error in CRSession._onMessage


sutidor:
## Summary

When connecting to LightPanda via CDP (port 9222) using Playwright, the Playwright CDP client crashes with an unhandled assertion error in `CRSession._onMessage`. The page is fetched successfully (confirmed via webhook.site HAR), but the CDP response message shape doesn't match what Playwright expects, causing the calling process to crash.

## Environment

- **LightPanda**: `lightpanda/browser:nightly` (built 2026-03-14, sha256:6d37e5f3bd1c4)
- **Playwright**: 1.58.2 (via Node.js application connecting to LightPanda's CDP endpoint)
- **Node.js**: 22.22.0
- **OS**: Linux 6.8.0-101-generic (Docker container)

## Steps to Reproduce

1. Run LightPanda exposing CDP:
   ```bash
   docker run -d --name lightpanda -p 9222:9222 lightpanda/browser:nightly
   ```

2. Connect Playwright to the CDP endpoint and navigate to any page:
   ```js
   const browser = await chromium.connectOverCDP('http://127.0.0.1:9222');
   const page = await browser.newPage();
   await page.goto('https://webhook.site/some-uuid');
   ```

3. Playwright crashes with an unhandled assertion error.

## Error

```
Unhandled promise rejection: Error: Assertion error
    at assert (playwright-core/lib/utils/isomorphic/assert.js:26:11)
    at CRSession._onMessage (playwright-core/lib/server/chromium/crConnection.js:129:31)
    at CRConnection._onMessage (playwright-core/lib/server/chromium/crConnection.js:67:15)
    at Immediate.<anonymous> (playwright-core/lib/server/transport.js:73:28)
```

## Analysis

The assertion is at [`crConnection.js:129`](https://github.com/nicknisi/playwright/blob/1ccfc16db0b1c3a277c5a5c tried/packages/playwright-core/src/server/chromium/crConnection.ts#L137), which checks:

```ts
assert(!object.id);
```

This assertion expects that CDP messages without an `id` field are events (method notifications), and messages with an `id` are command responses. It appears LightPanda is sending a CDP message that has an `id` but is being routed to a session that doesn't expect it, or is sending a response with unexpected fields.

## Expected Behavior

[...]

PR #3571, opened 2026-09-18
`cdp`: answer inspector commands on the session that sent them

nikneym:
Playwright's `browserContext.newCDPSession` must get its response on that session, not the primary one. Playwright keys its pending callbacks by session, finds none for that id, and throws "Assertion error", which takes the whole process down.

Addresses #1838 and #1839.


karlseguin:
I tried both issues on main, and they both work as-is. I do think there's a bug here, but I'm not sure this is the right fix. In large part because it doesn't (and can't) fix inspector events - they still get routed to the bc's session_id.

The proper solution seems to be to expand `AttachedSession` with a `*js.Inspector.Session` and create / cleanup the Inspector.Session on attach/detach. On its own, that seems pretty simple, the issue is that any RemoteObjectId needs to use the correct Inspector.Session. So I'd remove BrowserContext.inspector_session, store that as an `AttachedSession` (i.e. don't make the 'default' a special case), and then see what happens. For example, `dom.getNode
getNode` will break, since it needs the `inspector_session`, so it would need to either be passed the `inspector_session`, or at least be passed the `session_id` so that it can look it up in `bc.attached_sessions` (which, at this point, should probably be moved to a map).

Issue #1550, opened 2026-02-14
Creating a context with storage state fails


tmdh:
It waits for a minute then crashes. Error only happens when setting `origins` property in `storageState`.

## Minimal reproduction
```ts
import { chromium } from "playwright-core";

async function main() {
  const browser = await chromium.connectOverCDP("ws://127.0.0.1:9222");
  const context = await browser.newContext({
    storageState: {
      cookies: [],
      origins: [
        {
          origin: "https://example.com",
          localStorage: [
            {
              "name": "example_timestamp",
              "value": "1770479250644"
            },
          ]
        }
      ]
    }
  })
}

main().catch(console.error);

```

## Console output
```
[TargetClosedError: Error setting storage state:
Target page, context or browser has been closed
Call log:
  - navigating to "https://example.com", waiting until "load"
]
```

## Browser output
```
$time=1771105781504 $scope=app $level=info $msg="client connected" ip=127.0.0.1:9222
$time=1771105781538 $scope=not_implemented $level=warn $msg=Target.createBrowserContext disposeOnDetach=true has_proxyBypassList=false has_originsWithUniversalNetworkAccess=false
$time=1771105781542 $scope=cdp $level=debug $msg="inspector event" method="\"Runtime.executionContextCreated\""
$time=1771105781548 $scope=page $level=info $msg=navigate url=https://example.com method=GET reason=address_bar body=false req_id=2
$time=1771105781548 $scope=cdp $level=debug $msg="request intercept" state=paused id=2 url=https://example.com
$time=1771105781551 $scope=cdp $level=debug $msg="request intercept" state=fulfilled id=2 url=https://example.com status=200 body=true
$time=1771105781551 $scope=cdp $level=debug $msg="inspector event" method="\"Runtime.executionContextCreated\""
$time=1771105781551 $scope=cdp $level=debug $msg="inspector event" method="\"Runtime.executionContextCreated\""
$time=1771106098133 $scope=app $level=info $msg="CDP timeout"
$time=1771106098133 $scope=cdp $level=debug $msg="inspector event" method="\"Runtime.executionContextsCleared\""
```

PR #3572, opened 2026-09-18
`cdp`: move a kept isolated-world context onto the navigated origin

nikneym:


karlseguin:
#1550 hangs on main and in this PR. It hangs because it doesn't close the page and browser.

src/server/cdp/domains/page.zig
```diff
@@ -785,6 +785,9 @@ pub fn frameNavigated(arena: Allocator, bc: *CDP.BrowserContext, event: *const N
             if (!is_root_frame) {
                 continue;
             }
+            js_context.setOrigin(frame.origin) catch |err| {
```
karlseguin, line 788: Should this not be done for iframes too ?

nikneym: Right, I've changed the ordering.

nikneym:
> #1550 hangs on main and in this PR. It hangs because it doesn't close the page and browser.

Indeed, this made bookkeeping right on Zig-side but issue persists; will further investigate.

PR #3583, opened 2026-09-21
http: improve caching

karlseguin:
1 - Centralized cache-awareness into Cache and pulled header details out of
    SqliteCache and HttpClient

2 - Added support for expires header

3 - Support caching more status types (but not all, since HttpClient would need
    to be aware of what caching a 3xx/206 means)

4 - Revalidate cares about  "not specified" vs "no-store" vs "stale"
    (e.g. expires=0 means "stale", not fallthrough the last-modified logic)


src/network/cache/Cache.zig
```diff
@@ -281,197 +299,419 @@ pub const CachedResponse = struct {
...
+
+const testing = @import("../../testing.zig");
+test "Cache: ResponseDirectives.parse" {
+    try testing.expectEqual(300, ResponseDirectives.parse("max-age=300").max_age);
+    try testing.expectEqual(300, ResponseDirectives.parse("Max-Age=300").max_age);
+    try testing.expectEqual(300, ResponseDirectives.parse("MAX-AGE=300").max_age);
+    try testing.expectEqual(300, ResponseDirectives.parse("public, max-age=300").max_age);
+    try testing.expectEqual(300, ResponseDirectives.parse("  max-age=300  ").max_age);
+
+    // we are a private cache: s-maxage is for shared caches only
```
krichprollsch: Can't we use safely s-max-age if max-age isn't set?

krichprollsch (approved):

src/network/cache/Cache.zig
karlseguin: I went with `@min(s-maxage, maxage)`. In some cases that'll be worse, but when there's only an s-maxage it'll be better. We _do_ act more like a shared cache, so I think letting s-maxage lower the cache duration is right.

krichprollsch (approved):

PR #3587, opened 2026-09-21
Changes for `<template>` content

nikneym:
Template content belongs to an inert "template contents owner document" with no browsing context, related changes for #3328.


src/browser/webapi/CustomElementRegistry.zig
```diff
@@ -186,6 +186,10 @@ fn upgradeNode(self: *CustomElementRegistry, node: *Node, frame: *Frame) !void {
 }
 
 fn upgradeElement(self: *CustomElementRegistry, element: *Element, frame: *Frame) !void {
+    if (element.asNode().getDocument(frame)._frame == null) {
```
karlseguin, line 189: This is only called from `upgradeNode` which only walks children. It never crosses a document. If the root has a frame, than all children have a frame. So, this check can be done once, in `upgrade`.

src/browser/parser/Parser.zig
```diff
@@ -736,11 +762,66 @@ fn _appendCallback(self: *Parser, parent: *Node, node_or_text: h5e.NodeOrText) !
...
+    if (child._parent != null) {
+        child_pn.placed = true;
+    }
+
+    var document = parent.getDocument(frame);
+    const child_document = child.getDocument(frame);
+    if (child_document != document) {
+        if (child_pn.placed == false) {
+            child_pn.node = try self.rebuildIn(child, document);
+        } else if (parent_pn.placed == false) {
```
karlseguin, line 790: How does a child get placed but not its parent? Doesn't _reparentChildrenCallback (which places the parent) ensure this never happens

karlseguin (commented):
Looks good, just a couple small things.

src/browser/webapi/CustomElementRegistry.zig
nikneym: Done, also added a `customElements.upgrade(template.content)` test to cover it.

src/browser/parser/Parser.zig
nikneym: html5ever removes parent's last node and appends the node to a fresh copy; the copied element hasn't been placed, and the node is the furthest block (already placed). `_reparentChildrenCallback` only covers the later step where the furthest block's children move into the formatting element's clone. Added a test to make sure misadoption not happening.

PR #3588, opened 2026-09-21
Clamp scroll offsets, and latch a wheel to one scroll container

arrufat:
Two gaps left from #3523, found while answering "if the inner container is fully scrolled, do we scroll the outer one?". The answer was no, and the reason was a clamp that did not exist.

### Scroll offsets clamp to the scrollable extent

Every scroll write clamped at zero and nothing else, so an offset could exceed the scrollable extent without limit: a page probing `scrollTop >= scrollHeight - clientHeight` — the usual "am I at the bottom, load more?" test — got a number Chrome would never produce.

`Element.scrollExtent` is that bound, and `setScrollTop` / `setScrollLeft` / `scrollTo` / `scrollBy` now share one writer that applies it (four copies of the same `getOrPut` + compare + schedule dance collapse into it).

The extent is **optional, and null means unbounded** — the write behaves exactly as it did before. Without a layout engine there is no honest box otherwise: `getElementAxis` reads only inline `width`/`height`, so an element sized by a stylesheet measures 5px and its `clientHeight` collapses onto the same content sum `scrollHeight` uses; and `contentAxis` sums element children only, so a panel holding text reports no content at all. Clamping those to a zero extent would make the stylesheet scrollers #3523 just enabled unscrollable. Refusing a scroll we cannot prove impossible is worse than allowing one too many. `html`/`body` are excluded outright — their artificial giant defaults (`scrollHeight` 100,000,000) would fabricate an extent against the real viewport.

### A wheel latches to one scroll container

`wheelScroll` handed the whole delta to the nearest container on each axis whatever state it was in, so a saturated inner scroller trapped the wheel and the page never moved.

`scrollAxis` now walks outward per axis and gives the whole delta to the first container that can still move along it. **A delta is never split across scrollers**: a container that can only take part of it keeps the rest, and the page moves on the *next* wheel. That is Chrome's rule — `InputHandler::FindNodeToLatch`, `cc/input/input_handler.cc` — and it matches what Chrome 153 actually does: one `Input.dispatchMouseEvent` wheel of 1000px over a container with 416px of travel leaves `window.scrollY === 0`.

A container whose `overscroll-behavior` does not propagate on that axis takes the latch even when it cannot move, which ends the walk (`CanPropagate`, same file). So `overscroll-behavior` is now tracked in the cascade, following the `overflow-x`/`overflow-y` pattern exactly: two flags in `Props` (`packed struct(u10)` now), an entry in the `axis_shorthands` table `Slots.apply` folds, `StyleManager.overscrollContainAxes` as the probe. `CssParser.splitOverflow` became `splitAxisPair` now that two shorthands share it.

### Checked against Chrome

Everything above was recorded from Chrome 153 over CDP before being implemented, with a companion script in lightpanda-io/demo#253 (`puppeteer/scroll.js`). It passes against Chrome; against lightpanda it fails exactly three checks, which are the gaps this deliberately keeps:

- an element sized by a stylesheet has no extent, so it never saturates;
- an element holding only text has none either;
- the viewport has none at all.

All three wait on the cascade tracking `width`/`height`, which would move every box metric (`getBoundingClientRect`, `offsetHeight`, `elementFromPoint`, IntersectionObserver) and is its own project.

### Tests

`element/position.html` asserts the limit relationally (`scrollHeight - clientHeight`) rather than a number of our own, so it holds in a real browser too — Chrome reserves scrollbar space in `clientHeight` and lands a few px lower. The gap we keep is pinned in a Zig test instead, and the CDP test covers the latch, the chain to the viewport, `overscroll-behavior: contain`, and reversing direction.


src/browser/webapi/Element.zig
```diff
@@ -2014,55 +2019,111 @@ pub const ScrollToOpts = union(enum) {
...
     const gop = try owner.page.element_scroll_positions.getOrPut(owner.page.frame_arena, self);
     if (!gop.found_existing) {
         gop.value_ptr.* = .{};
     }
     const old_x = gop.value_ptr.x;
     const old_y = gop.value_ptr.y;
-    gop.value_ptr.x = @intCast(@max(0, @as(i32, @intCast(gop.value_ptr.x)) +| (o.left orelse 0)));
-    gop.value_ptr.y = @intCast(@max(0, @as(i32, @intCast(gop.value_ptr.y)) +| (o.top orelse 0)));
+
+    if (write.target(.width, old_x)) |target| {
```
karlseguin: the `gop` is now held over a lot more code. It's probably all safe (e.g. none of it calls into JS), but...you could do it _after_ you calculate the `x` and `y`, reducing any future risk (and, you can even avoid creating a map entry in that case where it hasn't moved.

src/browser/StyleManager.zig
```diff
@@ -1277,6 +1303,12 @@ fn foldDeclarations(block: []const u8, customs: ?*CustomSink) !TrackedProperties
     return slots.props();
 }
 
+// The `<x> [<y>]` shorthands the cascade expands into the tracked longhands.
+const axis_shorthands = [_]struct { name: []const u8, x: []const u8, y: []const u8 }{
+    .{ .name = "overflow", .x = "overflow-x", .y = "overflow-y" },
+    .{ .name = "overscroll-behavior", .x = "overscroll-behavior-x", .y = "overscroll-behavior-y" },
```
karlseguin: CSSStyleDeclaration doesn't know about this, so a `put("overscroll-behavior", "...")`  will cause issues.

Would be nice if StyleManager and CSSStyleDeclaration could share the short -> long  lookup/handling.

src/browser/frame/user_input.zig
```diff
@@ -487,16 +487,48 @@ pub fn wheel(frame: *Frame, target: *Element, x: f64, y: f64, delta_x: f64, delt
...
+        .height => .{ .y = true },
+    };
+
+    var current: ?*Element = target;
+    while (current) |el| {
+        const container = switch (el.scrollContainer(axes, frame)) {
+            .container => |c| c,
+            .viewport => break,
+        };
+        if (container.canScrollAxis(axis, delta, frame)) {
```
karlseguin: Have writeScroll returns whether it wrote (combined with not creating an entry if nothing's changed). Have scrollByAxis return that bool, and then you can delete `canScrollAxis`:

```zig
if (container.scrollByAxis(axis, delta, frame)) {
  return`
}
```

As-is,you're recomputing a lot of things to decide if you can write. Reverse it, try to write and return whether or not you did.

src/browser/webapi/Element.zig
arrufat: Done in fdee7d5. It now clamps both axes against a plain `get` and only takes the entry once an offset actually moves, so the `gop` is held over two field writes and `scheduleScrollEvents`, and a write that changes nothing no longer creates one.

src/browser/frame/user_input.zig
arrufat: Done in fdee7d5, that's much better. `writeScroll` returns whether it moved, `scrollByAxis` passes that through, and the walk is now:

```zig
if (try container.scrollByAxis(axis, delta, frame)) {
    return;
}
```

`canScrollAxis` is gone, and with it a second `scrollExtent` computation per container (each one reads inline styles and sums the children).

src/browser/StyleManager.zig
arrufat: Good catch: `put("overscroll-behavior", ...)` did leave `overscrollBehaviorX` reading empty, and a `style=` block round-tripping through the object lost it.

Shared in 93c551b. `CssParser.axis_shorthands` is now the single list, with `axisShorthand` / `axisLonghand` as the lookups both sides use: StyleManager's fold and the CSSOM's five overflow-only special cases (apply, set, remove, priority, serialize) all go through it, and `OverflowPair` became `AxisPair` carrying the shorthand name. Adding a third one is a line in that table.

Checked the behaviour against Chrome 153 rather than guessing — `overscroll-behavior: contain auto` reads back per axis, collapses to `contain` when both match, and serializes as a single declaration; `element/styles.html` asserts exactly that.

PR #3591, opened 2026-09-22
http: dont' re-use connections which are likely in a bad state.

karlseguin:
Some status-codes should never have a body except for a single trailing blank line. If we don't handle these, then we end up with a dirty connection in our connection pool:

1 - read the header, but not the body
2 - put the connection back in the pool
3 - try to read the header, but actually get the body from #1

WPT /fetch/api/basic/response-null-body.any.html exercises this path and is flaky (because it depends whether the request goes back out on a keep-alive connection)..but for a given run,you'll almost always get 1-3 failures.

This commit processes the request, but tells libcurl not to re-use the connection.


krichprollsch:
Is closing the connection driven by the WPT test? I'm wondering if it's better than draining the body and keeping the conn available.

karlseguin:
> Is closing the connection driven by the WPT test? I'm wondering if it's better than draining the body and keeping the conn available.

The WPT sends a 204 with a body. The point of the test is to verify that even if there is a body, because of the 204, we return null. That's a distinct behavior then the keepalive..but I don't think we should, I also don't think we can. curl itself has these rules baked-in.  It sees the 204 and enters a bodyless mode, which skips things, like reading the body.

krichprollsch (approved):

